PR #1876, opened 2018-01-30
copy insertSession based on DatabaseHashMap

njr-11:
Start experimenting with use of JCache for CacheHashMap.  Add the insertSession method copying from DatabaseHashMap.  I expect all of this code will be rewritten, but this is just something to start with in order to get through the code flow.


njr-11:
#build

njr-11:
Build has 1 test failure, which is a pre-existing issue in unrelated componentry

PR #1894, opened 2018-01-30
Remove unused classloader map

andymc12:


andymc12:
will likely fail due to "Auth cancel" infrastructure issue, but will try just in case: #build

andymc12:
#build

PR #1918, opened 2018-01-31
copy additional methods and enable basic cache test

njr-11:
After copying a sufficient number of methods from DatabaseHashMap/Session to CacheHashMap/Session, it is possible to execute some basic code path covered by the com.ibm.ws.session.cache_fat bucket.  This pull also switches over the fat bucket to use the very early stage and experimental sessionCache-1.0 feature, which will give us a starting point towards putting a proper implementation in place.


njr-11:
#build

PR #1930, opened 2018-02-01
JSR375:remove CWWKS1913W

toshiyamamoto:
CWWKS1913W message is being removed. No ID review is required since it's just a deletion.


toshiyamamoto:
#build

PR #1935, opened 2018-02-01
Java 2 Security issues in batch-1.0 feature

scottkurz:
The **batch-1.0** feature does not even allow a simple job execution to complete.

This problem is resolved mostly by adding doPrivileged callls to the  **com.ibm.jbatch.container** and ** com.ibm.websphere.javaee.batch.1.0**  bundles.

However JAXB unmarshaling (of the job XML) does not succeed even with these additions.   I learned that the server.xml needs to configure the **jaxb-2.2** feature in order to avoid requiring additional permissions (instead of using JAXB from the JRE).

### DIFF
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/jsl/impl/JobModelSerializerImpl.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/jsl/impl/JobModelSerializerImpl.java
@@ -41,17 +41,7 @@ public class JobModelSerializerImpl implements ModelSerializer<JSLJob> {
 
     @Override
     public String serializeModel(JSLJob model) {
-
-        final JSLJob finalModel = model;
-        String serializedModel = AccessController.doPrivileged(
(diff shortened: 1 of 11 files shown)


scottkurz:
#build #spawn.fullfat.buckets=com.ibm.ws.jbatch.jms_fat,com.ibm.ws.jbatch.joblog_fat,com.ibm.ws.jbatch.rest_fat,com.ibm.ws.jbatch.tck_fat,com.ibm.ws.jbatch.utility_fat,com.ibm.ws.jbatch_fat

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/jsl/impl/JobModelSerializerImpl.java
```diff
@@ -41,17 +41,7 @@
 
     @Override
     public String serializeModel(JSLJob model) {
-
```
ajm01: so i guess we had this protected but it is no longer necessary to protect this section?

Issue #1806, opened 2018-01-25
The openliberty build should output a small kernel zip


ewhoch:
To allow for more customized installations, it would be helpful to have a zip that contained only the kernel code that could then have features installed on top. Right now we package an "all" zip that has everything, and a "ga" zip that only has kind=ga features, this would be a 3rd type that consisted of a server minified with no features provisioned.

ewhoch:
Initial hacking at this by just copying over the packageOpenLiberty task into a second packageOpenLibertyKernel task seems to work, but duplicates a lot of logic. Also the server.xml default template containing jsp-2.3 then no longer makes sense. Going to see if i can factor out that logic to keep the script a little cleaner.

PR #1939, opened 2018-02-01
Make package code more general, add kernel packaging step

ewhoch:
First pass attempt. makes a featureless zip but still has some issues since the server.xml default template still contains jsp-2.3 which won't be in that zip. for issue #1806


ewhoch:
#build

ewhoch:
#build

ewhoch:
#build

PR #1943, opened 2018-02-01
Json logging test

Yushan-Lin:


fmhwong:
@Yushan-Lin Is this PR still active?

Issue #1944, opened 2018-02-01
Websockets for non-secure BASIC_AUTH follow session invalidation


billw1024:
If an HTTP session underlying a given websocket connection has been configured to use  a non-secure connection with BASIC_AUTH type of authentication, then close the websocket session when the underlying http session is invalidated/closed

PR #1946, opened 2018-02-01
websockets - close session when insecure BASIC_AUTH http session closes. 

billw1024:
For a websocket connection which was started from an HTTP Session that is using a non-SSL port, but has BASIC_AUTH enabled, then close the websocket session with a 1008 error code when
the HTTP session is closed/invalidated.


billw1024:
#build

billw1024:
59,000+ tests passed, 3 failures .  failures unrelated to this websocket change.

pnicolucci:
fixes #1944

PR #1948, opened 2018-02-01
Avoid overwriting updates made to the cache by another thread

njr-11:
Replace the Object[] value with a proper comparable structure such that we can use cache.replace to avoid overwriting updates that were made by another thread.


njr-11:
#build

njr-11:
#build

njr-11:
#build

PR #1956, opened 2018-02-01
update 'jdk.' to 'jdk' in com.ibm.ws.kernel.boot.ServerClasspathTest

WhiteCat22:
While upgrading to JDK8 SR5 FP7, a new class was added to
the JDK under "jdk" that can be loaded by JDK's classpath. The ServerClasspathTest
checks for all known packages on the JDK classpath to ensure we're not
adding anything that we shouldn't - so the "jdk" package needs to be
added to the "expected list" in the test. We already have "jdk." so we must update it to remove the '.'


WhiteCat22:
Build: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_5hf-UAePEeiIX6lO4t6XVQ

WhiteCat22:
build fixed the test

PR #1957, opened 2018-02-02
Convert headers to strings after processing ClientRequestFilters

andymc12:
The JAX-RS 2.1 TCK is failing because CXF is not converting objects put into the HTTP headers in ClientRequestFilters.  The spec requires that headers are converted into Strings (via either a HeaderDelegate or calling the object's toString method) before it is sent over the wire (or returned via the ClientRequestContext's abortWith method).  CXF converts header objects to strings earlier in the flow, but does not convert objects added in ClientRequestFilters.  This change fixes that.


andymc12:
#build

PR #1872, opened 2018-01-30
Don't call copyFileTo... when file already exists

benjamin-confino:

PR #1972, opened 2018-02-02
Revert "Don't call copyFileTo... when file already exists"

wraschke:
Reverts OpenLiberty/open-liberty#1872


wraschke:
#link 250894

PR #1974, opened 2018-02-02
httpSessionCache use bell in absence of libraryRef

njr-11:
Default JCache CachingProvider should be loaded from a bell in the absence of a configured libraryRef on httpSessionCache.  Need to arrange declarative services dependency such that libraryRef is prefered, bell is backup, and exactly one must be supplied to the service component.  This requires a creative filter and an intermediate service that re-registers the library of the bell in a place where httpSessionCache can default to it.


njr-11:
#build

njr-11:
#build

PR #1975, opened 2018-02-02
Remove SessionManager instance when app is stopped.

andymc12:
This avoids leaking the SessionManager instance which references
classes that reference the classloader, leaking it.


andymc12:
#build

Issue #1993, opened 2018-02-05
Test Failure (20180202-1159): testActivationSpecXARecovery FFDC for EJBException not found


njr-11:
Test Failure (20180202-1159): com.ibm.ws.jca.fat.derbyra.DerbyResourceAdapterTest.testActivationSpecXARecovery

junit.framework.AssertionFailedError: 2018-02-02-18:15:48:353 [An FFDC reporting javax.ejb.EJBException was expected but none was found.]
at componenttest.custom.junit.runner.FATRunner.blowup(FATRunner.java:289)
at componenttest.custom.junit.runner.FATRunner.access$500(FATRunner.java:65)
at componenttest.custom.junit.runner.FATRunner$1.evaluate(FATRunner.java:239)
at componenttest.custom.junit.runner.FATRunner$2.evaluate(FATRunner.java:307)
at componenttest.custom.junit.runner.FATRunner.run(FATRunner.java:159)

The expected FFDC still shows up in the log, it's just asynchronous to the test case completion and so the ExpectedFFDC checking doesn't find it.  Test case logic can be adjusted to allow for this.

PR #1994, opened 2018-02-05
testActivationSpecXARecovery is not waiting for expected EJBException to appear

njr-11:
Update test case to wait for the expected FFDC message to appear before ending the test such that the ExpectedFFDC validation will be able to see it.  Note that switching to AllowedFFDC did not seem like a good option, because it would leave open the possibility that the FFDC could appear during the next test that runs, which would not be expecting it, and could lead to other intermittent failures.


njr-11:
pull fixes #1993

PR #2007, opened 2018-02-05
Use 'debug' j2sec mode by default locally, but still cause failures when issues occur

aguibert:


aguibert:
#build

aguibert:
#build

aguibert:
#build

aguibert:
There were 2 FAT failures in the personal build which appear to be unrelated to the changes being delivered.  The failure in micirprofile.config_fat is a match for 250966 and the failure in wim.adapter.ldap_fat is a match to 250325.

PR #2015, opened 2018-02-05
Add non-jta-datasource ref to test persistence unit so table creation…

jgrassel:
… does not rel on taking place within a JTA tran.


jgrassel:
#build

jgrassel:
Not sure why the JPA  22 AppClient tests failed in the FAT run.  I only touched the Bean Val test bucket's assets.  BVal tests passed, so going out on a limb that it's safe.

Issue #2013, opened 2018-02-05
Fix HPACK Decoder when decoding empty header values


mrsaldana:
A header does not necessarily need to have a value. The H2 HPACK decoder is throwing a decompression exception when it tries to decode a Huffman encoded empty value.

PR #2016, opened 2018-02-05
Initial commit, decoder will now ensure header names are non empty an…

mrsaldana:
…d accepts empty header values.


mrsaldana:
#build

mrsaldana:
Resolves #2013

PR #2029, opened 2018-02-06
Initial support for multi-server session caching

aguibert:


aguibert:
#build

dev/com.ibm.ws.session.cache_fat/test-applications/multiServerApp/resources/META-INF/permissions.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
njr-11: Is it really necessary to duplicate all of the application files?  Seems like we could have a single copy of the application and put it onto all of the servers.  I realize that in one case you want to include testing of a data source and in the other case not, but if that's the only difference, that could be done with 2 different servlet methods rather than duplicating everything.

aguibert:
Although the personal build has not quite finished (an unrelated FAT has been running for 10 hours and is probably never going to complete), I'm going to merge because the session.cache_fat has ran successfully in my build and at this point the sessionCache-1.0 feature can't impact any other features or FATs.

PR #2048, opened 2018-02-06
 copy handlePropertyHits and getAllValues to make progress on cache-entry-per-session-property path

njr-11:
Implement handlePropertyHits and getAllValues based on DatabaseHashMapMR so that a basic test can complete the path where there is a cache entry for each session property.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
commits included under #2064

PR #2056, opened 2018-02-07
Pulling in ExceptionUtils for instanceof WAE check

andymc12:
This code change was made in CXF but right after 3.2.2 released, so
we need to make this change here and remove it when we consume CXF
3.2.3 or higher.  The change is on line 70, replacing a class check
with an instanceof check to ensure that WAE subclasses are treating
the same as WAEs.


andymc12:
#build

andymc12:
Added `@FFDCIgnore` to avoid FFDCs where none used to exist. New #build

andymc12:
Last build failed due to FTP issues... Trying new #build

andymc12:
Only failure is unrelated (EBA test missing the `timed exit` feature.

PR #2057, opened 2018-02-07
Handle null/empty contracts in Client.register(...) calls

andymc12:


andymc12:
#build

andymc12:
4 failures, but they are all unrelated and are related to other defects.

PR #2064, opened 2018-02-07
remove cache dependency on liberty classes

njr-11:
Replace the key/value classes with classes from the JDK such that the JCache server will not have any dependency on being able to load classes from Liberty.


dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
```diff
@@ -341,18 +394,18 @@ protected boolean persistSession(BackedSession d2, boolean propHit) {
     @Override
     protected BackedSession readFromExternal(String id) {
         String appName = getIStore().getId();
-        SessionKey key = new SessionKey(id, appName);
+        String key = createSessionKey(id, appName);
 
         CacheSession sess = null;
-        SessionData sessionData = cacheStoreService.cache.get(key);
+        ArrayList<?> sessionData = cacheStoreService.cache.get(key);
```
aguibert: It would still be useful to keep the `SessionData` class for the purpose of working with the field values in code, but instead of pushing the [de]serialization to the JCache layer, we could handle it in the session cache layer with some static helper method, such as:
```java
public static SessionData fromList(ArrayList<?> rawData);

public static ArrayList<?> toList(SessionData sessiondata);
```
This will be useful because we'll have type safety throughout the rest of our code, instead of needing to do things like `(Long) sessionData.get(LAST_ACCESS)`

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
```diff
@@ -90,22 +89,25 @@ protected void activate(ComponentContext context, Map<String, Object> props) {
 
         // load JCache provider from configured library, which is either specified as a libraryRef or via a bell
         CachingProvider provider = Caching.getCachingProvider(library.getClassLoader());
-        cacheManager = provider.getCacheManager(null, new CacheClassLoader()); // TODO When class loader is specified, it isn't being used for deserialization. Why?
-        cache = cacheManager.getCache("com.ibm.ws.session.cache", SessionKey.class, SessionData.class);
+        cacheManager = provider.getCacheManager(null, null, null);
```
aguibert: could just call `provider.getCacheManager()` now

njr-11: I'm intentionally using the signature that allows for sending the properties because we will eventually need to do that.

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
njr-11: I suppose we could do that.  I'll work on making this change.

njr-11:
#build

njr-11:
Build has 2 test failures, in unrelated componentry.

PR #2066, opened 2018-02-07
refactor gradle tasks to output javaagent jar on assemble

ewhoch:
From some suggestions Glenn had, this should also ensure that assemble builds all of the bin/tools JARs. One (i think javaagent) was missing before.

#build


ewhoch:
#build

ewhoch:
reviving this after it went stale/unmergeable

#build

ewhoch:
Build looks good. a couple unrelated FAT errors, one child FAT didn't run. acceptable risk.

PR #2080, opened 2018-02-08
test libraryRef and minimum features set for httpSessionCache

njr-11:
Update one of the test servers to supply the JCache provider via a libraryRef, in which case the bell feature is not needed.  Have this test server provide coverage for minimum feature set, which is servlet + sessionCache.


njr-11:
#build

njr-11:
#build

PR #2083, opened 2018-02-08
remove allow http config attribute for token requests

brutif:
During externals review it was requested that this attribute be removed.


brutif:
need to park this until fats are updated.

bcpswope:
Looks like we are only deleting content here... need an ID review?

brutif:
I think we're good, thanks

brutif:
once FAT updates on 10281 are in, this one can proceed.

brutif:
#build

brutif:
argh, I had a bug.
#build

brutif:
10281 is showing up in integration now, so let's try it again and see if that test has now  been deleted.
#build

brutif:
pb is good, FAT updates have finally taken effect.

Issue #1829, opened 2018-01-26
Provider properties for httpSessionCache


njr-11:
User story:
"As a user of JCache-backed HTTP sessions, I want to configure various provider-specific CacheManager properties, in order to manage and control various aspects of the JCache provider."

PR #2103, opened 2018-02-09
Initial prototype of jcache vendor property config for sessionCache

alexsm82:
Initial prototype to allow for vendor properties to be passed to the JCache provider used for session caching.  The properties can be passed in either as a URI to their vendor config file, or as vendor specific key value pairs on the properties element.

PR for story #1829

### DIFF
--- a/dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
+++ b/dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
@@ -10,7 +10,10 @@
  *******************************************************************************/
 package com.ibm.ws.session.store.cache;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Map;
+import java.util.Properties;
 
 import javax.cache.CacheManager;
 import javax.cache.Caching;
@@ -47,6 +50,9 @@ public class CacheStoreService implements SessionStoreService {
     private static final TraceComponent tc = Tr.register(CacheStoreService.class);
     
     Map<String, Object> configurationProperties;
+    private static final String BASE_PREFIX = "properties";
+    private static final int BASE_PREFIX_LENGTH = BASE_PREFIX.length();
+    private static final int TOTAL_PREFIX_LENGTH = BASE_PREFIX_LENGTH + 3; //3 is the length of .0.
 
     CacheManager cacheManager;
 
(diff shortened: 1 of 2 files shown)


dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
```diff
@@ -53,6 +56,9 @@
     private static final TraceComponent tc = Tr.register(CacheStoreService.class);
     
     private Map<String, Object> configurationProperties;
+    private static final String BASE_PREFIX = "properties";
+    private static final int BASE_PREFIX_LENGTH = BASE_PREFIX.length();
+    private static final int TOTAL_PREFIX_LENGTH = BASE_PREFIX_LENGTH + ".0.".length();
```
njr-11: Replace with +3 and a comment indicating that 3 is the length of ".0." rather than constructing a string for the sole purpose of counting its characters.

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
```diff
@@ -83,13 +89,36 @@
      * @param props service properties
      */
     @Activate
-    @FFDCIgnore(CacheException.class)
+    @FFDCIgnore({ CacheException.class, URISyntaxException.class })
```
njr-11: Why ignore this FFDC if there is a TODO in the catch block indicating that we will later be logging an error for it anyway?

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
```diff
@@ -83,13 +89,36 @@
...
     protected void activate(ComponentContext context, Map<String, Object> props) {
-        configurationProperties = props;
-
+    	configurationProperties = props;
```
njr-11: If I'm reading this correctly, this only modification to this line is to switch from standard 4 space indent to 3, such that it no longer aligns with the rest of this method.  As long as there are other fixes being made to this file, this can be corrected as well

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
alexsm82: Good catch, I copied this from the JDBC code without much thought, not sure why we're doing that there.  I'll update to use 3.

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
alexsm82: I wasn't entirely sure what to do here in terms of logging the error/FFDC, thus the todo.  I'll remove it from the FFDCIgnore

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
alexsm82: This shows up as the proper 4 spaces for me locally so hopefully Eclipse auto-fixed it so it is fixed when I push the updates for the previous comments.

njr-11: I see what is happening, it looks like you switched some of the spaces to a tab, which prints as 3 spaces in the browser where I'm reviewing this.

alexsm82: Yeah, I realized that later.  I switched it back to spaces so hopefully that resolves it.

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
The build above had 3 failures, all of which matched existing defects.

PR #2112, opened 2018-02-09
avoid costly iterator for session property cache

njr-11:
Avoid iterating through all values of the session property cache when we only need to update a subset of the properties.  This iteration is costly because it aggressively loads all of the cached values.

### DIFF
--- a/dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
+++ b/dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
@@ -62,12 +62,16 @@ public class CacheHashMap extends BackedHashMap {
      * @param app the application
      * @return the key
      */
+    @Trivial
     static final String createSessionKey(String sessionId, String app) {
-        return new StringBuilder(sessionId.length() + 1 + app.length())
+        String key = new StringBuilder(sessionId.length() + 1 + app.length())
                         .append(sessionId)
                         .append('@')
                         .append(app)
                         .toString();
+        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
+            Tr.debug(tc, "session key: " + key);
+        return key;
     }
 
     /**
@@ -77,12 +81,16 @@ public class CacheHashMap extends BackedHashMap {
      * @param propertyId the session property
      * @return the key
      */
+    @Trivial
     static final String createSessionPropertyKey(String sessionId, String propertyId) {
-        return new StringBuilder(sessionId.length() + 1 + propertyId.length())
+        String key = new StringBuilder(sessionId.length() + 1 + propertyId.length())
                         .append(sessionId)
                         .append('.')
                         .append(propertyId)
                         .toString();
+        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
+            Tr.debug(tc, "session property key: " + key);
+        return key;
     }
 
     /**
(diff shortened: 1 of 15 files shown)


njr-11:
#build

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
```diff
@@ -205,7 +213,7 @@ protected Object loadOneValue(String attrName, BackedSession sess) {
             byte[] bytes = cacheStoreService.getCache(appName).get(key);
 
             if (trace && tc.isDebugEnabled())
-                Tr.debug(this, tc, key.toString(), bytes.length);
+                Tr.debug(this, tc, key.toString(), bytes == null ? null : bytes.length);
```
aguibert: perhaps add a message here indicating what the trace value is?  Something like:
`"byte length for session=" + key.toString()`

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMapMR.java
```diff
@@ -152,67 +157,65 @@ protected Object getValue(String id, BackedSession s) {
...
-                            doWrite = true;
                             if (trace && tc.isDebugEnabled()) {
                                 Tr.debug(this, tc, "doing app changes for " + id + " on thread " + t);
                             }
-                            if (sht != null) {
-                                vEnum = sht.keys();
-                            } else {
-                                vEnum = (new Hashtable()).keys();
-                            }
+                            propsToWrite = sht.keySet();
```
aguibert: `sht` is an unfortunate variable name...  Change to `syncHt`?

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMapMR.java
```diff
@@ -311,6 +308,23 @@ boolean handlePropertyHits(BackedSession d2) {
...
+                    oldValue = cacheStoreService.cache.get(sessionKey);
+                    if (oldValue == null)
+                        throw new UnsupportedOperationException(); // TODO implement code path where cache entry for session is expired. Delete the property entries?
+                    SessionInfo sessionInfo = new SessionInfo(oldValue).clone();
+                    if (propsToWrite != null)
+                        sessionInfo.addSessionPropertyIds(propsToWrite);
+                    if (propsToRemove != null)
+                        sessionInfo.removeSessionPropertyIds(propsToRemove);
+                    newValue = sessionInfo.getArrayList();
+                } while (!cacheStoreService.cache.replace(sessionKey, oldValue, newValue));
```
aguibert: I'm worried about two clients updating the same cache and looping indefinitely here.  We should write a test that tries to force this situation.  Also, we should implement a max-retry count here where we blow up after something like 3 retries, in order to sniff out whether or not this situation actually occurs.

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMapMR.java
njr-11: I don't like the name either.  Currently it matches DatabaseHashMapMR from which this code was copied.  At this point, the code has diverged enough it should be okay to change it

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build has 3 test failures which are in unrelated componentry.

PR #2115, opened 2018-02-09
Improve build times with publishFeatureResources

fritze2:
- Changes publishFeatureResources to collect it's bnd Project dependencies, from those it gathers a list of it's dependencies' Gradle Projects, then using those it dependsOn their prereqFeatureResources task. Speeds up full :releaseNeeded by several minutes.
- Moves publishFeatureResources out from :assemble and into :publish so that local developers aren't slowed down. That drastically reduced :assemble time from 20 minutes to 9m 17s.


fritze2:
#build

fritze2:
#build

PR #2132, opened 2018-02-12
2106 performance and trace enhancements

### LINKED ISSUE #2101: MP Config: FAT bucket does not correctly run against both 1.1 and 1.2

### LINKED ISSUE #2106: MP Config Performance: ALL sources are dynamic and they refresh too much, invalidating the cache
The default environment source should be static
The default config.properties file source should be static
The default refresh rate should be longer than 500ms

tevans78:
This PR does multiple things, all aimed at improving test performance and reliablity
1. stop the internal config sources from being dynamic
2. only invalidate the cache if a source really has updated
3. remove entry exit trace from some methods
4. make the FAT tests only trace out at debug level (i.e. not dump/all)
5. by default run using EE8 and mpConfig-1.2 ... some tests also run with EE7 and/or mpConfig-1.1

fixes #2106 
fixes #2101 
#build

### DIFF
--- a/dev/com.ibm.ws.microprofile.config.1.2_fat/bnd.bnd
+++ b/dev/com.ibm.ws.microprofile.config.1.2_fat/bnd.bnd
@@ -21,6 +21,12 @@ javac.target: 1.8
 	
 fat.project: true
 
+tested.features: mpConfig-1.2,\
+                 cdi-1.2,\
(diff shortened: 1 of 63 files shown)


tevans78:
#build

tevans78:
#build

PR #2136, opened 2018-02-12
update test case to cover getAttributeNames

njr-11:
Current sets of tests don't have coverage for getAttributeNames and the underlying code path.  Add a test case that ensures that getAttributeNames matches the set of names of attributes created in the session.


njr-11:
#build

Issue #2140, opened 2018-02-12
global.debug.java2.sec defaults to true and runs exempt servers with security


tjwatson:
When running FATs locally the global.debug.java2.sec and that causes the bootstrap.properties file for the server to include java 2 security options even when `websphere.java.security.exempt=true` is set in the server bootstrap.properties file for the FAT test server.

PR #2141, opened 2018-02-12
Only enable java security if server is not exempt

tjwatson:


tjwatson:
This is for #2140

tjwatson:
#build

Issue #2109, opened 2018-02-09
CTS: EE8: JAX-RS bucket: Client side SseEventSource.close should cause the Liberty server to close the SseEventSink connection. 


jkoehler22:
trace log here indicates that even tho the client is no longer handling event the connection in never closed. 

[2/8/18 13:44:37:575 EST] 0000007f id=         SystemOut                                                    O ISOPEN true
[2/8/18 13:44:37:575 EST] 0000007f id=43509f36 com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               > isClosed Entry 
[2/8/18 13:44:37:575 EST] 0000007f id=43509f36 com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               < isClosed Exit  
                                                                                                               false
Here are all of the closes....Note that 43509f36 is not closed

	Line 3941085: [2/8/18 13:44:39:791 EST] 00000072 id=4a3d585f com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               > close Entry 
	Line 3943835: [2/8/18 13:44:45:893 EST] 00000049 id=a69e86af com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               > close Entry 
	Line 3965881: [2/8/18 13:44:53:511 EST] 00000044 id=86eec826 com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               > close Entry 
	Line 3968632: [2/8/18 13:44:59:518 EST] 00000073 id=2b58d6c0 com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               > close Entry 
	Line 3971383: [2/8/18 13:45:02:302 EST] 00000053 id=8c7942b2 com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               > close Entry 
	Line 3989341: [2/8/18 13:45:11:687 EST] 00000087 id=460bf564 com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl               > close Entry

PR #2150, opened 2018-02-12
Close sink on exception

andymc12:


andymc12:
#build

andymc12:
umm.... #build

wraschke:
#build

wraschke:
#build

dev/com.ibm.ws.jaxrs.2.1.common/src/com/ibm/ws/jaxrs21/sse/LibertySseEventSinkImpl.java
```diff
@@ -89,39 +90,48 @@ public boolean isClosed() {
...
-                    //TODO: convert to warning?
-                    if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                        Tr.debug(tc, "send - failed sending event " + event);
-                        future.completeExceptionally(ex);
+                } catch (WebApplicationException ex) {
+                    handleException(ex, future, event);
+                } catch (IOException ex) {
+                    handleException(ex, future, event);
+                } finally {
+                    if (os != null) {
```
jim-krueger: Shouldn't this be  "if (( os != null) || (os.size() == 0)) since you set o = new ByteArrayInputStream right away so I'm not sure when it would be "null".

dev/com.ibm.ws.jaxrs.2.1.common/src/com/ibm/ws/jaxrs21/sse/LibertySseEventSinkImpl.java
```diff
@@ -134,9 +144,17 @@ public boolean isClosed() {
             if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
                 Tr.debug(tc, "SseEventSink is closed - failed sending event:  " + event);
             }
-                throw new IllegalStateException("SseEventSink is closed.");
```
jim-krueger: Rather than removing the indent on this line shouldn't the if statement above it be indented to match?

dev/com.ibm.ws.jaxrs.2.1.common/src/com/ibm/ws/jaxrs21/sse/LibertySseEventSinkImpl.java
andymc12: No, we always want to close the ByteArrayOutputStream.  The case where it could be null is if the constructor itself threw an exception.

dev/com.ibm.ws.jaxrs.2.1.common/src/com/ibm/ws/jaxrs21/sse/LibertySseEventSinkImpl.java
andymc12: I let Eclipse handle the indenting using the Source -> Correct Indentation, but it looks right to me now.

andymc12:
Resolves #2109 - failures in builds above are unrelated and different intermittent issues.

dev/com.ibm.ws.jaxrs.2.1.common/src/com/ibm/ws/jaxrs21/sse/LibertySseEventSinkImpl.java
jim-krueger: That makes sense... agreed.

PR #2151, opened 2018-02-13
Clean up internal CDI api usage

### LINKED ISSUE #1619: Update beanValidation to use new CDIService instead of CDIRuntime
Currently beanVal goes after internal CDI APIs to check if the current module is CDI enabled.  

This was cleaned up in PR #1440, so the beanVal runtime should be updated accordingly to no longer go after the internal CDIRuntime class.

aguibert:
fixes #1619

### DIFF
--- a/dev/com.ibm.ws.beanvalidation.v20.cdi/src/com/ibm/ws/beanvalidation/v20/cdi/internal/LibertyHibernateValidatorExtension.java
+++ b/dev/com.ibm.ws.beanvalidation.v20.cdi/src/com/ibm/ws/beanvalidation/v20/cdi/internal/LibertyHibernateValidatorExtension.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 IBM Corporation and others.
+ * Copyright (c) 2017, 2018 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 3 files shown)


dev/com.ibm.ws.beanvalidation.v20.cdi/src/com/ibm/ws/beanvalidation/v20/cdi/internal/LibertyHibernateValidatorExtension.java
```diff
@@ -194,6 +194,20 @@ private ClassLoadingService getClassLoadingService() {
...
+        // Get the CDIService
+        Bundle bundle = FrameworkUtil.getBundle(CDIService.class);
+        CDIService cdiService = AccessController.doPrivileged((PrivilegedAction<CDIService>) () -> {
+            BundleContext bCtx = bundle.getBundleContext();
+            ServiceReference<CDIService> svcRef = bCtx.getServiceReference(CDIService.class);
+            return svcRef == null ? null : bCtx.getService(svcRef);
+        });
+        if (cdiService == null) {
+            return null;
+        }
```
tevans78: a) why get hold of CDIService the "hard" way? Why not just inject a Reference?
b) you're calling this method from within a CDI Extension method ... if the CDISerivce is null then there's something very odd happening!

aguibert: for (a) I didn't want to define a new service dependency, since this reference will only ever be used on an error path (most of the time it will never be needed at all).  
for (b) the extra null check was just my tinfoil hat here... wasn't sure what sort of things might be going wrong if we're having problems registering the CDI extension so I didn't want to make any extra assumptions.
I can resolve (b), but would prefer to keep (a) the way it is

aguibert:
#build

aguibert:
#build

aguibert:
#build

PR #2152, opened 2018-02-13
Initial version of jCacheContainer-1.1 feature

aguibert:
Split the javax.cache API into a separate public feature called `jCacheContainer-1.1` instead of exposing it via the `sessionCache-1.0` feature.


aguibert:
#build

PR #2155, opened 2018-02-13
use valid cache names and avoid unnecessary find or create

njr-11:
The : and / characters are not valid in JCache cache names.  To maintain a guarantee of uniqueness, we can't just swap the characters to another character, so I'm taking an approach of percent encoding them (along with the % character which it becomes necessary to encode as well to maintain uniqueness).  Also, I'm making updates to greatly reduce the number of find/create attempts that are made on these per-application caches.


njr-11:
#build

njr-11:
#build

njr-11:
Build has 5 test failures, all pre-existing issues in unrelated componentry

njr-11:
With the ongoing test infrastructure issues preventing deliveries, the commit for this are included in subsequent pulls, including #2335, which will be used for delivery.  Closing this pull.

Issue #1261, opened 2017-12-07
LDAP registry with global class mapping in groupMemberIdMap adds "objectclass=*" to Group searches


jvanhill:
Using the global class mapping in the 'groupMemberIdMap' setting such as:

`groupMemberIdMap="*:member"`

 will result in "objectclass=*" being appended to any group search filters. This will lead to potential ClassCastExceptions when trying to map a returned LDAP entity to a group since entities of any objectclass can now be returned.

```
[12/7/17 13:52:35:137 EST] 00000035 RegistryExcep >  <init> Entry  
                                  com.ibm.wsspi.security.wim.model.Entity incompatible with com.ibm.wsspi.security.wim.model.Group
                                  java.lang.ClassCastException: com.ibm.wsspi.security.wim.model.Entity incompatible with com.ibm.wsspi.security.wim.model.Group 
```

PR #2185, opened 2018-02-13
Added a check for "*" before adding an ObjectClass to the filter.

pubgagrimm:
Added a check for "*" before adding an ObjectClass to the filter.


pubgagrimm:
fixes #1261

dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/LdapConfigManager.java
```diff
@@ -866,7 +866,9 @@ private void setFilters(Map<String, Object> configProps) {
                     List<String> objClses = ldapEntity.getObjectClasses();
                     for (int j = 0; j < objClses.size(); j++) {
                         String objCls = objClses.get(j);
-                        objectClasses.add(objCls);
```
jvanhill: I don't believe this change should be here. We only want to make sure the objectclass "*" doesn't end backup on the group entity.

kristip17:
This was resolved under #3823

PR #2212, opened 2018-02-14
Don't wait for a feature update

brenthdaniel:
We don't need to wait for a feature update in the createHttpEndpoints method


brenthdaniel:
#build

PR #2236, opened 2018-02-15
Set log mark before deploying CDI extension app

Azquelt:
It seems like in some cases the app may be already deployed. The test
framework copies the app to dropins and then looks for the app started
message. Without setting the log mark, it will find the earlier message
from when the old app was started, meaning that it will start making
requests before the app is ready.


Azquelt:
Note: without this fix, the test usually passes because the server hasn't noticed that the app has been replaced at the point where the test sends the request, so it ends up being serviced by the old app rather than the new app.

In this case, I don't think it matters because the app will be the same, only the server config is changed by the `RepeatTests` class rule in the suite.

Azquelt:
#build

PR #2062, opened 2018-02-07
make MemoryPersistence tolerate a JPAQueryHelper object and perform t…

ajm01:
defect 250599 - WDT unable to access JobInstances using MemoryPersistence.
MemoryPersistence should tolerate a BASE JPAQueryHelper query and perform the equivalent retrieval.

PR #2237, opened 2018-02-15
Revert "make MemoryPersistence tolerate a JPAQueryHelper object and perform t…"

websterc87:
Reverts OpenLiberty/open-liberty#2062

The release build and continuous builds have been failing since Liberty Continuous Build - EBC 20180214-1741

251418: (HOT) Test Failure (20180214-2336): com.ibm.ws.ui.tool.javaBatch_fat.gui.tests.JavaBatchBasicTests.testInMemoryUnsupportedErrorPopup

The test is generating a FFDC with "java.lang.UnsupportedOperationException: The REST URL search parameters requesting this functionare not supported by the Java batch memory-based persistence configuration."

This seems directly related the above change


ewhoch:
Usually this would be an easy decision to revert, in spite of it being a fix for another high priority item. Keeping our builds good is paramount since problems spread to other people and cause noise in the system. However we didn't track down this introducing cause until it already reached the master branch, and the scope of impact is limited to a javabatch UI test which we've already turned off temporarily so this can probably stay in.

PR #2246, opened 2018-02-15
Make com.ibm.ws.cdi.visibility_fat pass java 2 security testing

Azquelt:
This test uses AnnotationLiteral which currently requires the app to
have the accessDeclaredMembers permission.

### DIFF
--- a/dev/com.ibm.ws.cdi.visibility_fat/fat/src/com/ibm/ws/cdi12/fat/tests/VisTest.java
+++ b/dev/com.ibm.ws.cdi.visibility_fat/fat/src/com/ibm/ws/cdi12/fat/tests/VisTest.java
@@ -220,7 +220,8 @@ public class VisTest extends LoggingTest {
                         .addAsModule(visTestWarAppClientLib17)
                         .addAsModule(visTestNonLib18)
                         .addAsLibrary(visTestFramework19)
-                        .addAsLibrary(visTestEarLib20);
+                        .addAsLibrary(visTestEarLib20)
+                        .addAsResource("com/ibm/ws/cdi12/fat/tests/permissions.xml", "permissions.xml");
 
        List paths = new ArrayList<String>();
        paths.add("publish/servers/visTestServer/apps");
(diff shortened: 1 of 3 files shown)


Azquelt:
#build

Azquelt:
The build shows failures with the client portion of the tests because fatTestPorts.xml isn't present.

I think this is a problem caused by a bug in the testing framework which can occur when a client test is the first thing to run on a given test server.

Kicked off a build including a fix for that: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_QiYK8RVwEeiANdIhjNYPnw

If that passes than this should be safe to deliver as the test hasn't been failing in this way previously.

Azquelt:
#build - better have a regular build as well

Azquelt:
That last build looks good. The only failure is something unrelated.

Azquelt:
Build running all fats rather than just the changed one. https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_KI_AEBZZEeiANdIhjNYPnw

I still think this fails on a build only if it's the first test to run, so I think would make it work. If this passes I'll deliver the change.

Azquelt:
That build ran the bucket twice, once on it's own as a full fat (which failed, due to the fatTestPorts.xml error) and once as a normal fat along with a bunch of others (which worked) so this is safe to deliver.

Azquelt:
@tevans78 asked me to ensure the whole bucket would pass with J2security enabled before merging

Azquelt:
Build submitted which will run all fats: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_v-ZfQBcYEeiANdIhjNYPnw

Expect VisTest to fail twice (in the full fat run) and pass twice (in the regular run)

Azquelt:
Same build with java2.sec enabled: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewDefinition&id=_IQRVkGV1EeaugeK15nw0Xw

Expect way more failures in this build overall but we're only interested in the com.ibm.ws.cdi.visibility_fat (though we can't run this on its own due to the fatTestPorts.xml problem)

Azquelt:
Builds look good

PR #2260, opened 2018-02-16
Issue #2248 - reduce config trace at highest levels

### LINKED ISSUE #2248: MP Config: Further reduce trace output
At the moment, if "all" trace is turned on then there is far too much "dump" level trace coming out. Need to reduce this down as much as possible so that "all" can be used in service without causing major performance or disk space issues.

Will try to keep a single "dump" output in the build() method. Config.toString() should not output dump level detail.

tevans78:
#build 
fixes #2248


tevans78:
#build

PR #2167, opened 2018-02-13
#1441 Merge JTS into BTS and format events at handlers

Yushan-Lin:
#build

Issue #1441, opened 2018-01-04
Merge JsonTraceService into BaseTraceService


Channyboy:
Merge logic introduced into JsonTraceService back into BaseTraceService.
Due to the impact of the new logic, the impact to HpelBaseTraceService and Jsr47TraceService should be investigated as well.

ewhoch:
reopened since PR 2167 was backed out.

Issue #1449, opened 2018-01-04
Unify formatting of message and console events into the MessageLogHandler and ConsoleLogHandler


Channyboy:
Currently Json formatting and write out happens in the handler and basic formatting and write out in BaseTraceService/JsonTraceService.

We would like to unify this so that the formatting happens in the _last step_ at the handler.

Reminder: only message/log is allowed to be emitted when in basic mode.

ewhoch:
reopened since PR 2167 was backed out.

PR #2282, opened 2018-02-19
Revert "#1441 Merge JTS into BTS and format events at handlers"

ewhoch:
Reverts OpenLiberty/open-liberty#2167

Under analysis, there is a 3x increase in the time it takes to run a logstash test. In combination with that, that's very close to when we have had some significant performance problems. I see the increase first on a PB, and I have strong suspicion it's this change in what looks like a frequently called method:

https://github.com/Yushan-Lin/open-liberty/commit/61af83f398d44620dde852525c4e1a0c62cec921#diff-473301b347975121e93d7d564e6abaf7


ewhoch:
#build

ewhoch:
Confirmed this causes a 3x increase in logstash.collector FATs, mainly due to timeouts looking for "TRAS0218I"

In isolation, this would be OK, but seeing how it's coincident with seeing some significant performance instability, i'm moving this PR out of the stream until we can do some more analysis on it. I'm very concerned with the new object create during array copy for what I think may be every logging event.

ewhoch:
reopens #1441 and #1449

fmhwong:
@ewhoch Do you have links to the builds that it was timed out looking for "TRAS0218I"?

ewhoch:
look at any build where it ran for 30+ minutes, literally any of them.

PR #2301, opened 2018-02-20
Mechanism to fail a FAT that generates more than X MB of trace while running

aguibert:
We have found out the hard way that some FATs are putting out great deal of trace (11 buckets >1GB in lite mode, with the highest being 3.4GB).  This has caused us to max out on disk IO in our build cloud and we need a mechanism to cap how much trace a FAT can put out in order to prevent individual FATs from saturating the build hardware and dragging builds down.

This PR introduces a mechanism to measure the size of log files during server stop (as they are brought into the autoFVT directory) and raise a JUnit failure if the threshold is exceeded.

- The limit is currently 4GB in LITE mode (intentionally set high).
- When running in FULL mode, the limit is increased by a factor of 3.
- When running locally, the limit is 85% of what it would be otherwise (in order to try and catch buckets running close to the limit locally)
- The limit can be overridden manually by setting the property: `fat.test.max.allowed.trace.mb` to the trace limit (in MB) for LITE mode.  This sets the pre-adjusted limit (i.e. before FULl or local mode is taken into account).
- The trace limit for a FAT can be overridden programatically for a specific FAT in the following way:
```java
static {
  LogPolice.setMaxAllowedTraceMB(2000); // sets trace limit for the bucket to 2GB
}
```

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/custom/junit/runner/LogPolice.java
+++ b/dev/fattest.simplicity/src/componenttest/custom/junit/runner/LogPolice.java
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2018 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package componenttest.custom.junit.runner;
+
+import com.ibm.websphere.simplicity.log.Log;
+
+import componenttest.custom.junit.runner.Mode.TestMode;
+
+/**
+ * Used to ensure a single FAT does not produce too much trace logs.
+ */
+public class LogPolice {
+
+    private static final long MB = 1024000L;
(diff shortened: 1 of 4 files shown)


aguibert:
#build

aguibert:
#build

aguibert:
#build

aguibert:
#build

wraschke:
#build

dev/fattest.simplicity/src/componenttest/custom/junit/runner/LogPolice.java
```diff
@@ -0,0 +1,70 @@
...
+import componenttest.custom.junit.runner.Mode.TestMode;
+
+/**
+ * Used to ensure a single FAT does not produce too much trace logs.
+ */
+public class LogPolice {
+
+    private static final long MB = 1024000L;
+    // TODO: Start out with 4GB cap on trace in LITE mode, incrementally constrict this down to ~300MB trace allowed per bucket
+    private static final long MAX_ALLOWED_TRACE = MB * Integer.getInteger("fat.test.max.allowed.trace.mb", 4000); // 4GB max (for now), override w/ system prop
```
ewhoch: Does this delegate eventually to a build prop? Seems like it does, but it's easier to just ask.

aguibert: Yes, the value here is parsed from the `fat.test.max.allowed.trace.mb` system property.  If that property is not set, then a default value of 4000 is used.

PR #2091, opened 2018-02-08
Create microProfile 1.3 grouping feature

NottyCode:

PR #2306, opened 2018-02-20
Add mp13 feature and change all dependent features to GA

ewhoch:
copies exactly the commits in #2091 but squashes them into one commit.


ewhoch:
This has all been built before, but I'll #build it to make sure i got the squash right.

PR #2313, opened 2018-02-20
Remove JaxRsModuleMetaData from overlay cache on app shutdown

andymc12:
It is possible to leak classloaders because the JaxRsModuleMetaData references an app's thread context classloader and is stored in the NonpersistentOverlayCache.  Normally this isn't a problem because the cache is removed when the classloader is removed - which happens at app shutdown, but common shared libraries have their own classloader that does not recycle when the app does - so it remains in the heap, referencing the overlay cache which references the JaxRsModuleMetaData, which references the thread context classloader - leaking it.  

The solution is to remove the JaxRsModuleMetaData from the cache when the metadata is destroyed (as part of app shutdown).


andymc12:
#build

andymc12:
#build

andymc12:
weird failures... trying again with a new #build after rebasing

andymc12:
Rebasing and new #build now that streams are re-opened.

andymc12:
#build

andymc12:
Two failures seem unrelated.

PR #2330, opened 2018-02-21
Reduce FAT tracing

andymc12:
Recent build analysis shows that the MP Rest Client FAT is generated a lot (~186MB) of trace which strains the build/test infrastructure.  This reduces some of that tracing.


andymc12:
#build

PR #2335, opened 2018-02-21
use cache entry per session property approach

njr-11:
Always use the approach of having a cache entry for each session property.


njr-11:
#build

njr-11:
#build

njr-11:
Build has 9 test failures, all of which are pre-existing issues in unrelated componentry.
Once the repository is opened up again this can be merged.

PR #2346, opened 2018-02-21
Set qualifiers to timestamps and upload a zip artifact of .feature files

fritze2:
- Sets the Bnd bundles version qualifiers to timestamps instead of branch names.
- Adds all .feature file sources to gradle-bootstrap.


fritze2:
#build

Issue #2352, opened 2018-02-21
Update buckets to use in-memory database


alexsm82:
Update those buckets which are able to use derby in-memory to do so if they are not already.  This will allow Derby to better function should infrastructure issues occur, reducing unnecessary defects.

alexsm82:
#build

PR #2355, opened 2018-02-21
Update to in-memory Derby where possible

alexsm82:
Update those buckets which are able to use derby in-memory to do so if they are not already. This will allow Derby to better function should infrastructure issues occur, reducing unnecessary defects.

Issue #2352


alexsm82:
#build

PR #2363, opened 2018-02-22
Fix cdi fat app deployments

### LINKED ISSUE #2302: Some tests don't deploy apps correctly with RepeatTests
When we're using the RepeatTests rule on the FATSuite, each test class is run more than once. Some tests don't cope with this properly.

In particular, tests which do the following will break:
- get a started server
- deploy the app
- wait for the app started message to appear in the log

On the second test run, the server starts, starts the old app, then the test deploys the new app (which is probably identical), then the server updates the app (and puts out an app updated rather than app started message).

One correct sequence would be:
 - get a started server
 - set the mark to the end of the log
 - deploy the app
 - wait for either an app started or app updated message

Side note: tests which deploy the app before starting the server should not hit this problem.

Azquelt:
Fix tests where we don't deploy the app properly when using RepeatTests

The general problem is that when the test runs a second time, the app will already be deployed and there will be an app started message in the log.

Some of the tests redeploy the app. In these cases, we need to set the mark before deploying and check for either the app started or app updated message.

This causes intermittent build breaks if the server processes the redeployment while a test is running.

This fixes #2302

### DIFF
--- a/dev/com.ibm.ws.cdi.beansxml_fat/fat/src/com/ibm/ws/cdi12/fat/tests/implicit/ImplicitWarLibJarsTest.java
+++ b/dev/com.ibm.ws.cdi.beansxml_fat/fat/src/com/ibm/ws/cdi12/fat/tests/implicit/ImplicitWarLibJarsTest.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package com.ibm.ws.cdi12.fat.tests.implicit;
  
+import static org.junit.Assert.assertNotNull;
+
 import java.io.File;
 
 import org.junit.AfterClass;
@@ -88,10 +90,11 @@ public class ImplicitWarLibJarsTest extends LoggingTest {
                         .addAsLibrary(implicitBeanExplicitArchive)
                         .addAsLibrary(utilLib);
 
+       server.setMarkToEndOfLog(server.getDefaultLogFile());
        ShrinkHelper.exportDropinAppToServer(server, implicitEJBInWar);
        ShrinkHelper.exportDropinAppToServer(server, implicitBeanDiscovery);
-
-       server.waitForStringInLogUsingMark("CWWKZ0001I.*Application implicitBeanDiscovery started");
+       assertNotNull("implicitBeanDiscovery started or updated message", server.waitForStringInLogUsingMark("CWWKZ000[13]I.*implicitEJBInWar"));
+       assertNotNull("implicitBeanDiscovery started or updated message", server.waitForStringInLogUsingMark("CWWKZ000[13]I.*implicitBeanDiscovery"));
     }
 
     /**
--- a/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi12/fat/tests/AroundConstructTestBase.java
+++ b/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi12/fat/tests/AroundConstructTestBase.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package com.ibm.ws.cdi12.fat.tests;
 
+import static org.junit.Assert.assertNotNull;
+
 import java.io.File;
 
 import org.junit.AfterClass;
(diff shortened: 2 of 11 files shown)


Azquelt:
#build

dev/com.ibm.ws.cdi.beansxml_fat/fat/src/com/ibm/ws/cdi12/fat/tests/implicit/ImplicitWarLibJarsTest.java
```diff
@@ -88,10 +90,10 @@ public static void buildShrinkWrap() throws Exception {
                         .addAsLibrary(implicitBeanExplicitArchive)
                         .addAsLibrary(utilLib);
 
+       server.setMarkToEndOfLog(server.getDefaultLogFile());
        ShrinkHelper.exportDropinAppToServer(server, implicitEJBInWar);
        ShrinkHelper.exportDropinAppToServer(server, implicitBeanDiscovery);
-
-       server.waitForStringInLogUsingMark("CWWKZ0001I.*Application implicitBeanDiscovery started");
+       assertNotNull("implicitBeanDiscovery started or updated message", server.waitForStringInLogUsingMark("CWWKZ000[13]I.*implicitBeanDiscovery"));
```
tevans78: will implicitBeanDiscovery always start after implicitEJBInWar? Do you actually need to check both?

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi12/fat/tests/AroundConstructTestBase.java
```diff
@@ -105,9 +107,11 @@ public static void setUp() throws Exception {
                         .add(new FileAsset(new File("test-applications/postConstructErrorMessageApp.war/resources/WEB-INF/beans.xml")), "/WEB-INF/beans.xml")
                         .addAsLibrary(utilLib);
 
+        server.setMarkToEndOfLog(server.getDefaultLogFile());
         ShrinkHelper.exportDropinAppToServer(server, aroundConstructApp);
         ShrinkHelper.exportDropinAppToServer(server, postConstructErrorMessageApp);
-        server.waitForStringInLogUsingMark("CWWKZ0001I.*Application aroundConstructApp started");
+        assertNotNull("aroundConstructApp started or updated message", server.waitForStringInLogUsingMark("CWWKZ000[13]I.*aroundConstructApp"));
+        assertNotNull("postConstructErrorMessageApp started or updated message", server.waitForStringInLogUsingMark("CWWKZ000[13]I.*postConstructErrorMessageApp"));
```
tevans78: see... you've checked for both here!

dev/com.ibm.ws.cdi.visibility_fat/fat/src/com/ibm/ws/cdi12/fat/tests/PackagePrivateAccessTest.java
```diff
@@ -54,8 +56,9 @@ public static void setUp() throws Exception {
                         .addClass("jp.test.bean.MyBean")
                         .add(new FileAsset(new File("test-applications/packagePrivateAccessApp.war/resources/WEB-INF/web.xml")), "/WEB-INF/web.xml");    
         server = LibertyServerFactory.getStartedLibertyServer("packagePrivateAccessServer");
+        server.setMarkToEndOfLog(server.getDefaultLogFile());
         ShrinkHelper.exportDropinAppToServer(server, packagePrivateAccessApp);
-        server.waitForStringInLogUsingMark("CWWKZ0001I.*Application packagePrivateAccessApp started");
+        assertNotNull("packagePrivateAccessApp started or updated message", server.waitForStringInLogUsingMark("CWWKZ000[13]I.*packagePrivateAccessApp"));
```
tevans78: is there anywhere common that could sensibly have a `dropInAndStartApplication` method? rather than repeating this same code?

dev/com.ibm.ws.cdi.beansxml_fat/fat/src/com/ibm/ws/cdi12/fat/tests/implicit/ImplicitWarLibJarsTest.java
Azquelt: Yeah, probably should check both

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi12/fat/tests/AroundConstructTestBase.java
Azquelt: I was more on the ball when I changed this file :wink:

dev/com.ibm.ws.cdi.visibility_fat/fat/src/com/ibm/ws/cdi12/fat/tests/PackagePrivateAccessTest.java
Azquelt: Yes, but I don't like that it changes the mark as a side effect :(

Azquelt: You could do a method which would drop in and wait for a start or updated message, but then you would still know that you should set the mark before calling it, so it would reduce the amount of code but hide what some of what was going on and still be easy to screw up.

Azquelt:
#build

PR #2367, opened 2018-02-22
Reduce trace in javaeesec fat bucket

kristip17:
Reduce to just ws.security trace. Bump some of the DB fat tests to full bucket.


kristip17:
#build

kristip17:
#build

kristip17:
fixes #2663

PR #2386, opened 2018-02-22
Upgrade lib versions, and remove legacy bundles

cbridgha:
This should finish off the work of upgrading and removing old version of commons libs


cbridgha:
Closing - no longer relevant

PR #2380, opened 2018-02-22
Change coordinates for arquillian plugin

andymc12:

PR #2409, opened 2018-02-26
Switch MicroProfile TCKs to use non-snapshot arquillian-wlp-managed-8.5

Azquelt:
The new version is identical to the current latest snapshot build so
should not change how the tests run.

The purpose of this change is to ensure that the 18.0.0.1 build is
reproducable in future.

PR #2413, opened 2018-02-26
Using 1.0.0.Final-SNAPSHOT to pull in what is in artifactory

ewhoch:
See PR #2380, this makes it (wrongly) consistent until we can fix these up to not point at snapshots.


andymc12:
> (wrongly) consistent

Correct, and @Azquelt is making the proper fix in PR #2409

PR #1986, opened 2018-02-05
1560 shrinkwrap helper should not default to overwiting when exporting artifacts to the server

benjamin-confino:

Issue #1560, opened 2018-01-11
ShrinkHelper defaults to overwriting artifacts


benjamin-confino:
With the new repeat actions for EE8 code we end up overwriting a lot of test artifacts. We should make it default to not overwriting artifacts

benjamin-confino:
Fixed under #1656

benjamin-confino:
The fix in 1656 was not enough. I am preparing a second fix

PR #2429, opened 2018-02-26
Revert "1560 shrinkwrap helper should not default to overwiting when exporting artifacts to the server"

ewhoch:
Reverts OpenLiberty/open-liberty#1986

Causes a failure among com.ibm.ws.CDI FATs due to a soft conflict with something @Azquelt delivered.


ewhoch:
reopens #1560

PR #2440, opened 2018-02-27
Microprofile Bulkhead: fix waitingTaskQueue default

### LINKED ISSUE #2400: MicroProfile Bulkhead: Incorrect default set for waitingTaskQueue
The default value of 10 is not being set for waitingTaskQueue.  Default for MaxThreads is also being incorrectly set to waitingTaskQueue default, though that's currently not causing issues since they both should have a default value of 10.

tevans78:
fixes #2400 
I can't run a build yet because of the cloud issues


tevans78:
#build

PR #2455, opened 2018-02-28
Fix MicroProfile FATs which repeat tests at the class level

Azquelt:
When using RepeatTests at the class level (rather than the suite level)
you need to specify which servers should be updated, otherwise your
updates don't get propagated to the server.

Also, RepeatTests needs the server to be shut down at the end of the test class, so we can't start the server once in the suite and use it in multiple classes.

Also, `ShrinkHelper.exportToServer()` writes apps to the `publish` directory. This means that when the test is repeated, the app is already deployed to the server. In some cases this isn't desirable, so we either need to not use `exportToServer()` or we need to manually clean out the `publish` directory at the end of the test.


Azquelt:
This still needs some work as two of the faulttolerance test classes fail

Azquelt:
#build

Azquelt:
#build

PR #2477, opened 2018-03-01
Issue 2165: Adding timeout test server

mswatosh:
Adding server for timeout tests.

### DIFF
--- a/dev/com.ibm.ws.session.cache_fat/fat/src/com/ibm/ws/session/cache/fat/SessionCacheTimeoutTest.java
+++ b/dev/com.ibm.ws.session.cache_fat/fat/src/com/ibm/ws/session/cache/fat/SessionCacheTimeoutTest.java
@@ -1,0 +1,61 @@
+/**
+ *
+ */
+package com.ibm.ws.session.cache.fat;
+
+import static org.junit.Assert.assertNotNull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import componenttest.annotation.Server;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+
+/**
+ * Tests related to Session Cache Timeouts, using a server with the following session settings:
+ * invalidationTimeout="1s"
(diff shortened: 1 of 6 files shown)


dev/com.ibm.ws.session.cache_fat/fat/src/com/ibm/ws/session/cache/fat/SessionCacheTimeoutTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.runner.RunWith;
+
+import componenttest.annotation.Server;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+
+/**
+ * Tests related to Session Cache Timeouts, using a server with the following session settings:
+ * invalidationTimeout="1s"
```
aguibert: I think a timeout of 1s is too small, as we may not be able to reliably hit that window for the requests we need to be successful.  Lets do something longer like 5s.

dev/com.ibm.ws.session.cache_fat/publish/servers/sessionCacheTimeoutServer/server.xml
```diff
@@ -0,0 +1,76 @@
...
+	</library>
+
+    <!-- Used for testing that data source can be stored in a session -->
+	<authData id="DerbyAuth" user="userName" password="userPassword"/>    
+    <dataSource id="DerbyDS" jndiName="jdbc/derby" containerAuthDataRef="DerbyAuth">
+	    <jdbcDriver libraryRef="DerbyLib"/>
+	    <properties.derby.embedded createDatabase="create" databaseName="memory:testdb"/>
+	</dataSource>
+
+	<javaPermission codebase="${shared.resource.dir}/derby/derby.jar" className="java.security.AllPermission"/>
```
aguibert: Looks like the set of `<javaPermission>` configs got duplicated somewhere along the way?  Not harming anything, so we can clean it up later if necessary.

dev/com.ibm.ws.session.cache_fat/fat/src/com/ibm/ws/session/cache/fat/SessionCacheTimeoutTest.java
mswatosh: I pick 1s to keep the test short, but since the reaper thread is minimum 30s, that's what we're most likely to be waiting on anyways, so I'll increase to 5s.

PR #2488, opened 2018-03-02
Make the timeout more lenient on FT testTimeout

Azquelt:
Make the timeout slightly more lenient on this first test which seems to do a bit of lazy initialization which can cause a spurious failure if the build hardware is particularly slow.

The test is now checking that the method returns within 3 seconds. If the timeout doesn't fire, it would take 20 seconds.

Also fixed up some of the error messages to make them more helpful.


Azquelt:
#build

Azquelt:
Copyright date updated

PR #2492, opened 2018-03-02
check for all SqlSyntaxErrorException classes when creating PSU

ajm01:
need to check for other SqlSyntaxErrorExceptions when determining level of the JobExecution Table


scottkurz:
Extension on PR #2492

PR #2496, opened 2018-03-02
#2497 JSONLogging Changed copy in read to copy on write for BufferManagerImpl

Arjun-Sharma1:
#2497


dev/com.ibm.ws.logging.core/src/com/ibm/ws/collector/manager/buffer/BufferManagerImpl.java
```diff
@@ -106,6 +87,11 @@ public void add(Object event) {
 
         } finally {
             RERWLOCK.readLock().unlock();
+            if (!synchronousHandlerSet.isEmpty()){
```
raymondlam: I don't think this check is needed if you are using the iterator

Arjun-Sharma1:
#build

PR #2509, opened 2018-03-05
Remove exception message check from test

idlewis:
The exception message is different in different JVMs, so
the test was failing in some builds. Modify the catch to
only catch ZipException to try and make the test a bit
more useful


idlewis:
#build

idlewis:
#build

PR #2521, opened 2018-03-05
Update session cache metatype to sentence case

alexsm82:
Update session cache metatype so that label attribute labels are sentence case.


chirp1:
I reviewed the string updates. They are OK. I added ID reviewed to this PR.

PR #2522, opened 2018-03-05
Avoid leaking Bus objects in the busSet

andymc12:
Prevent the memory leak by ensuring that we remove Bus objects from the
set when the bus is shutdown (usually at app stop).


andymc12:
#build

PR #2523, opened 2018-03-05
Ensure preShutdown invokes preShutdown of LibertyApplicationBusListeners

andymc12:
@jhanders34 and I discovered a copy/paste error `LibertyApplicationBusFactory` while investigating the fix for PR #2522.  The `preShutdown` and `postShutdown` methods in the `BusLifeCycleListenerAdapter` inner class were both calling `postShutdown` on the registered listeners.  Instead the `preShutdown` method should call `preShutdown`.


andymc12:
#build

andymc12:
#build

PR #2547, opened 2018-03-06
additional multihreaded tests for session cache

njr-11:
Various tests of session cache when multiple threads are attempting to perform updates/gets/removals of session attributes


njr-11:
@build

njr-11:
#build

PR #2553, opened 2018-03-07
Test that maxInactiveInterval is properly used for invalidation

alexsm82:
Adds a test to ensure that the value set on a session via setMaxInactiveInterval is used to invalidation in a session cache.


alexsm82:
#build

PR #2595, opened 2018-03-08
2255 master cache common fields for json logging

### LINKED ISSUE #2255: cache common fields to improve JSON logging performance
Looking at the JSON generated for a typical message event you can see certain things don't change...

```
{
 "type":"liberty_message",                          // same for all message events
 "host":"192.168.2.15",                             // same for life of server
 "ibm_userDir":"\/wlp\/usr\/",                      // same for life of server
 "ibm_serverName":"server1",                        // same for life of server
 "ibm_datetime":"2018-02-04T18:56:30.318-0500",
 "ibm_sequence":"1517788590318_000000003A4A7",
 "loglevel":"INFO",
 "module":"com.ibm.somelogger.QuickLogTest",
 "ibm_threadId":"00002db5",
 "message":"some message", 
 "ext_userName":"don", 
 "ext_isHere_bool":true
}
```

Rather than go through a lot of method calls for the first 4 fields we should cache the type/host/ibm_userDir/ibm_serverName the first time we generate them.  This cached string would look like:

```
{"type":"liberty_message","host":"192.168.2.15","ibm_userDir":"\/wlp\/usr\/","ibm_serverName":"server1",
```

Then we could use that same string to start each new liberty_message event.  Something like this:

```
private static String start_message_json_1_1 = null;

...

StringBuilder startMessageJson1_1(String hostName, String wlpUserDir, String serverName) {
    StringBuffer sb = new StringBuilder(512);

    if (start_message_json_1_1 != null) {
        sb.append(start_message_json_1_1)
    }
    else {
        addToJSON(sb, "type",           "liberty_message", false, false, false, true);
        addToJSON(sb, "host",           hostName,          false, false, false, false);
        addToJSON(sb, "ibm_userDir",    wlpUserDir,        false, true,  false, false);
        addToJSON(sb, "ibm_serverName", serverName,        false, false, false, false);
        start_message_json_1_1 = sb.toString();
    }
    
    return sb;
}
```

You would need one of the above methods for each of the different types (message, trace, ffdc, garbagecollection, accesslog).  
These would replace the addCommonFields* methods in CollectorJsonHelpers though you need to be aware that in some cases (eg. addCommonFieldsGC1_1) the addCommonFields method is adding more than the 4 fields (in which case you'll need the caller of startXXXJson method to handle adding the fields not in that group of 4 after calling the startXXXJson method).

levin-noro:
Fixes #2255

Fields such as host, wlpUserDir, ibm_userDir, serverName, ibm_serverName stay the same for the life of the server and fields such as type stay the same for each event type. Instead of generating the string for these fields everytime for every individual event, the string will be generated the first time this type of event occurs and then stored so that it can be used for all future occurences of this event.

### DIFF
--- a/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/collector/CollectorJsonHelpers.java
+++ b/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/collector/CollectorJsonHelpers.java
@@ -155,7 +155,7 @@ public class CollectorJsonHelpers {
     private static void addUnchangingFields(StringBuilder sb, String hostName, String wlpUserDir, String serverName) {
         if (unchangingFieldsJson == null) {
             StringBuilder temp = new StringBuilder(512);
-            addToJSON(temp, "host", hostName, false, false, false, false);
+            addToJSON(temp, "hostName", hostName, false, false, false, false);
             addToJSON(temp, "wlpUserDir", wlpUserDir, false, true, false, false);
             addToJSON(temp, "serverName", serverName, false, false, false, false);
             unchangingFieldsJson = temp.toString();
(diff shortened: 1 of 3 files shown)


levin-noro:
#build

PR #2596, opened 2018-03-08
test cache contents after multithreaded usage

njr-11:
Update tests to check contents of cache after multithreaded access to sessions.
This caught a bug where the bytes for the value null were being written to the cache in the case where an attribute was simultaneously being removed.


njr-11:
#build

njr-11:
build has 4 failures but are pre-existing issues in other componentry

PR #2605, opened 2018-03-09
2444 - fix user being unable to modify the EntityStream from a ClientRequestFilter

WhiteCat22:
CXF uses Phases to order the interceptors in the PhaseInterceptorChain. http://cxf.apache.org/docs/interceptors.html

A customer's client filter code gets executed in ClientRequestFilterInterceptor.handleMessage() during the PRE_LOGICAL phase of the outgoing chain.
We don't actually create the entity output  stream until HTTPConduit.prepare() which is called by MessageSenderInterceptor.handleMessage() during the PREPARE_SEND phase of the outgoing chain.

As a result, any interaction/modification of the entity stream by the user gets overwritten.

My solution is to create a proxy OutputStream to create a entity stream that the user can interact with and holds a reference that we can set our URLConnectionWrappedOutputStream to later.


WhiteCat22:
#link 2444

WhiteCat22:
build: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_WiMLMCO8EeiwitX0Zm6mOA

Issue #2601, opened 2018-03-09
Better handle JDBC driver throwing SQLException for unsupported methods


alexsm82:
According to the JDBC specification, JDBC drivers should throw a SQLFeatureNotSupportedException for unsupported optional methods.  However, we have found several JDBC drivers instead throw a SQLException.  While we have attempted to handle this in our code we have found that it is not handled for the DataDirect OpenEdge JDBC Driver.

PR #2607, opened 2018-03-09
Check SQLException error message for unsupported methods

alexsm82:
Although JDBC drivers should throw a SQLFeatureNotSupportedException when they don't support optional JDBC methods, some throw a SQLException instead and mention that the method is not supported in the message.  This PR adds a check for the string "support" when a JDBC driver throws a SQLException for an optional JDBC method, so that Liberty can handle this behavior.

Issue #2601


alexsm82:
#run-libby

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
There were two failures in the personal build, both of which are unrelated and connect to existing defects.

Issue #2620, opened 2018-03-12
When building lib.index.cache, preserve original jar names


sebratton:
The lib.index.cache currently renames the jars to sha-1 hashes of their contents. This is a little confusing when the jar names appear in error message. Preserve the original jar names in a folder named after the hash.

tjwatson:
This issue is fixed in master.

PR #2621, opened 2018-03-12
Change the lib-index-cache to preserve original jar names inside a di…

sebratton:
…rectory named for the jars's hash.


sebratton:
#build

sebratton:
fixes #2620

PR #2630, opened 2018-03-12
verify cached values in testConcurrentPutNewAttributesAndRemove

njr-11:
Enhance the testConcurrentPutNewAttributesAndRemove test case to confirm that session attribute values after the concurrent updates are consistent with what is actually in the cache.


njr-11:
#build

njr-11:
commit included in #2653, closing this pull

PR #2635, opened 2018-03-12
disallow advanced session manager properties and use default values

njr-11:
Disallow and set default values for session manager advanced properties that were available on httpSessionDatabase but which we are not externalizing on httpSessionCache.  These could later be exposed as the need arises, but if we do so, some could use better names.  Also, scheduleInvalidation isn't an advance property, but seems redundant because it can be inferred from the presence of configuration of any scheduled invalidation hours, so I'll omit that one as well and compute it that way.


njr-11:
#build

PR #2636, opened 2018-03-12
Use version qualifier for zip output for consistency with bundles

ewhoch:
previously a zip package could have a label distinct from the Bundle-Version and server version. This uses the same version qualifier that those do.

#build


ewhoch:
The build didn't quite work but i think it's because version qualifier is only set to the build label if we're in is.release mode. Am still confident this will work once it hits a release build. Have at least proven it doesn't regress.

PR #2640, opened 2018-03-12
Stop SourceManger from removing the BufferManager from LogSource or TraceSource

### LINKED ISSUE #2638: SourceManager removes bufferManager from special sources [logsource, tracesource]
Source Manager removes bufferManager from special sources [logsource, tracesource]:

If an asynchronous handler subscribes to the log source and trace source and then unsubscribes it this causes the sourceManager to remove the buffeManager from the source.

This causes an issue with with existing synchronouse handlers that are "subscribed" to these sources OR if an async handler subscribes to the message source or trace source again because a BufferManager is not assigned to the source again which causes messages/traces to not go through.

Channyboy:
#build 
fixes #2638


Channyboy:
#build

Channyboy:
Failures in the recent build do not relate to the changes made in this PR

PR #2616, opened 2018-03-10
Clean up test file before+after tests for windows on Oracle JDK

aguibert:
When building OpenLiberty on Windows I hit the following error:

```
java.lang.AssertionError: Repository with no file reported available
	at com.ibm.ws.repository.connections.test.SingleFileRepositoryConnectionTest.testNoFileCheckStatus(SingleFileRepositoryConnectionTest.java:45)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.runTestClass(JUnitTestClassExecuter.java:114)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.execute(JUnitTestClassExecuter.java:57)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassProcessor.processTestClass(JUnitTestClassProcessor.java:66)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:32)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:93)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.processTestClass(TestWorker.java:108)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:35)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:146)
[...]

PR #2645, opened 2018-03-13
Ensure SingleFileClient closes the input stream after reading

Azquelt:
Also update the test to ensure the test file gets deleted as problems
here can disrupt other tests.

Follow on from #2616


Azquelt:
#build

Azquelt:
@aguibert Can you check that this fixes the issue you were seeing building locally on Windows + OpenJDK in #2616?

I've added an assert to the file deletion in the test so you should get a clear exception if there's any issue removing the file at the end of the test.

Azquelt:
#build

Azquelt:
A fix for this was done in #3234

PR #2649, opened 2018-03-13
fix user name in generated token

brutif:


brutif:
#build

brutif:
pb is good, failures are unrelated.

PR #2653, opened 2018-03-13
test should poll cache instead of doing lengthy sleep

njr-11:
I noticed a 45 second sleep in SessionCacheTimeoutTest.testServletTimeout.  If we instead poll the cache for the condition to be met, we can reduce the wait to between 0 and 30 seconds depending on when the next cycle is.


njr-11:
#build

njr-11:
Build was abandoned for unknown reason. Submitting another:
#build

njr-11:
#build

njr-11:
Added commit for test updates that were needed after merge.  The merged-in permissions were hard coded to an app location of dropins whereas this pull had moved the application to apps in order to make it configurable.  Also, uri needs to be provided with accessing the cache from the application due to changes that the other pull made to add a uri instead of taking the default.
#build

njr-11:
The latest build, njr-11-2653-20180315-1454, only has 1 test failure and it matches a pre-existing failure in unrelated componentry.  This can be merged once the stream is opened up for deliveries again.

PR #2654, opened 2018-03-13
Extend CDI package import range to include CDI 2.0

aguibert:


ewhoch:
This should be OK to merge buildless, going to do that...

PR #2661, opened 2018-03-14
Import CDI packages with range [1.1,2.1)

aguibert:


aguibert:
Merging this PR without a personal build because we suspect the issue effecting the release builds will be resolved by this PR and therefore need to get it merged in asap.

PR #2665, opened 2018-03-14
always persist cdi session scoped

benjamin-confino:


benjamin-confino:
#build

NottyCode:
I thought the Weld attributes followed a convention of WELD_S#<number> where number starts at 0. If that is the case would it be faster to first check to see if there is a WELD_S#0 first and only do the logic if there is? Otherwise for a big session you are doing a lot of iterating if there aren't any session scoped items.

benjamin-confino:
This issue was eventually dealt with in weld itself.

PR #2629, opened 2018-03-12
Issue #2459 - create mp config 1.3

tevans78:
This PR is really for #2459 but also happens to fix #2507 
fixes #2507 
#build

PR #2666, opened 2018-03-14
2459 server xml config revert revert

### LINKED ISSUE #2459: MP Config: Investigate how to access elements from the server.xml
With reference to #2442 work out how to access ~variables and~ elements from the server.xml

~`<variable>` elements end up in Liberty's VariableRegistry~
`<appProperties>` elements nested inside `<application>` elements will end up in the OSGi Admin Service

edit: this issue is for the new appProperties element. #2643 will cover using variables.

tevans78:
This fixes #2459 
But the original PR (#2629) was reverted (#2662)
This is the revert-revert plus a fix


tevans78:
#build

Issue #2651, opened 2018-03-13
[spring boot feature] warm start uses old server configuration from previous launch


tjwatson:
When starting from a warm pre-populated cache it looks like the configuration admin is still using old configurations which were configured by using a OSGI-INF/wlp/defaultInstances.xml file (see com.ibm.ws.app.manager.springboot.internal.SpringBootApplicationImpl.SpringBootConfigImpl.installVirtualHostBundle(Bundle, ServerConfiguration)).  This should not happen because we uninstall the bundle which contains the configuration when the application is stopped.  We also look for any configuration bundles when the handler is activated and uninstall any that may have been left behind before starting the applications.

There may be something being cached in the configuration admin service which is not getting cleaned up properly when the configuration bundles are uninstalled.

The improper behavior that can be observed is that the web container reports that a web application is available at a specific URI before the spring application context has been fully initialized.  We must not start listening on the endpoint for the application until the spring application context is fully initialized.

tjwatson:
This is merged into master now.

PR #2679, opened 2018-03-14
Uninstall leftover app configurations on restart

tjwatson:
Must avoid uninstalling application configurations when
stopping an application during server shutdown.  This
is because config admin will not process the removal if
the framework is shutting down.  Instead the config bundle
must remain installed and will be cleaned up on restart.
This way config admin will have a chance to clean out
the application configuraiton before any applications
start.


tjwatson:
#build

tjwatson:
#build

tjwatson:
This is to fix issue #2651

PR #2686, opened 2018-03-15
2632 allow multiple app properties

### LINKED ISSUE #2632: MP Config 1.3: Allow multiple appProperties elements and don't re-use OSGi Configuration objects
Current code only allows for a single appProperties element...
```
<application>
   <appProperties myKey1="myValue1" myKey2="myValue2"/>
</application>
```
but should allow for multiple elements and merge the result
```
<application>
   <appProperties myKey1="myValue1" myKey2="myValue2"/>
   <appProperties myKey3="myValue3" myKey2="myValue4"/>
</application>
```
This probably results in multiple OSGi Configuration objects, the properties for which need merging. It also means that new elements can be added or old ones removed. Therefore, we need to look up all of the Configuration objects every time.

tevans78:
fixes #2632 
#build

### DIFF
--- a/dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/sources/AppPropertyConfigSource.java
+++ b/dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/sources/AppPropertyConfigSource.java
@@ -16,6 +16,7 @@ import java.util.Dictionary;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.SortedSet;
 
 import org.osgi.framework.BundleContext;
 import org.osgi.service.cm.Configuration;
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.microprofile.config.1.3_fat/test-applications/serverXMLApp/src/com/ibm/ws/microprofile/config13/serverXML/web/ServerXMLBean.java
```diff
@@ -48,9 +48,15 @@ public void appPropertiesSPITest() throws Exception {
         builder.addDiscoveredConverters();
```
Azquelt: Javadoc for this test says
> Tests the SPI that allows the priority to be set
Not sure what that's about.

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/sources/AppPropertyConfigSource.java
```diff
@@ -32,21 +33,21 @@
 public class AppPropertyConfigSource extends InternalConfigSource implements DynamicConfigSource {
```
Azquelt: You've added code to define the way that properties override each other if the same property is specified more than once.

It might be worth mentioning that in the class javadoc.

Could also include a sample of what the server.xml looks like.

tevans78:
review updates were javadoc only. previous build was good.

PR #2695, opened 2018-03-15
add gradle tasks in cnf/build.image to generate single json and maven…

rampras3:
… artifact from gaFeatures


gmarcy:
This cannot be delivered in it's current form, as the code depends on unapproved 3rd party json libraries that are not licensed under acceptable terms.

Issue #2700, opened 2018-03-16
MP FaultTolerance: Add some trace back into the TCK


tevans78:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=252460

In RTC 252460 there is another TCK undeploy error. However, the message.log for the server does show an unusually long gap when stopping the application. That could be a problem with our code. It could actually be the same as the config shutdown issue where the async threads don't die quickly enough.

tevans78:
Using this issue to add some FT and Config trace back in for the TCK

tevans78:
This issue is closed ... until the failure occurs again and we can look at the trace.

PR #2703, opened 2018-03-16
Issue #2700 - turn faulttolerance trace back on to help with RTC 252617

tevans78:
trace to help diagnose RTC 252617 / #2700 
also fixes some incorrect trace guards
#build


Azquelt:
The trace changes remove a couple of places where we have the condition `TraceComponent.isAnyTracingEnabled() && tc.isWarningEnabled()` in a guard around logging a warning message.

This doesn't make sense as Warnings are not trace. `isAnyTracingEnabled` will only be true if `fine` logging in enabled for some module.

Thus the old code results in warnings that only appear if tracing has been enabled somewhere (possibly in a completely unrelated module).

PR #2705, opened 2018-03-16
Direct stderr to the same output file when running Maven in the build

### LINKED ISSUE #2506: Current External Upsteam Liberty Arquillian Container 'Hangs' on exit of last test in Liberty FAT tests
In my experience the Fault Tolerance TCK hangs 9/10 times and I have also seen the Config TCK hang.
Track shows entry of final 'stop' method exit.

Azquelt:
This fixes a problem where _fat_tck projects would hang if they
outputted enough data to stderr. If nothing was consuming stderr, the
surefire thread that was processing messages would hang trying to write
to it and stop processing messages. This would either result in the
whole build hanging, or in surefire reporting that the forked vm did not
shut down properly at the end of the test run.

Fixes #2506


Azquelt:
#build

Azquelt:
I hadn't seen `redirectErrorStream()`, I think that makes more sense.

Azquelt:
#build

PR #2706, opened 2018-03-16
trace all jCache operations and use better cache names

njr-11:
I've noticed that when debugging errors, it is difficult to identify which operations are being invoked on the JCache provider in the Liberty trace and what is the result of those operations.  This pull attempts to help in this area by explicitly recording entry/exit trace of each operation we invoke on JCache.

Also, while a teammate was looking at this, they asked for a rename of the 'info' cache to 'meta' to avoid confusion.  The session attributes cache also needs a rename because referring to it as session properties rather than attributes has confused a number of people.  I'll cover the rename under this pull as well.


dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
```diff
@@ -112,60 +128,84 @@
...
-        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
-            Tr.debug(tc, "Using session info cache", sessionInfoCache);
+        tcSessionMetaCache = "MetaCache" + Integer.toHexString(System.identityHashCode(sessionMetaCache));
+        if (trace && tc.isDebugEnabled())
+            tcReturn(cacheStoreService.tcCacheManager, create ? "createCache" : "getCache", tcSessionMetaCache, sessionMetaCache);
+
+        // Session Attributes Cache
 
-        String propCacheName = new StringBuilder(24 + a.length()).append("com.ibm.ws.session.prop.").append(a).toString();
+        String propCacheName = new StringBuilder(24 + a.length()).append("com.ibm.ws.session.attr.").append(a).toString();
```
alexsm82: Update the variable name here to attrCacheName to match the updated cache name.

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
```diff
@@ -553,7 +664,15 @@ protected void insertSession(BackedSession session) {
      */
     @Override
     protected boolean isPresent(String id) {
-        return sessionInfoCache.containsKey(id);
+        final boolean trace = TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled();
```
alexsm82: Just nitpicking, but it is kind of inconsistent that in some methods in this class trace = TraceComponent.isAnyTracingEnabled(); but in other methods trace = TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled();.  It may be easier to maintain in the future if you switch to one or the other, at least per class.

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
```diff
@@ -876,8 +1094,18 @@ private void processInvalidListeners() {
      */
     @Override
     protected BackedSession readFromExternal(String id) {
+        final boolean trace = TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled();
```
alexsm82: As in my second comment, another spot where a different value for trace is used.

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
```diff
@@ -896,22 +1124,104 @@ protected BackedSession readFromExternal(String id) {
...
+        } else if (result.length > 0) {
+            byte[] bytes;
+            if (result[0] instanceof byte[]) {
+                bytes = (byte[]) result[0];
+                b.append("byte[").append(bytes.length).append("]");
+            } else {
+                bytes = null;
+                if (result[0] instanceof Collection)
+                    b.append(EOLN).append(' ');
+                b.append(result == null ? null : result[0]);
```
alexsm82: It isn't hurting anything, but I don't think this null check is needed, since it is in the else block of an if statement checking whether the result was null.

njr-11:
#build

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
njr-11: updated here and also updated other variables using the term property instead of attribute

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
njr-11: updated

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
njr-11: updated

dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheHashMap.java
njr-11: removed the redundant null check.
I added a commit with this and the other changes above

Issue #2701, opened 2018-03-16
[spring boot feature] JSP support


tjwatson:
Need to make sure JSP support works with a spring boot application.  So far all the examples I see of Spring Boot applications that use JSP are packaged as a WAR (but the WAR is executable).  I'm not sure if that is a hard requirement to using JSPs in Spring Boot but we need to make sure that path works.

I know one issue is that the JSP implementation will not find the TLDs from the spring libraries because the container used for the spring boot application no longer has the jars located under the WEB-INF/lib/ folder.  They all are moved to the library cache.  May need to modify the JSP implementation to find the TLDs without scanning the container for library jars.  Instead we could have some info object for the containers to scan similar to the info object web uses to scan for web annotations.  I would prefer not to use the same web info object we do not fully populate that in spring boot because we want to avoid scanning for web annotations.

PR #2715, opened 2018-03-19
Spring boot jsp

tjwatson:
Fix for issue #2701


tjwatson:
#build

tjwatson:
Note that the JSP support for Spring Boot applications required the introduction of a new interface:

`com.ibm.ws.container.service.app.deploy.extended.TagLibContainerInfo`

This is similar to the interface `com.ibm.ws.container.service.app.deploy.WebModuleClassesInfo` that is used by the web container to locate containers for scanning for web annotations.  I stuck the new interface in `com.ibm.ws.container.service.app.deploy.extended` because that package is not SPI like the `com.ibm.ws.container.service.app.deploy` package.  For the first pass at this I did not want to introduce SPI.  The `TagLibContainerInfo` implementation can be stored in the non-persistent cache of a module's container (again just like the `WebModuleClassesInfo` interface).  If found then the JSP implementation will use the containers returned by the `TagLibContainerInfo` to scan for tag libraries instead of searching for the libraries to scan under the default WEB-INF/lib/ folder.

This is necessary for Spring Boot because the libraries may not be located under the WEB-INF/lib/ folder at all.  And for our Spring Boot implementation we store the libraries required by the application into an external storage area that is separate from the application itself.  That makes it impossible to scan the application module's container itself for the libraries to search for JSP tag libraries.  The Liberty Spring Boot implementation knows exactly where the application libraries jars are located and what should be scanned.  A new `TagLibContainerInfo` is created by the Liberty Spring Boot implementation and stored in the non-persistent cache so the JSP implementation can locate the information and use it if present.  The normal path (and behavior) for other EE application types (e.g. WARs) is unchanged.

PR #2720, opened 2018-03-19
disable portion of test with intermittent failure while being investigated

njr-11:
SessionCacheOneServerTest.testConcurrentGetSetAndRemove has some test logic at the end of the test that intermittently fails.  Now that the failure has been identified and traces/logs collected during an occurrence of the failure and the proper team contacted to look into it, I'll be disabling that part of the test while the investigation is underway so that it stops causing trouble to automated FAT that runs with everyone's builds.  This part of the test can be re-enabled along with whatever fix ends up being found for the issue.


njr-11:
#build

PR #2730, opened 2018-03-19
Find latest version in maven repo for replacements

fritze2:
Rather than failing to generate the bootstrap, sort the distinct versions and pick the most recent version.


fritze2:
#build

PR #2734, opened 2018-03-19
simplicity objects for httpSessionCache

njr-11:
Create simplicity objects to model httpSessionCache config element and nested properties

### DIFF
--- a/dev/fattest.simplicity/src/com/ibm/websphere/simplicity/config/ServerConfiguration.java
+++ b/dev/fattest.simplicity/src/com/ibm/websphere/simplicity/config/ServerConfiguration.java
@@ -65,6 +65,9 @@ public class ServerConfiguration implements Cloneable {
     @XmlElement(name = "httpSession")
     private HttpSession httpSession;
 
+    @XmlElement(name = "httpSessionCache")
+    private ConfigElementList<HttpSessionCache> httpSessionCaches;
+
     @XmlElement(name = "httpSessionDatabase")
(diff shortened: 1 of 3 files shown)


dev/fattest.simplicity/src/com/ibm/websphere/simplicity/config/ServerConfiguration.java
```diff
@@ -65,6 +65,9 @@
     @XmlElement(name = "httpSession")
     private HttpSession httpSession;
 
+    @XmlElement(name = "httpSessionCache")
+    private ConfigElementList<HttpSessionCache> httpSessionCaches;
```
alexsm82: Is there a reason this is a list?  My understanding was that there would only ever be one HttpSessionCache element in the server configuration, and neither httpSession nor httpSessionDatabase are lists.

njr-11: While httpSessionCache is a singleton, it is perfectly valid to configure multiple elements, and they should merge.  Without the ability to define multiple, it would be impossible to cover merging scenarios, if someone wanted to test that.  So the approach of using a list is more correct.

njr-11:
#build

njr-11:
build has 2 test failures in com.ibm.ws.security.audit.fat.config.AuditUniqueServerIDTest and multiple in com.ibm.ws.microprofile.openapi_fat_tck, which are pre-existing failures and unrelated to these updates.

PR #2744, opened 2018-03-20
create fat bucket for config updates to httpSessionCcahe

njr-11:
Create a  new automated FAT bucket where we will cover configuration updates to httpSessionCache config while the server is running.  Initially, there will only be a basic test to ensure the bucket was set up correctly.  Subsequent pulls will add the meaningful tests.


njr-11:
#build

PR #2751, opened 2018-03-20
Call send listeners once writing thread has become idle.

ngmr:
This happens when there is no more data to write to the
network, or when a call to write has gone asynchronous,
freeing up this thread to safely call the send listeners
for messages whose data have been completely sent.

The surrounding code has also been refactored for clarity
and to eliminate code duplication.


ngmr:
Fixes #2332

ngmr:
#build

PR #2824, opened 2018-03-23
Create Fault Tolerance 1.1 feature

Azquelt:
Also create an autofeature for the planned integration with MP Metrics.

Things which might need changed or considered for future work
* I've included a copy of the latest faulttolerance API - is that the most sensible course of action?
  * I assume we're allowed to do that
* I've changed one of the FAT tests to also run against FaultTolerance-1.1 so that we at least have a sniff test but we should decide which tests we want to run against the new feature, as well as adding tests for new functionality.

I did some work prototyping integration with mpMetrics which helped iron out some mistakes with the autofeature, but the actual code isn't ready yet and isn't included in this PR.


Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

tevans78:
@Azquelt Are you working on your day off? If so, you have a merge conflict ... all the feature files were moved around!

Azquelt:
Not, but I saw my build started and hadn't finished so assume something broke over the weekend and I'd like to know whether it's broken when I get back.

Pity about the merge conflict, I had heard something about reorganising the feature field so they don't all have their own projects, which makes sense but the timing is a little unfortunate.

Anyway, I'm not really working and don't have my laptop so I won't be doing any imminent fixing 😉

Azquelt:
You should be able to push more commits onto my branch if you or someone else needs this resolved before I get back.

Azquelt:
Oh, is the merge conflict the reason the build isn't completing? If so we should fix the build so it'll still build my branch despite the conflict 😐

tevans78:
Yeah, looks like it, I will try to do the merge. Also, your branch name has a dot in which is a bad idea if i remember correctly since that is used as part of the build version and it gets confused by the extra dot!

tevans78:
This PR has been superceded by https://github.com/OpenLiberty/open-liberty/pull/2834

PR #2851, opened 2018-03-27
Fixes #2683: Add env var mapping

### LINKED ISSUE #2683: MP Config 1.3: Implement "define a mapping rule from config property name to environment variable"
https://github.com/eclipse/microprofile-config/issues/264

has all the details.

neilgsyoung:


neilgsyoung:
Fixes #2683

neilgsyoung:
#build

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/sources/EnvConfig13Source.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Enhancements to the EnvConfigSource class for MP Config 1.3.
+ */
+public class EnvConfig13Source extends EnvConfigSource {
+
+    private static final TraceComponent tc = Tr.register(EnvConfig13Source.class);
+    private Pattern p = null;
+
+    public EnvConfig13Source() {
+        p = Pattern.compile(ConfigConstants.ALLOWABLE_CHARS_IN_ENV_VAR_SOURCE);
```
tevans78: I would have thought that this could be statically compiled, not in the constructor?

tevans78: I wonder if we should also override the name of this config source to say 1.3 (just like you have done for the toString())??

tevans78: (The spec isn't that clear on how unique the names should be)

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/sources/OSGiConfigUtils.java
```diff
@@ -213,8 +213,11 @@ public static String getApplicationConfigFilter(BundleContext bundleContext, Str
             String applicationFilter = getApplicationConfigFilter(bundleContext, applicationName);
             ConfigurationAdmin admin = getConfigurationAdmin(bundleContext);
             Configuration[] osgiConfigs = admin.listConfigurations(applicationFilter);
-            for (Configuration cfg : osgiConfigs) {
-                configSet.add(cfg);
+
+            if (osgiConfigs != null) {
```
tevans78: huh, didn't realise that it was possible for this to be null ... the check doesn't hurt though

dev/com.ibm.ws.microprofile.config.1.3_fat/publish/servers/MapEnvVarServer/bootstrap.properties
```diff
@@ -0,0 +1,3 @@
+bootstrap.include=../testports.properties
+osgi.console=7773
+com.ibm.ws.logging.trace.specification=APPCONFIG=all
```
tevans78: I realise that this was copied from the other test but I think the new strategy was to leave trace turned off unless we really need it. Perhaps we should turn it off in both new tests?

dev/com.ibm.ws.microprofile.config.1.3_fat/publish/servers/MapEnvVarServer/server.env
```diff
@@ -0,0 +1,5 @@
+MPCONFIG_FATTEST_ENV_VARIABLE=mpconfigtestValue
+__MPCONFIG_FATTEST_ENV_VARIABLE=__mpconfigtestValue
+MPCONFIG_FATTEST_ENV_VARIABLE__=mpconfigtestValue__
+MPCONFIG_FATTEST__ENV_VARIABLE=mpconfig__testValue
+MPCONFIGFATTESTENVVARIABLE=mpconfigtestValue
```
tevans78: don't you need some lower case and mixed case variables to test as well?

tevans78: and then some other non-alpha numerics (e.g. dot)

tevans78: so some things found on the 1st pass, some on the 2nd and some on the 3rd

dev/com.ibm.ws.microprofile.config/src/com/ibm/ws/microprofile/config/interfaces/ConfigConstants.java
```diff
@@ -40,5 +40,6 @@
     public static final long DEFAULT_DYNAMIC_REFRESH_INTERVAL = 500;
     //The minimum refresh rate for dynamic sources ... if the refresh rate is set lower than this then it will be overridden by this value
     public static final long MINIMUM_DYNAMIC_REFRESH_INTERVAL = 500;
-
+    //The regex to specify allowable characters in an Environment Variable Config Source
+    public static final String ALLOWABLE_CHARS_IN_ENV_VAR_SOURCE = "[^A-Za-z0-9_]";
```
tevans78: while creating a separate constants file for 1.3 might be over-kill, the comment should probably make it clear that it is only used in 1.3

dev/com.ibm.ws.microprofile.config/src/com/ibm/ws/microprofile/config/sources/EnvConfigSource.java
```diff
@@ -76,4 +76,10 @@ public String run() {
     public String toString() {
         return "Environment Variables Config Source";
     }
+
+    @Override
+    public String getValue(String propertyName) {
+        return super.getValue(propertyName);
```
tevans78: seems a bit superfluous?

dev/com.ibm.ws.microprofile.config/src/com/ibm/ws/microprofile/config/sources/DefaultSources.java
```diff
@@ -43,6 +43,20 @@
...
+
+    /**
+     * Add resources of name {#link ConfigConstants.CONFIG_PROPERTIES} to a List of sources using
+     * the classloader's loadResources method to locate resources.
+     *
+     * @param classloader
+     * @param sources
+     */
+    public static void addPropertyConfigSources(ClassLoader classloader, ArrayList<ConfigSource> sources) {
+
```
tevans78: I'm clearly in a picky mood ... could this be a `get` method rather than an `add`? i.e.
`public static List<ConfigSource> getPropertyFileConfigSources(ClassLoader classloader)`

tevans78:
although I have put a bunch of comments in, the main reason for not approving right now is that I think you need more tests … to make sure you cover the 1st, 2nd and 3rd pass

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/sources/OSGiConfigUtils.java
neilgsyoung: The check is required. In the absence of appProperties attributes in the server.xml application element I see NPEs.

dev/com.ibm.ws.microprofile.config.1.3_fat/publish/servers/MapEnvVarServer/server.env
neilgsyoung: I have lots of interesting property names that get mapped to env vars in MapEnvVarServlet. I think that's the focus of the feature but I'll add a couple of new variables in server.env. I *think* they should work cross-platform.

neilgsyoung:
Ready for re-review.

neilgsyoung:
#build

tevans78:
Personal build was still pretty bad, running again.
#build

PR #2859, opened 2018-03-27
Repackage Commons Compress as com.ibm.ws.* package

andymc12:
Various kernel boot utilities like server package minify require the
Apache Commons Compress APIs.  However, by putting them in the kernel
boot bundle (which is also used in the classpath to start the Liberty
server) it ends up on the classpath for applications.  This causes
unintended classloading behavior when customer applications attempt to
package their own version of Commons Compress.  In order to ensure that
we can use it, and ensure that user apps can't load it, we need to 
re-package these classes into com.ibm.ws.* packages.


andymc12:
#build

hlhoots:
#build

hlhoots:
The latest build (andymc12-2859-20180329-1404) above, had 2 errors that looked to be slow downs during the build.

PR #2864, opened 2018-03-27
add sessionCache feature while server running

njr-11:
Add a test case that starts the server without the sessionCache-1.0 feature, uses a session, and then adds the feature at a later point, verifying that afterwards session updates are persisted.


njr-11:
#build

njr-11:
build has 3 test failures, which are pre-existing issues in unrelated components

PR #2880, opened 2018-03-28
config update to httpSessionCache uri

njr-11:
Create a test case that starts the server with an invalid uri for the httpSessionCache, then makes a configuration update to correct it, then makes a subsequent configuration update to specify it in a different way.  Validate that session data is persistent in the latter two cases.


njr-11:
#build

njr-11:
#build

njr-11:
#build

Issue #2884, opened 2018-03-28
NPE in push_promise error test


loriadi:
One of the error testcases for push_promise is generating an NPE

PR #2885, opened 2018-03-28
Unexpected NPE in push_promise error testcase

loriadi:
The error case test where a push_promise frame is sent from the client is causing an NPE during test execution.


loriadi:
Fixes #2884

loriadi:
#build

loriadi:
#build

loriadi:
#build

PR #2896, opened 2018-03-29
test application closes CachingProvider

njr-11:
Write a test case where an application closes the CachingProvider for the configured JCache provider library.  We need to update the session cache code, if possible, to use a different CachingProvider instance so that we are isolated from application code such as this.  The test will initially be failing because we don't have that level of isolation yet.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
build has a number of test failures, but none are in buckets that enable the sessionCache feature. The test buckets that use the sessionCache feature, which is the only code modified, all run successfully

PR #2898, opened 2018-03-29
Added code the throw a deployment exception if boolean was on the

billw1024:
onMessage signature where it is not allowed given the type of text or
binary parameter that was also on the onMessage signature, as per a
couple of CTS tests.


billw1024:
#build

billw1024:
build had 60727 test attempted with only 3 fail/errors, so 60724 passed.  failure/errors unrelated to these websocket changes.

PR #2907, opened 2018-03-30
Add feature units

ewhoch:
Adding some feature verification that runs directly against raw .feature files. This means it runs fast, and runs early so if someone is attempting to promote a new feature to GA it will crawl all dependencies for both kind and edition to make sure that the child features are not less visible than the parent. In other words, making sure a feature that "Servlet" (core) includes is not base/nd.

Leaves out the idea of autoFeatures (maybe for now, maybe always) since those are cases where a feature in a less visible mode would deliberately be left out of the direct dependency tree in order to selectively load.


ewhoch:
#build

ewhoch:
Looks like this is working fine. Merging in.
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/__T6vNTRNEein-4zb_x5ocw

PR #2919, opened 2018-04-02
force sessionCache feature to use different CachingProviding instances than applications

njr-11:
Avoid interference between sessionCache feature and applications using the same CachingProvider instance (which results in them closing eachothers' cache managers and so forth) by forcing the sessionCache feature to use a unique CachingProvider instance by wrapping the ClassLoader with which it obtains the CachingProvider.  The ClassLoader is used as key and so this will cause it not to match with application usage.


njr-11:
#build

njr-11:
Build has 2 test failures, both of which match pre-existing issues

PR #2936, opened 2018-04-02
Fix javadoc build errors on JDK 9

aguibert:


aguibert:
cancelling this PR as a dup of #2947

PR #2937, opened 2018-04-02
write tests for JCache MXBeans

njr-11:
Write tests for JCache spec defined MXBeans that can report statistics on the sessionCache-1.0 feature's usage of the the caches it creates.  One test will verify availability of each MXBean (management and statistics) for each cache, and another will verify that none of the MXBeans are available - the case where monitoring is not enabled.  The former test will be disabled for now, because we haven't implemented enablement of the MXBeans yet.  When that is done in a subsequent pull, the test can be switched on.


njr-11:
#build

njr-11:
Build was successful with all tests passing

PR #2957, opened 2018-04-04
management and statistics enablement based on monitoring enablement

njr-11:
Enable or disable management and statistics on the JCache caches used by the sessionCache feature based on whether the monitor configuration element includes monitoring for sessions.  The monitoring feature does this with a filter attribute.  If set to nothing, then all are enabled.  Otherwise it is a set of keywords, which can include Session.  If either of these are detected, then enable management and statistics.


njr-11:
commit was included under #2971, closing out

PR #2967, opened 2018-04-04
Issue 2165: Adding Cache verification invalidation tests

mswatosh:
Adding additional tests to verify the Session Cache after invalidation.


mswatosh:
#build

PR #2971, opened 2018-04-04
updating monitoring detection for multiserver

njr-11:
Enablement of statistics/management for the JCache provider used by sessionCache feature needs to be done in a way that is compatible with the possibility that multiple servers might be using the same cache.  All of the servers will need to have Session monitoring set the same (whether that is enabled or disabled) in order for the behavior to be meaningful - there is no way around that as long as monitor-1.0 is used as the triggering mechanism.  However, we need to ensure that when one server deactivates and unsets its reference to the monitor feature, that shouldn't cause disablement of collections of statistics in the cache itself.


njr-11:
#build

njr-11:
Build had all tests successful, although got stuck in a cleanup step.  Erik H agreed it can be considered good for delivery purposes.

PR #2975, opened 2018-04-05
Add FFDCIgnore to WebUtils.validateUriFormat

ayoho:
Having an FFDC for `java.security.PrivilegedActionException` could make someone believe there's a Java 2 Security problem when validating URIs when the real problem is that the URI is likely just formatted incorrectly. This change will cause that FFDC to be ignored.


ayoho:
#build

PR #2976, opened 2018-04-05
Save datasource on DBIDStore when datasourceLookup is set

### LINKED ISSUE #2966: Store dataSource on DatabaseIdentityStore when dataSourceLookup is set
When the dataSourceLookup is either a set String or an immediate EL expression, store the DataSource after it is looked up. If it is a deferred EL expression, then we will look up the datasource every time we need it.
Add extra protections on the Datasource connection. Set it to read only and set a max result.

kristip17:
fixes #2966 

Integrating testing into existing FAT tests.


kristip17:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.javaeesec_fat

kristip17:
The test failure in the PB is an existing issue. All security tests ran green.

PR #2980, opened 2018-04-05
test case to add and remove the monitor feature

njr-11:
Write a test case that adds and removes the monitor feature while the server is running, verifying that this cases management/statistics to be enabled/disabled for the caches used by HTTP session persistence


njr-11:
#build

njr-11:
#build

njr-11:
Build has 1 test failure and it is a pre-existing issue in an unrelated component

PR #2988, opened 2018-04-05
Determine API and SPI dependencies from ESAs

ericglau:
Determine API and SPI dependencies from ESAs instead of looking up from table for the feature Maven artifacts.


ericglau:
#build

ericglau:
Build failures are unrelated to the maven artifacts.

PR #2993, opened 2018-04-06
Add common security FAT web response utils class

ayoho:
Creates the `WebResponseUtils` class that can be used to get response text, status codes, titles, etc. from various web automation response objects. HtmlUnit, HttpUnit, and other web automation tools wrap requests and responses differently, so this is intended to provide a single, standard front end to extract useful information from these objects regardless of which testing framework is being employed.


ayoho:
#build

PR #2996, opened 2018-04-06
extending time for timeout for defect 252705

janamanoharan:


arturdzm:
Looks like this is fixed by adding global tck timeout settings

PR #3008, opened 2018-04-08
config update test for scheduled invalidation

njr-11:
Test case to update config to enable scheduled invalidation.
Given that scheduled invalidation happens on the hour, it won't be practical to wait for it to happen in the test, but we can validate that off-schedule invalidation doesn't cause the cache to be updated.


njr-11:
#build

PR #3020, opened 2018-04-09
useInvalidatedId should not be configurable

njr-11:
The useInvalidatedId option supported by httpSessionDatabase is not being externalized for httpSessionCache, so we should not permit it to be configurable.  This will better allow for it or a similar capability to be added in the future if there is a desire to do so.


njr-11:
#build

njr-11:
build has 1 test failure and it is a duplicate of a pre-existing issue in unrelated componentry

PR #3035, opened 2018-04-09
ready for build testing

gmarcy:
Add buildtools projects to OL workspace from the corresponding ant_* projects in CL.
Move all ant taskdefs into a common gradle file so we can update dependencies in just one place.
Fix up missing dependencies so that more tasks can be run right after the workspace is initialized.


gmarcy:
#build

PR #3012, opened 2018-04-09
Bells config

NottyCode:
Allow the META-INF/services file to have service properties defined. 
When reading META-INF/services process comment lines before the class and extract name=value from the comments.

PR #3051, opened 2018-04-10
Bells config - replacing #3012

NottyCode:
Allow the META-INF/services file to have service properties defined.
When reading META-INF/services process comment lines before the class and extract name=value from the comments.


NottyCode:
#build me please Libby

Issue #1827, opened 2018-01-26
Life cycle of cache for http sessions


njr-11:
User story:
"As a user of HTTP sessions, I want the life cycle of the session cache to be managed such that when the httpSessionCache feature shuts down on one server but others remain, the cache remains available/usable."
"As a user of HTTP sessions, I want the life cycle of the session cache to be managed such that when the only httpSessionCache feature/server that is using it shuts down, the cache is properly shut down and does not continue to remain around forever."

PR #3054, opened 2018-04-10
workaround for session manager stop overlapping deactivate

njr-11:
When a configuration update is made to httpSessionCache while the server is running, the service is deactivated and a new instance activated in order to honor the configuration update.  The session manager, which depends on the CacheStoreService, reacts by restarting all applications and performing other processing such as ensuring that any deferred writes are made to the persistent store.  Unfortunately, this logic overlaps the CachStoreService deactivate which closes the CachingProvider and CacheManager, making the concurrently executing processing of the deferred writes (and anything else that requires the persistent store fail).  A couple of lines of the config update test case are commented out for this reason.

Under this pull, we will be applying the same workaround that is used by DatabaseStoreService which also experienced this issue.  The workaround involves holding up the progress of deactivate until the other processing has finished.  This is not an ideal solution, and I would like to see this further looked into under user story #1827 which covers the life cycle of the CachingProvider, CacheManager, and Cache instances.

### DIFF
--- a/dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
+++ b/dev/com.ibm.ws.session.cache/src/com/ibm/ws/session/store/cache/CacheStoreService.java
@@ -17,6 +17,7 @@ import java.security.PrivilegedAction;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build

PR #3062, opened 2018-04-11
Use system.lineSeperator

benjamin-confino:


benjamin-confino:
#build

benjamin-confino:
This fixes the following error: 

[ERROR   ] SRVE0315E: An exception occurred: java.lang.Throwable: java.security.AccessControlException: Access denied ("java.util.PropertyPermission" "line.separator" "read")

PR #3092, opened 2018-04-12
Check in again JsonTraceService, BaseTraceService Merger + Fixes

Channyboy:
BASIC. Otherwise this could cause non-message source events to be
processed for output if these sources were enabled before in JSON format
configuration.
- Removed ThreadLocalhandler from LogSource and Trace Source. This would
cause Handler events (i.e. LogstashCollector) will not be emitted into
messages.log or console in BASIC. A subsequent change to this is to
introduce a recursion counter for any message event that is passed on to
LogSource.
- Removed the synchronized block around the message output in the
synchronizedWrite method in consoleLogHandler. This is to avoid a
dead-lock.
- Refactor to use LogFieldConstants' constants in LogSource,
TraceSource, ConsoleLogHandler, BaseTraceService,  and MessageLogHandler
- Use of (KeyValuePair) kvp.get*Value in BaseTraceFormatter. Old code
simply called kvp.getValue() and it was the responsibility of the caller
to cast and convert.
- In ConsoleLoghandler, when formatting BASIC console output, it filters
and suppresses stack traces. The formatter will return null if it is
suppressing a stack trace. The fix is to ensure that we are not printing
out 'null's out into the console.
- In messageLogHandler or consoleLogHandler, when JSON formatting is
activated and a non-message is a selected log source, a burst of
"default" formatted output of this type of source is output into the
messages.log or console when switching back to BASIC format. This is not
expected. Once in BASIC format, only message log events should be
emitted. This is caused by events that are "in-flight" through the
logging pipeline when the change to BAIC is configured. The fix is to
format into JSON any non-message source events that make it to the
handler.
- For ConsoleLogHandler JSON output, appended to the filter that checks
for consoleLogLevel to also check if the message originated form
sysout/syserr and if copysystemstreams is true. This is becauese
messages that originate from sysout/syserr while copysystemstreams are
valid even if they don't satisfy consoleloglevel.
- Add the commonMessageLogHandlerUpdates() and
commonConsoleLoghandlerUpdates() method call to
BaseTraceService.update()
- Unit Test Modifcation to CaputredOutputHolder. Need to modify the
overloaded publishTraceLogRecord to better reflect the mechanism used in
BaseTraceService.
- Modify the HPEL Logging Unit Test to use a traceSpecification oof
com.* and org.* over the default *=all. Using x.com.*=off did not work.
This is to avoid the presence of BufferManagerImpl add() Trace which
causes the test to fail.
- Keep ThreadLocalHandler's in use in TraceSource. This avoids the
triplicate of traces that occur through the collector framework.

#build 
fixes #


Channyboy:
Fat Failure is not related to changes made.
Will run one more #build

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceFormatter.java
```diff
@@ -560,6 +893,29 @@ private final String generateObjectId(Object id, boolean fixedWidth) {
         return objId;
     }
```
donbourne: this generateObjectId should just call the new generateObjectId with the value from System.identityHashCode

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceFormatter.java
```diff
@@ -736,13 +1092,100 @@ private String getMarker(LogRecord logRecord) {
...
+     * check casting to WsLogRecord
+     *
      * @return
      */
-    private WsLogRecord getWsLogRecord(LogRecord logRecord) {
+    //change to public
+    public WsLogRecord getWsLogRecord(LogRecord logRecord) {
         try {
             return (WsLogRecord) logRecord;
         } catch (ClassCastException ex) {
```
donbourne: should use instanceof instead

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceFormatter.java
```diff
@@ -736,13 +1092,100 @@ private String getMarker(LogRecord logRecord) {
...
+        String message = filterStackTraces(txt);
+        if (message != null) {
+            if (loglevel.equals("SystemErr")) {
+                message = "[err] " + message;
+            }
+        }
+        return message;
+    }
+
+    private String filterStackTraces(String txt) {
```
donbourne: should modify BaseTraceService.filterStackTraces to be public static (dealing with the instance variables that would need to be passed to it) and use that instead of copying the whole method here.

Channyboy:
last #build for concurrency of code

Channyboy:
The failing FAT is not related to the code included in this PR. Manual testing was conducted with both Oracle JDK and IBM JDK.

PR #3105, opened 2018-04-13
Init commit: logger will not delete backup files when 0 is set on max…

### LINKED ISSUE #3103: Access Log "maxFiles" attribute not working as intended with value of 0
The documentation for the Access Log's "maxFiles" attribute reads as follows: Maximum number of log files that will be kept, before the oldest file is removed; a value of 0 means no limit. However, setting the value to 0 does not cause an unlimited of backup files as intended. Instead, rollover does not occur, leaving just one current file of unlimited size.

mrsaldana:
…Files and will accept 0 as a valid value to invoke rollover when limit is reached.

Fixes #3103


mrsaldana:
#build #spawn.fullfat.buckets=com.ibm.ws.transport.http_test

Issue #3110, opened 2018-04-13
Test push method of PushBuilder from a secured and an unsecured servlet


ebreijo:

ebreijo:
Test the push method of PushBuilder from a secured and an unsecured servlet that tries to push a secured resource to the client. Note that this changes just contain the test app and part of the automation. This cannot be fully automated in the until Open Liberty includes an IBM JDK that supports ALPN.

PR #3112, opened 2018-04-13
Issue 3110: Test the push method of PushBuilder from a secured and an…

ebreijo:
… unsecured servlet that tries to push a secured resource to the client. Note that this changes just contain the test app and part of the automation.


ebreijo:
fixes #3110

PR #3118, opened 2018-04-13
config update test to remove sessionCache feature

njr-11:
Update the automated test for adding the sessionCache feature to also include removing it.  This exposes an issue where the shared library is not updated and still references previous jcache spec classes, and to work around this for now, we will need to also make an update to the library to force it to be refreshed.


njr-11:
#build

PR #3119, opened 2018-04-16
test infrastructure to capture and read server dump

njr-11:
Prepare for session cache introspector by getting test infrastructure in place to capture a server dump and read the output of an introspector.


njr-11:
#build

Issue #3093, opened 2018-04-12
CTS: include javax.enterprise.inject.se in the cdi ap jar


Emily-Jiang:
Need to include the package of javax.enterprise.inject.se in cdi api jar but just not export this package.

tevans78:
Done

PR #3137, opened 2018-04-16
#3093: add cdi se package to the cdi api jar

Emily-Jiang:
In OL, the package javax.enterprise.inject.se was not packaged in the api jar when repackaging CDI api jar. We need to package it but not export it so that all the classes in the released api jar are present.


Emily-Jiang:
#build

PR #3139, opened 2018-04-17
Expose CDI spec deployment errors in the server log

Azquelt:
When an exception is thrown during app deployment, check whether the
cause chain includes an exception defined by the CDI spec. If so, wrap
the thrown exception with a newly created instance of the CDI spec
exception.

This ensures that the CDI spec defined exception is shown in the server
log.


Azquelt:
We decided not to do this.

PR #3155, opened 2018-04-17
Create well-known Expectation classes for security FAT infra

ayoho:
Creates specific classes that extend the `Expectation` class, allowing specific expectation types to have their own behaviors (e.g. validating the expectation).


ayoho:
#build

PR #3161, opened 2018-04-17
add feature manifest generation to assemble task

gmarcy:
Run featureBnd tasks with createEsa: false during assemble task


gmarcy:
#build

ewhoch:
Pretty sure this causes an issue doing a cold releaseNeeded without any previous build, so this needs to be pulled for now....

PR #3166, opened 2018-04-18
Issue #2983: Eval config of SpringBootApplication

### LINKED ISSUE #2983: [spring boot feature] evaluate child elements of springBootApplication
The current springBootApp configuration element (to be renamed to springBootApplication see issue #2982) extends the <application/> pid `com.ibm.ws.app.manager`

The pid `com.ibm.ws.app.manager` has many child elements which may not all apply to spring boot applications.  A text search for `ibm:parentPid="com.ibm.ws.app.manager"` will find all the child elements that can be configured.  For example the `<classloader/>` element uses this.  That allows the classloader element to be configured for the application.  This enables support for configuring things like api visibility and shared libraries for the application.

Two things are needed here.
1) Evaluate if each child is applicable for spring boot applications, if not then add the child pid to `ibm:excludeChildren` attribute to prevent it from being used within the <springBootApplication/> element.
2) If the child element is applicable for spring boot applications then create a testcase that ensures it works.

dazavala:
fixes #2983 
#build

### DIFF
--- a/dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/ConfigSpringBootApplicationClassloaderTests.java
+++ b/dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/ConfigSpringBootApplicationClassloaderTests.java
@@ -1,0 +1,103 @@
+/*******************************************************************************
+ * Copyright (c) 2018 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.springboot.support.fat;
+
+import static org.junit.Assert.assertNotNull;
+
+import static componenttest.custom.junit.runner.Mode.TestMode.FULL;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.junit.After;
+import org.junit.Test;
+import org.junit.rules.TestName;
+import org.junit.runner.RunWith;
+
+import com.ibm.websphere.simplicity.config.ClassloaderElement;
+import com.ibm.websphere.simplicity.config.ServerConfiguration;
+import com.ibm.websphere.simplicity.config.SpringBootApplication;
+
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.custom.junit.runner.Mode;
+import componenttest.topology.utils.HttpUtils;
+
+@Mode(FULL)
+@RunWith(FATRunner.class)
+public class ConfigSpringBootApplicationClassloaderTests extends AbstractSpringTests {
+
+    @Override
+    public Set<String> getFeatures() {
+        return new HashSet<>(Arrays.asList("springBoot-1.5", "servlet-3.1"));
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/ConfigSpringBootApplicationClassloaderTests.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public AppConfigType getApplicationConfigType() {
+        return AppConfigType.SPRING_BOOT_APP_TAG;
+    }
+
+    // DO NOT OVERRIDE expectAppSuccess()
+
+    @Test
+    public void testSpringBootApplicationClassloaderDefaults() throws Exception {
+        server.stopServer(false);
```
tjwatson: I think we should have a more general way to do this that does not require a server restart in the test execution.  Create a new method in AbstractSpringTests:

```
public void modifyConfiguration(ServerConfiguration config) {
    // do nothing by default
}
```
Then this test class can override the method and add the classloader element to the SpringBootApplication already contained in the ServerConfiguration.
This method would get called by `AbstractSpringTests.configureServer()` just before doing the call to `server.updateServerConfiguration(config)`

dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/ConfigSpringBootApplicationClassloaderTests.java
```diff
@@ -0,0 +1,106 @@
...
+        List<SpringBootApplication> configuredApps = sc.getSpringBootApplications();
+        configuredApps.clear();
+        SpringBootApplication appConfig = new SpringBootApplication();
+        appConfig.setLocation(getApplicationFile().getName());
+        appConfig.setName("testClassloaderDefaults");
+
+        List<ClassloaderElement> configuredLoaders = appConfig.getClassloaders();
+        configuredLoaders.clear();
+        ClassloaderElement loader = new ClassloaderElement();
+        configuredLoaders.add(loader);
```
tjwatson: This seems to just test what putting an empty classloading element would do.  I expect that to just do the default behavior as if no classloading element was specified.  I think we should add some expanded visibility and verify that the expanded visibility is accessible from the application, something from a third-party API?  We probably should have a negative test endpoint in the `TestApplication` that is called from `CommonWebServerTests15`.  This endpoint should verify that some third-party API is not accessible from the application because when running `CommonWebServerTests15` the application should be using the default API visibility which excludes third-party API.

dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/ConfigSpringBootApplicationClassloaderTests.java
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import com.ibm.websphere.simplicity.config.ClassloaderElement;
+import com.ibm.websphere.simplicity.config.ServerConfiguration;
+import com.ibm.websphere.simplicity.config.SpringBootApplication;
+
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.utils.HttpUtils;
+
```
tjwatson: I think we should make this part of the FULL bucket

dazavala:
#build

dazavala:
#build

dazavala:
Added test class ConfigSpringBootApplicationClassloaderTests to the FULL bucket.  -Regards.

PR #2963, opened 2018-04-04
Lazily initialize CacheManager in sessionCache-1.0

aguibert:
Due to a race condition in dynamic the ShareLibrary implementation,
a modification event may propagate to a depending service before the
ClassLoader is updated. We have opened issue #2917 to hopefully get
this fixed.

In the immediate term, we will work around this race condition by lazily
initializing the CacheManager, instead of eagerly initializing it in
CacheStoreService activation.

Due to the number of test updates involved, this pull is being split apart and delivered in separate chunks under #3175, #3182, and other pull (add number here once opened)

PR #3175, opened 2018-04-18
minimal set of changes from Andy's #2963 in order to enable lazy init of JCache provider and cache manager

njr-11:
Andy started on some changes to enable lazy init of JCache providers (and cache manager), but they were intermixed with a number of other changes and didn't yet have sufficient updates to get the tests running cleanly.  This pull aims to deliver only the necessary updates for lazy init and updates to tests.  It should be noted that in some cases, tests were expecting different error messages relating to code points where expected failures happen on eager init, which now differ when using lazy init.  Also, the timing of some tests is thrown off by having the first operation do lazy init which makes that one operation take longer.  This was solved by updating the test to first make one access to sessions that is independent of the timing logic.  Also, the timing changes have gotten us into a situation where we much more frequently hit intermittent issues (all seem to have a single cause with extra versions of sessions being created due to synchronization issues with the last updated timestamp).  In this case, additional portions of the concurrency tests have been disabled with TODO comments indicating re-enablement should happen after that issue is fixed.


njr-11:
merged under #3182 which included this commit

PR #3182, opened 2018-04-18
config update to session cache fileset

njr-11:
Deliver Andy's test case from #2963 for updating the fileset of the JCache provider that is used by http session persistence while the server is running.


njr-11:
#build

njr-11:
#build

njr-11:
Build has two test failures, which are pre-existing issues in unrelated componentry

PR #3185, opened 2018-04-18
Update dependency for wmqJmsClient-1.1

ericglau:


ericglau:
Verified by building the project locally

PR #3200, opened 2018-04-19
add new method

brutif:


brutif:
#build

brutif:
pb is good, failure is unrelated.

PR #3202, opened 2018-04-19
increase jwtsso logged out cookie cache size to 10000 and enable cache by default

brutif:


brutif:
#build

brutif:
build had a disk space error but the tests ran and failures are unrelated,  so personal build is good.

PR #3212, opened 2018-04-19
Don't run the asserts on an excpetion if we see WELD-001210

benjamin-confino:


benjamin-confino:
#build

benjamin-confino:
This fixes RTC253554

benjamin-confino:
#build

PR #3243, opened 2018-04-23
Create CommonSecurityFat class

ayoho:
The `CommonSecurityFat` class is intended to be the fundamental base class that other FAT classes extend. The class is meant to include only the basic members and methods that should be common to _any_ security FAT such as common utility class members and logging entries/exits of test methods.


ayoho:
#build

PR #3215, opened 2018-04-20
Add new edtion and product id for Open Liberty

idlewis:
To distinguish features from WebSphere Liberty features
#build

PR #3246, opened 2018-04-23
Revert "Add new edtion and product id for Open Liberty"

wraschke:
Reverts OpenLiberty/open-liberty#3215


idlewis:
@ericglau Why are the new editions called Open and Open_Web? That is inconsistent with the existing Liberty editions which are all uppercase, e.g. 'EXPRESS'

ericglau:
@idlewis The edition for Open Liberty was already "Open".  To remain consistent with that, "Open_Web" attempts to keep the same convention.

PR #3282, opened 2018-04-24
Trace Enumeration<URL> returned by ContainerClassLoader.findResources

andymc12:
Diagnostic fix to print the URLs returned when calling `ContainerClassLoader.findResources(...)`.  Currently, we already trace the single URL returned when calling `findResource`, but without this fix, we don't have a good trace of how many or which URLs are returned in the enumeration.


andymc12:
#build

PR #3284, opened 2018-04-24
address several TODO comments in CacheHashMap

njr-11:
Clean up some of the TODO comments in CacheHashMap, including one around capping the backoff time for retries of the operation to add and remove attribute names from the main session entry.


njr-11:
#build

Issue #3278, opened 2018-04-24
Add JwtSSO credential provider


utle:
Add a new JwtSSO credential provider service to validate the subject that JwtSSOToken

PR #3303, opened 2018-04-25
add JwtSSOTokenCredentialProvider

utle:
Add a new OSGI service JwtSSOTokenCredentialProvider to validate the subject when jwtSSOToken feature enabled.


utle:
#3278

PR #3304, opened 2018-04-25
overQualLastAccessTimeUpdate is inconsistent with DatabaseHashMap

njr-11:
Update overQualLastAccessTimeUpdate to be consistent with DatabaseHashMap implementation.
Specifically, when the last access time already matches, we should return 1, not 0, for the number matching rather than the number changed.


njr-11:
#build

Issue #3250, opened 2018-04-23
[spring boot feature] Add tests for Spring boot Actuator


anjumfatima90:
We need to add tests for Spring Boot applications that use Spring Boot Actuator.

tjwatson:
Fix is in master now.

Issue #3122, opened 2018-04-16
[spring boot feature] ability to override server endpoint configuration with server.xml


tjwatson:
Today Spring Boot application support will configure an embedded server with the configuration provided by the application properties for things like the http port and ssl.  To do this IDs are generated for each of the server configurations.  For example, for the httpEndpoint, virtualHost, ssl and keystore elements in the server configuration.  And then the application is bound to the virtualHost that got generated.

This prevents any settings contained in the server.xml from being used for the various server configurations because the ID used is internal and not known by the deployer.  If we documented the IDs used and made them easy to use directly in the server.xml then we could support the IDs being used to override settings from the application with server.xml settings.

tjwatson:
Proposal:

Each embedded server created by the application gets assigned IDs for `httpEndpoint`, `virtualHost` and optionally for `ssl` and `keyStore` IDs.  The format used for the each ID is:
```
springBoot<name>-<port>
```
Where <name> is the config element name and <port> is the requested http port for the embedded server.  The following names will be used:

- `springBootVirtualHost-<port>` - The virtual host configuration, the alias will be set to `*:<port>` and the `allowFromEndpointRef` will be set to `springBootHttpEndpoint-<port>`
- `springBootHttpEndpoint-<port>` - The http endpoint configuration, optionally references `springBootSsl-<port>` if ssl is configured by the application
- `springBootSsl-<port>` - The ssl configuration, with references to `springBootKeyStore-<port>` for `keyStoreRef` and `springBootTrustStore-<port>` for `trustStoreRef`
- `springBootKeyStore-<port>` - the key store for the https endpoint
- `springBootTrustStore-<port>` - the trust store for the https endpoint

The spring boot handler will first check to see if any of these IDs exist for the configurations.  The discovery of existing configuration happens in this order:

1. Check for `springBootVirtualHost-<port>`, if found then use that and do not generate any additional configuration elements for the application (e.g. `springBootHttpEndpoint-<port>`); otherwise generate a `<virtualHost/>` configuration that matches the application settings
1. Check for `springBootHttpEndpoint-<port>`, if found then use that and do not generate any additional configuration elements for the application (e.g `springBootSsl-<port>`); otherwise generate a `<httpEndpoint/> configuraiton that matches the application settings
1. If keystore is configured by the application then check for `springBootKeyStore-<port>`, if found then use that; otherwise generate a `<keyStore/>` configuration that matches the application settings
[...]

PR #3315, opened 2018-04-26
Add support to override server config with server.xml

tjwatson:
This is to fix issues #3250 and #3122

An actuator test was added to confirm overrides work when multiple embedded servers are used.


tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

PR #3321, opened 2018-04-26
Issue #3320 - look for extensions in WEB-INF/classes/META-INF/services

### LINKED ISSUE #3320: CDI 2.0: Look for extensions in WEB-INF/classes/META-INF/services
In a WAR, we were only looking in `META-INF/services` (as it says in the spec) but the TCK puts the file in `WEB-INF/classes/META-INF/services`. Alasdair said the latter is where he would have expected it to be so changing the code to look in both places.

tevans78:
#build 
fixes #3320


tevans78:
#build

PR #3333, opened 2018-04-27
PR for Issue 2755: Delivers golden path JWT SSO FAT case

### LINKED ISSUE #2755: FAT: Initial golden path scenario(s) for JWT as SSO cookie
Deliver FATs that cover the initial golden path scenario(s) for the JWT as SSO cookie feature. This is intended to be limited to a simple successful case that verifies that a JWT cookie is created when user is successfully authenticated for a protected resource request.

ayoho:
Delivers the following golden path JWT SSO FAT scenario:
- `jwtSso-1.0` feature is enabled: Log in with valid credentials and verify that we receive a JWT SSO cookie, not an LTPA cookie

Resolves #2755


ayoho:
#build

PR #3334, opened 2018-04-27
ext.gitClone now clone is bucket is in spawn.experimentalfat

websterc87:
Changing ext.gitClone so it no longer calls an unimplemented ebcClone method.
We do not need a separate method for cloning on our production machines, 
we just need to make the clone optional based on if the bucket is experimental, so not every build will clone

#build


websterc87:
The Personal Build List above, Exercises the no-op case. Where the fat hasnt been selected to run in experimental mode

The following build: https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_E2HFsEn1EeiBcaHL18STjw
Exercises the case where the fat buckets have been set to run in experimental mode.

websterc87:
#build

websterc87:
Made changes to stop building on every single remote build
Testing with:
#build

PR #3338, opened 2018-04-27
avoid reducing last accessed time

njr-11:
Http session cache's overQualLastAccessTimeUpdate method can end up reducing the last accessed time the way it is currently written.  We can easily detect this by comparing the value to write, and avoid writing the older time.  Checked with Alan that we want to return an update count of 1 in this case, even though we didn't actually make an update, for the sake of the code using this method.


njr-11:
#build

njr-11:
#build

njr-11:
#build

PR #3343, opened 2018-04-27
Enhance TestValidationUtils in common security FAT for additional string validation

ayoho:
Adds string validation for the following cases:
- Validate that string is null
- Validate that string is not null
- Validate that string is equal to given string


ayoho:
#build

PR #3257, opened 2018-04-23
Generate and use keystore passwords

aguibert:
First part of story #1627

PR #3357, opened 2018-04-30
Grammar corrections for CWPKI0819I

aguibert:
Addresses these review comments:
https://github.com/OpenLiberty/open-liberty/pull/3257/files#r184994759


sishida:
Looking at this again, I'm thinking that making the verb negative would be better than making the noun negative. Thus...

"The default keystore is not created because **a** password is **not** configured on the <keyStore id="defaultKeyStore"/> element, and **the** 'keystore_password' environment variable is **not** set."

sishida:
Added ID review label! :)

Issue #3330, opened 2018-04-26
Whether we continue after failing process the jwtSSOTokken


utle:
When we process the jwtSSOToken and it's failed, do we want to fallback to the application authentication mechanism. The current code has flag to not allow fallback to LTPA but continue to the application authentication mechanism just like LTPA SSO scenario.

utle:
1) If there is a jwtSSOToken in a request, fallback to LTPA will not be allowed.
2) If there is NO jwtSSOToken in a request and shouldFallbackToLtpaCookie is true, fall back to LTPA will be allowed. 
if anything wrong with SSOAuthenticator, it will fallback to application authenticator

PR #3369, opened 2018-04-30
update fallback from jwt to ltpa

utle:
Alow fallback to ltpa only if there is no jwtSSOToken in the request and shouldFallbackToLtpaCookie is true.


utle:
#build

utle:
Fixed #3330

PR #3371, opened 2018-04-30
include session attribute names in session cache introspector

njr-11:
It would be helpful for the session cache introspector to include with session attributes are actually in the cache for debugging purposes.  Avoid values, which could include sensitive data, and in any case, cannot be guaranteed to deserialize from the introspector's class loader.


njr-11:
#build

njr-11:
#build

njr-11:
build has 1 test failure which is a pre-existing issue

PR #3402, opened 2018-05-02
Allow disabling individual annotations

benjamin-confino:


benjamin-confino:
#build

Azquelt:
The checks in `FaultToleranceCDIExtension` also need to be added to `FaultToleranceInterceptor` where we create the policies.

Otherwise it won't work if there are multiple annotations on one method and only some of them are disabled.

Azquelt:
Also, as it's written the cache will grow and grow because nothing is removed from it.

Azquelt:
@benjamin-confino looks good, there was one fix needed which was that the target class is usually a Weld proxy and we need to find the real class name to look up the correct config properties.

I also took the opportunity to do a little cleaning up. Could you review my changes when you have a moment?

Azquelt:
#build

benjamin-confino:
#build

benjamin-confino:
#build

PR #3430, opened 2018-05-03
Add JWT SSO FAT test for the feature not enabled

ayoho:
Adds the following test case:
- The `jwtSso-1.0` feature is NOT enabled: Invoke a protected resource and log in, ensuring that a JWT SSO cookie is not created


ayoho:
#build

PR #3441, opened 2018-05-04
update basic annotations test to drive code path

njr-11:
Update the annotations test to get it to attempt to drive the codepath where CDI would process JCache annotations if our extensions implemented them.


njr-11:
#build

PR #3447, opened 2018-05-04
Set SCC and AOT options regardless of JDK version

aguibert:
Previously we had to check the IBM JDK version to see if it was JDK 6
FP 10 or newer because this was the first release where SCC/AOT options
were supported. This check was written such that failure to parse the
JDK version would result in SCC/AOT not being set at all.

The new OpenJDK on OpenJ9 builds do not have the IBM JDK version property
file in them, so this check failed on these new builds. Since JDK 6 is
no longer supported, we can simply remove the JDK version check and set
SCC/AOT in IBM_JAVA_OPTIONS all the time regardless of JDK version.


aguibert:
#build

aguibert:
#build

aguibert:
#build

aguibert:
#build

Issue #2484, opened 2018-03-02
[spring boot feature] Support for context parameters


tjwatson:
Spring boot has properties for setting context parameters:

server.context-parameters

The type is Map<String,String>. For example:

server.context-parameters.key1=value1
server.context-parameters.key2=value2

This should result in the following ServletContext initialization parameters being set:

key1=value1
key2=value2

SpringBoot handles this in its own intializers.  We should have a testcase that verifies it works.

tjwatson:
Fixed in master

PR #3448, opened 2018-05-04
Fix override ssl test

tjwatson:
Fix for #2484

Also fixes an issue with the override SSL test and adds a test for setting the server header.

#build


tjwatson:
#build

Issue #3461, opened 2018-05-07
jwtsso token should reflect correct realm information


arunavemulapalli:
1. Build the jwt token with "realm" claim - subject credential should have this information already
2. When we are validating the token content and creating the subject, check and see if the "realm" claim exists first before using "issuer" as realm.

ayoho:
Need to follow up to see if this was included in the latest changes merged by Ut.

PR #3487, opened 2018-05-08
Issue #3461 : create jwtsso tokens with realm from the authenticated subject

arunavemulapalli:
If there is realm in the subject, then use that instead of "issuer" claim


arunavemulapalli:
fixes #3461

arunavemulapalli:
#build

ayoho:
Should have been fixed by some of Ut's changes.

Issue #3491, opened 2018-05-08
default jwtsso token issuer is not correct


arunavemulapalli:
Update default issuer logic to make sure that we use correct schema and port.

ayoho:
Ditto - check Ut's changes to see if this was included.

ayoho:
I will go back through our tests to see whether our issuer checking is accurate. The scheme and port should depend on initial request (i.e. http vs. https) and the server configuration (e.g. no SSL port).

PR #3492, opened 2018-05-09
Issue #3491 : update default issuer in the jwtsso token to use https schema and port when available

arunavemulapalli:


arunavemulapalli:
fixes #3491

ayoho:
Should have been fixed by some of Ut's changes.

PR #3509, opened 2018-05-09
Ensure mpRestClient-1.0 works with Filters in JAX-RS 2.1

andymc12:
When we upgraded the jaxrs-2.1 feature to CXF 3.2.4, that pulled in some
changes to FilterProviderInfo that are incompatible with CXF's 3.1.16
version that we use with jaxrs-2.0.  When a MP Rest Client attempts to
register a filter (when using jaxrs-2.1), the user will see a 
NoSuchMethodError for FilterProviderInfo's constructor.  

This fix changes the MP Rest Client implementation to use the newer
constructor and adds a "Stub" constructor for the 2.0 version of
FilterProviderInfo - this stub constructor should be removed when we
move to the next version of CXF 3.1.X.

This fix also adds new tests that verify that the fix works.

This is not fixing a release bug because JAX-RS 2.1 has not shipped yet.


andymc12:
#build

andymc12:
Rebased - new #build

andymc12:
Rebased on latest master - new #build

PR #3532, opened 2018-05-11
Fix leak of WABDeployedAppInfo when a wab is redeployed many times

tjwatson:
The issue is wabGroups map never clears out old WABGroup objects
when the WABGroup is empty.  This leaves a stale WABDeployedAppInfo
object in the wabGroups map. The next time the WAB is redeployed
it reuses the stale WABGroup this ends up leaking out the
new WABDeployedAppInfo even when it is not used, but should be.

The fix is to ensure the stale WABGroup is removed so it is not
reused.

This is for issue #3533


tjwatson:
A new FAT for this would be something along the lines of start/stop a bundle many times and make sure we are not leaking objects.  I'm not sure how to do that without some pretty long running test.

tjwatson:
#build

tjwatson:
#build

Issue #1175, opened 2017-12-02
Enabling javaee-8.0 gives SSL config errors out of the box


aguibert:
If someone is just getting started with Java EE 8 and OpenLiberty they typically do the following steps:
```
bin/server create myServer
# edit myServer's server.xml to enable the javaee-8.0 feature
bin/server run myServer
```

On server stratup, they will be met with the following errors:
> [ERROR   ] CWWKS9582E: The [defaultSSLConfig] sslRef attributes required by the orb element with the defaultOrb id have not been resolved within 10 seconds. As a result, the applications will not start. Ensure that you have included a keyStore element and that Secure Sockets Layer (SSL) is configured correctly. If the sslRef is defaultSSLConfig, then add a keyStore element with the id defaultKeyStore and a password.
> [ERROR   ] CWWKS9660E: The orb element with the defaultOrb id attribute requires a user registry but no user registry became available within 10 seconds.   As a result, no application will start. Ensure that you have configured an appropriate user registry in the server.xml file.

About 30 seconds after these errors pop out, a helpful AUDIT message comes along with some sample configuration that will likely resolve the errors:
> [AUDIT   ] CWPKI0817A: The default SSL configuration expects a <keyStore> element with an id value of defaultKeyStore, and a password. The defaultKeyStore <keyStore> element is missing, or the password is not specified. Either define the missing element: <keyStore id="defaultKeyStore" password="yourpassword" />, or change the default SSL configuration to use an existing keystore. See the following example: <ssl id="defaultSSLConfig" keyStoreRef="newKeyStore" />.

**Proposed solutions:**
1. Use JRE truststore
1. Add `<keyStore id="defaultKeyStore" password="CHANGEIT" />` to the default server template (would also require adding `ssl-1.0` to server template)
1. Add commented out config to the server template:
    ```xml
    <!-- if you enable ssl-1.0 (or anything that includes it, such as javaee-8.0 or webProfile-8.0) you must configure a <keystore> element -->
    <!-- <keyStore id="defaultKeyStore" password="CHANGEIT" /> -->
    ```
1. Have the `server create` command generate a random strong keystore password.  For example, have `<keyStore id="defaultKeyStore" password="${keystore.generated.password}" />` in the default config template and generate a random value for `keystore.generated.password` in the bootstrap.properties file.

NottyCode:
We don't use default passwords in Liberty, so that rules out doing option 2. It isn't secure.

aguibert:
[...]

PR #3558, opened 2018-05-14
Set default ID for keyStore element so configs merge properly

aguibert:
In the past, it has been possible to do the following to configure the default keystore:

<server>
  <keyStore password="whatever"/>
</server>

As a result of the following design issue: https://github.com/OpenLiberty/open-liberty/issues/1175 we made changes to auto-generate the default keystore out of the box.

In this case, both configurations fail because we are trying to create 2 keystores with different passwords at the same location:  one for the default instance with the generated password, and the other for the no-ID config element in server.xml.

To resolve this issue, we will default the ID to `defaultKeyStore` in metatype, so that no-ID configs will merge with the default instance.  This also has the side effect that multiple no-ID configs will now merge into one, but this sort of config is not usable currently, so it seems to be an acceptable tradeoff.


aguibert:
#build

aguibert:
#build

aguibert:
#build

aguibert:
#build

aguibert:
#build

PR #3584, opened 2018-05-16
JAX-RS performance improvements

dettmoney:
Fixed the Servlet3ContinuationProvider to eliminate loading a
class on every request. With EE8 we're always running servlet
4, so we don't need to attempt to load the 3.1 class. Similar
change in AbstractHTTPDestination as we know we will have
access to the method based on the servlet spec version supported.


dettmoney:
#build

dettmoney:
#build

PR #3616, opened 2018-05-17
Session cache changes

alexsm82:


alexsm82:
Both of these commits had been reviewed by @njr-11

Issue #3489, opened 2018-05-08
MP Rest Client does not use Liberty SSL config when making outbound requests


andymc12:
In Liberty 18.0.0.1, the `mpRestClient-1.0` feature was released.  It is possible to make requests to HTTPS sites - but by default they will only work if the SSL certificate used is a trusted CA certificate that is already in the JDK's cacert store.  

The JAX-RS Client in Liberty will use the `LibertyJaxRsClientSSLOutInterceptor` to utilize the Liberty SSL settings in the server.xml config - this allows users to define self-signed certificates and put them into server-specific trust stores.  Currently this does not work for MP Rest Client, but should.

PR #3618, opened 2018-05-17
Issue 3489: Allow outbound MP Rest Clients to use SSL configuration

andymc12:
This fix allows users to use custom SSL certificates by configuring
the key/trust store in server.xml with their MP Rest Client.  This
functionality has existed with JAX-RS 2.0 Clients for some time, but
did not work for MP Rest Clients.

This requires a new auto-feature that is enabled when mpRestClient-1.0
and ssl-1.0 are enabled.  This auto-feature will pull in the
jaxrsClient-2.0 (tolerating 2.1) feature and will register a listener
for new rest clients - that listener will then add the
LibertyJaxRsClientSSLOutInterceptor which utilizes the SSL config on
outbound calls.

This is tested by modifying the existing basic test case to use a 
non-default SSL configuration.

This PR resolves Issue #3489 which is a release bug.


andymc12:
#build

andymc12:
Rebase - new #build

andymc12:
#build

andymc12:
#build

Issue #3613, opened 2018-05-17
Display an error message when websocket feature is not enabled for spring boot application using websocket


anjumfatima90:
Look for the spring-boot-starter-websocket classes and if those are present, check if the websocket feature is enabled in server.xml. Throw an error when websocket feature is not enabled.

tjwatson:
This is fixed in master.

PR #3629, opened 2018-05-18
Issue #3613: Display an error message when websoket feature is not en…

anjumfatima90:
for issue #3613


sishida:
Messages look good to ID! I added the ID reviewed label. :)

Issue #3649, opened 2018-05-21
Filter out the default netty server when deployed on liberty for applications using spring-boot-starter-webflux 


anjumfatima90:
The spring boot applications using the spring-boot-starter-webflux, when deployed on liberty, starts up the netty server. We need to filter out the netty server when it is deployed on liberty just like we do for other embedded containers.

tjwatson:
This is fixed in master.

PR #3658, opened 2018-05-22
Issue#3649: Filter out the default netty server when deployed on libe…

anjumfatima90:
for issue #3649


anjumfatima90:
#build

tjwatson:
The changes only effect tests that enable springBoot features.  The other failures in the referenced build were in FATs that do not use springBoot features.

PR #3662, opened 2018-05-22
3042 report password expired user revoked exceptions

jarcher-IBM:
#build

### DIFF
--- a/dev/com.ibm.ws.webcontainer.security/src/com/ibm/websphere/security/web/PasswordExpiredException.java
+++ b/dev/com.ibm.ws.webcontainer.security/src/com/ibm/websphere/security/web/PasswordExpiredException.java
@@ -11,22 +11,24 @@
 package com.ibm.websphere.security.web;
 
 import javax.servlet.ServletException;
+
 /**
- *
+ * This exception is thrown when a programmatic login is performed against a user registry which is configured
+ * to report when a user's password is expired.
  */
 public class PasswordExpiredException extends ServletException{
 
-    private static final long serialVersionUID = 1L;
-
-    public PasswordExpiredException(String message) {
-        super(message);
-    }
-
     /**
-     * @param message
-     * @param e
+     * Comment for <code>serialVersionUID</code>
      */
-    public PasswordExpiredException(String message, Exception e) {
+    private static final long serialVersionUID = 12345622414392883L;
+    
+    /**
+     * Creates a new PasswordExpiredException with a message
+     *
+     * @param message Exception message about what caused the error
+     */
+    public PasswordExpiredException(String message) {
         super(message);
     }
 }
--- a/dev/com.ibm.ws.webcontainer.security/src/com/ibm/websphere/security/web/PasswordExpiredException.java
+++ b/dev/com.ibm.ws.webcontainer.security/src/com/ibm/websphere/security/web/PasswordExpiredException.java
@@ -1,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2018 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.websphere.security.web;
+
+import javax.servlet.ServletException;
+/**
+ *
+ */
+public class PasswordExpiredException extends ServletException{
+
+    private static final long serialVersionUID = 1L;
+
+    public PasswordExpiredException(String message) {
+        super(message);
+    }
+
+    /**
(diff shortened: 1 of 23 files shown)


dev/com.ibm.ws.webcontainer.security/src/com/ibm/websphere/security/web/PasswordExpiredException.java
```diff
@@ -0,0 +1,32 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.websphere.security.web;
+
+import javax.servlet.ServletException;
+/**
```
kaczyns: Since this is going to be API, customers will be able to see the code (or at least, javadoc will get generated from this).  We need to say something about what this exception is used for.  We could say something like "This exception is thrown when a programmatic login is performed against a user registry which is configured to report when a user's password is expired."

We should say something similar for the user revoked exception (the one that extends ServletException).

dev/com.ibm.ws.webcontainer.security/src/com/ibm/websphere/security/web/PasswordExpiredException.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+public class PasswordExpiredException extends ServletException{
+
+    private static final long serialVersionUID = 1L;
```
kaczyns: Since this is going to be API, we should probably generate a unique UID (or at least pick a number at random other than 1).  Same for the other exception.

dev/com.ibm.ws.webcontainer.security/src/com/ibm/websphere/security/web/PasswordExpiredException.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public PasswordExpiredException(String message) {
```
kaczyns: We need to be captain obvious here and say something in the javadoc like "Create a new PasswordExpiredException with a message".  same for the other constructors, and for the user revoked exception too.

dev/com.ibm.ws.webcontainer.security/test/com/ibm/ws/webcontainer/security/internal/AuthenticateApiTest.java
```diff
@@ -414,6 +425,112 @@ public void testLoginSuccess() throws Exception {
...
+                allowing(mockServletContext).getAttribute(with(any(String.class)));
+                will(returnValue(null));
+            }
+        });
+        
+        AuthenticateApi authApi = new AuthenticateApi(ssoCookieHelper, securityServiceRef, cu, null, new ConcurrentServiceReferenceMap<String, UnprotectedResourceService>("unprotectedResourceService"));
+        BasicAuthAuthenticator basicAuthAuthenticator = new BasicAuthAuthenticator(authnService, userRegistry, ssoCookieHelper, config);
+        
+        try {
+            authApi.login(req, resp, user, password, config, basicAuthAuthenticator);
```
kaczyns: It seems like you need a 'fail' in here somewhere, in case the login method doesn't throw an exception.

dev/com.ibm.ws.webcontainer.security/test/com/ibm/ws/webcontainer/security/internal/AuthenticateApiTest.java
```diff
@@ -414,6 +425,112 @@ public void testLoginSuccess() throws Exception {
...
+                allowing(mockServletContext).getAttribute(with(any(String.class)));
+                will(returnValue(null));  
+            }
+        });
+        
+        AuthenticateApi authApi = new AuthenticateApi(ssoCookieHelper, securityServiceRef, cu, null, new ConcurrentServiceReferenceMap<String, UnprotectedResourceService>("unprotectedResourceService"));
+        BasicAuthAuthenticator basicAuthAuthenticator = new BasicAuthAuthenticator(authnService, userRegistry, ssoCookieHelper, config);
+        
+        try {
+            authApi.login(req, resp, user, password, config, basicAuthAuthenticator);
```
kaczyns: Same comment here about failing if we don't throw an exception.  I think there is a junit `fail(message)` method that should work.

kaczyns:
Hi Joe - please check the build results.  There are failures in the feature checker because we added two classes to that API jar.  We need to increment the version on the API jar.  There are instructions (or should be instructions) in the test failure describing what to do.

```ERROR: com.ibm.websphere.appserver.webAppSecurity-1.0
[PACKAGE_COMPARE_MINOR com.ibm.websphere.security.web@null]

The class UserRevokedException has been added to the package.
For More Information, visit http://was.pok.ibm.com/xwiki/bin/view/Build/FeatureChecker```

jarcher-IBM:
#build

jarcher-IBM:
#build

PR #3537, opened 2018-05-11
add metatype attrib with one second granularity

brutif:
adds a new metatype attribute to the jwt builder, expiresInSeconds, so expiration can be set with one second granularity instead of the current one hour granularity.

PR #3674, opened 2018-05-22
convert default exp time to seconds

brutif:
#3537 changed the jwt default expiration time from hours to seconds, this change updates the jwtsso builder to also use seconds.


brutif:
#build

brutif:
looks like the pb has stalled, but I'll need to resolve the conflict anyway, so I'll proceed to do that and then kick off another pb

brutif:
#build

brutif:
pb is good

PR #3681, opened 2018-05-23
Remove method that converts connection properties

alexsm82:
Remove method that converts connection properties .... (test issue)


alexsm82:
#run-libby

PR #3701, opened 2018-05-24
JWT SSO FAT: Cookie manipulation and manually building JWT

ayoho:
Delivers the following tests:

- Log in with the JWT builder expiry set to a very short time, re-invoke protected resource with the JWT cookie within the allowed clock skew
- Log in with the JWT builder expiry set to a very short time, consumer clock skew set to an extremely short time, re-invoke protected resource with the JWT cookie outside of the allowed clock skew
- Log in, remove the entire signature segment from the JWT cookie, re-invoke protected resource with the signature-less cookie
- Log in, change the JWT cookie signature to be empty, re-invoke protected resource with the modified cookie
- Log in, access a different protected resource without the JWT cookie
- Log in, access a different protected resource with the JWT cookie
- Build a JWT that's missing required attributes (`upn` and `sub`) and invoke a protected resource with that JWT included as a cookie
- Build a JWT using a `jwtBuilder` with all default configuration values and invoke a protected resource with that JWT (all issuers _not_ trusted)
- Build a JWT using a `jwtBuilder` with all default configuration values and invoke a protected resource with that JWT (all issuers trusted)
- Build a JWT using a `jwtBuilder` that uses a unique signing key and invoke a protected resource with that JWT


ayoho:
Personal build https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_7zW8MF7DEeiK15wbvZUEzg completed 100% green.

PR #3719, opened 2018-05-24
adding support for resolver jar override

### LINKED ISSUE #3692: Override repository resolver jar in InstallMap
support a parameter to override the repository resolver jar file location, instead of using the one defined in the manifest.

hermankailey:
fixes #3692

### DIFF
--- a/dev/com.ibm.ws.install.map/src/com/ibm/ws/install/map/InstallMap.java
+++ b/dev/com.ibm.ws.install.map/src/com/ibm/ws/install/map/InstallMap.java
@@ -130,6 +130,7 @@ public class InstallMap implements Map {
     private static final String RUNTIME_INSTALL_DIR = "runtime.install.dir";
     private static final String INSTALL_KERNEL_INIT_CODE = "install.kernel.init.code";
     private static final String INSTALL_KERNEL_JAR = "install.map.jar";
+    private static final String OVERRIDE_JAR_BUNDLES = "override.jar.bundles";
     private static final String INSTALL_KERNEL_INIT_ERROR_MESSAGE = "install.kernel.init.error.message";
     private static final String MESSAGE_LOCALE = "message.locale";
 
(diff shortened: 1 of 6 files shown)


dev/com.ibm.ws.install.map/src/com/ibm/ws/install/map/InstallMap.java
```diff
@@ -406,17 +413,49 @@ else if (jars.length == 1) {
...
             Attributes attributes = manifest.getMainAttributes();
             String[] requireBundles = attributes.getValue("Require-Bundle").split(",");
+            String[] installMapVersionSplit = attributes.getValue("Bundle-Version").split("\\.");
+            String installMapVersion = "";
+            for (String mapVersionSplit : installMapVersionSplit) {
+                if (!(mapVersionSplit.length() > 3)) {
+                    installMapVersion = installMapVersion + mapVersionSplit;
+                }
+            }
+            if (resolverVersion.compareTo(installMapVersion) >= 0) {
```
alvinso: Currently we want to match the exact version i.e. Should not replace the jar even the version of the jar from Maven Central is higher than than the one in runtime.

dev/com.ibm.ws.install.map/src/com/ibm/ws/install/map/InstallMap.java
```diff
@@ -130,6 +130,8 @@
     private static final String RUNTIME_INSTALL_DIR = "runtime.install.dir";
     private static final String INSTALL_KERNEL_INIT_CODE = "install.kernel.init.code";
     private static final String INSTALL_KERNEL_JAR = "install.map.jar";
+    private static final String OVERRIDE_JAR_BUNDLE_NAME = "override.jar.bundle.name";
+    private static final String OVERRIDE_JAR_BUNDLE_FILE = "override.jar.bundle.File";
```
alvinso: The key should be all lower case

Issue #3706, opened 2018-05-24
Issue a warning if compression is configured for Spring Boot applications


tjwatson:
Spring boot has the following settings which are related to compression support:

    server.compression.enabled
    server.compression.excluded-user-agents
    server.compression.mime-types
    server.compression.min-response-size

For now we will issue a warning if these settings are used.  A separate issue #3705 is open to investigate how to enable compression.

sebratton:
As described above there are many different configuration settings.  A warning message will only be issued if compression is actually enabled

tjwatson:
Fixed in master

Issue #2487, opened 2018-03-02
[spring boot feature] session configuration 


tjwatson:
Spring boot many settings for controlling session behavior:

server.session.cookie.comment
server.session.cookie.domain
server.session.cookie.http-only
server.session.cookie.max-age
server.session.cookie.name
server.session.cookie.path
server.session.cookie.secure
server.session.persistent
server.session.store-dir
server.session.timeout
server.session.tracking-modes

Need to investigate what these settings do and if they can be configured on liberty.

tjwatson:
For session support we will be only supporting the settings in server.xml for things like:

server.session.persistent
server.session.store-dir
server.session.timeout

This task should be used to issue a warning if these options are used by a spring boot application.

sebratton:
There are many session setting. For purposes or generating a warning, there will only be one issue if one or more of the following conditions are true:
-- isPersistentSession == true
-- getSessionTimeout != 30*60 (30 mins in seconds)

tjwatson:
This is fixed in master.

PR #3722, opened 2018-05-25
Warn on not supported config

sebratton:


sebratton:
#build

tjwatson:
Fixes #3706 and #2487

PR #3731, opened 2018-05-25
Upgrade to Hibernate Validator 6.0.10

aguibert:


aguibert:
#build

aguibert:
Currently the beanValidation-2.0 feature uses Hibernate Validator 6.0.9.  While we were upgrading to the latest version (6.0.10) we noticed the following change that we are concerned will really negatively impact startup performance:
https://github.com/hibernate/hibernate-validator/pull/953/files#diff-d9d8b317ed14e228415c78d1b4785a75L222
 
Before, in the CDI extension the processAnnotatedType() used to filter on just BeanVal annotations:
`public <T> void processAnnotatedType(@Observes @WithAnnotations({ Constraint.class, Valid.class, ValidateOnExecution.class } ....)`
 
Now, they removed @WithAnnotations completely so they will observe ALL annotated types, and when they find one they do a very expensive looking reflective walk up the class hierarchy:
https://github.com/hibernate/hibernate-validator/pull/953/files#diff-66e9ba3cc4d09f83eb2ae799daeacab5R54

aguibert:
Raised a concern on the Hibernate Validator PR that introduced this performance regression:
https://github.com/hibernate/hibernate-validator/pull/953

aguibert:
The Hibernate Validator team was receptive of the performance concerns I raised and are reverting hibernate/hibernate-validator#953 in version 6.0.12.Final.  Cancelling this PR since we will skip past the 6.0.10.Final version.

PR #3736, opened 2018-05-25
Issue 1425: Suppress CWWKW1001W and CWWKW1002W msgs in JAX-RS 2.1

andymc12:
These messages indicate a mismatch between the lifecycle scopes of
JAX-RS and CDI for resources and providers, respectively.  In most cases
this mismatch is not worth logging since a user is indicating their
preference for the particular scope by adding a CDI or EJB annotation
to their resource/provider class.

The message will still be output, but only when tracing is enabled.

The CDITest verifies that this message is suppressed by no longer
allowing that message to be ignored - if the message appears during that
test, then it will fail.

This is _not_ a release bug because the JAX-RS 2.1 feature has not yet been released.


andymc12:
#build

WhiteCat22:
Looks good

andymc12:
#build

andymc12:
#build

andymc12:
#build

andymc12:
new tweak - #build

andymc12:
Failing tests are unrelated to this change.

PR #3753, opened 2018-05-29
connection leak on managed connection creation failure

### LINKED ISSUE #3752: Connection leak if failure occurs while managed connection is being constructed
A connection is obtained via the JDBC driver just before constructing a ManagedConnection instance for it.  If the code that constructs the ManagedConnection instance fails, the reference to the connection is lost and it is never closed, resulting in a connection leak.

njr-11:
pull fixes #3752


njr-11:
#build

njr-11:
#build

njr-11:
build has 1 test failure, which is a duplicate of a pre-existing issue in unrelated componentry

Issue #3646, opened 2018-05-21
Add AroundTimeout tests for EE8


brideck:
Java EE 7 level of the Interceptors specification required the @aroundtimeout method signature to throw Exception. All other callback methods do not require any exception be on the throws clause, so the Java EE 8 version of the specification has been updated to remove this requirement.

At a minimum, there needs to be at least one test that does not have the 'throws exception', and there should be one test that throws an 'application' exception rather than Exception.

PR #3790, opened 2018-05-30
3646 add around timeout ee8 tests

brideck:
Tested AroundTimeout methods without a throws clause
Tested AroundTimeout methods that throw an application exception


brideck:
fixes #3646

brideck:
#build

PR #3745, opened 2018-05-28
JSR375: use jaspicSession cookie instead of LTPAToken cookie

toshiyamamoto:
This PR includes:
Modify LoginToContinueInterceptor form conforming the JSR375 specification.
Use jaspicSession cookie for Form and Custom Form login.
Remove jaspicSession cookie upon logout.

PR #3798, opened 2018-05-31
JSR375: NPE when contextRootForFormAuthenticationMechanism value is not valid.

toshiyamamoto:


toshiyamamoto:
#build

toshiyamamoto:
fixes #3745

PR #3839, opened 2018-06-04
Update CXF binaries to improve behavior when running JDKs past JDK8

neuwerk:
Update the CXF binaries that the jaxws-2.2 feature uses in order to fix issues when running JDK levels higher than JDK8


neuwerk:
#build

neuwerk:
#build

PR #3850, opened 2018-06-04
JWT SSO FAT: Update JWK test to use different keystore

ayoho:
Ensure that the JWK test for the `jwtSso-1.0` feature sets the `mpJwt` configuration to use a keystore with a certificate that can't be used to validate the JWT cookie, but also configures the `jwksUri` attribute. This should make sure we're testing that if the `jwksUri` attribute is configured, we should use the JWK endpoint to get the right key to verify the JWT cookie instead of looking in the configured keystore.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.jwtsso_fat

Issue #3851, opened 2018-06-04
JAX-RS Client APIs fail when attempting PATCH method over HTTPS on IBM JDK


andymc12:
By default the JAX-RS 2.0 Client APIs will not allow users to invoke HTTP PATCH methods.  It is possible to make this work by specifying the system property `-Duse.httpurlconnection.method.reflection=true`.  This property is enabled by default in Liberty's JAX-RS 2.1 implementation.

However, if a user is using this property when making an HTTPS call when using the IBM JDK, the request may fail, and the user may see a NullPointerException like this:
```
0000004a id=         .ibm.ws.jaxrs.2.0.common:1.0.20.cl180120180309-2209(id=180)] W java.lang.NullPointerException
    at org.apache.cxf.common.util.ReflectionUtil$11.run(ReflectionUtil.java:196)
    at org.apache.cxf.common.util.ReflectionUtil$11.run(ReflectionUtil.java:194)
    at java.security.AccessController.doPrivileged(AccessController.java:594)
    at org.apache.cxf.common.util.ReflectionUtil.setAccessible(ReflectionUtil.java:194)
    at org.apache.cxf.transport.http.URLConnectionHTTPConduit.setupConnection(URLConnectionHTTPConduit.java:163)
    at org.apache.cxf.transport.http.HTTPConduit.prepare(HTTPConduit.java:504)
    at org.apache.cxf.interceptor.MessageSenderInterceptor.handleMessage(MessageSenderInterceptor.java:46)
    at org.apache.cxf.phase.PhaseInterceptorChain.doIntercept(PhaseInterceptorChain.java:308)
    at org.apache.cxf.jaxrs.client.AbstractClient.doRunInterceptorChain(AbstractClient.java:708)
    at org.apache.cxf.jaxrs.client.WebClient.doInvokeAsync(WebClient.java:951)
    at org.apache.cxf.jaxrs.client.WebClient.doInvokeAsyncCallback(WebClient.java:914)
    at org.apache.cxf.jaxrs.client.WebClient$AsyncInvokerImpl.method(WebClient.java:1437)
    at com.ibm.ws.test.MyTestClient.sendPatchRequest(MyTestClient.java:121)
```

This problem exists in the `jaxrs-2.1` feature (which is not yet released) and the `jaxrs-2.0` feature making this a release bug.

PR #3852, opened 2018-06-04
Issue 3851: Null check on reflective use of HTTP methods

andymc12:
This is required to support non-Oracle JDKs (specifically IBM JDK) when
using non-standard HTTP methods (like PATCH, etc) over HTTPS.  The
runtime changes are in URLConnectionHTTPConduit.java.

This is tested by modifying the existing PATCH test case to use HTTPS.

Resolves issue #3851


andymc12:
#build

andymc12:
#build - updated after testing locally on IBM JDK on Linux

andymc12:
including fix for jaxrs-2.0.  #build

andymc12:
a few more FFDCIgnores - #build

andymc12:
Last build showed one (unrelated) failure - and that test failure was not present in the second-to-last build (which only lacked a couple `@FFDCIgnore` annotations), so these changes should be safe to merge.

PR #3874, opened 2018-06-06
JPA 2.2 Container Feature needs JNDI feature dependency

jgrassel:

When testing the JPA 2.2 Container feature against HIbernate, I noticed that datasource lookup was failing due to in-availability of JNDI.  The JPA 2.2 feature declares a dependency on the JNDI feature, but the JPA 2.2 Container (which JPA 2.2 extends) did not.  This corrects that oversight.


jgrassel:
#build

jgrassel:
#build

PR #3887, opened 2018-06-06
Remove bad time restriction from test

### LINKED ISSUE #3886: Remove more resctrictions on sse delay test that are causing intermittent test failures
SSe delay testcase is intermittently failing due to some additional time restrictions that are not needed but will cause random failures when test machines are slow.

jim-krueger:
Fixes #3886


jim-krueger:
#build

PR #3891, opened 2018-06-06
Skip the JAXB test when Java 2 security is enabled.

andymc12:
This test fails when Java 2 security is enabled because the JDK does
not have doPriv blocks around things like checking for JAXB- or XML-
specific system properties, etc.

Since we cannot control the JDK, we will continue to run these tests,
but only when Java 2 security is not enabled.

This is a test case issue, so it is not a release bug.


andymc12:
#build

PR #3899, opened 2018-06-07
3017 MicroProfile FaultTolerance getFailure() and 1.1 TCK

hutchig:
Implements getFailure and adds the TCK as a FAT project.


hutchig:
#build

hutchig:
Note - the #build's TCK will fail in the Disable* packages tests as there is a further feature to get these working coming from Ben but I have verified all the rest of the TCK passes. 
Do not merge into integration/master without Ben's feature.

PR #3924, opened 2018-06-08
Add recursion counters to avoid infinite loop in trace for hpelbts

### LINKED ISSUE #3923: Need to add recursion counter into HpelBaseTraceService
There is no recursion counter currently for the `publishLogRecord` and the `publishTraceLogRecord`
which will cause an infinite loop if traces is enabled and bufferManagerImpl as part of the traces.

This scenario only happens if "\*" is used and/or "x.com.*" is used.
("x.com" because of BufferManagerImpl having the 'x' identifier")

Channyboy:
#build 
fix #3923 
The change is to add a recursion counter to the `publishTraceLogrecord()` in `HpelBaseTraceService` similar to `BaseTraceService`. The counter is lacking in `HpelBaseTraceService` due to the method having been overrode. 

The counter to is address any infinite traces that can occur if the `BufferManagerImpl` is traced.


Channyboy:
The fat test that are failing do not relate to the component being changed in this PR.
The changes made here are for the HPEL logging which can only be enabled when certain property is set into the boostrap.properties (e.g : `websphere.log.provider=binaryLogging-1.0`)

PR #3928, opened 2018-06-11
ensure warning is printed

andymc12:
Not a release bug because the change that introduced this problem (not logging unless trace is enabled) has not released yet.


andymc12:
#build

PR #3935, opened 2018-06-11
keep more careful track of FFDCs encountered in FATs

wtlucy:


wtlucy:
#build

wtlucy:
fixes #3943

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -478,12 +528,14 @@ private String getFFDCHeader(RemoteFile ffdcLogFile) throws Exception {
...
-                                        oldInfo.count += entry.getValue().count;
+                                    // check to see if a set of FFDCInfo already existed
+                                    Set<FFDCInfo> oldInfo = ffdcPrimaryInfo.get(entry.getKey());
+                                    if (oldInfo == null) {
+                                        Set<FFDCInfo> set = new HashSet<FFDCInfo>();
+                                        set.add(entry.getValue());
+                                        ffdcPrimaryInfo.put(entry.getKey(), set);
                                     } else {
-                                        ffdcPrimaryInfo.put(entry.getKey(), entry.getValue());
+                                        ffdcPrimaryInfo.get(entry.getKey()).add(entry.getValue());
```
aguibert: If there is already a `Set<FFDCInfo>` for this key, we should check if there is a matching `FFDCInfo` in the set, and if so, increment the count, and if not do the current line 538 code

wtlucy:
#build

PR #3951, opened 2018-06-12
Copy License and Notices to package staging dir

cbridgha:
Although License was previously added to the full build.image zip, They were not included in the packages, because that task uses a separate staging directory, with the feature Minify logic.
We now copy in these files before zipping contents.


cbridgha:
#build

PR #3959, opened 2018-06-13
3042 report password expired

### LINKED ISSUE #3042: Allow user registry to report password expired or user revoked to the web container
In some situations it is useful for the user registry to report that a user's password has expired or that the user is revoked.  Some user registries only give a user a certain number of attempts to log in before their user is revoked, and if their password is expired, no number of attempts will allow them to log in, leading to their ID being revoked.  This notification should be optional.  The user registry has no way to propagate this information today.

The `com.ibm.ws.security.registry.RegistryException` that is thrown from the user registry to the security code should be subclassed to a `PasswordExpiredException` and `UserRevokedException`.  The javadoc for these exceptions should make it clear that they should only be thrown if the user registry reporting the error has been explicitly configured to propagate this information.  

The `com.ibm.ws.security.authentication.jaas.modules.UsernameAndPasswordLoginModule` class `login()` method has a generic `catch (Exception e)` block.  We need to look specifically for subclasses of RegistryException now.  If we catch one of the two new subclasses, we should propagate this to the next 'layer' by subclassing `com.ibm.ws.security.authentication.AuthenticationException` in the same way we subclassed RegistryException.  Again, the javadoc should be specific about only throwing this if the user registry reporting the error has been explicitly configured to propagate this information.  Here is an example of how the new catch block in `login()` may look:
```Java
catch (Exception e) {
    if (e instanceof com.ibm.ws.security.registry.RegistryException) {
        if (e instanceof com.ibm.ws.security.registry.PasswordExpiredException) {
            throw new PasswordExpiredException(e.getLocalizedMessage(), e); /* where this PasswordExpiredException is a subclass of AuthenticationException */
        } else if (e instanceof com.ibm.ws.security.registry.UserRevokedException) {
            throw new UserRevokedException(e.getLocalizedMessage(), e); /* where this UserRevokedException is a subclass of authenticationException */
        } else {
            throw new AuthenticationException(e.getLocalizedMessage(), e);
        }
    }
}
```

We would like to report this information on a servlet programmatic login.  This uses the basic auth authenticator, `com.ibm.ws.webcontainer.security.internal.BasicAuthAuthenticator`.  The `basicAuthenticate()` method catches AuthenticationException, and converts it to an `AuthenticationResult`.  If it catches one of the two new subclasses, it should add this information to the AuthenticationResult.  Add two new booleans to the authentication result to contain this information (one for password expired and one for user revoked).

The AuthenticationResult is returned to class `com.ibm.ws.webcontainer.security.AuthenticateApi`  method `login()`.  If the password expired or user revoked booleans in the authentication result are set, we should include text in the resulting `ServletException` that says why.  The text can simply be [PWEXPIRED] or [IDREVOKED].

kaczyns:
This is a new PR replacing #3662 to resolve a merge conflict.  Fixes #3042.
#build


kaczyns:
@jarcher-IBM is going to approve the review record for this.  This is all his code, and I already reviewed #3662, but I had to open this PR to resolve a merge conflict, and I am not allowed to review my own PR.

Issue #3098, opened 2018-04-13
WELD-000416 messages on startup with FT and CDI 2.0


Azquelt:
```
[4/12/18 21:25:51:977 CDT] 0000088f org.jboss.weld.Bootstrap                                     W WELD-000146: BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) used for class com.ibm.ws.microprofile.faulttolerance.cdi.FaultToleranceInterceptor is deprecated from CDI 1.1!
[4/12/18 21:25:51:977 CDT] 0000088f org.jboss.weld.Bootstrap                                     W WELD-000146: BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) used for class com.ibm.ws.microprofile.faulttolerance.cdi.FaultToleranceInterceptor$ExecutorCleanup is deprecated from CDI 1.1!
```

tevans78:
```[4/12/18 21:25:51:977 CDT] 0000088f org.jboss.weld.Bootstrap                                     W WELD-000146: BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) used for class com.ibm.ws.microprofile.faulttolerance.cdi.FaultToleranceInterceptor is deprecated from CDI 1.1!
[4/12/18 21:25:51:977 CDT] 0000088f org.jboss.weld.Bootstrap                                     W WELD-000146: BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) used for class com.ibm.ws.microprofile.faulttolerance.cdi.FaultToleranceInterceptor$ExecutorCleanup is deprecated from CDI 1.1!
[4/12/18 21:25:51:977 CDT] 0000088f org.jboss.weld.Bootstrap                                     W WELD-000146: BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) used for class com.ibm.ws.security.mp.jwt.cdi.ClaimProducer is deprecated from CDI 1.1!```

Looks like an easy fix, just switch from using `addAnnotatedType(AnnotatedType<?> type)` to `addAnnotatedType(AnnotatedType<?> type, String id)`

Need to make sure that the id has the right level of uniqueness

Also looks like app security already fixed this in their interceptor: https://github.com/OpenLiberty/open-liberty/pull/1906

tevans78:
i just found the one other place we already create this id and we used the class’ own hashcode…
```clazz.getName() + ": " + clazz.hashCode()```
[...]

PR #4031, opened 2018-06-20
Use addAnnotatedType non-deprecated signature

hutchig:

Very minor bug, see #3098


hutchig:
#build

hutchig:
fixes #3098

dev/com.ibm.ws.microprofile.faulttolerance.cdi/src/com/ibm/ws/microprofile/faulttolerance/cdi/FaultToleranceCDIExtension.java
```diff
@@ -59,9 +59,9 @@ public void beforeBeanDiscovery(@Observes BeforeBeanDiscovery beforeBeanDiscover
         AnnotatedType<FaultTolerance> bindingType = beanManager.createAnnotatedType(FaultTolerance.class);
         beforeBeanDiscovery.addInterceptorBinding(bindingType);
         AnnotatedType<FaultToleranceInterceptor> interceptorType = beanManager.createAnnotatedType(FaultToleranceInterceptor.class);
-        beforeBeanDiscovery.addAnnotatedType(interceptorType);
+        beforeBeanDiscovery.addAnnotatedType(interceptorType, (interceptorType != null) ? interceptorType.getClass().getName() + ": " + interceptorType.getClass().hashCode() : "");
```
Azquelt: Is the null check actually needed here?

hutchig: I discussed this with Tom.  We reckon, that with the current system it probably is not - but we could not state that it definately was not or that this would not change. So I decided to be safe - it is one machine instruction. 
The other 'odd' thing you might be thinking is 'why call the weld method if the value IS null' and the reason is that I want weld to react and get a weld error (etc.)  rather than a NPE in a getClass().

PR #4142, opened 2018-06-28
Minor doc improvements to JMS & server cmd

NottyCode:
The online help for the server package tool doesn't list wlp as an option for
--include even though it is documented in the command description.

The WAS JMS configuration references 'full profile server' which is now referred
to as "WebSphere Application Server traditional"


NottyCode:
@cbridgha I haven't flagged this as a release bug even though I'm improving docs that has been released. I kind of think release bug is overkill for this, but I'd like your thoughts.

@LibbyBot please #build this.

Issue #4112, opened 2018-06-25
Need to propagate authorization header using MP Rest Client 


justineechen:
While attempting to update the rest client for open liberty's JWT guide, I ran into an issue propagating the authorization header. I tried using the fix found in this issue:
https://github.com/eclipse/microprofile-rest-client/issues/73

But received the following error:
```
javax.ws.rs.ProcessingException: javax.net.ssl.SSLException: SSLException invoking https://localhost:5050/system/properties: Unrecognized SSL message, plaintext connection?
        at it.io.openliberty.guides.jwt.SystemEndpointTest.testGetPropertiesWithJwt(SystemEndpointTest.java:49)
        at it.io.openliberty.guides.jwt.SystemEndpointTest.testSuite(SystemEndpointTest.java:43)
Caused by: javax.net.ssl.SSLException: SSLException invoking https://localhost:5050/system/properties: Unrecognized SSL message, plaintext connection?
        at it.io.openliberty.guides.jwt.SystemEndpointTest.testGetPropertiesWithJwt(SystemEndpointTest.java:49)
        at it.io.openliberty.guides.jwt.SystemEndpointTest.testSuite(SystemEndpointTest.java:43)
Caused by: javax.net.ssl.SSLException: Unrecognized SSL message, plaintext connection?
        at it.io.openliberty.guides.jwt.SystemEndpointTest.testGetPropertiesWithJwt(SystemEndpointTest.java:49)
        at it.io.openliberty.guides.jwt.SystemEndpointTest.testSuite(SystemEndpointTest.java:43)

```

Open Liberty 18.0.0.1/wlp-1.0.20.cl180120180309-2209 on IBM J9 VM, version 8.0.5.11

evelinec:
@andymc12 Tagging Andy FYI.

andymc12:
Hi @justineechen - I think the problem listed is that at least one of the requests is sending unencrypted data to a port that expects encrypted data.  Possibly the URL contains "http" instead of "https"?

That said, I think you will still be broken when using SSL until the change for PR #3618 is integrated.  This enabled MP Rest Clients to use custom SSL configurations.  I am also attempting to integrate a test case that shows HTTP header propagation (over HTTP or HTTPS - including the `Authorization` header) in PR #4153.  If you look at that code and test server config, it may help with the issue you are facing.

gkwan-ibm:
1. have to make the keystore files same in both backendServer and UIServer
1. For backendServer,
[...]

PR #4153, opened 2018-06-28
Test propagation of HTTP Headers from JAX-RS to MP Rest Client

andymc12:
This PR builds on PR #3618 (which provides SSL config options for MP Rest Client) and tests that HTTP headers can be propagated from an inbound JAX-RS request to a MP Rest Client request.  The test has a JAX-RS resource that uses MP Rest Client to invoke another RESTful endpoint - it passes if the specified headers from the original request are sent along to the ultimate endpoint. 

The magic is in HeaderPropagationFilter - if the same instance of this class is registered with both the JAX-RS application and the MP Rest Client, then it will propagate any header specified in the system property from the inbound JAX-RS request to the outbound MP Rest Client request.


andymc12:
#build

andymc12:
Addresses issue #4112

andymc12:
Failures are unrelated - and are already accounted for by existing defects.

Issue #4034, opened 2018-06-20
MP Fault Tolerance 1.1: Metrics are not removed when an application closes


Azquelt:
In general, we need better control over how metrics are managed when applications can start and stop.

This is additionally complicated by the possibility of applications having overlapping metrics.

Azquelt:
Turns out this wasn't accurate, the metrics feature will automatically remove metrics when the app is removed.

This does give us slightly awkward behaviour when create overlapping metrics, when the first app to register the metric is removed, the metric is removed along with it, even if the second app is still running, but this is the same behaviour that a user would see if they manually registered metrics with the same name from different apps.

PR #4156, opened 2018-06-29
FAT test for FT metrics being removed

Azquelt:
Ensure that FT metrics are removed when the app is undeployed

Related to #4034 - it turns out that liberty does behave as we want, this PR just adds a test to assert that behaviour and ensure we don't regress in the future.


Azquelt:
#build

PR #4159, opened 2018-06-29
Rename fields to make it clear which Mongo API is accessed reflectively

njr-11:
As a first step toward being able to update the Mongo integration code, rename the variables used to reflectively access the Mongo Java driver API so that it is clear exactly what the code is currently doing/using.


njr-11:
#build

njr-11:
The build completed with 2 test failures, both in unrelated componentry.

PR #4163, opened 2018-06-29
identify driver and version from presence of class

njr-11:
Identify the Mongo driver type and version based on the presence or absence of MongoClients classes, defaulting to the existing logic for pre v3 driver level


dev/com.ibm.ws.mongo/src/com/ibm/ws/mongo/internal/MongoService.java
```diff
@@ -492,6 +516,19 @@ private void init() throws Exception {
         DB_isAuthenticated = DB.getMethod("isAuthenticated");
     }
 
+    /**
+     * Lazy initialization for version 2.x of the Mongo Java driver.
```
alexsm82: I presume this should be version "3.x of the Mongo.."

njr-11:
#build

PR #4167, opened 2018-07-02
When FFDC checking is disabled for one server, continue checking other servers

scottkurz:
When `componenttest.topology.impl.LibertyServer.setFFDCChecking(false)` is called the FAT framework will avoid checking this server's FFDCs against the expected/allowed set.

However the logic in `componenttest.custom.junit.runner.FATRunner` is written to immediately abandon checking when any single server has FFDC checking disabled.   

This can prevent unexpected FFDCs from being detected and also prevent expected FFDCs from being matched.

The fix is to simply continue checking FFDCs with the next server, after seeing a server with checking disabled.


scottkurz:
#build

scottkurz:
The last build **scottkurz-4167-20180702-1526** looks OK.  Of the three FAT failures, only one is an FFDC..and it's not in an area where FFDC checking was disabled.   The javacore from the one unit test is suprising.  I'll do a rebuild to get further confidence here.

scottkurz:
The last build, *scottkurz-4167-20180703-1427*, looks good.  None of the three failures involve FFDC checking.   The previous javacore didn't reoccur.     A code search shows only a handful of `setFFDCChecking(false)` usages anyway.

Issue #140, opened 2017-09-21
Update the WIM entering / exiting bean trace output


jvanhill:
Currently, the WIM bean trace output for the data objects is very sketchy. It chops off parts of words and the multi-line formatting indentation is nearly unreadable.

Update WIMTraceHelper to use Apache Commons' ReflectionToStringBuilder.

It is imperative to remember that the trace needs to encode passwords.

PR #4189, opened 2018-07-03
Issue #140: Prototype for using JAXB for WIM bean trace. Ultimately t…

jvanhill:
…his failed since Java 2 security would prevent JAXB from making reflection calls during Marshalling. The only work-around was to add the permissions to the java.security file in the JRE, which I didn't see as feesible.


jvanhill:
#build

jvanhill:
#build

PR #4195, opened 2018-07-05
Exclude FT 1.1 timing sensitive tests

Azquelt:
Exclude a few of the tests from the FT 1.1 TCK which are sensitive to
timing since they cause occasional intermittent failures when the
test system is under load.

Notes:
The first two tests have been reported broken in the 1.1 TCK and will be fixed in 1.1.1
The `testBulkhead*` tests were already excluded in the 1.0
The last three tests are around timing metrics, so there's no real way to avoid them being timing sensitive. They're new for 1.1.


Azquelt:
#build

Azquelt:
Several errors occurred but all are due to a different deployment error which I hope to fix with OpenLiberty/liberty-arquillian#19

PR #3703, opened 2018-05-24
Tolerate JDBC 4.2 drivers not supporting Statement.getLargeUpdateCount() on cleanup code path

aguibert:
Fixes #3700

PR #4198, opened 2018-07-05
Also handle UnsupportedOperationException in getCompatibleUpdateCount

aguibert:
Fixes #3703 

Need to also handle the JDK default implementation of `java.sql.Statement.getLargeUpdateCount()` which throws an `UnsupportedOperationException` instead of a `SQLFeatureNotSupportedException`

@eweitlaner FYI


aguibert:
#build

Issue #2659, opened 2018-03-13
Capture security context from Java Batch thread when syncToOSThread is enabled


dmbelina:
The Java Batch container is not properly saving the security context from the running work thread when syncToOSThread is enabled.

The Java Batch context thread pool requires an update to capture the "com.ibm.ws.security.thread.zos.context.provider" provider.

Without this, a file access error such this will occur when syncToOSThread is enabled with Java Batch:
BatchContainerRuntimeException: com.ibm.jzos.ZFileException:
//'filename': fopen() failed; EDC5092I An I/O abend
was trapped. (errno2=0xC00B0162); errno=92 errno2=0xc00b0162
last_op=43 errorCode=0x9130038

PR #4199, opened 2018-07-05
Save security context for batch sync to OS thread

dmbelina:
For #2659


dmbelina:
Tested manually and via newly added FATs

Issue #4211, opened 2018-07-09
Java 2 security issue in org.apache.cxf.transport.https.HttpsURLConnectionFactory


andymc12:
When using Java 2 security and HTTPS in outbound JAX-RS client requests, it is possible to see a failure like this:

```
ERROR: Caught exception attempting to call test method testPatchOptions on servlet jaxrs21.fat.patch.PatchTestServlet
javax.ws.rs.ProcessingException: java.io.IOException: Error while initializing secure socket
at org.apache.cxf.jaxrs.client.AbstractClient.checkClientException(AbstractClient.java:643)
at org.apache.cxf.jaxrs.client.AbstractClient.preProcessResult(AbstractClient.java:619)
at org.apache.cxf.jaxrs.client.WebClient.doResponse(WebClient.java:1114)
at org.apache.cxf.jaxrs.client.WebClient.doChainedInvocation(WebClient.java:1051)
at org.apache.cxf.jaxrs.client.WebClient.doInvoke(WebClient.java:897)
at org.apache.cxf.jaxrs.client.WebClient.doInvoke(WebClient.java:866)
at org.apache.cxf.jaxrs.client.WebClient.invoke(WebClient.java:431)
at org.apache.cxf.jaxrs.client.SyncInvokerImpl.method(SyncInvokerImpl.java:135)
at org.apache.cxf.jaxrs.client.SyncInvokerImpl.method(SyncInvokerImpl.java:130)
at org.apache.cxf.jaxrs.client.SyncInvokerImpl.options(SyncInvokerImpl.java:70)
at org.apache.cxf.jaxrs.client.spec.InvocationBuilderImpl.options(InvocationBuilderImpl.java:142)
at jaxrs21.fat.patch.PatchTestServlet.testPatchOptions(PatchTestServlet.java:53)
at componenttest.app.FATServlet.doGet(FATServlet.java:67)
at javax.servlet.http.HttpServlet.service(HttpServlet.java:686)
at javax.servlet.http.HttpServlet.service(HttpServlet.java:791)
at com.ibm.ws.webcontainer.servlet.ServletWrapper.service(ServletWrapper.java:1255)
at com.ibm.ws.webcontainer.servlet.ServletWrapper.handleRequest(ServletWrapper.java:743)
at com.ibm.ws.webcontainer.servlet.ServletWrapper.handleRequest(ServletWrapper.java:440)
[...]

PR #4212, opened 2018-07-09
Issue 4211: Add doPriv around conn.setSSLSocketFactory

andymc12:
This is fixed in CXF 3.2.5 - this part can be removed once we upgrade.

This resolves the Java 2 security exception throw when creating a new SSL socket factory for outbound JAX-RS clients.


andymc12:
#build - this resolves issue #4211

PR #4230, opened 2018-07-10
minimal completion stage

njr-11:
Implement the minimalCompletionStage method for the managed completable future.  Implementation of CompletableFuture.minimalCompletionStage cannot be reused here because the resulting stage creates dependent stages that do not inherit the newIncompleteFuture type (ManagedCompletableFuture) of the stage that creates the minimal stage.


njr-11:
#build

PR #4238, opened 2018-07-11
Revert "force the use of tls 1.2"

fritze2:
Reverts OpenLiberty/open-liberty#4009

Since #4009 changes are delivered, there is no longer a need for this workaround.


fritze2:
#build

PR #4242, opened 2018-07-12
Enable overriding install map jar

ericglau:
In the install map, support a new key for specifying the File object of the install map jar itself, instead of using the one in the runtime dir.


ericglau:
#build

ericglau:
Build failures are unrelated.

PR #3952, opened 2018-06-12
Add global error when user registry is not found

kristip17:
Fixes #2489

PR #4243, opened 2018-07-12
Revert "Add global error when user registry is not found"

Azquelt:
Reverts OpenLiberty/open-liberty#3952

I think this broke the build because the version of `com.ibm.websphere.appserver.spi.federatedRepository` was updated, but the reference to the javadoc zip generated from this bundle was not updated in the `com.ibm.websphere.appserver.federatedRegistry-1.0.feature` file.

I'm not sure why this didn't cause the personal build to fail.


Azquelt:
No ID or L2 review needed because this is a revert

PR #4258, opened 2018-07-12
adding support for directory-Based Repositories

### LINKED ISSUE #4256: Support From-parameter in mapBasedInstallKernel
support the passing of a "from"-parameter by adding a DirectoryRepositoryConnection connection in the mapBasedInstallKernel

hermankailey:
fixes #4256

### DIFF
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
@@ -42,6 +42,7 @@ import com.ibm.ws.kernel.productinfo.ProductInfo;
 import com.ibm.ws.kernel.productinfo.ProductInfoParseException;
 import com.ibm.ws.kernel.productinfo.ProductInfoReplaceException;
 import com.ibm.ws.product.utility.extension.ifix.xml.IFixInfo;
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -521,6 +535,7 @@ private void resolve(Object installObject) {
             }
             for (List<RepositoryResource> item : resolveResult) {
                 for (RepositoryResource repoResrc : item) {
+                    featuresResolved.add(repoResrc.getId());
```
ericglau: If the resource is from the directory based repository, are its Maven coordinates null?  If so, should we return either Maven coordinates or the ID, but not both?  Otherwise each item would have two entries in the result, which doesn't cause any issues but is confusing.

hermankailey: When it is a directoryBasedRepository the ID is a relative path (regular ids from the maven repo esas are not paths), and Maven coordinates are null for directoryBasedRepository IDs. There are no duplicates and any nonconforming ids/nulls are removed after it passes through downloadEsas() in the plugin.

PR #4282, opened 2018-07-16
revert start of changes to detect Mongo 3.x

njr-11:
A better alternative has been identified.  We will no longer be updating the mongo feature for versions 3.x.  Instead it will be documented how the user can use CDI Producer to programmatically configure and share a single instance across their application.  The mongo feature is obsolete and will not be updated, therefore, the very few changes that were made to start looking into 3.x support will be reverted under this pull.


njr-11:
#build

njr-11:
#build

mhldr:
issues discussed in #was-liberty-ops caused the build to fail so I'm respinning this incase it's needed (you ***may*** already have the results you need but if not hopefully this respin will help)
#build

njr-11:
#build

PR #4299, opened 2018-07-17
JPA 22 FAT: Initial drop of multi-db support (updated Callback bucket)

jgrassel:

Updated the com.ibm.ws.jpa_22_fat to support the ability to drop/create/update tables without relying on the JPA provider implementation's drop-create table support.  Updated Callback bucket to leverage this fat-feature (also renamed callback tests to make them easier to spot among the other unit tests)


jgrassel:
#build

mhldr:
issues discussed in #was-liberty-ops caused the build to fail so I'm respinning this incase it's needed (you ***may*** already have the results you need but if not hopefully this respin will help)
#build

PR #4302, opened 2018-07-17
Update JPA Container features to include internal JAXB-2.2 feature.

jgrassel:

Reason as per ST discussion with @aguibert:

Andrew: at one point I found out that we don't trust the JDK's copy of jaxb in terms of j2sec, but we do trust our own copy from the jaxb-2.2 feature | 2:03:28 PM
so I'd prefer that we refactor our features to use our internal jaxb-2.2 feature (com.ibm.websphere.appserver.internal.jaxb-2.2) instead of the JDK copy. | 2:04:08 PM
this will also help us out in JDK 11 when JAX-B is removed from the JDK entirely | 2:04:23 PM

Updated to remove dynamic import in jpa container bnd.bnd file.


wraschke:
#build

aguibert:
custom pbuild here: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_IZIAII9nEeiN-KJL-9czGg

PR #4315, opened 2018-07-18
support ignoreCase for getUsers method

### LINKED ISSUE #3853: basicRegistry-1.0's 'ignoreCaseForAuthentication' attribute does not apply to getUsers(...) method
When federating the basic registry that has 'ignoreCaseForAuthentication' enabled, a user is unable to complete a login when using a case that does NOT match the user as in the basic registry. The checkPassword(...) method works, but federated repository does other lookups that include a getUsers(...) call that now fails b/c the case does not match.

The getUsers(...) method should be consistent with both the checkPassword(...) and isValidUser(...) methods, which both consider case insensitivity.

The convertToRegex(...) method could be updated to prepend the string "(?i)" to the filter if 'ignoreCaseForAuthentication' is enabled.

toshiyamamoto:
fixes #3853


toshiyamamoto:
#build

PR #4324, opened 2018-07-19
Port jdbc_bvt test bucket to Open Liberty

alexsm82:
Convert the jdbc_bvt test bucket to a FAT test and port it to Open Liberty.  This new test bucket will be used for testing accessing JDBC drivers via DriverManager.


alexsm82:
#build

PR #4339, opened 2018-07-23
use driver package to identify known data source class

njr-11:
When the existing mechanism to infer data source impl class names based upon the properties element type and the driver jar names don't match anything, we can load the JDBC vendor's java.sql.Driver and use its package name to help identify which data source impl class should be used.

This capability will be gated for internal testing only until it is determined that the feature is ready to GA.


njr-11:
#build

PR #4342, opened 2018-07-23
Issue #4341: Fix case where if PageControl's 'startIndex' is >0 and '…

### LINKED ISSUE #4341: PageControl's 'startIndex' is not honored when 'size' is greater than results
Federate repository search calls with a PageControl where 'startIndex' > 0 and 'size' is greater than the number of results will return all results and not the subset of results starting at 'startIndex'.

For example, the following PageControl:

```
com.ibm.wsspi.security.wim.model.PageControl=
[
   size=500
   startIndex=2
]
```

when there are less than 500 results would result in all results being returned, including the undesired results at index 0 and index 1.

**Note:** this bug is only reachable using the SCIM feature.

jvanhill:
…size' is greater than the results returned from the search, all results will be returned. The results should not contain those prior to the 'startIndex'.

fixes #4341


jvanhill:
#build

jvanhill:
Re-building since I changed one line of code slightly as a result of a code review.

#build

PR #4344, opened 2018-07-24
use driver class name to guess data source class names

njr-11:
When Liberty does not have built-in knowledge of the JDBC driver being used, it can attempt to guess the data source class names based on the java.sql.Driver impl class name by replacing "Driver" with *DataSource and checking for its existence and that it implements the proper type.  This will only work some of the time, but it seems to be a pattern that a number of JDBC drivers use, and will provide a better user experience in those cases, because then the user will not need to go digging through vendor docs or JAR file to find the data source class names on their own.


njr-11:
#build

Issue #4111, opened 2018-06-25
Create a SymbolResolver interface to allow other symbol types other than environment variables


kaczyns:
The `SymbolRegistry` allows environment variables to be specifying by prefixing the variable name with `env.`.  In this case, the registry will do a `System.getenv()` to look up the environment variable value.  There are other types of variables that could do the same if we provided a way for others to define a way to lookup something for a prefix other than 'env'.

In the `com.ibm.ws.kernel.service.location` project, we can create a new package that will be exported.  The `bnd.bnd` will need to be updated to export this package.  The new package can contain a new interface, `SymbolResolver` which is capable of looking up some string and retrieving its value.  The interface should contain two methods, one that returns a list of prefixes which the resolver supports, and the second that takes the symbol to resolve as a String, and returns the value (if any) (ie.  `public String resolveSymbol(String symbol)`).  

With this interface created, the existing code in `SymbolRegistry` which deals with env variables can be changed to use an implementation of `SymbolResolver` that deals with environment variables.  We should create this implementation, and register it in the bundle activator for this project (`com.ibm.ws.kernel.service.location.internal.Activator.java`).  The `SymbolRegistry` class will need to somehow receive implementations of the `SymbolResolver` interface.  Remember that there can be more than one of these symbol resolvers registered, and if so we'll need to look at which one is appropriate to use.  It is possible that we can use declarative services, but, it may be too early for that and so we might need to use a service tracker instead.  In either case, when we get into the `resolveStringSymbols()` method towards the bottom where we check to see if the symbol has an `env.` prefix, we should instead look thru the registered `SymbolResolver` implementations and see if one is registered for the prefix we have (in this case the `env` prefix) and if we do, call it to get the value.

At this point the framework should be extensible enough that we can add a second SymbolResolver if we wish.

PR #4355, opened 2018-07-24
4111 symbol resolver #4111

mmerali96:
#4111 Create a SymbolResolver interface to allow other symbol types other than environment variables.


dev/com.ibm.ws.kernel.service.location/src/com/ibm/ws/kernel/service/location/internal/EnvVariableSymbolResolver.java
```diff
@@ -31,10 +34,15 @@ public EnvVariableSymbolResolver() {}
...
 	/**
 	 * Gets environment variables from system using variable name prefix as input
+	 * If symbol could not be resolved, print error message and return null string
 	 */
 	@Override
 	public String resolveSymbol(String symbol) {
-		return System.getenv(symbol);
+		String envVar = System.getenv(symbol);
+		if (envVar == null) {
+			System.out.println("EnvVariableSymbolResolver: ENV variable could not be resolved");
```
tevans78: use Tr.debug rather than System.out

tevans78: or possibly Tr.warning ... but then the message would need to be NLS enabled

PR #4357, opened 2018-07-25
test for inferring data source class from driver class

njr-11:
Write a test that ensures Liberty can infer a data source class name based on just the driver class name.


njr-11:
#build

PR #4365, opened 2018-07-25
Issue 4364: Allow filling existing messages.log/trace.log files before rolling

kgibm:
This changes the default behavior of `messages.log` and `trace.log` (but **not** `console.log`) processing so that, if said file already exists, then append to it before rolling (taking into account [maxFileSize](https://openliberty.io/config/rwlp_config_logging.html)).

This behavior may be reverted by specifying `com.ibm.ws.logging.fillExistingFile=false` in `bootstrap.properties` (although reverting does **not** work by setting `fillExistingFile="false"` in the `logging` element of `server.xml`).


kgibm:
#build

kgibm:
#build

kgibm:
#build

kgibm:
@OpenLiberty/message-reviewer Don't review NLS yet; this is going back to the drawing board.

PR #4376, opened 2018-07-26
scan for datasource implementation classes in driver JAR

njr-11:
If all other ways of inferring the data source implementation class fail, we can scan for classes in the same or a similar package to the Driver impl class with "DataSource" in their name and check if they implement the right type of data source.  This isn't certain to locate the data source classes, but seems like a nice compromise over loading every class in the JAR.


njr-11:
#build

PR #4379, opened 2018-07-27
Add units for bulkhead.waiting.duration metric

### LINKED ISSUE #4378: Bulkhead waiting queue duration metrics reported without units
The metric for the time tasks spend in the bulkhead queue is reported without specifying any units. It should specify that it's reported in nanoseconds.

## Symptoms

* queue duration metrics have the wrong name in the prometheus output `ft_name_bulkhead_waiting_duration` instead of `ft_name_bulkhead_waiting_duration_seconds`
* queue duration metrics are reported in nanoseconds instead of seconds in the prometheus output
* queue duration metrics do not report the units in the json output

Azquelt:
This metric is reported in nanoseconds. Not supplying the unit meant
that the metric name was wrong and the value was not converted to
seconds in the prometheus output format.

Fixes #4378


Azquelt:
#build

hutchig:
Merged as per Andrew Rouse [8:26 PM]
"Can someone review https://github.com/OpenLiberty/open-liberty/pull/4379
and merge if it's ok it has a green build."

PR #4384, opened 2018-07-30
move jdbc init code towards tolerance of driver class

njr-11:
The updates under this pull are still very early on and form some basic steps to start moving the JDBC code in the directly of accepting a java.sql.Driver instead of a data source impl class.  No new capability is made available here - many more steps are necessary before that can happen - but we can at least demonstrate that we are maintaining non-breakage of existing code path as we head in this direction.


njr-11:
#build

njr-11:
Build has 3 test failures due to timeout in an unrelated test bucket that does not even enable the jdbc feature.

PR #4391, opened 2018-07-30
locate driver and stringify properties

njr-11:
When URL is specified and the built-in logic for inferring data source classes doesn't find any, locate the Driver implementation that accepts the configured URL.  Make adjustments to the configured properties (stringify values and decode passwords) such that they can be supplied to the JDBC driver.  This gets us up to the point where there data source can be used to attempt a connection, which will fail indicating that the JDBC driver impl class isn't a data source because we haven't implemented code path beyond that point yet.


njr-11:
#build

njr-11:
#build

Issue #4392, opened 2018-07-30
Fix server hang issue when bootstrap.properties variable is incorrectly specified


hlhoots:

hlhoots:
Whenever a user specifies a variable incorrectly in bootstrap.properties and uses the variable in server.xml for a substitution, the server can potentially hang even though an error messages is output.  This fix resolves that issue, and ensure the server is started.

PR #4394, opened 2018-07-30
4392: Update to BundleProcessor to fix server hang for bad config

hlhoots:
PR for #4392


hlhoots:
#build

hlhoots:
Build had 5 failures but none of them are related to this change and/or had existing defects for the issues.

PR #4405, opened 2018-07-31
 Add initial support for getting a connection using Driver

alexsm82:
Initial support for getting a connection using java.sql.Driver.

### DIFF
--- a/dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
+++ b/dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
@@ -331,7 +331,6 @@ public class JDBCDriverService extends Observable implements LibraryChangeListen
      * <li>javax.sql.ConnectionPoolDataSource
      * <li>javax.sql.DataSource
      * <li>javax.sql.XADataSource
-     * <li>java.sql.Driver // TODO
      * </ul>
      * 
      * @param props typed data source properties
@@ -588,6 +587,39 @@ public class JDBCDriverService extends Observable implements LibraryChangeListen
             }
 
             return create(className, props);
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+    
+    /**
+     * Create a Driver
+     * 
+     * @param props typed data source properties
+     * @return the driver
+     * @throws SQLException if an error occurs
+     */
+    public Object createDriver(PropertyService props) throws SQLException {
+        lock.readLock().lock();
+        try {
+            if (!isInitialized)
+                try {
+                    // Switch to write lock for lazy initialization
+                    lock.readLock().unlock();
+                    lock.writeLock().lock();
+
+                    if (!isInitialized) {
+                        if (!loadFromApp())
+                            classloader = AdapterUtil.getClassLoaderWithPriv(sharedLib);
+                        isInitialized = true;
+                    }
+                } finally {
+                    // Downgrade to read lock for rest of method
+                    lock.readLock().lock();
+                    lock.writeLock().unlock();
+                }
+            
+            return loadDriver(props, classloader);
         } finally {
             lock.readLock().unlock();
         }
--- a/dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSManagedConnectionFactoryImpl.java
+++ b/dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSManagedConnectionFactoryImpl.java
@@ -22,6 +22,7 @@ import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException; 
 import java.security.PrivilegedExceptionAction;
 import java.sql.Connection;
+import java.sql.Driver;
 import java.sql.SQLException;
 import java.sql.SQLNonTransientConnectionException;
 import java.sql.Statement;
@@ -118,7 +119,7 @@ public class WSManagedConnectionFactoryImpl extends WSManagedConnectionFactory i
      * The underlying data source implementation from the JDBC vendor.
      * It could be a javax.sql.DataSource, javax.sql.ConnectionPoolDataSource, or javax.sql.XADataSource
      */
-    transient private CommonDataSource dataSource;
+    transient private Object dataSourceOrDriver;
 
     /**
      * Class name of data source or driver implementation from the JDBC vendor.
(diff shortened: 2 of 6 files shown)


dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -331,7 +331,7 @@ public T run() throws Exception {
      * <li>javax.sql.ConnectionPoolDataSource
      * <li>javax.sql.DataSource
      * <li>javax.sql.XADataSource
-     * <li>java.sql.Driver // TODO
+     * <li>java.sql.Driver
```
njr-11: It is incorrect to place Driver in this ordered list of data source precedence due to our reliance on the URL attribute

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -399,6 +399,7 @@ public Object createAnyDataSourceOrDriver(Properties props) throws SQLException
      * <li>javax.sql.XADataSource
      * <li>javax.sql.ConnectionPoolDataSource
      * <li>javax.sql.DataSource
+     * <li>java.sql.Driver
```
njr-11: It is incorrect to place Driver in this ordered list of data source precedence due to our reliance on the URL attribute

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -129,7 +129,7 @@
      * We should remove these classes from the Introspector cache
      * when the jdbcDriver is deactivated or modified.
      */
-    private final HashSet<Class<? extends CommonDataSource>> introspectedClasses = new HashSet<Class<? extends CommonDataSource>>();
```
njr-11: We do not introspect the Driver impl.  This ought to remain typed as CommonDataSource.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -227,7 +227,7 @@ private SQLException classNotFound(Throwable cause) {
      * @return the data source
      * @throws SQLException if an error occurs
      */
-    private <T extends CommonDataSource> T create(final String className, final Hashtable<?, ?> props) throws SQLException {
```
njr-11: This should not have changed from CommonDataSource to Object.
This method is very specific to data sources and should not be used to load a Driver impl.  Driver implementations are not populated with JavaBean properties via reflection, which is what this method does.  In the case where you are given a Driver class name (which is overshooting the scope of this delivery), it should be trivial to either use the ServiceLoader with the given class loader to load java.sql.Driver classes and pick the one that matches the desired class name, or otherwise directly load the Driver impl class & instantiate via reflection.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -499,7 +500,7 @@ public ConnectionPoolDataSource createConnectionPoolDataSource(Properties props)
                 }
             }
 
-            return create(className, props);
+            return (ConnectionPoolDataSource) create(className, props);
```
njr-11: unnecessary casts once create signature is restored

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -612,7 +646,7 @@ protected void deactivate(ComponentContext context) {
                         shutdownDerbyEmbedded();
                     classloader = null;
                 }
-                for (Iterator<Class<? extends CommonDataSource>> it = introspectedClasses.iterator(); it.hasNext(); it.remove())
```
njr-11: Can be restored to CommonDataSource after prior comments are addressed

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSManagedConnectionFactoryImpl.java
```diff
@@ -975,13 +1056,13 @@ private Object getTraceable(Object d) throws ResourceException {
...
 
     /**
-     * Retrieve the underlying DataSource for this ManagedConnectionFactory.
+     * Retrieve the underlying DataSource or Driver for this ManagedConnectionFactory.
      * 
-     * @return the underlying DataSource.
+     * @return the underlying DataSource or Driver.
      */
-    public CommonDataSource getUnderlyingDataSource() throws SQLException
+    public Object getUnderlyingDataSourceOrDriver() throws SQLException
```
njr-11: Don't change this signature until the consequences of doing so are investigated.  I believe this is used for wrapper pattern, and I don't want to see a driver instance being returned.  I would rather see NULL returned to ensure that unwrap doesn't allow operations on the Driver

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
alexsm82: I was just removing the TODO but I can remove the entry from the list entirely

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
alexsm82: Same as above, I was adding the entry to match the other list but I'll remove.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
alexsm82: Correct, I initially updated the create method to also be used for drivers but decided against it when refactoring with your changes.  I apparently forgot to update the method signature when I made that change.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
alexsm82: Updated

alexsm82:
The build had three failures, which matched to existing defects in unrelated components.  Note that there haven't been any changes since the build was run.  I had to rebase to resolve what git thought was a merge conflict, but I had already built on top of the commits so there wasn't any merging that actually needed to be done.

PR #4411, opened 2018-08-01
add null checks when adding to jeeComponentClass

benjamin-confino:


dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/BeanDeploymentArchiveImpl.java
```diff
@@ -369,6 +377,9 @@ private void initializeJEEComponentClasses(Set<String> allClassNames) throws CDI
         for (EjbDescriptor<?> ejb : ejbs) {
             if (ejb.isMessageDriven()) {
                 classes.add(ejb.getBeanClass());
+                if (ejb.getBeanClass() == null) {
```
Azquelt: For this case (and the one for managed beans) do we want to add the class to `classes` if it's `null`?

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/BeanDeploymentArchiveImpl.java
```diff
@@ -786,6 +797,9 @@ public void addEjbDescriptor(EjbDescriptor<?> ejbDescriptor) {
     public void addManagedBeanDescriptor(ManagedBeanDescriptor<?> managedBeanDescriptor) {
         if (getBeanDiscoveryMode() != BeanDiscoveryMode.NONE) {
             this.managedBeanClasses.add(managedBeanDescriptor.getBeanClass());
+            if (managedBeanDescriptor.getBeanClass() -= null) {
```
Azquelt: `-=` is probably not what you meant here

benjamin-confino: Correct, that's a typo

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/BeanDeploymentArchiveImpl.java
benjamin-confino: As I said in slack. I think for now we stick to the existing behaviour, but log it.

PR #4423, opened 2018-08-01
detect driver class based on DataSourceDefinition url

njr-11:
When DataSourceDefinition is specified without a className, we can infer the Driver class based on the URL by iterating through the Drivers available to the ServiceLoader and checking which one accepts the URL.


njr-11:
#build

PR #4435, opened 2018-08-02
infer data source class for DataSourceDefinition

njr-11:
When DataSourceDefinition lacks a className and url, we can discover available JDBC drivers and infer/discover available data source implementations.  This will make it unnecessary for the user to know the implementation class name in many cases.


njr-11:
#build

PR #4437, opened 2018-08-02
Update FeatureReplacementAction so ALL_SERVERS works at the test class level with injected servers

aguibert:
Previously the following would not work:

```java
public class ThirdPartyJAXBTest extends FATServletClient {

    @ClassRule
    public static RepeatTests r = RepeatTests.with(new FeatureReplacementAction("ldapRegistry-3.0").withID("LDAP"))
                    .andWith(new FeatureReplacementAction("ldapRegistry-3.0", "jpa-2.1").withID("JPA"))
                    .andWith(new FeatureReplacementAction("jpa-2.1", "jaxrs-2.0").withID("JAXRS"));

    @Server("jaxb_fat.no-jaxb-feature")
```

The workaround was to specify `.forServers(SERVER)` on each FeatureReplacementAction, which is more tedious than it needs to be.  This PR makes it so the server name does not need to be specified and that the above code will work.


aguibert:
#build

aguibert:
#build

aguibert:
#build

PR #4442, opened 2018-08-02
Issue #3679: Replace GenericData ArrayList with array

### LINKED ISSUE #3679: Fix GenericData so that we don't need null value placeholders

levin-noro:
Fixes #3679

### DIFF
--- a/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/data/AccessLogData.java
+++ b/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/data/AccessLogData.java
@@ -66,21 +66,6 @@ public class AccessLogData extends GenericData {
         setPair(index, NAMES1_1[index], l);
     }
 
-    private String getStringValue(int index) {
-        KeyValueStringPair kvp = (KeyValueStringPair) getPairs().get(index);
-        return kvp.getStringValue();
-    }
-
-    private int getIntValue(int index) {
-        KeyValueIntegerPair kvp = (KeyValueIntegerPair) getPairs().get(index);
-        return kvp.getIntValue();
-    }
-
-    private long getLongValue(int index) {
-        KeyValueLongPair kvp = (KeyValueLongPair) getPairs().get(index);
-        return kvp.getLongValue();
-    }
-
     public void setRequestStartTime(long l) {
         setPair(0, l);
     }
--- a/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/data/GenericData.java
+++ b/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/data/GenericData.java
@@ -10,68 +10,107 @@
  *******************************************************************************/
 package com.ibm.ws.logging.data;
 
-import java.util.ArrayList;
-
 public class GenericData {
 
-    private final ArrayList<KeyValuePair> pairs;
+    private final static int DEFAULT_SIZE = 16;
+
+    private KeyValuePair[] pairs;
 
     private String sourceType;
 
     private String jsonMessage = null;
 
+    private Integer lastIndex = -1;
+
     public GenericData() {
-        pairs = new ArrayList<KeyValuePair>();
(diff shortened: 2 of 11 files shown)


levin-noro:
#build

levin-noro:
#build

levin-noro:
#build

levin-noro:
#build

levin-noro:
#build

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/data/GenericData.java
```diff
@@ -10,68 +10,94 @@
...
-        KeyValueStringPair kvp = new KeyValueStringPair(key, value);
-        pairs.set(index, kvp);
+        pairs = new KeyValuePair[size];
     }
 
     public void setPair(int index, String key, String value) {
         KeyValueStringPair kvp = new KeyValueStringPair(key, value);
-        pairs.add(index, kvp);
+        ensureCapacityAndSetPair(index, kvp);
+        setLastIndex(index);
```
donbourne: code to set the lastIndex (currently in setLastIndex) should be inside ensureCapacityAndSetPair since every time you want to setLastIndex you want to potentially set lastIndex.  This would also be consistent with how you're handling ensureCapacityAndAddPair.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/data/AccessLogData.java
```diff
@@ -67,17 +67,17 @@ private void setPair(int index, long l) {
     }
 
     private String getStringValue(int index) {
```
donbourne: should move getStringValue / getIntValue / getLongValue into GenericData.  We can then remove those methods from all of the GenericData subclasses

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/data/LogTraceData.java
```diff
@@ -93,22 +93,22 @@ private void setPair(int index, long l) {
     }
 
     private String getStringValue(int index) {
-        KeyValueStringPair kvp = (KeyValueStringPair) getPairs().get(index);
-        return kvp.getStringValue();
+        KeyValueStringPair kvp = (KeyValueStringPair) getPairs()[index];
+        return kvp == null ? null : kvp.getStringValue();
```
donbourne: to be consistent with getIntValue / getLongValue, we should not check for nulls here and just let an NPE happen in cases where a kvp doesn't exist.  That way anyone using this class can rely on NPEs as the way of indicating a field is missing.

levin-noro: I changed this to handle null because there were a few fields in CollectorJSONUtils that were calling get<FieldName> on some string KVPs that were running into NPEs. For example, getMessageId(), getThrowable(), and getSequence()

donbourne: This is a case where Java's Optional class could help, but I'm hesitant to introduce that as it would result in a lot of extra method calls for this use case.  Instead, could we add a second getStringValue that takes two parms, the second indicating the value to return if the key isn't present?  If we do that, should provide same for int/long for consistency.

```
public String getStringValue(int index, String default) { ... }
```

levin-noro: As per our discussion, I'll create a new issue for this change and leave my changes as is.

levin-noro:
Manual Testing:
- [ ] Compared LogstashCollector 1.1 output for audit, accessLog, ffdc, message, and trace events with master
- [ ] Compared LogstashCollector 1.0 output for audit, accessLog, ffdc, message, and trace events with master
- [ ] Compared trace.log, message.log, console.log audit.log, trace.log, and ffdc logs with master

Issue #4399, opened 2018-07-31
18.0.0.2 CWWKZ0002E: NoClassDefFoundError: javax/ws/rs/core/Application


mikekuzak:
HI .. 

I'm getting a strange exception when starting up the server. It looks like the microprofile is not loaded fully before the app starts. 

```
Launching defaultServer (Open Liberty 18.0.0.2/wlp-1.0.21.cl180220180619-0403) on Java HotSpot(TM) 64-Bit Server VM, version 1.8.0_181-b13 (en_US)
[AUDIT   ] CWWKE0001I: The server defaultServer has been launched.
[AUDIT   ] CWWKG0093A: Processing configuration drop-ins resource: /opt/ol/wlp/usr/servers/defaultServer/configDropins/defaults/open-default-port.xml
[AUDIT   ] CWWKS4104A: LTPA keys created in 0.834 seconds. LTPA key file: /opt/ol/wlp/output/defaultServer/resources/security/ltpa.keys
[AUDIT   ] CWWKZ0058I: Monitoring dropins for applications.
[AUDIT   ] CWPKI0803A: SSL certificate created in 1.574 seconds. SSL key file: /opt/ol/wlp/output/defaultServer/resources/security/key.jks
[AUDIT   ] CWWKT0016I: Web application available (default_host): http://app-scheduler-gateway-85bb5759f7-ffcmd:9080/jwt/
[AUDIT   ] CWWKT0016I: Web application available (default_host): http://app-scheduler-gateway-85bb5759f7-ffcmd:9080/openapi/ui/
[AUDIT   ] CWWKT0016I: Web application available (default_host): http://app-scheduler-gateway-85bb5759f7-ffcmd:9080/ibm/api/
[AUDIT   ] CWWKT0016I: Web application available (default_host): http://app-scheduler-gateway-85bb5759f7-ffcmd:9080/health/
[AUDIT   ] CWWKT0016I: Web application available (default_host): http://app-scheduler-gateway-85bb5759f7-ffcmd:9080/metrics/
[AUDIT   ] CWWKT0016I: Web application available (default_host): http://app-scheduler-gateway-85bb5759f7-ffcmd:9080/openapi/
[ERROR   ] CWWKZ0002E: An exception occurred while starting the application mcm-scheduler-gateway. The exception message was: java.lang.NoClassDefFoundError: javax/ws/rs/core/Application
[...]

PR #4407, opened 2018-07-31
Remove dependency on internal JAX-B from wimcore-1.0

aguibert:
fixes https://github.com/OpenLiberty/open-liberty/issues/4404

PR #4445, opened 2018-08-02
Issue 4399: Ensure GW bundle is started before allowing classloads

andymc12:
This root problem of this issue appears to be that the gateway bundle
is not activated until after the application has attempted to use it's
classloader to load classes. This fix resolves the issue by preventing
the createGatewayBundleClassLoader method from exiting until the bundle
is started or a timeout (default 10 seconds) has expired.

This pull request should resolve issue #4399.


andymc12:
#build

andymc12:
#build

andymc12:
This looks like a misdirection.  It seems like #4407 may resolve this issue without needing changes to the classloader.  I'll leave this open for a few days just in case 4407 does not solve this.

Issue #4404, opened 2018-07-31
JAXB classloading issue after upgrading to 18.0.0.2


lmsurpre:
After upgrading from 18.0.0.1 to 18.0.0.2, we started hitting a strange JAXB error when we try to explicitly load the JAXBContext for our model classes: 
```
[7/28/18 6:11:07:486 UTC] 0000002a id=         watsonhealth.fhir.server.listener.FHIRServletContextListenerEEncountered an exception while initializing the servlet context.
java.lang.Error: javax.xml.bind.JAXBException: Unable to create context
 - with linked exception:
[java.lang.reflect.InvocationTargetException]
	at com.ibm.watsonhealth.fhir.model.util.FHIRUtil.createContext(FHIRUtil.java:225)
	at com.ibm.watsonhealth.fhir.model.util.FHIRUtil.<clinit>(FHIRUtil.java:147)
	at com.ibm.watsonhealth.fhir.server.listener.FHIRServletContextListener.contextInitialized(FHIRServletContextListener.java:74)
	at com.ibm.ws.webcontainer.webapp.WebApp.notifyServletContextCreated(WebApp.java:2376)
	at com.ibm.ws.webcontainer31.osgi.webapp.WebApp31.notifyServletContextCreated(WebApp31.java:514)
	at com.ibm.ws.webcontainer.webapp.WebApp.initialize(WebApp.java:1004)
	at com.ibm.ws.webcontainer.webapp.WebApp.initialize(WebApp.java:6566)
	at com.ibm.ws.webcontainer.osgi.DynamicVirtualHost.startWebApp(DynamicVirtualHost.java:467)
	at com.ibm.ws.webcontainer.osgi.DynamicVirtualHost.startWebApplication(DynamicVirtualHost.java:462)
	at com.ibm.ws.webcontainer.osgi.WebContainer.startWebApplication(WebContainer.java:1132)
	at com.ibm.ws.webcontainer.osgi.WebContainer.startModule(WebContainer.java:937)
	at com.ibm.ws.app.manager.module.internal.ModuleHandlerBase.deployModule(ModuleHandlerBase.java:99)
	at com.ibm.ws.app.manager.module.internal.DeployedModuleInfoImpl.installModule(DeployedModuleInfoImpl.java:49)
	at com.ibm.ws.app.manager.module.internal.DeployedAppInfoBase.deployModules(DeployedAppInfoBase.java:429)
[...]

PR #4446, opened 2018-08-02
Add required-osgi-ee bundle attribute to feature resolver

### LINKED ISSUE #4403: Handling removed JDK APIs in Java 9 and beyond
As of Java 9, a number of modules were removed from the default bootclasspath:
 - javax.activation
 - CORBA API
 - JTA API (javax.transaction)
 - JAX-B API
 - JAX-WS API (and annotations)
Although they were not on the default bootclasspath, they were still present in the JDK.  In java 9-10 we worked around this issue by specifying `--add-modules=java.se.ee` which added them back onto the bootclasspath.
 
In Java 11, these APIs are gone completely from the JDK, so our workaround got taken away.  Now we need to provide these APIs to the Liberty runtime and its apps on Java 11 and beyond.
 
Here are the options I have examined thus far for the JAX-B case:
 1) create new internal JAX-B feature that does not expose API to app classloader, refactor Liberty components to use this instead of JDK
 2) use bundle-fragments that only resolve on JDK9+ and export JAX-B packages
 3) add a new feature in wlp/lib/platform/ which pulls in all of the removed APIs as features as type="boot.jar", which adds the jars to the classpath of the OSGi framework.  The kernel bootstrap code would only provision this feature on JDK9+.
 
I stopped pursuing option (1) once I hit unresolvable classloading conflicts and realized this could cause a breaking behavior change for customers that remain on Java 8.
Tom suggested option (2) which I never fully understood but he realized it may cause a similar issue to option (1) where we cause a behavior change for customers that remain on java 8.
Right now option (3) is the leading approach, since it does not effect existing Java 8 users in any way, and tricks the rest of Liberty into thinking that the removed APIs are still coming from the JDK/system.bundle.

aguibert:
Fixes #4403


aguibert:
@tjwatson adding you for code review here.  Note that this PR builds on top of some commits in PR #4407 which are already being reviewed by Andy McCright.  So you only need to review these commits:
https://github.com/OpenLiberty/open-liberty/pull/4446/commits/1165cd92042d53f883d4fcd0b264178b3ae22b7c
https://github.com/OpenLiberty/open-liberty/pull/4446/commits/bb0ef068999c5d33e8399eddc6bb6c001095af4c

aguibert:
#build

aguibert:
#build

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/FeatureResourceImpl.java
```diff
@@ -373,4 +373,20 @@ public String getBundleRepositoryType() {
...
+            if (result == null)
+                result = "";
+            if (!_requiredOSGiEE.compareAndSet(null, result))
+                result = _requiredOSGiEE.get();
+        }
+        if ("".equals(result))
+            return null;
+        else
+            return result;
+    }
```
tjwatson: This seems overly complicated to cache a map lookup of the key "required-osgi-ee:" into an atomic reference.  I think we should just go with a simple map lookup here:
```
    public String getRequiredOSGiEE() {
        // Directive names are in the attributes map, but end with a colon
        return _rawAttributes.get("required-osgi-ee:");
    }
```
The tolerates method does this caching to avoid splitting the raw string to a List<String> each call.  But here we just have a simple String so no need to have a complicated cache for that.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/Provisioner.java
```diff
@@ -270,6 +274,36 @@ public boolean handle(FeatureResource fr) {
...
+                            public Map<String, Object> getAttributes() {
+                                return Collections.emptyMap();
+                            }
+                        });
+                        if (foundEEs.isEmpty()) {
+                            // require-osgi-ee capability not satisfied, do not install
+                            return true;
+                        }
+                    }
+
```
tjwatson: I would like to see the code to check for required OSGi EEs split out into a separate method.  This bundleList.foreach code is already too long in my opinion.  So we can reduce the code in this method to something like this:

```
                    if (missingRequiredOSGiEE(fr)) {
                            // require-osgi-ee capability not satisfied, do not install
                            return true;
                    }
```

dev/com.ibm.websphere.appserver.features/visibility/private/com.ibm.websphere.appserver.internal.optional.jaxb-2.2.feature
```diff
@@ -0,0 +1,16 @@
...
+visibility=private
+IBM-App-ForceRestart: uninstall, \
+ install
+IBM-Process-Types: client, \
+ server
+Subsystem-Name: Java XML Bindings 2.2 for Java 9 and above
+-features=\
+  com.ibm.websphere.appserver.classloading-1.0
+-bundles=\
+  com.ibm.websphere.javaee.jaxb.2.2; required-osgi-ee:="(&(osgi.ee=JavaSE)(version>=9))"; location:="dev/api/spec/,lib/"; mavenCoordinates="javax.xml.bind:jaxb-api:2.2.12", \
```
tjwatson: Shouldn't these technically use >=11?  My understanding is the jaxb stuff is going away in java 11.

dev/com.ibm.websphere.appserver.features/visibility/public/jaxb-2.2/com.ibm.websphere.appserver.jaxb-2.2.feature
```diff
@@ -14,8 +14,11 @@ IBM-Process-Types: client, \
  server
 Subsystem-Name: Java XML Bindings 2.2
 -features=\
-  com.ibm.websphere.appserver.classloading-1.0,\
-  com.ibm.websphere.appserver.internal.jaxb-2.2
```
tjwatson: Why did we lose the internal.jaxb-2.2 feature that always installed the bundle.  Is this not used by any other feature that always wants our jaxb bundles even on Java 8?

dev/com.ibm.websphere.appserver.features/visibility/private/com.ibm.websphere.appserver.internal.optional.jaxb-2.2.feature
aguibert: It is gone from the JDK completely in Java 11, but it is not on the default bootcp in Java 9/10.  I don't want to specify `--add-modules java.se.ee` for just 9/10 and have to do something different for 11+.  So this approach will give us consistent behavior accross 9/10/11

dev/com.ibm.websphere.appserver.features/visibility/public/jaxb-2.2/com.ibm.websphere.appserver.jaxb-2.2.feature
aguibert: I'm getting rid of the internal.jaxb-2.2 feature because it doesn't keep JAX-B internal (issue #4404 shows that JAX-B API still leaked out into the app classloader).  So the new approach will be this "optional and conditional" JAXB feature which only resolves on JDK 9+, but the public jaxb-2.2 feature needs to continue to resolve JAX-B regardless of JDK version.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/Provisioner.java
aguibert: agreed

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/FeatureResourceImpl.java
aguibert: I agree, but I wanted to follow the convention of the other attributes in this class:
```java
    private final AtomicReference<VersionRange> _range = new AtomicReference<VersionRange>();
    private final AtomicReference<String> _location = new AtomicReference<String>();
    private final AtomicReference<List<String>> _osList = new AtomicReference<List<String>>();
    private final AtomicReference<Map<String, String>> _attributes = new AtomicReference<Map<String, String>>();
    private final AtomicReference<Map<String, String>> _directives = new AtomicReference<Map<String, String>>();
    private final AtomicInteger _startLevel = new AtomicInteger(-1);
    private final AtomicReference<List<String>> _tolerates = new AtomicReference<List<String>>();
```
(Specifically, the `_location` attribute is what I followed the convention of since it is just a single String)

tjwatson: I see, but I think the way `_location` does it is unnecessary also.  I would prefer not to follow that pattern if we think it is unnecessary work.

aguibert:
custom pbuild: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_rYklJJc0EeiA46HHn3eUcA

aguibert:
custom pbuild: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_CQtb8JkhEeig4deLkOQTuA&tab=com.ibm.team.build.web.ui.internal.editors.result.overview.OverviewPage

Issue #4447, opened 2018-08-03
WABInstaller throws IllegalStateException when swtiching features


navidsh:
The problem happens if one feature is replaced with another feature AND both features include a bundle that includes a dynamic `Web-ContextPath` e.g. `Web-ContextPath: @contextRoot`. In this case the following exception will be thrown:
```
[02/08/18 16:33:57:977 EDT] 0000000e LogService-0-org.eclipse.osgi                                E CWWKE0701E: FrameworkEvent ERROR Bundle:org.eclipse.osgi(id=0) java.lang.IllegalStateException: A bundle is already attempting to use the name: publicMetricsURL - com.ibm.ws.microprofile.metrics.public_1.0.22.201808021436 [152]
    at com.ibm.ws.app.manager.wab.internal.WABInstaller$WABTrackerCustomizer.createConfigurableWAB(WABInstaller.java:706)
    at com.ibm.ws.app.manager.wab.internal.WABInstaller$WABTrackerCustomizer.addingBundle(WABInstaller.java:669)
    at com.ibm.ws.app.manager.wab.internal.WABInstaller$WABTrackerCustomizer.addingBundle(WABInstaller.java:648)
    at org.osgi.util.tracker.BundleTracker$Tracked.customizerAdding(BundleTracker.java:469)
    at org.osgi.util.tracker.BundleTracker$Tracked.customizerAdding(BundleTracker.java:1)
    at org.osgi.util.tracker.AbstractTracked.trackAdding(AbstractTracked.java:256)
    at org.osgi.util.tracker.AbstractTracked.track(AbstractTracked.java:229)
    at org.osgi.util.tracker.BundleTracker$Tracked.bundleChanged(BundleTracker.java:444)
    at org.eclipse.osgi.internal.framework.BundleContextImpl.dispatchEvent(BundleContextImpl.java:908)
    at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:230)
    at org.eclipse.osgi.framework.eventmgr.ListenerQueue.dispatchEventSynchronous(ListenerQueue.java:148)
    at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEventPrivileged(EquinoxEventPublisher.java:213)
    at org.eclipse.osgi.internal.framework.EquinoxEventPublisher.publishBundleEvent(EquinoxEventPublisher.java:120)
[...]

PR #4448, opened 2018-08-03
Issue #4447 - Ensure WAB bundles are always unconfigured

navidsh:
PR for Issue #4447:
- Ensure WAB bundles are always unconfigured


navidsh:
#build

Issue #4458, opened 2018-08-03
ORMDiagnostics: Add EclipseLink API annotations


dazey3:
The JPA ORMDiagnostics Tool currently filters out EclipseLink annotations. We want to include those annotations in what we collect.

dazey3:
see #5612

PR #4460, opened 2018-08-03
Issue 4458: Add EclipseLink API to ORMDiagnostics

dazey3:
for #4458 


dazey3:
It'd be nice if there was some kind of FAT that would allow me to add a test for this. Perhaps we can address this in a future task when we feel the reader portion is more complete

dazey3:
Delivered under #5612

Issue #4373, opened 2018-07-26
JAX-RS endpoint needs to support interceptor


Emily-Jiang:
JAX-RS resources should support the standard CDI support, which should include field, method and constructor injections and the use of interceptors. At the moment, it seems only field and method inejctions are supported. Most MicroProfile interceptors are not working as the interceptor support is not there. This put Liberty in the odd situation as Payara and other app server work except Liberty. MP code generator has to workaround this.

PR #4461, opened 2018-08-03
Issue 4373: Allow `@Inject` on JAX-RS resource constructors

andymc12:
Currently, JAX-RS resources only allow no-arg constructors or constructors with parameters annotated with `@Context` or `@*Param` (if per-request scope).  To be compliant with CDI, we really should allow CDI to inject constructor parameters.  This change enables that.

This fixes one part of Issue #4373.


andymc12:
#build

andymc12:
#build

PR #4476, opened 2018-08-06
Handle loginTimeout with Driver type

alexsm82:
Handle and test loginTimeout with Driver type.  This includes rejecting a loginTimeout when specified in the URL or as a property of a DataSource using the Driver type.


dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -594,5 +594,11 @@ KERBEROS_NOT_SUPPORTED_WARNING.explanation= The Application Server does not supp
 KERBEROS_NOT_SUPPORTED_WARNING.useraction= Disable Kerberos, or use a DB2 backend with the DB2 universal JDBC driver.
 
 # 9544-9548 deleted
+
+# {0} name of the DataSource
+INVALID_LOGINTIMEOUT=DSRA9549E: The Application Server does not support the loginTimeout property specified in the URL or properties of the {0} DataSource.
+INVALID_LOGINTIMEOUT.explanation=The Application Server does not support the loginTimeout property when specified for a DataSource that uses the Driver type.
+INVALID_LOGINTIMEOUT.useraction=Remove the loginTimeout property or modify the DataSource configuration to use a DataSource type rather than a Driver type.
```
njr-11: Reverse the order of the suggestions to assume that the customer wants to keep the functionality that they configured.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -377,6 +377,11 @@ public Object createAnyDataSourceOrDriver(Properties props) throws SQLException
             if (nonshipFunction) {
                 String url = props.getProperty("URL", props.getProperty("url"));
                 if (url != null) {
+                    if(url.toLowerCase().contains("logintimeout") || props.containsKey("loginTimeout")) {
+                        SQLNonTransientException failure = connectorSvc.ignoreWarnOrFail(tc, null, SQLNonTransientException.class, "INVALID_LOGINTIMEOUT", dataSourceID);
```
njr-11: ignoreWarnOrFail is not appropriate here unless you are able to parse the loginTimeout out of the URL

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -632,6 +642,12 @@ public Object getDriver(String url) throws SQLException {
                     lock.writeLock().unlock();
                 }
 
+            if(url.toLowerCase().contains("logintimeout") || props.containsKey("loginTimeout")) {
+                SQLNonTransientException failure = connectorSvc.ignoreWarnOrFail(tc, null, SQLNonTransientException.class, "INVALID_LOGINTIMEOUT", dataSourceID);
+                if (failure != null)
+                    throw failure;
+            }
+
```
njr-11: This code block is copied 3 times, each time right before loadDriver(url, classloader).
Why not move it to the loadDriver method and have it in a single place?

dev/com.ibm.ws.jdbc_fat_driver/fat/src/com/ibm/ws/jdbc/fat/driver/JDBCDriverManagerTest.java
```diff
@@ -62,6 +62,7 @@ public static void setUp() throws Exception {
 
     @AfterClass
     public static void tearDown() throws Exception {
-        server.stopServer("DSRA8020E.*internal.nonship.function"); // TODO remove once the capability becomes GA
+        server.stopServer("DSRA8020E.*internal.nonship.function", // TODO remove once the capability becomes GA
+                          "CWWKE0701E"); //Expected in testGetSetLoginTimeout
```
njr-11: I assume this should be removed once ignoreWarnFail logic is replace with always failing

dev/com.ibm.ws.jdbc_fat_driver/publish/servers/com.ibm.ws.jdbc.fat.driver/server.xml
```diff
@@ -28,7 +28,7 @@
     
     <dataSource id="fatDataSource" jndiName="jdbc/fatDataSource">
       <jdbcDriver libraryRef="FATDriverLib" internal.nonship.function="This is for internal development only. Never use this in production"/>
-      <properties databaseName="memory:jdbcdriver1" autoCreate="true" user="dbuser1" password="{xor}Oz0vKDtu" />
+      <properties databaseName="memory:jdbcdriver1" autoCreate="true" user="dbuser1" password="{xor}Oz0vKDtu" loginTimeout="660"/>
```
njr-11: Will not be able to add loginTimeout to this config once ignoreWarnFail logic is replaced

dev/com.ibm.ws.jdbc_fat_driver/publish/servers/com.ibm.ws.jdbc.fat.driver/server.xml
```diff
@@ -37,6 +37,12 @@
       <containerAuthData user="dbuser1" password="{xor}Oz0vKDtu"/>
     </dataSource>
     
+    <dataSource id="fatDriverInvalidURLLoginTimeout" jndiName="jdbc/fatDriverInvalidURLLoginTimeout">
+      <jdbcDriver libraryRef="FATDriverLib" internal.nonship.function="This is for internal development only. Never use this in production"/>
+      <properties url="jdbc:fatdriver:memory:jdbcdriver1;create=true;LoginTimeout=10" user="dbuser2" password="{xor}Oz0vKDtt"/>
+      <containerAuthData user="dbuser1" password="{xor}Oz0vKDtu"/>
```
njr-11: Given that this data source is only meant to test loginTimeout, why the complexity with different users?

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSManagedConnectionFactoryImpl.java
```diff
@@ -1426,8 +1427,9 @@ public final int getLoginTimeout() throws SQLException {
             if(!Driver.class.equals(type)) {
                 return ((CommonDataSource) dataSourceOrDriver).getLoginTimeout();
             }
-            //TODO behavior for java.sql.Driver
-            return 0;
+            //When using Driver, loginTimeout is blocked from being set via URL or properties
+            //Therefore we know the loginTimeout being used is from the DriverManager
+            return DriverManager.getLoginTimeout();
```
njr-11: Returning 0 here was actually correct. This means the system is defaulting the value, as is expected when we have not permitted the configuration of it to a non-zero value.  Even though DriverManager.getLoginTimeout value might end up being used as the default, note that the value can change over time and it would be very confusing to the user. If they see a value, for example of 60 here, it implies they configured it to 60 and that the value will remain 60.  The former is untrue and the latter cannot be guaranteed.

dev/com.ibm.ws.jdbc_fat_driver/publish/servers/com.ibm.ws.jdbc.fat.driver/server.xml
alexsm82: This config is used to obtain a datasource impl, not one that is driver based, so it will work even after changing the ignoreWarnFail logic.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
alexsm82: Agreed, that would be better.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -594,5 +594,11 @@ KERBEROS_NOT_SUPPORTED_WARNING.explanation= The Application Server does not supp
...
+
 # 9600-9603 deleted
```
njr-11: This message was placed in the wrong section of the JDBC messages file.  This messages file has different numeric ranges depending on whether messages came from traditional WebSphere code or whether they are from Open Liberty.  Open Liberty messages belong in the range 4000-4999.

dev/com.ibm.ws.jdbc_fat_driver/fat/src/com/ibm/ws/jdbc/fat/driver/JDBCDriverManagerTest.java
alexsm82: No, it won't be because there will still be an injection failure.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -245,6 +245,11 @@ DSRA4004.onconnect.sql=DSRA4004E: Unable to execute onConnect SQL command \"{0}\
 DSRA4004.onconnect.sql.explanation=An error occurred while attempting to execute the specified SQL command.
 DSRA4004.onconnect.sql.useraction=Refer to the exception message to help determine the cause of the problem.
 
+# {0} name of the DataSource
+INVALID_LOGINTIMEOUT=DSRA4005E: The Application Server does not support the loginTimeout property specified in the URL or properties of the {0} DataSource.
```
njr-11: Should follow the precedent of preceding messages and name the key
DSRA4005.invalid.login.timeout=...
I assume this is done to make scanning the source code for message keys simpler.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -719,10 +719,26 @@ public ClassLoader getClassLoaderForLibraryRef() {
...
-    private Driver loadDriver(String url, ClassLoader classloader) throws SQLException {
+    private Driver loadDriver(String url, ClassLoader classloader, Properties props, String dataSourceID) throws SQLException {
         final boolean trace = TraceComponent.isAnyTracingEnabled();
         if (trace && tc.isDebugEnabled())
             Tr.entry(this, tc, "loadDriver", url, classloader);
+        int index = url.toLowerCase().indexOf("logintimeout");
+        if(index != -1) {
+            int length = url.length();
+            index += 13;  //length of logintimeout=
+            if(index+1 < length && !url.substring(index, index+1).equals("0")) {
```
njr-11: Is it worth it to check that the subsequent character is non-numeric to rule out the possibility of loginTimeout=060&...
?

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
alexsm82: Updated, although the only messages in the file that follow that format are the preceding three messages.

alexsm82:
#build

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -245,6 +245,11 @@ DSRA4004.onconnect.sql=DSRA4004E: Unable to execute onConnect SQL command \"{0}\
 DSRA4004.onconnect.sql.explanation=An error occurred while attempting to execute the specified SQL command.
 DSRA4004.onconnect.sql.useraction=Refer to the exception message to help determine the cause of the problem.
 
+# {0} name of the DataSource
+DSRA4005.invalid.logintimeout=DSRA4005E: The Application Server does not support the loginTimeout property specified in the URL or properties of the {0} DataSource.
```
lzdanski: The Application Server does not support the loginTimeout property **that are** specified in the URL or **in the** properties of the {0} DataSource.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -245,6 +245,11 @@ DSRA4004.onconnect.sql=DSRA4004E: Unable to execute onConnect SQL command \"{0}\
...
+DSRA4005.invalid.logintimeout.explanation=The loginTimeout property is not supported when specified for a DataSource that uses the Driver type because the Application Server is unable to call getLoginTimeout on the Driver class.
```
lzdanski: The loginTimeout property is not supported when **it is** specified for a DataSource that uses the Driver type because the Application Server is unable to call getLoginTimeout on the Driver class.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -245,6 +245,11 @@ DSRA4004.onconnect.sql=DSRA4004E: Unable to execute onConnect SQL command \"{0}\
...
+DSRA4005.invalid.logintimeout.useraction=Modify the DataSource configuration to use a DataSource type rather than a Driver type or remove the loginTimeout property.
```
lzdanski: DSRA4005.invalid.logintimeout.useraction=Modify the DataSource configuration to use a DataSource type instead of a Driver type, or remove the loginTimeout property.

alexsm82:
The one test failure was unrelated (JDBC is not enabled in the bucket with the failing test).

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
alexsm82: Hi Laura, after a team discussion we think updating to this is more clear, what are your thoughts:
The loginTimeout property is not supported when it is specified for a DataSource that is backed by a java.sql.Driver implementation because the server is unable to call getLoginTimeout on the java.sql.Driver class.

PR #4483, opened 2018-08-06
Upgrade from ASM 6.1.1 to 6.2.1 for JDK 11 support

aguibert:
This upgrades us to ASM 6.2.1 and will allow us to scan JDK 11 bytecode.  ASM 6.2 introduced this capability, but they had a nasty regression in 6.2.0 so I had to wait for 6.2.1 for the regression to be fixed.


aguibert:
#build

PR #4484, opened 2018-08-06
JDBC DataSource allows for IsolationLevel = TRAN_NONE

KyleAure:
Added new isolation level to Metatype that allows a user to set IsolationLevel = Transaction_None.  This will allow a user to use a JDBC Driver that does not support transactions.

When using an isolation level of Transaction_None no attempt will be made to enlist the connection into a global transaction and autoCommit will reject a setting of false. 

A new "TRANSACTION_NONE_UNSUPPORTED" error message has been added for cases when the isolation level in data source is set to Transaction_None, but the driver does not support this isolation level.

### DIFF
--- a/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
+++ b/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
@@ -65,7 +65,7 @@
    <Option value="TRANSACTION_REPEATABLE_READ"    label="%isoLevel.REPEATABLE_READ.desc"/>
    <Option value="TRANSACTION_SERIALIZABLE"       label="%isoLevel.SERIALIZABLE.desc"/>
    <Option value="TRANSACTION_SNAPSHOT"           label="%isoLevel.SNAPSHOT.desc"/>
-   <Option value="TRANSACTION_NONE"     		  label="%isoLevel.NONE.desc" />
+   <Option value="TRANSACTION_NONE"               label="%isoLevel.NONE.desc" />
   </AD>
   <AD id="statementCacheSize"                     name="%stmtCacheSz"  description="%stmtCacheSz.desc"  required="false" type="Integer" default="10" min="0"/>
   <AD id="transactional"                          name="%transactnl"   description="%transactnl.desc"   required="false" type="Boolean" default="true"/>
--- a/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
+++ b/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
@@ -65,7 +65,7 @@
    <Option value="TRANSACTION_REPEATABLE_READ"    label="%isoLevel.REPEATABLE_READ.desc"/>
    <Option value="TRANSACTION_SERIALIZABLE"       label="%isoLevel.SERIALIZABLE.desc"/>
    <Option value="TRANSACTION_SNAPSHOT"           label="%isoLevel.SNAPSHOT.desc"/>
-   <Option value="TRANSACTION_DRIVER_DEFAULT"     label="%isoLevel.DRIVER_DEFAULT.desc" />
+   <Option value="TRANSACTION_NONE"     		  label="%isoLevel.NONE.desc" />
   </AD>
   <AD id="statementCacheSize"                     name="%stmtCacheSz"  description="%stmtCacheSz.desc"  required="false" type="Integer" default="10" min="0"/>
   <AD id="transactional"                          name="%transactnl"   description="%transactnl.desc"   required="false" type="Boolean" default="true"/>
--- a/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
+++ b/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
@@ -65,6 +65,7 @@
    <Option value="TRANSACTION_REPEATABLE_READ"    label="%isoLevel.REPEATABLE_READ.desc"/>
    <Option value="TRANSACTION_SERIALIZABLE"       label="%isoLevel.SERIALIZABLE.desc"/>
    <Option value="TRANSACTION_SNAPSHOT"           label="%isoLevel.SNAPSHOT.desc"/>
+   <Option value="NON-TRANSACTIONAL"              lable="%isoLevel.NON-TRANSACTIONAL" />
   </AD>
   <AD id="statementCacheSize"                     name="%stmtCacheSz"  description="%stmtCacheSz.desc"  required="false" type="Integer" default="10" min="0"/>
   <AD id="transactional"                          name="%transactnl"   description="%transactnl.desc"   required="false" type="Boolean" default="true"/>
--- a/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
+++ b/dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
@@ -65,7 +65,7 @@
    <Option value="TRANSACTION_REPEATABLE_READ"    label="%isoLevel.REPEATABLE_READ.desc"/>
    <Option value="TRANSACTION_SERIALIZABLE"       label="%isoLevel.SERIALIZABLE.desc"/>
    <Option value="TRANSACTION_SNAPSHOT"           label="%isoLevel.SNAPSHOT.desc"/>
-   <Option value="NON-TRANSACTIONAL"              lable="%isoLevel.NON-TRANSACTIONAL" />
+   <Option value="TRANSACTION_DRIVER_DEFAULT"     label="%isoLevel.DRIVER_DEFAULT.desc" />
   </AD>
   <AD id="statementCacheSize"                     name="%stmtCacheSz"  description="%stmtCacheSz.desc"  required="false" type="Integer" default="10" min="0"/>
   <AD id="transactional"                          name="%transactnl"   description="%transactnl.desc"   required="false" type="Boolean" default="true"/>
--- a/dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
+++ b/dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
@@ -250,6 +250,11 @@ DSRA4005.invalid.logintimeout=DSRA4005E: The server does not support the loginTi
 DSRA4005.invalid.logintimeout.explanation=The loginTimeout property is not supported when it is specified for a DataSource that is backed by a java.sql.Driver implementation because the server is unable to call getLoginTimeout on the java.sql.Driver class.
 DSRA4005.invalid.logintimeout.useraction=Modify the DataSource configuration to use a DataSource implementation instead of a java.sql.Driver implementation, or remove the loginTimeout property.
 
+#4008			TRANSACTION_NONE_UNSUPPORTED
+TRANSACTION_NONE_UNSUPPORTED=DSRA4008: The JDBC driver does not support an isolation level of TRANSACTION_NONE.  The isolation level was set to TRANSACTION_NONE because of the data source configuration.
+TRANSACTION_NONE_UNSUPPORTED.explanation= The JDBC driver does not support using an isolation level of TRANSACTION_NONE.  An attempt to set the isolation level to TRANSACTION_NONE was the result of the data source configuration.
+TRANSACTION_NONE_UNSUPPORTED.useraction=  Either update the data source configuration to use an isolation level that is supported by the driver, or use a driver that supports an isolation level of TRANSACTION_NONE. 
+
 # ----------------------------------------------------------------------------
 # 
 # Using 7000-7599 for DataStore Helper exception messages and warnings
@@ -602,8 +607,3 @@ KERBEROS_NOT_SUPPORTED_WARNING.useraction= Disable Kerberos, or use a DB2 backen
 
 # 9600-9603 deleted 
 # 9900 deleted
-
-#9549			TRANSACTION_NONE_UNSUPPORTED
-TRANSACTION_NONE_UNSUPPORTED=DSRA9549W: The JDBC driver does not support an isolation level of TRANSACTION_NONE.  The isolation level was set to TRANSACTION_NONE because of the data source configuration.
-TRANSACTION_NONE_UNSUPPORTED.explanation= The JDBC driver does not support using an isolation level of TRANSACTION_NONE.  An attempt to set the isolation level to TRANSACTION_NONE was the result of the data source configuration.
-TRANSACTION_NONE_UNSUPPORTED.useraction=  Either update the data source configuration to use an isolation level that is supported by the driver, or use a driver that supports an isolation level of TRANSACTION_NONE. 
--- a/dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
+++ b/dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
@@ -254,17 +254,17 @@ DSRA4008.tran.none.unsupported=DSRA4008E: The JDBC driver does not support an is
 DSRA4008.tran.none.unsupported.explanation=The TRANSACTION_NONE isolation level that is configured on the data source is not supported by the JDBC driver.
 DSRA4008.tran.none.unsupported.useraction=Either update the data source configuration to use an isolation level that is supported by the driver, or use a driver that supports an isolation level of TRANSACTION_NONE. 
 
-DSRA4009.tran.none.transactional.unsupported=DSA4009E: The data source {0} cannot be configured with an isolation level of TRANSACTION_NONE and transactional=\"true\".
+DSRA4009.tran.none.transactional.unsupported=DSRA4009E: The data source {0} cannot be configured with an isolation level of TRANSACTION_NONE and transactional=true.
 DSRA4009.tran.none.transactional.unsupported.explanation=A JDBC driver that is configured with an isolation level of TRANSACTION_NONE will not support transactions.
-DSRA4009.tran.none.transactional.unsupported.useraction=Set transactional=\"false\" on the data source.
+DSRA4009.tran.none.transactional.unsupported.useraction=Set transactional=false on the data source.
 
 DSRA4010.tran.none.autocommit.required=DSRA4010E: Auto commit is required when using an isolation level of TRANSACTION_NONE, which is configured on data source {0}.
 DSRA4010.tran.none.autocommit.required.explanation=The JDBC driver is configured with an isolation level of TRANSACTION_NONE, and requires the use of auto commit.
 DSRA4010.tran.none.autocommit.required.useraction=Keep auto commit set to true.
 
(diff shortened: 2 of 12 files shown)


KyleAure:
#build

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -245,6 +245,11 @@ DSRA4004.onconnect.sql=DSRA4004E: Unable to execute onConnect SQL command \"{0}\
 DSRA4004.onconnect.sql.explanation=An error occurred while attempting to execute the specified SQL command.
 DSRA4004.onconnect.sql.useraction=Refer to the exception message to help determine the cause of the problem.
 
+#4008			TRANSACTION_NONE_UNSUPPORTED
+TRANSACTION_NONE_UNSUPPORTED=DSRA4008: The JDBC driver does not support an isolation level of TRANSACTION_NONE.  The isolation level was set to TRANSACTION_NONE because of the data source configuration.
```
alexsm82: Update the message key to match that of the preceding keys (DSRA4008.description) as that makes it easier to search for the keys in files.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -245,6 +245,11 @@ DSRA4004.onconnect.sql=DSRA4004E: Unable to execute onConnect SQL command \"{0}\
...
+TRANSACTION_NONE_UNSUPPORTED.explanation= The JDBC driver does not support using an isolation level of TRANSACTION_NONE.  An attempt to set the isolation level to TRANSACTION_NONE was the result of the data source configuration.
```
lzdanski: TRANSACTION_NONE_UNSUPPORTED.explanation= The JDBC driver does not support using an isolation level of TRANSACTION_NONE. **The data source configuration caused an attempt to set the isolation level to TRANSACTION_NONE.**

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
lzdanski: TRANSACTION_NONE_UNSUPPORTED=DSRA4008: The JDBC driver does not support an isolation level of TRANSACTION_NONE.  **The data source configuration caused the isolation level to be set to TRANSACTION_NONE.**

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
```diff
@@ -1160,15 +1191,19 @@ public final boolean isGlobalTransactionActive() {
...
      * 
-     * @return true if transaction enlistment is enabled, false otherwise.
+     * @return true if transaction enlistment is enabled and supported, false otherwise.
      */
     public final boolean isTransactional() {
         // Take a snapshot of the value with first use (or reuse from pool) of the managed connection.
         // This value will be cleared when the managed connection is returned to the pool.
         if (transactional == null) {
             transactional = mcf.dsConfig.get().transactional;
+            if (useDriverIsoLvl && transactional)
```
alexsm82: I have some concerns here about switching the transactional value.  If it is only true because that is the default that I think it's okay to switch the value, but if they have configured TRANSACTON_NONE and transactional=true in their config it is probably better to reject that configuration rather than switch the value under the covers.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
```diff
@@ -485,8 +491,33 @@ else if (!DataSource.class.equals(mcf.type) && !Driver.class.equals(mcf.type) &&
...
+                        if(isTraceOn && tc.isDebugEnabled()) {
+                            Tr.debug(this, tc, "DataSource configured with an isolation level of 'NONE (0)' and the driver's isolation level has been set to 'NONE (0)'.");
+                        }
+                        useDriverIsoLvl = true;
+                    } else { //attempt failed
+                        throw new SQLException(AdapterUtil.getNLSMessage("TRANSACTION_NONE_UNSUPPORTED"));
+                    }
+                }
+            } catch (SQLException sqle) {
+                Tr.error(tc, "TRANSACTION_NONE_UNSUPPORTED");
```
alexsm82: You don't need to log an error here since you are throwing the exception.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/jdbc/WSJdbcDataSource.java
```diff
@@ -104,7 +105,7 @@ public final Connection getConnection() throws SQLException {
         final boolean isTraceOn = TraceComponent.isAnyTracingEnabled(); 
 
         if (isTraceOn && tc.isDebugEnabled()) 
-            Tr.debug(this, tc, "getConnection");
+            Tr.debug(this, tc, "getConnection and determine isolation level");
```
alexsm82: This addition doesn't seem necessary.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
```diff
@@ -3821,6 +3855,10 @@ public final void setAutoCommit(boolean value) throws SQLException
     {
         if (value != currentAutoCommit || helper.alwaysSetAutoCommit()) 
         {
+            if(useDriverIsoLvl && (value == false) )
+                throw new SQLException("DataSource configured with an isolation level of 'NONE (0)'. "
```
alexsm82: This message needs to be translated.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
```diff
@@ -356,6 +356,12 @@ public boolean isMCStale() {
      * flag to keep track of isolation level switching on connection support or not
      **/
     boolean supportIsolvlSwitching = false; 
+    
+    /**
+     * flag if dsConfig is set to isolation level 'NONE (0)'
+     * Used to ensure we use the drivers isolation level
+     */
+    boolean useDriverIsoLvl = false;
```
alexsm82: Rename variable so it is more obvious that it is being used for transaction_none (I imagine the name is leftover from when we were using the JDBC driver default).

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
alexsm82: Also the explanation and message shouldn't be quite so similar (and could use a little more detail).  Maybe something more along the lines of:
message: DSRA4008: The datasource {0} was configured with an isolation level of TRANSACTION_NONE, but that isolation level is not supported by the JDBC driver.
explanation: The datasource was configured with with an isolation level of TRANSACTION_NONE, but that is not the JDBC driver's default isolation level and the Application Server was unable to set the driver's isolation level to TRANSACTION_NONE.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
alexsm82: It would probably be appropriate to use ignorewarnfail if they have TRANSACTION_NONE configured with transactional=true

dev/com.ibm.ws.jdbc/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -65,6 +65,7 @@
    <Option value="TRANSACTION_REPEATABLE_READ"    label="%isoLevel.REPEATABLE_READ.desc"/>
    <Option value="TRANSACTION_SERIALIZABLE"       label="%isoLevel.SERIALIZABLE.desc"/>
    <Option value="TRANSACTION_SNAPSHOT"           label="%isoLevel.SNAPSHOT.desc"/>
+   <Option value="TRANSACTION_NONE"     		  label="%isoLevel.NONE.desc" />
```
njr-11: usage of tabs causes this to not line up with the other labels in some viewers

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
```diff
@@ -245,6 +245,11 @@ DSRA4004.onconnect.sql=DSRA4004E: Unable to execute onConnect SQL command \"{0}\
...
+TRANSACTION_NONE_UNSUPPORTED.useraction=  Either update the data source configuration to use an isolation level that is supported by the driver, or use a driver that supports an isolation level of TRANSACTION_NONE.
```
njr-11: remove extra white space at beginning of message

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
njr-11: Need to remove extra white space at beginning of message.  Otherwise, this seem a bit wordy and awkward. Can we just say,
"The TRANSACTION_NONE isolation level that is configured on the data source is not supported by the JDBC driver."
?

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
njr-11: The message prefix needs an indicator of informational/error/warning/audit.  I'm not sure which you intend it to be because I haven't reviewed the rest of the code yet, but assuming it is error, then it needs to be:
DSRA4008E: ...

njr-11: I don't like the language stating "caused the isolation level to be set to TRANSACTION_NONE".  The isolation level didn't get set to TRANSACTION_NONE - the attempt was unsuccessful, so this is misleading.  Recommend the following:
"The JDBC driver does not support an isolation level of TRANSACTION_NONE, which is configured on data source {0}."
where {0} is a substitution parameter for the data source identifier that the user ought to be looking at to fix the problem

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
njr-11: When you do write a translatable message, leave out the part about the connection not being enlisted in a global transaction.  What the user needs to know is that setAutoCommit(false) is disallowed with TRANSACTION_NONE regardless of other circumstances.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
```diff
@@ -485,8 +491,33 @@ else if (!DataSource.class.equals(mcf.type) && !Driver.class.equals(mcf.type) &&
...
+        // Check to make sure a data source configured with 'NONE (0)' is compatible with the driver being used. 
+        if(config.isolationLevel == Connection.TRANSACTION_NONE) {
+            try {
+                if(conn.getTransactionIsolation() == Connection.TRANSACTION_NONE) {
+                    if(isTraceOn && tc.isDebugEnabled()) {
+                        Tr.debug(this, tc, "DataSource configured with an isolation level of 'NONE (0)' and the driver's isolation level is already 'NONE (0)'."); 
+                    }
+                    useDriverIsoLvl = true; //Expected behavior
+                } else {
+                    //attempt to set isolation level of the driver to 'NONE (0)'
```
njr-11: java.sql.Connection.setTransactionIsolationLevel clearly states that TRANSACTION_NONE cannot be supplied as a parameter.  Why are we going out of our way to attempt what the JDBC API itself says cannot be done?  Is there a particular JDBC driver we are trying to accommodate here?  If not, we should remove this path that deviates from the JDBC spec unless/until there is shown to be a pressing need for it.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
njr-11: Also, I would like to see this check moved from this location to the point where we accept the configuration, in which case we would simply compare the values of the transactional and isolationLevel config properties.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
njr-11: I don't think we need this variable at all.  Just check dsConfig.isolationLevel == TRANSACTION_NONE in the one or two places where this information is needed.  It will be much more straightforward.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
lzdanski: "The JDBC driver does not support an isolation level of TRANSACTION_NONE, which is configured on data source {0}." Looks good to me.

lzdanski: @njr-11, This review corresponds to the line that it's commented on, line 249. We don't review the message prefixes.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
lzdanski: "The TRANSACTION_NONE isolation level that is configured on the data source is not supported by the JDBC driver." Looks good to me.

dev/com.ibm.ws.jdbc/resources/com/ibm/ws/rsadapter/resources/IBMDataStoreAdapterNLS.nlsprops
njr-11: @lzdanski  - I don't understand your most recent comment above that mentions me.  I certainly wouldn't argue with what you stated, and I don't believe I wrote anything that contradicts it, so I'm wondering if you meant to place that comment elsewhere or direct it towards a different person?

lzdanski: @njr-11 I think that there are multiple threads of comments about the same line. It's not clear what's a response to the message review comments vs. the code. Sorry about that!

KyleAure:
#build

dev/com.ibm.ws.jdbc_fat_derby/test-applications/jdbcapp/src/jdbc/fat/derby/web/JDBCDerbyServlet.java
```diff
@@ -10,37 +10,244 @@
...
+    @Resource(name = "jdbc/dsfat3ref", lookup = "jdbc/dsfat3")
+    DataSource ds3; //DSConfig (TRAN_NONE) + Res-ref (TRAN_NONE) + TRAN_NONE JDBC Driver
+
+    @Resource
+    private UserTransaction tran;
+
+    /**
+     * Create the default table used by the tests.
+     */
+    private void createTable(DataSource ds) throws SQLException {
```
alexsm82: In the future switch to built in createTable method if possible

dev/com.ibm.ws.jdbc_fat_derby/test-applications/jdbcapp/src/jdbc/fat/derby/web/JDBCDerbyServlet.java
```diff
@@ -10,37 +10,244 @@
...
+    @Test
+    @ExpectedFFDC({ "java.sql.SQLException" })
+    public void testRejectIsolationChange() throws Throwable {
+        Connection con = null;
+        try {
+            con = ds1.getConnection();
+            con.setTransactionIsolation(Connection.TRANSACTION_READ_COMMITTED);
+            fail("Exception should have been thrown when swiching isolation levels on a driver configured with TRANSACTION_NONE.");
+        } catch (SQLException sql) {
+            assertTrue("Exception message should have contained", sql.getMessage().contains("DSRA4011E"));
```
alexsm82: Although we have the FFDC, on a subsequent PR update this to rethrow so that we have the stack more easily for debugging.

dev/com.ibm.ws.jdbc_fat_derby/test-applications/jdbcapp/src/jdbc/fat/derby/web/JDBCDerbyServlet.java
```diff
@@ -10,37 +10,244 @@
...
+            tran.begin();
+            /*
+             * This should throw an exception because we are not enlisted in a global
+             * transaction (see previous testTransationEnlistment) so autoCommit(false)
+             * would prevent changes from being committed overall.
+             *
+             * If this behavior is undesired in the future consider a requirement to set
+             * CommitOrRollbackOnCleanup on dataSource.
+             */
+            con.setAutoCommit(false);
```
alexsm82: You don't need all the code after this since you are expecting this call to result in an exception.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/WSRdbManagedConnectionImpl.java
```diff
@@ -356,6 +356,12 @@ public boolean isMCStale() {
      * flag to keep track of isolation level switching on connection support or not
      **/
     boolean supportIsolvlSwitching = false; 
+    
+    /**
+     * Flag if datasource is configured with isolation level = TRANSACTION_NONE
+     * Used to ensure JDBC driver supports using a datasource with this isolation level 
+     */
+    boolean isDSTranNone = false;
```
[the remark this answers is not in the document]
alexsm82: This can be addressed in a future PR.

lzdanski:
@KyleAure looks good to me! I'll add the reviewed label. Thanks!

PR #4511, opened 2018-08-07
null check before adding to ringBuffer

### LINKED ISSUE #4510: Add null check when flushing earlyBuffer in BufferManagerImpl
Add null check when flushing earlyBuffer in BufferManagerImpl to an asynchronous handler.
If a hard-referenced early queue is used then null objects will be sent through.

Channyboy:
fixes #4510 
#build


Channyboy:
The failure experienced here does not relate to the issue/error reported in the build. There already exists an item already tracking the issue.

PR #4516, opened 2018-08-08
Add provided scope to BOM

rampras3:


rampras3:
BOM includes scope(provided)for esa 

screen shot 2018-08-08 at 10 22 59 am

PR #4532, opened 2018-08-08
tests for driver not accessible to application classloader

njr-11:
Add test cases to cover scenarios where the JDBC driver library is only available to the dataSource/jdbcDriver and not on the application class loader, so that we can be sure that the added code to detect JDBC driver and data source classes based on it does not require the classes to be available on the thread context class loader.


njr-11:
#build

njr-11:
#build

njr-11:
I've forgotten to add the JDBC driver that is used by this test. Adding a commit for that,

njr-11:
#build

Issue #4550, opened 2018-08-09
Injection race condition in JAX-RS during startup


andymc12:
There is a race condition in Liberty with both `jaxrs-2.0` and `jaxrs-2.1` where a CDI proxy of an `@Context` object is not properly injected into JAX-RS providers in the app.  This can occur when multiple threads are accessing a JAX-RS resource (which flows through the provider(s)) while the app is still starting.

One occurrence might look like this:
```
@Provider
@ApplicationScoped
public class Filter1 implements ContainerRequestFilter, ContainerResponseFilter {

    @Context
    ResourceInfo resourceInfo;
    
    @Override
    public void filter(ContainerRequestContext reqCtx) throws IOException {
        Class<?> resourceClass = resourceInfo.getResourceClass();
        System.out.println("Filter1(request) - resourceClass=" + resourceClass);
        if (resourceClass == null) {
            reqCtx.abortWith(Response.serverError().build());
        }
    }
...
```

In this case, the `resourceClass` should always be non-null, but due to the bug in this issue, it is possible that the injected instance of `ResourceInfo` would return a null resource class.

PR #4551, opened 2018-08-09
Inject at startup issue

andymc12:
This resolved issue #4550 

This will synchronize the injection processing code at JAX-RS app start time so that the expected CDI proxy will be injected into the provider instance - avoiding a race condition.


dev/com.ibm.ws.jaxrs.2.0.common/src/com/ibm/ws/jaxrs20/injection/InjectionRuntimeContextHelper.java
```diff
@@ -140,43 +140,44 @@ public static Object getBeanCustomizerContext(JaxRsFactoryBeanCustomizer beanCus
             return;
         }
```
jim-krueger: Maybe add a comment here explaining this synchronization?  And shouldn't you maybe do the synchronized(resource) instead at the top of the method?  (ie. couldn't the "isInit()" also be subject to the race condition?)

andymc12: I think we should limit the sync block to the minimum necessary - otherwise we slow down performance.  The sync ensures that we only "initialize" the ProviderInfo class once.  Once it is initialized, we shouldn't need to reset the provider - keep in mind that the provider is already set when the ProviderInfo was created - so we set it here only when there is a CDI (or EJB) proxy to the provider - so that those technologies can manage the lifecycle.

andymc12:
Added test cases - new #build

andymc12:
#build

andymc12:
#build

andymc12:
#build - minor test case tweak.

andymc12:
Single failure is unrelated to these changes.

PR #4558, opened 2018-08-09
mpJwt issuer is no longer required in server.xml

c00crane:
In prep for mpJwt 1.1 support, make the issuer in the mp_jwt config optional.

#build


c00crane:
Failures in the personal build are not related to the metadata change I'm making.

PR #4559, opened 2018-08-09
test coverage for data source backed by java.sql.Driver in loadfromapp bucket

njr-11:
The loadfromapp test bucket is a convenient place to gain some test coverage for data sources backed by (or inferred from) java.sql.Driver because it has a couple of mock JDBC drivers, of which one can easily be extended to also provide a Driver implementation.  In addition to enhancing coverage of our feature, this will also demonstrate its compatibility with the loadfromapp feature for which that bucket was originally written.


njr-11:
#build

PR #4561, opened 2018-08-09
Add new test to ensure response thread is from the specified ExecSvc

andymc12:
Adding a new test that verifies that the response from an async JAX-RS client invocation occurs on a thread from the specified `ExecutorService` (via the `ClientBuilder`'s `executorService(...)` method.

Just a new test, not a release bug.


andymc12:
#build

Issue #4444, opened 2018-08-02
Port async EJB tests to OL - error/warning tests


brideck:
Creating OL tests in com.ibm.ws.ejbcontainer.session.async_fat:
AsyncErrTest
AsyncWarnTest

PR #4580, opened 2018-08-10
Issue #4444: Port async EJB tests to OL - error/warning tests

brideck:
AsyncErrTest
AsyncWarnTest


brideck:
fixes #4444

brideck:
Olen/Tracy, this is the streamlined version of the test port I've been working on.  I'm maintaining the version with extra activate/deactivate cycles in a local branch to continue that work.

olendvcook:
@brideck, In the middle of reviewing right now and I noticed something, and these seem like test infrastructure bugs:

First thing is that in the build, we are not appending what mode it ran on at the end of the tests such as _EE7_FEATURES, i looked at the lastest LAOS builds to make sure they actually show up there for other tests and they do, I can see the tests running multiple times so repeat tests is working, I guess whatever I did to append that only works with `@Servlet` which I noticed you aren't using.

Second is that all the tests are being ran in the non full bucket even though the classes are marked as FULL tests.

brideck:
#build

brideck:
#build

PR #4581, opened 2018-08-10
test class-name optional in deployment descriptor

njr-11:
Write a test that shows class-name is optional in the deployment descriptor for application defined data source, in that when it is absent, the same rules are applied as when omitted from @DataSourceDefinition.


njr-11:
#build

njr-11:
#build

njr-11:
build has 4 test failures which are pre-existing issues in unrelated tests

PR #4582, opened 2018-08-10
remove nonship guards for datasource backed by driver feature

njr-11:
Code path for the feature to have data sources backed by driver
is currently guarded with a hidden nonship property.
This pull removes the hidden property so that the function can be usable
and updates our test cases accordingly such that they no longer specify the hidden property.


njr-11:
#build

njr-11:
#build

PR #4597, opened 2018-08-13
address TODO comments around message parameters

njr-11:
There are a couple of TODO comments left in the code which this pull addresses:

1. Ensure that the dataSource identifier, not the jdbcDriver, is supplied to the message.  The jdbcDriver config element is not precise enough because multiple dataSource configurations can share a single jdbcDriver config, and those dataSource configurations could have different values for type.

2. Supplying the lists of searched-in libraries/packages to the proper message when DataSourceDefinition fails to locate/infer a data source or Driver implementation.


njr-11:
#build

njr-11:
Build has 1 test failure which is a pre-existing issue in unrelated tests/componentry.

Issue #4529, opened 2018-08-08
JPA ORM Analyzer: Scan ORM XML documents for types that should not be trimmed.


jgrassel:
The class com.ibm.ws.jpa.diagnostics.puscanner.PersistenceUnitScanner is responsible for scanning the ORM XML files that are part of the same persistence unit for persistent-capable and persistent-aware types so that those types are retained from the Class Scanner result set.

There remains a number of ORM XML elements that need to be analyzed for collecting these type names (many of which are identified as TODOs in the PersistenceUnitScanner class).

These include:
  - Embeddable, Entity, and MappedSuperclass Attributes
  - Converters
  - Named Query Constructor Results
  - Named Entity Graphs
  - Named Native Query Result Class
  - Named Stored Procedure Result Class and Parameters
  - Named Subgraph complex type
  - SQL Result Mapping

jgrassel:
Cancelling this.  After the updates to how class fingerprints are collected, the size of the output was dramatically reduced, so this isn't as much of an issue as it was originally.

PR #4607, opened 2018-08-13
OpenLiberty 4529: Scan ORM XML documents for types that should not be…

jgrassel:
Delivery pull request for #4529


jgrassel:
#build

Issue #4504, opened 2018-08-07
Add new test cases for mpMetrics's authentication config


navidsh:
Add FAT tests for the MicroProfile Metrics.

For more info about the actual changes made, see https://github.com/OpenLiberty/open-liberty/issues/4406

Scenarios to be tested:
- Setting `authentication` to `true` or `false`.
- Removing `<mpMetrics ... />` and ensuring default is the secure endpoint.
- Switching between `mpMetrics-1.0` and `mpMetrics-1.1`

I would suggest you add a configuration element to the FAT framework to make your life easier when switching `mpMetrics` configuration. See [com.ibm.websphere.simplicity.config.OpenAPIElement](https://github.ibm.com/was-liberty/WS-CD-Open/blob/integration/dev/fattest.simplicity/src/com/ibm/websphere/simplicity/config/OpenAPIElement.java) for an example of how you could do it.

tkburroughs:
Is this issue ready to be closed?

navidsh:
Delivered. Closing.

PR #4615, opened 2018-08-14
For issue #4504: mpMetrics FAT test cases for authetication config

a-saf:
This PR contains additional test cases for authentication config in com.ibm.ws.microprofile.metrics_fat for mpMetrics-1.0 and mpMetrics-1.1 features


a-saf:
This PR is for issue #4504

dev/fattest.simplicity/src/com/ibm/websphere/simplicity/config/mpMetricsElement.java
```diff
@@ -0,0 +1,76 @@
...
+            for (WebModuleDocElement webModuleDoc : webModuleDocs) {
+                sb.append(webModuleDoc.toString()).append(", ");
+            }
+        }
+        sb.append("]]");
+        return sb.toString();
+    }
+
+    @XmlType(name = "mpMetricsWebModuleDoc")
+    public static class WebModuleDocElement extends ConfigElement {
```
navidsh: I don't think we need this class.

navidsh: Since `<mpMetrics ... />` won't have a `webModuleDoc` element.

dev/com.ibm.ws.microprofile.metrics_fat/publish/servers/MetricsAuthenticationServer/server.env
```diff
@@ -0,0 +1 @@
+MP_METRICS_TAGS=tier=integration
```
navidsh: This is probably not needed in this test.

a-saf:
@arturdzm @navidsh With regards to Artur's suggestion to add a setter for MetricsElement in ServerConfig - I'm not sure if this is necessary; The only field that element has at this point is `authentication`, and that can already be set to `null`, and is set to `null` in one of the tests, which results in an empty `<mpMetrics />` in server.xml and works as intended. So, what do you think - should the setter still be added so that the whole element can be set to `null`?

PR #4627, opened 2018-08-14
Install Kernel exception fix for matching Feature shortnames

### LINKED ISSUE #4521: Install kernel does not throw exception if already installed features are specified again with a different capitalization
When InstallKernelMap.singleFileResolve() is called with FEATURES_TO_RESOLVE that has feature shortnames with different capitalizations compared to the actual installed features (e.g. FEATURES_TO_RESOLVE contains "localconnector-1.0" or "LOCALCONNECTOR-1.0" when the actual installed feature shortname is "localConnector-1.0"), it currently does not match them, so tries to call the resolver with the features.

Instead, it should be case insensitive in order to detect that the features are already installed and throw the "ASSETS_ALREADY_INSTALLED" exception.

Specifically, I am referring to this code in InstallKernelMap:
```java
if (featureToInstall.contains(feature.getIbmShortName()) || featureToInstall.contains(feature.getFeatureName())) {
```

To reproduce this issue, use this pom https://github.com/ericglau/ci.maven-355/blob/master/simplepom/pom.xml and run `mvn package` twice.  The second time fails with NPE because the "ASSETS_ALREADY_INSTALLED" exception is not thrown.

AnshulPatel1:
Fixes #4521

### DIFF
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
@@ -470,6 +470,15 @@ public class InstallKernelMap implements Map {
         }
     }
 
+    public boolean containsIgnoreCase(Collection<String> featureToInstall, String existingFeature) {
+        for (String current : featureToInstall) {
+            if (current.equalsIgnoreCase(existingFeature)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     @SuppressWarnings("unchecked")
     public Collection<String> singleFileResolve() {
         data.put(ACTION_INSTALL, null);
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -470,6 +470,15 @@ private void resolve(Object installObject) {
         }
     }
 
+    public boolean compareIfContains(Collection<String> item1, String item2) {
```
ericglau: - Call the method something like containsIgnoreCase, since current name does not tell you much.
- Avoid parameter names "item1" and "item2", since the parameter types are not the same.

AnshulPatel1:
#build

AnshulPatel1:
#build

AnshulPatel1:
#build

AnshulPatel1:
#build

PR #4637, opened 2018-08-15
Create default mp jwt config

arunavemulapalli:
First update to runtime to create default mpjwt configuration


lzdanski:
@arunavemulapalli Libby bot flagged a file as requiring ID review, but I don't see it in the diff? Is ID review required?

arunavemulapalli:
I think this is already done.

PR #4638, opened 2018-08-15
Run Fault Tolerance tests against 1.1

Azquelt:
Change all fault tolerance FAT tests to run against 1.1, with a few also
run against 1.0, rather than the reverse.


Azquelt:
#build

Azquelt:
In the repeat build, we got no failures from 50 iterations.

For now, I'll put down the initial failure to an infrastructure glitch but will leave the trace enabled so that if it occurs again we can investigate further.

PR #4639, opened 2018-08-15
Update mpjwt runtime to handle default config

arunavemulapalli:
Handle default config


arunavemulapalli:
Personal build has 4 unrelated failures

PR #4642, opened 2018-08-15
Add switching of virtual host for MP features

NottyCode:
Allow a WAB to have an OL-Virtaul-Host header than indicates the virtual host the app should run on. If it contains a variable then resolve the variable. This only supports a total variable substitution.

### DIFF
--- a/dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WAB.java
+++ b/dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WAB.java
@@ -76,6 +76,8 @@ class WAB implements BundleTrackerCustomizer<WAB> {
     private final Version wabBundleVersion;
     private final String wabContextPath;
     private final WABInstaller installer;
+    private final String rawVirtualHost;
+    private final String resolvedVirtualHost;
 
     private final WABTracker<WAB> trackerForThisWAB;
 
@@ -92,6 +94,8 @@ class WAB implements BundleTrackerCustomizer<WAB> {
         this.wabBundleVersion = wabBundle.getVersion();
         this.wabContextPath = wabContextPath;
         this.installer = installer;
+        this.rawVirtualHost = wabBundle.getHeaders().get("OL-VirtualHost");
+        this.resolvedVirtualHost = resolveVirtualHost();
         trackerForThisWAB = installer.getTracker(this);
     }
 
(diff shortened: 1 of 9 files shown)


NottyCode:
#build this please Libby

dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WABInstaller.java
```diff
@@ -459,6 +466,23 @@ public boolean getUseJandex() {
                         DeployedModuleInfo deployedMod = deployedApp.getDeployedModule(moduleInfo);
                         wab.setDeployedModuleInfo(deployedMod);
 
+                        WebAppConfiguration appConfig = (WebAppConfiguration)((WebModuleMetaData)mmd).getConfiguration();
```
tjwatson: This doesn't handle when configuration of the variable changes.  But I think you know that.  Do we intend to handle that in the future?

dev/com.ibm.ws.microprofile.metrics.private/bnd.bnd
```diff
@@ -23,6 +23,7 @@ IBM-Authorization-Roles: com.ibm.ws.management
 IBM-Web-Extension-Processing-Disabled: true
 
 Web-ContextPath: @privateMetricsURL
+OL-Virtual-Host: ${admin.virtual.host}
```
tjwatson: What does the server.xml config look like to set this?  Are we intending for this to be configurable by the user?  Or are we going to have some default value set?  If we have some default value set for admin.virtual.host where is that set?  Or is the default `null` and when null then we just fallback to the default virtual host?

dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WABInstaller.java
NottyCode: Right, I've just updated the PR to deal with that. The PR from this morning was incomplete, I wanted to get a build running so Navid could run some testing. I should have commented in the PR on that part.

NottyCode:
#build the latest version of the code please Ms Libby

dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WAB.java
```diff
@@ -76,6 +76,8 @@
     private final Version wabBundleVersion;
     private final String wabContextPath;
     private final WABInstaller installer;
+    private final String rawVirtualHost;
+    private String resolvedVirtualHost;
```
tjwatson: I think resolveVirtualHost can be final.  At least I don't see where you modify it outside the constructor.  My understanding is that a new WAB will be create because of a restart on the bundle so it will pick up the new resolved host then.

NottyCode:
I ran a few extra builds:

https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_fuOcMKIcEeiH1sOX9zTQgw&tab=com.ibm.team.build.web.ui.internal.editors.result.test.TestProvider
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_Xxbf8KJ4EeiH1sOX9zTQgw&tab=com.ibm.team.build.web.ui.internal.editors.result.test.TestProvider

they each had failures, but they were different each time and in areas unrelated to the change, or which are known intermittents.

PR #4645, opened 2018-08-15
3917 Clean up tests for TRANSACTION_NONE

KyleAure:
Cleaned up tests for readability and removed tests where an error path had been checked by another test.  No behavior change except for testTNTransationEnlistmentModified, which now looks up the data source and ensures that the correct exception message is thrown.


KyleAure:
#build

KyleAure:
#build

PR #4668, opened 2018-08-16
Get realm name from one of the registries if primary realm not defined

kristip17:
fixes #4645

If a default or primary realm name is not defined, get the realm name from one of the registries to add to the successful operation VMM Audits.


kristip17:
#build

mtamboli:
@kristip17: thanks for opening this issue and fixing it in OL.

kristip17:
PB tests failures are unrelated. Monica was able to test with my PB this morning and confirmed that the realm now appears in the audit record.
FAT test will be in Commercial Liberty space.

PR #4670, opened 2018-08-16
Update MP Config tests to target 1.3

Azquelt:
A selection of tests will still be run against older versions but
everything will now run against 1.3.


Azquelt:
#build

PR #4680, opened 2018-08-16
Handle the required-osgi-ee properly when the Java level changes

tjwatson:
Restructured the solution to do the check reqiured-osgi-ee check
while adding the bundle resources to the BundleList.

This allows the bundle list to only have things we actually
want installed instead of having the Provisioner do the check.
Doing it this way allows the code that determines what bundles
to uninstall to work properly.  I also had to added a persistent
field to the bundle cache file to store the java specification
level.  That is checked each startup to see if it changed.
If it changed from one launch to the next then it forces
a new feature resolution to happen even when no feature
configuration has changed.  This is so we can properly
determine the list of desired bundles to install based on the
current java level being run.

OpenLiberty Pull Requester,

list of bugs fixed.


tjwatson:
Please #build this fix libby

tjwatson:
Libby please #build again after I fixed up the feature manager unit tests to expect calls to `Bundle.adapt(FrameworkWiring.class)`.

tjwatson:
#build

Issue #4519, opened 2018-08-08
MicroProfile Rest Client 1.1


andymc12:
MP Rest Client 1.1 is part of the MP 1.4 and 2.0 release, and provides a few new functional improvements over MP Rest Client 1.0, including (stolen from the [MP Rest Client 1.1 release page](https://github.com/eclipse/microprofile-rest-client/releases)):

* Asynchronous Methods - to use, just update the return type of your Rest Client methods to be CompletionStage.
* `RestClientBuilderListener` - now products can plug in and register providers, update config properties, etc. prior to the creation of new client instances. This enables other MicroProfile technologies, like Open Tracing to automatically add their client request/response filters to MP Rest Clients.
* Supports `java.net.URI` in addition to `java.net.URL`.
* `@RegisterRestClient` is now a bean-defining annotation - no longer need to add `@Dependent` annotation to the client interface.

sabolo:
@andymc12 :  Please could you complete the feature test summary issue under this Epic once all code and test deliveries are complete.  Many thanks.

andymc12:
[WAD-MPRestClient1.1-20180815.pdf](https://github.com/OpenLiberty/open-liberty/files/2293998/WAD-MPRestClient1.1-20180815.pdf)

Attached WAD.

@gkwan-ibm 
1) Yes - the most common issues will likely be with understanding the `CompletionStage` and the intricacies of async methods.
2) Not yet - I am working with the support organization for skills transfer.  During that education session I will be presenting/demoing the functionality and how to service it.  
3) Brian H from SVT is conducting the testing.  He is prepared to approve after his 3 day run completes.
4) WASWEB (L2) and WL3JRS (L3) will support this feature - both teams know that they will be supporting this feature.

andymc12:
Request for new guide is at: https://github.com/OpenLiberty/guides-common/issues/162

bsbyrd1:
My approval is based on the sizing of the ID work, and the expectation that the ID work can be completed before the last ID translation drop.

andymc12:
Latest (and hopefully final) revision of the WAD:
[WAD-MPRestClient1.1-20180821.pdf](https://github.com/OpenLiberty/open-liberty/files/2307890/WAD-MPRestClient1.1-20180821.pdf)

dave-waddling:
[...]

PR #4685, opened 2018-08-16
MP Rest Client CompletionStage completing for subsequent stages

andymc12:
This fixes an issue where the CompletionStage returned by instances of
MicroProfile Rest Client interfaces do not execute subsequent stages.
The method completes normally and the remote invocation occurred
asynchronously, but if the user attempted a `then*` type method on the
returned `CompletionStage`, that stage would not execute because the
internal implementation assumed that the initial stage was not complete.
This fix resolves that issue.

This is a fix for a feature that is not yet released, so it is not a release bug.

This PR also include FAT tests for the above issue, but also for `RestClientBuilderListener` and exception path cases with `CompletionStage` responses.


andymc12:
#build

andymc12:
Adding a new fix, and more tests - new #build

andymc12:
This change is part of MP Rest Client 1.1 Epic Issue #4519

PR #4704, opened 2018-08-17
Ignore some FFDCs in Fault Tolerance Async

### LINKED ISSUE #4662: FFDC created when Async Fault Tolerance method throws exception
QueuedFuture.get is creating FFDCs when catching exceptions thrown by user code.

Catching exceptions at this point is expected, the whole point of the feature is to help with handling exceptions, so it shouldn't be creating FFDCs here.

Azquelt:
Looking at the history, I just forgot to add these FFDCIgnores when
I added some catch blocks.

In each case, the catch would be triggered if the user's method threw an
exception. As Fault Tolerance is concerned with handling failures, this
isn't unexpected so it doesn't make sense to FFDC here.

Fixes #4662


Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:

PR #4710, opened 2018-08-17
Ensure mpRestClient-1.1 works with EE7/MP1.4 and EE8/MP2.0

andymc12:
Additional tests to ensure different feature configurations so that we test that mpRestClient-1.1 passes the same tests that mpRestClient-1.0 passes.  Also run the mpRestClient-1.1 tests (normal FATs & TCK) with Java EE 7 features as well as Java EE 8 features.


andymc12:
Only failure is unrelated to this change.

PR #4733, opened 2018-08-20
Update date insert in OIDC client NLS message

ayoho:
When the `net.oauth` dependency was removed from the OIDC client code, one of the NLS message inserts was changed from using a `net.oauth.jsontoken.SystemClock` object to simply inserting `System.currentTimeMillis()`. Inserting `System.currentTimeMillis()` produces a `long` rather than an actual date format, so this change will update the insert to be formatted as a date.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.openidconnect.client_fat.config,com.ibm.ws.security.openidconnect.server_fat.jaxrs

ayoho:
PB had known failures in OIDC buckets that weren't related to my change.

PR #4736, opened 2018-08-20
Revert getLocation change - should only be non-null if explicitly set

andymc12:
or if the Location header was set on the HTTP response.

This change was made in development of 18.0.0.3, and then reverted - so not a release bug.


andymc12:
#build

andymc12:
#build

PR #4751, opened 2018-08-21
Issue #4750 - only depend on config 1.3

### LINKED ISSUE #4750: Limit MicroProfile Fault Tolerance 1.1 to only work with MP Config 1.3
MP FT 1.1 currently works with all versions of MP Config. It should only work with MP Config 1.3.

tevans78:
#build 
fixes #4750


tevans78:
#build

tevans78:
PB looks good

PR #4768, opened 2018-08-21
Print stack traces for CNFE from parent/follow-on classloaders

andymc12:
Adding trace for occasions where the UnifiedClassLoader's superclass or follow-on classloaders return a ClassNotFoundException.


andymc12:
#build

andymc12:
Test case failures are unrelated to these changes.

After talking with @jaridk, there are some concerns about flooding the trace logs with stack traces for expected exceptions.
Since this diagnostic is for an issue that is not critical to the 3Q release, we will plan to put the change in the stream after the release, resolve the problem, and then remove these changes so that we don't flood logs in a release build.

andymc12:
Merging now - but will plan to revert it after we have the requested diagnostics.

PR #4785, opened 2018-08-22
Improve Java prereq detection in feature list

NottyCode:
The generated feature doc uses some Java prereq detection from the feature list.
There are two issues with the fl.xml detection:

1. It requires you to have all dependent features. This means if you tolerate a
   newer feature, but don't have the base one it fails. Given Open Liberty doesn't
   have the EE 6 features this is a problem.
2. It still reports Java SE 6 as a valid version.

both of these are fixed in this change

#build this Libby

### DIFF
--- a/dev/com.ibm.ws.kernel.feature.cmdline/src/com/ibm/ws/kernel/feature/internal/generator/FeatureList.java
+++ b/dev/com.ibm.ws.kernel.feature.cmdline/src/com/ibm/ws/kernel/feature/internal/generator/FeatureList.java
@@ -73,25 +73,24 @@ public class FeatureList {
 
         if (writingJavaVersion) {
             // When we add Java 9 we need to add it here and update the 1.6 and 1.7 and 1.8 filters. Need to ensure this is captured in docs.
-            addJVM(possibleJavaVersions, "1.6", "1.5", "1.4", "1.3", "1.2", "1.1");
             addJVM(possibleJavaVersions, "1.7", "1.6", "1.5", "1.4", "1.3", "1.2", "1.1");
             addJVM(possibleJavaVersions, "1.8", "1.7", "1.6", "1.5", "1.4", "1.3", "1.2", "1.1");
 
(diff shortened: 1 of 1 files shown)


NottyCode:
I'm not marking this as a release bug since the logic only runs to generate documentation so no user would ever see this problem.

dev/com.ibm.ws.kernel.feature.cmdline/src/com/ibm/ws/kernel/feature/internal/generator/FeatureList.java
```diff
@@ -404,13 +403,12 @@ private String toJavaEE(Map<String, Object> version) {
             }
             for (FeatureResource res : fd.getConstituents(SubsystemContentType.FEATURE_TYPE)) {
                 ProvisioningFeatureDefinition otherFD = mfp.getFeatureDefinitions().get(res.getSymbolicName());
-                // This will NPE when we call getJavaVersion, so fail more helpfully
-                if (otherFD == null)
+                // This will NPE when we call getJavaVersion, so skip
```
aguibert: What is the meaning of "skip" here?  Does it mean that the feature will not be considered?  if we aren't able to detect the java version, I'd like to assume we are just on the newest known version of java, which is what other libs like commons-lang3 do now given the new release cadence of java

NottyCode: what skip means is we ignore the fact that the feature definition isn't found. The reason this was added was because for compatibility with WebSphere Liberty a load of Open Liberty features define EE 6 feature versions as a baseline, but those features never exist in Open Liberty. As a result the feature list generation just failed. By skipping we will ignore it and move onto the next feature. Including looking at the Java EE 7 one.

For a concrete scenario if you have a feature that depends on EE 7, but tolerates EE 8 if you don't have EE 8 installed it'll list Java SE 8 as being the prereq. If you have EE 7 installed you'll get Java SE 7 & 8.

NottyCode:
The test failures in the build are unrelated to this change.

PR #4810, opened 2018-08-23
update oidc clients common code to hash and encode the oidccode cookie

arunavemulapalli:
Before this change - we are taking the hash and encoding.
Receiving side, the order is decode, calculate hash and compare. This may cause some extraneous effects (especially if the cookie is tampered with...).

After this change - We encode first and take the hash
Receiving side - calculate hash and compare to make sure that the data is not tampered with and then decode


arunavemulapalli:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.openidconnect.client_fat,com.ibm.ws.security.openidconnect.client_fat.config,com.ibm.ws.security.openidconnect.client_fat.e2e,com.ibm.ws.security.openidconnect.client_fat.jaxrs

arunavemulapalli:
fixes #4835

PR #4820, opened 2018-08-24
create test bucket for nonship jdbc-4.3 feature

njr-11:
Create a test bucket that can be used for the new jdbc-4.3 feature.
Initially there will only be one test, showing that a connection can be obtained when the jdbc-4.3 feature is enabled, and that the 4.3 version reported by the JDBC driver is also reported by the DatabaseMetaData as the JDBC specification major/minor version.


njr-11:
ID reviewer : please disregard the request to review.  ID team already reviewed the first commit under a different pull, but it has been blocked from delivery due to instability of the stream.

Issue #4806, opened 2018-08-23
Springboot apps start on port 8080 when WebServer factory bean is called from child context


bpaskin-ibm:
When creating a SpringBoot application in Liberty it is possible that the application will not start on the defined port for the httpEndpoint, but on port 8080.  
```
<httpEndpoint id="defaultHttpEndpoint" httpPort="9080" httpsPort="9443" host="*"/>
<springBootApplication location="mySpringBoot.jar" id="myApp" type="spring" name="myApp"/>

```
This is happening when the WebServer factory bean is being called by a child of the ApplicationContext and not the root ApplicationContext.  Liberty should handle both cases.

A small workaround is to define a virtual host for the springBootVirtualHost to override the behavior.
```
<virtualHost id="springBootVirtualHost-8080">
   <hostAlias>*:9443</hostAlias>
</virtualHost>
```

Brian

tjwatson:
The fix for this has been merged for the next release 18.0.0.3 as well as for integration to master.

PR #4826, opened 2018-08-24
Fix issue #4806 - First WebServer should use default host

tjwatson:
Before we only allowed the first WebServer from the root
ApplicationContext to use the dfault host. This is an issue
for applications that use a hierarchy of ApplicationContext

For issue #4806 

Libby please #build


tjwatson:
The build passed all Spring Boot tests.  Only failure was in a child FAT build (that contained no Spring Boot tests), it appeared to fail bringing up some cloud environment to run the tests.

Issue #4666, opened 2018-08-16
Correct getServletPath for default mapping


pnicolucci:
Make this change specific to the Servlet 4.0 feature.

There exists a WebContainer custom property : `<webContainer com.ibm.ws.webcontainer.removetrailingservletpathslash=“true”/>` that defaults to (false). This property when enabled will:

Remove the trailing '/' on the servlet path and add it to the beginning of the path info value. Also, if the servlet mapping url pattern is not a '/*' then the servlet path value will not be an empty string.

By default (value false) the Web Container will have the following behavior:
The path information does not always begin with a '/' because the servlet path is ending with a '/'.  Also, unless the servlet mapping url pattern is a '/*', the getServletPath call should not return an empty string.

Add a property to the metatype and perhaps revisit the name to ensure it makes sense once it is in the metatype. Ensure that this property defaults to true and will ensure that the behavior will be default for Servlet 4.0+.

The Servlet 4.0 Spec section 3.5 states the following:

Servlet Path: The path section that directly corresponds to the mapping which
activated this request. This path starts with a ’/’ character except in the case
where the request is matched with the ‘/*’ or ““ pattern, in which case it is an
empty string.

PathInfo: The part of the request path that is not part of the Context Path or the
Servlet Path. It is either null if there is no extra path, or is a string with a leading
‘/’.

In addition Section 12.2 states the following:

A string containing only the ’/’ character indicates the "default" servlet of the
application. In this case the servlet path is the request URI minus the context path
and the path info is null.

pmd1nh:
https://github.com/OpenLiberty/open-liberty/pull/4686

pnicolucci:
Basically two things addressed here:

(1) append the pathInfo (if not empty/null) with a leading /
(2) If default servlet was used for a request (i.e mapping to / ), the getServletPath() will be the URI - contextRoot

pnicolucci:
Current behavior is not default for servlet-4.0 as expected. This looks to be because the SET_SERVLET_PATH_FOR_DEFAULT_MAPPING is not null in WCCustomProperties and instead is returning an empty string since the metatype default value is an empty String.

We need to check for an empty string otherwise we never set the value to true for the servlet-4.0 feature.

pmd1nh:
https://github.com/OpenLiberty/open-liberty/pull/4921

Add check for empty WCCustom property string.

pnicolucci:
[...]

PR #4828, opened 2018-08-24
Correct getServletPath for a default mapping case

pnicolucci:


pnicolucci:
#4666 delivery to 18.0.0.3

Issue #4835, opened 2018-08-25
Update oidccode cookie


arunavemulapalli:
Update common oidc clients code to encode the cookie first and then calculate the hash value.

ayoho:
Looks like this has been merged.

PR #4836, opened 2018-08-25
Issue #4835 (18.0.0.3) : update oidccode cookie content

arunavemulapalli:
Before this change - we are taking the hash and encoding.
Receiving side, the order is decode, calculate hash and compare. This may cause some extraneous effects (especially if the cookie is tampered with...).

After this change - We encode first and take the hash
Receiving side - calculate hash and compare to make sure that the data is not tampered with and then decode


arunavemulapalli:
PR to deliver the change to stabilization stream
Fixes #4835

arunavemulapalli:
#build

PR #4851, opened 2018-08-27
remove artifactory dependency

brutif:


brutif:
#build

brutif:
other build seems hung so let's 
#build 
again

Issue #1438, opened 2018-01-04
JAAS login module shared library is missing protection domain


teddyjtorres:
There is a timing issue where the JAAS login module class is obtained from the shared library before the Java 2 Security permissions are processed. This causes the shared library to be created without a protection domain that contains the permissions granted in the server.xml, resulting in Java 2 Security access denied errors.

PR #4854, opened 2018-08-27
Update ClassLoadingService to depend on PermissionManager instead of the other way around for issue #1438

teddyjtorres:
For issue #1438


teddyjtorres:
#build

teddyjtorres:
#build

dev/com.ibm.ws.classloading_test/bnd.bnd
```diff
@@ -33,16 +33,20 @@ test.project: true
 	com.ibm.ws.kernel.metatype.helper;version=latest,\
 	com.ibm.ws.kernel.security.thread;version=latest,\
 	com.ibm.ws.logging;version=latest,\
+	org.jmock:jmock-legacy;version=2.5.0, \
```
andymc12: Are these new dependencies required?  I would think that the existing JMock dependencies should be sufficient.

teddyjtorres: There is a new mock to a class, not an interface, that requires this package when setting the mockery object with {
             {
                 setImposteriser(ClassImposteriser.INSTANCE);
             }
         };

teddyjtorres:
#build

teddyjtorres:
FAT failures were caused by the current changes. The javaPlatform features are not installed by default, thus resulting in the following error,
[ERROR   ] CWWKE0702E: Could not resolve module: com.ibm.ws.classloading [47]
  Unresolved requirement: Import-Package: com.ibm.ws.security.java2sec; version="[1.0.0,2.0.0)"

The code will be reworked to not depend directly on the com.ibm.ws.security.java2sec bundle, but rather in the existing java2sec interface in the kernel boot core.

teddyjtorres:
#build

teddyjtorres:
@arkarkala @andymc12 
Please review the newest changes. There were test failures because the javaPlatform-X.0 is not available in all the FAT server configurations. The code was modified to remove the class loader dependency on the java2sec and instead rely on the existing PermissionsCombiner interface in kerne.boot.core. The PermissionsCombiner interface was updated to return the ProtectionDomain map and the ClassLoadingService implementation to have a programmatic cardinality minimum set for the service reference.

teddyjtorres:
Reviewing java2sec FAT failure.

teddyjtorres:
The changes were causing the ClassLoadingServiceImpl to deactivate when the PermissionManager deactivated because of the removal of the javaPermission from the server.xml.

teddyjtorres:
#build

teddyjtorres:
#build

teddyjtorres:
#build

teddyjtorres:
#build

websterc87:
Reverting this pull request see #5084 for more details

dev/com.ibm.ws.classloading_test/bnd.bnd
aguibert: @teddyjtorres we need to stop using jmock-legacy, it will not work reliably with JDK 11

PR #4860, opened 2018-08-27
initial implementation of beginRequest and endRequest

njr-11:
To start out with, invoke the beginRequest operation when associating or getting the first connection handle and invoke the endRequest operation when cleaning up or destroying a managed connection.  Add a few test cases to start experimenting with different scenarios where beginRequest/endRequest should be sent to the JDBC driver.


njr-11:
#build

Issue #4873, opened 2018-08-28
Allow CXF-specific client properties for the JAX-RS 2.X Client APIs


andymc12:
The underlying implementation of the JAX-RS 2.0 and 2.1 Liberty features is Apache CXF, however many custom CXF properties that could be useful for configuring the Client APIs and inaccessible to users.  For example, a user might want to specify whether or not to use "Connection: Keep-Alive", or whether to disable or configure chunking of large entities, etc.  

The list of client configuration properties are available here:
http://cxf.apache.org/docs/client-http-transport-including-ssl-support.html#ClientHTTPTransport(includingSSLsupport)-Theclientelement

PR #4874, opened 2018-08-28
Allow CXF-specific client properties for JAX-RS Client

andymc12:
With this change, users can specify CXF-specific client properties when
using the jaxrs-2.0 or jaxrs-2.1 feature.  These properties should be
specified on the Client or ClientBuilder via the property(...) method,
and the property and value should be specified as a String (i.e. 
`client.property("client.ChunkingThreshold", "8000")` ).

The property name should be prefixed with "client." - i.e. 
"client.AllowChunking", "client.ConnectionTimeout", etc.

This includes runtime changes and test cases.

This will resolve issue #4873, and is an enhancement, not a bug fix.


andymc12:
#build

PR #4892, opened 2018-08-29
ignore multi-release jar error message from Bnd. Bnd has not been tau…

WhiteCat22:
This PR is to suppress a Bnd error caused by multi-release jars as maven dependencies. Bnd has not been taught about multi release jars yet so adding "-fixupmessages.mrjar: Classes found in the wrong directory:*META-INF/versions" to the global bnd file at cnf\build.bnd to suppress.


WhiteCat22:
I tested these changes by adding new projects with multi-release jars as maven dependencies. The gradle 'jar' command failed with the message "Classes found in the wrong directory: {META-INF/versions/9/javax/xml/bind/ModuleUtil.class=javax.xml.bind.ModuleUtil}". After setting this property the 'jar' command succeeds and I get a clean build.

WhiteCat22:
build failed due to single unrelated jms failure.

PR #4895, opened 2018-08-29
test cases for beginRequest and endRequest where LTC is used after suspend

njr-11:
Write test cases for beginRequest and endRequest where a global transaction is suspended and the same connection handle is then used within an LTC to commit and roll back changes independently of the suspended global transaction.  Then resume the global transaction, make additional updates within it and commit.


njr-11:
Build has 2 test failures which match pre-existing issues, both of which are in unrelated test buckets that do not even enable the nonship jdbc-4.3 feature that is added by this pull.

PR #4896, opened 2018-08-30
SseBroadcaster fix

andymc12:
This fix resolves an issue discovered late in the release cycle that was introduced as part of the upgrade to CXF 3.2.6 - CXF changed it's SseBroadcasterImpl code to require that the instance of `SseEventSink` be a CXF-specific `SseEventSinkImpl` - this change breaks our implementation which uses a custom `LibertySseEventSinkImpl`.  The change in this pull request reverts the `SseBroadcasterImpl` to the 3.2.4 version so that we can continue to function normally.  After this release, we should update the Liberty code to use the CXF-specific classes.


andymc12:
#build

PR #4904, opened 2018-08-30
 Avoid TCKs creating a top level failure

Azquelt:
Avoid creating a test failure when mvn returns with a non-zero return code, as long as there are tests that failed.

Usually, a non-zero return code means that a test failed. In this case, we don't want to create an additional test failure since we will report the actual failures and having an extra failure confuses the build monitoring tooling.

In addition, allow custom properties to be passed through to the mvn command to move towards more TCKs being able to use this method of launching tests.


Azquelt:
#build

Azquelt:
#build

Azquelt:
Previously I had a change in this PR to make metrics and openAPI tests use the common code. However, this will need more work than I originally thought so I'll do that in a follow up PR.

Azquelt:

PR #4907, opened 2018-08-30
Add and test JDBC 4.3 pass through methods and block builder access

alexsm82:
Adds and tests the JDBC 4.3 pass through methods on DatabaseMetadata and the Statement variants.  It also adds tests to ensure the 4 builder methods cannot be called on the wrapper or underlying datasource objects.

### DIFF
--- a/dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43CallableStatement.java
+++ b/dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43CallableStatement.java
@@ -18,6 +18,7 @@ import com.ibm.websphere.ras.TraceComponent;
 import com.ibm.ws.rsadapter.AdapterUtil;
 import com.ibm.ws.rsadapter.impl.StatementCacheKey;
 import com.ibm.ws.rsadapter.jdbc.WSJdbcConnection;
+import com.ibm.ws.rsadapter.jdbc.WSJdbcUtil;
 import com.ibm.ws.rsadapter.jdbc.v42.WSJdbc42CallableStatement;
 
 public class WSJdbc43CallableStatement extends WSJdbc42CallableStatement implements CallableStatement {
@@ -39,27 +40,51 @@ public class WSJdbc43CallableStatement extends WSJdbc42CallableStatement impleme
     public String enquoteLiteral(String val) throws SQLException {
         // KEEP CODE IN SYNC: This method is duplicated in WSJdbc43Statement, WSJdbc43PreparedStatement,
         // and WSJdbc43CallableStatement because multiple inheritance isn't allowed.
-        throw new UnsupportedOperationException();
+        try {
+            return cstmtImpl.enquoteLiteral(val);
+        } catch (SQLException ex) {
+            throw WSJdbcUtil.mapException(this, ex);
+        } catch (NullPointerException nullX) {
+            throw runtimeXIfNotClosed(nullX);
+        }
     }
 
     @Override
     public String enquoteIdentifier(String identifier, boolean alwaysQuote) throws SQLException {
         // KEEP CODE IN SYNC: This method is duplicated in WSJdbc43Statement, WSJdbc43PreparedStatement,
         // and WSJdbc43CallableStatement because multiple inheritance isn't allowed.
-        throw new UnsupportedOperationException();
+        try {
+            return cstmtImpl.enquoteIdentifier(identifier, alwaysQuote);
+        } catch (SQLException ex) {
+            throw WSJdbcUtil.mapException(this, ex);
+        } catch (NullPointerException nullX) {
+            throw runtimeXIfNotClosed(nullX);
+        }
     }
 
     @Override
     public boolean isSimpleIdentifier(String identifier) throws SQLException {
         // KEEP CODE IN SYNC: This method is duplicated in WSJdbc43Statement, WSJdbc43PreparedStatement,
         // and WSJdbc43CallableStatement because multiple inheritance isn't allowed.
-        throw new UnsupportedOperationException();
+        try {
+            return cstmtImpl.isSimpleIdentifier(identifier);
+        } catch (SQLException ex) {
+            throw WSJdbcUtil.mapException(this, ex);
+        } catch (NullPointerException nullX) {
+            throw runtimeXIfNotClosed(nullX);
+        }
     }
 
     @Override
     public String enquoteNCharLiteral(String val) throws SQLException {
         // KEEP CODE IN SYNC: This method is duplicated in WSJdbc43Statement, WSJdbc43PreparedStatement,
         // and WSJdbc43CallableStatement because multiple inheritance isn't allowed.
-        throw new UnsupportedOperationException();
+        try {
+            return cstmtImpl.enquoteNCharLiteral(val);
+        } catch (SQLException ex) {
+            throw WSJdbcUtil.mapException(this, ex);
+        } catch (NullPointerException nullX) {
+            throw runtimeXIfNotClosed(nullX);
+        }
     }
 }
--- a/dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43PreparedStatement.java
+++ b/dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43PreparedStatement.java
@@ -18,6 +18,7 @@ import com.ibm.websphere.ras.TraceComponent;
 import com.ibm.ws.rsadapter.AdapterUtil;
 import com.ibm.ws.rsadapter.impl.StatementCacheKey;
 import com.ibm.ws.rsadapter.jdbc.WSJdbcConnection;
+import com.ibm.ws.rsadapter.jdbc.WSJdbcUtil;
 import com.ibm.ws.rsadapter.jdbc.v42.WSJdbc42PreparedStatement;
 
 public class WSJdbc43PreparedStatement extends WSJdbc42PreparedStatement implements PreparedStatement {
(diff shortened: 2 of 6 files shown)


dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43CallableStatement.java
```diff
@@ -39,27 +41,63 @@ public WSJdbc43CallableStatement(CallableStatement cstmtImplObject, WSJdbcConnec
...
     @Override
     public String enquoteIdentifier(String identifier, boolean alwaysQuote) throws SQLException {
         // KEEP CODE IN SYNC: This method is duplicated in WSJdbc43Statement, WSJdbc43PreparedStatement,
         // and WSJdbc43CallableStatement because multiple inheritance isn't allowed.
-        throw new UnsupportedOperationException();
+        try {
+            return cstmtImpl.enquoteIdentifier(identifier, alwaysQuote);
+        } catch (SQLException ex) {
+            FFDCFilter.processException(ex,
+                                        "com.ibm.ws.rsadapter.jdbc.v43.WSJdbc43CallableStatement.enquoteIdentifier(String, String)", "64", this);
```
njr-11: Misleading method signature (String,String), when the actual signature is (String, boolean).
Also note that common practice for FFDC is to avoid the method name and signature altogether and just do
FFDCFilter.processException(ex, getClass().getName(), "64", this);

dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43PreparedStatement.java
```diff
@@ -46,27 +48,63 @@ public WSJdbc43PreparedStatement(PreparedStatement pstmtImplObject, WSJdbcConnec
...
     @Override
     public String enquoteIdentifier(String identifier, boolean alwaysQuote) throws SQLException {
         // KEEP CODE IN SYNC: This method is duplicated in WSJdbc43Statement, WSJdbc43PreparedStatement,
         // and WSJdbc43CallableStatement because multiple inheritance isn't allowed.
-        throw new UnsupportedOperationException();
+        try {
+            return pstmtImpl.enquoteIdentifier(identifier, alwaysQuote);
+        } catch (SQLException ex) {
+            FFDCFilter.processException(ex,
+                                        "com.ibm.ws.rsadapter.jdbc.v43.WSJdbc43PreparedStatement.enquoteIdentifier(String, String)", "71", this);
```
njr-11: Same as previous comment.  I also wonder if it is even valuable to record FFDC for simple operations that fail in the JDBC driver like this.  Liberty never invokes these of its own choosing. They are only used when the application invokes them, and any exception goes right back to the application which is probably good enough.

dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43Statement.java
```diff
@@ -38,27 +40,63 @@ public WSJdbc43Statement(Statement stmtImplObject, WSJdbcConnection connWrapper,
...
     @Override
     public String enquoteIdentifier(String identifier, boolean alwaysQuote) throws SQLException {
         // KEEP CODE IN SYNC: This method is duplicated in WSJdbc43Statement, WSJdbc43PreparedStatement,
         // and WSJdbc43CallableStatement because multiple inheritance isn't allowed.
-        throw new UnsupportedOperationException();
+        try {
+            return stmtImpl.enquoteIdentifier(identifier, alwaysQuote);
+        } catch (SQLException ex) {
+            FFDCFilter.processException(ex,
+                                        "com.ibm.ws.rsadapter.jdbc.v43.WSJdbc43Statement.enquoteIdentifier(String, String)", "63", this);
```
njr-11: previous comments apply here as well

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/jdbc/WSJdbcProxyMethod.java
```diff
@@ -129,6 +129,8 @@
         // oracle.jdbc.pool.OracleConnectionPoolDataSource.createPooledConnectionBuilder
         // oracle.jdbc.xa.client.OracleXADataSource.createPooledConnectionBuilder
         // oracle.jdbc.xa.OracleXADataSource.createPooledConnectionBuilder
+        unsafeMethods.add("createShardingKeyBuilder");
```
njr-11: Why is this blocked?  I don't see why there would be a concern with the user using an Oracle interface to construct a sharding key.

dev/com.ibm.ws.jdbc_fat_v43/publish/servers/com.ibm.ws.jdbc.fat.v43/server.xml
```diff
@@ -44,6 +44,7 @@
   </library>
 
   <javaPermission codebase="${server.config.dir}/drivers/d43driver.jar" className="java.security.AllPermission"/>
+  <javaPermission className="java.lang.RuntimePermission" name="getClassLoader"/>
```
njr-11: codebase is missing - who are we needing to grant this permission to and why?

dev/com.ibm.ws.jdbc_fat_v43/test-applications/d43driver/src/org/test/d43/jdbc/D43Handler.java
```diff
@@ -63,6 +63,17 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                 return this;
         }
 
+        //overriding default impls for the below methods to show they are being properly delegated to the driver
```
njr-11: Why did we override the default implementation with something that is less functional and less correct?  If we only need to show that our wrapper is being used, we could do that by having the test application first close the Statement and then attempting the operation, expecting the exception message to start with the typical DSRA####

dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43CallableStatement.java
alexsm82: Good catch.

dev/com.ibm.ws.jdbc.4.3/src/com/ibm/ws/rsadapter/jdbc/v43/WSJdbc43PreparedStatement.java
alexsm82: I'm fine with not logging an FFDC, I was just going off what we did for JDBC 4.1/4.2.

dev/com.ibm.ws.jdbc_fat_v43/test-applications/app43/src/jdbc/fat/v43/web/JDBC43TestServlet.java
```diff
@@ -582,4 +590,152 @@ public void testSerialUnsharedInLTC() throws Exception {
...
+
+    /**
+     * Tests the new JDBC 4.3 DatabaseMetaData method supportsSharding.
+     */
+    @Test
+    public void testDBMDSupportsSharding() throws Exception {
+        Connection con = defaultDataSource.getConnection();
+        try {
+            DatabaseMetaData mdata = con.getMetaData();
+            assertFalse(mdata.supportsSharding());
```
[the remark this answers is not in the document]
alexsm82: It could be, I don't generally have strong feelings one way or the other.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/jdbc/WSJdbcProxyMethod.java
alexsm82: You are correct, it shouldn't be...

dev/com.ibm.ws.jdbc_fat_v43/publish/servers/com.ibm.ws.jdbc.fat.v43/server.xml
alexsm82: The application needs it due to the unwrap, I can add the war as the codebase

dev/com.ibm.ws.jdbc_fat_v43/test-applications/d43driver/src/org/test/d43/jdbc/D43Handler.java
alexsm82: We can update the test to do that, although I wasn't particularly concerned with how "correct" this driver is.  It will make the test cases slightly more convoluted, but I guess that is ok.

dev/com.ibm.ws.jdbc_fat_v43/test-applications/app43/src/jdbc/fat/v43/web/JDBC43TestServlet.java
```diff
@@ -582,4 +590,152 @@ public void testSerialUnsharedInLTC() throws Exception {
...
+        XADataSource xaDS = unsharableXADataSource.unwrap(XADataSource.class);
+        try {
+            xaDS.createXAConnectionBuilder();
+            fail("Call to createXAConnectionBuilder on XADataSource should result in an exception");
+        } catch (SQLFeatureNotSupportedException ex) {
+            if (!ex.getMessage().contains("DSRA9130E"))
+                throw ex;
+        }
+
+        try {
```
njr-11: Given that this has nothing to do with unwrap, why is it under this test method?
I expect that when we do the work to support sharding, the mock JDBC driver will implement this method and return an actual sharding key builder, in which case this part of the test will need to be removed anyway.

alexsm82: Correct, it shouldn't be there, I removed it when I unblocked the method from above.

dev/com.ibm.ws.jdbc_fat_v43/test-applications/d43driver/src/org/test/d43/jdbc/D43Handler.java
njr-11: Given that other changes are being made anyway, I think we should fix this.  We should try to avoid getting in the way of already-correct implementation that the driver or JDBC spec default impl provides so that test applications can use all of the JDBC methods on the driver as naturally as possible.

dev/com.ibm.ws.jdbc_fat_v43/publish/servers/com.ibm.ws.jdbc.fat.v43/server.xml
alexsm82: Opening an issue for this.

alexsm82:
#build

alexsm82:
All the failures are unrelated due to the LDAP outage.

Issue #4888, opened 2018-08-29
Fix intermittent failure in ejbcontainer.session.passivation_fat


brideck:
StatefulTimeoutTest.testXMLOnly can sometimes fail with:
javax.ejb.NoSuchEJBException: Stateful bean CMStatefulBeanO timed out

This can happen when the amount of time it takes for recovery log processing to occur exceeds the 15 sec stateful timeout that is configured by the test.

tkburroughs:
merged

PR #4920, opened 2018-08-30
Issue #4888: Work around slow recovery log initialization in FAT

brideck:
- Added EJB to initialze recovery log before running tests


brideck:
Fixes #4888

brideck:
#build

PR #4922, opened 2018-08-30
New test case that `@Produces` and `@Consumes` affects the right header

andymc12:
Unlike JAX-RS resources, the MP Rest Client `@Produces` and `@Consumes` annotation determines the value for the `Accept` and `Content-Type` header, respectively.  The REST client represents a "server view", so the `@Produces` annotation indicates that the _server_ is producing that MediaType.  This test case validates this behavior.

This test case was added due to discussion in the MP Rest Client spec project in Issue [109](https://github.com/eclipse/microprofile-rest-client/issues/109).
Also related to MP Rest Client issue [113](https://github.com/eclipse/microprofile-rest-client/issues/113).


andymc12:
#build

PR #4925, opened 2018-08-31
Handle TCKs which use JUnit in the same way as those that use TestNG

Azquelt:
Some TCKs use JUnit rather than TestNG as their test framework. Both
frameworks generate JUnit result files but put them in different
locations.

Make MvnUtils handle both frameworks by detecting the location and
avoiding relying on the TestNG-specific result files.

Then update the affected TCK FAT buckets to use MvnUtils.


Azquelt:
Looks like this causes some issues with the rest client 1.1 TCK which is using `RepeatTests`.

While not directly related to the problem I'm trying to solve in this PR, it represents a deeper issue that we don't handle repeating TCKs at all (newer results will overwrite older ones).

Azquelt:
#build

dev/com.ibm.ws.microprofile.openapi_fat_tck/publish/tckRunner/tck/pom.xml
```diff
@@ -26,6 +26,8 @@
     </properties>
 
     <build>
+        <directory>${targetDirectory}</directory>
+        <defaultGoal>clean test</defaultGoal>
```
Azquelt: These lines are in all of the other TCKs.

It's necessary to add at least the `<directory>` line to ensure that the results end up in the same location as all other TCKs so that we can use common code to extract and publish the results.

Issue #4912, opened 2018-08-30
Fix missing doPriv in unwrap


alexsm82:
The unwrap code is missing a doPriv around `Proxy.newProxyInstance` as documented below:
```
("java.lang.RuntimePermission" "getClassLoader")
Stack: 
java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "getClassLoader")java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
java.base/java.security.AccessController.checkPermission(AccessController.java:895)
java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:335)
com.ibm.ws.kernel.launch.internal.MissingDoPrivDetectionSecurityManager.checkPermission(MissingDoPrivDetectionSecurityManager.java:45)
java.base/java.lang.reflect.Proxy.checkProxyAccess(Proxy.java:457)
java.base/java.lang.reflect.Proxy.getProxyConstructor(Proxy.java:411)
java.base/java.lang.reflect.Proxy.newProxyInstance(Proxy.java:999)
com.ibm.ws.rsadapter.jdbc.WSJdbcWrapper.unwrap(WSJdbcWrapper.java:472)
jdbc.fat.v43.web.JDBC43TestServlet.testMultipleOpenSharableHandlesInTransaction(JDBC43TestServlet.java:118) 
```
The code is:
```
result = Proxy.newProxyInstance(priv.getClassLoader(interfaceClass),
                                                new Class[] { interfaceClass },
                                                this);
```

Should also remove that privilege from FAT server.xmls once the product code is fixed.

PR #4930, opened 2018-08-31
Add missing doPriv in unwrap path

alexsm82:
Add missing do priv in the unwrap path.  Also removes the assertion that `createConnectionBuilder()` should result in a SQLFeatureNotSupportedException.  Fix for issue #4912.


alexsm82:
#build

alexsm82:
#build

alexsm82:
The build failures are unrelated, either due to the infra outage or known existing defects.

PR #4943, opened 2018-09-03
 Handle TCKs which use RepeatTests

Azquelt:
When a TCK run finishes, add the current RepeatAction ID to the results
file name and to each test name.

This is similar to the behaviour of the FATRunner and allows results
from multiple repeats to be retained and distinguished from each other.

Note: only the last commit is part of this PR, the others will be removed once the prereq #4925 is merged.


Azquelt:
#build

Azquelt:

Azquelt:
Commits are shown after the PB because they were rebased on integration before merging.

PR #4948, opened 2018-09-04
Check for ibm.tolerates versions when required feature is not found

ericglau:
When installing features with installUtility (on WebSphere Liberty), if a required feature dependency is not found in the repository, then also check for the existence of other versions that can satisfy the `ibm.tolerates` directive before throwing an error message.


ericglau:
#build

ericglau:
Test failures are not related to feature installation.

PR #4950, opened 2018-09-04
Interceptor test fix

andymc12:
Resolves a test bug a failure could occur if a test sets the interceptors used on one thread, and then a different test ends up running on that same thread.  This clears the thread local's set of invoked interceptors.


andymc12:
#build

PR #4954, opened 2018-09-04
beginRequest and endRequest when aborted from same thread

njr-11:
Coverage for begin/endRequest on the path where Connection.abort is invoked from the same thread that is using the connection.  This can be within an LTC or a global transaction, and could be on a sharable or unsharable connection.

### DIFF
--- a/dev/com.ibm.ws.jdbc_fat_v43/test-applications/app43/src/jdbc/fat/v43/web/JDBC43TestServlet.java
+++ b/dev/com.ibm.ws.jdbc_fat_v43/test-applications/app43/src/jdbc/fat/v43/web/JDBC43TestServlet.java
@@ -458,7 +458,8 @@ public class JDBC43TestServlet extends FATServlet {
     }
 
     /**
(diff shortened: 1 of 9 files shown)


njr-11:
#build

dev/com.ibm.ws.jdbc_fat_v43/test-applications/app43/src/jdbc/fat/v43/web/JDBC43TestServlet.java
```diff
@@ -445,6 +457,131 @@ public void testMultipleUnsharedConnectionsInTransaction() throws Exception {
         }
     }
 
+    /**
+     * Abort an unsharable connection from the same thread that is using it.
```
alexsm82: This should be shareable connection.

njr-11:
#build

njr-11:
Build has 2 test failures which are pre-existing issues in unrelated componentry

PR #4973, opened 2018-09-05
beginRequest/endRequest when aborted from another thread

njr-11:
Add test coverage for beginRequest/endRequest when sharable and unsharable connections are aborted from a different thread.


njr-11:
#build

PR #4980, opened 2018-09-05
Update EJB to expose CORBA apis on Java 11

alexsm82:
Add an optional feature for exposing the CORBA APIs to the application for EJB.  Still need to determine if APIs should be added to dev/api/spec.


alexsm82:
Rather than expose the CORBA APIs the EJB component will be updated to no longer no use since they weren't actually used in the affected path.

PR #4981, opened 2018-09-05
Add oidc client userinfo endpoint support

brutif:


brutif:
#build

brutif:
Interesting build failure. 
com.ibm.ws.security.openidconnect.client/src/com/ibm/ws/security/openidconnect/client/internal/OidcClientConfigImpl.java (at line 56)
error: OidcClientConfigImpl is not abstract and does not override abstract method isUserInfoEnabled() in ConvergedClientConfig
public class OidcClientConfigImpl implements OidcClientConfig

So the closed stuff would need to be delivered first.

dev/com.ibm.ws.security.social/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -56,6 +56,12 @@ uniqueId.desc=The unique ID.
 authorizationEndpoint=Authorization end point URL
 authorizationEndpoint.desc=Specifies an Authorization end point URL.
 
+userInfoEndpoint=UserInfo end point URL
+userInfoEndpoint.desc=Specifies a UserInfo end point URL.
+
+userInfoEndpointEnabled=User Info endpoint enabled
+userInfoEndpointEnabled.desc=Specifies whether the User Info endpoint should be contacted.
```
bsbyrd1: Avoid "should." Suggestion:
Specifies whether the User Info endpoint is contacted.

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
```diff
@@ -180,5 +180,12 @@ JWK_RETRIEVE_FAILED=CWWKS1747E: A JSON Web Key (JWK) was not returned from the U
 JWK_RETRIEVE_FAILED.explanation=A JSON Web Key cannot be returned from the specified URL. The URL might not be valid, the URL might not be configured to return a JWK, the response from the URL might be empty, or an unknown error occurred.
```
bsbyrd1: Is it necessary to capitalize Web Key? Is it capitalized because of some industry standard? If not, then we should make it lowercase.

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
```diff
@@ -180,5 +180,12 @@ JWK_RETRIEVE_FAILED=CWWKS1747E: A JSON Web Key (JWK) was not returned from the U
...
 JWK_RETRIEVE_FAILED.useraction=Verify that the URL is formatted correctly and specifies a location that is capable of returning JSON Web Keys. Check the status code and content of the response for more information.
 
+USERINFO_RETREIVE_FAILED=CWWKS1748E: An attempt to retrieve UserInfo from from the URL [{0}] failed.  The response status was [{1}] and the content returned was [{2}].
```
bsbyrd1: The object precedes the descriptor; for example:
An attempt to retrieve UserInfo from from the [{0}] URL failed.

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
```diff
@@ -180,5 +180,12 @@ JWK_RETRIEVE_FAILED=CWWKS1747E: A JSON Web Key (JWK) was not returned from the U
...
+USERINFO_RETREIVE_FAILED.explanation=User Info data could not be obtained from the specified URL. The URL might not be valid, the supplied access token might not be valid, the response from the URL might be empty, or an unknown error occurred.
+USERINFO_RETREIVE_FAILED.useraction=Verify that the URL is formatted correctly and specifies a location that is capable of returning User Info data. Check the status code and content of the response for more information.
+
+USERINFO_INVALID=CWWKS1749E: The User Info data [{0}] is invalid because the sub claim does not match the sub claim of the Id Token [{1}].
```
bsbyrd1: The object precedes the descriptor; for example:
The [{0}] User Info data is invalid because the sub claim does not match the sub claim of the [{1}] ID Token.

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
```diff
@@ -180,5 +180,12 @@ JWK_RETRIEVE_FAILED=CWWKS1747E: A JSON Web Key (JWK) was not returned from the U
...
+USERINFO_INVALID.explanation=The sub claim of user info data is required to match the sub claim of the Id token, but it does not.
```
bsbyrd1: Make sure to use consistent capitalization. Info Data has been capitalized to this point. The same is true for ID Token. Use consistent capitalization throughout.

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
brutif: this is an old message already shipped and it is convention.

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
brutif: will fix

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
brutif: It won't be apparent what the data is about so the descriptor needs to lead. 
For example, in the old message 
OIDC_CLIENT_BAD_PARAM_COOKIE=CWWKS1745E: The WASOidcCode cookie [{0}]
is more understandable than"the 4q5890uawerlkfjaxlwu4oiqfrn;adpzov;k/ WASOidcCode cookie...

dev/com.ibm.ws.security.openidconnect.clients.common/resources/com/ibm/ws/security/openidconnect/clients/common/resources/OidcClientMessages.nlsprops
brutif: It won't be apparent what the object is,  so the descriptor needs to lead. 
For example, in the old existing message 
OIDC_CLIENT_BAD_PARAM_COOKIE=CWWKS1745E: The WASOidcCode cookie 4q5890uawerlkfjaxlwu4oiqfrn;adpzov;k/...

is more understandable than "the 4q5890uawerlkfjaxlwu4oiqfrn;adpzov;k/ WASOidcCode cookie...
I'll update to make it a little more declarative though.

brutif:
13287 has merged, so
#build

brutif:
#build

brutif:
let's build again to check latest changes
#build

brutif:
pb is good

dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelper.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    protected void updateAuthenticationResultPropertiesWithUserInfo(ProviderAuthenticationResult oidcResult, String userInfoStr) {
+        oidcResult.getCustomProperties().put(Constants.USERINFO_STR, userInfoStr);
+    }
+
+    // per oidc-connect-core-1.0 sec 5.3.2, sub claim of userinfo response must match sub claim in id token.
+    protected boolean isUserInfoValid(String userInfoStr, String subClaim) {
+        String userInfoSubClaim = getUserInfoSubClaim(userInfoStr);
+        if (userInfoSubClaim == null || subClaim == null || userInfoSubClaim.compareTo(subClaim) != 0) {
```
ayoho: Are there benefits to using `String.compareTo()` as opposed to `String.equals()`?

dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelper.java
```diff
@@ -0,0 +1,130 @@
...
+     * Obtain userInfo from an OIDC provider
+     *
+     * @return the userInfo string, or null
+     */
+    protected String getUserInfoFromURL(ConvergedClientConfig config, SSLSocketFactory sslsf, String accessToken) {
+        String url = config.getUserInfoEndpointUrl();
+        boolean hostnameVerification = config.isHostNameVerificationEnabled();
+
+        // https required by spec, use of http is not spec compliant
+        if (!url.toLowerCase().startsWith("https:") && config.isHttpsRequired()) {
```
ayoho: Is HTTPS always required by the spec? Or only if the client requires it? If the former is the case, should the `config.isHttpsRequired()` check be taken out?

dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelper.java
```diff
@@ -0,0 +1,130 @@
...
+            return null;
+        }
+
+        OidcClientUtil oidccu = new OidcClientUtil();
+        int statusCode = 0;
+        String responseStr = null;
+        try {
+            Map<String, Object> resultMap = oidccu.getUserinfo(url, accessToken, sslsf, hostnameVerification);
+            if (resultMap == null) {
+                throw new Exception("result map from getUserinfo is null");
```
ayoho: This needs to be an NLS message. This relies on our FFDC catching/logging framework to automatically generate an FFDC for an error condition using a string that won't be translated. That goes against our serviceability guidelines.

dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelper.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+            HttpResponse response = (HttpResponse) resultMap.get(ClientConstants.RESPONSEMAP_CODE);
+            if (response == null) {
+                throw new Exception("HttpResponse from getUserinfo is null");
```
ayoho: Similar comment - this should be an NLS message. Hard coded English exception strings go against serviceability guidelines.

dev/com.ibm.ws.security.openidconnect.clients.common/test/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelperTest.java
```diff
@@ -0,0 +1,169 @@
...
+        setExpectations();
+        UserInfoHelperMock uihm = new UserInfoHelperMock(convClientConfig);
+        boolean result = uihm.getUserInfo(null, null, "accesstoken", null);
+        assertFalse("method return should be false", result);
+        assertFalse("updateAuthResult should not have been called",
+                uihm.updateAuthResultCalled);
+    }
+
+    @Test
+    public void testGetUserInfoSubClaim() {
```
ayoho: Would suggest adding tests for when the string argument is:
- Empty
- Not in JSON format
- Doesn't contain the `"sub"` entry
- Contains a `"sub"` entry whose value isn't a string

dev/com.ibm.ws.security.openidconnect.clients.common/test/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelperTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * retrieve userinfo where no subject from id token is present.
+     * userinfo should be ignored, since id token subject and userinfo subject don't match.
+     */
+    @Test
+    public void testRetrieveNoSujbectFromIdToken() {
+        setExpectations();
+        UserInfoHelperMock uihm = new UserInfoHelperMock(convClientConfig);
+        boolean result = uihm.getUserInfo(null, null, "accesstoken", null);
```
ayoho: Would suggest tests be added for:
- Config does not have a user info endpoint configured or user info is disabled by the config
- Access token is null

dev/com.ibm.ws.security.social/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -56,6 +56,12 @@ uniqueId.desc=The unique ID.
 authorizationEndpoint=Authorization end point URL
 authorizationEndpoint.desc=Specifies an Authorization end point URL.
 
+userInfoEndpoint=Userinfo end point URL
+userInfoEndpoint.desc=Specifies a UserInfo end point URL.
```
ayoho: I know the other endpoint descriptions don't have it, but should we provide more information about what the User Info endpoint is, what its format should be, what it's expected to return, etc.?

dev/com.ibm.ws.security.social/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -331,7 +331,12 @@
         <AD id="authorizationEndpoint" name="%authorizationEndpoint" description="%authorizationEndpoint.desc"
             required="false" type="String" />
         <AD id="tokenEndpoint" name="%tokenEndpoint" description="%tokenEndpoint.desc"
+            required="false" type="String" />
+        <!--  userInfoEndpoint is not required, but if used successfully, results are available in an api.  -->
+        <AD id="userInfoEndpoint" name="%userInfoEndpoint" description="%userInfoEndpoint.desc"
```
ayoho: I think it's better practice to leave new configuration options as internal until we have adequate FATs in place that can verify the attribute behaves as expected.

dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelper.java
brutif: it is required by spec but we allow non-spec behavior with isHttpsRequired, so we'll warn about it.

dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelper.java
brutif: it should never happen so I didn't think it worthy of nls but did put some text in there just in case I was wrong.

dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelper.java
brutif: same thing

dev/com.ibm.ws.security.social/resources/OSGI-INF/l10n/metatype.properties
brutif: perhaps, but paraphrasing the spec is a slippery slope.  Also most users will never see these fields, only those using WDT.

dev/com.ibm.ws.security.social/resources/OSGI-INF/metatype/metatype.xml
brutif: you are right, and I should have said the fat's are coming on #13378, but this one has to go first.

dev/com.ibm.ws.security.openidconnect.clients.common/test/com/ibm/ws/security/openidconnect/clients/common/UserInfoHelperTest.java
brutif: interesting tests but they exercise JsonObject.parse which is a mature api.

ayoho: Those aren't testing how the API handles those inputs, they're testing how your method handles the result. As it is, there's nothing that verifies what value `UserInfoHelper.getUserInfoSubClaim()` returns in the case of an unexpected response format.

brutif:
fixes #5149

PR #4995, opened 2018-09-06
make XADataSource default for jdbc-4.3 feature and above

njr-11:
Detect the version of Liberty jdbc feature enabled, and for 4.3 or higher, make XADataSource the default type when otherwise unspecified.


njr-11:
#build

njr-11:
#build

PR #5017, opened 2018-09-07
initial ConnectionBuilder prototype with user/password only

njr-11:
Prototype ConnectionBuilder allowing only user/password and driving through existing codepath rather than through the driver's ConnectionBuilder/PooledConnectionBuilder/XAConnectionBuilder.


njr-11:
#build

njr-11:
Build has 4 test failure in unrelated test buckets that do not even enable this feature.

PR #5030, opened 2018-09-10
New test for plain socket timeout (not using JAX-RS client)

andymc12:
This new test is to be used for ensuring that a given environment is capable of using socket timeouts directly - so as to differentiate between a problem with the environment and a problem with JAX-RS client connection timeouts.


andymc12:
#build

PR #5038, opened 2018-09-10
Updates to LibertyServer in FAT framework

ayoho:
Adds some new startup/reconfig methods to `LibertyServer` and updates the `LogMonitor.verifyStringNotInLogUsingMark()` method to return the line in the log that was found to match the provided string (or `null` if no such line was found). The latter allows us to include the line that was found to match the string in a test failure as opposed to the simple boolean "yes, the string was there" or "no, the string was not there."


ayoho:
#build

PR #5053, opened 2018-09-11
test connectionbuilder for xa and pooled connection with user/password only

njr-11:
Write tests for using JDBC 4.3 ConnectionBuilder to request connections from XADataSource and ConnectionPoolDataSource specifying user/password and verify that connections are established with the correct user/password and are correctly matching & sharing based on user/password.


njr-11:
#build

PR #5057, opened 2018-09-12
Fix the top-level help message for securityUtility. 

barecode:
The --arg format is --arg=value, not --arg value.


barecode:
Failed, but with unrelated errors

PR #5058, opened 2018-09-12
ConnectionBuilder tests with user and password for Driver and DataSource

njr-11:
Add tests for using ConnectionBuilder where user/password are specified (no sharding) and the Liberty data source is backed by either java.sql.Driver or javax.sql.DataSource (separate tests for each).


njr-11:
#build

njr-11:
#build

PR #5062, opened 2018-09-12
pull request to update trace in OidcClientHttpUtil

barbj:
#build


barbj:
Tested with local server in build to make sure trace output is what I expect.
Test errors in build are not related to my update.

PR #5065, opened 2018-09-12
Keep creation TTL for attributesCache on update

### LINKED ISSUE #5064: When updating the Ldap AttributesCache, retain creation time time to live (TTL) behavior
On Ldap requests, when the search cache (SearchResultsCache) is updated, the attributes cache (AttributesCache) is also updated. The VMM Cache implementation was incorrectly placing the attribute entries into the first tier of the cache, making it behave like a last access cache instead of creation time to live cache.

Depending the the timeout of the SearchResultsCache and the AttributesCache, this could cause a delay in when the AttributesCache evicts entries. Unexpected or stale values would remain in the cache when they should have been evicted at an earlier time. For example, if the SearchResultsCache is set to 3 minutes and the AttributesCache is set to 6 minutes, if a search request comes in at minute 4 (after the search entry evicts, but before the attribute entry), then the search will be inserted into the cache and the attribute entry will be reinserted into the cache, starting its creation time over at 6 minutes.

This fix will update the AttributesCache, but not alter the time it remains in the cache.

kristip17:
Fixes 5064

Create an update option on the Cache implementation for the Attributes and Search cache.
Call insert or update depending on whether the attributes value is new.
Add a unit and fat test.

### DIFF
--- a/dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/LdapConnection.java
+++ b/dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/LdapConnection.java
@@ -201,9 +201,9 @@ public class LdapConnection {
      * Returns a hash key for the name|filter|cons tuple used in the search
      * query-results cache.
      *
-     * @param name   The name of the object from which to retrieve attributes.
+     * @param name The name of the object from which to retrieve attributes.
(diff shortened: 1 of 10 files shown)


kristip17:
#build
#spawn.fullfat.bucket=com.ibm.ws.security.wim.adapter.ldap_fat

kristip17:
The two fat failures in my build seem to be existing problems. I verified that my new Fat and Unit test passed.

PR #5066, opened 2018-09-12
Capture global shared library contents in CLS dump introspector

andymc12:
Adding new diagnostics that will print the contents of the global shared library (if any) to the ClassLoadingService dump introspector.  This should help identify when the GSL is enabled - and help determine what files are there.


andymc12:
#build

andymc12:
Previous build failed due to unrelated changes that were merged as part of `ClassLoadingServiceImpl` - since those changes were reverted, I've re-merged to not include those changes.  New #build

PR #5067, opened 2018-09-13
Update old DSA signed certs to use RSA algorithm

aguibert:
This updates all default keystores in the repo with the password "Liberty" and a self-signed DSA certificate to keystores with the same pass and a RSA signed certificate.

To do this update I took a checksum of all *.jks files and updated keystores with a checksum that evaluated to` fd38df41e1a106508c96f5aceee3a00d`, which is the checksum for a self-signed DSA certificate with the password "Liberty"


aguibert:
#build

PR #5068, opened 2018-09-13
Issue #5011: Avoid Loading CORBA classes if remote EJB is not enabled

### LINKED ISSUE #5011: Avoid classloading CORBA APIs in the EJB Container
The ejbcontainer.cdi_fat and com.ibm.ws.ejbcontainer.cdi_ee6_fat buckets are failing on Java 11 because the CORBA APIs are not exposed to the application classloader.  The EJB container needs to be updated to avoid classloading the APIs since they are not actually used in this path.

tkburroughs:
There is no need to generate and/or load the Tie classes for remote
EJB interfaces if the ejbRemote feature is not enabled. This is
important for Java 11, which doesn't provide the CORBA classes
used by the Ties.

The remote wrappers are still generated/loaded to allow an appropriate
error message if an attempt is made to use them when the ejbRemote
feature is not enabled.

fixes #5011


tkburroughs:
#build

tkburroughs:
The failures in the build do not appear to be related to this change; however one of the failures is a classloading failure in an EJB related bucket, so running another build.

#build

PR #5097, opened 2018-09-14
Add jaxb and activation apis to rest client tck fats

alexsm82:
Add the jaxb and activation APIs to the rest client TCK FATs so they can run on Java 9+.


alexsm82:
#build

PR #5098, opened 2018-09-14
Update Saaj config policy

alexsm82:
Fix from the code review comment in #5041 - update the SAAJ config policy to ignore.


alexsm82:
#build

Issue #5099, opened 2018-09-15
EJB timer requires datasource definition even for persistent = false


sdaschner:
Creating a (singleton, startup) EJB with a non-persistent timer definition fails if there is no default datasource defined. The logs show the following line:

```
[ERROR   ] CWWKZ0002E: An exception occurred while starting the application foobar. The exception message was: com.ibm.ws.container.service.state.StateChangeException: java.lang.IllegalStateException: The ejbPersistentTimer feature is enabled, but the defaultEJBPersistentTimerExecutor persistent executor cannot be resolved. The most likely cause is that the DefaultDataSource datasource has not been configured. Persistent EJB timers require a datasource configuration for persistence.
```

However, in this case the EJB timer was actually defined as non-persistent:

```java
@Singleton
@Startup
public class TimerDefinition {

    @Schedule(second = "*/5", minute = "*", hour = "*", persistent = false)
    public void doSomething() {
        // ...
    }
}
```

This should not require a default datasource to be setup.

Tested with OpenLiberty `18.0.0.3/wlp-1.0.22.20180911-0300` on OpenJDK 64-Bit Server VM, version `1.8.0_111-8u111-b14-2~bpo8+1-b14 (en)`

aguibert:
I'm guessing you have the full `ejb-3.2` (or the full javaee 7/8) features enabled, in which case the `ejbPersistentTImer-3.2` feature will be pulled in (that's the thing requiring a DataSource to be configured).

The EJB features are broken down into a few sizes:
 - ejbLite-3.2 (pulled in by `webProfile-7/8`) barebones EJB stuff, including non-persistent EJB timers)
 - ejbPersistentTimer-3.2
 - ejb-3.2 (pulled in by `javaee-7/8`)

If you downgrade your feature list to something that just includes `ejbLite-3.2` and not the full `ejb-3.2` feature, then you can use non-persistent EJBs without needing to configure a DataSource.

However, this still seems like quirky behavior.  If we just use a regular EJB (no `@Schedule` anno) with the full `ejb-3.2` feature, then the app starts fine without requiring a DataSource.  It seems like the EJB feature could do a little better by only looking for a DataSource if an EJB contains `@Schedule(persistent=true, ...)`.

sdaschner:
Thanks, good to know.
 
 
> However, this still seems like quirky behavior. [...]
[...]

PR #5100, opened 2018-09-16
Only require DataSource when persistent timers are present

aguibert:
Proposed fix for issue #5099

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ws/metadata/ejb/EJBMDOrchestrator.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ws/metadata/ejb/EJBMDOrchestrator.java
@@ -4896,6 +4896,7 @@ public abstract class EJBMDOrchestrator {
 
         List<TimerMethodData> timerMethods = new ArrayList<TimerMethodData>();
         Method timeoutMethod = getTimeoutMethod(bmd);
+        boolean hasPersistentTimers = false;
 
         if (timeoutMethod != null) {
             boolean oneParm = timeoutMethod.getParameterTypes().length == 1; //F743-15870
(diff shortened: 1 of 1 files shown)


aguibert:
#build

tkburroughs:
So, the concern with this fix is that it ignores programmatically created persistent timers. If timers were created in the database in the past, server is re-started but database cannot be accessed, then those persistent timers will not run... and the application likely won't behave properly.

Unlike the 'automatic' (i.e. `@Schedule`) timers that are processed in EJBMDOrchestrator, the only way to determine if the application supports programmatic timers is to scan all of the bytecode for all classes in the application and look for calls to the EJBTimerService methods that programmatically create timers, etc, which really isn't practical.

So, with the proposed fix, let's say the app did create programmatic persistent timers in the past.  When the server is restarted.... they won't ever run, because the persistent timer service won't be started... unless the app also happened to have an `@Schedule` of a persistent timer.

I think a more appropriate change would be to detect when a persistent timer datasource just hasn't been configured at all, and if so, then just log a warning, not an error, and allow the application to start.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ws/metadata/ejb/EJBMDOrchestrator.java
```diff
@@ -5011,7 +5015,10 @@ else if (!bmd.metadataComplete) {
 
             // F743-13022 - Check now if the runtime environment supports timers
             // rather than waiting until the automatic timers are created.
-            bmd.container.getEJBRuntime().setupTimers(bmd);
+            if (isTraceOn && tc.isDebugEnabled())
+                Tr.debug(tc, "Are persistent timers present? " + hasPersistentTimers);
+            if (hasPersistentTimers)
```
tkburroughs: Note: if persistent timers exist in the database, they won't run if the server is restarted and there are no `@Schedule` persistent timers.

aguibert:
thanks for the feedback @tkburroughs! That makes perfect sense about programmatic timers.

> I think a more appropriate change would be to detect when a persistent timer datasource just hasn't been configured at all, and if so, then just log a warning, not an error, and allow the application to start.

Sounds good to me.  I'll go ahead and cancel this PR in favor of the above solution.

PR #5102, opened 2018-09-16
Viewer (previously Monitor) role

barecode:
This PR implements a Monitor role for the /metrics, /ibm/api and /IBMJMXConnectorREST endpoints. This monitor role permits only HTTP GET method access to these web resources, allowing for read-only access to Administrator-level information. This addresses a common request from users for  "read only" to this information.


barecode:
Build error appears unrelated.

com.ibm.ws.persistence.utility_fat FAT tests
com.ibm.ws.persistence.utility.fat.DDLGenerationUtilityTest.testScriptSuccess ERROR

java.lang.Exception: 2018-09-16-21:26:41:902 Errors/warnings were found in server com.ibm.ws.persistence.utility.fat logs:
 <br>[9/16/18 21:26:40:158 UTC] 00000031 com.ibm.ejs.j2c.PoolManager                                  E J2CA0046E: Method reserve caught an exception during creation of the ManagedConnection for resource consumerdsNonTxDs, throwing ResourceAllocationException. Original exception: Pool requests blocked, connection pool is being shut down.
 <br>[9/16/18 21:26:40:182 UTC] 00000031 eclipselink.ps.ejb                                           E CWWKD0292E: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.6.7.WAS-v20180816-0dccd4c): org.eclipse.persistence.exceptions.DatabaseException

barecode:
@NottyCode you had mentioned you didn't care for the "Monitor" name for the role. I've got a few alternates:
* Viewer Role
* Read-Only Role
* No-Touchy Role (super serious about this one)
* Observer Role

I'm open to any of these (specially No-Touchy!) or something better. Getting the externals right is key.

mbroz2:
An obvious choice: https://www.youtube.com/watch?v=V4uV3icrmw0

barecode:
@mbroz2 so then the configuration would look like:
<viewer-role>
  <user>bob</user>
</viewer-role>

The reason for the <viewer-role> tag is to indicate A) its a role B) consistent with <administrator-role>

barecode:
The good news is the term "Viewer" is also much more greppable in the source code - spinning up a commit for the "Viewer Role".

barecode:
#build

barecode:
#build

dev/com.ibm.ws.management.security/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -18,6 +18,9 @@
 role.administrator=Administrator Role
 role.administrator.desc=A collection of users and/or groups assigned the server administrator role.
 
+role.viewer=Viewer Role
+role.viewer.desc=A collection of users and/or groups assigned the server viewer role.
```
bsbyrd1: A collection of users, groups, or both that are assigned the server viewer role.

barecode:
@bsbyrd1 if you could re-review please! :)

mbroz2:
Not sure if relevant, but do we need to consider how this Viewer role would work with Java Batch roles (which have their own set of roles, including 2 monitor roles)?  I'm guessing we'll just need to update the Java Batch docs, but worth thinking through it if it hasn't been considered yet.

barecode:
I have not explored it. It definitely needs more thought. However, I was pretty sure the Java Batch roles were totally separate (by design) from the Management roles.

dev/com.ibm.ws.microprofile.metrics.private/resources/WEB-INF/web.xml
```diff
@@ -19,17 +19,17 @@
     <servlet-class>com.ibm.ws.microprofile.metrics.privateapi.PrivateMetricsRESTProxyServlet</servlet-class>
     <load-on-startup>1</load-on-startup>
   </servlet>
-  
+
```
mbroz2: I know for 18.0.0.3 mp metrics added a config flag (<mpMetrics authentication="false|true" />) that allows a user to turn off authentication.  Not familiar with the implementation of that, but it would be good to make sure that the flag works correctly with the viewer role as well

barecode: I believe the test flow in MetricsAuthenticationTest covers this. Specifically, it tests accessing the metrics with Administrator, Viewer and then sets the config flag to false and accesses as unauthenticated.

PR #5108, opened 2018-09-17
Update security FAT infra to allow immediate expectations

ayoho:
Updates the common security FAT infrastructure to allow for immediate expectations, meaning we can create an expectation but don't have to specify a test action associated with it. That allows us to do a couple things:
1. We can set an expectation that will be verified every time we attempt to validate a result, not just if the current test action matches the action associated with the expectation. This is helpful for things like checking 200 status codes that we might expect for every web invocation in a sequence of actions.
2. We can be agnostic about whatever the "current test action" is. If my test performs a single action (e.g. a web resource invocation), I no longer have to name that action, pass it as the action I'm performing in the test case, _and_ associate that test action with whatever expectations I set.


ayoho:
#build

ayoho:
#build

PR #5120, opened 2018-09-18
Update JWK code to support getting key by use entry

ayoho:
The OIDC Basic RP certification tests have a scenario where the JWKS URI used by the RP to verify the ID token returns multiple JWKs, but only one of those JWKs matches the signature algorithm used by the test and specifies `"sig"` as the `"use"` value. The common JWK code needs to be updated to iterate through the JWKs that are returned to determine if there is one, and only one, key that matches the signature algorithm and use combination and return that key.


ayoho:
#build

ayoho:
#build

ayoho:
New personal build: https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_po6uELzTEeieisOzGH_nZQ

PR #5122, opened 2018-09-18
Write fail.log in well-known location to match non-TCK FAT behavior

aguibert:
Currently if a TCK-style FAT fails with only TCK-style tests, the FAT output will not be uploaded to the automated build for investigation, and if running the FAT locally the gradle build will report SUCCESS even if tests failed.

Both of these environments rely on the presence of a file at `autoFVT/output/fail.log` if there is a test failure.  The patch here is simply to produce a `fail.log` file in the event we detect failures in TCK-style tests.


aguibert:
#build

PR #5129, opened 2018-09-18
Remove OnlyRunInJava7Rule and related rules

aguibert:
We have some rules in the FAT framework that were created back when we only tested on Java 6/7, but were not extensible to later java versions.

Now that we no longer test on Java 6, the OnlyRunInJava7 rules are all unnecessary and can be removed.


aguibert:
#build

Issue #5076, opened 2018-09-13
NullPointerException in ClassLoadingServiceImpl


tkburroughs:
Saw the following FFDC during in com.ibm.ws.ejbcontainer.security_fat.features FAT tests. Tests seemed to pass otherwise; just reported this unexpected FFDC

Exception = java.lang.NullPointerException
Source = com.ibm.ws.library.internal.SharedLibraryImpl
probeid = 260
Stack Dump = java.lang.NullPointerException
at com.ibm.ws.classloading.internal.ClassLoadingServiceImpl$3.update(ClassLoadingServiceImpl.java:520)
at com.ibm.ws.classloading.internal.providers.WeakLibraryListener.libraryNotification(WeakLibraryListener.java:51)
at com.ibm.ws.library.internal.SharedLibraryImpl.notifyListeners(SharedLibraryImpl.java:259)
at com.ibm.ws.library.internal.SharedLibraryImpl.publishGeneration(SharedLibraryImpl.java:377)
at com.ibm.ws.library.internal.FilesetListener.filesetNotification(FilesetListener.java:98)
at com.ibm.wsspi.config.internal.FilesetImpl.setListener(FilesetImpl.java:148)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:95)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:55)
at java.lang.reflect.Method.invoke(Method.java:508)
at org.apache.felix.scr.impl.inject.BaseMethod.invokeMethod(BaseMethod.java:229)
at org.apache.felix.scr.impl.inject.BaseMethod.access$500(BaseMethod.java:39)
at org.apache.felix.scr.impl.inject.BaseMethod$Resolved.invoke(BaseMethod.java:650)
at org.apache.felix.scr.impl.inject.BaseMethod.invoke(BaseMethod.java:506)
at org.apache.felix.scr.impl.inject.BindMethod.invoke(BindMethod.java:658)
at org.apache.felix.scr.impl.manager.DependencyManager.doInvokeBindMethod(DependencyManager.java:1727)
at org.apache.felix.scr.impl.manager.DependencyManager.invokeBindMethod(DependencyManager.java:1704)
[...]

PR #5130, opened 2018-09-18
Issue 5076: null check on aclStore

andymc12:
It is possible for aclStore to be null when the server is shutting down.
In that case, we don't need to remove any classloaders from it.

This resolves a (benign) FFDC that is generated intermittently when the server shuts down at the same time that a library update occurs.

This also disables the save action for the com.ibm.ws.classloading project.

This resolves issue #5076.


andymc12:
#build

andymc12:
Failing test is unrelated.

PR #5131, opened 2018-09-18
prototype matching of shardingkey based on equals

njr-11:
Create a prototype implementation of matching for sharding keys based on the existing approach that is used for other connection properties, so that we can start experimenting with writing some tests and further investigating how to best add support for sharding keys.

For now, we will make the assumption that .equals can be used for comparisons and that default value of sharding keys will always be null (the spec has no way to ask for a current value).


njr-11:
#build

njr-11:
#build

PR #5134, opened 2018-09-19
basic test of sharding key matching

njr-11:
Write an initial test of connection matching based on sharding key, using the mock JDBC 4.3 driver, which doesn't actually support sharding, but instead just reports the sharding key values that were set on it, which it does via the JDBC client info mechanism.


njr-11:
#build

njr-11:
Commit is included under #5139 and reviewed as part of that pull, so closing this out

PR #5160, opened 2018-09-21
Individual esa fix

kathrynkodama:
Added install.individual.esa to map based install kernel to allow for an individual esa file to be specified in a pom.xml

### DIFF
--- a/dev/com.ibm.ws.install/resources/com/ibm/ws/install/internal/resources/InstallKernel.nlsprops
+++ b/dev/com.ibm.ws.install/resources/com/ibm/ws/install/internal/resources/InstallKernel.nlsprops
@@ -771,6 +771,14 @@ ERROR_UNKNOWN_OSLIST_VALUE=CWWKF1387E: The value {0} is not a valid entry for th
 ERROR_UNKNOWN_OSLIST_VALUE.explanation=This value will be ignored as it is not a recognized operating system name.
 ERROR_UNKNOWN_OSLIST_VALUE.useraction=Please modify or remove this value from the feature manifest file.
 
+ERROR_ESA_NOT_FOUND=CWWKF1388E: The specified ESA file {0} did not contain any features.
+ERROR_ESA_NOT_FOUND.explanation=A feature will not be installed from the ESA file.
+ERROR_ESA_NOT_FOUND.useraction=Ensure that the individual ESA file provided is valid and contains a feature.
+
+ERROR_SINGLE_REPO_CONNECTION_FAILED=CWWKF1389E: Unable to create a JSON file at {0} for the ESA file {1}.
+ERROR_SINGLE_REPO_CONNECTION_FAILED.explanation=The feature in the ESA file will not be installed.
+ERROR_SINGLE_REPO_CONNECTION_FAILED.useraction=Ensure that write access is available for the directory where the JSON file will be created.
+
 TOOL_DOWNLOAD_FEATURE_OK=CWWKF1500I: The feature has been successfully downloaded.
 TOOL_DOWNLOAD_FEATURE_OK.explanation=The feature was downloaded.
 TOOL_DOWNLOAD_FEATURE_OK.useraction=No action is required.
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
@@ -11,6 +11,13 @@
 package com.ibm.ws.install.internal;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -22,6 +29,8 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.logging.Level;
+
+import org.apache.tools.ant.BuildException;
 
 import com.ibm.ws.install.CancelException;
 import com.ibm.ws.install.InstallConstants;
@@ -48,7 +57,9 @@ import com.ibm.ws.repository.connections.RepositoryConnection;
 import com.ibm.ws.repository.connections.RepositoryConnectionList;
 import com.ibm.ws.repository.connections.SingleFileRepositoryConnection;
 import com.ibm.ws.repository.connections.liberty.ProductInfoProductDefinition;
+import com.ibm.ws.repository.exceptions.RepositoryBackendException;
 import com.ibm.ws.repository.exceptions.RepositoryException;
+import com.ibm.ws.repository.parsers.internal.CreateJsonRepositoryFiles;
 import com.ibm.ws.repository.resolver.RepositoryResolutionException;
 import com.ibm.ws.repository.resolver.RepositoryResolver;
 import com.ibm.ws.repository.resources.RepositoryResource;
@@ -88,6 +99,8 @@ public class InstallKernelMap implements Map {
     private static final String FEATURES_TO_RESOLVE = "features.to.resolve";
     private static final String SINGLE_JSON_FILE = "single.json.file";
     private static final String MESSAGE_LOCALE = "message.locale";
+    private static final String INSTALL_INDIVIDUAL_ESAS = "install.individual.esas";
+    private static final String INDIVIDUAL_ESAS = "individual.esas";
 
     // Return code
     private static final Integer OK = Integer.valueOf(0);
@@ -365,9 +378,20 @@ public class InstallKernelMap implements Map {
             } else {
                 throw new IllegalArgumentException();
             }
+
         } else if (DIRECTORY_BASED_REPOSITORY.equals(key)) {
             if (value instanceof File) {
                 data.put(DIRECTORY_BASED_REPOSITORY, value);
+            }
+        } else if (INDIVIDUAL_ESAS.equals(key)) {
+            if (value instanceof List) {
+                data.put(INDIVIDUAL_ESAS, value);
+            } else {
+                throw new IllegalArgumentException();
+            }
+        } else if (INSTALL_INDIVIDUAL_ESAS.equals(key)) {
+            if (value instanceof Boolean) {
+                data.put(INSTALL_INDIVIDUAL_ESAS, value);
             } else {
                 throw new IllegalArgumentException();
             }
@@ -513,6 +537,13 @@ public class InstallKernelMap implements Map {
             for (File jsonRepo : singleJsonRepos) {
                 RepositoryConnection repo = new SingleFileRepositoryConnection(jsonRepo);
                 repoList.add(repo);
+            }
+            //TODO
+            //check if we do the deleteOnExit or not
+            if (data.get(INSTALL_INDIVIDUAL_ESAS).equals(Boolean.TRUE)) {
+                Path tempDir = Files.createTempDirectory("individualEsas");
+                tempDir.toFile().deleteOnExit();
+                repoList.add(individualESAInstall(tempDir));
             }
 
             ManifestFileProcessor m_ManifestFileProcessor = new ManifestFileProcessor();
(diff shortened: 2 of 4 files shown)


dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -520,6 +551,41 @@ private static boolean containsIgnoreCase(Collection<String> featureToInstall, S
...
+                    }
+
+                }
+            } catch (NullPointerException e) {
+                data.put(ACTION_RESULT, ERROR);
+                data.put(ACTION_ERROR_MESSAGE, e.getMessage());
+                data.put(ACTION_EXCEPTION_STACKTRACE, ExceptionUtils.stacktraceToString(e));
+            }
+
+            for (String feature : featureToInstall) {
```
ericglau: Remove println

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +774,70 @@ private Integer uninstall() {
...
+            while (zipEntries.hasMoreElements()) {
+                ZipEntry nextEntry = zipEntries.nextElement();
+                if ("OSGI-INF/SUBSYSTEM.MF".equalsIgnoreCase(nextEntry.getName())) {
+                    subsystemEntry = nextEntry;
+                    break;
+                }
+            }
+            if (subsystemEntry == null) {
+                ;
+            } else {
```
ericglau: Just use `if (subsystemEntry != null) { Manifest m ...... }`

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +774,70 @@ private Integer uninstall() {
         }
         return OK;
     }
-}
+
+    private static void getShortNameFromManifest(File esa, Map<String, String> shortNameMap) throws IOException {
```
ericglau: Change method name to be more descriptive, since it is currently named "get..." but does not return anything.  Perhaps call it "populateShortNameFromManifest"

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +774,70 @@ private Integer uninstall() {
...
+            }
+        } finally {
+            if (zip != null) {
+                zip.close();
+            }
+        }
+    }
+
+    //accept List of Files
+    private File individualESAInstall(Path generatedJson, Map<String, String> shortNameMap) throws IOException, BuildException, RepositoryException {
```
ericglau: Change method name to be more descriptive, since the current name doesn't tell you what it does.  It doesn't install anything at this point.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +774,70 @@ private Integer uninstall() {
...
+            getShortNameFromManifest(esa, shortNameMap);
+
+            SingleFileRepositoryConnection mySingleFileRepo = null;
+            if (singleJson.exists()) {
+                mySingleFileRepo = new SingleFileRepositoryConnection(singleJson);
+            } else {
+                try {
+                    mySingleFileRepo = SingleFileRepositoryConnection.createEmptyRepository(singleJson);
+                } catch (IOException e) {
+                    e.printStackTrace();
```
ericglau: Don't print stacktrace, should handle the exception

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +774,70 @@ private Integer uninstall() {
...
+                }
+            }
+            System.out.println("Esa name: " + esa.getName());
```
ericglau: Remove println

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -520,6 +551,41 @@ private static boolean containsIgnoreCase(Collection<String> featureToInstall, S
...
+                if (data.get(INSTALL_INDIVIDUAL_ESAS).equals(Boolean.TRUE)) {
+                    Path tempDir = Files.createTempDirectory("generatedJson");
+                    tempDir.toFile().deleteOnExit();
+                    Map<String, String> shortNameMap = new HashMap<String, String>();
+                    File individualEsaJson = individualESAInstall(tempDir, shortNameMap);
+                    RepositoryConnection repo = new SingleFileRepositoryConnection(individualEsaJson);
+                    repoList.add(repo);
+                    for (String feature : featureToInstall) {
+                        if (feature.endsWith(".esa")) {
+                            if (shortNameMap.containsKey(feature)) {
```
ericglau: The `shortNameMap` was populated by `individualESAInstall()` with the keys being absolute paths to the ESA files.  But can ESA files be specified (e.g. in the Maven plugin's `<features>` section) with relative paths?  (You can try this with the existing plugin, which will use installUtility for individual esas.)  If so, then the `feature` string here could be a relative path, so `shortNameMap.containsKey(feature)` would be false even though we want it to match.  You could get around this by comparing canonical paths instead (not absolute paths, because absolute paths would not resolve things such as "." or "..")

kathrynkodama: This was tested and ESA files cannot be specified with relative paths with installUtility.

ericglau: What if they have absolute paths but with things such as "." or "..".  For example, `/folder/../folder/file.esa` and `/folder/file.esa` point to the same file, but comparing absolute paths would give false.  So we should still use canonical paths (so that both will be resolved to `/folder/file.esa`)

kathrynkodama: This structure also results in a failed to invoke error.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
kathrynkodama: This change results in error: Only one "/" symbol is allowed in the resourceString, do you have alternate suggestions? Individual esa install works when this change is reverted.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +764,68 @@ private Integer uninstall() {
...
+                shortNameMap.put(esa.getAbsolutePath(), shortNameAttr);
+            }
+        } finally {
+            if (zip != null) {
+                zip.close();
+            }
+        }
+    }
+
+    //accept List of Files
```
ericglau: Add javadoc style comments.  Generally they are not needed for private methods, but it would help with readability in this case since the method does quite a few things.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +764,68 @@ private Integer uninstall() {
...
+    private File individualESAInstall(Path generatedJson, Map<String, String> shortNameMap) throws IOException, BuildException, RepositoryException, InstallException {
```
ericglau: Perhaps rename method to something like generateJsonsFromIndividualESAs

ericglau: Rename parameter `generatedJson` since the parameter is not a json but is just the temp dir where the jsons will be stored.  Perhaps use `jsonDirectory`

dev/com.ibm.ws.install/resources/com/ibm/ws/install/internal/resources/InstallKernel.nlsprops
```diff
@@ -771,6 +771,14 @@ ERROR_UNKNOWN_OSLIST_VALUE=CWWKF1387E: The value {0} is not a valid entry for th
 ERROR_UNKNOWN_OSLIST_VALUE.explanation=This value will be ignored as it is not a recognized operating system name.
 ERROR_UNKNOWN_OSLIST_VALUE.useraction=Please modify or remove this value from the feature manifest file.
 
+ERROR_ESA_NOT_FOUND=CWWKF1388E: The specified ESA file did not contain any features.
+ERROR_ESA_NOT_FOUND.explanation=A feature will not be installed from the ESA file.
+ERROR_ESA_NOT_FOUND.useraction=Ensure that the individual ESA file provided is valid and contains a valid feature.
```
ericglau: Change to:
```
ERROR_ESA_NOT_FOUND.useraction=Ensure that the ESA file provided is valid and contains a feature.
```

dev/com.ibm.ws.install/resources/com/ibm/ws/install/internal/resources/InstallKernel.nlsprops
```diff
@@ -771,6 +771,14 @@ ERROR_UNKNOWN_OSLIST_VALUE=CWWKF1387E: The value {0} is not a valid entry for th
 ERROR_UNKNOWN_OSLIST_VALUE.explanation=This value will be ignored as it is not a recognized operating system name.
 ERROR_UNKNOWN_OSLIST_VALUE.useraction=Please modify or remove this value from the feature manifest file.
 
+ERROR_ESA_NOT_FOUND=CWWKF1388E: The specified ESA file did not contain any features.
```
ericglau: It would be helpful for the message to have a parameter to say what file it is referring to.
e.g. `The specified ESA file {0} did not contain any features.` where {0} is the file path

dev/com.ibm.ws.install/resources/com/ibm/ws/install/internal/resources/InstallKernel.nlsprops
```diff
@@ -771,6 +771,14 @@ ERROR_UNKNOWN_OSLIST_VALUE=CWWKF1387E: The value {0} is not a valid entry for th
...
+ERROR_ESA_NOT_FOUND.explanation=A feature will not be installed from the ESA file.
+ERROR_ESA_NOT_FOUND.useraction=Ensure that the individual ESA file provided is valid and contains a valid feature.
+
+ERROR_SINGLE_REPO_CONNECTION_FAILED=CWWKF1389E: Unable to establish a single file repository connection for the single JSON.
```
ericglau: [SingleFileRepositoryConnection](https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.repository/src/com/ibm/ws/repository/connections/SingleFileRepositoryConnection.java) throws IOException if the json already exists or there is a problem creating the file.  We can assume it doesn't already exist since InstallKernelMap line 810 checked for that, so the issue would be a problem creating the file.  So the message can be like:

```
ERROR_SINGLE_REPO_CONNECTION_FAILED=CWWKF1389E: Unable to create a JSON file at {0} for the ESA file {1}.
ERROR_SINGLE_REPO_CONNECTION_FAILED.explanation=The feature in the ESA file will not be installed.
ERROR_SINGLE_REPO_CONNECTION_FAILED.useraction=Ensure that write access is available for the directory where the JSON file will be created.
```

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +764,79 @@ private Integer uninstall() {
...
+
+        String dir = jsonDirectory.toString();
+        List<File> esas = (List<File>) data.get(INDIVIDUAL_ESAS);
+        File singleJson = new File(dir + "/SingleJson.json");
+
+        for (File esa : esas) {
+            try {
+                populateShortNameFromManifest(esa, shortNameMap);
+            } catch (IOException e) {
+                throw new InstallException(Messages.INSTALL_KERNEL_MESSAGES.getLogMessage("ERROR_ESA_NOT_FOUND", esa.getCanonicalPath()));
```
ericglau: Use `esa.getAbsolutePath()` for the message, otherwise getting canonical path could throw another exception which is not caught.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -708,4 +764,79 @@ private Integer uninstall() {
...
+
+            SingleFileRepositoryConnection mySingleFileRepo = null;
+            if (singleJson.exists()) {
+                mySingleFileRepo = new SingleFileRepositoryConnection(singleJson);
+            } else {
+                try {
+                    mySingleFileRepo = SingleFileRepositoryConnection.createEmptyRepository(singleJson);
+                } catch (IOException e) {
+                    throw new InstallException(Messages.INSTALL_KERNEL_MESSAGES.getLogMessage("ERROR_SINGLE_REPO_CONNECTION_FAILED", dir,
+                                                                                              esa.getCanonicalPath()));
```
ericglau: Use `esa.getAbsolutePath()` for the message, otherwise getting canonical path could throw another exception which is not caught.

dev/com.ibm.ws.app.manager.springboot/src/com/ibm/ws/app/manager/springboot/internal/LibIndexCache.java
```diff
@@ -66,7 +66,7 @@ protected void setAdaptableModuleFactory(AdaptableModuleFactory adaptableFactory
         this.adaptableFactory = adaptableFactory;
     }
 
-    public File getLibrary(Map.Entry<String, String> LibIndexEntry) {
+    private File getLibrary(Map.Entry<String, String> LibIndexEntry) {
```
ericglau: This and some other files seem unrelated to your changes.  Can you merge or rebase from integration?

kathrynkodama:
build: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_iPJOkMM-Eei9re3pB293Wg&tab=com.ibm.team.build.web.ui.internal.editors.result.overview.OverviewPage

PR #5165, opened 2018-09-23
connection builder cannot allow sharding key when backed by java.sql.Driver

njr-11:
When data source is backed by java.sql.Driver, no connection builder is involved on the driver side.  Without any way to set the specified sharding key on the driver, we should reject this.  The shardingKey(key) and shardingKeyBuilder(key) methods don't declare that they raise SQLException, so we will raise UnsupportedOperationException


njr-11:
#build

njr-11:
build has 1 test failure which is a duplicate of a pre-existing issue

PR #4867, opened 2018-08-28
Update ShrinkWrapSharedServer to allow List and handle servers already copied to build.image

benjamin-confino:
This is preparatory work for https://github.ibm.com/was-liberty/WS-CD-Open/issues/12523

PR #5167, opened 2018-09-24
Fix AfterTypeDiscoveryTest and EnablingBeansXmlValidationTest

Azquelt:
I think AfterTypeDiscoveryTest was missed in #4867

Fixes the path in the test and also updates some of the doc in ShrinkWrapSharedServer.

It looks like EnablingBeansXmlValidationTest fails due to inconsistent errors thrown when validating an invalid beans.xml. It looks like the invalid file references a remote schema and the server is now throwing an error when trying to load that, rather than throwing a validation error. Either is acceptable and stops the app from starting so I've loosened the constraint on the test.


Azquelt:
For RTC 257198

Azquelt:
#build

Azquelt:
Note this is a full FAT test so check that the test was run in the build.

Azquelt:
From the build result, it looks like there's another bug with `testEnablingBeansXmlValidation` which wasn't reported because it was masked by this defect.

Azquelt:
#build

PR #5173, opened 2018-09-24
Add Java permissions for test using plain Java sockets to connect.

andymc12:
The `CxfClientPropsTest`'s `testUrlConnectTimeout` fails with Java 2 security enabled.  This fix resolves the issue, and describes what permissions would be required to make a `URLConnection` to a remote endpoint.


andymc12:
#build

Issue #4464, opened 2018-08-04
APSFOUND-18: Liberty shutdown improvements


brenthdaniel:
Liberty shutdown currently has a quiesce feature, but it doesn't do what one would expect a quiesce feature to do. We need to ensure that sources of incoming work are shutdown and that all existing work is allowed to complete.

gkwan-ibm:
**Serviceability Approval Comment** - Please answer the following questions for serviceability approval (note that this is a new approval step -- comments welcome on how to improve it):

1.  WAD -- does the WAD identify the most likely problems customers will see and identify how the feature will enable them to diagnose and solve those problems without resorting to raising a PMR?  Have these issues been addressed in the implementation?

2.  Test and Demo -- As part of the new serviceability process we're asking feature teams to test and analyze common problem paths for serviceability and demo those problem paths to someone not involved in the development of the feature (eg. L2, test team, or another development team).  
a) What problem paths were tested and demonstrated? 
b) Who did you demo to?
c) Do the people you demo'd to agree that the serviceability of the demonstrated problem scenarios is sufficient to avoid PMRs for any problems customers are likely to encounter, or that L2 should be able to quickly address those problems without need to engage L3?

3.  SVT -- SVT team is often the first team to try new features and often encounters problems setting up and using them.  Note that we're not expecting SVT to do full serviceability testing -- just to sign-off on the serviceability of the problem paths they encountered.
a) Who conducted SVT tests for this feature?
b) Do they agree that the serviceability of the problems they encountered is sufficient to avoid PMRs, or that L2 should be able to quickly address those problems without need to engage L3?  

4.  Queues -- Which L2 / L3 queues will handle PMRs for this feature? Do the respective L2 / L3 teams know they are supporting it?

gscottj:
List of focal points:

Area	Focal Point

Accessibility	Scott Johnston

Globalization/TVT	 ??? 

Legal 	Cass Tucker

WDT 	Leonard Theivendra

FAT	Kevin Smith

IEA/STE/FFW	Swati Kasundra

ID	Brandy Gantt

SVT / CPT	Chris Regan / Brian Hanczaryk

CTS	Scott Highbarger

Performance	Tim Pickett

Serviceability	Don Bourne

Launch Readiness	Dave Sudlik

WAD Approvals	Alasdair Nottingham


Accessibility Focal Point tab in Liberty Dashboard:
[...]

PR #5184, opened 2018-09-25
Shutdown improvements

brenthdaniel:
PR for #4464

### DIFF
--- a/dev/com.ibm.ws.runtime.update/resources/com/ibm/ws/runtime/update/internal/resources/Messages.nlsprops
+++ b/dev/com.ibm.ws.runtime.update/resources/com/ibm/ws/runtime/update/internal/resources/Messages.nlsprops
@@ -58,5 +58,5 @@ quiesce.listeners.not.complete.explanation=One or more operations that were init
 quiesce.listeners.not.complete.useraction=No action is required.
 
 quiesce.waiting.on.threads=CWWKE1107W: {0} threads did not complete during the quiesce period. 
-quiesce.waiting.on.threads.explanation=One or more threads that was active before shutting down the server was still active at the end of the quiesce period.
+quiesce.waiting.on.threads.explanation=One or more threads that were active before shutting down the server were still active at the end of the quiesce period.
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.runtime.update/resources/com/ibm/ws/runtime/update/internal/resources/Messages.nlsprops
```diff
@@ -47,4 +47,16 @@ client.quiesce.begin.useraction=No action is required.
...
+notifications.not.complete=CWWKE1105W: {0} runtime updates did not complete during the quiesce period. 
+notifications.not.complete.explanation=One or more updates to features, configuration, or applications did not complete before the server shutdown.
+notifications.not.complete.useraction=No action is required. 
+
+quiesce.listeners.not.complete=CWWKE1106W: {0} shutdown operations did not complete during the quiesce period. 
+quiesce.listeners.not.complete.explanation=One or more operations that were initiated during the shutdown operation did not complete. 
+quiesce.listeners.not.complete.useraction=No action is required.
+
+quiesce.waiting.on.threads=CWWKE1107W: {0} threads did not complete during the quiesce period. 
+quiesce.waiting.on.threads.explanation=One or more threads that was active before shutting down the server was still active at the end of the quiesce period.
```
bsbyrd1: Suggestion:
One or more threads that were active before shutting down the server were still active at the end of the quiesce period.

brenthdaniel:
#build

olendvcook:
I would like to point out that this PR is directly causing the MDB Failure. I know I brought this up in the issue but I should have paid more attention on this response:

"I think that in general components that implement both Pauseable and QuiesceLIstener should be able to handle situations like this, probably by keeping some state variable or just checking to see if a resource is nulled out."

Our problem is that the component that pauses the resource is a different one than the one that destroys it during shutdown.

olendvcook:
I'll have to dig into it a bit more but it probably makes more sense that I should be changing something in our pause/ server stopping code.

PR #5221, opened 2018-09-27
Revert "Revert "Shutdown improvements""

brenthdaniel:
Reverts OpenLiberty/open-liberty#5219

#5184 was reverted because of an intermittent failure that managed to fail all three attempts in the continuous build. The change that resolves the intermittent is in commercial liberty, so the easiest thing to do here is to revert the revert and redeliver when that change has been integrated.

### DIFF
--- a/dev/com.ibm.ws.channelfw/src/com/ibm/websphere/channelfw/ChannelUtils.java
+++ b/dev/com.ibm.ws.channelfw/src/com/ibm/websphere/channelfw/ChannelUtils.java
@@ -46,7 +46,7 @@ import com.ibm.wsspi.channelfw.exception.RetryableChannelException;
  * <ul>
  * <li>printDebugStackTrace - for pretty-printing a debug stack trace
  * </ul>
- * 
+ *
  * Utility methods for displaying Channel and Channel Chain configuration
  * <ul>
  * <li>displayChannels - displays all configured channels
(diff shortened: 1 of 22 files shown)


brenthdaniel:
#build

Azquelt:
I am not sure why this was delivered. The personal build did not run.

Please could you take a look at RTC 258461 and see if it could have been caused by this change? (It's a reference lookup failure in batch during shutdown)

brenthdaniel:
@Azquelt I don't think we typically  run builds for revert PRs, and in fact "POUND build" won't work because it won't be able to find the branch. I tried to run one just to be thorough. 

This code is likely to surface intermittent issues in components that don't shut down correctly, and that may be the case with 258461. We've run a *lot* of builds with this code over the past few months,  though, and I've never seen this particular warning before.

Azquelt:
In general we don't, but as in this case you're reverting a revert, i.e. re-delivering code that has already caused problems in the build once, I expected one to be run. When I saw the failure in the build, I was hoping to verify that this PR hadn't caused a hard failure by verifying that the same failure hadn't occurred in your personal build.

I do see your point about having run a lot of builds for this change though. If we're expecting it to flush out problems, has anyone been told about it? I linked to this change from the RTC defect I raised, but I only knew about it since I happened to have triaged the build where this change went in.

brenthdaniel:
That's fair. FYI, I took a look at this defect and it seems to be a rare DS intermittent rather than anything caused by this work.

PR #5223, opened 2018-09-27
Common security FAT: Implement JsonObjectExpectation

ayoho:
Implements the `JsonObjectExpectation` class to allow us to verify JSON object values in FATs.


ayoho:
#build

PR #5226, opened 2018-09-27
test connection sharing according to MatchCurrentState with sharding

njr-11:
Write a test case that changes the sharding keys and expect connection matching to happen based on the current state of the connection rather than the original request.


njr-11:
#build

PR #5229, opened 2018-09-28
Common security FAT (Social): Create social-specific formlogin application

ayoho:
Create a social-specific test application that extends the common security `formlogin` application. This allows us to add compilation dependencies on the social product code and output social-specific information in the test application without (unnecessarily) adding all of those dependencies to the common security FAT project.


ayoho:
#build

PR #5236, opened 2018-09-28
Use more performant regex to search logs for errors and warnings

aguibert:
When a FAT test produces a lot of system.out, @jgrassel noticed that a lot of time was being taken scanning the messages.log file for errors and warnings.  This turned out to be the regex string used to search the logs was not being limited a single-line search.


aguibert:
#build

PR #5256, opened 2018-10-02
Align Java EE API features with corresponding EE versions

aguibert:
Currently many of our private javax.* features that provide Java EE APIs are not restricted to their corresponding Java EE version(s).

@dazavala pointed out that by default we may have `servlet-4.0 -> anno-1.0 -> javax.annotations-1.1`, which means Servlet from EE8 is linking with common annotations from Java EE 6.


aguibert:
#build

aguibert:
Although these changes are technically correct, they would tighten up behavior in a way that may break existing applications (as shown by some of our automated tests).  

Closing this PR in favor of a more focused solution on alignment between the servlet and javax.annotation features

PR #5269, opened 2018-10-02
Common security FAT (Social): Create UserInfoJsonExpectation

ayoho:
Creates the `UserInfoJsonExpectation` class to extract and validate UserInfo data from the social login FAT protected resource output.


ayoho:
None of the failures in the linked build appear to be related in any way to these changes.

PR #5272, opened 2018-10-03
Move reactive lib to stable

### LINKED ISSUE #5055: Move org.reactivestreams API jar to dev/api/stable
The com.ibm.websphere.org.reactivestreams.reactive-streams.1.0/bnd.bnd
imports the org.reactivestreams API into the third-party directory. 

This means that all MicroProfile applications that use reactive streams will
need to have "classloader apiTypeVisibility="+third-party" and 
need to be installed via server.xml and cannot be put in the dropins folder.

For example:
```
   <application id="mpReactiveStreamsFAT" name="ReactiveStreamsTest" type="war" location="ReactiveStreamsTest.war">
   </application>
```

I propose that use of this jar does not need the protection that the third-party safety
catch provides and there is more customer benefit to removing the need for third-party
to load it by putting it in a place that is on the classpath of a dropped in app - for example
stable (or somewhere else if preferred to keep that 100% Microprofile). The MicroProfile
Reactive Streams Operators Spec specifically discussed this API at the core and its
API classes will not compile if this API is materially changed as they implement 100%
of the the interfaces directly.

The other reasons this API is stable is:

   the exact version is source form maven central 
   it is old and has not changed for years
   it is used by multiple vendors and products for interoperability
   it has 4 interfaces each of which has one or two simple methods 
   it is 100% interfaces 
   no one is likely to change it as it becomes redundant with the java Flow API in Java 9
   there is no active work on the API and no open issues on it
  
I propose that it is moved back to the stable directory so that reactive apps
can be dropped in to liberty that this will not introduce any risk due to changes
in the jar

hutchig:
fixes #5055

### DIFF
--- a/dev/com.ibm.ws.microprofile.reactive.streams.operators_fat/publish/servers/ReactiveStreamsTestServer/server.xml
+++ b/dev/com.ibm.ws.microprofile.reactive.streams.operators_fat/publish/servers/ReactiveStreamsTestServer/server.xml
@@ -13,7 +13,6 @@
 	</featureManager>
 
    <application id="mpReactiveStreamsFAT" name="ReactiveStreamsTest" type="war" location="ReactiveStreamsTest.war">
-  		<classloader apiTypeVisibility="+third-party"/>
    </application>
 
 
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.microprofile.reactive.streams.operators_fat/publish/servers/ReactiveStreamsTestServer/server.xml
```diff
@@ -13,7 +13,6 @@
 	</featureManager>
 
    <application id="mpReactiveStreamsFAT" name="ReactiveStreamsTest" type="war" location="ReactiveStreamsTest.war">
```
tevans78: now that you don't need the classloader bit, do you need the application entry at all? can it go back to using dropins? Fine if that's a separate PR

PR #5278, opened 2018-10-03
tests to experiment with vendor-specific sharding key defaults

njr-11:
Write automated tests to experiment with how vendor-specific sharding key defaults will interact with connection matching.  In this pull, we address the scenario where a null sharding key (or super sharding key) means that a default that is configured at data source level applies.  In this case, there is no distinction between unspecified and null -- both mean use the default.


njr-11:
#build

PR #5281, opened 2018-10-03
CXF-7860: Reprocess `@FormParam`s prior to resource method invocation

andymc12:
This is a temporary fix (until [CXF-7860](
https://issues.apache.org/jira/browse/CXF-7860) is pulled into a Liberty release).  This fixes the issue documented in JAX-RS API Issue [659](
https://github.com/eclipse-ee4j/jaxrs-api/issues/659) - where a parameter annotated with `@FormParam("value")` might have an injected value that does not match a `Form` parameter's value.

This PR includes the same change submitted to the CXF community, and also a test case.


andymc12:
#build

andymc12:
#build - previous failed with unrelated ANT compilation errors...

PR #5286, opened 2018-10-03
Fail startup if tmp dir is full.

### LINKED ISSUE #4132: full tmp dir prevents server from reading server.env during startup
If /tmp directory is full and you want to start a server, the server starts but with wrong environment vars. during startup you see exceptions like this:

wlp/bin/server: line 161: cannot create temp file for here document: No space left on device

According to the return code server started successfully which is almost true (according to messages.log) and "server status" command also says "server is running". However, the server initializes without the given server.env which could result in a wrong behaviour of the application. Think what happens if you put db-connection-urls / usernames etc. in the server.env and they aren't available. In the worst case the application switches to some default "developer environment" behaviour in production ...

You could easily recreate this problem just with a full /tmp directory.

https://github.com/OpenLiberty/open-liberty/blob/29aa36b798700d30808f535aa32cd2e32d745d98/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server#L159

hibell:
It was reported that starting the server with a full tmp directory
prevents the server from reading server.env causing the server to start
with the wrong environment vars.

fixes #4132

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -161,7 +161,13 @@ safeEcho()
   cat <<EOF
 $*
 EOF
+
+  return $?
 }
+
+# Starting the server with a full tmp directory may prevent the server from
+# reading server.env during startup.
+safeEcho "" || exit $?
 
 # Define escapeForEval functions using ${var//find/replace} and ${var#suffix}
 # if possible since those constructs are significantly faster than safeEcho+sed


hibell:
## Example Output
```
$ hwibell@hwibell-ubuntu:~/Development/wlp$ df -h $TMPDIR
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           1.0M  1.0M     0 100% /mnt/ramdisk

$ hwibell@hwibell-ubuntu:~/Development/wlp$ bin/server start
bin/server: line 161: cannot create temp file for here-document: No space left on device

$ hwibell@hwibell-ubuntu:~/Development/wlp$ echo $?
1
```

hibell:
#build

hibell:
Test failures are unrelated to change.

PR #5292, opened 2018-10-04
begin and end request are not safe to invoke

njr-11:
Given how we have seen JDBC driver implementations of the JDBC 4.3 spec methods Connection.beginRequest/Connection.endRequest with behavior that is contrary to the intent of the JDBC spec, and which will destabilize the connection pool and statement cache in a way that will adversely impact users, this pull updates OpenLiberty to stop invoking these optional methods.  We would be open to reintroducing invocation of these "hints" to the JDBC driver in future spec levels where the language around these methods is clarified to guarantee that connection state will not be corrupted.


njr-11:
#build

PR #5296, opened 2018-10-04
should not assume that JDBC driver will allow reuse of connection builders

njr-11:
We should not be making the assumption that a JDBC driver will allow reuse of ConnectionBuilder/PooledConnectionBuilder/XAConnectionBuilder after it has been used to create a connection.  The JDBC 4.3 spec is silent in this area and so the behavior is left to the JDBC driver.  To enforce that we are not assume any particular behavior when using the JDBC driver's connection builders to create connections, I'll be updating our Mock JDBC 4.3 driver under this pull to reject build operations as well as mutator operations after build has been invoked for the first time. This will ensure we only use a connection builder instance once and that we have not (and will not in the future) coded any dependencies on vendor-specific JDBC driver behavior around connection builders.


njr-11:
#build

PR #5298, opened 2018-10-04
New test for ensuring that a ContainerRequestFilter can modify headers

andymc12:
The JAX-RS spec requires that a `ContainerRequestFilter` should be able to modify the HTTP headers via the `ContainerRequestContext`.  A recent PR in the CXF community is seeking to improve handling of cases where multiple values for the same header exists (see [CXF-7842](
https://issues.apache.org/jira/browse/CXF-7842) and CXF PR [445](
https://github.com/apache/cxf/pull/445) but is failing the JAX-RS TCK because it (in it's current form) is making the headers map immutable.  This test ensures that the map remains mutable.

New test - not a release bug.


andymc12:
#build

andymc12:
Merged with changes in master - new #build

PR #5304, opened 2018-10-04
Update JPA ORM Diagnostics Trace to utilize function for Introspector…

jgrassel:

Updated the JPA ORM Diagnostic collector on the trace side of things, utilizing the code updates which were earlier dropped in for the dump introspector.


jgrassel:
#build

dev/com.ibm.ws.jpa.container.core/src/com/ibm/ws/jpa/management/JPAIntrospection.java
```diff
@@ -26,30 +26,116 @@
...
 import com.ibm.ws.jpa.diagnostics.JPAORMDiagnostics;
 import com.ibm.ws.jpa.diagnostics.class_scanner.ano.jaxb.classinfo10.ClassInfoType;
-import com.ibm.ws.jpa.diagnostics.puscanner.PersistenceUnitScannerResults2;
+import com.ibm.ws.jpa.diagnostics.puscanner.PersistenceUnitScannerResults;
 
 /**
  *
  */
 public class JPAIntrospection {
+    private static final TraceComponent tc = Tr.register(JPAORMDiagnostics.class,
```
tkburroughs: This is the wrong class. Should be JPAIntrospection.class

dev/com.ibm.ws.jpa.container.core/src/com/ibm/ws/jpa/management/JPAIntrospection.java
```diff
@@ -166,92 +252,133 @@ public void executeIntrospectionAnalysis(final PrintWriter dout) {
...
+                sb.append(baos.toString());
+
+                Tr.debug(tc, "JPAORMDiagnostics Dump", sb.toString());
+            }
+        } catch (Throwable t) {
+            FFDCFilter.processException(t, JPAIntrospection.class.getName() + ".executeTraceAnalysis", "185");
+        }
+    }
+
+    private void doExecuteIntrospectionAnalysis(final PrintWriter dout) {
```
tkburroughs: Should methods like this have a try/catch to ensure any error we encounter during introspection does not prevent the application itself from working?  Seems that an application that could start and run just fine may fail to start if we have a bug in some of our introspection code, like this method.

dev/com.ibm.ws.jpa.container.core/src/com/ibm/ws/jpa/management/JPAIntrospection.java
jgrassel: Ah!  Great catch, thanks!

dev/com.ibm.ws.jpa.container.core/src/com/ibm/ws/jpa/management/JPAIntrospection.java
jgrassel: Good point, wrapping the static public method body impls with a catch block that logs to ffdc.

jgrassel:
Pushed a new commit with review suggestions.

jgrassel:
Test failure looks unrelated.  Merging.

PR #5327, opened 2018-10-05
Add javax.activity package and provision to apps on Java 9+

aguibert:
The `javax.activity` package has been in JDK 5-8, but was removed in Java 9.  

I wasn't able to find an existing open source dependency of the exception classes, but since this package only contains 3 simple exception classes we will just re-code them based on javadoc spec.

Since these exceptions don't really belong with any particular feature and is a very small dependency, I'm going to provide this package to the app classloader always.


aguibert:
#build

aguibert:
#build

aguibert:
#build

aguibert:
#build

dev/com.ibm.websphere.javaee.activity.1.0/src/javax/activity/ActivityCompletedException.java
```diff
@@ -0,0 +1,34 @@
...
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package javax.activity;
+
+import java.rmi.RemoteException;
+
+/**
+ * A mirror of the corresponding class that was provided in JDK 5-8, and removed in JDK 9+
+ */
+@SuppressWarnings("serial")
```
tkburroughs: This should not be used on an Exception class, especially one that extends RemoteException.....  if the serialversionUID doesn't match, then this class cannot be returned to remote clients.
Really need to find the serialversionUID that is in the classes removed from the JDK to ensure interoperability.

dev/com.ibm.websphere.javaee.activity.1.0/src/javax/activity/ActivityRequiredException.java
```diff
@@ -0,0 +1,34 @@
...
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package javax.activity;
+
+import java.rmi.RemoteException;
+
+/**
+ * A mirror of the corresponding class that was provided in JDK 5-8, and removed in JDK 9+
+ */
+@SuppressWarnings("serial")
```
tkburroughs: This should not be used on an Exception class, especially one that extends RemoteException..... if the serialversionUID doesn't match, then this class cannot be returned to remote clients.
Really need to find the serialversionUID that is in the classes removed from the JDK to ensure interoperability.

dev/com.ibm.websphere.javaee.activity.1.0/src/javax/activity/InvalidActivityException.java
```diff
@@ -0,0 +1,34 @@
...
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package javax.activity;
+
+import java.rmi.RemoteException;
+
+/**
+ * A mirror of the corresponding class that was provided in JDK 5-8, and removed in JDK 9+
+ */
+@SuppressWarnings("serial")
```
tkburroughs: This should not be used on an Exception class, especially one that extends RemoteException..... if the serialversionUID doesn't match, then this class cannot be returned to remote clients.
Really need to find the serialversionUID that is in the classes removed from the JDK to ensure interoperability.

dev/com.ibm.websphere.javaee.activity.1.0/src/javax/activity/ActivityCompletedException.java
tkburroughs: It appears the original classes did not provide a serialVersionUID either; nor any instance variables... so I'm approving.  May be nice to have a unit test that serializes using the Java 8 class, then deserializes using this class; but that is a nice to have.

dev/com.ibm.websphere.javaee.activity.1.0/src/javax/activity/ActivityRequiredException.java
tkburroughs: It appears the original classes did not provide a serialVersionUID either; nor any instance variables... so I'm approving.  May be nice to have a unit test that serializes using the Java 8 class, then deserializes using this class; but that is a nice to have.

dev/com.ibm.websphere.javaee.activity.1.0/src/javax/activity/ActivityCompletedException.java
aguibert: Discussed with Tracy on slack -- these exceptions surprisingly do not define a svUID, and do not appear to have any instance fields/methods.

dev/com.ibm.websphere.javaee.activity.1.0/src/javax/activity/InvalidActivityException.java
tkburroughs: It appears the original classes did not provide a serialVersionUID either; nor any instance variables... so I'm approving.  May be nice to have a unit test that serializes using the Java 8 class, then deserializes using this class; but that is a nice to have.

PR #5330, opened 2018-10-05
tests with jdbc-4.3 feature and older JDBC driver

njr-11:
Update the automated test bucket for the JDBC 4.3 feature to also use a JDBC driver that is of a spec level earlier than 4.3, such that it runs with the jdbc-4.3 feature enabled.  This will help verify that the jdbc-4.3 feature allows drivers at prior spec levels, as its intent is to support JDBC drivers up to and including 4.3.


njr-11:
Per Alex's comment, updated this pull to include tests within the existing JDBC 4.3 feature bucket so that we do not need to force any other buckets to use Java 11.  We are not ready to take the latter step at this time.

njr-11:
#build

Issue #5339, opened 2018-10-09
H2: Fix race condition in multi-stream writing logic


billw1024:
Race condition between the dedicated per connection thread controlling writes off of a write work queue and a stream thread putting working onto the queue.

PR #5340, opened 2018-10-09
fix a write race condition for multiple streams writing on a connection

billw1024:
Also, while updating the code to fix the race condition, I remove the internal async writing code of the H2 code.  This code is not used, but was present in case it would be required by the design, which it is not.  It was removed for this PR since it would have needed a similar update as then internal sync path to fix a probable race condition.


billw1024:
fixes #5339

billw1024:
#build

billw1024:
Green Build, 64000+ testcases passed

billw1024:
#build

billw1024:
#build

billw1024:
#build

billw1024:
#build

billw1024:
build had 64000+ tests pass.  3 failures not related to this Http2 change.  And like a cat coughing up a fur ball I finally got a good build.

PR #5344, opened 2018-10-09
create internal features that we can use to no-op MicroProfile Concurrency

njr-11:
MicroProfile Concurrency cannot be implemented when Java 7 is used because it requires Java 8 or higher.  Yet, we still want our implementation of EE Concurrency (ManagedExecutorService and ContextService) to be interchangeable with MicroProfile Concurrency.  This pull is the start of coming up with a way that our implementations can conditionally implement MicroProfile Concurrency based on the Java SE level.


njr-11:
#build

njr-11:
build has 2 test failures which are both matches for pre-existing issues in unrelated test buckets

PR #5351, opened 2018-10-10
include pre-release version of microprofile concurrency spec

njr-11:
updates to get open liberty to pull in a pre-release version of the MicroProfile concurrency spec that we can start to experiment with initially under a non-ship feature


njr-11:
#build

njr-11:
#build

njr-11:
In the build result, the feature checker didn't like this pull because one of the bundles providing spec API isn't yet used by a feature.  That feature is provided under pull #5357, so I'll close out this pull and let 5357 deliver both commits

PR #5352, opened 2018-10-10
rebased with new branch, added delete file wrapper for exceptions

bberry1675:
Wrapped the file delete operation in RemoteFile.java with the java.nio.Files class so that in the case of failure the reason is printed as an exception to the logs in order to help with debugging.


bberry1675:
#build

bberry1675:
#build

PR #5359, opened 2018-10-10
update JAXRSClientSSLDefaultTest to get the root server directory fro…

WhiteCat22:
…m the LibertyServer API instead of relying on System.getenv() like it did in WS-CD-OPEN


WhiteCat22:
SERVER_CONFIG_DIR is not exported by the *nix server script, so unless a user has set it in their environment it won't show up in System.getenv.  On windows anything that gets set in the script ends up showing up in getenv, so SERVER_CONFIG_DIR would always be there.

WhiteCat22:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_FXEygMy_EeiVwtHPioAdTw

WhiteCat22:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_YAjy4M1_EeiVwtHPioAdTw

Green build

PR #5360, opened 2018-10-10
Improve support for installing individual ESAs

ericglau:
- Avoid ConcurrentModificationException when installing two individual ESAs, by not changing the contents of a collection while looping over its items.
- For individual ESAs that do not have Maven coordinates, set its file path instead, so that the file path can be retrieved by the plugin after features are resolved.
- For individual ESAs that do not have a short name, support using its symbolic name as the feature name instead.


ericglau:
#build

PR #5365, opened 2018-10-11
Run on ee8 but only run on ee7 in full

benjamin-confino:


benjamin-confino:
#build

dev/com.ibm.ws.microprofile.faulttolerance_fat_tck/bnd.bnd
```diff
@@ -19,6 +19,9 @@ src: \
 
 fat.project: true
 
+tested.features:\
```
tevans78: if this sometimes runs on ee7 then doesnt the tested.features need to include cdi-1.2 and servlet-3.1?

dev/com.ibm.ws.microprofile.faulttolerance_fat_tck/fat/src/com/ibm/ws/microprofile/faulttolerance/tck/FATSuite.java
```diff
@@ -10,10 +10,20 @@
...
+import componenttest.rules.repeater.FeatureReplacementAction;
+import componenttest.rules.repeater.RepeatTests;
+
 @RunWith(Suite.class)
 @SuiteClasses({ FaultToleranceTckLauncher.class })
-public class FATSuite {}
+public class FATSuite {
+
+    @ClassRule
+    public static RepeatTests r = RepeatTests.withoutModification()
```
tevans78: Would it be better to say repeat with EE7 andWith EE8 ... then the bit below says skip for EE7

dev/com.ibm.ws.microprofile.faulttolerance_fat_tck/fat/src/com/ibm/ws/microprofile/faulttolerance/tck/FaultToleranceTckLauncher.java
```diff
@@ -96,7 +99,16 @@ public static void tearDown() throws Exception {
      */
     @Test
     @AllowedFFDC // The tested exceptions cause FFDC so we have to allow for this.
-    public void launchFaultToleranceTCK() throws Exception {
+    @Mode(TestMode.FULL)
```
tevans78: this is clever but please add a comment to explicitly say what each test does

dev/com.ibm.ws.microprofile.faulttolerance_fat_tck/bnd.bnd
benjamin-confino: I believe the way it works is this is for features that are not in any of the server.xml files but are swapped in later. At the moment it's set up to start on EE7 and move to EE8 so cdi-1.2 and servlet-3.1 are in the server.xml.

dev/com.ibm.ws.microprofile.faulttolerance_fat_tck/fat/src/com/ibm/ws/microprofile/faulttolerance/tck/FATSuite.java
benjamin-confino: Hmmm. I've never tried a test suite without a `withoutModification()` so I don't know how it would behave. If I had to guess it would go through server.xmls and swap cdi-1.2 for cdi-1.2

Harmless but less optimised than `withoutModification()`.

PR #5357, opened 2018-10-10
mpConcurrency-1.0 feature initially as nonship

njr-11:
Create the mpConcurrency-1.0 feature as non-ship so that we can experiment with an implementation.

PR #5368, opened 2018-10-11
move existing managed completable future impl into concurrent mp bundle and enable resource injection as MP concurrency types

njr-11:
So that we can reuse the existing ManagedCompletableFuture implementation in our MicroProfile Concurrency implementation, copy it over to the com.ibm.ws.concurrent.mp.1.0 bundle.  Then update the existing test bucket to use it directly, so we maintain that coverage.  This pull does not include any work to transition the implementation/tests to the MP Concurrency API.  That work will instead be done in a series of steps as we gradually transform ManagedCompletableFuture into our implementation of MP Concurrency.


njr-11:
Note to @OpenLiberty/message-reviewer it is not necessary for you to review any changes under this pull if you cover them under #5357 which contains the commit needing ID review.  Delivery of #5357 is still blocked on an ID review which is why the commits are stacking up and also appearing under this pull.

njr-11:
#build

njr-11:
commits were merged under #5380, which also had to disable resource injection of MP concurrency types due to code that validates the types in creates.objectClass.  We will need a different solution for that.

PR #5382, opened 2018-10-12
 Make Asynchronous methods return when they timeout

### LINKED ISSUE #3919: Future does not return immediately when timeout fires when using Timeout with Async
On a method like this:

```
@Asynchronous
@Timeout(500)
public Future<String> doWork() {
    while (true) {
        try {
            Thread.sleep(10000);
            break;
        } catch (InterruptedException e) {}
    }
}
```
a call to `doWork().get()` will take over 10 seconds to complete, it should take half a second.

While the method had made itself uninterruptable, we should still allow the returned `Future` to throw a `TimeoutException` when the timeout fires, even if the requested task remains running.

We should also add a TCK for this for the next version.

Azquelt:
Fixes #3919 

Make Asynchronous methods as soon as they timeout, rather than waiting for the method to actually return.

This change adds the ability to abort the execution of an asynchronous method and set its result immediately.

We can then use this to abort execution of an asynchronous method when a timeout occurs.

However, we can't do this if Retry or Fallback are in use as they need to capture and handle the exception and return a different result.

In addition, some refactoring of ExecutionContextImpl and QueuedFuture was required to ensure that we take the correct code paths when starting execution of an asynchronous method. The exising code could cause the calling thread to be interrupted if the timeout occurred before the asynchronous method started running.

I think this also makes the async double call scenario work (either way, the test now passes so I've enabled it).

### DIFF
--- a/dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/async/QueuedFuture.java
+++ b/dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/async/QueuedFuture.java
@@ -12,7 +12,7 @@ package com.ibm.ws.microprofile.faulttolerance.impl.async;
 
 import java.util.ArrayList;
 import java.util.concurrent.Callable;
-import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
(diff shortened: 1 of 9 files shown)


Azquelt:
#build

dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/async/QueuedFuture.java
```diff
@@ -31,72 +31,90 @@
...
-                return outerCancelled;
-            }
+        Future<Future<R>> resultFuture = getResultFuture();
+        boolean cancellationResult = resultFuture.cancel(false); // Result will always be a cancellation exception
 
+        if (cancellationResult) {
+            getTaskFuture().cancel(mayInterruptIfRunning); // Attempt to cancel the queued/running task
+        } else if (resultFuture.isDone()) {
+            try {
+                Future<R> innerFuture = resultFuture.get();
```
tevans78: Catch CancellationException

tevans78: e.g. if method called twice

dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/async/QueuedFuture.java
```diff
@@ -218,44 +206,78 @@ public R get(long methodTimeout, TimeUnit methodUnit) throws InterruptedExceptio
      * @throws Exception
      */
     @Override
-    public Future<R> call() throws Exception {
+    @FFDCIgnore(Throwable.class)
+    public Void call() throws Exception {
```
tevans78: This class can be a Runnable rather than a Callable

dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/async/QueuedFuture.java
Azquelt: Added `!isCancelled()` instead.

dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/async/QueuedFuture.java
Azquelt: Done

dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/async/QueuedFuture.java
```diff
@@ -52,27 +51,23 @@
      * <p>
      * Don't use it to get the result, use {@link #resultFuture} instead.
      */
-    private Future<Void> taskFuture;
+    private Future<?> taskFuture;
 
     //threadContext is the context retrieved from the originating thread
     private ThreadContextDescriptor threadContext;
     //executionContext is the overall FT execution context, covering both synchronous halves of the asynchronous execution
```
tevans78: you removed the executionContext but not the comment?

Azquelt: doh

Azquelt: Fixed

Azquelt:
#build

Azquelt:

PR #5391, opened 2018-10-15
Minor fix and logging in auto extract test

brenthdaniel:
Set mark to end of log in auto extract test and add a log message.


brenthdaniel:
#build

PR #5394, opened 2018-10-15
allow external builds to find pre-release mp concurrency binary

njr-11:
The pre-release MicroProfile Concurrency binary that we are using to build against was only available internally.  This pull should make it possible to build this code externally as well.


njr-11:
#build

njr-11:
build has 2 test failures which are preexisting issues in unrelated test buckets

PR #5274, opened 2018-10-03
Change CWSIP05800W to CWSIP0580W

AndrewDJBanks:
Fixes #4202

PR #5401, opened 2018-10-16
Change CWSIP05800W to CWSIP0580W

### LINKED ISSUE #4202: Migration of JMS delivery delay.
com.ibm.ws.sib.msgstore.persistence.objectManager.PersistableMetadata assumes there is no space after the end of the data in the restored serialised buffer.
It uses the remaining length to determine if  jms deliveryDelayTime is present instead of introducing a new serialVersionUid..
Unfortunately if there  are 8 or more bytes present a bogus delivery time will be created. If 1-7 bytes are present readOblect fails
with EOFException

AndrewDJBanks:
Fixes #4202

### DIFF
--- a/dev/com.ibm.ws.messaging.common/resources/com/ibm/ws/sib/processor/CWSIPMessages.nlsprops
+++ b/dev/com.ibm.ws.messaging.common/resources/com/ibm/ws/sib/processor/CWSIPMessages.nlsprops
@@ -1298,7 +1298,7 @@ MESSAGE_EXISTS_ERROR_CWSIP0572.useraction=None
 # {1} The system message id
 # {2} The destination name
 # {3} The action taken
-DELIVERY_DELAY_TIME_WARNING_CWSIP0580=CWSIP05800W: A message with ID {0} system message ID {1} was found in destination {2} with a delivery delay exceeding the limit set by the property sib.msgstore.maximumAllowedDeliveryDelayInterval, that action taken was ''{3}''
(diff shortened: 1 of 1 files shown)


AndrewDJBanks:
#build

ngmr:
Supercedes #5274

dev/com.ibm.ws.messaging.common/resources/com/ibm/ws/sib/processor/CWSIPMessages.nlsprops
```diff
@@ -1298,7 +1298,7 @@ MESSAGE_EXISTS_ERROR_CWSIP0572.useraction=None
 # {1} The system message id
 # {2} The destination name
 # {3} The action taken
-DELIVERY_DELAY_TIME_WARNING_CWSIP0580=CWSIP05800W: A message with ID {0} system message ID {1} was found in destination {2} with a delivery delay exceeding the limit set by the property sib.msgstore.maximumAllowedDeliveryDelayInterval, that action taken was ''{3}''
+DELIVERY_DELAY_TIME_WARNING_CWSIP0580=CWSIP0580W: A message with ID {0} system message ID {1} was found in destination {2} with a delivery delay exceeding the limit set by the property sib.msgstore.maximumAllowedDeliveryDelayInterval, that action taken was ''{3}''
```
chirp1: I see that just the message number was changed, but here are some comments for the message text, explanation, and user action:
Change "A message with ID {0} system message ID {1} was found in destination {2} with a delivery delay exceeding the limit set by the property sib.msgstore.maximumAllowedDeliveryDelayInterval, that action taken was ''{3}'' to "A message with ID {0}  and system message ID {1} was found in destination {2} with a delivery delay exceeding the limit set by the property sib.msgstore.maximumAllowedDeliveryDelayInterval. The action taken was ''{3}'' ".

Also, if there is truly no action for the user to take, might you make the message an informational message instead of a warning message? If it is really a warning message, then you would need a user action that actually has the user do  something.

PR #5408, opened 2018-10-16
Unzip app to temp directory to prevent test issue

brenthdaniel:
We need to unzip snoop.war to a temp location rather than directly to "apps" so that we don't start the app before classes are available. Otherwise, we can get ClassNotFound exceptions until the app updates again.


brenthdaniel:
#build

PR #5413, opened 2018-10-16
Change acceptLicense requirement depending on feature license and runtime

ericglau:
Change acceptLicense to not be required if any of the following are true:
- feature license is EPL
- feature license is standard feature terms
- feature license is restricted feature terms **and** runtime is ND


ericglau:
#build

ericglau:
#build

PR #5414, opened 2018-10-16
reduce references to internals in test code

njr-11:
ConcurrentRxTestServlet currently has a pattern where it directly references ManagedCompletableFuture (product internals) to work around the fact that the test code must run on both Java 11 as well as Java 8 and so is unable to use Java 9+ methods, despite needing to test some of them.  This pull will update the tests to access many of these methods reflectively, hopefully in a somewhat clean way so that we can eventually stop exporting the product internals.

This pull will cover maybe half of the occurrences, although the majority of unique Java 9+ methods that are needed by the tests.


njr-11:
#build

njr-11:
#build

njr-11:
build has 1 test failure which is a duplicate of a preexisting issue in an unrelated test bucket

Issue #5439, opened 2018-10-19
SimpleDateFormat usage in JAX-RS Headers class causing ArrayIndexOutOfBoundsException


andymc12:
An internal customer ran into an ArrayIndexOutOfBoundsException that appears to be the result of multiple threads using the same SimpleDateFormat instance.  According to the JavaDocs, SimpleDateFormat instances are not thread safe and can result in these types of exceptions.  The stack trace reported was:
```
Caused by: java.lang.ArrayIndexOutOfBoundsException
at java.util.Calendar.getDisplayName(Calendar.java:2125)
at java.text.SimpleDateFormat.subFormat(SimpleDateFormat.java:1136)
at java.text.SimpleDateFormat.format(SimpleDateFormat.java:977)
at java.text.SimpleDateFormat.format(SimpleDateFormat.java:947)
at java.text.DateFormat.format(DateFormat.java:356)
at org.apache.cxf.transport.http.Headers.toHttpDate(Headers.java:591)
at org.apache.cxf.transport.http.Headers.headerObjectToString(Headers.java:559)
at org.apache.cxf.transport.http.Headers.copyToResponse(Headers.java:534)
at org.apache.cxf.transport.http.AbstractHTTPDestination.flushHeaders(AbstractHTTPDestination.java:669)
at org.apache.cxf.transport.http.AbstractHTTPDestination.flushHeaders(AbstractHTTPDestination.java:641)
at org.apache.cxf.transport.http.AbstractHTTPDestination$WrappedOutputStream.onFirstWrite(AbstractHTTPDestination.java:836)
at org.apache.cxf.io.AbstractWrappedOutputStream.write(AbstractWrappedOutputStream.java:47)
at org.codehaus.jackson.impl.Utf8Generator._flushBuffer(Utf8Generator.java:1754)
at org.codehaus.jackson.impl.Utf8Generator._writeBytes(Utf8Generator.java:1164)
at org.codehaus.jackson.impl.Utf8Generator._writeFieldName(Utf8Generator.java:457)
at org.codehaus.jackson.impl.Utf8Generator.writeFieldName(Utf8Generator.java:304)
at org.codehaus.jackson.map.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:444)
at org.codehaus.jackson.map.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:150)
... 124 more
```

The fix should be to create a new SimpleDateFormat when needed.

[...]

PR #5440, opened 2018-10-19
Issue 5439: Use new DateFormat for each request

andymc12:
The DateFormat is not thread safe, so using the same one for multiple
requests risks issues like ArrayIndexOutOfBoundsExceptions.

This change removes the Headers.java file, as the single static usage
of SimpleDateFormat was the only unique change from what gets pulled in
from the Apache CXF community. All of the other changes had been
previously contributed back to the open source community.


andymc12:
#build

andymc12:
One failure - unrelated.

andymc12:
This resolves issue #5439.

PR #5442, opened 2018-10-19
ManagedExecutor.newIncompleteFuture

njr-11:
Add implementation for the newIncompleteFuture method of ManagedExecutor, and update the test bucket so that it uses the MicroProfile ManagedExecutor API rather than internals. The test bucket will still require internals in one place, which is where it constructs a CompletableFuture around a non-managed executor, which is not something that MicroProfile does.  That will likely be exposed as an internal interface at some point, but for now, I'll just let the test case reflectively invoke a method to obtain it.


njr-11:
#build

njr-11:
build has 2 test failures which are duplicates of preexisting issues

PR #5455, opened 2018-10-22
Remove ant dependency

ericglau:
InstallKernelMap has a dependency on the library `com.ibm.ws.org.apache.ant`, which doesn't exist in the runtime.  This dependency can be removed, since the code does not actually depend on any of its classes.


ericglau:
#build

PR #5460, opened 2018-10-23
consolidate the old concurrent rx package under mp concurrency and don't export it

njr-11:
What was previously the concurrent.rx package contains the implementation for the concurrent.mp package, so we should combine these and adjust visibility accordingly.  Neither of these packages should be exported.


njr-11:
#build

njr-11:
#build

njr-11:
build has 7 test failures which are duplicates of preexisting issues

PR #5479, opened 2018-10-24
OSGi service component for MP ConcurrencyProvider

njr-11:
Create an OSGi service component that registers Liberty as a ConcurrencyProvider.  Add just enough stubbed out interfaces to be able to obtain a ThreadContext instance (albeit unusable) from ThreadContextBuilder.instance() which will prove that our ConcurrencyProvider implementation was successfully registered.  Include a test case that demonstrates this.


njr-11:
#build

njr-11:
build has 3 failures which are duplicates of pre-existing issues

Issue #5049, opened 2018-09-11
JAX-RS sub-resources seems to loose message body writer


kdaham:
Imagine the following simple root and sub-resource.

```code
@RequestScoped
@Path("/root")
@Produces(MediaType.APPLICATION_JSON)
public class RootResource {
    @Context ResourceContext rc;

    @Path("sub")
    public SubResource getSubResource() {
        return rc.initResource(new SubResource());
    }

    @GET
    @Path("id/{id}")
    public HashMap<String, String> getId(@PathParam("id") String id) {
        HashMap<String, String> map = new HashMap<>();
        map.put("root", id);
        return map;
    }
}

public class SubResource {

    @GET
    @Path("id/{id}")
    public HashMap<String, String> getId(@PathParam("id") String id) {
        HashMap<String, String> map = new HashMap<>();
        map.put("subId", id);
        return map;
    }
}
```
Deploy that to 18.0.0.2 or latest build produces the following result.

$ curl http://localhost:9080/subresource/resources/root/id/1                                                                                                                                                                                                               20:05:26
HTTP/1.1 200 OK
X-Powered-By: Servlet/4.0
Content-Type: application/json
Date: Tue, 11 Sep 2018 18:05:33 GMT
Content-Language: en-US
Content-Length: 12

{"root":"1"}

$ curl http://localhost:9080/subresource/resources/root/sub/id/1                                                                                                                                                                                                           20:05:33
HTTP/1.1 500 Internal Server Error
X-Powered-By: Servlet/4.0
Content-Type: text/plain
Date: Tue, 11 Sep 2018 18:05:42 GMT
Content-Language: en-US
Content-Length: 83
Connection: Close

No message body writer has been found for class java.util.HashMap, ContentType: */*

If I deploy the same .war to wildfly 11 I get the hashmap converted to json.
Ive also tried annotating the sub-class and methods with @Produces but it doesnt seem to help.

WhiteCat22:
Taking a look, Andy was able to reproduce.

WhiteCat22:
[9/13/18 16:30:28:643 CDT] 00000041 id=bd4c1855 com.ibm.ws.jaxrs21.providers.json.JsonBProvider              > isJsonType Entry  
                                                                                                               */*
[9/13/18 16:30:28:643 CDT] 00000041 id=bd4c1855 com.ibm.ws.jaxrs21.providers.json.JsonBProvider              < isJsonType Exit  
                                                                                                               false
[...]

Issue #2084, opened 2018-02-08
CTS: EE8: JAX-RS bucket: NullPointerException during SseEventSink.send


jkoehler22:
[2/8/18 10:05:06:487 EST] 00000060 com.ibm.ws.webcontainer.srt                                  W WARNING: Cannot set status. Response already committed.
[2/8/18 10:05:06:556 EST] 00000060 com.ibm.ws.webcontainer.srt                                  W SRVE8094W: WARNING: Cannot set header. Response already committed.  Stack trace of errant attempt to set header: 
        at com.ibm.ws.webcontainer.srt.SRTServletResponse.setHeader(SRTServletResponse.java:1748)
        at [internal classes].(Unknown Source)

[2/8/18 10:05:09:356 EST] 0000005a g.apache.cxf.cxf.rt.frontend.jaxrs.3.2:1.0.20.master(id=97)] W java.lang.NullPointerException
	at org.apache.cxf.jaxrs.ext.MessageContextImpl.get(MessageContextImpl.java:88)
	at org.apache.cxf.jaxrs.impl.tl.ThreadLocalMessageContext.get(ThreadLocalMessageContext.java:42)
	at org.apache.cxf.jaxrs.provider.JAXBElementProvider.getAttachments(JAXBElementProvider.java:540)
	at org.apache.cxf.jaxrs.provider.JAXBElementProvider.addAttachmentMarshaller(JAXBElementProvider.java:518)
	at org.apache.cxf.jaxrs.provider.JAXBElementProvider.marshal(JAXBElementProvider.java:443)
	at org.apache.cxf.jaxrs.provider.JAXBElementProvider.writeTo(JAXBElementProvider.java:319)
	at org.apache.cxf.jaxrs.sse.OutboundSseEventBodyWriter.writePayloadTo(OutboundSseEventBodyWriter.java:131)
	at org.apache.cxf.jaxrs.sse.OutboundSseEventBodyWriter.writeTo(OutboundSseEventBodyWriter.java:112)
	at org.apache.cxf.jaxrs.sse.OutboundSseEventBodyWriter.writeTo(OutboundSseEventBodyWriter.java:40)
	at com.ibm.ws.jaxrs21.sse.LibertySseEventSinkImpl.send(LibertySseEventSinkImpl.java:98)
	at com.sun.ts.tests.jaxrs.jaxrs21.ee.sse.sseeventsink.MBWCheckResource.sendJAXBElement(MBWCheckResource.java:129)
[...]

PR #5486, opened 2018-10-24
5049 fix sse npe

WhiteCat22:
https://github.com/OpenLiberty/open-liberty/issues/5049

This PR reverts #2084 and performs a null check in MessageContextImpl.java instead.

### DIFF
--- a/dev/com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2/src/org/apache/cxf/jaxrs/ext/MessageContextImpl.java
+++ b/dev/com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2/src/org/apache/cxf/jaxrs/ext/MessageContextImpl.java
@@ -86,11 +86,10 @@ public class MessageContextImpl implements MessageContext {
         }
         if (keyValue.equals("WRITE-" + Message.ATTACHMENTS)) {
             // Liberty Change Start - #5049
-            if (m.getExchange().getOutMessage() == null) {
-                return null;
+            if (m.getExchange().getOutMessage() != null) {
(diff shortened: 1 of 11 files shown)


WhiteCat22:
#build

dev/com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2/src/org/apache/cxf/jaxrs/ext/MessageContextImpl.java
```diff
@@ -0,0 +1,322 @@
...
+                m.getExchange().put("cxf.io.cacheinput", Boolean.FALSE);
+                throw new WebApplicationException(e, 413);
+            } catch (HeaderSizeExceededException e) {
+                throw new WebApplicationException(e, 413);
+            }
+        }
+        if (keyValue.equals("WRITE-" + Message.ATTACHMENTS)) {
+            // Liberty Change Start - #5049
+            if (m.getExchange().getOutMessage() == null) {
+                return null;
```
andymc12: Hmm... if we return null here, it seems like we might be preventing this method from returning something that might be on the "in" message from the code below - what if we did something like:
```
if (keyValue.equals("WRITE-" + Message.ATTACHMENTS) && m.getExchange().getOutMessage() != null) {
    return m.getExchange().getOutMessage().get(Message.ATTACHMENTS);
```

WhiteCat22:
#build try not returning null

WhiteCat22:
build successful in both cases. We are going to err on the side of caution and not return null when there is no outMessage, but continue through MessageContextImpl.get().

PR #5495, opened 2018-10-25
create custom thread context provider for tests

njr-11:
Create a custom thread context provider that we can use for tests.
It should be possible to include it within a shared library, made available to the application.


njr-11:
#build

njr-11:
build has 3 failures which are duplicates of preexisting issue in unrelated componentry

PR #5520, opened 2018-10-26
Ignore port conflict messages on server stop

brenthdaniel:
The test case doesn't need to actually use the ports it defines, so ignore any port conflict errors.


brenthdaniel:
#build

brenthdaniel:
#build

PR #5526, opened 2018-10-28
Fixes #4387 ignore wlp user dir for runnable

### LINKED ISSUE #4387: Runnable JAR execution fails when WLP_USER_DIR env var is set to "other" location with CWWKE0005E
With WLP_USER_DIR set, running the server as executable JAR gives:

```
$ java -jar target/batch-bonuspayout-application.jar
Extracting files to .....\wlpExtract\batch-bonuspayout-application_621724882767969\wlp
Successfully extracted all product files.
CWWKE0005E: The runtime environment could not be launched.
CWWKE0031E: Specified server BonusPayout does not exist; use the --create option to create a new server. serverPath: /some/other/location/servers/BonusPayout
```

This is pretty straightforward to recreate:

1. Build a server as a runnable JAR 
    * One way is to take a Maven build such as [the BonusPayout sample](https://github.com/WASdev/sample.batch.bonuspayout) and do `mvn package`
2. Set WLP_USER_DIR to some other location:
> export WLP_USER_DIR=/some/other/location
3. Run server with
>    java -jar target/batch-bonuspayout-application.jar

Not sure if there's any complicated interplay between WLP_USER_DIR, and WLP_OUTPUT_DIR, the latter of which is [documented here](https://www.ibm.com/support/knowledgecenter/en/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/rwlp_setup_jarserver.html).

At a naive, first take, though, it seems that the user dir is going to effectively be set to the extracted location, so what should it matter that a WLP_USER_DIR is set in the environment.   It seems it should maybe be ignored.

scottkurz:
Fixes #4387

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/cmdline/Utils.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/cmdline/Utils.java
@@ -64,7 +64,7 @@ public class Utils {
             File resultDir = null;
 
             // If this property is set we want to ignore the WLP_USER_DIR env var and use the <install>/usr dir
-            if (Boolean.getBoolean(BootstrapConstants.LOC_PROPERTY_INSTANCE_DIR_IGNORE_ENV)) {
+            if (Boolean.getBoolean(BootstrapConstants.LOC_PROPERTY_IGNORE_INSTANCE_DIR_FROM_ENV)) {
                 resultDir = getDefaultInstallBasedUserDir();
             } else {
                 userDirLoc = System.getenv(BootstrapConstants.ENV_WLP_USER_DIR);
@@ -72,7 +72,7 @@ public class Utils {
                     resultDir = new File(userDirLoc);
                 } else {
 
-                    // PI20344: Check if the Java property is set, which is the normal case when
+                    // Check if the Java property is set, which is the normal case when
                     // the server is embedded; i.e. they didn't launch it from the command line.
                     userDirLoc = System.getProperty(BootstrapConstants.ENV_WLP_USER_DIR);
                     if (userDirLoc != null) {
--- a/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/cmdline/Utils.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/cmdline/Utils.java
@@ -62,26 +62,39 @@ public class Utils {
         // when the server is started from the command line.
         if (userDir.get() == null) {
             File resultDir = null;
-            userDirLoc = System.getenv(BootstrapConstants.ENV_WLP_USER_DIR);
-            if (userDirLoc != null) {
-                resultDir = new File(userDirLoc);
-            } else {
 
-                // PI20344: Check if the Java property is set, which is the normal case when
-                // the server is embedded; i.e. they didn't launch it from the command line.
-                userDirLoc = System.getProperty(BootstrapConstants.ENV_WLP_USER_DIR);
+            // If this property is set we want to ignore the WLP_USER_DIR env var and use the <install>/usr dir
+            if (Boolean.getBoolean(BootstrapConstants.LOC_PROPERTY_INSTANCE_DIR_IGNORE_ENV)) {
+                resultDir = getDefaultInstallBasedUserDir();
+            } else {
+                userDirLoc = System.getenv(BootstrapConstants.ENV_WLP_USER_DIR);
                 if (userDirLoc != null) {
                     resultDir = new File(userDirLoc);
                 } else {
-                    File installDir = Utils.getInstallDir();
-                    if (installDir != null) {
-                        resultDir = new File(installDir, "usr");
+
+                    // PI20344: Check if the Java property is set, which is the normal case when
+                    // the server is embedded; i.e. they didn't launch it from the command line.
+                    userDirLoc = System.getProperty(BootstrapConstants.ENV_WLP_USER_DIR);
+                    if (userDirLoc != null) {
+                        resultDir = new File(userDirLoc);
+                    } else {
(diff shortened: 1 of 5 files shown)


scottkurz:
#build

scottkurz:
TODOs:
1. Look for other places in the code reading WLP_USER_DIR.
2. Maybe should set new prop unconditionally, not just in the 1-JVM case.

scottkurz:
#build

scottkurz:
Some timeout tests, plus non-timeout tests in:
  - com.ibm.ws.webserver.plugin.runtime_fat FAT tests
  - com.ibm.ws.jsonb_fat

Both these failures seem to reflect issues with DS dependency graphs, which I don't see I could have affected.   Going to rebase and re-run.

scottkurz:
#build

dev/wlp.lib.extract_fat/fat/src/wlp/lib/extract/PackageRunnableTest.java
```diff
@@ -49,24 +49,30 @@
     /*
      * return env as array and add WLP_JAR_EXTRACT_DIR=extractDirectory
      */
-    private static String[] runEnv(String extractDirectory) {
+    private static String[] runEnv(String extractDirectory, boolean useDummyUserDir) {
```
shighbar: Some comments on when to useDummyUserDir might be helpful. If we have this Boolean I'm wondering why not have a test where we test the true scenario as well?

dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtractRun.java
```diff
@@ -339,6 +339,8 @@ private static int runServerInline(String extractDirectory,
 
         Properties props = System.getProperties();
 
+        // Setting this tells us the boostrap to ignore any WLP_USER_DIR that happens to be set and use the <install>/usr from the extracted server install.
+        props.setProperty("wlp.prefer.default.user.dir", "true");
```
shighbar: Is there a case where we would want some other directory in a self extracted run? This locks that in - which may be ok in this scenario but wondering if there are cases where it might make sense to have some other dir even in self runable jar?

dev/wlp.lib.extract_fat/fat/src/wlp/lib/extract/PackageRunnableTest.java
scottkurz: Thanks for taking a look.

I added some comments in my updated (and rebased) PR .

I actually did refactor the `testRunnableJar` test so the net result is both true/false boolean values are passed through this method (both are passed to the `executeTheJar` method as well).

Also this test class has some pretty heavy-weight methods with forking processes and the cleanup involved (including a sleep for 30 seconds)... which makes me think this isn't going to be one where we add too many more different test methods to.

dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtractRun.java
scottkurz: @NottyCode  any opinions here?

scottkurz:
#build

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/cmdline/Utils.java
```diff
@@ -62,20 +62,23 @@ public static File getUserDir() {
...
+            } else {
+                userDirLoc = System.getenv(BootstrapConstants.ENV_WLP_USER_DIR);
                 if (userDirLoc != null) {
                     resultDir = new File(userDirLoc);
                 } else {
-                    File installDir = Utils.getInstallDir();
-                    if (installDir != null) {
-                        resultDir = new File(installDir, "usr");
+
+                    // PI20344: Check if the Java property is set, which is the normal case when
```
NottyCode: Please remove PI20344.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/BootstrapConstants.java
```diff
@@ -99,6 +99,14 @@
     public static final String LOC_INTERNAL_LIB_DIR = "wlp.lib.dir";
 
     public static final String LOC_PROPERTY_INSTANCE_DIR = "wlp.user.dir";
+
+    /**
+     * Used (especially in the runnable JAR case at present) to specify that we should ignore WLP_USER_DIR set from the environment, and use
+     * something else as user dir. E.g. we might use: "${wlp.install.dir}/usr"
+     */
+    public static final String LOC_PROPERTY_INSTANCE_DIR_IGNORE_ENV = "wlp.user.dir.ignore.env";
```
NottyCode: I think ignore would be better earlier since it would match how you read it. So wlp.ignore.user.dir.env makes more sense to me since  it says 'ignore user dir from environment'

scottkurz: I could do that or:  `wlp.ignore.user.dir.from.env`

scottkurz: Used `wlp.ignore.user.dir.from.env`

scottkurz:
Final name for new property:   `wlp.ignore.user.dir.from.env`

scottkurz:
As far as documentation, I'm thinking it doesn't need any?   

The OL doc:

https://github.com/OpenLiberty/docs/blob/master/ref/config/serverConfiguration.adoc
https://github.com/OpenLiberty/docs/blob/master/ref/commands/server/server-commands.adoc
https://github.com/OpenLiberty/docs/blob/master/ref/general/runnablejarfiles.adoc

doesn't go into this level of detail.   

From the perspective of the typical developer, the use of this property is basically hidden as an implementation detail.   

I suppose an embedder could make use of this though?

PR #5531, opened 2018-10-29
Set Default Read Timeout to 1min

### LINKED ISSUE #5341: Modify default ldapRegistry-3.0 read timeout to be 1 minute
The ldapRegistry-3.0 uses an infinite read timeout. This can lead to user requests hanging for long periods of time. This consumes worker threads and contexts from the LDAP registry's context pool, both resulting in poor performance.

Issue #3383 was completed to allow the customer to specify their own timeout.

Modify the default read timeout to be 1 minute to match both the JNDI search timeout and the connect timeout.

See issue #4949 for more details on the decision to make this change.

agiridharan:
fixes #5341.

Updated default readTimeout in metatype.xml. Updated LDAPRegistry ReadTimeout constant to be set to 1 min. Added Fat Test to test for Default Value

### DIFF
--- a/dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/context/ContextManager.java
+++ b/dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/context/ContextManager.java
@@ -88,10 +88,10 @@ public class ContextManager {
     private static final int DEFAULT_PREF_POOL_SIZE = 3;
 
     /** The default connect time limit - 1 minute. */
-    private static final long DEFAULT_CONNECT_TIMEOUT = 600000;
+    private static final long DEFAULT_CONNECT_TIMEOUT = 60000;
 
-    /** The default read time limit - infinite. */
-    private static final long DEFAULT_READ_TIMEOUT = 0;
+    /** The default read time limit - 1 minute. */
+    private static final long DEFAULT_READ_TIMEOUT = 60000L;
 
     /** Key to use for accessing the active URL from the JNDI environment table. */
     private static final String ENVKEY_ACTIVE_URL = "_ACTIVE_URL_";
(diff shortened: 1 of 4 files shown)


agiridharan:
Set default ReadTimeout to be 1 min. Modified LDAPRegistryContextPoolNoConfigTest.java to check if ReadTimeout was being set to the default 1 min

agiridharan:
#build

dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/context/ContextManager.java
```diff
@@ -1389,7 +1389,7 @@ public Void run() {
      * @param prefPoolSize The preferred context pool size. Not required when <code>enableContextPool == false</code>.
      * @param maxPoolSize The maximum context pool size. A size of '0' means the maximum size is unlimited. Not required when <code>enableContextPool == false</code>.
      * @param poolTimeOut The context pool timeout in milliseconds. This is the amount of time a context is valid for in
-     *            the context pool is valid for until it is discarded. Not required when <code>enableContextPool == false</code>.
+     *           the context pool is valid for until it is discarded. Not required when <code>enableContextPool == false</code>.
```
jvanhill: Might want to undo this change. Appears to be a spacing change.

dev/com.ibm.ws.security.wim.adapter.ldap_fat/fat/src/com/ibm/ws/security/wim/adapter/ldap/fat/LDAPRegistryContextPoolNoConfigTest.java
```diff
@@ -149,6 +149,10 @@ public void testCustomContextNoConfig() throws Exception { // for issue
         tr = "CacheResultSizeLimit: 2000";
         errMsgs = server.findStringsInLogsAndTrace(tr);
         assertFalse("Should have found, " + tr, errMsgs.isEmpty());
+
```
jvanhill: I don't think this belongs here. We probably want to put it in its own FAT or something more closely related. This FAT deals with the context pool. Additionally alls this tests is whether some trace is printed out and not that it mirrors the value passed in to the read timeout.

I would write a test that tries to call out to an LDAP server that is unavailable and verify that you get a connection timeout after 60 seconds. I would then modify the the read timeout dynamically to something short and verify that again you get a connection timeout, but this time much quicker.

jvanhill: Actually, this is read timeout, so you might have to just open your own connection and never respond to LDAP requests. That should get you a read timeout I imagine.

PR #5533, opened 2018-10-29
rename groupConfiguration to match xml element

ryanesch:


ryanesch:
This doesn't change any functionality. I renamed a class to reflect the xml element.

PR #5535, opened 2018-10-29
Use non-JDK version of JAX-B on Java 9+

aguibert:
Resolves an issue where CXF attempts to use ASM to create a synthetic extension of one of the JDK's JAX-B impl classes and uses an incorrect package name.

Here we update the CXF code to use non-internal package for Java 9+ (see 2nd commit)


aguibert:
#build

aguibert:
#build

PR #5084, opened 2018-09-14
Revert "Update ClassLoadingService to depend on PermissionManager instead of the other way around for issue #1438"

websterc87:
Reverts OpenLiberty/open-liberty#4854

I believe this item is responsible for high frequency intermittent current being seen in Open and Commercial liberty builds in com.ibm.ws.persistence.utility.fat.DDLGenerationUtilityTest.testScriptSuccess (RTC Defect 258096)

The reason i believe this item is likely responsible is that in the last 14 days the issue only started occuring 4 days ago, where it was only seen in PR #4854.  until it was pushed in the continuous then it start to infect the other builds corresponding to when this change was being adopted.

It would be extremely coincindental for this progression if that work item was not the cause.

"build":"Liberty Personal Build - EBC - teddyjtorres-4854-20180910-0236",
"build":"Liberty Personal Build - EBC - teddyjtorres-4854-20180910-1515",
"build":"LAOS Continuous build - 20180912-0725",
OL PUSH
         "build":"Liberty Personal Build - EBC - aguibert-13365-20180913-0044"
		 "build":"Liberty - Mac EBC - 20180913-0117",
		 "build":"Liberty Full Fat - Java 2 security -  EBC - 20180913-0322",
		 "build":"Liberty Personal Build - EBC - tkburroughs-5068-20180913-0410",
		 "build":"LAOS Continuous build - 20180913-0326",
		 "build":"Liberty Continuous Build - EBC - 20180913-1006",
         "build":"LAOS Continuous build - 20180913-1007",
         "build":"Liberty - Mac EBC - 20180913-1310",
         "build":"Liberty Personal Build - EBC - pmd1nh-5060-20180913-1424",
		 "build":"Liberty Personal Build - EBC - tkburroughs-5068-20180913-1601",
		 "build":"Liberty Continuous Build - EBC - 20180913-1648",
		 "build":"LAOS Continuous build - 20180913-1649",
         "build":"Liberty Personal Build - EBC - andymc12-5066-20180913-1731",
         "build":"Liberty Personal Build - EBC - billw1024-5079-20180913-1955",
         "build":"Liberty Personal Build - EBC - brutif-4981-20180913-2122",
		 "build":"LAOS Continuous build - 20180913-2328",
		 "build":"Liberty Personal Build - EBC - pmd1nh-5060-20180914-0028",
		 "build":"Liberty Personal Build - EBC - jdmcclur-5046-20180914-0143",
         "build":"Liberty Official Release Build - Static - 20180914-0300",

PR #5540, opened 2018-10-30
Move PermissionManager and related artifacts to classloading bundle to fix temporal dependency for issue #1438

teddyjtorres:
For issue #1438


teddyjtorres:
@OpenLiberty/message-reviewer  the message files were moved from one bundle to another. There are no new messages or message changes.

teddyjtorres:
#build

teddyjtorres:
Note to reviewers: An earlier solution in https://github.com/OpenLiberty/open-liberty/pull/4854 that used dynamic reference counting was not appropriate since dependents of the ClassLoadingService would get notified of changes after activation causing unexpected behaviors. For example, see https://github.com/OpenLiberty/open-liberty/pull/5084 where that pull request was reverted because it caused the DDLGenerationUtilityTest to fail when the datasources were no longer available because of the services coming and going.

In addition, the creation of protection domains is only for the usage of the ClassLoadingService and therefore it is appropriate that the PermissionManager also resides in the same bundle. Further enhancements could be done to refactor the code that uses the PermissionManager (DeployedAppInfoBase, DeployedAppInfoFactoryBase, ResourceAdapterService, etc.) to obtain the ProtectionDomain to not do that and instead have the ClassLoadingService obtain the ProtectionDomain itself from the PermissionManager.

teddyjtorres:
@andymc12 Hi Andy. Do you have any comments about this code move? This is a rework of #4854 you already reviewed and that could not be used. Thank you for your review.

teddyjtorres:
JASPIFormLoginAuditTest error in PB is not related to the code being delivered and it has an internal defect already opened for it. The Full Java 2 Security build shows that the jca_derbyra FAT tests are no longer failing and that there has been a reduction in the total number of failures seen.

PR #5547, opened 2018-10-30
use ConcurrencyManager to load context providers separately per class loader

njr-11:
Start experimenting with ConcurrencyManager from the MP Concurrency spec.  The design of the spec around ConcurrencyManager needs some improvement and it will help to experiment with how well it fits in to the Liberty implementation.  To start out with, most of the non-essential methods will be left as a no-op or exception, and we won't be concerned with clearing cached ConcurrencyManager/provider lists when a class loader goes away, although that will eventually need to be added.


njr-11:
#build

PR #5554, opened 2018-10-31
Use FATRunner for all Fault Tolerance tests

### LINKED ISSUE #2737: MP Fault Tolerance: Use RunWith(FATRunner) in faulttolerance.cdi_fat and cdi_fat_metrics suites
Need to make sure to use the FATRunner and "standard" FAT framework where possible to ensure that functions such as RepeatTests work propertly and that tests marked as FULL mode only run in full mode

Azquelt:
Some of the Fault Tolerance tests weren't using FATRunner where they should be.

Fixes #2737


Azquelt:
#build

PR #5582, opened 2018-11-01
test case with custom thread context provider

njr-11:
Basic test of ThreadContextBuilder, including one built-in container context type (APPLICATION) and one custom context provider type (TestContextTypes.STATE). Build a MicroProfile ThreadContext instance and use it to contextualize a task based on current context of the servlet thread. Change the custom "State" context of the servlet thread and run the contextualized task, verifying that the originally captured context is used. After the task ends, verify that the custom "State" context on the servlet thread has been restored to what we most recently set. Submit the same contextual task to an asynchronous unmanaged executor thread (lacking context) and verify that it runs with the context that was originally captured.


njr-11:
#build

njr-11:
Build has 15 failures which look like duplicates of preexisting issues.  However, given that this commit was also included in under #5596 which has a better build, I'll do the merge under that pull instead.

PR #5595, opened 2018-11-01
Issue #5585: Fix EJB ScheduleExpression serialization

### LINKED ISSUE #5585: EJB timer ScheduleExpression serialization incompatibility
The serializable class javax.ejb.ScheduleExpression changed in an incompatible way from the EJB 3.1 API jar provided and the EJB 3.2 API jar, so instances that were serialized and stored persistently with the ejb-3.1 feature cannot be properly deserialized after moving to the ejb-3.2 feature.

Since ScheduleExpression is serialized and stored in the EJB timer database for persistent calendar based EJB timers, this means that persistent calendar based EJB timers that were created using ejb-3.1 will not function properly if the application is migrated to the ejb-3.2 feature.

Specifically, the names of the instance variables in the class were changed (i.e. second to second_). Since the class uses default serialization, the names of the fields are written to the serialized bytes, and therefore do not match up with the new version of the class. The serial version UID does match, so the new class instance is created, but all fields are left default (null/0).

tkburroughs:
Provide a custom implementation of ScheduleExpression that handles the
serialization incompatibility between the RI implementation for EJB 3.2 and
the version used for EJB 3.1. The serial version identifier is the same for
both implementations, however the instance variables have different names,
so default serialization is not able to restore the values when deserializing
bytes from the older implementation using the RI implementation.

Since an instance of ScheduleExpression is serialized and stored in a
database for persistent calendar based EJB timers, supporting the
deserialization of both versions of the class is needed for applications
migrating from EJB 3.1 to EJB 3.2.

fixes #5585


tkburroughs:
#build

tkburroughs:
Test failures are unrelated to EJB.

PR #5596, opened 2018-11-01
prerequisite thread context providers

njr-11:
Complete the code that allows for and enforces the ordering of prerequisite thread context providers, and add a test case to verify.


njr-11:
#build

njr-11:
Build has 5 test failures which are duplicates of preexisting issues

PR #5601, opened 2018-11-01
Social oidc login updates to do OP discovery

arunavemulapalli:


arunavemulapalli:
pb looks good except 3 unrelated failures.

PR #5612, opened 2018-11-02
Refactor OpenJPA and Eclipselink specific Introspection

jgrassel:

Refactor OpenJPA introspection, add Eclipselink introspection.

### DIFF
--- a/dev/com.ibm.ws.jpa.container.ormdiagnostics/src/com/ibm/ws/jpa/diagnostics/AbstractIntrospection.java
+++ b/dev/com.ibm.ws.jpa.container.ormdiagnostics/src/com/ibm/ws/jpa/diagnostics/AbstractIntrospection.java
@@ -1,0 +1,298 @@
+/*******************************************************************************
+ * Copyright (c) 2018 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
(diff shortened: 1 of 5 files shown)


jgrassel:
#build

dev/com.ibm.ws.jpa.container.ormdiagnostics/src/com/ibm/ws/jpa/diagnostics/AbstractIntrospection.java
```diff
@@ -0,0 +1,298 @@
...
+    /*
+     * Utility methods used for introspecting JPA Persistence Provider Implementation Data Structures
+     */
+    protected static Object reflectObjValue(Object o, String field) throws Exception {
+        final Class<?> c = o.getClass();
+        final Field f = findField(c, field); //        c.getField(field);
+        if (f == null) {
+            return null;
+        }
+        final boolean accessible = f.isAccessible();
```
tkburroughs: Note: java.lang.Class.getDeclaredField will return a copy of the Field.... so changes made here with setAccessible are only visible here.... and do not need to be undone.

If it is possible to cache this Field, that would be nice, but probably not important if this won't run very often.

Will this run in a doPriv? or an a privileged thread? If running as regular trace, you may fail Java 2 security authorization.

jgrassel: I believe the introspection dump threads are privileged threads, I've been running the FATs (which enable java2sec) and the dump has not been failing with ACE involving reflection (only ACE is from the use of the jdk provided JAXB provider impl)

jgrassel:
Failing fats not related to updated JPA diagnostic code.

PR #5617, opened 2018-11-04
Temporarily downgrade feature suggestion from WARN to INFO

aguibert:
Delivering the feature suggestion message at WARN level has caused many test failures in the Java 11 builds (because a FAT will fail if we find unexpected errors/warnings in server logs).

This delivery will be followed up with another delivery that will move the message back to WARN (after the expected warnings have been mapped out for the effected FATs).


aguibert:
#build

PR #5632, opened 2018-11-05
keystore update should be immediate

arunavemulapalli:


ayoho:
Is this pull request still needed?

ayoho:
@arunavemulapalli Is this pull request needed?

ayoho:
@arunavemulapalli Checking again - is this pull request needed?

arunavemulapalli:
There were improvements in keystore service/management and so we don't need this change in the components that depend on keystore service

PR #5656, opened 2018-11-07
tests for dependent stage actions running on thread that invokes complete or get

njr-11:
Add some tests that aim to get completion stage actions running on the thread that invokes complete or get.  We cannot force the latter to happen consistently, but should be able to at least write a test that encourages the possibility and validates thread context propagation in any case.


njr-11:
#build

PR #5659, opened 2018-11-07
prerequisite context removed from the MP concurrency spec

njr-11:
The concept of prerequisite context has been removed from the MicroProfile Concurrency spec.  While we don't have a refresh of the spec binary yet, we can prepare for this by removing our implementation and updating our tests.


njr-11:
#build

PR #5672, opened 2018-11-08
Add java vendor to service properties

utle:
Add java vendor to service properties so we can target an OSGI service with Java version and vendor


utle:
#build

PR #5682, opened 2018-11-08
error paths for managed executor builder

njr-11:
Ensure that ManagedExecutorBuilder methods are raising proper errors per the spec, and that even after errors are raised, the ManagedExecutorBuilder continues to be usable, retaining previous values that were validly/successfully set on it.  Also, clean up an unused test feature from the tests, clean up compiler warnings in the tests, and add coverage of newIncompleteFuture invoked on a CompletableFuture from a managed executor because that scenario isn't previously covered elsewhere.


njr-11:
#build

PR #5722, opened 2018-11-12
Add support to server package for tar & tar.gz

NottyCode:
Support tar and tar.gz archive formats.


NottyCode:
#build this please Libby

PR #5723, opened 2018-11-12
Add feature dependency on appSecurity-2.0 to mpJwt-1.0

### LINKED ISSUE #5720: Feature dependency issue with appSecurity
There was a change in Liberty and we had to add appSecurity ourselves. 

Deployed the application on `wlp-webProfile7 18.0.0.3` using Maven. I don't have any errors and it is running fine. But when the same application is deployed on `wlp-webProfile8 18.0.0.3` using Maven, I see the below error.
 
```
[ERROR   ] CWWKS5604E: A JsonWebToken Principal can't be injected because one is not available. Protect the requesting resource so authentication occurs before the resource is accessed.
```
 
This error is resolved by adding `appSecurity-2.0` in server.xml.

How are these two different and why should we add the `appSecurity-2.0` here?

ayoho:
Should fix #5720


ayoho:
#build

Issue #5726, opened 2018-11-12
Limit concurrent number of threads installing/resolving/starting gateway bundles


tjwatson:
There have been cases where Liberty can hit a performance issue with massive amounts of applications installed.  It looks like it may be hit by the Equinox bug [541048](https://bugs.eclipse.org/bugs/show_bug.cgi?id=541048).  Liberty could work around this issue by limiting the number of concurrent threads installing/resolving/starting gateway bundles for the applications.

PR #5727, opened 2018-11-12
Limit the number of threads to 10 for install/resolve/start

tjwatson:
For issue #5726 
#build


tjwatson:
This is no longer needed.

PR #5728, opened 2018-11-12
Issue #5099: Allow nonpersistent EJB timers

### LINKED ISSUE #5099: EJB timer requires datasource definition even for persistent = false

tkburroughs:
When the ejbPersistentTimer feature is enabled, but not configured
with a datasource, applications that do not use persistent timers
should be allowed to staert. There are two scenariows addressed

1 - Automatically created timers

a) if persistent=true; a new error will be logged in addition to
   the current exception
b) if persistent=false; the application will start; no messages or errors

2 - Programmtic timer method exists

It is not obvious if programmatic timers will be persistent, so
in this scenario, a single informational message will be logged
to indicate that even though the ejbPersistentTimer feature has
been enabled, it will not function. Applications will start, but
will fail if they attempt to create a persistent timer.

fixes #5099


tkburroughs:
#build

tkburroughs:
#build

tkburroughs:
@OpenLiberty/message-reviewer Please review.

dev/com.ibm.ws.ejbcontainer/resources/com/ibm/ws/ejbcontainer/osgi/internal/resources/EJBContainerMessages.nlsprops

tkburroughs:
#build

PR #5730, opened 2018-11-12
clear microprofile context types when unspecified and clean up tracing of context

njr-11:
When the user uses the old server-configuration based method of defining contextServer/managed executors, we should be clearing/restoring MicroProfile-supplied context types because the user's tasks/actions will not expect to see these new types remain on threads.  This will also enforce a more predictable, consistent behavior because these context types will always be cleared rather than sometimes having one value vs another or not being present at all.

Also, looking of the trace output for context propagation, it has so many trace points that it is nearly unreadable.  In this pull, I'm reducing/improving trace to more concisely show the most important and useful information.


njr-11:
#build

PR #5732, opened 2018-11-12
Add test delay to ensure update WAR file doesn't have the same timestamp

brenthdaniel:
In the app manager's auto extract FAT tests, it's possible for an updated WAR file to have the same timestamp as the old WAR file. This causes the app manager to believe that the extracted directory doesn't need to be updated. A short delay in the test will prevent this (though we should eventually get around to using a better mechanism than timestamps in the app manager code.)


brenthdaniel:
#build

PR #5750, opened 2018-11-13
remove difficult-to-reach paths for static methods of ManagedCompletableFuture

njr-11:
Static methods on ManagedCompletableFuture are not required by spec and would be extremely difficult for application code to reach.  It would need to access these static methods (probably via reflection) from the impl class of CompletableFuture.  Because these are not spec requirements, we should do away with this code path and just raise UnsupportedOperationException.  This pull cleans up several of these paths.  It might be possible to go further, but that will need to be done separately.


njr-11:
#build

njr-11:
Build has 2 test failures which are preexisting issues

PR #5756, opened 2018-11-14
support wildcards

### LINKED ISSUE #5755: FeatureReplacementAction should support removing feature by wildcards
We have four different versions of microprofile config, which has resulted in eight different feature action classes to set the correct version of mpConfig and java EE features. 

Allowing the removal of features by wildcard would make it possible to condense these duplicate classes.

benjamin-confino:
fixes #5755

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
+++ b/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
@@ -16,6 +16,7 @@ import static org.junit.Assert.assertTrue;
 import java.io.File;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Scanner;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
```diff
@@ -347,6 +363,20 @@ private static String getReplacementFeature(String originalFeature, Set<String>
         return null;
     }
 
+    private static boolean removeWildcardFeature(Set<String> features, String removeFeature) {
+        String matcher = removeFeature.substring(0, removeFeature.length() - 1); //This assumes the feature string is in the format "feature-*"
```
aguibert: I don't think we should make this assumption that wildcards are always `feature-*`, mainly because if I do something slightly different but valid like `feature*` or `feature-1.*` it would be tricky to figure out why the features are not being removed.

PR #5768, opened 2018-11-14
implement ThreadContext.withContextCapture(CompletableFuture)

njr-11:
Implement the withContextCapture method that takes a CompletableFuture on ThreadContext.  A test case for this method, although passing, is not included under this pull in order to avoid a conflict with a test bucket rename that another developer is working on.


njr-11:
#build

PR #5771, opened 2018-11-14
Fix span name for http-path

### LINKED ISSUE #5767: mpOT-1.2 Fix http-path span name
If the class has `@Path("/foo")` but the method doesn't have the `@Path` annotation, the current path generated will be `get:/foo/`.  It should be `get:/foo` without the trailing slash.

fmhwong:
Fixes #5767


fmhwong:
#build

fmhwong:
The messaging client container failure in the FAT is not related to open tracing.

Issue #5772, opened 2018-11-14
AppClassLoader does not correctly handle null response from ClassFileTransformers 


andymc12:
In `AppClassLoader`, the `findClass` method will acquire the bytes from the underlying class container and will then pass those bytes to registered `ClassFileTransformers`.  Per the Java 7 and 8 javadoc for `ClassFileTransformer`, the `transform` method should return null when no transformation should occur.  However, the `findClass` code will simply replace the original bytes loaded with the return from the transformer - which can result in a NullPointerException - and prevents the class from being loaded.

The NPE would look something like this:
```
java.lang.NullPointerException
	at com.ibm.ws.classloading.internal.AppClassLoader.definePackageAndClass(AppClassLoader.java:329)
	at com.ibm.ws.classloading.internal.AppClassLoader.findClass(AppClassLoader.java:304)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at com.ibm.ws.classloading.internal.AppClassLoader.findOrDelegateLoadClass(AppClassLoader.java:482)
	at com.ibm.ws.classloading.internal.AppClassLoader.loadClass(AppClassLoader.java:443)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        at ...
```

PR #5774, opened 2018-11-14
Issue 5772: Check for null response from class file transformer

andymc12:
ClassFileTransformer.transform(...) should return null if it does not transform the byte code passed in.  The AppClassLoader needs to check for this before processing the null bytes.

Adds tests for cases where transformer returns null, returns the same bytes passed-in, and returns transformed bytes.


andymc12:
#build

andymc12:
#build

andymc12:
This resolves issue #5772

PR #5790, opened 2018-11-15
test case for withContextCapture(CompletableFuture) and honor unchanged context

njr-11:
Add a test case for ThreadContext.withContextCapture(CompletableFuture)
and update the code to honor the configured unchanged context.


njr-11:
Build has 2 test failures which are duplicates of pre-existing issues.

PR #5795, opened 2018-11-16
add test for ThreadContext.withContextCapture(CompletionStage)

njr-11:
Need to add a test for ThreadContext.withContextCapture(CompletionStage).
In addition to testing that Liberty thread pool and context propagation are used, this should test that the resulting CompletionStage and its dependent stages do not allow CompletableFuture methods that are not present on CompletionStage (conforms to minimalCompletionStage of Java SE 11 CompletableFuture)


njr-11:
#build

PR #5796, opened 2018-11-16
Initial version of ManagedExecutor CDI injectio

aguibert:
Initial pass on CDI injection of ManagedExecutor.

Allow applications to use just `@Inject` or `@Inject @ManagedExecutorConfig(...)` to obtain `ManagedExecutor` instances.


aguibert:
#build

aguibert:
#build

aguibert:
#build

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ManagedExecutorProducer.java
```diff
@@ -0,0 +1,53 @@
...
+
+    // A new instance of this class will be constructed each time the producer is
+    // called so we need to keep the config->executor map in a static variable
+    private static final Map<ManagedExecutorConfigContainer, ManagedExecutor> configMap = new ConcurrentHashMap<>();
+
+    @Produces
+    public ManagedExecutor produceManagedExecutor(InjectionPoint point) {
+        ManagedExecutorConfig configAnno = point.getAnnotated().getAnnotation(ManagedExecutorConfig.class);
+        ManagedExecutorConfigContainer config = ManagedExecutorConfigContainer.get(configAnno);
+        ManagedExecutor exec = configMap.computeIfAbsent(config, this::createManagedExecutor);
```
njr-11: I wonder if the createManagedExecutor function & ManagedExecutorBuilderImpl.build is consistent with the intent of computeIfAbsent's requirement that it be "short and simple".  I suppose it might become simpler as the code evolves.

dev/com.ibm.ws.concurrent.mp_fat/publish/servers/MPConcurrentCDITestServer/server.xml
```diff
@@ -0,0 +1,33 @@
...
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+   
+    Contributors:
+        IBM Corporation - initial API and implementation
+ -->
+<server>
+
+	<featureManager>
```
njr-11: The indentation is strange here. Top level element names don't line up, and neither do that last 2 child elements.

dev/com.ibm.ws.concurrent.mp_fat/test-applications/MPConcurrentCDIApp/src/concurrent/mp/fat/cdi/web/MPConcurrentCDITestServlet.java
```diff
@@ -0,0 +1,222 @@
...
+    @Inject
+    @ManagedExecutorConfig(maxAsync = 2, maxQueued = 2)
+    ManagedExecutor max2;
+
+    @Inject
+    @ManagedExecutorConfig(propagated = {}, cleared = { ThreadContext.APPLICATION })
+    ManagedExecutor noAppCtx;
+
+    @Inject
+    @ManagedExecutorConfig(propagated = { ThreadContext.APPLICATION, ThreadContext.TRANSACTION }, cleared = {})
```
njr-11: Liberty will not (at least initially) support propagating TRANSACTION context.  If this is permitted, we need to figure out where error checking is missing.  The same applies to the next injection point as well.
Also, correct spelling for field name is propagated.

dev/com.ibm.ws.concurrent.mp_fat/test-applications/MPConcurrentCDIApp/src/concurrent/mp/fat/cdi/web/MPConcurrentCDITestServlet.java
```diff
@@ -0,0 +1,222 @@
...
+        assertEquals(propogatedAB, propogatedBA);
+    }
+
+    @Test
+    public void testBasicMEWorks() throws Exception {
+        CompletableFuture<String> cf = noAnno.supplyAsync(() -> {
+            try {
+                System.out.println("testBasicMEWorks: Performing lookup of 'foo'");
+                // TODO: This confirms JEE context is on the thread, need to find an operation that
+                // confirms 'application' context is on the thread
```
njr-11: Application context also includes thread context class loader, so that could be used as well.  If we do that, JNDI lookups could be dropped and we could disable the jndi-1.0 feature, which would be nice because the other test server has it enabled.

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ManagedExecutorProducer.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
+public class ManagedExecutorProducer {
+
+    private static final TraceComponent tc = Tr.register(ManagedExecutorProducer.class);
+
+    // A new instance of this class will be constructed each time the producer is
+    // called so we need to keep the config->executor map in a static variable
```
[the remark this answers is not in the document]
aguibert: good catch, I'll add a TODO for now and be sure to add test for this in a follow-up PR.  However, this made me realize that the sharing of identical config across various injection points is probably undesirable in most situations.

If I was writing an app and say I fire async events for payments and notifications, and I happen to configure them the same like this:

```java
public class PaymentProcessor {
    @Inject
    @ManagedExecutorConfig(maxAsync = 5)
    ManagedExecutor paymentExectuor;
}

public class NotificationProcessor {
    @Inject
    @ManagedExecutorConfig(maxAsync = 5)
    ManagedExecutor notificationExecutor;
}
```

I would think that I'm allocating up to 5 threads for payments, and up to 5 threads for notifications.  However, since these ManagedExecutors are the same instance, I would have a max of 5 threads between Payments+Notifications.  Looking at the code, this is not intuitive at all. 
I'd like to discuss this one more, and understand if it's just Liberty's implementation doing this or if the MP Concurrency spec mandates this.

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ManagedExecutorProducer.java
```diff
@@ -0,0 +1,53 @@
...
+        return exec;
+    }
+
+    private ManagedExecutor createManagedExecutor(ManagedExecutorConfigContainer config) {
+        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
+            Tr.debug(tc, "Building a new ManagedExecutor for the requested configuration: " + config);
+        return ManagedExecutorBuilder.instance()
+                        .maxAsync(config.maxAsync)
+                        .maxQueued(config.maxQueued)
+                        .propagated(config.propogated.toArray(new String[config.propogated.size()]))
```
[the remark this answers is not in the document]
aguibert: Doing this would require the main MP Concurrent bundle to export its impl package so the CDI producer could call the impl-specific builder method. Since these arrays will typically be small (1-4 elements) I don't think this is a big enough optimization to warrant exposing packages from the core MP Concurrent bundle

aguibert:
link to new build with review comment changes included

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ManagedExecutorProducer.java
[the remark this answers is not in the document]
njr-11: Yes, the MP concurrency spec currently mandates it, but the spec is still being developed if the community agrees it is wrong mandated.  This is best to discuss in the spec group.

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ManagedExecutorProducer.java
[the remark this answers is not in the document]
njr-11: I don't think we would do this by exporting an impl package. Instead, it would export an interface to our builder impl for this as well as other potential optimizations.  That said, I wouldn't bother making any of these changes until more discussion has happened in the spec to determine if this path is needed.

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ManagedExecutorProducer.java
njr-11: See the [javadoc for the computeIfAbsent method](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-).  The spec uses that language, although left to interpretation, so it is unclear if our builder constructor and related processing fits the definition.

PR #5801, opened 2018-11-16
Issue 5772: Check for null response from class file transformer - release-18.0.0.4

andymc12:
ClassFileTransformer.transform(...) should return null if it does not
transform the byte code passed in.  The AppClassLoader needs to check
for this before processing the null bytes.

Duplicate of PR #5774 (issue #5772), but pushing to release-18.0.0.4 branch.


andymc12:
#build

PR #5803, opened 2018-11-18
Compress upgrade cleanup - stabilize

cbridgha:
Removing temp project needed for co-req delivery


cbridgha:
Same as integration branch PR #5783

Issue #5804, opened 2018-11-18
Require "--" prior to command line variables


brenthdaniel:
After some discussion we've decided that variables specified on the command line should use argument form ("--variableName=value") rather than accepting everything after "--" as a variable. This aligns config variables with the way spring boot variables are specified.

PR #5805, opened 2018-11-18
Require argument form for command line variables

brenthdaniel:
PR for #5804


brenthdaniel:
#build

brenthdaniel:
Looks like there was an error downloading an artifact in the build, so 
#build again

PR #5675, opened 2018-11-08
Broaden search for Fault Tolerance fallback methods

Azquelt:
* allow the fallback method to be private or protected
* allow the fallback method to be on a superclass or interface
* add tests for validation
* add tests for calling fallback methods in valid locations

PR #5814, opened 2018-11-19
Improve error handling for Fault Tolerance 2.0 async execution

Azquelt:
**Recommend reviewing this as separate commits**

The goal of this task was to improve the error handling of FT 2.0 async executions. We do this by placing error barriers around code which is run asynchronously and ensuring that any errors caught there are logged and propagated asynchronously back to the user's application (assuming that a result has not yet been returned).
 * logging the error ensures that we can provide adequate servicing and helps with problem determination
 * propagating the error back to the user ensures that the user is not left with a Future or CompletionStage will never complete, in the event that an error prevents execution of the user's code.

In the course of the work some additional changes were required:
 * Use PolicyExecutor to enforce the bulkhead policy (error handling is complex here and the PolicyExecutor already takes care of 
 * Enhance the FAT framework so that a developer can easily restrict `RepeatTests` rules to a single configuration.
 * Re-target the Fault Tolerance FAT tests to run against both 1.1 and 2.0 to ensure that we don't create regressions in 2.0.
 * Make a few fixes to the 2.0 implementation which were revealed by running the FAT tests.


Azquelt:
No ID review needed. Modified file should be reviewed under #5675

Azquelt:
#build

Azquelt:
Java 7 build: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_SQ11gewUEeih-KCxmjCsJg

(To check validity of changes made to FAT framework)

Azquelt:
#build

dev/fattest.simplicity/bnd.bnd
```diff
@@ -56,6 +56,9 @@ Private-Package: \
 
 test.project: true
 
+# Allow access to @Repeatable annotation from Java 8, while still compiling for Java 7 compatibility
+-buildpath.bootclasspath: ${javac.bootclasspath.1.8}
```
aguibert: This gives the simplicity framework more than just access to `@Repeatable`, and I don't think it's worth the risk for FATs that still need to run on Java 7

dev/fattest.simplicity/src/componenttest/rules/repeater/RepeatTests.java
```diff
@@ -111,5 +111,14 @@ public void evaluate() throws Throwable {
                 }
             }
         }
+
+        private static boolean shouldRun(RepeatTestAction action) {
+            String repeatOnly = System.getProperty("fat.repeat.only");
+            if (repeatOnly == null) {
```
aguibert: if `repeatOnly == null`, shouldn't we be returning `action.isEnabled()` instead of `true` all the time? (existing behavior before this sys prop was introduced)

dev/fattest.simplicity/src/componenttest/annotation/TestServlet.java
```diff
@@ -19,6 +20,7 @@
 
 @Target({ ElementType.FIELD })
 @Retention(RetentionPolicy.RUNTIME)
+@Repeatable(TestServlets.class)
```
aguibert: This is a good change and I really wanted to do this when I first wrote `@TestServlet`, but unfortunately core components like fattest.simplicity still need to use Java 7

dev/fattest.simplicity/src/componenttest/rules/repeater/RepeatTests.java
Azquelt: Switched this around

dev/fattest.simplicity/bnd.bnd
Azquelt: Removed this

dev/fattest.simplicity/src/componenttest/annotation/TestServlet.java
Azquelt: Removed this

Azquelt:
#build

Azquelt:
#build

Azquelt:
Personal build failures appear to be unrelated

PR #5815, opened 2018-11-19
implement and test withContextCapture methods for server config based ThreadContext/ContextService

njr-11:
Need to implement the withContextCapture methods for when the ThreadContext/ContextService instance is configured at server level.  This implementation should be largely the same as the builder/CDI path, and tests should be similar as well, except that there is no way of configuring to leave context unchanged on the thread in that path.  Write similar tests, expecting this.

### DIFF
--- a/dev/com.ibm.ws.concurrent.mp.1.0/src/com/ibm/ws/concurrent/mp/ContextServiceImpl.java
+++ b/dev/com.ibm.ws.concurrent.mp.1.0/src/com/ibm/ws/concurrent/mp/ContextServiceImpl.java
@@ -15,6 +15,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


njr-11:
#build

dev/com.ibm.ws.concurrent.mp.1.0/src/com/ibm/ws/concurrent/mp/ContextServiceImpl.java
```diff
@@ -75,6 +91,39 @@ protected void deactivate(ComponentContext context) {
...
+            PolicyExecutor policyExecutor = concurrencyProvider.policyExecutorProvider.create(name);
+            policyExecutor.maxConcurrency(-1).maxQueueSize(-1);
+            // TODO these policy executor instances, as well as those created via ManagedExecutorBuilder are never shut down
+            // and removed from PolicyExecutorProvider's list. This is a memory leak and needs to be fixed.
+
+            executor = new ManagedExecutorImpl(name, policyExecutor, this, concurrencyProvider.transactionContextProvider.transactionContextProviderRef);
+
+            if (!managedExecutorRef.compareAndSet(null, executor)) {
+                // Another thread updated the reference first. Discard the instance we created and use the other.
+                executor.shutdown();
```
alexsm82: Will this result in an IllegalStateException since ManagedExecutorImpl extends ManagedExecutorServiceImpl?

njr-11:
#build

njr-11:
#build

PR #5830, opened 2018-11-20
Updated SelfExtract so that '~' is expanded during interactive mode

AnshulPatel1:

### DIFF
--- a/dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
+++ b/dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
@@ -181,9 +181,6 @@ public class SelfExtract {
                 String pathNameToReturn = pathName.substring(1);
                 String home = System.getenv("HOME");
                 if (home == null) {
-                    home = System.getProperty("user.home");
-                }
-                if (home == null) {
                     err("invalidInstall", pathName);
                     System.exit(ReturnCode.BAD_INPUT);
                 }
--- a/dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
+++ b/dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
@@ -176,12 +176,18 @@ public class SelfExtract {
                 outputDirFromUser = new File(targetString.trim());
             }
 
-            if (outputDirFromUser.getPath().charAt(0) == '~') {
-                String pathName = outputDirFromUser.getPath();
+            String pathName = outputDirFromUser.getPath();
+            if (pathName.startsWith("~")) {
                 String pathNameToReturn = pathName.substring(1);
                 String home = System.getenv("HOME");
-                pathNameToReturn = home + pathNameToReturn;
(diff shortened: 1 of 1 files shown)


AnshulPatel1:
#build

dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
```diff
@@ -167,6 +175,15 @@ public static int doMain(String[] args) {
             } else {
                 outputDirFromUser = new File(targetString.trim());
             }
+
+            if (outputDirFromUser.getPath().indexOf('~') >= 0) {
+                String pathName = outputDirFromUser.getPath();
+                String pathNameToReturn = pathName.replaceAll("~", "");
```
ericglau: Should check for `~` only at the beginning of the string, since `~` is a legal character in the middle of a path (e.g. it could be part of a directory name).  Then in that case, don't do `replaceAll` but only replace the first instance of that char.

AnshulPatel1:
#build

dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
```diff
@@ -167,6 +175,15 @@ public static int doMain(String[] args) {
             } else {
                 outputDirFromUser = new File(targetString.trim());
             }
+
```
alvinso: Is the following better?
            
           String pathName = outputDirFromUser.getPath();
            if (pathName.trim().startWith("~")) {

dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
```diff
@@ -167,6 +175,15 @@ public static int doMain(String[] args) {
...
+            if (outputDirFromUser.getPath().charAt(0) == '~') {
+                String pathName = outputDirFromUser.getPath();
+                String pathNameToReturn = pathName.substring(1);
+                String home = System.getenv("HOME");
+                pathNameToReturn = home + pathNameToReturn;
```
alvinso: We can create File directly with the parent path

outputDirFromUser = new File(home, pathNameToReturn);

dev/wlp.lib.extract/src/wlp/lib/extract/SelfExtract.java
```diff
@@ -167,6 +175,15 @@ public static int doMain(String[] args) {
             } else {
                 outputDirFromUser = new File(targetString.trim());
             }
+
+            if (outputDirFromUser.getPath().charAt(0) == '~') {
+                String pathName = outputDirFromUser.getPath();
+                String pathNameToReturn = pathName.substring(1);
+                String home = System.getenv("HOME");
```
alvinso: does System.getenv("HOME") work on Windows

AnshulPatel1:
#build

AnshulPatel1:
#build

AnshulPatel1:
#build

AnshulPatel1:
#build

PR #5838, opened 2018-11-21
Verify application component availability before applying context

njr-11:
Verify that the application component that submitted a task/action is still available before applying context to run the task or action.

This will comply with EE Concurrency 3.3.4, which states:
"All invocations to any of the proxied interface methods will fail with a java.lang.IllegalStateException exception if the application component is not started or deployed."


njr-11:
#build

njr-11:
#build

PR #5849, opened 2018-11-26
5507 ft unit tests

benjamin-confino:


benjamin-confino:
#build

dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance20/impl/FutureShellTest.java
```diff
@@ -0,0 +1,262 @@
...
+            futureShell.get(100, TimeUnit.MILLISECONDS);
+            fail("FutureShell did not throw a timeout exception");
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        } catch (ExecutionException e) {
+            e.printStackTrace();
+        } catch (TimeoutException e) {
+            timedOut = true;
+        }
+        assertTrue("No TimeoutException was thrown", timedOut);
```
Azquelt: If you allow other exceptions to be thrown, you shouldn't actually need this assert, further simplifying the test.

dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance20/impl/MethodResultTest.java
```diff
@@ -0,0 +1,42 @@
...
+        assertThat("Sucessful result had the wrong toString", result.toString(), containsString("Method returned value"));
+        assertEquals("Sucessful result had the wrong result", result.getResult(), ("test"));
+    }
+
+    @Test
+    public void testUnsuccessfulMethodResult() {
+        Throwable testThrowable = new NullPointerException("test");
+        MethodResult<String> result = MethodResult.failure(testThrowable);
+        assertTrue("Unsucessful result returned false on isFailure", result.isFailure());
+        assertEquals("Unsucessful result contained the wrong failure", result.getFailure(), testThrowable);
```
Azquelt: JUnit's `assertEquals` wants the expected value first and the actual value second. If they're the wrong way round, then the failure message is confusing.

Annoyingly, this is the opposite way round from `assertThat`.

If you prefer, you can change them all to `assertThat(foo, equalTo(bar))`

benjamin-confino:
#build

PR #5854, opened 2018-11-26
implement shutdown for MicroProfile managed executors

njr-11:
Allow life cycle operations (shutdown, shutdownNow, awaitTermination, isShutdown, isTerminated) for managed executors that are created via MicroProfile.   Continue to disallow the life cycle operations for server-wide managed executors created via EE Concurrency, which is necessary for compliance with that specification.


njr-11:
Build has 2 test failures in unrelated test buckets

PR #5864, opened 2018-11-27
added a check to make sure jax-ws components are not tested with java11

Erikkre:
Add an if block to make sure jaxws components are only run with java versions 8 and below.
Issue #5757

### DIFF
--- a/dev/com.ibm.ws.microprofile.metrics.monitor_fat/fat/src/com/ibm/ws/microprofile/metrics/monitor_fat/MetricsMonitorTest.java
+++ b/dev/com.ibm.ws.microprofile.metrics.monitor_fat/fat/src/com/ibm/ws/microprofile/metrics/monitor_fat/MetricsMonitorTest.java
@@ -231,7 +231,7 @@ public class MetricsMonitorTest {
       		new String[] {}, 
       		new String[] { "vendor:" });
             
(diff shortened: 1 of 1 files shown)


Erikkre:
#build

dev/com.ibm.ws.microprofile.metrics.monitor_fat/fat/src/com/ibm/ws/microprofile/metrics/monitor_fat/MetricsMonitorTest.java
```diff
@@ -260,6 +230,40 @@ public void testEnableDisableFeatures() throws Exception {
       	checkStrings(getHttpsServlet("/metrics"), 
       		new String[] {}, 
       		new String[] { "vendor:" });
+            
+        if (componenttest.topology.impl.JavaInfo.forServer(server).JAVA_VERSION<=8){
```
fmhwong: `JAVA_VERSION` just returns `JavaInfo.forCurrentVM().majorVersion();`. I think you should use `JavaInfo.forServer(server).majorVersion()` instead.

Erikkre: Change tested and committed.

PR #5876, opened 2018-11-28
Fault Tolerance Async Cancellation

Azquelt:
Add support for cancelling methods annotated with `@Asynchronous` which return a `Future`


Azquelt:
#build

PR #5878, opened 2018-11-28
enforce builder constraints on context configuration

### LINKED ISSUE #5707: Container provided thread context
User story:
"As a user of Liberty, I would like the built-in, container-provided thread context types to be available for thread context capture/propagation/clearing when using MicroProfile Concurrency ManagedExecutor and ThreadContext, so that actions/tasks can operate the same as the requesting thread."

For built-in context types that do not map directly to spec defined context types, need to decide whether to externalize constants/context types names, or whether the only way of configuring them will be with the catch-all ThreadContext.REMAINING constant.

njr-11:
fixes #5707

Per spec, ThreadContext builder and ManagedExecutor builder must raise IllegalStateException when context configuration overlaps between cleared/propagated/unchanged and must also raise this exception if cleared or propagated sets contain any unavailable context types.


njr-11:
Build has 1 test failure which is a duplicate of an unrelated pre-existing issue

PR #5894, opened 2018-11-30
test case is missing code to wait for termination before verifying the state

njr-11:
Inserted a line to await termination before the test case validates that the executor has reached the terminated state.


njr-11:
#build

PR #5901, opened 2018-11-30
remove unused attributes from threading metatype

garypicher:


garypicher:
#build

garypicher:
#build

garypicher:
#build

garypicher:
Failures in PB are related to defects 243004 and 259888.  The latter defect is not an exact match, but this JAX-WS bucket is new and there were several failures in the Full FAT builds over the weekend since the bucket was added.  None of the failures look at all related to this change.

PR #5902, opened 2018-11-30
change withCurrentContext completion stages to run on same thread, per spec

njr-11:
Update our implementation of completion stages returned by withCurrentContext to match a recent change to the spec that requires them to be backed by the ThreadContext rather than a managed executor so that they must run actions on the same thread.


njr-11:
#build

njr-11:
#build

njr-11:
build has 5 test failures, all pre-existing

Issue #5900, opened 2018-11-30
Http/2 test issue - broken pipe


loriadi:
testInvalidPaddingValue sends goaway frame before the expected end of the test

PR #5925, opened 2018-12-03
Reordered the testInvalidPaddingValue test to put the expected frame in

loriadi:
Need to add a goaway frame to the expected list before the start up sequence. 

Because of timing, the test will sometimes fail if the start up sequence completes before the goaway is added as expected.


loriadi:
#build

loriadi:
fixes #5900

PR #5951, opened 2018-12-05
Exclude test testClassLevelRetryMaxDuration temporarily as unreliable on slow hardware

hutchig:
Signed-off-by: Gordon Hutchison <gordon.hutchison@gmail.com>

PR #5956, opened 2018-12-05
5955 test bulkhead method asynch future done without get slow macs

hutchig:
This branch is built on top of https://github.com/OpenLiberty/open-liberty/pull/5951
for issue 5950. So the first two commits are from that. 5951 is in review, I will rebase this
if there are additional commits there.


hutchig:
#build

PR #5962, opened 2018-12-05
Add post invoke delay before checking if timer is deleted in fat test

olendvcook:
Test checks that a single action timer is cancelled after running. Test correctly uses a CountDownLatch to wait for timeout before checking, but because of a slight timing window from postinvoke the timer might be in the middle of being cancelled when the test asserts that it's deleted.

Adding a slight sleep to account for post invoke.


olendvcook:
#build

PR #5966, opened 2018-12-05
Add unsafe method from Oracle 18c Driver

alexsm82:
Block access to the new `setSafelyClosed` method on the new Oracle 18c driver on an unwrapped connection, as this methods may close out underlying resources from under Liberty.


alexsm82:
#build

alexsm82:
The personal build failures are all unrelated.

PR #5986, opened 2018-12-07
Fix issue where timeout.stop() can be called before timeout.start()

Azquelt:
Previously, the AsyncExecutor could not start the timeout until after it
had submitted the task, because the timeout callback requires the
ExecutionReference for the running task.

However, this meant that in rare cases the task could run, finish and
timeout.stop() could be called before timeout.start() was called.

To prevent this, we split the starting of the timeout and the setting of
the timeout callback. This allows the timeout to be started before
submitting the task and the callback to be set afterwards.

Also took the opportunity to tighten up the state handling logic in
Timeout to help ensure that any more incorrect handling of timeout
states will be caught.

Also fix message output from FT 2.0


Azquelt:
#build

Azquelt:
Build failures do not appear to be related to this change.

Issue #6011, opened 2018-12-12
Test Failure (Liberty Continuous Build - EBC - 20181201-1307): com.ibm.websphere.microprofile.faulttolerance_fat.tests.CDITimeoutTest.testTimeoutClassScopeConfig_MICROPROFILE13


hutchig:
See https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=259566

hutchig:
This has not re-occurred since #6012 (the fix) was merged.

PR #6012, opened 2018-12-12
ft_fat smooth over outlier timing glitch

hutchig:

This change is an attempt to discount errors that arise from a slow
period in underlying test hardware.

The basic idea is that is a timeout takes too long, the test will wait for a little
while and then take 10 timing measurements. Only if they take too long on
average will a test fail and a defect be raised.


hutchig:
#build

hutchig:
#build

hutchig:
Fixes #6011

PR #6015, opened 2018-12-12
Do not throw exception when output stream does not match

naumanna:
The ConcurrentModificationException thrown in CapturedOutputHolder is causing periodic unit test errors in the Liberty builds.  Changing it to a warning message to avoid the problem breaking everything.


naumanna:
This is not the fix we want to make, closing this out.

PR #6016, opened 2018-12-12
Sync changes in fattest.simplicity

aguibert:


aguibert:
#build

aguibert:
#build

aguibert:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -1071,10 +1181,12 @@ public ProgramOutput startServerWithArgs(boolean preClean, boolean cleanStart,
         JVM_ARGS += " -Djava.security.egd=file:///dev/urandom";
 
         JavaInfo info = JavaInfo.forServer(this);
-        // Debug for a highly intermittent problem on IBM JVMs.
+        // Debug for a Java 8 problem on IBM JVMs. See defect 181046 for details.
```
ayoho: Did you want to have a defect number in the code here?

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -2510,7 +2644,7 @@ protected void recursivelyCopyDirectory(RemoteFile remoteDirectory, LocalFile de
             RemoteFile toCopy = new RemoteFile(machine, remoteDirectory, l);
             LocalFile toReceive = new LocalFile(destination, l);
             String absPath = toCopy.getAbsolutePath();
-            Log.finest(c, "recursivelyCopyDirectory", "Getting: " + absPath);
+            Log.finest(c, "recursivelyCopyDirectory", "Getting: " + toCopy.getAbsolutePath());
```
ayoho: Not sure why this change was made.

dev/fattest.simplicity/src/componenttest/topology/impl/LogMonitor.java
```diff
@@ -247,15 +248,22 @@ protected String waitForStringInLogUsingMark(String regexp, long intendedTimeout
...
      *
-     * @param timeout Timeout (in milliseconds)
+     * @param regexp a regular expression to search for
+     * @param intendedTimeout a timeout, in milliseconds, within which the wait should complete. Exceeding this is a soft fail.
+     * @param extendedTimeout a timeout, in milliseconds, within which the wait must complete. Exceeding this is a hard fail.
+     * @param outputFile file to check
      * @return line that matched the regexp
      */
-    public String verifyStringNotInLogUsingMark(String regexToSearchFor, long timeout) {
+    public boolean verifyStringNotInLogUsingMark(String regexp, long timeout) {
```
ayoho: I disagree with this change. I deliberately changed this method in https://github.com/OpenLiberty/open-liberty/pull/5038 to return a String. I can understand having a method that returns a boolean for this case because we're verifying whether or not a string was in the log - that's clearly best represented by a boolean. But I believe it's arguably more useful to return the actual string that was found in the log because calling methods can then  log or do additional checking on the line that was found. It's been frustrating to me in some cases where I've used this method when it returned a boolean because it didn't tell me what the line was that it matched. Having that information helps me determine if I wrote my regex incorrectly, if the marks in the log haven't been set as I thought, or if it's finding some other line that I wasn't aware would match.

aguibert:
@ayoho thanks for reviewing, I amended my commit with fixes to the comments you made

aguibert:
#build

aguibert:
Merging changes, all FAT failures are close matches to existing defects

PR #6030, opened 2018-12-13
Dump and restore output streams between unit tests

naumanna:
We are intermittently seeing ConcurrentModificationExceptions thrown from captureStreams() in CapturedOutputHandler called from SharedOutputHandler apply() in the automated unit tests.  

This will dump and restore the output streams between each test, which will hopefully prevent that intermittent error.


naumanna:
#link 248532

PR #6056, opened 2018-12-17
cleared context improperly cloned

njr-11:
There is a bug in the clone method of 
com/ibm/ws/concurrent/mp/MicroProfileClearedContextSnapshot.java
where it makes a shallow copy of the restore context types, which are not supposed to be shared across snapshots.


njr-11:
#build

njr-11:
build has 14 test failures, all within a single test bucket that is unrelated to this change and is a preexisting issue.

PR #6063, opened 2018-12-18
Fix file name runnable

NottyCode:
In 18.0.0.3 and earlier setting --archive=fileName.zip and --include=runnable didn't generate a file called fileName.zip. This might seem weird, but the maven/gradle tools were depending on it so PR fixes it and creates a test.


NottyCode:
#build this please Miss Libby

Issue #6067, opened 2018-12-19
Fix NPE in com.ibm.ws.app.manager.module.internal.DeployedAppInfoBase.getProtectionDomain


c00crane:
If the list of merged permissions is empty, the code still attempt to do an add a permission that does not exist resulting in an NPE.
This change will resolve that issue.

james-ro-williams:
Would it be possible if I could work on this? I have an idea as to a possible fix.

c00crane:
@james-ro-williams sure thing!

PR #6068, opened 2018-12-19
only add java permissions if they exist

c00crane:
#6067 
Fix NPE that occurs when we attempt to add an empty list of permissions.

### DIFF
--- a/dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
+++ b/dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
@@ -512,8 +512,10 @@ public abstract class DeployedAppInfoBase extends SimpleDeployedAppInfoBase impl
 
             ArrayList<Permission> mergedPermissions = permissionManager.getEffectivePermissions(applicationInformation.getLocation());
             int count = mergedPermissions.size();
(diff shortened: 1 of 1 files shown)


c00crane:
#build 
#spawn.fullfat.buckets=com.ibm.ws.security.saml.sso_fat.logout

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
```diff
@@ -512,8 +512,10 @@ protected ProtectionDomain getProtectionDomain() {
 
             ArrayList<Permission> mergedPermissions = permissionManager.getEffectivePermissions(applicationInformation.getLocation());
             int count = mergedPermissions.size();
-            for (int i = 0; i < count; i++)
-                perms.add(mergedPermissions.get(i));
+            if (count > 0) {
+                for (int i = 0; i < count; i++)
+                    perms.add(mergedPermissions.get(i));
+            }
```
ayoho: This change doesn't look like it will do anything. In the original code, if `count == 0`, the code never enters the `for` loop and nothing happens. Introducing an `if (count > 0)` check here is redundant.

ayoho:
Is this pull request needed?

c00crane:
no - not sure how they hit the npe where they did.  
We haven't seen it since

PR #6069, opened 2018-12-19
Avoid NPE for repogen on Java 11

aguibert:
#build


aguibert:
Cancelling this PR in favor of #6074

PR #6082, opened 2018-12-20
Allow for FFDC from intentional exception raised by test

njr-11:
The MP Concurrency TCK includes coverage of a tasks/actions that intentionally raises an exception, and this is causing an FFDC in the Liberty server logs, which is detected as a test failure.  Need to add the proper annotation indicating that FFDC for this exception should be allowed.


njr-11:
#build

PR #6092, opened 2018-12-27
Update connection wait timeout exception to include number of waiting requests and the total connection count

jmstephensgit:
For improved serviceability and consumability providing the number of waiting threads with the total connection count may avoid requiring trace for tuning recommendations.   If only a few connection wait timeouts are occurring, if the waiting threads numbers are low, increasing the maximum connections by just a few connections may be enough to avoid the connection wait timeouts.  If the waiting threads numbers are high relative to the total connections, then several considerations may be required to fix the problem correctly.  When the waiting thread numbers are high, database performance, network performance, number of threads using the datasource and any other performance related issue around connections will need to be reviewed.


jmstephensgit:
#build

jmstephensgit:
#build

PR #6093, opened 2018-12-29
CompletableFuture unaware of executor shutdown and other abort paths

njr-11:
CompletableFutures created by a ManagedExecutor which are still queued for execution are unaware when shutdownNow is invoked on the executor.  This causes the CompletableFutures (plus dependent stages) to remain incomplete even though they will never be able to run.  This pull adds the necessary code so that these CompletableFutures can be reliably canceled/aborted for the Java 11 implementation, and cancel/aborted on a best effort basis for the Java 8 implementation.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build has 5 test failures (in a single bucket) which are duplicates of a pre-existing issue.

Issue #5970, opened 2018-12-06
trackLoggedOutSSOCookies setting causing multiple login failure


una-tapa:
Issue:   
When application performs HttpServletRequest.logout(), Liberty fails to remove LtpaToken from its AuthCache.  This causes unexpected behavior when trackLoggedOutSSOCookies flag is set to true.  
When a user login from a browser for the first time (without LtpaToken),  Liberty would still find LtpaToken in the AuthCache.  Since this token is now tracked as logged-out-token, the user can no longer login.

una-tapa:
Port service PR here.  https://github.ibm.com/was-liberty/open-liberty-service/pull/171

PR #6100, opened 2019-01-02
trackLoggedOutCookies setting cause multiple login failure

una-tapa:
trackLoggedOutCookies setting could cause multiple login failure.  

This is caused by following 2 reasons. 

1) There are cases that authCache entry not being removed when trackLoggedOutCookies setting is configured, and subsequent login hit the authCache retrieving the same cookie (that is considered logged out) 

and/or 

2) The logic of LtpaToken expiration time allows the same token to be created within 1 min login-window.   The new cookie may be the same value as the logged out cookie.


una-tapa:
#build

una-tapa:
Fixes #5970

PR #6117, opened 2019-01-04
create new test application/server instance for testing MP Config overrides of MP Concurrency

njr-11:
Create a new test application and server instance with both mpConfig and mpConcurrency enabled where we can test MicroProfile Config being used to override configuration of ManagedExecutor and ThreadContext.  For now, I'll start with a single test case, where the application uses MP Config directly, which will demonstrate everything is working before we later add the tests of MP Config being used by the MP Concurrency implementation for container produced instances.


njr-11:
#build

njr-11:
#build

njr-11:
Build has 9 test failures which are pre-existing issues in unrelated areas.

PR #6126, opened 2019-01-07
MicroProfile Config overrides int attribute of ManagedExecutorConfig

njr-11:
This is the first set of changes to get MicroProfile Config overrides of ManagedExecutor injection points working on one basic code path: the code path where `@ManagedExecutorConfig` is used on a field without any qualifiers.

### DIFF
--- a/dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/MPConfigAccessorImpl.java
+++ b/dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/MPConfigAccessorImpl.java
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2018 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.concurrent.mp.cdi;
+
+import java.util.Optional;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+import org.osgi.service.component.annotations.Component;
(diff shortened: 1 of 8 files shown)


njr-11:
#build

njr-11:
#build

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/MPConfigAccessorImpl.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Reads a String[] or Integer property value from MicroProfile Config.
+     *
+     * @param name config property name.
+     * @param defaultValue value to use if a config property with the specified name is not found.
+     * @return value from MicroProfile Config. Otherwise the default value.
+     */
+    @Override
+    public <T> T get(String name, T defaultValue) {
+        Config config = configProviderResolver.getConfig();
```
aguibert: Looking through our implementation of ConfigProviderResolver, this seems like a rather expensive operation with several synchronized blocks in it.  Lets lazily initialize and cache the `Config` instance per `MPConfigAccessorImpl` instance

njr-11:
build has 5 test failures which are pre-existing issues in another bucket

njr-11:
Sadly, there are some pretty severe merge conflicts on this, so it will need yet another build.  I should have delivered it yesterday instead of trying to be safe and resubmitting the build after the general MP bucket failures.

njr-11:
#build

naumanna:
#build

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/MPConfigAccessorImpl.java
njr-11: We will look into improving this path under subsequent pulls, however, it isn't as simple as caching a single config instance for the MPConfigAccessorImpl singleton because that singleton is used to load config that can differ by application due to microprofile-config.properties.  Before we do anything in this area, I would like to have added a test that utilizes MP config per application to ensure we aren't breaking it.

njr-11:
Build has 37 test failures which are pre-existing issues in unrelated test buckets.

PR #6146, opened 2019-01-09
test MP Config override of array type attribute for unqualified injection point without ManagedExecutorConfig

njr-11:
This test covers both override of array type attributes via MicroProfile Config as well as MicroProfile Config override of ManagedExecutor injection points that have neither qualifiers nor the ManagedExecutorConfig annotation.


njr-11:
#build

njr-11:
build has 36 test failures which are pre-existing issues in unrelated test buckets

PR #6149, opened 2019-01-10
FT 2.0 CompletionStage new semantics

### LINKED ISSUE #6139: FT 2.0: New CompletionStage semantics
Change AsyncExecutor for methods which return a `CompletionStage` so that result processing is done when the returned `CompletionStage` is complete, rather than when the method returns.

Azquelt:
The FT 2.0 spec has been updated to say that when an `@Asynchronous` method returns a `CompletionStage`, any other fault tolerance logic is applied when the returned `CompletionStage` completes, rather than when the method returns.

This change adds this behaviour for everything except `@Bulkhead`.

See the details in the changes under eclipse/microprofile-fault-tolerance#381

Fixes #6139


Azquelt:
#build

Azquelt:
#build

Azquelt:
The failures in the build do not appear to be related to this change.

PR #6153, opened 2019-01-10
Propagate NamingException when enumerating through JNDI bindings

### LINKED ISSUE #6152: NamingException masked when listing entries in a JNDI context
When using the OSGi console's `jndilist` command, a NullPointerException may be received instead of the underlying NamingException being propagated.

chackoj:
fixes #6152


chackoj:
#build

PR #6157, opened 2019-01-10
Allow MP Config to override ManagedExecutor injection points which are method parameters

njr-11:
Implement the code path to allow MicroProfile Config to be used to override ManagedExecutor injections which are method parameters without the NamedInstance qualifier (that combination will be a separate pull and will require some additional refactoring).  Write a test case to validate.


njr-11:
#build

njr-11:
Build has 37 test failures which are preexisting issues in unrelated test buckets

PR #6138, opened 2019-01-09
Upgrade existing Guava project to 26.0

ayoho:
This will result in a `com.ibm.ws.com.google.guava` and `com.ibm.ws.com.google.guava.26.0` project that are effectively identical (the latter will have `com.ibm.ws.com.google.guava.26.0` as its BSN to distinguish it.) Once all references in the commercial product are updated, the `com.ibm.ws.com.google.guava.26.0` project should be able to be removed.

PR #6162, opened 2019-01-10
Update Guava to 27.0.1-android for Java 7 support

ayoho:
The 27.0.1-jre version of Guava is compiled with Java 8 so will not work on systems running Java 7. The x.x.x-android artifacts are built with Java 7, so I'm updating the library to use that version.


ayoho:
I think I'm actually going to include this in #6138.

ayoho:
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_gM9hoBUSEemJj5Oykri-vA shows that this fix will work.

Issue #6147, opened 2019-01-10
FT 2.0: Correct semantics for Async Bulkhead


Azquelt:
This arose when implementing Asynchronous Bulkhead with CompletionStage but it became apparent that there are wider implications which require a rethink about how we implement bulkhead for FT 2.0.

Currently we delegate to a PolicyExecutor which restricts the number of running threads for us.

However, within the task we submit to a policy executor, we do more than just run the user's code. We also do the finalization logic including checking the result, scheduling a retry, running the fallback method and committing the result.

This has the following problems:
* The bulkhead reservation is held while the fallback is run
* The bulkhead reservation is still held at the point when the result is committed and becomes visible to the user.
  * This causes metrics tests to fail because the bulkhead reservation is still held and execution times have not been recorded after we signal to the user that everything is complete
* We have no way to hold the bulkhead reservation _beyond_ the point when the method returns. This is reqired for the new CompletionStage semantics.

I can see two options:
* Ensure that we only run the user's code within the policy executor task
  * This would mean enqueueing another task to do the attempt finalization logic
* Implement our own logic to limit the number of simultaneously running tasks
  * In this case we have to be really careful about ensuring that we always complete the user's result (either successfully or exceptionally)

Note: the spec says that when using CompletionStage, fault tolerance policies are not applied until the returned completion stage completes. This means that limiting the number of running threads is insufficient as we need to retain the bulkhead reservation even after our thread has finished.

PR #6166, opened 2019-01-11
Add fault tolerance 2.0 metrics

### LINKED ISSUE #5492: Add Metric recording to FT 2.0
Having replaced the executors in FT 2.0, we need to make them call the metric recorder methods at the appropriate time.

This could either be done by the executor, or by the stateImpls

Azquelt:
Adds metrics support to Fault Tolerance 2.0

Test coverage is provided by the TCK

Still missing are the execution and queuing time histograms from asynchronous bulkhead which will be added under #6147. (Currently the Async Bulkhead is broken in a fundamental way so these metrics can't be added correctly. It doesn't seem worth the effort of adding them incorrectly given that they won't pass the tests anyway.)

Fixes #5492

Ignore the first two commits, they will be removed when #6149 is merged


Azquelt:
Rebased on integration

Azquelt:
#build

Azquelt:

PR #6169, opened 2019-01-11
avoid repeatedly resolving MPConfig

njr-11:
Performance improvement to avoid repeating resolving of MicroProfile Config.

### DIFF
--- a/dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ConcurrencyCDIExtension.java
+++ b/dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ConcurrencyCDIExtension.java
@@ -61,11 +61,13 @@ public class ConcurrencyCDIExtension implements Extension, WebSphereCDIExtension
     private final Set<Throwable> deploymentErrors = new LinkedHashSet<>();
     private final Set<String> appDefinedProducers = new HashSet<>();
 
-    private final MPConfigAccessor mpConfig = AccessController.doPrivileged((PrivilegedAction<MPConfigAccessor>) () -> {
+    final MPConfigAccessor mpConfigAccessor = AccessController.doPrivileged((PrivilegedAction<MPConfigAccessor>) () -> {
         BundleContext bundleContext = FrameworkUtil.getBundle(ManagedExecutorBean.class).getBundleContext();
         ServiceReference<MPConfigAccessor> mpConfigAccessorRef = bundleContext.getServiceReference(MPConfigAccessor.class);
         return mpConfigAccessorRef == null ? null : bundleContext.getService(mpConfigAccessorRef);
     });
+
+    Object mpConfig;
 
     @Trivial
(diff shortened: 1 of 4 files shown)


njr-11:
#build

dev/com.ibm.ws.concurrent.mp.cdi/src/com/ibm/ws/concurrent/mp/cdi/ConcurrencyCDIExtension.java
```diff
@@ -61,12 +61,14 @@
...
     private final Set<String> appDefinedProducers = new HashSet<>();
 
-    private final MPConfigAccessor mpConfig = AccessController.doPrivileged((PrivilegedAction<MPConfigAccessor>) () -> {
+    final MPConfigAccessor mpConfigAccessor = AccessController.doPrivileged((PrivilegedAction<MPConfigAccessor>) () -> {
         BundleContext bundleContext = FrameworkUtil.getBundle(ManagedExecutorBean.class).getBundleContext();
         ServiceReference<MPConfigAccessor> mpConfigAccessorRef = bundleContext.getServiceReference(MPConfigAccessor.class);
         return mpConfigAccessorRef == null ? null : bundleContext.getService(mpConfigAccessorRef);
     });
 
+    Object mpConfig;
```
aguibert: instead of storing the variable on the extension, I think it would be cleaner to lazily get/cache it within the MPConfigAccessor instance

njr-11: That wouldn't be valid. There is a single MPConfigAccessor instance for the server.  And there are many different cdi extension classes that need to use it, each with a different MP Config instance.

aguibert: right, I missed that. But is it necessary to get the accessor from the service registry? As opposed to just creating a new instance of the accessor per extension?  That way each accessors and MPConfig instances have a simple 1-1 relationship.

njr-11: Even though MPConfigAccessorImpl is in the same bundle, it still matters that it accesses classes which are only sometimes available.  We cannot invoke its constructor directly from code that might not have access to those classes without causing class loading problems JDKs that aggressively load.  Sure, reflection could be used, but that makes the code much less clear.

njr-11: Unless we want to use reflection to construct instances (and that would be much less clear), the MPConfigAccessorImpl cannot be constructed by our code where there might or might not be access to MP Config classes.  Otherwise class loading will fail on more aggressive JDKs.

njr-11:
Build has 18 test failures, which are preexisting issues in unrelated test buckets

PR #6171, opened 2019-01-11
Issue #6148 - use BSN where version contains special chars

### LINKED ISSUE #6148: Update com.ibm.websphere.javaee.jaxb.2.2 to use correctly formatted BSN dependency
The buildpath currently says
```
-buildpath: \
	javax.xml.bind:jaxb-api;strategy=exact;version=2.2.12.b140109_1041
```
this is in G:A format but it should be in BSN format...
```
-buildpath: \
	jaxb-api;strategy=exact;version=2.2.12.b140109_1041
```
i.e. it should refer to the bundle symbolic name and not the maven group:artifact identifier

tevans78:
fixes #6148 
#build


tevans78:
Personal Build has failures but none look related to this change. If this change broke something then it would be far more obvious, possibly even a compilation failure. Merging.

PR #6184, opened 2019-01-14
Use in memory derby for injection_fat

olendvcook:
Use in memory derby for injection_fat for build performance reasons


olendvcook:
#build

PR #6203, opened 2019-01-15
update MP Concurrency TCK bucket for running tests that use MP Config

njr-11:
MicroProfile Config needs to be enabled in the MicroProfile Concurrency TCK test bucket because some of the TCK tests rely on it.


njr-11:
Build has 15 test failures, which are preexisting issues in unrelated test buckets.

PR #6216, opened 2019-01-16
Move FireOnEventListenerVisitor classes inside a parent to help prevent compilation issues

tevans78:
Occasionally when building OpenLiberty locally, it takes two attempts to build the webcontainer code. We think it is because FireOnEventListenerVisitors.java does not contain a single class called FireOnEventListenerVisitors but rather has multiple classes named something else. We hope that changing these to be static child classes will mean the compilation works correctly every time.
#build


tevans78:
PB has no relevant failures

PR #6220, opened 2019-01-17
Fixed merging problem in Reader.java file

Rachelwrq:
There was a minor merging issue with `Content` and `Responses` in the previous `setpathItemOperation` method. If there were multiple methods in one class with the same HTTP method annotation, then the produced document that `Reader.java` would not be complete; since one will overwrite the other. This pr fixed the issue by adding `putIfAbsent` to required Objects.


Rachelwrq:
The changes were tested through the following methods:
- Running the `ant` build locally on the the files, this step created a new `build.image/wlp` folder against new changes
- Created a test application on eclipse and point the Liberty server to the new `build.image/wlp`. The document produced by the test app proved that the changes fixed the mentioned issue.
- Ran the fat bucket on personal build on RTC 30 times to ensure that no failures occurs

Issue #6106, opened 2019-01-03
Update EclipseLink trace to be more informative


dazey3:
I updated WebSphere Traditional to be a bit more informative and allow for the same "eclipse." notation available on Open Liberty already. In the process, I added some more content and I want to keep the strategies in sync.

PR #6235, opened 2019-01-17
Issue 6106: Fix EclipseLink null channel bug

dazey3:
for #6106 
Bug found in trace where the channel was not being set


dazey3:
#build

PR #6236, opened 2019-01-17
Issue #5472 - check OSGi Framework state

### LINKED ISSUE #5472: MP Config: Call FrameworkState.isStopping inside config code that uses OSGi
`com.ibm.wsspi.kernel.service.utils.FrameworkState.isStopping` can be used to guard against the OSGi Framework shutting down around us.

tevans78:
#build 
fixes #5472


tevans78:
PB is good, merging

PR #6250, opened 2019-01-18
tolerate empty array or empty string within array

njr-11:
Update our MP Concurrency implementation's code that processes config values from MP Config such that both empty array as well as size 1 array with empty string are valid and mean the same thing.  This is done in order to tolerate an ambiguity in the MP Config spec.


njr-11:
#build

njr-11:
#build

njr-11:
Build has 5 test failures, which are preexisting issues

PR #6269, opened 2019-01-22
New implementation of AsyncBulkheadState

### LINKED ISSUE #6147: FT 2.0: Correct semantics for Async Bulkhead

Azquelt:
Featues:
 - doesn't use PolicyExecutor
 - bulkhead permit must be released manually

Manual release of the bulkhead permit is an important feature which
allows the permit to be held even after the async task has returned.
This is to allow the release of the bulkhead reservation to be attached
to the completion of a CompletionStage returned by another component
(e.g. JAX-RS client) which is required by the spec.

Fixes #6147

### DIFF
--- a/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncCompletionStageExecutor.java
+++ b/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncCompletionStageExecutor.java
@@ -22,7 +22,7 @@ import com.ibm.ws.microprofile.faulttolerance.spi.FallbackPolicy;
 import com.ibm.ws.microprofile.faulttolerance.spi.MetricRecorder;
 import com.ibm.ws.microprofile.faulttolerance.spi.RetryPolicy;
 import com.ibm.ws.microprofile.faulttolerance.spi.TimeoutPolicy;
-import com.ibm.ws.threading.PolicyExecutorProvider;
+import com.ibm.ws.microprofile.faulttolerance20.state.AsyncBulkheadState.BulkheadReservation;
 
 /**
  * Executor for asynchronous methods which return a {@link CompletionStage}{@code<R>}
@@ -34,8 +34,8 @@ public class AsyncCompletionStageExecutor<R> extends AsyncExecutor<CompletionSta
     private static final TraceComponent tc = Tr.register(AsyncCompletionStageExecutor.class);
 
     public AsyncCompletionStageExecutor(RetryPolicy retry, CircuitBreakerPolicy cbPolicy, TimeoutPolicy timeoutPolicy, FallbackPolicy fallbackPolicy, BulkheadPolicy bulkheadPolicy,
-                                        ScheduledExecutorService executorService, PolicyExecutorProvider policyExecutorProvider, MetricRecorder metricRecorder) {
-        super(retry, cbPolicy, timeoutPolicy, fallbackPolicy, bulkheadPolicy, executorService, policyExecutorProvider, metricRecorder);
+                                        ScheduledExecutorService executorService, MetricRecorder metricRecorder) {
+        super(retry, cbPolicy, timeoutPolicy, fallbackPolicy, bulkheadPolicy, executorService, metricRecorder);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncCompletionStageExecutor.java
```diff
@@ -63,16 +63,16 @@ protected void commitResult(AsyncExecutionContextImpl<CompletionStage<R>> execut
     }
 
     @Override
-    protected void finalizeAttempt(AsyncAttemptContextImpl<CompletionStage<R>> attemptContext, MethodResult<CompletionStage<R>> result) {
+    protected void processMethodResult(AsyncAttemptContextImpl<CompletionStage<R>> attemptContext, MethodResult<CompletionStage<R>> result, BulkheadReservation reservation) {
```
Azquelt: Note switch from overriding `finalizeAttempt` to overriding `processMethodResult`.

`processMethodResult` occurs after the user's method has executed and has returned a result. When this happens, if the user's code has returned a CompletionStage, we want to defer processing until that CompletionStage is complete.

In contrast, `finalizeAttempt` is called when we've decided that the attempt is over for any reason (method complete, timeout, user cancellation, execution blocked by circuit breaker or bulkhead). Overriding this method was incorrect for several reasons:
1. The only situation where we want to defer processing is when the user's method completes
2. Releasing the bulkhead needs to occur when the method actually completes, not when the result is returned to the user (required by spec)

dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncExecutor.java
```diff
@@ -143,8 +143,7 @@ private void enqueueAttempt(AsyncExecutionContextImpl<W> executionContext) {
 
         timeout.start();
 
-        ExecutionReference ref = bulkhead.submit(handleExceptions(() -> runExecutionAttempt(attemptContext), attemptContext, executionContext),
-                                                 getExceptionHandler(attemptContext));
+        ExecutionReference ref = bulkhead.submit((reservation) -> runExecutionAttempt(attemptContext, reservation), getExceptionHandler(attemptContext));
```
Azquelt: Removal of `handleExceptions()` is due to `runExecutionAttempt` now doing its own exception handling, since it has slightly more to do to ensure that the bulkhead is always released.

dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/state/impl/AsyncBulkheadStateNullImpl.java
```diff
@@ -18,22 +18,23 @@
 public class AsyncBulkheadStateNullImpl implements AsyncBulkheadState {
```
Azquelt: Changes to this class just reflect the changes to the `AsyncBulkheadState` interface

Azquelt:
I've added a few notes (above) to explain why changes were made (where the current code makes sense, but it's important to understand why the previous code needed to change).

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
Failures in the build do not appear to be related to this change. Merging.

PR #6274, opened 2019-01-22
Globally apply ossindex plugin to all non-test projects

aguibert:
Currently every product-project needs to include this snippet in their build.gradle file:
```groovy
plugins {
  id "net.ossindex.audit" version "0.3.8-beta"
}

audit {
    failOnError = true
}
```

This should be specified once at a global level so that each project does not need to specify it, and developers don't need to remember to include it when writing new product projects.


aguibert:
#build

aguibert:
#build

PR #6277, opened 2019-01-22
clear cached classloaders when application is stopped

njr-11:
ConcurrencyProvider caches ConcurrencyManager instances based on the class loader (which is used as a key).  Need to clear entries from this cache when the application is stopped.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
build has 6 test failures which are duplicates of preexisting issues in unrelated test buckets

PR #6315, opened 2019-01-28
tests for security context

njr-11:
Write tests for security context propagation/clearing when using MicroProfile Concurrency.


njr-11:
#build

PR #6318, opened 2019-01-29
Increase the timeout of testMultiRequestBulkhead

Azquelt:
This test occasionally fails in the build due to the request taking too
long. The time taken is not critical to the correct running of the test
so we can improve reliability by increasing the timeout.


Azquelt:
#build

Issue #4945, opened 2018-09-04
Config: Run appropriate Config TCK with EE7 and EE8 features


Azquelt:
Config is included in Microprofile 1.4 and 2.0. We should run the TCK in both these scenarios.

Issue #4946, opened 2018-09-04
Run appropriate FT TCK with EE7 and EE8 features


Azquelt:
Fault Tolerance 1.1 is included in both Microprofile 1.4 and 2.0 so we should run the TCK with both EE7 and EE8 features.

PR #6322, opened 2019-01-29
4946 tck repeat

benjamin-confino:


benjamin-confino:
#build

benjamin-confino:
Fixes #4945 #4946

Issue #5785, opened 2018-11-15
CWWKS9582E: The [defaultSSLConfig] sslRef attributes required by the orb element with the defaultOrb id have not been resolved within 10 seconds.


dburdine:
During Server initialization the following error is seen:
CWWKS9582E: The [defaultSSLConfig] sslRef attributes required by the orb element with the defaultOrb id have not been resolved within 10 seconds.

dburdine:
The fix will be to enable the timeout value to be configurable. Currently it is hardcoded at 10 seconds.

dburdine:
PR#5788

dburdine:
Property orbSSLInitTimeout has been added. Here is an example setting the timeout to 33 seconds:
    <orb id="defaultOrb" orbSSLInitTimeout="33">
        <serverPolicy.csiv2>
            <layers>
                <attributeLayer identityAssertionEnabled="false"/>
                <authenticationLayer mechanisms="LTPA" establishTrustInClient="Never"/>
                <transportLayer sslRef="requiredClientAuthenticationSSLConfig" />
            </layers>
        </serverPolicy.csiv2>
    </orb>

toshiyamamoto:
The original fix was reverted, so this issue is reopened.

dburdine:
Resolved by PR 5983

PR #6328, opened 2019-01-29
Make ORB startup timeout value configurable

toshiyamamoto:
Occasionally, CWWKS9582E message which indicates that SSL is not initialized within 10 seconds is logged. Most of the time, it is due to the hardware performance. This PR intends to provide a way to modify the timeout value other than 10.


toshiyamamoto:
#build
fixes #5785

toshiyamamoto:
All errors in the personal build are known defects. No regression is found.

Issue #5119, opened 2018-09-18
Transaction Recovery Log bundle startup sequence needs refinement


neilgsyoung:
Specifically in the case where Tran Logs are being written to an RDBMS, I have found a window where recovery is attempted before the Database interface code is ready.

PR #6334, opened 2019-01-30
Enable waitForRecovery test with transaction bundle set to CONTAINER_LATE

neilgsyoung:
The WaitForRecoveryTest confirms that the waitForRecovery configuration option is honoured. Prior to issue #5119 's resolution waitForRecovery processing could hang. 

The combination of #5119 - which means that the OSGi DS services do not conflict - and the setting of CONTAINER_LATE for the com.ibm.ws.transaction bundle - which means that the Liberty bundles can start without conflict - allow waitForRecovery to work.


neilgsyoung:
#build

neilgsyoung:
#build

PR #6345, opened 2019-01-30
beans for container-produced ThreadContext and ManagedExecutor must not have an EL name

njr-11:
Per section 2.6.3 "Beans with no name" of the CDI 2.0 spec doc, there is no EL name for beans without the `@Named` annotation.  This pull updates our implementation to be consistent with this requirement.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
build has 2 test failures which are duplicates for preexisting issues in unrelated test buckets

PR #6356, opened 2019-01-30
Move JPA Spec 1.0 tests to a seperate test bucket.

jgrassel:

Updated Callback bucket to include EJB.  Updated buckets to run against EE7 and EE8.  Moved JPA 1.0 Spec buckets to their own test project.


jgrassel:
#build

PR #6357, opened 2019-01-30
Increase JobWaiter polling timeout to 30

scottkurz:
Just a FAT change to decrease false failures due to missing a timing window.


scottkurz:
#build

scottkurz:
@cgianfra  just be a bit more careful reviewing this, since there are tests in here where (tran) timeouts matter.   I don't think the JobWaiter timeout matters but appreciate your help proving I'm not full of it.  Thanks.

PR #6359, opened 2019-01-31
Port dual peer recovery tests from Commercial Liberty

neilgsyoung:
Add the dual server peer recovery tests.


neilgsyoung:
#build

Issue #6364, opened 2019-01-31
NPE possible in TrOSGiLogForwarder.logged if LogEntry.getBundle is null


tjwatson:
The following NPE is possible if `LogEntry.getBundle()` returns null

```
1/30/19 6:47:43:009 UTC] 0000001f SystemErr                                                    R LogListener.logged threw a non-fatal unchecked exception as follows:
[1/30/19 6:47:43:009 UTC] 0000001f SystemErr                                                    R java.lang.NullPointerException
[1/30/19 6:47:43:009 UTC] 0000001f SystemErr                                                    R   at java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:947)
[1/30/19 6:47:43:010 UTC] 0000001f SystemErr                                                    R   at com.ibm.ws.logging.internal.osgi.TrOSGiLogForwarder.getTraceComponent(TrOSGiLogForwarder.java:58)
[1/30/19 6:47:43:010 UTC] 0000001f SystemErr                                                    R   at com.ibm.ws.logging.internal.osgi.TrOSGiLogForwarder.logged(TrOSGiLogForwarder.java:99)
[1/30/19 6:47:43:010 UTC] 0000001f SystemErr                                                    R   at org.eclipse.osgi.internal.log.ExtendedLogReaderServiceFactory.safeLogged(ExtendedLogReaderServiceFactory.java:111)
[1/30/19 6:47:43:010 UTC] 0000001f SystemErr                                                    R   at org.eclipse.osgi.internal.log.ExtendedLogReaderServiceFactory.logPrivileged(ExtendedLogReaderServiceFactory.java:253)
[1/30/19 6:47:43:011 UTC] 0000001f SystemErr                                                    R   at org.eclipse.osgi.internal.log.ExtendedLogReaderServiceFactory.log(ExtendedLogReaderServiceFactory.java:225)
[1/30/19 6:47:43:011 UTC] 0000001f SystemErr                                                    R   at org.eclipse.osgi.internal.log.ExtendedLogServiceFactory.log(ExtendedLogServiceFactory.java:94)
[1/30/19 6:47:43:011 UTC] 0000001f SystemErr                                                    R   at org.eclipse.osgi.internal.log.LoggerImpl.log(LoggerImpl.java:80)
[...]

PR #6365, opened 2019-01-31
Fix NPE for when LogEntry.getBundle returns null

tjwatson:
For #6364 
#build


tjwatson:
This was merged as part of another PR

PR #6377, opened 2019-02-01
Revert "Move JPA Spec 1.0 tests to a seperate test bucket."

Azquelt:
Reverts OpenLiberty/open-liberty#6356

This bucket is too big and has timed out in the continuous build multiple times.

In your personal build, it took 54 minutes to complete (see the activity tab on child f12) and we timeout FAT buckets after an hour.

It looks like the PR is splitting up a bucket, but also adds RepeatTests to run all the tests twice? It's hard to see as there are so many files changed and github doesn't do a great job of showing them to you.


Azquelt:
Git on the command line shows the correct number of files, once the setting is added to allow it to find all the files which have moved.

PR #6386, opened 2019-02-01
precontextualized actions supplied to managed executor and completion stage

njr-11:
When a pre-contextualized action (for example, threadContext.contextualRunnable(action)) is supplied as the action for a ManagedExecutor or managed completion stage, the already-captured context from the action should be used rather than newly captured context according to the managed executor configuration.  However, the ManagedExecutor is still the default asynchronous execution facility and provides context propagation for other dependent actions which are not pre-contextualized.

Here is an example where 'supplier' and 'function2' run with context captured by 'myManagedExecutor', but 'function1' runs with context captured by 'myThreadContext',
```
myManagedExecutor
    .supplyAsync(supplier)
    .thenApplyAsync(myThreadContext.contextualFunction(function1))
    .thenApplyAsync(function2);
```


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build has 7 test failures, which are duplicates of preexisting issues in unrelated test buckets.

njr-11:
#build

PR #6393, opened 2019-02-04
Test fix: ensure builder timeout is reset after test that changes it.

andymc12:
Fix a test to reset the read timeout after a test lowers it.  This comes into play when the test cases are executed "out of order" and a non-timeout-related test attempts to run after the read timeout has been set to a low value.  The tests should not assume any ordering.  This fix ensures that the timeout gets reset properly.


andymc12:
#build

PR #6396, opened 2019-02-04
precontextualized tasks

njr-11:
When ThreadContext is used to precontextualized Runnable and Callables that are supplied to ManagedExecutor.submit/execute/invoke methods, or to ManagedScheduledExecutorService.schedule methods, ensure that the already-captured thread context is used rather than capturing context again per the managed executor's configuration.


njr-11:
#build

njr-11:
build has 14 test failures, which are duplicates of preexisting issues in unrelated test buckets

PR #6397, opened 2019-02-05
FT 2.0 - propegate thread context for asynchronous executions

Azquelt:
This ensures that when a method annotated with `@Asynchronous` is run, it has the correct security, metadata and classloader context.


Azquelt:
#build

dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AbstractFTTest.java
```diff
@@ -25,7 +29,9 @@
     @Before
     public void before() {
         System.setProperty(FTConstants.JSE_FLAG, "true");
-        FaultToleranceProviderResolver.setInstance(new ProviderResolverImpl20());
+        ProviderResolverImpl20 providerResolver = new ProviderResolverImpl20();
```
tevans78: Should this be @Before or @BeforeClass? Does it matter?

Azquelt: I think it could be either, as long is it matches up with the `after` method.

I don't think it's important as it's not doing anything expensive.

dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AbstractFTTest.java
```diff
@@ -25,7 +29,9 @@
...
+        setField(ProviderResolverImpl.class, "contextService", providerResolver, new DummyContextService());
```
tevans78: Wouldn't setters and getters on ProviderResolverImpl be better rather than trying to access protected fields like this? I think those fields should be private anyway

dev/com.ibm.ws.microprofile.faulttolerance.cdi_fat/publish/servers/FTFallback/server.xml
```diff
@@ -11,15 +11,17 @@
...
-	    <feature>componenttest-1.0</feature>
+		<feature>componenttest-1.0</feature>
 		<feature>osgiconsole-1.0</feature>
 		<feature>localConnector-1.0</feature>
 		<feature>appSecurity-3.0</feature>
 		<feature>cdi-2.0</feature>
 		<feature>servlet-4.0</feature>
 		<feature>mpFaultTolerance-2.0</feature>
+		<feature>ejblite-3.2</feature>
+		<feature>appSecurity-3.0</feature>
```
tevans78: I wonder if a separate server would be better rather than introducing ejb to an existing one?

dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AbstractFTTest.java
Azquelt: I don't think they're the fields you're thinking of.

These fields are annotated `@Reference` and populated with services provided by DS.

I think setting the references by reflection is an appropriate approach for a unit test. I'd prefer to avoid adding further boilerplate to the implementation where it isn't necessary.

dev/com.ibm.ws.microprofile.faulttolerance/src/com/ibm/ws/microprofile/faulttolerance/impl/ExecutorBuilderImpl.java
```diff
@@ -35,7 +35,7 @@
     protected FallbackPolicy fallbackPolicy = null;
     protected TimeoutPolicy timeoutPolicy = null;
     protected MetricRecorder metricRecorder = DummyMetricRecorder.get();
-    private final WSContextService contextService;
+    protected final WSContextService contextService;
```
tevans78: grump!

dev/com.ibm.ws.microprofile.faulttolerance.cdi_fat/publish/servers/FTFallback/server.xml
Azquelt: Mmm, possibly, though that would also mean separating out the tests into different classes.

I'm happy to do it, I didn't bother to in the first instance because I'm lazy.

tevans78:
split the test servers, otherwise ok

Azquelt:
@tevans78 servers have been split

Azquelt:
#build

Azquelt:

Azquelt:
#build

Azquelt:

PR #6422, opened 2019-02-05
reject double contextualization until spec is clarified

njr-11:
The MicroProfile Concurrency spec does not define a behavior for what happens when you attempt to re-contextualize an instance that has already been contextualized.  There is an issue open to determine this, but in case it doesn't not make it into the initial spec release, we want to have the flexibility of updating our implementation later, so the safest approach is to reject all attempts at double contextualization.


njr-11:
#build

njr-11:
build has 8 test failures which are duplicates of pre-existing issues in unrelated test buckets.
Also, one of the async test runs stopped responding, but I resubmitted it and then it ran cleanly.

PR #6432, opened 2019-02-06
FT 2.0: Fix async fallback

### LINKED ISSUE #5493: Fix asynchronous fallback handling in FT 2.0
Currently, the fallback method is called on the current thread at the point we decide we need to fall back.

Often this is fine as it's the thread that has just run the method that failed. However, in the case where we get a BulkheadException or a CircuitBreakerOpenException, this may be the calling thread which is not allowed.

Azquelt:
Ensure that the AsyncExecutor always runs the fallback method
asynchronously, even if the bulkhead is full.

Run fallback with the correct thread context.

Fixes #5493

### DIFF
--- a/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncExecutor.java
+++ b/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncExecutor.java
@@ -255,10 +255,10 @@ public abstract class AsyncExecutor<W> implements Executor<W> {
                 Thread.interrupted(); // Clear interrupted thread
             }
         } catch (Throwable t) {
-            // Handle any Unexpected exceptions
+            // Handle any unexpected exceptions
             // Manual FFDC since we've ignored it for the catch above
-            String sourceId = AsyncExecutor.class.getName() + "runExecutionAttempt";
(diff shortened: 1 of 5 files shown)


Azquelt:
#build

Azquelt:
#build

dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncExecutor.java
```diff
@@ -369,32 +367,60 @@ protected void finalizeAttempt(AsyncAttemptContextImpl<W> attemptContext, Method
...
+
+            executionContext.setFailure(failedResult.getFailure());
+
+            ThreadContextDescriptor contextDescriptor = executionContext.getThreadContextDescriptor();
+            ArrayList<ThreadContext> context = contextDescriptor.taskStarting();
+
+            MethodResult<W> fallbackResult;
+            try {
+                fallbackResult = MethodResult.success((W) fallbackPolicy.getFallbackFunction().execute(executionContext));
+            } catch (Throwable ex) {
```
tevans78: IgnoreFFDC??

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:

PR #6443, opened 2019-02-06
NullPointerException when backed by unmanaged executor

njr-11:
The following NullPointerException occurs (Java SE 8 path only) when ManagedCompletableFuture is backed by an unmanaged executor that was obtained via supplyAsync,

java.lang.NullPointerException
at com.ibm.ws.concurrent.mp.ContextualSupplier.get(ContextualSupplier.java:37)
at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1590)
at ...

Also, we don't have any test case for this code path, so I'm adding one.  This code path would only be used internally by Liberty when a completion stage is wanted with the Liberty thread pool as its default asynchronous execution facility.


njr-11:
#build

njr-11:
Build has 4 test failures, which are preexisting issues in unrelated test buckets

Issue #6437, opened 2019-02-06
Remove FFDC generation from JaxRsFactoryBeanEJBCustomizer to comply with the JAX-RS spec regarding ExceptionMapper.


jkoehler22:
Exception = java.lang.reflect.InvocationTargetException
Source = com.ibm.ws.jaxrs20.ejb.JaxRsFactoryBeanEJBCustomizer
probeid = 334
Stack Dump = java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:90)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:55)
	at java.lang.reflect.Method.invoke(Method.java:508)
	at com.ibm.ws.jaxrs20.ejb.JaxRsFactoryBeanEJBCustomizer.serviceInvoke(JaxRsFactoryBeanEJBCustomizer.java:333)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsServerFactoryBean.performInvocation(LibertyJaxRsServerFactoryBean.java:637)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsInvoker.performInvocation(LibertyJaxRsInvoker.java:160)
	at org.apache.cxf.service.invoker.AbstractInvoker.invoke(AbstractInvoker.java:96)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsInvoker.invoke(LibertyJaxRsInvoker.java:273)
	at org.apache.cxf.jaxrs.JAXRSInvoker.invoke(JAXRSInvoker.java:205)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsInvoker.invoke(LibertyJaxRsInvoker.java:444)
	at org.apache.cxf.jaxrs.JAXRSInvoker.invoke(JAXRSInvoker.java:111

PR #6444, opened 2019-02-06
Fix Issue #6437: remove FFDC

jkoehler22:


jkoehler22:
#build

dev/com.ibm.ws.jaxrs.2.0.ejb/src/com/ibm/ws/jaxrs20/ejb/JaxRsFactoryBeanEJBCustomizer.java
```diff
@@ -244,6 +244,7 @@ public boolean isCustomizableBean(Class<?> clazz, Object context) {
     }
 
     @Override
+    @FFDCIgnore(Exception.class)
```
andymc12:
```suggestion
    @FFDCIgnore({NamingException.class, Exception.class, Exception.class})
```

The `@FFDCIgnore` annotation scans through the method and looks for catch blocks based on the exception(s) in the value of the annotation.  It ignores based on matches in order, so for this method, it would only ignore the first catch block where it specifically catches an `Exception`.  This method has three catch blocks, first a `NamingException`, then two for `Exception`.  The FFDC reported by the user was from the last catch block, but we should probably ignore exceptions for all three - in the first catch block, we log the exception to trace, and in the second and third we re-throw the exception so it will be handled further upstream.

jkoehler22:
#build

jkoehler22:
None of the errors found doing the build are related to my change.  Merging...

Issue #6246, opened 2019-01-18
Apply "useAuthenticationDataForUnprotectedResource" to jwtSso cookie


chunlongliang-ibm:
We need apply "useAuthenticationDataForUnprotectedResource" to JWT SSO cookie, so it has the equivalent behavior as LTPA cookie

ayoho:
@teddyjtorres Can this be closed since the PR has been merged?

PR #6447, opened 2019-02-06
Use JWT SSO Cookie for unprotected resource if useAuthenticationDataForUnprotectedResource is enabled for issue #6246

teddyjtorres:
For issue #6246


teddyjtorres:
#build

teddyjtorres:
Note that this is a temporary solution. A generic framework using the UnprotectedResourceService should be considered.

teddyjtorres:
Errors in build are not related to the fix being delivered.

PR #6454, opened 2019-02-07
Add null check to each merged permission

james-ro-williams:
#6067 
First ensure the list of merged permissions isn't
empty before iterating over it. Then, due to the fact
that null can be added to a List, null check each
Permission.


james-ro-williams:
Changes tested by running ./gradlew build in order to perform unit tests.

james-ro-williams:
#build

PR #6460, opened 2019-02-07
Real flush

### LINKED ISSUE #6095: Ability to extend the size of the log buffer beyond 8k on WAS Liberty Profile
WAS Liberty Profile has an 8K limit for logs sent to console and it breaks it into 8K chunks. This limitation doesn't allow fluentd to scan and parse 8K+ log entries correctly.

see https://www.ibm.com/developerworks/rfe/execute?use_case=viewChangeRequest&CR_ID=115284 
see https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemName/com.ibm.team.workitem.WorkItem/250381

AidanPolese:
#build
Fixes #6095

### DIFF
--- a/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
+++ b/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
@@ -1200,72 +1200,42 @@ public class BaseTraceService implements TrService {
 
     public final static class TeePrintStream extends PrintStream {
         protected final TrOutputStream trStream;
+        protected final boolean autoFlush;
 
         public TeePrintStream(TrOutputStream trStream, boolean autoFlush) {
             super(trStream, autoFlush);
+            this.autoFlush = autoFlush;
             this.trStream = trStream;
         }
 
-        public void realFlush() {
-            try {
-                trStream.realFlush();
-            } catch (IOException e) {
-                e.printStackTrace();
+        public void realFlush(String s) {
+            if (autoFlush) {
+                try {
+                    trStream.realFlush();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
             }
+            if (autoFlush && (s.indexOf('\n') >= 0)) {
+                try {
+                    trStream.realFlush();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+            }
+
+        }
+
+        @Override
+        public void println(String s) {
+            super.print(s);
+            realFlush(s);
         }
 
         @Override
         public void print(String s) {
             super.print(s);
-            realFlush();
-        }
-
-        @Override
-        public void print(boolean b) {
-            super.print(b);
-            realFlush();
-        }
-
-        @Override
-        public void print(char c) {
-            super.print(c);
-            realFlush();
-        }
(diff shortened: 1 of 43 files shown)


dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -233,7 +233,7 @@ public BaseTraceService() {
         systemOut = new SystemLogHolder(LoggingConstants.SYSTEM_OUT, System.out);
         systemErr = new SystemLogHolder(LoggingConstants.SYSTEM_ERR, System.err);
 
-        earlyMessageTraceKiller_Timer.schedule(new EarlyMessageTraceCleaner(), 5 * MINUTE); // 5 minutes wait time
+        earlyMessageTraceKiller_Timer.schedule(new EarlyMessageTraceCleaner(), 1 * MINUTE); // 5 minutes wait time
```
donbourne: @AidanPolese , is change to 1 min temporary?  I'm guessing that's just to make your testing easier somehow.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -1200,11 +1200,102 @@ public boolean isEnabled() {
...
+                }
+            }
+            if (autoFlush && (s.indexOf('\n') >= 0)) {
+                try {
+                    trStream.realFlush();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
```
donbourne: this method looks like it's flushing twice in the case of \n.  check for \n should probably be in the print methods (which decide whether or not to call realFlush), not in the realFlush method itself.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -1200,11 +1200,102 @@ public boolean isEnabled() {
...
+        @Override
+        public void print(char c) {
+            super.print(c);
+            realFlush(String.valueOf(c));
+        }
+
+        @Override
+        public void print(int i) {
+            super.print(i);
+            realFlush(String.valueOf(i));
```
donbourne: another reason to check for \n in the print methods... here you're having to convert i to string even though you know it will never have a \n in it.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -1200,11 +1200,102 @@ public boolean isEnabled() {
...
             this.trStream = trStream;
         }
+
+        public void realFlush(String s) {
+            if (autoFlush) {
+                try {
+                    trStream.realFlush();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
```
donbourne: what happens when you call e.printStackTrace()?  could we end up in an infinite loop if the thing that caused the original exception doesn't clear before the prints from e.printStackTrace get here?

PR #6469, opened 2019-02-07
Add missing DDL files to JPA Spec 10 bucket for non-derby databases.

jgrassel:

Added missing DDL files for non-derby databases for inheritance, many-to-many, and one-to-one buckets.

There are no compilable or executable artifacts in this update.


jgrassel:
#build

PR #6472, opened 2019-02-07
Change to remove specific cached object from the configuration

acdemyers:


acdemyers:
#build

acdemyers:
The change has to do with clearing caches on when keystone file monitoring.   The SSL test cases do tests on file monitoring.

PR #6476, opened 2019-02-07
Add optional imports for sun.* packages

### LINKED ISSUE #6473: RemoteException thrown when sending objects over IIOP on Java 11
When using Liberty with Java 11, if there is a serialization-compatible change between the classes on the server and those on the client, objects of that class may not be transmitted correctly, and a `java.rmi.RemoteException` may be received instead.

chackoj:
fixes #6473


chackoj:
#build

chackoj:
These changes are only tangentially related to the error they were intended to fix — they do NOT fix the error.
There is no value in merging these changes on their own, so I am closing this PR.

PR #6483, opened 2019-02-08
update usage of MP Config to align with convention from MP Fault Tolerance

njr-11:
This pull switches us to the same pattern of using MP Config as MP Fault Tolerance for overriding annotation attributes of field injection points,

`org.example.MyBean.exec1.maxAsync=10`
-->
`org.example.MyBean/exec1/ManagedExecutorConfig/maxAsync=10`

Here is how we will cover method parameter injection points, although I did not see this explicitly addressed by Fault Tolerance convention,

`org.example.MyBean.setCompletableFuture.1.propagated=Security`
-->
`org.example.MyBean/setCompletableFuture/1/ManagedExecutorConfig/propagated=Security`


njr-11:
#build

Issue #6478, opened 2019-02-08
Invalid schema generated when an attribute references a beta element


alexsm82:
In #6434 I attempted to add a new properties.oracle.ucp element as beta, but it created errors in the schema validator and build.image profile tests.  Talking with Brent, it appears the issue is that we're generating an invalid schema because we include the attribute pointing to properties.oracle.ucp, but we aren't including the referenced type because it's marked beta.

PR #6487, opened 2019-02-08
Add a separate check for beta attributes to catch extended types that…

brenthdaniel:
… are marked as beta

PR for #6478


brenthdaniel:
#build

PR #6495, opened 2019-02-11
Update JDBC 4.1 FAT Java Permissions

alexsm82:
We should restrict the granting of the java permission for abort to the application to ensure we have regression testing against abort calls that initiate from Open Liberty itself.


alexsm82:
#build

PR #6256, opened 2019-01-21
6165 transaction scoped

benjamin-confino:

PR #6497, opened 2019-02-12
Revert 6256

Azquelt:
Reverts #6256 as it causes a leak of the application classloader

The TransactionContext holds a reference to the bean manager which will keep the app classloader alive.

The mistake appears not to be here, but in TransactionImpl where a static reference is kept to a single TransactionContext. This looks like it would cause problems anyway if multiple applications were used.


tevans78:
Merging without PB to get the OL builds moving again

PR #6498, opened 2019-02-12
Configure gradle build scan plugin

aalmiray:
This small PR configures the `build-scan` Gradle plugin (https://guides.gradle.org/creating-build-scans/) which yields further insight into the build, allowing the team to make informed decisions on where and when the build may be tweaked to get faster, more performant builds.

If this PR is accepted then scan generation must be enabled on the command line, like so

    ./gradlew build --scan

### DIFF
--- a/dev/.build-scans.gradle
+++ b/dev/.build-scans.gradle
@@ -1,0 +1,33 @@
+def acceptFile = new File(gradle.gradleUserHomeDir, "build-scans/open-liberty/gradle-scans-license-agree.txt")
+def env = System.getenv()
+boolean isCI = env.CI || env.TRAVIS
+boolean hasAccepted = isCI || env.GRADLE_SCANS_ACCEPT=='yes' || acceptFile.exists() && acceptFile.text.trim() == 'yes'
+boolean hasRefused = env.GRADLE_SCANS_ACCEPT=='no' || acceptFile.exists() && acceptFile.text.trim() == 'no'
+
+buildScan {
+    if (hasAccepted) {
+        termsOfServiceAgree = 'yes'
+    } else if (!hasRefused) {
+        gradle.buildFinished {
+            println """
+This build uses Gradle Build Scans to gather statistics, share information about 
+failures, environmental issues, dependencies resolved during the build and more.
+Build scans will be published after each build, if you accept the terms of 
+service, and in particular the privacy policy.
+
+Please read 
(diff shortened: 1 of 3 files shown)


aalmiray:
Ran the build locally by invoking `gradlew clean build --scan`.
This produced the following scan link https://gradle.com/s/ysn4gs22hmt26
Notice that there are test failures that are unrelated to the changes brought by this PR.

aguibert:
hi @aalmiray, thanks for the PR. One question I have on the ToS URL, what happens if that website cannot be reached? A lot of our builds run on internal networks and don't necessarily have access to that URL, and if they can, we don't want our builds to fail if that ToS URL is unreachable for some other reason.

aalmiray:
The build will have its outcome (success or failure) independent of the build-scan plugin. If the URL can't be reached then you'll get the following result.

Here's the output when the URL can't be reached

    BUILD SUCCESSFUL in 12s
    15 actionable tasks: 14 executed, 1 up-to-date

    The build scan was not published due to a configuration problem.

    The buildScan extension 'termsOfServiceUrl' value must be exactly the string 'https://gradle.com/terms-of-service' (without quotes).
    The value given was 'https://gradle.com/terms-of-service1'.

    To agree to the Gradle Terms of Service, include the following in your root project's configuration:
    buildScan { termsOfServiceUrl = 'https://gradle.com/terms-of-service'; termsOfServiceAgree = 'yes' }

    For more information, please see https://gradle.com/scans/help/plugin-terms-of-service.

    Alternatively, if you are using Gradle Enterprise, specify the server location.
    For more information, please see https://gradle.com/scans/help/plugin-enterprise-config.

aalmiray:
FWIW received a confirmation from Harry Hoots that the signed CLA has been processed.

fritze2:
Thanks for the CCE fix. I will merge after this passes an internal build.
#build

fritze2:
The prior build failed because this version of the plugin requires Gradle 5.x. Until this week we had a manual override setting internal builds to use Gradle 4.10.2. Internal builds are back to using Gradle 5.1.1, so I expect this to pass.
#build

aalmiray:
I see. Well if Gradle 4.x is desired then version `1.16` of the build-scan plugin should be used instead.

aalmiray:
Thanks! 😄

PR #6503, opened 2019-02-12
Add unit test to validate mavenCoordinates in feature manifests

### LINKED ISSUE #6501: Create test for Feature Manifest "mavenCoordinates"  to resolve dependencies
The "mavenCoordinates" directive in our feature Manifest's help map our liberty features to actual maven central downloads for our maven "bom" support.

We need to create a test that helps resolve these coordinates at buildtime

aguibert:
fixes #6501

To implement this properly, I had to write code that pings maven central to check if the dependency is available. The net is that every time we run `:com.ibm.websphere.appserver.features:test`, we will ping maven central ~300 times (once for each unique mavenCoord).

The 300 pings don't add significant time to the build (less than 10s on my machine), but it still seems like a heavyweight check, considering people have been pretty careful about using correct mavenCoords so far (we've only had 1 bad mavenCoordinate sneak in).

Also note that if for some reason maven central cannot be reached, the test will not fail.


aguibert:
#build

cbridgha:
@aguibert is there a way to set this up to run only on Integration / release builds?  
I agree with you it seems to be a heavyweight test - I was going to suggest this be part of the feature checker task - but actually that also runs in PB's...

aguibert:
we want release / integration / personal builds to be as in-sync as possible. It would be bad if this didn't run in PBs, because then people might deliver hard breaks that are missed by PBs.

aguibert:
Since the issue of bad mavenCoordinates has only happened once, and validating them would require >300 maven pings per build (local or remote), the burden on public infra seems to outweigh the benefits.
Closing this PR without a merge for now. We can always re-visit later if bad mavenCoordinates in feature manifests become problematic.

PR #6511, opened 2019-02-13
Don't perform server dump all the time in plugin.utility_fat

aguibert:
On remote builds, taking a server dump takes about 5 minutes, along with producing ~25MB of compressed data. The test infrastructure will already automatically take a server dump if some catastrophic failure occurs, so it's not necessary for a FAT to be taking server dumps on every run.


aguibert:
#build

PR #6512, opened 2019-02-13
Disable dumps in JPA Spec10 Bucket

jgrassel:

Disabling the dumps that are always requested by the jpa spec10 fat, jpa spec20 fat, and jpa_22 fat buckets.


jgrassel:
#build

PR #6514, opened 2019-02-13
Anno release

tbitonti:


cbridgha:
Hey @tbitonti this PR is setting off alarms. :). Please cleanup the IBM copyrights, and do we really need to check in 100's of jar's for the FAT's?   We need to look at storing these in artifactory.

bsbyrd1:
I'm adding the ID Review Not Needed tag to this PR. I don't see where any new messages have been added. If I'm wrong, please point out where an ID Review is needed. Thanks!

skasund:
L2 has reviewed the messages

PR #6520, opened 2019-02-13
Correctly handle case where we need to wait for multiple updates

brenthdaniel:
An app manager test can fail if the file update monitor picks up a change while we are still unzipping files. We have code to handle this case that attempts to wait for a second update to the application before running the test, but it doesn't work as intended. The log offset is not updated by the initial search for the "application updated" message, so the test finds the first "updated" message again rather than waiting for a new one.


brenthdaniel:
#build

PR #6492, opened 2019-02-11
Publish runtime, javaee8, and webprofile8 images from release builds

naumanna:
To be able to run daily Docker image builds for Open Liberty (https://github.com/OpenLiberty/ci.docker.daily) based on the latest release build output on DHE (public.dhe.ibm.com/ibmdl/export/pub/software/openliberty/runtime/nightly/), we need to publish the runtime, javaee8, and webProfile8 images from the release builds, since those images are used to generate the official Open Liberty Docker images.

PR #6528, opened 2019-02-14
Publish runtime, javaee8, and webProfile8 images to DHE

naumanna:
Redo of the fix in #6492, which did not work and had to be reverted.  This change will publish the runtime, javaee8, and webProfile8 images to DHE from release builds only.  It is needed for the daily Open Liberty Docker images to run properly.

### DIFF
--- a/dev/build.image/build.gradle
+++ b/dev/build.image/build.gradle
@@ -282,6 +282,11 @@ task zipOpenLiberty(type: Zip) {
     from packageDir
     destinationDir distsDir
     version releaseVersion
+    doLast{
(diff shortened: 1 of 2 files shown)


naumanna:
#build

naumanna:
Awesome, the extra image files were published as expected to DHE - http://public.dhe.ibm.com/ibmdl/export/pub/software/openliberty/runtime/personal/_SeDMQDCKEemMuOyAk-O3PQ/

I'm going to remove the test only changes to publish those files from personal builds and then ask for a review.

PR #6553, opened 2019-02-15
When ucp is specified with ConnectionPoolDataSource, do not search ba…

alexsm82:
…sed on classname or infer from driver name.  Otherwise we will return an impl of ConnectionPoolDataSource from the Oracle JDBC driver, when the user configured UCP.


alexsm82:
#build

alexsm82:
Both failures match to existing known defects.

PR #6570, opened 2019-02-15
added trace specification

AidanPolese:
added tracing specs to help debug


dev/com.ibm.ws.microprofile.metrics.monitor_fat/publish/servers/MetricsMonitorServer/server.xml
```diff
@@ -6,6 +6,8 @@
     	<feature>mpMetrics-1.1</feature>
     </featureManager>
 
+    <logging traceSpecification="com.ibm.ws.config.*=all;logservice=all"/>
```
fmhwong: From Knowledge Center, "If a configuration property is specified in both the bootstrap.properties file and the server.xml file, the value in bootstrap.properties is used until the server.xml file is processed. Then, the value in the server.xml file is used. Avoid specifying different values for the same configuration property in both the bootstrap.properties and the server.xml file."

PR #6582, opened 2019-02-18
Issue 6583: Update autoExtract to not re-extract with no changes.  AutoExtract co…

tbitonti:
See issue #6583: AutoExpand does not respect time stamps.

Note: This update adds a retry (for stability), but does **not** enable re-use of the extracted files across server restarts.  That function is present, but is currently disabled by an extra check.  At the least, enabling reuse across server starts impacts FATS, including not yet delivered annotation caching FATs which have code to work-around the extraction behavior.


tbitonti:
#build

tbitonti:
#build

tbitonti:
The update does rely on the time stamp of the un-expanded archive.  Without additional information (e.g., a sha file along side the archive, which would be very nice to have) that's the way this would have to work.

The expanded archive timestamp is updated (by an explicit call to File.setLastModified) to be exactly the same as the timestamp of the un-expanded archive.

Prior to this update, autoExpand would expand **always** after a server restart (because the time stamp wasn't recorded internally, which turned into an "always different" result), and would be expanded **sometimes** after an application restart, based on whether the expanded timestamp was the same or different than the internally recorded time stamp.  That seems to already cause one of the behaviors that you note: A change to a file in the expanded application folder will cause a new expansion.

There are at least two (three, when annotation caching is added) places we rely on time stamps to invalidate expanded resources. 

See:
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.app.manager.springboot/src/com/ibm/ws/app/manager/springboot/internal/SpringBootApplicationImpl.java
around line 612:
                if (thinSpringAppFile.lastModified() != springAppFile.lastModified()) {

See also:
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/ZipFileContainer.java
Around line 1429:
                    if ( isModified(inputEntry, outputFile) ) {
(Curiously, that 'isModified' only detects a difference if the last modified times are more than a second apart:
    private boolean isModified(ArtifactEntry entry, File file) {
        long fileLastModified = FileUtils.fileLastModified(file);
        long entryLastModified = entry.getLastModified();
        return ( Math.abs(fileLastModified - entryLastModified) >= 1010L );
    }
)

The main problem that I'm seeing is the case that you point out: Making an update to the un-expanded archive without updating the last updated timestamp.  I know that this definitely causes some annotation caching test code (which includes cases where autoExpand is true) to break, at least as currently written.

This issue became of note because we noticed during performance testing of the annotation caching code that around 10% of the time being spend doing extractions, for some of the test applications.

tbitonti:
#build

tbitonti:
Hi; have updated the change to revert to the prior time stamp checking.

The important remaining change is the addition of the retry loop for the delete step.  This no longer resolves the issue (which is looking to be a bigger discussion), but it should help with test stability.

tbitonti:
#build

tbitonti:
Failures prior to the merge with integration:

com.ibm.ws.classloading_fat FAT tests 113 1 0
  com.ibm.ws.classloading.UpdatingAppClassesViaMBeanFatTest 12 1 0
    testUpdatePostConstructAnnotationInEJB_loose 1 1 0
com.ibm.ws.jsonb_fat FAT tests 34 0 1
  junit.framework.TestSuite 1 0 1
    com.ibm.ws.jsonb.fat.JSONBContainerTest 1 0 1

tbitonti:
Not finding these as known problems; retrying with new merge of integration.

tbitonti:
#build

tbitonti:
The failure in:
testUpdatePostConstructAnnotationInEJB_loose

Is associated with pull-request #6510
https://github.com/OpenLiberty/open-liberty/pull/6510

No issue seems to have been created (yet) for the failure, but there is an RTC build break defect:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=261136

tbitonti:
Still seeing a number of failures; rebasing on integration an retrying.

tbitonti:
#build

tbitonti:
Last build completely fell over.  Retrying.

tbitonti:
#build

PR #6584, opened 2019-02-18
create nonship configValidator feature, porting over the Config REST API handler first

njr-11:
Create a nonship feature under which we can port the config/validator REST APIs that have been in beta for a while, and further experiment with them without making any of the function GA.  This nonship feature might end up being used for a beta in open liberty, or might eventually be removed in favor of an approach using autofeatures.


njr-11:
#build

njr-11:
The ID review should be unnecessary at this time given that the feature created under this pull is a nonship feature with the sole purpose of keeping the function out of GA as we experiment with it.

skasund:

njr-11:
#build

Issue #6590, opened 2019-02-19
NPE occurs when having an invalid entry in the web-ext


bpaskin-ibm:
When there is an invalid moduleName in web-ext within an application entry causes the application to fail to start and throws an NPE.  This is occurring in 19.0.0.1, but was ignored in 16.0.0.4.

```
   <application location="someear.ear" id="ear" type="ear">
        <web-ext moduleName="invalidname.war" context-root="/mycontext" />
   </application>
```

This would lead to ` E CWWKZ0002E: An exception occurred while starting the application ent-submission-tool-ear. The exception message was: com.ibm.ws.container.service.metadata.MetaDataException: java.lang.NullPointerException` in the messages.log and the following in the FFDC:

```
Caused by: java.lang.NullPointerException
        at com.ibm.ws.javaee.ddmodel.webext.WebExtAdapter.getConfigOverrides(WebExtAdapter.java:132)
        at com.ibm.ws.javaee.ddmodel.webext.WebExtAdapter.adapt(WebExtAdapter.java:61)
        at com.ibm.ws.javaee.ddmodel.webext.WebExtAdapter.adapt(WebExtAdapter.java:35)
        at com.ibm.ws.adaptable.module.internal.AdapterFactoryServiceImpl.adapt(AdapterFactoryServiceImpl.java:196)
        at com.ibm.ws.adaptable.module.internal.AdaptableContainerImpl.adapt(AdaptableContainerImpl.java:170)
        at com.ibm.ws.adaptable.module.internal.InterpretedContainerImpl.adapt(InterpretedContainerImpl.java:203)
        at com.ibm.ws.webcontainer.osgi.container.config.WebAppConfigurator.<init>(WebAppConfigurator.java:197)
        at com.ibm.ws.webcontainer.osgi.container.config.WebAppConfigurationAdapter.adapt(WebAppConfigurationAdapter.java:111)
        at com.ibm.ws.webcontainer.osgi.container.config.WebAppConfigurationAdapter.adapt(WebAppConfigurationAdapter.java:43)
        at com.ibm.ws.adaptable.module.internal.AdapterFactoryServiceImpl.adapt(AdapterFactoryServiceImpl.java:196)
        at com.ibm.ws.adaptable.module.internal.AdaptableContainerImpl.adapt(AdaptableContainerImpl.java:170)
        at com.ibm.ws.adaptable.module.internal.InterpretedContainerImpl.adapt(InterpretedContainerImpl.java:203)
[...]

PR #6605, opened 2019-02-19
Adjust web extension module name validation.

tbitonti:
For issue #6590 NPE occurs when having an invalid entry in the web-ext

### DIFF
--- a/dev/com.ibm.ws.javaee.ddmodel/src.gen/com/ibm/ws/javaee/ddmodel/webext/WebExtAdapter.java
+++ b/dev/com.ibm.ws.javaee.ddmodel/src.gen/com/ibm/ws/javaee/ddmodel/webext/WebExtAdapter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 IBM Corporation and others.
+ * Copyright (c) 2017,2019 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


tbitonti:
#build

tbitonti:
#build

tbitonti:
These tests are known unrelated failures:
com.ibm.ws.jmx.connector.client.rest_fat FAT tests | 303 | 0 | 38
com.ibm.ws.transaction_fat_java7 FAT tests | 467 | 0 | 1

The relationship of this test is unclear:
com.ibm.ws.microprofile.rest.client12_fat_tck FAT tests | 106 | 1 | 0

dev/com.ibm.ws.javaee.ddmodel/src.gen/com/ibm/ws/javaee/ddmodel/webext/WebExtAdapter.java
```diff
@@ -33,117 +39,300 @@
...
                return null;
-          appInfo = moduleInfo.getApplicationInfo();
-     }
-     NestedConfigHelper configHelper = null;
-     if (appInfo != null && appInfo instanceof ExtendedApplicationInfo)
-          configHelper = ((ExtendedApplicationInfo) appInfo).getConfigHelper();
-      if (configHelper == null)
+            }
+            appInfo = moduleInfo.getApplicationInfo();
+            if ( appInfo == null ) {
```
brenthdaniel: So this means if there is no application.xml we won't return a config override for the web module?

tbitonti: That 'applicationInfo' is not the same as the application descriptor.  Later on, the application descriptor is retrieved (line 300 in the update code) with:

Application appDD = appInfo.getContainer().adapt(Application.class);

The (appInfo == null) check was previously next to the type check.  I moved the check to more clearly indicate that the goal was obtain application information.

Original lines 92-97:

          appInfo = moduleInfo.getApplicationInfo();
     }
     NestedConfigHelper configHelper = null;
     if (appInfo != null && appInfo instanceof ExtendedApplicationInfo)
          configHelper = ((ExtendedApplicationInfo) appInfo).getConfigHelper();
      if (configHelper == null)

There is a lot of particular code in this method which has to do with different cases of obtaining module information and application information, but I don't have any explanation other than what is in the code: Whether module information is available impacts both the override selection, and impacts whether the name matching validation is done.  But there is no explanation of when the module information would or would not be null.

tbitonti:
#build

tbitonti:
Agreed on the problem of extensions missing a name: We won't discover them if they occur at the end of the list, since we stop looping once we find a match.

PR #6610, opened 2019-02-19
fix lambda bug and some minor changes.

toshiyamamoto:
this modification contains:
1) fix test case error in com.ibm.ws.kernel.boot_fat when serialFilter is enabled.
2) fix serialFilter error when lambda expression is used.
3) remove duplicated messages when discovery mode is used.
4) set default value for mode and permission configuration attributes.


toshiyamamoto:
#build

toshiyamamoto:
#build

PR #6612, opened 2019-02-20
avoid warning for feature version that does not exist in OpenLiberty

njr-11:
Update feature file so that it doesn't default to servlet-3.0 because there is no such version in OpenLiberty.  This is causing a warning to appear in server logs when the validator feature is enabled.


njr-11:
#build

njr-11:
closing this out because it was included under #6622

PR #6615, opened 2019-02-20
update message file, comments only

brutif:


brutif:
#build

brutif:
WebSphere Liberty corresponding change set is 15570.

brutif:
pb is good

PR #6619, opened 2019-02-20
Make FaultTolerance 2.0 depend on CDI-2.0

Azquelt:
Previously it used an autofeature for CDI integration. However, there's
no way of using Fault Tolerance without using CDI and we depend
specifically on CDI-2.0, so it makes more sense for this to be a hard
dependency.


Azquelt:
#build

Azquelt:
Errors in the build do not appear to be related to this change

PR #6620, opened 2019-02-20
tolerate Java 12 CompletableFuture methods

njr-11:
Java 12 adds 5 new methods to CompletionStage, which are implemented on CompletableFuture, which our implementation extends.  We should either provide a valid implementation or reject as unsupported if unable to provide a valid implementation now.  Open Liberty isn't currently compiled against Java 12, so would need to use reflection.

From the CompletionStage interface:
```
CompletionStage<T> exceptionallyAsync​(Function<Throwable,​? extends T> fn)
CompletionStage<T> exceptionallyAsync​(Function<Throwable,​? extends T> fn, Executor executor)
CompletionStage<T> exceptionallyCompose​(Function<Throwable,​? extends CompletionStage<T>> fn)
CompletionStage<T> exceptionallyComposeAsync​(Function<Throwable,​? extends CompletionStage<T>> fn)
CompletionStage<T> exceptionallyComposeAsync​(Function<Throwable,​? extends CompletionStage<T>> fn, Executor executor)
```


njr-11:
#build

PR #6630, opened 2019-02-21
port tests for config REST API

njr-11:
Port tests to OpenLiberty that cover the config REST API.  A test involving a user feature with configuration is currently disabled and will require the additional porting of the associated user feature before it can be turned back on.


njr-11:
#build

njr-11:
Closing this because the same commit is included under #6642 due to the stream being locked and changes stacking up.

PR #6631, opened 2019-02-21
add trace to connection pool mbeans

jmstephensgit:
Add some additional trace for mbeans showPoolConents and purgePoolConents.

When using scripts its easy to have typos or syntax errors that will be easier to fix with the additional trace.

### DIFF
--- a/dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/PoolManagerMBeanImpl.java
+++ b/dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/PoolManagerMBeanImpl.java
@@ -104,16 +104,28 @@ public class PoolManagerMBeanImpl extends StandardMBean implements ConnectionMan
     @Override
     public Object invoke(String actionName, Object[] params, String[] signature) throws MBeanException, ReflectionException {
         if (TraceComponent.isAnyTracingEnabled() && tc.isEntryEnabled())
-            Tr.entry(this, tc, "invoke");
+            Tr.entry(this, tc, "invoke", params, signature);
 
         Object o = null;
 
         if ("showPoolContents".equalsIgnoreCase(actionName)) {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "Displaying the connection pool");
+            }
             o = showPoolContents();
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "Connection pool contents", o);
+            }
         } else if ("purgePoolContents".equalsIgnoreCase(actionName)) {
             if (params == null) {
+                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


jmstephensgit:
#build

dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/PoolManagerMBeanImpl.java
```diff
@@ -104,16 +104,28 @@ public void unregister() {
     @Override
     public Object invoke(String actionName, Object[] params, String[] signature) throws MBeanException, ReflectionException {
         if (TraceComponent.isAnyTracingEnabled() && tc.isEntryEnabled())
-            Tr.entry(this, tc, "invoke");
+            Tr.entry(this, tc, "invoke", params, signature);
```
aguibert: I think it would be good to trace the `actionName` here at some point also, otherwise we won't get the method name

dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/PoolManagerMBeanImpl.java
```diff
@@ -104,16 +104,28 @@ public void unregister() {
...
+            }
         } else if ("purgePoolContents".equalsIgnoreCase(actionName)) {
             if (params == null) {
+                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                    Tr.debug(tc, "Normal purge of the connection pool");
+                }
                 purgePoolContents("normal");
             } else if (params.length == 1 && params[0] instanceof java.lang.String) {
+                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                    Tr.debug(tc, "Purging the connection pool using option", (String) params[0]);
```
aguibert: for single variable trace statements, string concatenation looks a little nicer because the trace statement stays all on one line. If we pass in variables on the last arg of `Tr.debug()` it ends up on a newline which is clunkier to read

dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/PoolManagerMBeanImpl.java
jmstephensgit: The actionName is covered by an exception, which is easy to debug,  throw new MBeanException(new IllegalArgumentException(actionName), Tr.formatMessage(tc, "INVALID_MBEAN_INVOKE_ACTION_J2CA8061", actionName, "ConnectionManager"));

dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/PoolManagerMBeanImpl.java
jmstephensgit: This is a personal preference since I see it both ways in code.   I normally go with the string concatenation, but was not sure what values will actually show up for the parms.   

Recently this was being used first parm "${server.output.dir}/server.xml" which of course was not working.

Leaving it on the next line in this case.

PR #6635, opened 2019-02-22
disable testAsyncClassLevelTimeout

benjamin-confino:


benjamin-confino:
#build fixes https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=259667

Azquelt:
I think the change to add the `doPriviliged` call makes sense, but we probably want to run it past someone who is more familiar with that area.

Azquelt:
Actually reading it again, I'm not sure whether that commit was meant to end up in this PR at all? It seems to be addressing an unrelated problem.

PR #6616, opened 2019-02-20
FT 2.0 FATs for CompletionStage and interceptor interactions

Azquelt:
Test some functionality that isn't checked by the TCK.

Fixes #6592 
Fixes #6595

PR #6636, opened 2019-02-22
Add tests for FT 2.0 + JAX-RS 

### LINKED ISSUE #6594: FT 2.0: Test Fault Tolerance with JAX-RS reactive APIs
Fault Tolerance support for `@Asynchronous` and `CompletionStage` should allow fault tolerance to be applied around JAX-RS reactive operations, which wasn't previously possible.

This is a valuable integration point and should be explicitly tested.

Azquelt:
Only the last commit is part of this change. The other two will be removed when #6616 is merged.

Fixes #6594


Azquelt:
#build

Azquelt:
Rebased on integration, no other changes

PR #6644, opened 2019-02-22
test coverage for TransactionScoped bean

njr-11:
Add test coverage for maintaining the state of a TransactionScoped bean when CDI context is propagated vs cleared.  TransactionScoped follows the transaction that is on the thread, not the propagated CDI context or lack thereof.


njr-11:
#build

njr-11:
#build

PR #6696, opened 2019-02-26
jsonp-1.1 should depend on json-api 1.1.3

### LINKED ISSUE #6255: jsonp-1.1 api dependencies incorrect
Note, this issue may also apply to jsonp-1.0.

I should be able to depend on the jsonp-1.1 feature like this:

```
        <dependency>
            <groupId>io.openliberty.features</groupId>
            <artifactId>jsonp-1.1</artifactId>
            <type>esa</type>
            <scope>provided</scope>
        </dependency>
```

And have it pull in the javax.json API automatically for me.  Unfortunately, the jsonp-1.1 feature depends on version 1.0 of the API artifact which is from 2013 and does not contain the correct level of the jsonp APIs (e.g. I can't use `javax.json.stream.JsonCollectors` because that came in a later version.).

To work around this issue, I have to add this dependency.

```
        <dependency>
            <groupId>javax.json</groupId>
            <artifactId>javax.json-api</artifactId>
            <version>1.1.4</version>
            <scope>provided</scope>
        </dependency>
```

ericglau:
Fixes #6255


ericglau:
#build

ericglau:
Build has the maven artifact with the updated version. Test errors are not related.

Issue #6428, opened 2019-02-06
Extraneous message with Rest Client 1.1 in 19.0.0.1


mbroz2:
@kwsutter commented on [Tue Feb 05 2019](https://github.com/OpenLiberty/openliberty.io/issues/962)

I am getting the following extraneous message logged when using mpRestClient-1.1 in 19.0.0.1:

```
[2/5/19 16:30:24:615 CST] 00000026 .cxf.rt.rs.mp.client.3.2:1.0.24.cl190120190124-2339(id=195)] W Could not load provider, , configured for Rest Client interface, eu.agilejava.dukes.weather.WeatherService 
```

When I run with 18.0.0.4, I don't receive that message.  I have talked this over with @andymc12 and he has reproduced it.  (He actually has a patch for it already.)

PR #6421, opened 2019-02-05
Avoid warning when no providers are specified for MP Rest Client

andymc12:
Resolves this warning message when no providers are specified via MP Config:
```
[2/5/19 16:30:24:615 CST] 00000026 .cxf.rt.rs.mp.client.3.2:1.0.24.cl190120190124-2339(id=195)] W Could not load provider, , configured for Rest Client interface, io.mycompany.MyClient 
```

This resolves issue [962](https://github.com/OpenLiberty/openliberty.io/issues/962).

PR #6700, opened 2019-02-26
Avoid warning when no providers are specified for MP Rest Client 1.2

andymc12:
This PR is a follow-on to PR #6421 - part of issue #6428.  The fix in 6421 only dealt with MP Rest Client 1.0 and 1.1, not 1.2 (which is not yet shipped).  The problem is fixed in CXF, but after the 3.3.0 release, so we need a temporary solution for this until we can pull in CXF 3.3.1 or later.

Since MP Rest Client 1.2 has not yet shipped, this is not a release bug.


andymc12:
#build

andymc12:
Lots of failures in that build, but they seem to be a server startup problem unrelated to my changes.  Normally, I might merge, but there were so many failures that it's possible that the startup issues might be hiding a potential issue of this change.  Will submit a new #build.

PR #6710, opened 2019-02-27
Improve tracing of Fault Tolerance

### LINKED ISSUE #6526: FT 2.0: Trace reason for not retrying
Currently we have fairly good trace for most events which occur during fault tolerance processing, including whether or not we retry. However, we don't currently trace the reason why we didn't retry, which would have been useful to me recently.

### LINKED ISSUE #6664: FT 2.0: Use better Ids for tracing calls
In our trace messages, we currently use the fully qualified method name to identify the execution which is taking place.

This has two downsides:
1) The fully qualified method name is very long
2) The trace is very confusing when the same method is being run multiple times concurrently

It might be better to assign to random identifier at the start of each invocation, have the first trace message include both the ID and the fully qualified method name, and then just use the identifier in future trace messages.

Azquelt:
Generates a random ID for each execution, resulting in shorter messages
which are easier to track when there are multiple executions running
concurrently.

For FT 1.x, change from using a UUID to using a shorter pseudorandom
string

Also trace out the reason why when a retry does not occur.

Fixes #6664 
Fixes #6526

### DIFF
--- a/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncExecutor.java
+++ b/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncExecutor.java
@@ -134,12 +134,12 @@ public abstract class AsyncExecutor<W> implements Executor<W> {
 
     @Override
     public W execute(Callable<W> callable, ExecutionContext context) {
-        if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
-            Tr.event(tc, "Fault tolerance asynchronous execution started for {0}", context.getMethod());
-        }
-
         @SuppressWarnings("unchecked")
         AsyncExecutionContextImpl<W> executionContext = (AsyncExecutionContextImpl<W>) context;
+
+        if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
+            Tr.event(tc, "Execution {0} Fault tolerance asynchronous execution started for {1}", executionContext.getId(), context.getMethod());
+        }
 
         executionContext.setCallable(callable);
 
@@ -179,7 +179,7 @@ public abstract class AsyncExecutor<W> implements Executor<W> {
      */
     private void prepareExecutionAttempt(AsyncExecutionContextImpl<W> executionContext) {
         if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
-            Tr.event(tc, "Method {0} enqueuing new execution attempt", executionContext.getMethod());
+            Tr.event(tc, "Execution {0} enqueuing new execution attempt", executionContext.getId());
         }
 
         AsyncAttemptContextImpl<W> attemptContext = new AsyncAttemptContextImpl<>(executionContext);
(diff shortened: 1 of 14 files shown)


Azquelt:
#build

dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncExecutor.java
```diff
@@ -246,14 +246,14 @@ private void runExecutionAttempt(AsyncAttemptContextImpl<W> attemptContext, Bulk
             }
 
             if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
-                Tr.event(tc, "Method {0} attempt execution reuslt: {1}", executionContext.getMethod(), methodResult);
+                Tr.event(tc, "Execution {0} attempt execution reuslt: {1}", executionContext.getId(), methodResult);
```
tevans78: "reuslt"

tevans78: Also, do you need to say "execution" twice? Maybe just "Execution {0} result: {1}"?

dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/state/impl/RetryStateImpl.java
```diff
@@ -55,31 +55,31 @@ public RetryResult recordResult(MethodResult<?> methodResult) {
...
             }
         } else {
             // Successful case
-            result.shouldRetry = false;
+            result.reason = RetryResultReason.NO_EXCEPTION;
         }
 
         // Capture whether this result was considered a failure by the Retry
-        boolean resultWasFailure = result.shouldRetry;
+        boolean resultWasFailure = result.shouldRetry();
```
tevans78: you're asking the result if it thinks it should retry and then inferring that the result was a failure. Perhaps the method should be `result.wasFailure()`?

dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/state/impl/RetryStateImpl.java
```diff
@@ -171,6 +171,31 @@ public long getDelay() {
         public TimeUnit getDelayUnit() {
             return delayUnit;
         }
+
+        @Override
+        public String toString() {
+            return reason.toString();
```
tevans78: `reason` is not guaranteed to be non-null, you can't use it like this in a toString

dev/com.ibm.ws.microprofile.faulttolerance.cdi/src/com/ibm/ws/microprofile/faulttolerance/cdi/FaultToleranceInterceptor.java
```diff
@@ -242,6 +244,15 @@ private Object execute(InvocationContext invocationContext, AggregatedFTPolicy a
         return result;
     }
 
+    private String generateId(Method method) {
```
tevans78: `@Trivial`?

Azquelt:
squashed review changes

Azquelt:
#build

Azquelt:
Last build failed with an infrastructure communication failure.

Azquelt:
#build

Azquelt:

PR #6720, opened 2019-02-27
Add other Oracle UCP vendor props

alexsm82:
Add the Oracle UCP vendor config that was left out of the first pull request pending further investigation and testing.


alexsm82:
#build

alexsm82:
#build

alexsm82:
Adding ID Review Not Needed label as we are only adding a JDBC driver vendor property which follows a standard format and Brandy said one wasn't needed the last time I added one.

alexsm82:
The test failures in the build are all unrelated, known defects.

PR #6747, opened 2019-03-01
Upgrade Felix DS to OSGi R7

brenthdaniel:
Upgrade Felix SCR to a new version based on OSGi R7


dev/com.ibm.ws.logging.osgi/src/com/ibm/ws/logging/internal/osgi/TrOSGiLogForwarder.java
```diff
@@ -128,12 +131,16 @@ public void logged(LogEntry le) {
...
-                        Tr.info(tc, "OSGI_MSG001", getObjects(logEntry, true));
+                        if(shouldBeLogged(logEntry, tc)) {
+                            Tr.info(tc, "OSGI_MSG001", getObjects(logEntry, true));
+                        }
                     }
                     break;
     
                 case WARN:
-                    Tr.warning(tc, "OSGI_WARNING_MSG", getObjects(logEntry, true));
+                    if(shouldBeLogged(logEntry, tc)) {
```
sebratton: I don't think this requires a change prior to merging, since I don't have any sense that this is impactful to performance. But perhaps later add an isWarnEnabled guard before calling shouldBeLogged since that method causes additional object to be created even if WAR disabled.

sebratton:
#build

PR #6751, opened 2019-03-01
start of autofeature where we can put OpenAPI discovery

njr-11:
Check in the start of this autofeature so that other team members can pick up on this work.


njr-11:
#build

PR #6727, opened 2019-02-28
JSP slash slash comment fix

epj:
fixes #6725

PR #6767, opened 2019-03-04
Revert "JSP slash slash comment fix"

ayoho:
Reverts OpenLiberty/open-liberty#6727


ayoho:
The original PR makes use of the `\h` regex character class. However that character class is not supported in Java 7 (see: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html) so there are tons of test failures on Java 7 and Windows.

PR #6769, opened 2019-03-04
corresponding rollback tests for transaction propagation

njr-11:
Create tests for rollback corresponding to the existing tests for commit, so that we can further experiment with transaction context propagation.


njr-11:
#build

PR #6774, opened 2019-03-04
Fix custom parameter configuration for OIDC clients

ayoho:
The `<authzParameter>` and related configuration elements were not having their values refreshed if you configure a server with those elements and then reconfigure the server to remove those elements; the custom parameters will continue to be added after the configuration is updated. This change fixes that behavior.


ayoho:
#build

ayoho:
#build

ayoho:
Personal build had only two known failures in an unrelated bucket.

PR #6775, opened 2019-03-04
Add support for UCP with datasource definition

alexsm82:
Add support for using Oracle UCP with the DataSourceDefinition annotation.


alexsm82:
#build

alexsm82:
The two failures in the pb are unrelated and match to existing known defects.

PR #6784, opened 2019-03-05
Alphabetize list of public features installed/removed from servers

aguibert:
I find it easier to locate specific features when they are in a deterministic order. Also, this will make it easier to diff 2 different feature lists from two different sets of logs to quickly pick out differences.


aguibert:
#build

PR #6791, opened 2019-03-05
Verify that MP Rest Client 1.2 will propagate headers

andymc12:
This PR fixes a bug where headers were not being propagated using the default ClientHeadersFactory (via `@RegisterClientHeaders`) and verifies it with new FAT tests.


andymc12:
#build

andymc12:
1 error, and it is unrelated to these changes.

PR #6799, opened 2019-03-06
specify defaults globally via MP Config

njr-11:
Implement recent spec update to allow MP Config to be used to specify global defaults for ManagedExecutor and ThreadContext configuration attributes.


njr-11:
#build

Issue #6800, opened 2019-03-06
Allow users to configure a cert less then a year lifespan in the certificateUtility


acdemyers:
Currently the certificate utility only allows users to create with a minimal validity period of 365 days.   It has been requested to allow for certs less then a year lifespan.  The default will continue to be 365 days.  Users will be allowed to specify a low as 15 days with a command line options.

PR #6801, opened 2019-03-06
Lower validity period allowed

acdemyers:
Allow a lower validity period on certs created with the certificate utility.


acdemyers:
#build

acdemyers:
fixes #6800

PR #6802, opened 2019-03-06
Divide DataSourceDefinition maxStatements when using UCP

alexsm82:
The `DataSourceDefinition` annotation definitions its `maxStatements` property as "The total number of statements that a connection pool should keep open.".  UCP also has a `maxStatements` property, which it defines as “Sets the maximum number of statements that may be pooled or cached on a connection”.  Since UCP uses `maxStatements` per connection, we will divide the value from `DataSourceDefinition` by `maxPoolSize`, exactly as we do in Liberty when converting `maxStatements` to `maxCacheSize`.


alexsm82:
#build

alexsm82:
#build

PR #6806, opened 2019-03-06
Handle JDK12 when invoking default interface methods.

andymc12:
This removes the check that the Throwable must be an IllegalAccesssException. This is because the IBM JDK 12 removes the accessMode field that we used to in JDK 8 - that ends up throwing a NoSuchFieldException.

The mpRestClient-1.2 feature has not yet shipped (not has OL with JDK12 support), so this PR is not a release bug.


andymc12:
#build

andymc12:
Rebased - new #build

aguibert:
Thanks for looking into this one @andymc12, just FYI there is no such thing as IBM JDK 12. This issue occurs because of a change in OpenJDK source, and will occur on Hotspot JDK 12 or OpenJ9 JDK 12 (once available).

PR #6809, opened 2019-03-06
Block unsafe operations on UCP

alexsm82:
As with other JDBC interfaces, block the potentially unsafe methods on the UCP PoolDataSource interface.  Note on the property setters, we generally block all setter methods on the datasource but these were not covered by the general rules since they are two arg, so we special case them here.  The connection properties can be set using the connectionProperties config and the connection factory properties can be set using the connectionFactoryProperties config.


alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

PR #6814, opened 2019-03-07
spring boot FAT failure fix

sebratton:
fix an intermittent failure in FAT issue due incorrect asserts


sebratton:
#build

sebratton:
@tjwatson I hope this change will at least get past the first failure when neither "web application available" or "web application removed" are seen. If the channel is really messed up, the test may still fail later when the test app starts the second time.

PR #6824, opened 2019-03-07
Avoid test ordering problems in config fat

brenthdaniel:
The child alias tests can fail if run in a different order.


brenthdaniel:
#build

brenthdaniel:
publishing failure...
#build

PR #6828, opened 2019-03-07
Look for property that unittests are disabled before trying to unzip JUnit zip

wraschke:
This is right!


wraschke:
Depends on https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=261355, which is delivered now.

PR #6830, opened 2019-03-08
Increase arquillian undeploy timeout

arturdzm:
MP Open API TCK fails while undeploying application with the timeout
Default timeout is 20 seconds.

Looking at last failed builds, server stopped tck application in 26 seconds


arturdzm:
#build

PR #6832, opened 2019-03-08
FT 2.0: Handle exceptions when applying thread context

### LINKED ISSUE #6776: FT 2.0: Async CompletionStage doesn't return a result if thread context can't be applied
If the following happens:
* java2 security is enabled
* the user obtains a `CompletionStage` by using `CompletableFuture.supplyAsync` or one of the `async` methods on `CompletionStage`
* they then pass that CompletionStage back as the result from a Fault Tolerance `@Asynchronous`

Then:
* The completion of the execution can happen on a thread from the common fork join pool, rather than on a liberty worker thread
* A security exception is thrown when the async executor attempts to apply the correct thread context before completing the result CompletionStage
* An FFDC is raised
* The result CompletionStage is never completed (!)

There are two problems which need fixed:
1) An unexpected exception should always result in the CompletionStage being completed with that exception
2) We should cope gracefully with the case where we end up running on a fork join pool thread (probably best to asynchronously schedule the task to run on the liberty thread pool)

Azquelt:
Fixes #6776 

* Catch and handle exceptions which occur when applying thread context before completing the result of an asynchronous method which returns a completion stage.
* Add a workaround for the fairly common (but not permitted) case where a completion stage is completed on a fork join thread and thread context cannot be applied if a security manager is present.

### DIFF
--- a/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncCompletionStageExecutor.java
+++ b/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncCompletionStageExecutor.java
@@ -15,6 +15,8 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ScheduledExecutorService;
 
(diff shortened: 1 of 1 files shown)


Azquelt:
#build

Azquelt:
@hutchig yep, `internal.error.CWMFT4998E` is a pre-existing message for reporting unexpected internal exceptions.

Azquelt:

PR #6838, opened 2019-03-08
Issue 6822: Fail Automatic EJB Timers if no tables exist

### LINKED ISSUE #6822: Automatic EJB Timer creation skipped if database tables do not exist
When using the ejbPersistentTimer-2.1 feature and `@Schdule` to define automatic persistent EJB timers, if the database tables for EJB Timer persistence do not exist, then the application starts successfully without creating the persistent timers.  Attempts to create persistent timers programmatically do fail.

Expectation is that the application should fail to start if the feature is enabled and the application contains automatic timers.  Instead, just the following message is logged:

CNTR0219I: The server created 0 persistent automatic timer or timers and 0 non-persistent automatic timer or timers for the test.war module.
CNTR4001I: The test.war EJB module in the test application has started successfully.

Enabling trace does show that the following error occurs:

Caused by: org.h2.jdbc.JdbcSQLException: Table "WLP_PROP" not found; SQL statement:
INSERT INTO WLP_PROP (ID, VAL) VALUES (?, ?) [42102-197]

However, the error is ignored, and the application is allowed to start without creating the automatic persistent EJB timers.

Note: normally the database tables would be created automatically, but this does not occur with the following configuration:

  <databaseStore createTables="false" dataSourceRef="h2" id="timerstore" tablePrefix="WLP_"/>
 
Specifically, when createTables is set to "false".

tkburroughs:
Fixed the createProperty method on the DatabaseTaskStore to properly
return false only when the row exists in the database.

EJB Container will:
- true : know row was inserted, so create automatic timers
- false : know row already exists, so do nothing
- exception : fail application start since required timers cannot be created

Also, added a new test that verifies the exception case when
no tables exist in the database.

fixes #6822

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.timer.persistent_fat/fat/src/com/ibm/ws/ejbcontainer/timer/persistent/fat/tests/NoTablesTest.java
+++ b/dev/com.ibm.ws.ejbcontainer.timer.persistent_fat/fat/src/com/ibm/ws/ejbcontainer/timer/persistent/fat/tests/NoTablesTest.java
@@ -1,0 +1,91 @@
+/*******************************************************************************
+ * Copyright (c) 2019 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package com.ibm.ws.ejbcontainer.timer.persistent.fat.tests;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.EnterpriseArchive;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import com.ibm.websphere.simplicity.ShrinkHelper;
+
+import componenttest.annotation.ExpectedFFDC;
+import componenttest.annotation.Server;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+
+/**
+ * Test ejbPersistentTimer feature when a datasource tables have not been created.
+ */
+@RunWith(FATRunner.class)
+public class NoTablesTest extends FATServletClient {
+
+    private static final String CNTR0218E = "CNTR0218E.* " + "NoDBPersistAutoTimerEJB.jar";
+    private static final String CNTR4002E = "CNTR4002E.* " + "NoDBPersistAutoTimerEJB.jar.* " + "NoDBPersistAutoTimerApp";
+    private static final String CWWKZ0106E = "CWWKZ0106E.* " + "NoDBPersistAutoTimerApp";
+    private static final String CWWKZ0002E = "CWWKZ0002E.* " + "NoDBPersistAutoTimerApp";
+
+    @Server("com.ibm.ws.ejbcontainer.timer.persistent.fat.NoTablesServer")
+    public static LibertyServer server;
+
+    @After
+    public void cleanUp() throws Exception {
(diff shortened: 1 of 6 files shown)


tkburroughs:
#build

tkburroughs:
FAT failure appears unrelated to concurrency change; also seems to match a failure seen in another build.

dev/com.ibm.ws.ejbcontainer.timer.persistent_fat/fat/src/com/ibm/ws/ejbcontainer/timer/persistent/fat/tests/NoTablesTest.java
```diff
@@ -0,0 +1,91 @@
...
+        server.setMarkToEndOfLog();
+
+        // Use ShrinkHelper to build the ear and export to server dropins
+        JavaArchive NoDBPersistAutoTimerEJB = ShrinkHelper.buildJavaArchive("NoDBPersistAutoTimerEJB.jar", "com.ibm.ws.ejbcontainer.timer.nodb.pauto.ejb.");
+        EnterpriseArchive NoDBPersistAutoTimerApp = ShrinkWrap.create(EnterpriseArchive.class, "NoDBPersistAutoTimerApp.ear");
+        NoDBPersistAutoTimerApp.addAsModule(NoDBPersistAutoTimerEJB);
+        ShrinkHelper.exportToServer(server, "dropins", NoDBPersistAutoTimerApp);
+
+        // Verify the application failed to start with correct messages
+        assertNotNull(CNTR0218E, server.waitForStringInLogUsingMark(CNTR0218E)); // persistent timers not supported
```
olendvcook:
```
* Wait for the specified regexp in the default logs from the last mark.
* The offset is also incremented every time this method is called.
```

All the exceptions are probably printed out in this order but if they ever print out in a different order the test will fail, I'd consider using the not mark version

dev/com.ibm.ws.ejbcontainer.timer.persistent_fat/fat/src/com/ibm/ws/ejbcontainer/timer/persistent/fat/tests/NoTablesTest.java
```diff
@@ -0,0 +1,91 @@
...
+        assertNotNull(CNTR4002E, server.waitForStringInLogUsingMark(CNTR4002E)); // EJB module failed to start
+        assertNotNull(CWWKZ0106E, server.waitForStringInLogUsingMark(CWWKZ0106E)); // App failed to start
+        assertNotNull(CWWKZ0002E, server.waitForStringInLogUsingMark(CWWKZ0002E)); // Exception occurred starting app
+
+        // Verify an info message did NOT occur that persistent timers not configured
+        assertNull("CNTR4021I", server.verifyStringNotInLogUsingMark("CNTR4021I", 0));
```
olendvcook: there's no verifyStringNotInLog (not using mark) so i'd call `resetLogMarks()` before this

dev/com.ibm.ws.ejbcontainer.timer.persistent_fat/fat/src/com/ibm/ws/ejbcontainer/timer/persistent/fat/tests/NoTablesTest.java
tkburroughs: Actually, the 'mark' version is the version to use if the order may change.  waitForStringInLogUsingLastOffset will increment the offset with each call and force the order. waitForStringInLogUsingMark updates the last offset, but NOT the 'mark'.... so will search from last mark, which will be beginning of log for this test.  

For this particular scenario, the messages should always be in this order; and while switching to the Offset version of method would enforce the order, that approach seems to be deprecated in preference of using the Mark methods.

dev/com.ibm.ws.ejbcontainer.timer.persistent_fat/fat/src/com/ibm/ws/ejbcontainer/timer/persistent/fat/tests/NoTablesTest.java
tkburroughs: The test calls 'setMarkToEndOfLog' after starting server, and before installing the app.  The test is intending to verify the message is not in the logs from the action of starting the apps, so 'verifyStringNoInLogUsingMark' is correct, as it will search the logs from the 'mark' that was set just prior to installing the apps.  Calling the 'mark' version of the methods does not update the mark, just the offset.

olendvcook:
@tkburroughs merged

PR #6841, opened 2019-03-08
remove tests for config annos and NamedInstance

njr-11:
Given that ManagedExecutorConfig, ThreadContextConfig, and NamedInstance annotations are being removed from the MP Concurrency specification, we should remove our tests for these, so that our implementation can later be removed as well.


njr-11:
#build

PR #6842, opened 2019-03-08
Pull in changes for CXF-7992 to resolve memory leak in MP RC 1.2

andymc12:
Resolve memory leak when using Async methods in MP Rest Client 1.2 - this has not been released yet, so not a release bug.


andymc12:
#build

andymc12:
There were some WIM failures that appear to be due to a downed LDAP server - these failures are unrelated to these changes.  I confirmed that all MP Rest Client 1.2 FATs continue to pass - as well as other MP technologies' FATs that depend on MP RC (like OpenTracing, etc.).

PR #6844, opened 2019-03-09
remove support for config annos and NamedInstance because they were removed from MP Concurrency

njr-11:
The following are being removed from MP Concurrency, and so we should remove our implementation/support for them,
ManagedExecutorConfig
ThreadContextConfig
NamedInstance


njr-11:
#build

njr-11:
ID message review is not needed because the message files are being removed, not updated or added to.

PR #6853, opened 2019-03-11
add tests for serial usage of local transaction across multiple threads

njr-11:
Add test cases that attempt to use a connection performing a local transaction (scenarios for both shared and unshared) across multiple async stages, committing or rolling back in the final stage.


njr-11:
#build

Issue #6761, opened 2019-03-04
Custom JAX-RS ParamConverter doesn't work for collection and array types


sithmein:
I was trying to create a custom `javax.ws.rs.ext.ParamConverter<List<String>>`. The idea is to convert query parameters such as "ids=i1,i2,i3" into a list of strings i1, i2, i3 in the resource class. The convert is instantiated correctly and used but the call actually fails with

> java.lang.ClassCastException: Cannot cast java.util.Arrays$ArrayList to java.lang.String

I did a bit of debugging and it seems there is some special treatment for collections and arrays in `org.apache.cxf.jaxrs.utils.InjectionUtils` which partially ignores the custom `ParamConverter`. It's called but the returned value as assume to be the type inside the collection.
Attached is a small example project to reproduce the issue:
* `mvn package`
* `java -jar com.knime.enterprise.cloud.service.repository.app/target/cloud.service.repository.app-0.0.2-SNAPSHOT.jar`
* `curl http://localhost:9080/repository/foo?ids=1,2,3`

The relevant code is in _com.knime.enterprise.cloud.service.repository.impl/src/main/java/com/knime/enterprise/cloud/service/repository/impl/v4_
This occurs in Open Liberty 18.0.0.3.

[jar-zip.zip](https://github.com/OpenLiberty/open-liberty/files/2926729/jar-zip.zip)

bpaskin-ibm:
This has nothing to do with the findings of the IBM team, but I believe you need a `?` instead of a `/` after `foo` to read in the query parameters, which `1,2,3` will be read in as a `String` first.

WhiteCat22:
Looks like some of the logic we are doing in InjectionUtils.injectIntoCollectionOrArray() to get the realType could use some work.

paramType=interface java.util.List and genericType=java.util.List<java.lang.String> until that point.

WhiteCat22:
I tested with CXF 3.2.8 and the issue does not occur, so this must be the result of some change we've made to CXF.

Uploading the war file (with source) I used to reproduce with vanilla CXF (change .zip to .war)
[CXFParamConverter.zip](https://github.com/OpenLiberty/open-liberty/files/2963410/CXFParamConverter.zip)

andymc12:
This issue should be resolved in the next Liberty release.  Thanks.

PR #6855, opened 2019-03-11
6761 - fix javax.ws.rs.ext.ParamConverter not working for collection …

WhiteCat22:
…and array types

#6761

It seems that InjectionUtils.injectIntoCollectionOrArray() tries to build a list, but when a param converter is specified it blows up because it tries to cast the list returned from the param converter to the generic type of the list.

To fix this, I check for a param converter right away and skip all of the extra work if we find one. This passes all of our param tests locally and I wrote some new tests to test param converter functionality.

### DIFF
--- a/dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/jaxrs/utils/InjectionUtils.java
+++ b/dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/jaxrs/utils/InjectionUtils.java
@@ -971,6 +971,7 @@ public final class InjectionUtils {
     }
 
     //CHECKSTYLE:OFF
+    @FFDCIgnore(value = { IndexOutOfBoundsException.class }) // Liberty Change
     private static Object injectIntoCollectionOrArray(Class<?> rawType,
                                                       Type genericType,
                                                       Annotation[] paramAnns,
@@ -978,6 +979,27 @@ public final class InjectionUtils {
                                                       boolean isbean, boolean decoded,
                                                       ParameterType pathParam, Message message) {
         //CHECKSTYLE:ON
+        // Liberty change start
+        ParamConverter<?> pm = null;
+        if (message != null) {
+            ServerProviderFactory pf = ServerProviderFactory.getInstance(message);
+            pm = pf.createParameterHandler(rawType, genericType, paramAnns, message);
+            if (pm != null) {
+                if (tc.isDebugEnabled() && values.size() > 1) {
+                    Tr.debug(tc, "injectIntoCollectionOrArray unexpected: size of values > 1, values.size()=" + values.size());
+                }
+                for (Map.Entry<String, List<String>> entry : values.entrySet()) {
+                    // always only ever 1?
+                    // if a user specifies the wrong id we should return null instead of blowing up
+                    try {
+                        return pm.fromString(entry.getValue().get(0));
+                    } catch (IndexOutOfBoundsException e) {
+                        return null;
+                    }
+                }
+            }
+        }
+        // Liberty change end
         Class<?> type = getCollectionType(rawType);
 
         Class<?> realType = null;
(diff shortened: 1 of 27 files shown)


WhiteCat22:
#build

dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/jaxrs/utils/InjectionUtils.java
```diff
@@ -971,13 +971,32 @@ private static Object mergeCollectionsOrArrays(Object first, Object second, Type
...
                                                       ParameterType pathParam, Message message) {
         //CHECKSTYLE:ON
+        // Liberty change start
+        ParamConverter<?> pm = null;
+        if (message != null) {
+            ServerProviderFactory pf = ServerProviderFactory.getInstance(message);
+            pm = pf.createParameterHandler(rawType, genericType, paramAnns, message);
+            if (pm != null) {
+                for (Map.Entry<String, List<String>> entry : values.entrySet()) {
+                    // always only ever 1?
```
jim-krueger: I would suggest checking the size of "values" and putting a debug trace in here indicating that there are more entries than we expect and listing them

dev/com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2/src/org/apache/cxf/jaxrs/utils/InjectionUtils.java
```diff
@@ -1001,13 +1001,32 @@ private static Object mergeCollectionsOrArrays(Object first, Object second, Type
...
                                                       ParameterType pathParam, Message message) {
         //CHECKSTYLE:ON
+        // Liberty change start
+        ParamConverter<?> pm = null;
+        if (message != null) {
+            ServerProviderFactory pf = ServerProviderFactory.getInstance(message);
+            pm = pf.createParameterHandler(rawType, genericType, paramAnns, message);
+            if (pm != null) {
+                for (Map.Entry<String, List<String>> entry : values.entrySet()) {
+                    // always only ever 1?
```
jim-krueger: same comment as above

dev/com.ibm.ws.jaxrs.2.0_fat/test-applications/paramconverter/src/com/ibm/ws/jaxrs/fat/paramconverter/objects/NoPublicConstructorListObject.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.ArrayList;
+
+import javax.json.bind.annotation.JsonbCreator;
+
+public class NoPublicConstructorListObject extends ArrayList<String> {
+
+    private static final long serialVersionUID = 646543532165146L;
+
+    @JsonbCreator
```
andymc12: Does this work?  When we chatted, I thought you said that `@JsonbCreator` only worked on other (non-no-arg) constructors.  If so, we should remove it here - or change the static getter to be a public constructor with an argument.

dev/com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2/src/org/apache/cxf/jaxrs/utils/InjectionUtils.java
```diff
@@ -1001,13 +1001,35 @@ private static Object mergeCollectionsOrArrays(Object first, Object second, Type
...
+                    if (tc.isDebugEnabled() && values.size() > 1) {
+                        Tr.debug(tc, "injectIntoCollectionOrArray unexpected: size of values > 1, values.size()=" + values.size());
```
andymc12: Same as above - this should be moved outside of the loop.

dev/com.ibm.ws.jaxrs.2.0_fat/test-applications/paramconverter/src/com/ibm/ws/jaxrs/fat/paramconverter/objects/NoPublicConstructorListObject.java
WhiteCat22: It didn't work and I missed taking this out. I adjusted the test to return a string instead and it works.

WhiteCat22:
#build

PR #6858, opened 2019-03-12
tests for unchanged transaction context

njr-11:
Add tests for unchanged transaction context, which allows us to run the scenario where unshared connection is used serially across threads.
Also, update an existing test to ensure that a new transaction is attempted after having run within the same transaction inline on the same thread.


njr-11:
#build

PR #6867, opened 2019-03-12
Synchronize all variations of JDBC 4.2 Statement wrappers

aguibert:
Fixes an issue raised on StackOverflow here: https://stackoverflow.com/questions/55126974/liberty-java-lang-unsupportedoperationexception-getlargeupdatecount-not-impleme


aguibert:
#build

PR #6876, opened 2019-03-12
Increase timeout for shutdown

andymc12:
On slow test machines, the arquillian shutdown timeout is firing before the server can shutdown, resulting in errors.  This fix increases the shutdown timeout to account for these slower test machines.


andymc12:
#build

andymc12:
Previous build failed due to an infrastructure issue - new #build

andymc12:
Same infrastructure build, but I've heard that the problem is resolved, so new #build

PR #6877, opened 2019-03-12
restore of already-completed transaction should be a no-op

njr-11:
Update our experimental transaction propagation code path such that an attempt to restore an already-completed transaction onto a thread (during the context restore phase) is a no-op rather than failing with InvalidTransactionException.  This happens when the completion stage commits or rolls back the transaction during its execution.  In this case, the proper transaction context to restore is empty because the transaction that had previously been active has ended.


njr-11:
#build

njr-11:
#build

njr-11:
Build has 2 test failures.  One is a preexisting issue in an unrelated bucket.  The other is a preexisting test infrastructure issue that is erroneously trying to run a bucket named "all".  This impacts other personal builds as well and so isn't from my changes.

PR #6878, opened 2019-03-12
Ensure CompletionStage is handled correctly in JsonBProvider

andymc12:
We use JSON-B to deserialize JSON into object, and use the generic type in order to ensure that collections (i.e. List<MyObject>) are properly deserialized. CompletionStages should be treated differently since they are used for the asynchronous operations, not part of the data model.


andymc12:
#build

andymc12:
Previous build failed due to an infrastructure issue - new #build

andymc12:
More infrastructure issues... new #build

PR #6880, opened 2019-03-13
Don't register file monitors while shutting down

brenthdaniel:
If a server is stopped immediately after starting it, it's possible to get an IllegalStateException in the FFDC logs.


brenthdaniel:
#build

brenthdaniel:
Build infrastructure issues, trying again
#build

PR #6885, opened 2019-03-13
Don't search for Driver with UCP

alexsm82:
Similar to what we are doing for UCP with `ConnectionPoolDataSource` we should throw an exception if someone tries to configure UCP with `type=java.sql.Driver` and the `properties.oracle.ucp` element.  Because UCP has no driver interface this will result in them using the Driver interface from the Oracle driver, when they are attempting to use UCP.


alexsm82:
#build

alexsm82:
The pr failed with an unrelated existing defect.

PR #6893, opened 2019-03-13
tests for transaction propagation with one-phase resource used serially with multiple handles involved

njr-11:
Add tests case where multiple sharable and unsharable (not expected to work due to enlisting multiple one-phase resources) connection handles are used serially to perform transactional work within a transaction that is propagated across threads.


njr-11:
#build

njr-11:
#build

PR #6895, opened 2019-03-13
Pass initialPoolSize to UCP

alexsm82:
Ensure that we don't remove initialPoolSize from the vendor properties when using UCP so that it is applied to the UCP driver.


alexsm82:
#build

PR #6902, opened 2019-03-14
Pull in change for CXF-7980 - avoid static collection of clients

andymc12:
This fix pulls in the change from the Apache CXF community that avoids leaking Rest Clients when apps stop. 

More details can be found in the JIRA [CXF-7980](https://issues.apache.org/jira/browse/CXF-7980).


andymc12:
#build

PR #6903, opened 2019-03-14
tests for transaction propagation with one-phase commit unsharable cached handle

njr-11:
Add some test cases for caching an unsharable handle across a propagated transaction, for use as a one-phase only resource.


njr-11:
#build

PR #6913, opened 2019-03-14
Fix config update path where only the properties element is modified …

alexsm82:
…switching to or from UCP.  Previously in this path the already initialized connection manager service would be used, with could incorrectly enable/disable Liberty's connection pooling.


alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

PR #6932, opened 2019-03-15
test cases with two-phase commit where resources used on different threads

njr-11:
Add test cases where two-phase capable resources are used on different threads within the same transaction, with transaction context being propagated by MP Concurrency.  Include scenarios for commit and rollback.


njr-11:
#build

PR #6935, opened 2019-03-15
Set connectionFactoryClassName for UCP when not specified

alexsm82:
When using UCP, if a connectionFactoryClassName is not provided, set it based on the datasource type being used.


alexsm82:
#build

PR #6949, opened 2019-03-18
write test for last participant support in combination with transaction being used across multiple threads

njr-11:
Write tests that include enlisting a single one-phase resource (last participant support) alongside a two phase resource in a transaction that is propagated across threads.  Test both commit and rollback for such a transaction.


njr-11:
#build

PR #6960, opened 2019-03-19
Add stall warning to the zip reaper

tbitonti:


tbitonti:
#build

tbitonti:
Put in a warning that lets us know when the zip reaper thread isn't allowed to run in a timely fashion.  This is one possible cause of the windows file locking problems that we've been seeing.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/cache/internal/ZipFileReaper.java
```diff
@@ -200,15 +212,33 @@ public void run() {
...
+                    reapAt = SystemUtils.getNanoTime();
+                    if ( doDebug ) {
+                        Tr.debug(tc, methodName + " Reap [ " + toRelSec(startAt, reapAt) + " (s) ]");
+                    }
+
+                    if ( reapDelay > 0L ) {
+                        long actualDelay = reapAt - lastReapAt;
+                        if ( actualDelay > reapDelay ) {
+                            long overage = actualDelay - reapDelay;
+                            if ( overage > STALL_LIMIT ) {
```
tkburroughs: Are we not concerned about the first time though, where reapDelay == -1?  This is an indefinite wait.... what if the first time through is taking much longer than it needs to be?  What exactly breaks the wait out of the wait forever?

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/cache/internal/ZipFileReaper.java
```diff
@@ -200,15 +212,33 @@ public void run() {
...
+                                Tr.warning(tc, methodName +
```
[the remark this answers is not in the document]
tbitonti: Hi; yes!  I'll get this in as a translated message.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/cache/internal/ZipFileReaper.java
tbitonti: An indefinite wait occurs when there are no pending closes.  The wait is broken when a new pending close appears.  In this case, the reaper thread should awaken immediately, as the sleep was broken by a call to notify.  For other cases, the sleep is broken after the requested time has expired.

Most often, the initial wait will be short, but it could grow to be quite long, for example, if the server quiesces and sees no activity for an expended period of time: The server should eventually process all pending closes, and the reaper thread will go back to an indefinite wait.

I could have the code set a time when the notification for a first pending close is made, but that would introduce new state, and would be more complex than handling the definite wait case.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/cache/internal/ZipFileReaper.java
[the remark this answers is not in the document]
tbitonti: Updates for warnings are now posted.

tbitonti:
#build

dev/com.ibm.ws.artifact.zip/resources/com/ibm/ws/artifact/zip/internal/resources/ZipArtifactMessages.nlsprops
```diff
@@ -1,3 +1,10 @@
+[ant:validate]   (error) reaper.unregistered.path is duplicate key
```
tkburroughs: This heading seems problematic.... doesn't seem these errors should be here, and could cause problems processing the file.

tbitonti: Agh, a stray paste while adding the new warnings.  That header is not supposed to be there.

tbitonti:
#build

tbitonti:
#build

PR #6963, opened 2019-03-19
Don't send auth alias message when using UCP

alexsm82:
When using Oracle UCP don't output the message telling users to use an auth alias when the user and password are specified as datasource properties.  This is because certain versions of Oracle UCP require that a user and password are set as datasource properties.


alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
Two pbs were run, each had a different unrelated failure, the first due to a db2 machine issue, and the second due to an intermittent issue in the mp config unit tests.

PR #6965, opened 2019-03-19
temporary approximate enforcement of limitation against resuming transaction onto multiple threads at once

njr-11:
The transaction manager doesn't yet have the ability to stop a transaction from being resumed onto multiple threads at once, so I'm adding some temporary code that aims to at least sometimes be able to enforce it on our own, which will allow us to start to get our tests updated according to this expectation.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build

PR #6979, opened 2019-03-21
Server variable substitution for validator parameters

aguibert:


aguibert:
#build

dev/com.ibm.ws.rest.handler.validator/src/com/ibm/ws/rest/handler/validator/internal/ValidatorRESTHandler.java
```diff
@@ -308,4 +313,16 @@ private JSONObject toJSONObjectForThrowable(Map<String, ?> errorInfo, Throwable
 
         return json;
     }
+
+    @Trivial
+    private String resolvePotentialVariable(String value) {
+        if (value == null || !value.startsWith("${") || !value.endsWith("}"))
```
njr-11: Rather than us owning the logic to determine what is a valid variable and what is not, it would be nice to keep that logic in the variable registry itself.  Can we just send non-null values to VariableRegistry regardless?

dev/com.ibm.ws.rest.handler.validator_fat/fat/src/com/ibm/ws/rest/handler/validator/fat/ValidateDataSourceTest.java
```diff
@@ -81,31 +81,37 @@ public void testAppAuth() throws Exception {
...
-        assertEquals(err, "DefaultDataSource", json.getString("id"));
-        assertNull(err, json.get("jndiName"));
-        assertTrue(err, json.getBoolean("successful"));
-        assertNull(err, json.get("failure"));
-        assertNotNull(err, json = json.getJsonObject("info"));
-        assertEquals(err, "Apache Derby", json.getString("databaseProductName"));
-        assertTrue(err, json.getString("databaseProductVersion").matches(VERSION_REGEX));
-        assertEquals(err, "Apache Derby Embedded JDBC Driver", json.getString("jdbcDriverName"));
-        assertTrue(err, json.getString("jdbcDriverVersion").matches(VERSION_REGEX));
+        assertSuccessResponse(json, "DefaultDataSource", "DefaultDataSource");
```
njr-11: I don't see why this test code was consolidated elsewhere.  The indirection lessens the readability/debugability of the test case.

aguibert: I disagree here, its still easy to track where the test failed with a stack trace, and having a bunch of redundant code makes the file larger and more cumbersome to navigate. 

If we have 1-3 sets of duplicated code, I usually don't refactor out into helper methods, but when we're doing the same thing 4+ times then a helper method is warranted

aguibert:
#build

dev/com.ibm.ws.rest.handler.validator/src/com/ibm/ws/rest/handler/validator/internal/ValidatorRESTHandler.java
```diff
@@ -308,4 +313,16 @@ private JSONObject toJSONObjectForThrowable(Map<String, ?> errorInfo, Throwable
 
         return json;
     }
+
+    @Trivial
+    private String resolvePotentialVariable(String value) {
+        if (value == null)
+            return value;
+
+        String resolvedVariable = variableRegistry.resolveString(value);
```
alexsm82: You need to use `variableRegistry.resolveRawString(value)` so that the value is not path normalized.

aguibert: why don't we want path normalization here?

alexsm82: Well for one, it's different behavior than we have for server config and datasource definition, so it may result in a different behavior than they'd actually get in their app. It also messes up some JDBC urls (especially for Oracle) among possibly other variables.

aguibert:
#build

Issue #6954, opened 2019-03-19
Removing web-bnd stanza from server.xml doesn't remove binding


frowe:
On the 2019002 beta, I noticed something unexpected with server specified bindings. I define a resource reference binding in server.xml like this:
```
   <webApplication id="RestfulConnectionTest"
       location="RestfulConnectionTest.war"
       name="RestfulConnectionTest">
       <web-bnd>
           <resource-ref binding-name="jdbc/myDB" name="myDBRef"></resource-ref>
       </web-bnd>
   </webApplication>

```
The resource is injected as expected and everything works.  If I modify the binding name or the name and save server.xml, the resource injection fails as expected.   But, if I completely remove the web-bnd stanza and save server.xml, resource injection still works as if the binding still existed.  In fact, even after restarting the server, the binding still exists as resource injection continues to work.  It's not until I clean the server and restart, does the binding go away and the resource injection starts to fail.  I'm assuming that not expected behavior?

brenthdaniel:
Thanks. It looks like there's an issue with correctly processing deletes of child-first configuration elements that are nested under elements that extend another element. In this case, the child first element "web-bnd" is nested under "webApplication" which extends "application".  The bindings and extensions elements may be the only ones that have this particular issue. 

As a workaround, you could use "application" instead of "webApplication".

PR #6996, opened 2019-03-22
Look up metatype entries for deleted elements correctly

brenthdaniel:
PR for #6954 

Tests for this function still need to be moved to open liberty, so I'm only making the runtime change here.


brenthdaniel:
#build

PR #6998, opened 2019-03-22
Fixing missing PS Exception Filter

cbridgha:
My initial Docker test was invalid, so an additional fix was required.  (Filter for the proper IOException was invalid)

This fix is now validated on the minimal Docker image.


cbridgha:
#build

PR #7008, opened 2019-03-23
update to renamed spec packages and classes

njr-11:
When the MicroProfile Concurrency spec was renamed to MicroProfile Context Propagation, the spec packages and a couple of class names were updated, as well as the packages for the TCK tests.  Pull in an updated MP Context Propagation jar and update OpenLiberty code accordingly.


njr-11:
#build

njr-11:
#build

tevans78:
@njr-11 are there any plans to rename the implementation bundles as well? Something like `com.ibm.ws.microprofile.context.*` ?

njr-11:
> @njr-11 are there any plans to rename the implementation bundles as well? Something like `com.ibm.ws.microprofile.context.*` ?

We haven't decided regarding the internal package names yet, however there are certainly more renames coming.  For example, the mpConcurrency-1.0 beta feature will need to be renamed to mpContextPropagation-1.0.  I'm not aiming to cover all aspects under this pull. This pull is rather the first step of switching to the new spec interfaces and demonstrating that all of our tests will work.

njr-11:
The two failures in the build indicate that allowedOSPackages must be updated to list the new spec package names.  Will need to wait for that to go in (separate build) before we can attempt to submit another build for this.

njr-11:
#build

njr-11:
@tevans78 I'm reversing some of my previous reply to you given that it's taking quite a long time for the prerequisite changes to make it into master.  Given the delay, this pull is including a number of additional corresponding renames, including the mpContextPropagation-1.0 feature name.  Unfortunately the old feature name must be be left in place for additional cycles through commits to master, when this pull will eventually need to commit to OpenLiberty master, then I can do an update to switch the feature name in the beta feature list, and then wait for that to go into master, and then we can finally remove the mpConcurrency-1.0 feature name from OpenLiberty.  I expect it will be quite a while to make it through all of these cycles.

Issue #6982, opened 2019-03-21
JAX-RS 2.1 Performance


jdmcclur:
There is a small JAX-RS 2.1 server regression in the last few weeks with the primitives that I run. So far analysis has not been conclusive. Continuing to investigate.

jdmcclur:
This looks to be caused by https://github.com/OpenLiberty/open-liberty/pull/6485.

Profile shows time added in two places.

1.
```
  Parent      0   0.01   0.01           1           1    J:org/apache/cxf/jaxrs/utils/JAXRSUtils.runContainerRequestFilters(Lorg/apache/cxf/jaxrs/provider/ServerProviderFactory;Lorg/apache/cxf/message/Message;ZLjava/util/Set;)Z
  Parent      0   0.04   0.04           6           6    J:org/apache/cxf/jaxrs/JAXRSInvoker.invoke(Lorg/apache/cxf/message/Exchange;Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
  Parent      0   0.08   1.25          11         173    J:org/apache/cxf/jaxrs/provider/ServerProviderFactory.injectContextValues(Lorg/apache/cxf/jaxrs/model/ProviderInfo;Lorg/apache/cxf/message/Message;)V
 
    Self      0   0.13   1.30          18         180    J:org/apache/cxf/jaxrs/utils/InjectionUtils.injectContexts(Ljava/lang/Object;Lorg/apache/cxf/jaxrs/model/AbstractResourceInfo;Lorg/apache/cxf/message/Message;)V
 
   Child      0   0.19   1.00          26         138    J:org/apache/cxf/jaxrs/utils/InjectionUtils.injectContextFields(Ljava/lang/Object;Lorg/apache/cxf/jaxrs/model/AbstractResourceInfo;Lorg/apache/cxf/message/Message;)V
   Child      0   0.09   0.09          12          12    J:org/apache/cxf/jaxrs/utils/InjectionUtils.injectContextMethods(Ljava/lang/Object;Lorg/apache/cxf/jaxrs/model/AbstractResourceInfo;Lorg/apache/cxf/message/Message;)V
[...]

PR #7025, opened 2019-03-25
Improve performance of JsonBProvider

andymc12:
This change avoids Context injection of Providers - instead it captures the ContextResolvers at construction time iterating through to determine if the customer has provided their own Jsonb implementation.

This resolves issue #6982.

### DIFF
--- a/dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/jaxrs/provider/ProviderFactory.java
+++ b/dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/jaxrs/provider/ProviderFactory.java
@@ -189,7 +189,7 @@ public abstract class ProviderFactory {
                              //new StringProvider<Object>(), // Liberty Change for CXF
                              //new JAXBElementSubProvider(),
                              createJsonpProvider(), // Liberty Change for CXF Begin
-                             createJsonBindingProvider(),
+                             createJsonBindingProvider(null),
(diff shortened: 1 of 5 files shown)


andymc12:
#build

andymc12:
#build

dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/jaxrs/provider/ProviderFactory.java
```diff
@@ -283,7 +283,7 @@ public ClassLoader run() {
         return c;
     }
 
-    public static Object createJsonBindingProvider() {
+    public static Object createJsonBindingProvider(Iterable<ProviderInfo<ContextResolver<?>>> contextResolvers) {
```
jim-krueger: I see you add this contextResolvers parameter, but I don't see it being used in this method.

WhiteCat22: We don't need it for Jackson in Jaxrs-2.0, hence it's not referenced. But the method API is shared between Jaxrs-2.0 and Jaxrs-2.1 so it has to be here.

jim-krueger: works for me... thanks

PR #7026, opened 2019-03-25
Add test for app compiled at Java 12 bytecode level

aguibert:
The main builds still compile everything to the lowest supported Java (Java 7) by default, and can opt into Java 8 if needed. However, projects cannot opt into Java 9+ bytecode levels, as our main builds still run on JDK 8.  For this reason, to test JDK 12 bytecode apps, we need to check in a pre-compiled app.


aguibert:
#build

PR #7034, opened 2019-03-26
removed unused LdapConfigManager.getLdapRDNFilter(LdapEntity,String) method

### LINKED ISSUE #7014: Remove unused LdapConfigManager.getLdapRDNFilter(LdapEntity,String) method
I removed what I believe is the only call to this method in #5376, but didn't realize it at the time. The method should be removed from the source.

agiridharan:
We previously removed the last call to this method. Fixes #7014


agiridharan:
#build

PR #7048, opened 2019-03-27
Avoid forcing async responses on the AutomaticWorkQueue

andymc12:
Currently, async responses are handled using the `forceWQ` option which goes through a separate executor before actually executing the async JAX-RS or MP Rest Client response.  This is not necessary and might be a factor causing deadlocks in cases where the ForkJoinPool.commonPool() is filled up waiting for responses to be handled.


andymc12:
#build

andymc12:
#build

andymc12:
#build

andymc12:
Resolves issue #7083

Issue #7053, opened 2019-03-28
NullPointerException is thrown when serialFilter is enabled under JDK12


toshiyamamoto:
When serialFilter is enabled on JDK12 environment, the following exception is caught:

ava.lang.NullPointerException
 at java.base/java.io.ObjectInputStream.<init>(ObjectInputStream.java:352)
 at java.rmi/sun.rmi.server.MarshalInputStream.<init>(MarshalInputStream.java:123)
The exception is not observed in the previous JDK versions.

PR #7059, opened 2019-03-28
remove final from ValidatorFactory field.

toshiyamamoto:
final modifier is removed from serializationValidatorFactory field, and also modifiers fields won't be updated unless final is set.


toshiyamamoto:
fixes #7059 
#build

toshiyamamoto:
fixes #7053

toshiyamamoto:
#build

PR #7063, opened 2019-03-29
Deploy arquillian support feature and use it for FT 2.0 TCK

Azquelt:
This change deploys the arquillian support feature as a test feature (rather than a user feature) and uses it to enable the disabled tests in the Fault Tolerance 2.0 TCK


Azquelt:
#build

Azquelt:
Failures in the build correspond to a known defect.

PR #7066, opened 2019-03-29
Set RMI response timeout for TCK tests

Azquelt:
Follow up to #5203 to ensure that the system property set on the command line (in class `MvnUtils`) is propagated to the test process.


Azquelt:
#build

PR #6629, opened 2019-02-21
Revert PR #6611 and PR # 5241 - Issue5019-Java11DefaultToPKCS12

tevans78:
Reverts https://github.com/OpenLiberty/open-liberty/pull/6611
and https://github.com/OpenLiberty/open-liberty/pull/5241

Issue #6391, opened 2019-02-04
Building .tar.gz server package fails on Windows


gcharters:
When trying to build a .tar.gz on Windows using maven, I get the following error and it fails.  The identical build works find on Mac . Produced on 18.0.04 and 19.0.0.1

[INFO] CWWKM2001I: Invoke command is ["C:\open-cloud-native-intro\target\liberty\wlp\bin\server.bat", package, defaultServer, --archive=C:\open-cloud-native-intro\target\mpservice.tar.gz, --include=usr].
[INFO] Packaging server defaultServer.
[INFO] in Launcher, about to error.unknownException
[INFO] CWWKE0005E: The runtime environment could not be launched.
[INFO] CWWKE0018E: An exception occurred while launching the runtime environment: java.lang.RuntimeException: mode '-1' is too big ( > 2097151 ).
[INFO] java.lang.RuntimeException: mode '-1' is too big ( > 2097151 ).
[INFO]  at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.failForBigNumber(TarArchiveOutputStream.java:636)
[INFO]  at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.failForBigNumber(TarArchiveOutputStream.java:624)
[INFO]  at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.failForBigNumbers(TarArchiveOutputStream.java:616)
[INFO]  at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.putArchiveEntry(TarArchiveOutputStream.java:369)
[INFO]  at com.ibm.ws.kernel.boot.archive.internal.ZipArchive.processEntry(ZipArchive.java:124)
[INFO]  at com.ibm.ws.kernel.boot.archive.internal.ZipArchive.addDirEntry(ZipArchive.java:85)
[INFO]  at com.ibm.ws.kernel.boot.archive.DirEntryConfig.configure(DirEntryConfig.java:77)
[INFO]  at com.ibm.ws.kernel.boot.archive.internal.AbstractArchive.create(AbstractArchive.java:49)
[INFO]  at com.ibm.ws.kernel.boot.internal.commands.PackageProcessor.execute(PackageProcessor.java:228)
[INFO]  at com.ibm.ws.kernel.boot.internal.commands.PackageCommand.packageServerRuntime(PackageCommand.java:281)
[...]

PR #7075, opened 2019-03-29
disable jaxrs-2.0 test who's test method was removed from the test se…

WhiteCat22:
…rvlet resulting in a java.lang.NoSuchMethodException

caused by #6629


WhiteCat22:
#6391

WhiteCat22:
#build

PR #7077, opened 2019-03-31
Remove stray JAX-B dependency from repository bundle

aguibert:
Since the repository bundle is a rather low-level dependency, and JAX-B was removed from the JDK in Java 9+. Since JAX-B is several MB in size, we want to minimize dependencies on it where possible.
Since the repository bundle only used JAX-B for Base64 conversion, we can use the java.util.Base64 utility class instead which was added in JDK 8 when running on newer JDKs.

#build


aguibert:
#build

PR #7079, opened 2019-04-01
Correct Circuit Breaker metric reporting

### LINKED ISSUE #7052: mpFT 2.0: Circuit Breaker metrics updated incorrectly when non-failure exception thrown
While reviewing the solution to eclipse/microprofile-fault-tolerance#419, I found that our FT 2.0 implementation doesn't record the metrics correctly either.

Exceptions which are not included in the `failOn` list should increment the circuit breaker success metric but instead they increment the fail metric.

Looking at the code, we'll process changes to the circuit breaker state correctly, we just won't update the metrics correctly.

Azquelt:
Ensure that the metrics reported for passing/failing executions reflect the value of the `failOn` attribute of the circuit breaker.

Without this change, any exception is logged as a failure, even if the circuit breaker is configured to only treat certain exceptions as failures.

Fixes #7052

### DIFF
--- a/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncAttemptContextImpl.java
+++ b/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/impl/AsyncAttemptContextImpl.java
@@ -23,6 +23,7 @@ public class AsyncAttemptContextImpl<W> {
 
     private final AsyncExecutionContextImpl<W> executionContext;
     private TimeoutState timeoutState;
+    private boolean circuitBreakerPermittedExecution = false;
     private final AtomicBoolean complete = new AtomicBoolean(false);
 
     public AsyncAttemptContextImpl(AsyncExecutionContextImpl<W> executionContext) {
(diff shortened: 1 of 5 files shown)


Azquelt:
#build

Azquelt:
@hutchig Hmm, maybe "an execution must be attempted" isn't the most helpful way of putting it. The circuit breaker doesn't really care whether an execution is attempted, it just requires that a result is recorded for every time that `requestPermissionToExecute` is called.

E.g. it's fine for an execution not to be run because the bulkhead was full and for the `BulkheadException` to be reported as the result, but it's not ok to call `requestPermissionToExecute` and then not call `recordResult` at all.

Azquelt:
Under normal circumstances, an exception should not be thrown anywhere in the executors. (All execptions thrown from user code will be caught so they can be handled).

The AsyncExecutor has been coded very carefully so that if an unexpected exception is thrown (e.g. a NullPointerException caused by a programming error or an OutOfMemoryError raised by the JVM) it will log an FFDC and try to execute all of the finishing up logic for the current execution attempt, which includes recording the circuit breaker result.

(Obviously it may not be possible to do this if you've hit an OutOfMemoryError, but at that point your whole server is probably going down anyway)

The SyncExecutor's handling for unexpected exceptions is not so robust and could be improved.

PR #7092, opened 2019-04-02
Restrict HTTP methods on config and validator endpoints

aguibert:
Restrict /config endpoint to GET requests.
Restrict /validator endpoint to POST requests.

This will align better with the wider REST API principles of Liberty where POST requires an admin role, and GET is considered a "read only" operation.


aguibert:
#build

aguibert:
for issue #7452 

Not going to deliver this PR since it has merge conflicts at this point. Will hand off this delivery to @KyleAure

PR #7094, opened 2019-04-02
FT 2.0: add tests for circuitbreaker metrics

Azquelt:
Adds test for #7079 and updates other metrics tests to also run against fault tolerance 2.0


Azquelt:
#build

PR #7103, opened 2019-04-02
FT 2.0: Make circuit breaker limit executions when half-open

### LINKED ISSUE #7080: FT 2.0: Circuit breaker doesn't correctly restrict executions when in half-open state
The spec says (Chapter 9):
> In half-open state, trial executions of the service are allowed. By default one trial call
to the service is permitted.

However, our implementation permits any number of executions to start while it's in half-open state. As soon as one of them _returns_ it may switch into open or closed state, but until an execution returns, new executions are permitted to start.

This could be problematic in a situation where a service either succeeds quickly or fails slowly. Even if the first trial execution fails, a successful execution may have already gone through, putting the breaker into closed state instead of into open state as expected.

Azquelt:
When half-open, the circuit breaker should limit the number of concurrent executions to the success threshold.

Fixes #7080


tevans78:
Might need some more comments and trace. Optionally can extract out the BitSet circular buffer.

Azquelt:
squashed

Azquelt:
#build

Azquelt:
Failures in build caused by missing call to increment a metric. Now fixed.

#build

Issue #7095, opened 2019-04-02
HTTP/2 does not handle some error cases with synchronous write


loriadi:

loriadi:
Redelivering #6756 with a test case update

PR #7123, opened 2019-04-03
HTTP/2 will only use synchronous write

loriadi:
Add check for H2 in asynchronous write path.  Update one test to check for HTTP error instead of H2 reset.


loriadi:
#7095

loriadi:
#build

loriadi:
The test failures are all covered by other defects.

PR #7124, opened 2019-04-03
In jaxb-2.3 feature at IBM-API-Package, require-java:="9" is added fo…

### LINKED ISSUE #7117: jaxb-2.3 class loading problem
jaxb-2.3 javax.activation package causing class loading problems for Java versions less than 9

BerksanAtes:
…r javax.activation since before java 9, java SDKs already came with this library. And loading same library twice was causing class loading issue

fixes #7117


BerksanAtes:
#build

PR #7139, opened 2019-04-04
Add clear warning message and instructions that OpenLiberty build will soon req Java 11

aguibert:
#build 

Since OpenLiberty now requires Java 11 to build, make it clear that people need to be using Java 11 or newer, in case they try to build with anything less. Sample output:
```
$ ./gradlew :build.example_fat:compileJava

FAILURE: Build failed with an exception.

* Where:
Settings file '/Users/aguibert/dev/git/open-liberty/dev/settings.gradle' line: 237

* What went wrong:
A problem occurred evaluating settings 'dev'.
> Building this repository requires Java 11. To use Java 11 follow these steps:
    1) You can download a copy for your OS here if you don't already have it: https://adoptopenjdk.net/index.html?variant=openjdk11&jvmVariant=openj9
    2) run 'export JAVA_HOME=/path/to/your/java11' in the shell, or set in ~/.bashrc
    3) [MACOS ONLY] Set 'ulimit -Sn 1024' in your ~/.bashrc
    4) Restart your gradle daemon with './gradlew --stop'
```

### DIFF
--- a/dev/settings.gradle
+++ b/dev/settings.gradle
@@ -13,7 +13,6 @@
  *
  * Depends on bnd_* values from gradle.properties.
  */
-
 pluginManagement {
   def userProperties = {
     Properties props = new Properties()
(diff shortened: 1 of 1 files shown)


aguibert:
Since the Java 11 requirement has been lifted for now, I'm going to downgrade this to a warning message instead of a hard failure. We can leave this warning in place for a while and then swap it to an error that halts the build when we do require Java 11.

Running a new #build to make sure OpenLiberty still builds properly

PR #7143, opened 2019-04-04
Update wrong type error message

alexsm82:
From the serviceability review update to use the same message for when a datasource with UCP is configured with type=Driver or type=ConnectionPoolDataSource.


alexsm82:
#build

alexsm82:
#build

PR #7151, opened 2019-04-05
Remove DynamicSourceCleanupTest because it doesn't work

tevans78:
#build


tevans78:
#build

tevans78:
The original unit test was too complex to be a unit test and was very error prone. We tried moving the test to be a FAT test but it turns out that it just can't be made to work. We managed to get a heap dump out at the point when the test was failing and it showed that the logging framework was holding on to references to the key objects. We can't clear those references and while they are being held, the objects will not be garbage collected.

The good news is that the heap dump also showed that there were no other references to the key objects. If it hadn't been for the logging, the objects would have been garbage collected. Therefore we are confident that there is no memory leak ... which is the reason that the test was created in the first place.

Therefore we have decided to just remove the test.

tevans78:
#build

dev/com.ibm.ws.microprofile.config.1.1_fat/test-applications/dynamicSources.war/src/com/ibm/ws/microprofile/appConfig/dynamicSources/test/DynamicSourcesTestServlet.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 IBM Corporation and others.
+ * Copyright (c) 2016, 2019 IBM Corporation and others.
```
Azquelt: Could remove the changes in this file

Azquelt:
Worth adding that the references from the logging framework are soft references, which explains why (a) they don't constitute a memory leak and (b) why they weren't cleaned up when the test requests a GC.

dev/com.ibm.ws.microprofile.config.1.1_fat/test-applications/dynamicSources.war/src/com/ibm/ws/microprofile/appConfig/dynamicSources/test/DynamicSourcesTestServlet.java
```diff
@@ -218,5 +218,4 @@ public void testClosedConfigStopsRefreshing() throws Exception {
             // Expected
         }
     }
-
 }
```
Azquelt: Could remove all the changes from this file...

PR #7164, opened 2019-04-08
 add JDBC connection support to mock adapter for testing JCA validator

njr-11:
Update the mock resource adapter that is used for testing the JCA validator
so that the adapter also provides a JDBC managed connection factory.  This makes possible a different set of validation info (from DatabaseMetaData) and the connection.validate(timeout) mechanism for validating a connection.


njr-11:
#build

PR #7165, opened 2019-04-08
Skip copying server dump temp files during dump processing.

andymc12:
Test infrastructure fix - this resolves a problem where a server dump is occurring when the LibertyServer test infrastructure is attempting to copy the server directory.  It results in an exception which causes a test failure.  The exception looks like:
```
java.lang.Exception: Cannot copy a file or directory that does not exist: /home/jazz_build/_U_kFAFSgEem8NdNdaSGUMw-EBC.PROD.WASRTC-6CX-000-00-00/jbe/build/dev/image/output/wlp/usr/servers/mpRestClient11.async/dump_19.04.01_21.07.41/introspections/RegionIntrospection.txt: localhost
at com.ibm.websphere.simplicity.RemoteFile.copy(RemoteFile.java:841)
at com.ibm.websphere.simplicity.RemoteFile.copyFromSource(RemoteFile.java:513)
at componenttest.topology.impl.LibertyServer.recursivelyCopyDirectory(LibertyServer.java:2722)
at componenttest.topology.impl.LibertyServer.recursivelyCopyDirectory(LibertyServer.java:2685)
at componenttest.topology.impl.LibertyServer.recursivelyCopyDirectory(LibertyServer.java:2685)
at componenttest.topology.impl.LibertyServer.postStopServerArchive(LibertyServer.java:2612)
at componenttest.topology.impl.LibertyServer.stopServer(LibertyServer.java:2422)
at componenttest.topology.impl.LibertyServer.stopServer(LibertyServer.java:2214)
at componenttest.topology.impl.LibertyServer.stopServer(LibertyServer.java:2189)
at com.ibm.ws.microprofile.rest.client.fat.AsyncMethodTest.afterClass(AsyncMethodTest.java:75)
at componenttest.rules.repeater.RepeatTests$CompositeRepeatTestActionStatement.evaluate(RepeatTests.java:115)
at componenttest.custom.junit.runner.FATRunner$2.evaluate(FATRunner.java:314)
at componenttest.custom.junit.runner.FATRunner.run(FATRunner.java:167)
```

The fix is to avoid copying the `dump_*` directory during the post stop server archive step.  This is test infrastructure code, and therefore not a release bug.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -2616,7 +2616,7 @@ public class LibertyServer implements LogMonitorClient {
         runJextract(serverFolder);
 
         // Copy the log files: try to move them instead if we can
-        recursivelyCopyDirectory(serverFolder, logFolder, false, true, true);
+        recursivelyCopyDirectory(serverFolder, logFolder, true, true, true);
 
         deleteServerMarkerFile();
 


andymc12:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -2679,6 +2679,11 @@ protected void recursivelyCopyDirectory(RemoteFile remoteDirectory, LocalFile de
                 continue;
             }
 
+            if (remoteDirectory.getName().startsWith("dump_")) {
```
aguibert: I'm worried that this will skip out cases where we actually do want to copy folders that start with `dump_`, since this is added to a  generic method. It seems the root problem here is that `server.dumpServer()` returns while a server dump is not yet complete?

andymc12: @aguibert It's also an issue when the dump was done on a different thread than the test case (which is calling stop server and post stop server archive).  I considered using synchronization to prevent the server from stopping during a dump, but that seemed like overkill.  Do you prefer the synchronization?  Or have another suggestion?  

Thanks for the review comments

aguibert: Is there a particular reason why the server dump is done async? I think it would be better to turn `server.dumpServer()` into a blocking call. On the test client side we generally prefer to do things sequentially because it produces more reliable/repeatable tests and is easier to debug.

aguibert: side note: doing a server dump is a very resource intensive process, and generally we try to avoid doing it in the main path because it consumes so much CPU and disk space. Unless you are explicitly testing the contents of a server dump here, I'd recommend you remove the server dump call, or at least turn it into a failure-only code path (although in general if a server hangs the test framework will take a server dump automatically)

andymc12: This was made to be async so that server dumps could be acquired at various phases of the test at runtime.  It enables users to get thread dump info, as well as other introspection data at various times throughout the test.  The "unpredictable" timing of the dumps is ok in most cases because it is only for diagnosing tests that are already failing.  Ideally, we would get the data from the dumps in all cases, but what we have now are tests failing because we are collecting the dumps - that is no good.

andymc12:
#build

PR #7174, opened 2019-04-08
Add more info when validating cloudant

alexsm82:
Provide more information about the cloudant/couchDB server being validated.

### DIFF
--- a/dev/com.ibm.ws.cloudant/src/com/ibm/ws/cloudant/internal/CloudantService.java
+++ b/dev/com.ibm.ws.cloudant/src/com/ibm/ws/cloudant/internal/CloudantService.java
@@ -246,7 +246,7 @@ public class CloudantService implements ApplicationRecycleComponent, LibraryChan
      * @throws Exception
      */
     @FFDCIgnore(value = {InvocationTargetException.class, PrivilegedActionException.class })
-    Object createResource(String databaseName, boolean createDatabase, int resAuth, List<? extends ResourceInfo.Property> loginPropertyList) throws Exception {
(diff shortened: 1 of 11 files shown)


dev/com.ibm.ws.cloudant/src/com/ibm/ws/cloudant/internal/CloudantService.java
```diff
@@ -597,4 +606,33 @@ protected void setSsl(ServiceReference<Object> ref) { // com.ibm.wsspi.ssl.SSLCo
...
+        if (containerAuthData == null) {
+            SerializableProtectedString protectedPwd = (SerializableProtectedString) props.get(PASSWORD);
+            if (protectedPwd != null) {
+                password = String.valueOf(protectedPwd.getChars());
+                password = PasswordUtil.getCryptoAlgorithm(password) == null ? password : PasswordUtil.decode(password);
+            }
+        } else
+            password = String.valueOf(containerAuthData.getPassword());
+
+        final ClientKey key = new ClientKey(null, userName, password);
```
njr-11: The null parameter here is the class loader for the path where the Cloudant driver is loaded from the application.  I don't think the validators will work properly in combination with the load-from-app feature, although the latter is currently disabled, so it would be more a concern if attempting to enable it at a later point.

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

alexsm82:
#build

PR #7195, opened 2019-04-10
 add validation for custom login modules for JCA connection factories

njr-11:
Add test case to cover validation of custom login module/login properties for JCA connection factories.
Make updates to validator code to support this. It was mostly previously copied from the JDBC data source validator, but needs minor updates which will also be added to the data source validator. 
Fix the JCA code on a path where doPrivileged is lacking, which I found to be failing when running the tests for this.


njr-11:
#build

njr-11:
Build failed with an infrastructure issue, but by now there is another commit on top of this change for #7196, so I'm closing those out because it will deliver with that pull.

PR #7199, opened 2019-04-10
Fix intermittent issue with config updates

brenthdaniel:
It's possible for the delayed variable tests to fail because the test code isn't waiting for a config update to complete.


brenthdaniel:
#build

PR #7200, opened 2019-04-10
Reduce max trace allowed per bucket to 1GB

aguibert:


aguibert:
#build

aguibert:
Tested with 750MB limit and there are 3 buckets that are threatened by the limit:

- com.ibm.ws.microprofile.faulttolerance.1.1_fat_tck -> 950MB
- com.ibm.ws.microprofile.faulttolerance.2.0_fat_tck -> 2.2GB
- com.ibm.ws.microprofile.rest.client_fat -> 950 MB

PR #7210, opened 2019-04-11
transaction times out and releases locks if whenComplete stage unable to run

njr-11:
The limitation to reject propagation of a transaction to multiple threads in parallel can get it in the way of a whenComplete (or other) stage that is intended to resolve the transaction, when the app is poorly coded to allow the completion stage processing to happen while the transaction remains on the original thread.  In this case, we want to ensure that the transaction does eventually roll back (after the transaction timeout) so that locks are not kept on the data forever.  This pull adds a test for this scenario.


njr-11:
#build

njr-11:
Added a commit to switch to upgrade to running the TCK from release candidate 2, which is required for us to be able to force roll back in response to failed stages.
#build

njr-11:
#build

njr-11:
Build has 3 test failures.  All are pre-existing issues in unrelated test buckets.  This includes the known package verification error for open-liberty-adoc.zip

PR #7213, opened 2019-04-11
Enable jdbc_fat_postgresql to run with a remote Docker Host

aguibert:


aguibert:
#build

dev/com.ibm.ws.jdbc_fat_postgresql/fat/src/com/ibm/ws/jdbc/fat/postgresql/FATSuite.java
```diff
@@ -78,10 +53,31 @@ private static boolean tryDockerHost(ExternalTestService dockerService) {
         String m = "tryDockerHost";
         String dockerHostURL = "tcp://" + dockerService.getAddress() + ":" + dockerService.getPort();
         System.setProperty("DOCKER_HOST", dockerHostURL);
+        try {
+            File certDir = new File("docker-certificates");
```
websterc87: Im guessing this directory isnt included in the FAT Output.

dev/fattest.simplicity/src/componenttest/topology/utils/ExternalTestService.java
```diff
@@ -561,9 +561,16 @@ public void writePropertyAsFile(String keyName, File file) throws IOException {
             throw new IllegalStateException("Key not found in service properties: " + keyName);
         }
 
+        String decryptedProp = null;
+        try {
+            decryptedProp = prop.getStringValue();
```
websterc87: I think this will break some mongodb tests, as they have some properties that contain non-UTF-8 characters

Perhaps we can do some switch based on if its encrypted

aguibert:
#build

dev/com.ibm.ws.jdbc_fat_postgresql/fat/src/com/ibm/ws/jdbc/fat/postgresql/FATSuite.java
aguibert: it is, but there are already secrets like this included in FAT output

dev/fattest.simplicity/src/componenttest/topology/utils/ExternalTestService.java
aguibert: good point. I've undone this change and instead done `externalService.getProperties().getProperty(propName)` (which does already decrypt) and then I write that string to a file using `java.io.Files`

PR #7241, opened 2019-04-15
1792 unclear error message mp metrics

### LINKED ISSUE #1792: Unclear error message on mp metrics
If name for the annotations of Timed and Counted are not set but used in combination, you will get the error message:
IllegalArgumentException and the message "[INFO] test.ServiceC.getProperty is already used for a different type of metric"

It is unclear on what is the action required to fix the problem.

Channyboy:
#build 
fixes #1792


Channyboy:
Woops

Channyboy:
#build

PR #7242, opened 2019-04-15
resource adapter for testing config viewer api and initial test

njr-11:
Create a mock resource adapter, from which configuration will be generated, which can be used to test the ibm/api/config REST endpoint with JCA connection factories (and later other JCA resources).  Write an initial test case, or at least what is reasonably possible without significant additional changes.


njr-11:
#build

PR #7247, opened 2019-04-16
Handle async case where app shuts down

Azquelt:
If an application shuts down, no more asynchronous operations can be
started for that application. With FT and async, new async operations
can be started at several points for one method call.

This change ensures that in this case, the user gets a sensible error
message, rather than an FFDC and an internal error exception.


skasund:
Message review completed by L2

Azquelt:
Hi @dmuelle, sorry I hadn't been able to get back to this. I agree with the changes proposed by @jantley-ibm and have added a commit with those changes.

Azquelt:
#build

dmuelle:
@Azquelt looks good! ID reviewed label added

Azquelt:
Build did not complete correctly. Likely an infrastructure issue.

#build

PR #7249, opened 2019-04-16
omit internal JCA properties and fix bug with character type values

njr-11:
The /ibm/api/config  REST endpoint is currently including internal properties such as creates.objectClass for flattened JCA config properties.  These need to be removed.
Also, properties of type character are broken (improperly converted to JSON) and needs to be fixed as well.


njr-11:
#build

PR #7275, opened 2019-04-17
remove mpConcurrency feature after rename completed

njr-11:
Removes the mpConcurrency-1.0 feature which was replaced by/renamed to mpContextPropagation-1.0


njr-11:
#build

Issue #7279, opened 2019-04-18
mpJWT: Improve performance if JWK location is a file.


jdmcclur:
If the jwk location is set as a file, there is a performance issue looking up the file repeatedly. This does not affect a string or remote lookup.

jdmcclur:
Fixed by PR 7283.

PR #7283, opened 2019-04-18
add cache check when jwk is in file

brutif:
resolves performance problem #7279


brutif:
#build

brutif:
pb is good, failures are unrelated

PR #7285, opened 2019-04-18
JCA admin objects via config rest api

njr-11:
Add tests for accessing JCA admin objects via the /ibm/api/config REST API.


njr-11:
#build

PR #7297, opened 2019-04-22
hide password values of JCA resources from config REST endpoint output

njr-11:
Even if the resource adapter doesn't declare a property as confidential, if the property name contains "password", then don't include its value in the output of the /ibm/api/config/ REST endpoint.


njr-11:
#build

PR #7302, opened 2019-04-22
tests for querying activationSpec via the config REST endpoint

njr-11:
Create an activation spec for the mock resource adapter.
Configure a couple of activationSpec config elements in server configuration.
Write tests that access them via the /ibm/api/config REST endpoint and verify the output.


njr-11:
#build

PR #7321, opened 2019-04-23
validator can use vendor-specific testConnection method for CICS and IMS

njr-11:
CICS and IMS ManagedConnection classes have a vendor-specific method named testConnection, which takes no parameters, which must be used to validate a connection.  It is not enough just to use the CCI API to obtain a Connection and Interaction. These do not actually cause a connection to be established to the backend, and would thus result in false positives for the validator.


njr-11:
#build

PR #7327, opened 2019-04-24
 tests for vendor-specific testConnection method used by validator

njr-11:
Write test cases for a vendor-specific testConnection() method supplied by the ManagedConnection implementation that can be used by the validator of JCA connection factories to cover the case where the resource adapter lazily connects to the backend.


njr-11:
#build

njr-11:
Build has 1 test failure, which is the known Packaging Verification error about open-liberty-adoc.zip

PR #7341, opened 2019-04-25
Stop reading feature cache

anjumfatima90:
Stop reading feature.cache file for the installed features. Add a trace message which displays all installed features in FeatureManager. Read the installed features from trace.log


anjumfatima90:
#build

PR #7348, opened 2019-04-26
Use a faster way to get current application name

fmhwong:
Use a faster way to get current application name.  Keeping the existing way as a fallback if the new way returns null.


fmhwong:
#build

fmhwong:
#build

fmhwong:
The packaging error in PB is not related to this changes.

PR #7364, opened 2019-04-29
Build OS specific package (rpm/deb) integration

zhangm13:
We would like to integrate the changes to the Open Liberty build that create rpms and debs from the already generated Openliberty All zip


zhangm13:
Added some more changes to another branch and will open a new PR

PR #7366, opened 2019-04-29
Update feature description to indicate stabilization.

frowe:
Change feature description to make it obvious it is stabilized.

### DIFF
--- a/dev/com.ibm.websphere.appserver.features/visibility/public/mongodb-2.0/resources/l10n/com.ibm.websphere.appserver.mongodb-2.0.properties
+++ b/dev/com.ibm.websphere.appserver.features/visibility/public/mongodb-2.0/resources/l10n/com.ibm.websphere.appserver.mongodb-2.0.properties
@@ -15,4 +15,7 @@
 #
 
 description=This feature has been stabilized. \
-The strategic alternative to this feature is described at https://openliberty.io/blog/2019/02/19/mongodb-with-open-liberty.html.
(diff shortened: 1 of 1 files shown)


frowe:
#build

dev/com.ibm.websphere.appserver.features/visibility/public/mongodb-2.0/resources/l10n/com.ibm.websphere.appserver.mongodb-2.0.properties
```diff
@@ -14,6 +14,5 @@
 #NLS_MESSAGEFORMAT_NONE
 #
 
-description=This feature enables the usage of the MongoDB Java Driver and allows DB instances to be \
- configured in the server configuration, injected into managed components such as EJBs, and accessed \
- through JNDI. Applications interact with these DB instances via the MongoDB APIs
+description=This feature has been stabilized. \
```
aguibert: Can we append this information to the existing feature description? We should still have the description of what this feature does, even if it is stabilized and there is a more strategic alternative.

frowe: We could, I was just trying to make it harder to use since we want people to move off.  I'll prepend this info so it comes first.

frowe:
The only failure in PB was a packaging verification error which Andy pushed a fix for yesterday after my build.

PR #7367, opened 2019-04-29
Remove extra debug when updating test counts

naumanna:
Removing some unnecessary debug I had added under #7337.  Just keeping the updated test counts in case there is another problem in the area in the future.


naumanna:
#build

PR #7370, opened 2019-04-29
Add Java SE 12 as a support Java level for feature doc

aguibert:


aguibert:
#build

NottyCode:
This might affect how the Liberty repository displays Java baselines, but it won’t affect the generated feature doc for KC. That is based on the feature list generated data. 

Can the min version ever be 1.6.0? We don’t support it anymore so if it is wouldn’t that suggest our ESAs aren’t quite built correctly?

NottyCode:
You need to update this class for the documentation to list Java SE 12:

https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.kernel.feature.cmdline/src/com/ibm/ws/kernel/feature/internal/generator/FeatureList.java

aguibert:
#build

aguibert:
#build

PR #7381, opened 2019-04-30
objects should not be wrapped in json arrays when cardinality is 1

njr-11:
JSON generated by the /config/ endpoint is inconsistent between cardinality 0 and 1, where from the user's perspective they are configuring a single value.  When cardinality is 1 (or -1), an array is being used to wrap the JSON object.  When cardinality is 0, the JSON object is not wrapped with an array.  We should detect the cardinality and generate the JSON consistently.  The array seems unnecessary for something that can never have more than one value, and so should be omitted.


njr-11:
#build

njr-11:
Build has 3 test failures which all match preexisting issues in unrelated test buckets.

PR #7395, opened 2019-05-01
 test case for JAX-RS context propagation

njr-11:
Write a test for propagation of JAX-RS context.  The test will be delivered in working state where context propagation is not needed because JAX-RS creates a new instance of the `@Path` resource per request, which can then be used consistently from any thread.  The test case includes some TODOs for switching it to singleton or CDI usage patterns, where it will not work, along with comments regarding a possible solution involving CDI context propagation that would require the URIInfo (or other) to be a CDI request scoped bean, which could then be propagated by MP Context Propagation when CDI context propagation is enabled.

It should be noted that we are not currently trying to offer a solution that makes `@Context` work in all cases because there are plans to deprecate `@Context` from JAX-RS in favor of a CDI solution to be added in the future.


njr-11:
#build

PR #7419, opened 2019-05-03
add security trace to jaxrs SecurityAnnotationTest

WhiteCat22:
adding security trace to debug a build break


WhiteCat22:
#build

PR #7435, opened 2019-05-06
Issue #7434-Fix NPE with Stateful access-timeout

### LINKED ISSUE #7434: NullPointerException in MethodAttribUtils.getXMLCMCLockAccessTimeout
In an EJB 2.x stateful session bean specifies an access timeout in ejb-jar.xml (<concurrent-method><access-timeout>), then a NullPointerException will occur in MethodAttribUtils.getXMLCMCLockAccessTimeout() when starting the application.

tkburroughs:
EJB 2.x stateful beans (with a home) have an extra method added to the
list of method attributes to handle the remove method on the home. metadata
for the method is set, but not the method reflect object, which is
used by MethodAttribUtils; so an NPE occurs processing <access-timeout>
for such a scenario.

A null check has been added to skip further checking; access-timeout would
never apply to the home method.

Please excuse the eclipse re-formatting; the only change is on line 2371 (new version); which was to add a null check for method not being set for home remove.

Note, a test will be added soon in the ejbcontainer.remote fat project.

fixes #7434

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2014 IBM Corporation and others.
+ * Copyright (c) 1998, 2019 IBM Corporation and others.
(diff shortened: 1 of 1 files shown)


tkburroughs:
#build

tkburroughs:
Test failures are not related; EJB features are not in use.  Can merge once buildcon improves.

PR #7468, opened 2019-05-09
Custom auth filter

utle:
Provide customer a way to specify the request header without the value.

### DIFF
--- a/dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties
+++ b/dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties
@@ -30,6 +30,8 @@ name.desc=Specifies the name.
 
 value=Value
 value.desc=Specifies the value.
+valueOptional.desc=Specifies the value. If value is not specify, matching is done with the key not value.
+
 
 userAgent=User Agent
 userAgent.desc=Specifies the user agent.
--- a/dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties
+++ b/dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties
@@ -30,8 +30,7 @@ name.desc=Specifies the name.
 
 value=Value
 value.desc=Specifies the value.
-valueOptional.desc=Specifies the value. If value is not specify, matching is done with the key not value.
-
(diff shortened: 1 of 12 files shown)


utle:
@chirp1, can you review the  dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties

utle:
#build

dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -30,6 +30,8 @@ name.desc=Specifies the name.
 
 value=Value
 value.desc=Specifies the value.
+valueOptional.desc=Specifies the value. If value is not specify, matching is done with the key not value.
```
Charlotte-Holt: Hi @utle - Could you help me fill in a few blanks here? I'm thinking this will end up something like:

"The value **<is this an attribute, etc.?>** specifies the value of **<what value does this specify?>**. If the value is not specified, then **<what is matched with the key?>** is matched with the key. "

What would be the correct information to insert for the bolded items?

Charlotte-Holt:
Hi @utle - I asked you a few questions about the message in dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties. After we work through those and come to an agreement, I'll add the ID reviewed label. Thanks!

dev/com.ibm.ws.security.authentication.filter/resources/OSGI-INF/l10n/metatype.properties
utle: This is the HTTP request header name and value. For example email=myemail@xyz.com... so we don't really know the attribute name or value. Before this change, the value is required but now optional so I create a new valueOtional.desc.

Charlotte-Holt: @utle Agreed on:
"The value attribute specifies the HTTP request header value. If the value is not specified, then matching is done with the name attribute, not the value attribute."
Thanks!

utle: Done.

utle:
#build

andrea-pichler:
No message review from L2 required.

PR #7483, opened 2019-05-09
app defined resources

njr-11:
Add implementation of generated nested config for app-defined resources, such as connectionManager and jdbcDriver.  Add tests for these to show in the /config/ REST endpoint.


njr-11:
#build

njr-11:
Build has 1 failure, which is a duplicate of a pre-existing issue in an unrelated test bucket.

PR #7486, opened 2019-05-10
Fix to include specific classes and add OSGi 1.8 to version range

fritze2:
- Copy specific classes from com.ibm.ws.kernel.feature.core via build.gradle rather than using bnd.bnd Import-Package header.
- Change version range for OSGi to include 1.8 to fix WDT on Eclipse Oxygen.
- Compile at Java 6 for migration tools.


fritze2:
#build

fritze2:
#build

PR #7492, opened 2019-05-10
Support for PostgreSQL setDefaultFetchSize and setPrepareThreshold props

aguibert:


dev/com.ibm.ws.jdbc_fat_postgresql/test-applications/postgresqlApp/src/jdbc/fat/postgresql/web/PostgreSQLTestServlet.java
```diff
@@ -404,6 +405,75 @@ public void testPropertyCleanup_schema() throws Exception {
...
+            verifyClean(con);
+            PGConnection pgCon = con.unwrap(PGConnection.class);
+            defaultFetchSize = pgCon.getDefaultFetchSize();
+
+            assertTrue("Was expecting default value to be anything other than 5", defaultFetchSize != 5);
+            pgCon.setDefaultFetchSize(5);
+            assertEquals(5, pgCon.getDefaultFetchSize());
+        }
+        try (Connection con = ds.getConnection()) {
+            verifyClean(con);
```
njr-11: I know that 'verifyClean' was written as part of this pull, however, the test case uses it, and so I was curious what it was doing.  I see this at the start of it which seems very suspicious,
```
        // Always "do some work" with the connection before we verify it's underlying state.
        // The JDBC code intentionally lazily resets connection values, so our wrapper may be out of sync with the underlying
        // connection between getting the initial connection and actually driving some work on it
        con.createStatement().close();
```
Are we covering up a bug where our connection wrapper reports wrong values until someone invokes `createStatement`?  If so, that seems like a serious problem.  For example, imagine if the user has code like this - they could run under the wrong isolation level,
```
if (con.getTransactionIsolation() != Connection.TRANSACTION_REPEATABLE_READ)
   con.setTransactionIsolation(Connection.TRANSACTION_REPEATABLE_READ);
con.createStatement().execute(sql) ...
```

dev/com.ibm.ws.jdbc/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -768,6 +768,8 @@ portNumber=Port number
 portNumber.desc=Port on which to obtain database connections.
 prefetch=Prefetch
 prefetch.desc=JDBC driver property: prefetch.
+preparedStatementCacheQueries=Prepared Statement Cache Queries
```
Charlotte-Holt: Change 
"Prepared Statement Cache Queries" 
to sentence case
"Prepared **s**tatement **c**ache **q**ueries"

dev/com.ibm.ws.jdbc/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -768,6 +768,8 @@ portNumber=Port number
...
+preparedStatementCacheQueries.desc=JDBC driver property: preparedStatementCacheQueries.
```
Charlotte-Holt: Change
"JDBC driver property: preparedStatementCacheQueries."
to
"The JDBC driver property is preparedStatementCacheQueries."

njr-11: I don't agree with this change unless you are willing to change every other occurrence in this file.  This file currently has 431 messages using the form, "JDBC driver property: ...", and 0 occurrences using the recommended form, "The JDBC driver property is ...".   Please ensure that all of them remain consistent.

dev/com.ibm.ws.jdbc_fat_postgresql/test-applications/postgresqlApp/src/jdbc/fat/postgresql/web/PostgreSQLTestServlet.java
aguibert: no, I don't think we are covering up a bug here. We have an optimization in the JDBC code (which causes more complexity than it was probably worth IMO) where we intentionally let our connection handle and the underlying connection impl be out of sync until the user attempts to do something meaningful with the connection. 

The connection handle that the user app deals with always has correct values, it's the underlying impl object which may have "wrong" values at first, but we correct this lazily as the connection is used. In the code example you gave with setting the tx isolation level, it would play out like this:
1. handle has isolation=RR, underlying impl may have something else (let's say SER)
2. when we attempt to create a statement, we first align connection properties by calling setTxIsolation(RR) on the underlying connection
3. statement is created/executed with RR isolation level

aguibert:
#build

andrea-pichler:
No message review from L2 required.

dev/com.ibm.ws.jdbc/resources/OSGI-INF/l10n/metatype.properties
Charlotte-Holt: @njr-11 I understand, and thanks for this information. Disregard my previous suggestion on this one.

dev/com.ibm.ws.jdbc_fat_postgresql/test-applications/postgresqlApp/src/jdbc/fat/postgresql/web/PostgreSQLTestServlet.java
njr-11: Because verify clean does a con.createStatement, which syncs up values, it is depriving the test of actually checking the value on the connection handle prior to the sync.  If it is included, it should be moved further down in the test until after the value of the handle is checked.  Also, the createStatement within verifyClean should probably be moved further down within that method until after the values of the handle are checked.

aguibert:
#build

PR #7494, opened 2019-05-10
 Add comments and change how generated.properties is stored

fritze2:
- Changed some of the Gradle scripts commenting.
- Moved where generated.properties is written to disk.
- Changed generated.properties to only include properties we need to save for future builds.
- Moved gradle to wlp-gradle so that autocompletion on `./gradlew` works.


fritze2:
#build

fritze2:
#build

PR #7507, opened 2019-05-13
default config for app defined resource

njr-11:
Include default values for config attributes when an application-defined resource does not specify any value for the attribute.


njr-11:
#build

njr-11:
build has 1 test failure, which is a pre-existing issue

PR #7516, opened 2019-05-14
type conversion for app defined resource config

njr-11:
Config for application-defined resources often has the value represented as a String (as it was input by the user) and needs to be converted to the correct type per the metatype so that generated JSON for the /config/ REST endpoint can be consistent between app-defined data sources and dataSource elements defined in server config.


njr-11:
#build

njr-11:
Build has 4 test failures which are pre-existing issues

PR #7519, opened 2019-05-14
app-defined connectionManager jdbcDriver and parse error

njr-11:
Add tests for application-define connectionManager and jdbcDriver being accessed with the /config/ REST endpoint.  Also handle and test what happens when there is a parsing error on a duration type attribute.  And include coverage of the /config/ endpoint for app-defined resources in the java:app and java:comp namespaces.


njr-11:
#build

njr-11:
Build has 1 test failure, which is a preexisting issue

PR #7525, opened 2019-05-15
REST endpoints for app-defined datasource in java:global

njr-11:
Test the /config/ and /validator/ REST endpoints for an app-defined DataSource in the java:global namespace.


njr-11:
#build

PR #7532, opened 2019-05-15
validate QueueConnection and TopicConnection

njr-11:
Validation code for QueueConnectionFactory and TopicConnectionFactory is currently validating only with a generic JMS connection, not the QueueConnection and TopicConnection that are specific to these connection factories.  Need to update the validation code to include this coverage.


njr-11:
#build

PR #7543, opened 2019-05-16
test config REST endpoint for multiple app-defined data sources with the same JNDI name

njr-11:
It is possible to define multiple app-defined data sources with the same JNDI name, if they are scoped to different components/modules/applications.  Write a test to ensure that the /config/ REST endpoint handles this correctly and returns the correct configuration for the respective data source, even if both have the same JNDI name.


njr-11:
#build

PR #7356, opened 2019-04-27
Update localConnector-1.0 to not depend on JDK internal APIs

aguibert:

Issue #7542, opened 2019-05-16
Unable to retrieve connector address for localConnector of started VM


evelinec:
All of the OL guides are tested with the daily OL development builds. Today we found that the `arquillian-managed` guide tests with JDK11 failed (test error message below). After some investigation, the failure was reproducible locally starting with the recently daily build (described below). 

Test error message: 
```[AUDIT   ] CWWKF0012I: The server installed the following features: [cdi-2.0, jaxrs-2.1, jaxrsClient-2.1, jndi-1.0, jsonp-1.1, localConnector-1.0, servlet-4.0].
[INFO    ] CWWKF0008I: Feature update completed in 6.005 seconds.
[AUDIT   ] CWWKF0011I: The ArquillianManagedServer server is ready to run a smarter planet. The ArquillianManagedServer server started in 8.390 seconds.
[INFO    ] SRVE9103I: A configuration file for a web server plugin was automatically generated for this server at /home/travis/build/OpenLiberty/guide-arquillian-managed/finish/target/liberty/wlp/usr/servers/ArquillianManagedServer/logs/state/plugin-cfg.xml.
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 31.488 s <<< FAILURE! - in it.io.openliberty.guides.inventory.InventoryArquillianTests
[ERROR] it.io.openliberty.guides.inventory.InventoryArquillianTests  Time elapsed: 31.484 s  <<< ERROR!
org.jboss.arquillian.container.spi.client.container.LifecycleException: Could not start container
Caused by: org.jboss.arquillian.container.spi.client.container.LifecycleException: Unable to retrieve connector address for localConnector of started VM
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0 s <<< FAILURE! - in it.io.openliberty.guides.system.SystemArquillianTests
[ERROR] it.io.openliberty.guides.system.SystemArquillianTests  Time elapsed: 0 s  <<< ERROR!
java.lang.RuntimeException: Arquillian initialization has already been attempted, but failed. See previous exceptions for cause
Caused by: org.jboss.arquillian.container.spi.client.container.LifecycleException: Could not start container
Caused by: org.jboss.arquillian.container.spi.client.container.LifecycleException: Unable to retrieve connector address for localConnector of started VM
[ERROR] Errors: 
[...]

PR #7544, opened 2019-05-17
Revert "Update localConnector-1.0 to not depend on JDK internal APIs"

cbridgha:
Reverts OpenLiberty/open-liberty#7356


cbridgha:
Deciding to revert this for now, and resolve problems that caused #7542

PR #7565, opened 2019-05-20
Increase wait between docker host attempts from 15s -> 120s

aguibert:


aguibert:
Merging this PR with no pbuild, since we are simply increasing a timeout that only occurs on an error path, so unless my build happened to hit the error path (less than 1/100 chance) this code change wouldn't be tested anyway.

Issue #7180, opened 2019-04-09
JPMS options are not applied when JAVA_HOME is not set


aguibert:
Steps to reproduce:

1. Add Java 11 to $PATH
2. Ensure $JAVA_HOME is unset
3. Create a server and enable `javaee-8.0`
4. Start server, should see JPMS warning in logs:
```
$ serv run --clean

Launching defaultServer (WebSphere Application Server 19.0.0.4/wlp-1.0.27.201904011424) on Eclipse OpenJ9 VM, version 11.0.2+9 (en_US)
[AUDIT   ] CWWKE0001I: The server defaultServer has been launched.
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.ibm.ws.util.ThreadContextAccessor (file:/Users/aguibert/dev/git/open-liberty/dev/build.image/wlp/lib/com.ibm.ws.container.service_1.0.27.jar) to field java.lang.Thread.contextClassLoader
WARNING: Please consider reporting this to the maintainers of com.ibm.ws.util.ThreadContextAccessor
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
...
```

aguibert:
While chatting with some of the JDK folks recently I was informed about the `JDK_JAVA_OPTIONS` environment variable which is automatically read by the JDK launcher for Java 9+, making it the perfect place to specify JPMS options.

We should go with this solution instead of the approach where we search for `${JAVA_HOME}/lib/modules`

PR #7567, opened 2019-05-20
Apply JPMS options for JDK 9+ even if JAVA_HOME is not set

aguibert:


aguibert:
Fixes Linux/Mac side of #7180

aguibert:
#build

PR #7569, opened 2019-05-20
match up the impl for the config and validator REST handlers so that they can later be consolidated

njr-11:
As a first step toward consolidation of ConfigBasedRESTHandler and ConfigRESTHandler, merge the two approaches as much as possible such that the two are no longer taking divergent (and incompatible) paths for identifying the config elements to validator/display.  A future pull will perform the actual consolidation.


njr-11:
#build

njr-11:
Build has 1 test failure which is a duplicate of a preexisting issue.

PR #7584, opened 2019-05-21
rebase ConfigRESTHandler onto ConfigBasedRESTHandler

njr-11:
ConfigRESTHandler duplicates a lot of code from ConfigBasedRESTHandler.  The latter was intended to be a common implementation for configuration-based REST endpoints, so we ought to be using it for the former. This will also eliminate the possibility that the two become out of sync again.


njr-11:
#build

njr-11:
commit was also included under #7591 and merged with that pull, so I'm closing this out

PR #7591, opened 2019-05-21
test that validates app defined datasources

njr-11:
Add test coverage for validation of app-defined data sources at the various scopes.


njr-11:
#build

njr-11:
A change is needed to a commit that was included in this pull because we cannot make a fix to a connection management exception path until tests in another repo are pre-emptively updated to accommodate.  I added a commit to temporarily comment out this part of the changes.
#build

Issue #7595, opened 2019-05-22
java.lang.IllegalStateException: This FAT has used up too much trace!


Rachelwrq:
The voluminous logging is caused by this message being traced thousands of times per second:

```jmx.connector.client.rest.internal.RESTMBeanServerConnection 1 run Connection refused (Connection refused)```

PR #7596, opened 2019-05-22
Added callThreadWait method to prevent quick exit

Rachelwrq:
Added a new method called `callThreadWait`. 
This method is invoked when there are multiple threads trying to do the same connection recovery step. The first thread would the do the connection recovery while the later thread would wait until it gets notified, `waitTime` is up, or the servers shuts down after maximum retry time. 
This method is necessary because it allows later threads to wait instead of quick exiting the `recoverConnection` method and retry immediately without resting.


Rachelwrq:
The changes were tested through personal builds on RTC. I build the associated fat bucket 20 runs each time for 2 times, and all tests have passed.

Rachelwrq:
fixes #7595

PR #7599, opened 2019-05-22
tracing for configuration-based REST endpoints

njr-11:
Standardize the trace strings (currently we have config.rest, rest.config, rest.validator).
The former two sound like the same thing and are awkward.
The latter one is inconsistent with the REST endpoint (validation, not validator).
Also, it would be nice if at least one part of the tracing formatted the JSON in a readable way.


njr-11:
#build

Issue #7601, opened 2019-05-22
Mongo should wait for configured SSL instances to be available


brenthdaniel:
The MongoDB service has a tendency to bounce repeatedly as a result of SSL activation. We should be using cardinality.minimum to prevent the MongoDB component from activating before SSL (assuming SSL is configured.)

PR #7602, opened 2019-05-22
Use minimum cardinality property to prevent mongo from activating bef…

brenthdaniel:
...ore SSL

PR for #7601


brenthdaniel:
#build

PR #7615, opened 2019-05-23
ConnectionFactoryDefinition properties are lost if there are no default values to override

njr-11:
Add an app-defined connection factory to the automated test bucket for the config REST endpoint, including an initial test of the portions that are already working after making an update to the config endpoint code to allow for how the ConnectionFactoryResourceBuilder only registers configuration for app-defined connection factories under the supertype pid.

This is also fixing a bug where properties of `@ConnectionFactoryDefinition` are lost if the resource adapter does not also supply a default value for the configured property.


njr-11:
Accidentally pushed another commit to this pull instead of creating a new pull. I'll need to submit another build now.
#build

njr-11:
Build has 3 test failures (pre-existing timeout issue) in an unrelated bucket.

PR #7630, opened 2019-05-24
app-defined connectionfactory props for config rest endpoint

njr-11:
Name/value pair properties that are specified via `@ConnectionFactoryDefinition(properties=...)` are all showing up as String instead of the proper types per the generated metatype.  Also, the JSON object containing all of the properties is being placed within an extra array, meaning it's not getting the cardinality correct.  Make these improvements and enable the tests for them.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build has 26 failures across 5 buckets, all of which are preexisting infrastructure issues - inability to reach external resources (selenium servers or ldap).

PR #7641, opened 2019-05-28
config endpoint for second app-defined connection factory using same jndiName

njr-11:
Test coverage for multiple app-defined connection factories with the same JNDI name, but different scopes.  Specify different connection factory interfaces and other properties, but with the same JNDI name. Verify that the config REST endpoint shows the correct set of properties for each.


njr-11:
#build

PR #7655, opened 2019-05-29
REST endpoint for app-defined JMS connection factory

njr-11:
Add test case and update code to support using the /config/ REST endpoint to access information about an application-defined JMS connection factory (JMSConnectionFactoryDefinition)


njr-11:
#build

PR #7662, opened 2019-05-29
implementing unit conversion for Timed metric in JSON

### LINKED ISSUE #7440: Investigate possible difference in values between Prometheus and JSON format metrics
per https://github.com/eclipse/microprofile-metrics/issues/320 the value shown in prometheus format from `/metrics` isn't the same value as from JSON format.  

```
prometheus value = 0.000043196000000000005
json value = 43196.0
```
Example endpoint
```
@Path("/")
public class ResourceMeteredTimed {
    @Timed(name = "timed",
            description = "Metrics to monitor the times of timed method. - @Timed",
            unit = MetricUnits.MINUTES,
            absolute = true)
    @GET
    @Path("/timed")
    public Response timed() {
        return Response.ok().build();
    }
}
```

jennifer-c:
fixes #7440 
#build


jennifer-c:
#build

fmhwong:
I will defer the review to @Channyboy since he is the SME.

PR #7666, opened 2019-05-30
Fix ignoreOutliers method in mpFT timeout tests

Azquelt:
Bug in getSampleOfConnectTimesAverage meant that the time calculated
included the delay waited before running the test samples.


Azquelt:
#build

PR #7668, opened 2019-05-30
REST endpoint for app-defined TopicConnectionFactory

njr-11:
Add test coverage for using the /config/ REST endpoint to access the configuration of an application-defined javax.jms.TopicConnectionFactory.  Also, update the implementation so that it works properly.


njr-11:
#build

PR #7682, opened 2019-05-31
Allow internal ConfigSources to be provided using DS

tevans78:
This PR is a prototype which allows other internal components to provide ConfigSources as DS Services. It is not yet tested and not yet ready for delivery.


tevans78:
The prototype Config 1.4 code will need to be slightly updated too, it doesn't compile right now!

tevans78:
#build

tevans78:
Don't know why the previous PB failed so running again.
#build

tevans78:
Another odd failure...
#build

PR #7690, opened 2019-05-31
Perf improvement during getLogger flow

NottyCode:
Try to remove call to getStackTrace to speed up creation of new Loggers.


NottyCode:
Looks like this removes ~20ms in server startup flow

#build this again please Libby

NottyCode:
#build this again because the last build failed.

PR #7699, opened 2019-06-02
Give more clear error message when attempting to unwrap to non-interface class

aguibert:


aguibert:
#build

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/jdbc/WSJdbcWrapper.java
```diff
@@ -447,18 +447,19 @@ else if (ifcToDynamicWrapper.containsKey(interfaceClass))
...
-            {
+        else {
+            try {
                 if (tc.isDebugEnabled())
                     Tr.debug(this, tc,
                              "No existing wrappers found. Need to create a new wrapper.");
 
+                if (!interfaceClass.isInterface()) {
+                    throw new SQLException(AdapterUtil.getNLSMessage("NOT_AN_INTERFACE", interfaceClass.getName()));
+                }
```
njr-11: The placement of this condition has the potential to break vendor-specific enhancements that go beyond the spec to allow unwrap to non-interface classes.  It should be moved to with the block,
```
if (implObject == null) { // Not found
```
in order to preserve compatibility here.

aguibert: this would not break existing behavior, because even if `getJDBCImplObject(interfaceClass);` did return non-null on the next line, we still throw a SQLException if the passed in interfaceClass is not an interface.

njr-11: ok, so the check was already in there.  I was looking at the new code rather than the original and missed that.

PR #7710, opened 2019-06-03
Stop writing plugin-cfg.xml on every startup

NottyCode:


NottyCode:
It seems this speeds up server startup by around 30ms

#build this please Libby

PR #7714, opened 2019-06-03
hide internal properties from config REST endpoint and remove TODO

njr-11:
The application-defined resources path adds some internal properties (jndiName.unique and creates.objectClass) that are not identified within the metatype and thus were not being recognized as internal properties to omit from the /config/ REST endpoint output.  This pull adds handling for these properties.

Also, remove a TODO comment about a possible bug which turned out not to be the case.  JMS connection factory properties aren't ignored in the absence of defaults because code is already in place which steps through all of the properties that are defined in the metatype, looking for each within the annotation values.


njr-11:
#build

Issue #7563, opened 2019-05-20
Better handling of command line error scenarios


brenthdaniel:
There are a couple of error scenarios with command line scripts where it is nearly impossible to diagnose the actual problem. 

1) A "bad certificate" error can prevent the client from communicating the command to the server, but the error is not reported on the client side. 

2) If the client does not have write permissions on the workarea directory the server will not accept the command (by design.) The client does not indicate why the command fails -- it just says to check the logs.

PR #7715, opened 2019-06-03
Add error messages, tests, and delete duplicate property files

brenthdaniel:
PR for #7563


brenthdaniel:
#build

brenthdaniel:
random unrelated failure
#build again

dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
```diff
@@ -756,6 +756,8 @@ info.resumeFailedException=Resuming the {0} server failed. Check the server logs
 info.resumingListeners.target=Resuming the specified target components of the {0} server.
 info.resumedListeners.target=Resuming the specified target components of the {0} server completed.
 info.resumeFailedException.target=Resuming the specified target components of the {0} server failed. Check the server logs for details.
+info.serverCommandAuthFailure=The command {0} failed because the process could not delete the file {1}.
```
Charlotte-Holt: Change
"The command {0} failed because the process **could** not delete the file {1}."
to
"The command {0} failed because the process **might** not delete the file {1}."

brenthdaniel: Hi, 
I'm not sure that change makes sense here. If this error message is issued, it is definite that the file was not able to be deleted. 'Might' would imply that it may or may not have been able to delete the file. It would also be inconsistent with other failures from the same script, all of which use 'could not' when something was not able to be completed.

Charlotte-Holt: Hi @brenthdaniel. I understand, and thanks for your feedback. Okay to leave it as is.

PR #7718, opened 2019-06-03
config REST endpoint for app-defined AdministeredObjectDefinition

njr-11:
Configure an app-defined administered object in a test case and access it via the /config/ REST endpoint, verifying its output.  Make necessary updates to product code such that this is possible.


njr-11:
#build

PR #7731, opened 2019-06-04
Fix ability to run FATs without executing all unit tests

naumanna:
We received reports that running the `buildandrun` task on fat projects suddenly also caused lots of unit tests to run across all of the OL projects, causing lots of wasted time executing FATs locally.

This change fixes the dependencies so that running `buildandrun` doesn't also try to execute the `test` task against all subprojects.


naumanna:
#build

naumanna:
#build

PR #7756, opened 2019-06-05
update test JavaDoc that still refers to MP Concurrency

njr-11:
JavaDocs for various test cases still refer to MP Concurrency or MicroProfile Concurrency.  These need to be updated to instead refer to MP Context Propagation, so as not to cause confusion for developers who are maintaining these tests in the future.


njr-11:
#build

KyleAure:
Looks good to me.  Per the issue I did not see any other references to MP Concurrency or MicroProfile Concurrency in these test buckets.

Issue #7764, opened 2019-06-05
Possible deadlock in ConfigurationStore


tjwatson:
PR #7551 introduced a possible deadlock that looks like the following:

```
3XMTHREADINFO      "Default Executor-thread-6" J9VMThread:0x0000000002304500, omrthread_t:0x00007FC20C66F150, java/lang/Thread:0x00000000E092CD00, state:B, prio=5
3XMJAVALTHREAD            (java/lang/Thread getId:0x37, isDaemon:true)
3XMTHREADINFO1            (native thread ID:0x32C3, native priority:0x5, native policy:UNKNOWN, vmstate:B, vm thread flags:0x00000281)
3XMTHREADINFO2            (native stack address range from:0x00007FC1C9ED0000, to:0x00007FC1C9F11000, size:0x41000)
3XMCPUTIME               CPU usage total: 1.096510017 secs, current category="Application"
3XMTHREADBLOCK     Blocked on: com/ibm/ws/config/admin/internal/ConfigurationStore@0x00000000E014D668 Owned by: "Config-Thread-10" (J9VMThread:0x0000000001E3E700, java/lang/Thread:0x00000000E0486660)
3XMHEAPALLOC             Heap bytes allocated since last GC cycle=0 (0x0)
3XMTHREADINFO3           Java callstack:
4XESTACKTRACE                at com/ibm/ws/config/admin/internal/ConfigurationStore.save(ConfigurationStore.java:171)
5XESTACKTRACE                   (entered lock: com/ibm/ws/config/admin/internal/ConfigurationStore@0x00000000E014D668, entry count: 1)
4XESTACKTRACE                at com/ibm/ws/config/admin/internal/ExtendedConfigurationImpl.updateCache(ExtendedConfigurationImpl.java:480)
4XESTACKTRACE                at com/ibm/ws/config/xml/internal/ConfigUpdater.updateConfiguration(ConfigUpdater.java:265)
4XESTACKTRACE                at com/ibm/ws/config/xml/internal/ConfigUpdater.updateConfiguration(ConfigUpdater.java:194)
4XESTACKTRACE                at com/ibm/ws/config/xml/internal/ConfigUpdater.update(ConfigUpdater.java:109)
4XESTACKTRACE                at com/ibm/ws/config/xml/internal/ConfigUpdater.update(ConfigUpdater.java:95)
[...]

PR #7765, opened 2019-06-05
Use read/write locks for synchronization

tjwatson:
Also add a separate lock for saveTask to avoid out of
order locks with the configStore configurations map lock

For issue #7764


tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

PR #7776, opened 2019-06-05
validation for app-defined JCA connection factories

njr-11:
Test coverage for the validation REST endpoint for JCA connection factories (includes javax.resource.cci.ConnectionFactory and javax.sql.DataSource as a JCA connection factory).
This also includes code updates to make the validation REST endpoint possible for app-defined connection factories.


njr-11:
#build

PR #7794, opened 2019-06-06
first step toward using standard MPConfig property names

njr-11:
In order to align with standards being defined for MicroProfile specs that define their own MicroProfile Config properties, we need to change the name of 7 properties that MicroProfile Context Propagation uses to supply default values for ManagedExecutor/ThreadContext instances.  Per https://github.com/eclipse/microprofile/pull/100 these property names should be prefixed with mp.[spec-name]. and per eclipse/microprofile-config#420 these property names should avoid the use of /.

This pull will make OpenLiberty tolerate either form of the property names for now.  Then, once the spec changes goes through, we can remove support for the old names.


njr-11:
#build

PR #7795, opened 2019-06-06
Changed TBWSessionStorer to non-configurable 1 second

### LINKED ISSUE #7634: "writeInterval" in httpSessionDatabase configuration is not honored
I tested following configuration with writeInterval set to 1s, 5s and 30s. 
The trace shows the actual writeInterval is 10 secs regardless of the configured value. 

```
    <httpSessionDatabase dataSourceRef="sessionDS"
        useMultiRowSchema="true" writeContents="ONLY_UPDATED_ATTRIBUTES"
        writeFrequency="TIME_BASED_WRITE" writeInterval="1s"/>
```

Following repo has a Docker file and recreation steps 
https://github.com/una-tapa/LibertySessionDBTest

trace
https://github.com/una-tapa/LibertySessionDBTest/blob/master/trace.log

ellen-lau:
Fixes #7634 

#build

### DIFF
--- a/dev/com.ibm.ws.session/src/com/ibm/ws/session/SessionContext.java
+++ b/dev/com.ibm.ws.session/src/com/ibm/ws/session/SessionContext.java
@@ -352,7 +352,12 @@ public class SessionContext {
     public IStorer createStorer(SessionManagerConfig smc, IStore store) {
         IStorer storer = null;
         if (smc.getEnableTimeBasedWrite()) {
-            storer = new TBWSessionStorer(store, 10); // 10 seconds is fixed - not configurable
+            if (smc.getPropertyWriterInterval() < 10) {
+                storer = new TBWSessionStorer(store, (int)smc.getPropertyWriterInterval());
+            }
+            else {
+                storer = new TBWSessionStorer(store, 10); // 10 seconds is fixed - not configurable
+            }
         } else if (smc.getEnableManualWrite()) {
             storer = new ManualSessionStorer();
         } else {


una-tapa:
Please do not merge yet.  @kgibm   Please help review.

zhang-ca:
writeInterval with 10 sec or above will always work before the change.  writeInterval to 10 mins worked as always.
Please do not look at the trace of TimeBasedWriteTask time, it will always 10 sec apart before the change. 
With the change, it will show up every second (TimeBasedWriteTask ) in the trace.log.

The actual session write is done by trace of :
DatabaseHashMap                                              < persistSession RETURN true
which is based on writeInterval.

kgibm:
I haven't looked at the code, but is my understanding correct that there is a `TimeBasedWriteTask` which runs every X seconds and checks if a session was updated `<= now() - writeInterval` and calls persistSession?

zhang-ca:
@kgibm That is correct.

kgibm:
@zhang-ca Thanks, that makes sense. @una-tapa Hiroko, so basically the interval of the `TimeBasedWriteTask` controls the "resolution" of `writeInterval`. In other words, with the current 10 second interval, the session updates will be written within `[${TIME_OF_UPDATE} + writeInterval, ${TIME_OF_UPDATE} + writeInterval + 10]`. Reducing the resolution to 1 second makes sense to me and means that a `writeInterval` of 1 second should write updates between `${TIME_OF_UPDATE} + 1 second` and `+ 2 seconds`, depending on when the timer runs. This change seems good but I suggest running it by @gjdeval and @jhanders34 just in case. If there is an equivalent to `TimeBasedWriteTask` in tWAS and it's also 1 second then this change seems to be very low risk.

una-tapa:
Thanks for looking @kgibm.  @zhang-ca 

Found equivalent code in tWAS => WsTBWSessionStorer.java in tsession.shell component.   
In this class,  WAS scheduler AlarmManager is set with  (1 sec x  interval) already. 

   Snippet >  _interval = interval * 1000; // input is  seconds, we need milliseconds 

Looks like tWAS does allow 1 sec interval.  I feel much better.

zhang-ca:
@una-tapa tWAS code is same as Liberty. The call came from com.ibm.ws.session.WsSessionContext.createStorer() : 
    /*
     *  Create the storer.  Overrides method in SessionContext.  In full WAS environment
     *  we can have a time-based write store, a manual storer, or an end-of-service (EOS)
     *  storer.
     */    
     IStorer createStorer(SessionManagerConfig smc, IStore store) {
         IStorer storer = null;
         if (smc.getEnableTimeBasedWrite()) {
             **storer = new WsTBWSessionStorer(store, 10); // 10 seconds is fixed - not configurable**
         } else if (smc.getEnableManualWrite()) {
             storer = new WsManualSessionStorer();
         } else {
             storer = new EOSSessionStorer();
         }
         return storer;
     }

kgibm:
Interesting, so tWAS has always had a "jitter" of up to 10 seconds. There is some risk in reducing to 1 second as it may change the dynamics and push the DB harder (or softer if writes are less bursty) for existing customers. I'd like to get @gjdeval's opinion on this. I like the idea of giving customers the option of as low as 1 second, but maybe it's better to do through a configurable property instead and keep the default of 10 seconds.

zhang-ca:
If Liberty run good with Timer.schedule(TimeBasedWriteTask, 0, 1); 
https://docs.oracle.com/javase/7/docs/api/java/util/Timer.html#schedule(java.util.TimerTask,%20long,%20long)
That is, Timer period of 1 second. We willl retrofit the change in tWAS.

zhang-ca:
@kgibm we can create custom property called "timerPeriod".

gjdeval:
There are two sort of obvious possible performance impacts from this change:
1) Currently a timer pops every 10 secs to check for changes; if we move it to 1 sec then there will be 10X as many timer pops even when no work is being done ... for distributed customers this would pass unnoticed, but zOS customers are very sensitive to activity occurring when the system is idle because they pay by the OPS so they might notice this as an increase in "idle cpu"
2) If the app causes frequent session changes (as suggested by the original customer complaint) then there will be work to do when the timer pops, and it is more expensive to persist the session state more often, compared to doing it on a longer interval or at EOS. In this case at least the customer gets some benefit (better fidelity of session persistence) for the performance cost, so it is up to them whether the trade-off is worth it. Probably this possible performance impact is documented in the KC as part of the description around choosing the persistence behavior? 
There are of course other possible impacts if the session persistence design were suboptimal ...
Maybe we could leave the default timer cycle at 10 secs, and if the customer specifies a smaller replication interval than 10, we can push the timer interval down accordingly? That way most users will see no difference, only those who chose a short-duration replication interval will see a performance impact.

zhang-ca:
@gjdeval Thanks for the suggestion. we will push the timer interval down accordingly.

if (smc.getEnableTimeBasedWrite()) {
            if (smc.getPropertyWriterInterval() < 10)
            {
                storer = new TBWSessionStorer(store, (int)smc.getPropertyWriterInterval()); 
            }
            else
            {
                storer = new TBWSessionStorer(store, 10); // 10 seconds is fixed - not configurable
            }
}

ellen-lau:
#build

ellen-lau:
#build

PR #7805, opened 2019-06-06
validation for app-defined JMS connection factories

njr-11:
Add test coverage for the /validation/ REST endpoint for app-defined JMS connection factories.
Also, fix a ClassCastException when enableSharingForDirectLookups is specified on app-defined resources.


njr-11:
#build

PR #7808, opened 2019-06-06
remove non static public field

benjamin-confino:


benjamin-confino:
#build

benjamin-confino:
In more recent JVMs weld's validation will complain about non-static public fields. So I'm making this public field private application.

PR #7834, opened 2019-06-10
Specify hostname in server debug command

### LINKED ISSUE #7833: Remote debugging does not work with Java 11+
Previously to attach a remote debugger, we could specify JVM arguments such as this:
```
-agentlib:jdwp=transport=dt_socket,address=8787,server=y,suspend=n
```

However, this does not allow connecting to the VM with a remote debugger. To allow this, we need to slightly adjust the `address` property to be this:
```
-agentlib:jdwp=transport=dt_socket,address=*:8787,server=y,suspend=n
```
(Note the extra `*:`)

Unfortunately, the `*:` syntax is not backwards compatible to Java 7/8, but if we specify the hostname instead of `*` then it does work on all JDKs.

aguibert:
Fixes #7833

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -55,14 +55,17 @@
 #              or set to n to startup without waiting for a debugger to attach.
 #              The default value is y. 
 #
+# WLP_DEBUG_REMOTE - Whether to allow remote debugging or not. This can be set
+#              to y to allow remote debugging. The default value is n. 
+#
 # WLP_SKIP_UMASK -  Skip setting the umask value to allow the default value 
-#                   to be used.
+#              to be used.
 #
 # WLP_ZOS_PROCEDURE - The name of the STC procedure to be used to start the 
-#                     server in a z/OS environment.
+#              server in a z/OS environment.
 #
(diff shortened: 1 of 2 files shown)


NottyCode:
I am uneasy about enabling remote debug for everyone because the Java debug port isn't secure. I'd prefer this were opt in so it can be set for those environments where it is known it is needed. I think most people are still doing local debug.

aguibert:
Requested manual personal build here:
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_U82jAI3qEem6Jvl9oDV4Sw

This will run 2 FATs: build.example_fat and com.ibm.ws.kernel.boot_fat

On the following platforms:
> RHEL7_X86_IBMJDK8_64_EBC,UBUNTU18_X86_SUNJDK8_64_EBC,UBUNTU16_X86_OPENJDK8_OPENJ9_64_EBC,UBUNTU16_X86_OPENJDK11_OPENJ9_64_EBC,UBUNTU16_X86_OPENJDK11_HOTSPOT_64_EBC,UBUNTU16_X86_OPENJDK12_OPENJ9_64_EBC,UBUNTU16_X86_OPENJDK12_HOTSPOT_64_EBC,WIN2016_IBMJDK8_64_EBC,WIN2016_OPENJDK11_OPENJ9_64_EBC,WIN2016_OPENJDK11_HOTSPOT_64_EBC

PR #7837, opened 2019-06-10
start of yaml for validation endpoint

njr-11:
Create a basic yaml file that will be used to describe the /validation/ endpoint.  Initially, it will just be the start of a document.  I'll include it in an existing test bucket so that it isn't shipped, but it will still be tracked in git, and possible to write tests against it.


njr-11:
#build

PR #7855, opened 2019-06-11
validation endpoint must avoid logging password to FFDC and trace

njr-11:
In preparing the /validation/ REST point to become product-worthy code, I checked for, and found, a couple of places where it allows the password value to be logged to trace and FFDC logs.  This pull includes updates to prevent the password value from being logged.


njr-11:
#build

njr-11:
Build has 1 test failure, which is a pre-existing issue in an unrelated test bucket.

PR #7858, opened 2019-06-11
include header and query parameters in the yaml for the validation REST endpoint

njr-11:
header params for yaml:
  X-Validation-User
  X-Validation-Password

query params for yaml:
  auth
  authAlias
  loginConfig


njr-11:
#build

PR #7862, opened 2019-06-12
Improve performance of service fingerprint

NottyCode:
It seems that writing the service fingerprint on startup takes around 10ms and most
of that is creating the timestamp that the java.util.Properties.store method writes.
So rewriting to use a DataInputStream/DataOutputStream in the expectation that it will
be faster than using a Properties object.


NottyCode:
#build this please Miss Libby

tjwatson (dismissed):
The change looks fine functionally.  I would like to have confirmation that it improves startup time before merging.

I think we could do additional work to avoid writing the file every startup, possibly multiple times when things have not changed.

NottyCode:
@tjwatson based on our discussion I've made a further change which should remove additional reads and writes without affecting behaviour. I'm waiting for feedback from performance. I'm aiming to merge next week.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/provisioning/ServiceFingerprint.java
```diff
@@ -28,46 +29,49 @@
  */
 public class ServiceFingerprint {
     private static volatile File fingerPrintFile;
-    private static Properties props = new Properties();
+    private static Map<String, String> props = new HashMap<>();
```
tjwatson: Changed this from a thread safe Properties object to a non-thread safe HashMap.  Need to add external synchronization or simply wrap in java.util.Collections.synchronizedMap(Map<K, V>).  Or we need to be convinced this class is only ever accessed from a single thread.

tjwatson: Why is fingerPrintFile volatile but props is not?  Why can we not make props final instead?

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/provisioning/ServiceFingerprint.java
```diff
@@ -28,46 +29,49 @@
...
         boolean result = true; // default to true so we do a clean start if we can't work out if service is applied
 
         fingerPrintFile = new File(workarea, "platform/service.fingerprint");
         if (fingerPrintFile.exists()) {
-            props = new Properties();
-            FileReader reader = null;
-            try {
-                reader = new FileReader(fingerPrintFile);
-                props.load(reader);
+            props = new HashMap<>();
```
tjwatson: Why not clear props here instead of creating a new HashMap?

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/provisioning/ServiceFingerprint.java
```diff
@@ -197,26 +176,21 @@ public static void put(File file) {
...
-                props.store(writer, null);
-            } catch (IOException e) {
-            } finally {
-                if (writer != null) {
-                    try {
-                        writer.close();
-                    } catch (IOException e) {
-                    }
+            }
+            try (DataOutputStream out = new DataOutputStream(new FileOutputStream(fingerPrintFile))) {
```
tjwatson: props.store would use a BufferedWriter under the covers.  I would wrap the FileOutputStream in a BufferedOutputStream

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/provisioning/ServiceFingerprint.java
```diff
@@ -28,46 +29,49 @@
...
         fingerPrintFile = new File(workarea, "platform/service.fingerprint");
         if (fingerPrintFile.exists()) {
-            props = new Properties();
-            FileReader reader = null;
-            try {
-                reader = new FileReader(fingerPrintFile);
-                props.load(reader);
+            props = new HashMap<>();
+
+            try (DataInputStream in = new DataInputStream(new FileInputStream(fingerPrintFile))) {
```
tjwatson: Consider wrapping the FileInputStream in a BufferedInputStream

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/provisioning/ServiceFingerprint.java
```diff
@@ -226,13 +200,16 @@ public static void putInstallDir(String name, File installDir) {
             name = "core";
         }
 
-        internalPut(new File(installDir, "lib/versions/service.fingerprint"));
-        try {
-            props.setProperty(name, installDir.getCanonicalPath());
-        } catch (IOException ioe) {
-            // ignore this. If it happens for core we will do an auto clean next time, otherwise we don't really care.
+        if (!props.containsKey(name)) {
+            internalPut(new File(installDir, "lib/versions/service.fingerprint"));
```
tjwatson: I don't understand the check here.  So if the install dir name (e.g. "core") is already in the prop we loaded from disk we don't need to recheck its service.fingerprint hash to make sure it is not different from when we loaded the props?

PR #7872, opened 2019-06-12
dataSource validation result and example values

njr-11:
Enhance the yaml for dataSource validation to include example values and schema for the validation result.


njr-11:
#build

PR #7892, opened 2019-06-14
Explicitly initialize PlatformMBeanServer prior to JVM self-attach

aguibert:


aguibert:
Cancelling this PR in favor of #7968

PR #7893, opened 2019-06-14
consistent type for error code

njr-11:
Use a consistent type for error codes across exception info that is display by the validation REST endopint, so that we can more straightforwardly represent the exception chain, regardless of exception type, in schema.


njr-11:
#build

Issue #6692, opened 2019-02-26
Implement Hibernate's ExtendedBeanManager


benjamin-confino:

benjamin-confino:
Done

PR #7899, opened 2019-06-16
Jpa extended bean manager

benjamin-confino:
This PR makes it so that there is one IBMHibernateExtendedBeanManager per deployment, rather than one per BeanManager. 

This is possible because IBMHibernateExtendedBeanManager only handles listeners for a couple of CDI lifecycle events. 

In addition this PR uses information available during the cdi lifecycle event method to direct the event to the appropriate IBMHibernateExtendedBeanManager, rather than relying on the connection between a IBMHibernateExtendedBeanManager and an underlying bean manager. Which makes this function earlier in the lifecycle.


benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
This fixed #6692

benjamin-confino:
Fixes #6692

PR #7912, opened 2019-06-17
Make mpjwt wait

c00crane:
Add a check for the Security Service ready message to the server startup of the MP JWT FAT.
We had a case where a test tried to run before the server was fully up.


c00crane:
#link 264151

c00crane:
#build

c00crane:
The mp jwt fat tests 1) all ran in both lite and full, 2) all were successful.
Failures in other projects are unrelated (this was a mp jwt fat only change)

c00crane:
#link 264151

PR #7929, opened 2019-06-18
Check for null properties

acdemyers:
Check for null properties before using them to set SSL parameters.


acdemyers:
#build

acdemyers:
#build

PR #7930, opened 2019-06-18
Restrict REST config/validation endpoints to HTTP GET method

KyleAure:
Restricted both Config/Validator endpoints to GET only.
All local tests and build passed. 
Two tests will need to be reworked in the future: 
- testCustomLoginModuleProperties
- testCustomLoginModuleForJCADataSource

### DIFF
--- a/dev/com.ibm.websphere.rest.handler/src/com/ibm/wsspi/rest/config/ConfigBasedRESTHandler.java
+++ b/dev/com.ibm.websphere.rest.handler/src/com/ibm/wsspi/rest/config/ConfigBasedRESTHandler.java
@@ -140,7 +140,8 @@ public abstract class ConfigBasedRESTHandler implements RESTHandler {
      * @see com.ibm.wsspi.rest.handler.RESTHandler#handleRequest(com.ibm.wsspi.rest.handler.RESTRequest, com.ibm.wsspi.rest.handler.RESTResponse)
      */
     @Override
-    public final void handleRequest(RESTRequest request, RESTResponse response) throws IOException {
+    public void handleRequest(RESTRequest request, RESTResponse response) throws IOException {
+
         String path = request.getPath();
         final boolean trace = TraceComponent.isAnyTracingEnabled();
         if (trace && tc.isEntryEnabled())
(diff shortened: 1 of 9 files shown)


KyleAure:
#build

KyleAure:
#build

dev/com.ibm.websphere.rest.handler/src/com/ibm/wsspi/rest/config/ConfigBasedRESTHandler.java
```diff
@@ -134,13 +134,21 @@ private static String getUID(String configDisplayId, String id) {
...
+     * Common implementation of handleRequest which limits ability to make only GET requests, locates configured instances
+     * matching the request, delegating each to handleSingleInstance, and, after all of which have completed, invokes
+     * createResponse.
      *
      * @see com.ibm.wsspi.rest.handler.RESTHandler#handleRequest(com.ibm.wsspi.rest.handler.RESTRequest, com.ibm.wsspi.rest.handler.RESTResponse)
      */
     @Override
     public final void handleRequest(RESTRequest request, RESTResponse response) throws IOException {
+
+        if (!"GET".equals(request.getMethod())) {
```
njr-11: We shouldn't put the limitation here because it will limit all REST API that is based on server config to only allowing GET.  Instead, it should be place on the specific REST endpoints (config and validation) that we are trying to limit. And yes, I realize that config & validation are current the only config-based REST endpoints, but the ConfigBasedRESTHandler is intended for more general use in the future, and this would limit it.

dev/fattest.simplicity/src/componenttest/topology/utils/HttpsRequest.java
```diff
@@ -182,6 +183,8 @@ public void checkServerTrusted(X509Certificate[] certs, String authType) {}
 
             if (json != null) {
                 con.setRequestProperty("Content-Type", "application/json");
+                //TODO Look for alternatives to output stream for writing JSON data to message body.
+                //This line will change request method to POST
```
aguibert: since we can break backwards compatibility in the test framework, we could save others the confusion and throw an `IllegalStateException` or something if they try to specify `json` and `method("GET")` at the same time

PR #7945, opened 2019-06-19
Remove testClassLoaderCache

tevans78:
#build 
testClassLoaderCache is unreliable because the logging and tracing code keeps hold of a SoftReference to the ClassLoader.

testMultiApplication appears to be OK so leaving that in place for now.


Azquelt:
I didn't think this was the same test that we had the logging and tracing issue with. Nothing should keep a reference to the classloader after the app is restarted.

tevans78:
This isn't the right solution, closed.

Issue #7426, opened 2019-05-06
com.ibm.wsspi.kernel.embeddable.Server stop() does not honor its arguments


kaczyns:
The javadoc for `Server.java` says that `stop(String... args)` takes argument as the `server` script does.  However the `--force` argument does not cause the server to force stop.

kaczyns:
In https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/wsspi/kernel/embeddable/Server.java the stop() method takes arguments. According to the javadoc, these are command line arguments for stop. However the implementation ignores the arguments. This makes it impossible to do a stop --force.

In EmbeddedServerImpl.java there is an embedded class StopOperation that throws away the arguments in its constructor. It should save the arguments, or at least parse them for the --force flag and set a boolean indicating force was requested (at some point we'll have to parse the arguments, could do it here or could pass them thru as strings until later. Lets just do it now. If we get unknown arguments just ignore them so we don't break anyone who is currently passing unknown arguments). When the stop operation calls server.shutdown() we'll have to add the the new boolean to the call. This method then calls thru to kernelBootstrap.shutdown().

The KernelBoostrap class should add a version of shutdown() that takes a single boolean representing the force option. This calls thru to LauncherDelegate.shutdown(). We should add another version of shutdown() to LauncherDelegate that takes the boolean for force. The implementation of LauncherDelegate is in com.ibm.ws.kernel.launch.internal.LauncherDelegateImpl. It should add the new shutdown method. The shutdown method calls thru to FrameworkManager.shutdownFramework(). Fortunately there is already a method on the FrameworkManager that takes a boolean argument for force: FrameworkManager.shutdownCommand(boolean force). We can call this method instead.

Note that there may be other implementations of LauncherDelegate that need to change. This is not an SPI interface so there should not be external implementations.

PR #7952, opened 2019-06-19
Issue #7426: Make server stop take arguments

mmerali96:
#7426

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/EmbeddedServerImpl.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/EmbeddedServerImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2016 IBM Corporation and others.
+ * Copyright (c) 2012, 2016, 2019 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -82,9 +83,9 @@ public class EmbeddedServerImpl implements Server {
      * enhancements need new constructors.</p>
      *
      * @param serverName ServerName: defaultServer will be used if this is null
-     * @param userDir WLP_USER_DIR equivalent, may be null
-     * @param outputDir WLP_OUTPUT_DIR equivalent, may be null
-     * @param listener ServerEventListener that should receive notifications of Server lifecycle changes, may be null.
+     * @param userDir    WLP_USER_DIR equivalent, may be null
+     * @param outputDir  WLP_OUTPUT_DIR equivalent, may be null
+     * @param listener   ServerEventListener that should receive notifications of Server lifecycle changes, may be null.
      *
      * @see {@link ServerEventListener}
      */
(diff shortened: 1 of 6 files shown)


kaczyns:
#build

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/EmbeddedServerImpl.java
```diff
@@ -380,12 +381,32 @@ public Result call() {
...
+
         /**
-         * No arguments yet: someday there may be, we're just being flexible.
+         * Handles argument parsing for stop command.
          *
          * @param arguments
          */
-        StopOperation(String... arguments) {}
+        StopOperation(String... arguments) {
+            List<String> cmdArgs = new ArrayList<String>(Arrays.asList(arguments));
```
kaczyns: There's nothing wrong with what you did here, I'm just curious why you created an ArrayList when the `Arrays.asList()` is going to give you back a `List`.  I didn't see you using any specific methods on `ArrayList`.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/EmbeddedServerImpl.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2016 IBM Corporation and others.
+ * Copyright (c) 2012, 2016, 2019 IBM Corporation and others.
```
kaczyns: Thanks for remembering to update the copyright!  For changes to parts, we keep the first and the last copyright date.  So this should be `2012, 2019` instead of `2012, 2016, 2019`.  This applies to all of the copyrights you changed here.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/EmbeddedServerImpl.java
```diff
@@ -380,12 +381,32 @@ public Result call() {
...
-        StopOperation(String... arguments) {}
+        StopOperation(String... arguments) {
+            List<String> cmdArgs = new ArrayList<String>(Arrays.asList(arguments));
+
+            if (0 < cmdArgs.size()) {
+                Iterator<String> i = cmdArgs.listIterator();
+                while (i.hasNext()) {
+                    String arg = i.next();
+
+                    if (arg.equalsIgnoreCase("--force")) {
```
kaczyns: I was not sure how `Arrays.asList()` worked and so I was playing with it.  I noticed that I could add null strings to the list.  I think we need to check `if ((arg != null) && (arg.equalsIgnoreCase("--force"))) {` here instead, to make sure that `arg` is not null.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/KernelBootstrap.java
```diff
@@ -296,6 +296,27 @@ public ReturnCode shutdown() throws InterruptedException {
...
     }
 
+    /**
+     * Force stop the server. This emulates the path that the server stop action takes:
+     * it calls shutdown on the LauncherDelegate, and then waits until it can obtain
+     * the server lock to ensure the server has stopped.
+     *
+     * @throws InterruptedException
+     */
+    public ReturnCode shutdown(boolean force) throws InterruptedException {
```
kaczyns: Probably should have the no-arg shutdown method call this one, with a false argument. Something like:
```java
public ReturnCode shutdown() throws InterruptedException {
  return shutdown(false);
}
```

dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/LauncherDelegateImpl.java
```diff
@@ -139,6 +139,24 @@ public boolean shutdown() throws InterruptedException {
         return true;
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * @throws InterruptedException
+     */
+    @Override
+    public boolean shutdown(boolean force) throws InterruptedException {
```
kaczyns: Same comment here about having the no-arg version call this version with `false`.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/EmbeddedServerImpl.java
mmerali96: I think I based my arguments parsing off server start which uses an ArrayList. That's a good point though, I didn't realize that I am not using specific ArrayList methods. Going to get rid of the ArrayList cause we don't need it.

mmerali96:
#build

PR #7960, opened 2019-06-19
Set mark to end of log between server config update / restore

olendvcook:
Set mark to end of log between server config update / restore


olendvcook:
#build

PR #7964, opened 2019-06-19
Modification needed in server.xml due to lack of java.net.URLPermission class for Java 7

BerksanAtes:
Java 7 throws "java.lang.ClassNotFoundException[java.net.URLPermission]" due to java.net.URLPermission defined in server.xml. Using java7_server.xml in which java.net.URLPermission settings are removed solved this test run problem for com.ibm.ws.jaxws.ejb_fat bundle


BerksanAtes:
#build

PR #7967, opened 2019-06-19
experiment with unicode chars in parameters

njr-11:
Add tests for including unicode characters in query parameters and header parameters.
Query parameters can take advantage of URL encoding and are automatically decoded for us.
However, with header parameters, we would need to do the decoding within our code.  I've included a hidden (for now) parameter that can switch on/off the ability to decode header parameters in order to experiment with that approach, which should be fine while this feature is still in beta.


njr-11:
#build

PR #7968, opened 2019-06-19
Explicitly initialize PlatformMBeanServer prior to JVM self-attach

aguibert:


aguibert:
#build

aguibert:
requested an additional build with a more focused set of FATs but a large sampling of JDK/OS/VM levels:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_W7o00JLVEemx0IgRKfGg0w

aguibert:
builds completed OK, except for the Mac build which has not started yet due to lack of hardware available to run the build it seems.

Proceeding with merging

Azquelt:
This appears to have broken the Java 2 Security build again (went from 10 errors in the previous build to 1700 errors in the build this PR was delivered). From this builds in this PR, it doesn't look like it was tested with J2S enabled.

Azquelt:
Looks like here: https://github.com/OpenLiberty/open-liberty/pull/7968/files#diff-f5e0d616c085fb479138de802a3e94cdR79 we're loading the Java 8 class whereas the previous code chooses a different class for Java 9+. I assume those checks were removed when you started using the new public APIs for Java 9+ and were not put back in when we discovered a few edge cases where the new code path does not work.

So I think this needs backed out. Could you run a pbuild with j2sec enabled before redelivering?

dev/com.ibm.ws.jmx.connector.local/src/com/ibm/ws/jmx/connector/local/LocalConnectorActivator.java
```diff
@@ -62,26 +62,50 @@ public String run() {
...
-                            Class<?> clazz2 = JavaInfo.majorVersion() < 9 
-								? Class.forName("sun.misc.VMSupport", true, systemClassLoader) 
-								: Class.forName("jdk.internal.vm.VMSupport", true, systemClassLoader);
-                            Method m2 = clazz2.getMethod("getAgentProperties");
-                            Properties props = (Properties) m2.invoke(null);
-                            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                                Tr.debug(tc, "Attempting to retrieve the connector address from agent properties.");
+                        if (JavaInfo.majorVersion() < 9 ||
+                            (JavaInfo.majorVersion() >= 9 && !Boolean.getBoolean("jdk.attach.allowAttachSelf") ||
+                             System.getSecurityManager() != null)) {
```
tevans78: @aguibert we think the problem is here ... if `System.getSecurityManager() != null` then it should use `jdk.internal.agent.Agent`??

Azquelt: Kinda. We think that if `JavaInfo.majorVersion() >= 9` and this whole condition is also true, then on line 79 we should be using `jdk.internal.agent.Agent` and on line 84 we should be using `jdk.internal.vm.VMSupport`, as the old code did.

aguibert: yep, @Azquelt is spot on -- my changes were preserving existing behavior for the Java 7/8 code path but not the Java 11+ code path

PR #7969, opened 2019-06-20
Create common bundle for mpMetrics so that they share metatype files

### LINKED ISSUE #7948: Avoid duplicating metatype between mpMetrics-1.1 and mpMetrics-2.0
Having separate metatype may cause issues with the server.xsd, which is used by WDT, by the admin server config tool, and by our doc generation.

Channyboy:
fixes #7948 
#build


tevans78:
Failure in the PB looks like the same as this
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=264237

Channyboy:
Fats indicate that a liberty message does not appear. However, manual inspection shows that the message is present. This may be an error with the FAT (i.e timing).

Channyboy:
#build

Channyboy:
The `MetricsAuthenticationTest` is failing once again in the second build with the same issue.
The test is expecting an `8010` port, but `8011` port is used instead. 
It is related to the RTC item that @tevans78 has mentioned.

Another FAT failure has occured in the second build, but has not relation to the changes made here for metrics. There also seems to be existing RTC items :
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=264062

The fix should be good.

dev/com.ibm.ws.microprofile.metrics.common/bnd.bnd
```diff
@@ -0,0 +1,37 @@
+#*******************************************************************************
+# Copyright (c) 2017, 2018 IBM Corporation and others.
```
fmhwong: 2019

PR #7973, opened 2019-06-20
possible fixed code for issue#7971

### LINKED ISSUE #7971: performance concern of com.ibm.ws.security.jca.internal.AuthDataServiceImpl
@una-tapa suggested me to open an issue cc: @kgibm

There is a performance concern of com.ibm.ws.security.jca.internal.AuthDataServiceImpl, especially, in Liberty + MQ system where container authentication is used.
This might be applicable to other types of ESI, but let me explain the concern by using the MQ case.

When a secure servlet(ServletA) request a connection, a new connection is created. When the connection manager allocate a connection, it associates a subject with the connection.
And the subject contains the following information. The credential is created by using the container auth information.

Subject:
    Principal:  login user of ServletA
    Credential: container-auth info

Then, if other login user requests a connection, the connection matching logic in JCA layer takes the entire subject into consideration. So the existing connection can't be reused for the other login user even if it's in the free pool.
So frequent connection creation and destroy will happen in the pool if many login users have a chance to request connections. This is not good from performance perspective.

The credential generated by the container-auth is sufficient for MQ RAR to connect to MQ. So I was wondering why Liberty is adding the caller principal into the subject.
When AuthDataServiceImpl create a subject using AuthData, it adds the caller principal in obtainSubject method. It seems this behavior was introduced by an old APAR PI61592 which was opened for a customer's specific resource adapter.
But I'm feeling that the Liberty is losing a performance benefit for general use-cases by corresponding to a specific use-case.

e30532:
fixes #7971

### DIFF
--- a/dev/com.ibm.ws.security.jca/src/com/ibm/ws/security/jca/internal/AuthDataServiceImpl.java
+++ b/dev/com.ibm.ws.security.jca/src/com/ibm/ws/security/jca/internal/AuthDataServiceImpl.java
@@ -148,7 +148,6 @@ public class AuthDataServiceImpl implements AuthDataService {
 
     private Subject obtainSubject(ManagedConnectionFactory managedConnectionFactory, AuthData authData) {
         Subject subject = createSubject(managedConnectionFactory, authData);
-        addInvocationSubjectPrincipal(subject);
         optimize(subject);
         return subject;
     }


e30532:
@una-tapa would you please review this?

teddyjtorres:
This code cannot be removed since the principal is needed by resource adapters that can make use of it. Removing this code will cause a regression and potential outages.

una-tapa:
Per slack conversation, working to add switch to this.

PR #7977, opened 2019-06-20
Use correct system property for FAT server HTTP port

andymc12:
The GraphQL FAT tests currently look for the wrong system property when determining the server's HTTP port - it looks for "bvt.prop.HTTP" when it should look for "bvt.prop.HTTP_default" - if unspecified, it defaults to 8010 (which is the default test port) - that's why these tests pass most of the time.  But when a port other than 8010 is used, the tests fail.  This change uses the correct system property for determining the server port.


andymc12:
#build

andymc12:
#build

PR #7983, opened 2019-06-20
initial yaml for config endpoint, plus improvements to yaml for login config properties

njr-11:
Create initial version of the schema for the ibm/api/config REST endpoint.
Move the schema for login config properties of the validation REST endpoint to be a header parameter instead of in the body so that the request doesn't auto-convert from GET to POST.
Add a parameter to let the user specify whether or not they URL-encoded their header parameters (which is needed to supply non-ASCII).


njr-11:
#build

Issue #7974, opened 2019-06-20
Port Default Bindings Tests to Open Liberty


olendvcook:
Port default bindings tests to Open Liberty

PR #7985, opened 2019-06-20
Create ejbcontainer.bindings_fat and port default bindings test

olendvcook:
Created a new project for bindings tests and ported default bindings test.  Failing tests currently aren't ran because the TestClass code that adds the servlets is commented out.

Added Java:global lookup tests as a sanity check that the application is correctly installed and current bindings work.


olendvcook:
#build

dev/com.ibm.ws.ejbcontainer.bindings_fat/test-applications/EJB3DefBndWeb.war/src/com/ibm/ws/ejbcontainer/bindings/defbnd/web/DefaultJavaColonBindingsServlet.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+@SuppressWarnings("serial")
+@WebServlet("/DefaultJavaColonBindingsServlet")
+public class DefaultJavaColonBindingsServlet extends FATServlet {
+
+    /**
+     * lookup and call local business interface using java: global.
+     */
+    @Test
+    public void testJavaColonLocalBI() throws Exception {
```
brideck: I know you're just porting a legacy test here, but we probably shouldn't be using remote lookups, etc. for these two local tests now that this is going to open source.  Do we have a lookup method that doesn't do a narrow that we could use?  The assertion text should also be updated to reflect the correct interface names.

olendvcook: @brideck I've changed it, thanks for spotting.

olendvcook:
#build

olendvcook:
for #7974

PR #7988, opened 2019-06-21
add yaml for root config element and filtering by generic query params

njr-11:
Add schema for /ibm/api/config
Update schema to allow for generic query parameters to be supplied for filtering, both on this endpoint, as well as the /ibm/api/config/{elementName} endpoint

### DIFF
--- a/dev/com.ibm.ws.rest.handler.validator_fat/test-applications/testOpenAPIApp/resources/META-INF/openapi.yaml
+++ b/dev/com.ibm.ws.rest.handler.validator_fat/test-applications/testOpenAPIApp/resources/META-INF/openapi.yaml
@@ -16,12 +16,29 @@ info:
 servers:
   - url: https://127.0.0.1:8020/ibm/api
 paths:
+  /config/:
+    get:
+      tags:
+      - Config
+      summary: "Shows configuration of all elements"
+      description: "Retrieves configuration detail for instances of all configuration element types."
+      parameters:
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.rest.handler.validator_fat/test-applications/testOpenAPIApp/resources/META-INF/openapi.yaml
```diff
@@ -16,12 +16,29 @@ info:
...
+      responses:
+        200:
+          description: "Configuration info retrieved"
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/config.result"
   /config/{elementName}:
```
nmittles: I know it's not related to these changes, but would `{elementType}` be better than `{elementName}`?

njr-11: I'm not really sure which way to go with the terminology.
I could see someone saying that the type of configuration element that you put in server config is `dataSource` or `connectionManager`.  I could also see someone saying that the type of configuration that you put in server config is a data source or a connection manager, and that the name of these configuration elements that you put into server config is `dataSource` and `connectionManager`.  I suppose it would be best to go with whichever way the majority see it, but I'm unsure what that is.

njr-11:
#build

PR #7992, opened 2019-06-21
update loginConfigProps impl per the yaml and re-enable tests

njr-11:
Update REST endpoint to read login config properties from the header instead of the body, and then re-enable tests for login config properties.


njr-11:
#build

PR #8009, opened 2019-06-24
add basic auth and error message

njr-11:
Add option for basic authentication to yaml.
Add test case for URL encoded characters in header parameters.
Also add an error message for when the delimiter is missing.


njr-11:
#build

njr-11:
#build

PR #8011, opened 2019-06-24
remove erroneous override of library id

njr-11:
The metatype for the library element overrides the id attribute and defines a default value for the id, being a variable reference to the service.pid.  However, it isn't valid to use variables within id values. After consulting with Brent Daniel, I've opened this pull to remove this invalid default and submit a build against it to help confirm that nothing will break if we fix it.

### DIFF
--- a/dev/com.ibm.ws.classloading/resources/OSGI-INF/metatype/library.xml
+++ b/dev/com.ibm.ws.classloading/resources/OSGI-INF/metatype/library.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-    Copyright (c) 2011 IBM Corporation and others.
+    Copyright (c) 2011,2019 IBM Corporation and others.
     All rights reserved. This program and the accompanying materials
     are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


njr-11:
#build

njr-11:
Hundreds of tests failed, demonstrating that code is relying on the erroneously-specified id.  I'm closing out this pull (it cannot be delivered as is) and will document the findings in the corresponding issue.

njr-11:
Using the pull to try out another test, which involves fixing the OSGi filter syntax error which was due to unescaped characters
#build

njr-11:
#build

njr-11:
@brenthdaniel - this pull contains some experimentation with removing the erroneous default id for library containing the variable.  I initially thought it wouldn't work given the number of test failures, but after further experimentation found that all of the tests pass after fixing a couple of OSGi filters that were built without escaping the () characters.  After including that, it got a green build with all tests passing.  You're the expert in this area, and I'll leave it to you where you believe this change is safe.  I'm out all of next week, however, if you think the change is valid, just add yourself as a reviewer and merge it.

njr-11:
#build

PR #8020, opened 2019-06-25
 Porting error paths test bucket to open liberty

KyleAure:
Squashed Commit History
Original port
Fixed creation of apps/db location
Local build and tests successful

Waiting to build/pull until bucket is wrapped up in CL


dev/com.ibm.ws.concurrent.persistent_fat_errorpaths/fat/src/com/ibm/ws/concurrent/persistent/fat/errorpaths/PersistentExecutorErrorPathsTest.java
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * After completing all tests, stop the server.
+     *
+     * @throws Exception
+     */
+    @AfterClass
+    public static void tearDown() throws Exception {
+        if (server != null && server.isStarted())
+            server.stopServer(LibertyServer.DISABLE_FAILURE_CHECKING);
```
aguibert: I don't think it's necessary to have special code or a special value for disabling error checking. Instead, we could just pass in the regex `.*` which will match everything without needing special handling

KyleAure:
#build

PR #8021, opened 2019-06-25
WIP: Default trustStoreType to PKCS12

andymc12:
Test bug in the MP Rest Client 1.3 TCK - breaks with IBM JDK 8 because the test uses PKCS12 trust store type for the embedded Jetty test server for the SSL tests.  IBM JDK 8 assumes a default of JKS trust store type.  Oracle JDK 8 and all JDK versions >8 either default to PKCS12 or can detect it and roll with it.  This is a workaround so that Liberty passes the TCK with IBM JDK 8.


andymc12:
Not necessary - the fix should come from the TCK test case - see PR https://github.com/eclipse/microprofile-rest-client/pull/205

PR #7990, opened 2019-06-21
Kafka connector for reactive messaging

Azquelt:
Add a kafka connector for microprofile reactive messaging.

Fixes #7783 
Fixes #7782 
Fixes #7828 
Fixes #7740 
Fixes #4086

PR #8026, opened 2019-06-25
Update kafka client test dependency

Azquelt:
Update our kafka client test dependency to 2.3.0. This dependency is not shipped.

Also remove the project that was wrapping the kafka client up into a bundle as it was not used anywhere and isn't needed.

Only the last two commits are part of this PR. The rest will be removed once #7990 is merged.


Azquelt:
#build

Azquelt:
Rebased on integration

PR #8030, opened 2019-06-25
standardize and enforce parameters/results for validator implementations

njr-11:
Define constants to standardize parameter names and result keys across validator implementations.
Add enforcement requiring only recognized parameter names to be used, along with the ability for implementations to extend or replace the list of valid parameters.


njr-11:
#build

Issue #7394, opened 2019-05-01
Use MP Context Propagation for JAX-RS Reactive Client


andymc12:
We need to use the CompletionStage/CompleteableFuture from MP Context Propagation in the JAX-RS Reactive Client so that the CompletionStages will allow context propagation - and will retain the correct executor service beyond the initial stage(s).

This will require use of the internal MP Context Propagation feature, and then enabling some currently disabled tests.  

We may need to do the same thing for MP Rest Client async methods.

jkoehler22:
PR: https://github.com/OpenLiberty/open-liberty/pull/8040

jkoehler22:
Nate - Can you please review com.ibm.websphere.appserver.features/visibility/private/com.ibm.websphere.appserver.concurrent.mp-1.0.feature.  Working with AndyM, he had me add features and bundles just like com.ibm.websphere.appserver.concurrent-1.0.feature....thanks!

PR #8040, opened 2019-06-25
Fix for Issue #7394: Use MP Context Propagation for JAX-RS Reactive Client

jkoehler22:


jkoehler22:
#build

jkoehler22:
Nate - Can you please review com.ibm.websphere.appserver.features/visibility/private/com.ibm.websphere.appserver.concurrent.mp-1.0.feature. Working with AndyM, he had me add features and bundles just like com.ibm.websphere.appserver.concurrent-1.0.feature....thanks!

dev/com.ibm.ws.org.apache.cxf.cxf.rt.rs.client.3.2/src/org/apache/cxf/jaxrs/client/CompletionStageRxInvokerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    public CompletionStage<Response> method(String name, Entity<?> entity) {
+        return method(name, entity, Response.class);
+    }
+
+    @Override
+    public <T> CompletionStage<T> method(String name, Entity<?> entity, Class<T> responseType) {
+      //Liberty code change start
+        if (ex == null) {
+         // Run on Liberty executor thread and do not propagate the thread context
+            return (ManagedExecutor.builder().propagated(ThreadContext.NONE).build()).supplyAsync(() -> wc.sync().method(name, entity, responseType));
```
njr-11: This will leak ManagedExecutor instances that will accumulate without ever being shut down, eventually causing out-of-memory.

It's also important to note that by using a ManagedExecutor with context propagation disabled, you still get context clearing.  I'm not sure if that is what you wanted here.  If so, then a better way to do this would be to construct a single, shared ManagedExecutor instance with context propagation disabled and reuse it.
Alternatively, if the interference with thread context is unwanted on this path, then the following would be the most efficient way to get what you are looking for:
com.ibm.ws.concurrent.mp.ManagedCompletableFuture.supplyAsync(Supplier<U>, Executor)
I'm not sure if the above method will be accessible to your code, but I'm certainly open to exporting an a helper interface or something that will give you access to it.

dev/com.ibm.ws.jaxrs.2.1_fat_executor/test-applications/jaxrsapp/src/web/jaxrstest/JAXRSExecutorTestServlet.java
```diff
@@ -170,7 +171,10 @@ public void testCompletionStageRxInvokerSynchronousFunctionSwitchManagedExecutor
                 results.add(result);
             if (error != null)
                 results.add(error);
-            results.add(Thread.currentThread().getName());
+            String threadName = Thread.currentThread().getName();
+            results.add(threadName);
+            _log.info(m + " currentThread=" + threadName);
+            new Exception("handleAsync is running on " + threadName).printStackTrace(System.out);
```
njr-11: Was this code that dumps the stack meant to remain in the test code?   It's fine if it is.  I just wanted to draw your attention to it in case you meant to remove it before delivery.

dev/com.ibm.ws.org.apache.cxf.cxf.rt.rs.client.3.2/bnd.overrides
```diff
@@ -13,6 +13,7 @@ Import-Package: \
   !org.springframework.*,\
   !javax.cache.*,\
   !org.osgi.service.blueprint.*,\
+  com.ibm.ws.concurrent.mp.*;resolution:=optional,\
```
njr-11: I'm confused by this line because com.ibm.ws.concurrent.mp is a private package (not exported) and so will never be available for import.  This won't fail, because the import is optional.  The line just doesn't seem to serve any purpose.

dev/com.ibm.websphere.appserver.features/visibility/private/com.ibm.websphere.appserver.concurrent.mp-1.0.feature
```diff
@@ -2,8 +2,17 @@
 symbolicName=com.ibm.websphere.appserver.concurrent.mp-1.0
 visibility=private
 singleton=true
+-features=com.ibm.websphere.appserver.containerServices-1.0, \
+ com.ibm.websphere.appserver.appLifecycle-1.0, \
+ com.ibm.websphere.appserver.concurrencyPolicy-1.0, \
+ com.ibm.websphere.appserver.concurrent.mp-0.0.0.noImpl; ibm.tolerates:=1.0, \
```
njr-11: Some of the changes to this file are very confusing.  Do you have the reasoning behind why they are being made.
On this particular line, it seems like the com.ibm.websphere.appserver.concurrent.mp-1.0 feature is including an older version of itself and then tolerating the newest version of itself?  Are all of these changes needed?  If not, can the ones where are not needed be omitted?

dev/com.ibm.ws.org.apache.cxf.cxf.rt.rs.client.3.2/src/org/apache/cxf/jaxrs/client/CompletionStageRxInvokerImpl.java
jkoehler22: Would this disable context propagation?
defaultManagedExecutor = ManagedExecutor.builder().propagated().build();

dev/com.ibm.websphere.appserver.features/visibility/private/com.ibm.websphere.appserver.concurrent.mp-1.0.feature
jkoehler22: Looks like I only need com.ibm.websphere.appserver.contextService-1.0.  If I remove com.ibm.websphere.appserver.contextService-1.0 I get:  E **CWWKE0702E**: Could not resolve module: com.ibm.ws.concurrent.mp.1.0 [120]

Thanks for all of the great comments!

dev/com.ibm.ws.org.apache.cxf.cxf.rt.rs.client.3.2/src/org/apache/cxf/jaxrs/client/CompletionStageRxInvokerImpl.java
njr-11: Yes, `.propagated()` and `.propagated(ThreadContext.NONE)` are identical.  The `ThreadContext.NONE` constant is just there for users who prefer a more explicit way of writing it.  With all context propagation disabled on this executor, it will resort to clearing all thread context types for the duration of the completion stage action.

jkoehler22:
#build

dev/com.ibm.ws.org.apache.cxf.cxf.rt.rs.client.3.2/bnd.overrides
```diff
@@ -13,7 +13,7 @@ Import-Package: \
   !org.springframework.*,\
   !javax.cache.*,\
   !org.osgi.service.blueprint.*,\
-  *
+   *
```
andymc12: No need to add a space here - probably Eclipse did it to "help" you when you were making another change, but now that there are no other changes, it would be better to remove this change from the commit to keep the history cleaner.

dev/com.ibm.ws.org.apache.cxf.cxf.rt.rs.client.3.2/src/org/apache/cxf/jaxrs/client/CompletionStageRxInvokerImpl.java
```diff
@@ -29,19 +29,24 @@
...
+    private boolean isExecutorNull = false;
+    private ExecutorService defaultManagedExecutor = null;
     CompletionStageRxInvokerImpl(WebClient wc, ExecutorService ex) {
         this.ex = ex;        
         if (this.ex instanceof ManagedExecutor) {            
             isManagedExecutor = true;
         }
+        if (this.ex == null) {
+            isExecutorNull = true;
+            defaultManagedExecutor = ManagedExecutor.builder().propagated().build();
```
njr-11: This is definitely an improvement on creating less ManagedExecutor instances, but I suspect there is still a memory leak here, albeit a lesser one.  Am I right in assuming that a new instance of CompletionStageRxInvokerImpl would be created with each clientbuilder.build()?  If so, every time that is invoked, then once the user is done with the client, it will leak one ManagedExecutor.  These could build up over time.  It looks like javax.ws.rs.client.Client has a `close` method, where you could shut down the executor instance if one is created for it, and that would prevent the memory leak.  However, that would have some other implications, which might or might not be desirable - specifically that it will attempt to cancel/interrupt ongoing async operations on the executor and prevent further async operations from being requested.  I should probably chat with Andy about what sort of behavior he wants to see for the default (no executor specified) path, both with respect to this as well as with regard to context clearing (mentioned in other comments).

dev/com.ibm.ws.org.apache.cxf.cxf.rt.rs.client.3.2/src/org/apache/cxf/jaxrs/client/CompletionStageRxInvokerImpl.java
```diff
@@ -0,0 +1,215 @@
...
+        if (isExecutorNull) {
+         // Run on Liberty executor thread and do not propagate the thread context
+            return ((ManagedExecutor) defaultManagedExecutor).supplyAsync(() -> wc.sync().method(name, responseType));
+        }      
+        if (isManagedExecutor) {
+            return ((ManagedExecutor) ex).supplyAsync(() -> wc.sync().method(name, responseType));
+        }
+      //Liberty code change end
+        return CompletableFuture.supplyAsync(() -> wc.sync().method(name, responseType), ex);
+    }
```
andymc12: It looks like all of the `method` methods are essentially doing the same thing.  You could probably save some code by adding something like this:
```
private <T> CompletionStage<T> supplyAsync(Supplier<T> supplier) {
        if (isExecutorNull) {
          // Run on Liberty executor thread and do not propagate the thread context
             return ((ManagedExecutor) defaultManagedExecutor).supplyAsync(supplier);
         }      
         if (isManagedExecutor) {
             return ((ManagedExecutor) ex).supplyAsync(supplier);
         }
         return CompletableFuture.supplyAsync(supplier, ex);
}
```

then all of the `method` methods could just call:
`return supplyAsync( () -> wc.sync().method(name, responseType) );` (with the appropriate parameters for each method.

dev/com.ibm.ws.jaxrs.2.1_fat_executor/bnd.bnd
```diff
@@ -26,4 +26,7 @@ javac.target: 1.8
   com.ibm.websphere.javaee.jsonb.1.0,\
   com.ibm.websphere.javaee.jsonp.1.1,\
   com.ibm.websphere.javaee.servlet.3.1;version=latest,\
-  com.ibm.ws.cdi.interfaces;version=latest
+  com.ibm.ws.cdi.interfaces;version=latest,\
+  com.ibm.websphere.org.eclipse.microprofile.contextpropagation.1.0;version=latest,\
+  com.ibm.ws.concurrent.mp.1.0;version=latest,\
+  com.ibm.websphere.javaee.concurrent.1.0;version=latest
```
andymc12: I think only `com.ibm.websphere.javaee.concurrent.1.0;version=latest` is necessary here.  The test cases shouldn't need to depend on MP context propagation or the concurrent.mp project.  If they do, then it sounds like user apps might also need these dependencies, which would be a breaking change.

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
Do a build with ga added to mp.feature

#build

jkoehler22:
Previous build failed with:
Caused by: com.ibm.elastic.build.cloud.api.requirement.exceptions.TemporaryProcessingException

#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
#build

jkoehler22:
All test passed.  Only failure is in the grade.log.  I will ignore this failure and merge.

PR #8054, opened 2019-06-26
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

I know this code is intended for Jython which is currently only supports Python 2 but we can all hope that someday Jython will support Python 3.


cclauss:
I tested my changes by running:
* python2 -m flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
* python3 -m flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
before and after making my changes to ensure that the syntax error on Python 3 was fixed without causing any new breakage,

I am an IBMer and I consider the addition of these four characters to be __CLA Trivial__

PR #8060, opened 2019-06-26
enable JAX-RS to create completable futures backed by any executor

njr-11:
Export an interface that will allow JAX-RS to create a completable future that uses any executor service as its default asynchronous execution facility.


njr-11:
#build

njr-11:
Build machine ran out of space and aborted the first attempt. Retrying the build,
#build

PR #8081, opened 2019-06-28
escape special characters from element name when used in OSGi filter

njr-11:
Special characters should be escaped when supplying the provided values to an OSGi filter.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
build has 5 test failures which are preexisting issues in unrelated test buckets.

PR #8087, opened 2019-06-28
include application/module/component in validation result for app-defined resources

njr-11:
App-defined data sources have JNDI names that can be scoped to the application, module and component, in which case the JNDI name itself isn't unique and must be paired with this other information.  We already include application/module/component in the /config/ output for app-defined resources, and should do the same for the /validation/ REST endpoint.


njr-11:
#build

PR #8089, opened 2019-06-28
WIP: Server startup notifications

NottyCode:
We need a way to provide sync (and async) notification of Liberty startup. There is an internal async, but there are people building user features that would like a public way to do this. Also need a sync way.

The sync approach is being done using a listener service registered in the SR. This is called before the message is output. The intent is this will be used by the channel framework to start ports on startup.

The async approach is just creating an SPI version of the existing internal ServerStarted interface. Essentially this means that you can watch the service registry and react based on the service being registered.


NottyCode:
#build this please Libby

PR #8095, opened 2019-06-28
Give <properties.postgresql> default values for serverName and portNumber

aguibert:


aguibert:
#build

dev/com.ibm.ws.jdbc/src/com/ibm/ws/jdbc/internal/JDBCDriverService.java
```diff
@@ -1053,6 +1055,28 @@ private void shutdownDerbyEmbedded() {
...
+        Arrays.sort(props, new Comparator<PropertyDescriptor>() {
+            @Override
+            public int compare(PropertyDescriptor p1, PropertyDescriptor p2) {
+                if (p1.getName().equalsIgnoreCase("url")) {
+                    if (p2.getName().equalsIgnoreCase("url"))
+                        return p1.getName().compareTo(p2.getName()); 
+                    else
+                        return 1;
+                }
+                if (p2.getName().equalsIgnoreCase("url"))
```
njr-11: If we are only doing this for PostgreSQL, do we really need the equalsIgnoreCase?  I imagine its more costly than normal equals.

aguibert: other JDBC drivers may benefit from this as well, so I kept it generic. I don't think the performance difference will be significant.

aguibert:
Cancelling this PR because the PostgreSQL JDBC driver has built-in defaults for serverName and portNumber, so it is not necessary for us to set default values in our metatype.

PR #8100, opened 2019-07-01
Fix MP Config classloader cache test

Azquelt:
Several fixes here:

1. Check the return value of `LibertyServer` methods which return `false` on error rather than throwing an exception
1. When deleting an app from the server, manually remove it from the list of apps which should be validated when the server starts. Without this, if the server is restarted it will wait for the deleted app to start.
1. Fix the retry loop used when deleting a file from the server so that it waits for the requested amount of time.

PR is broken down into one commit for each logical part


Azquelt:
#build

dev/fattest.simplicity/src/com/ibm/websphere/simplicity/RemoteFile.java
```diff
@@ -110,6 +110,9 @@ public boolean retry(Operation op, long retryNs) throws Exception {
         }
 
         for ( int retryNo = 0; !didAct && retryNo < retryCount; retryNo++ ) {
+            if (retryNo != 0) {
```
tevans78: remove this guard ... want to sleep before all retries

Azquelt: Correct, I didn't realise it did an initial attempt outside of the loop.

Azquelt: Now fixed

Azquelt:
#build

Azquelt:

PR #8103, opened 2019-07-01
Prevent Batcher trying to commit batches after the app has closed

Azquelt:
Add some shutdown logic to the batcher so that if we lose the partition
or the app is shut down we don't try to commit any more offsets.

In particular this prevents the problem where the batcher schedules an
update for some time in the future and the application shuts down in the
mean time so the CDI beans aren't valid by the time it runs.


Azquelt:
No tests as the change is in response to a test failing in the build.

Azquelt:
#build

Azquelt:
Last build failed with a network error.

#build

PR #8105, opened 2019-07-01
Fix test to work in non-English locales

brenthdaniel:
Test case fix to correctly check logs in non-English locales


brenthdaniel:
#build

PR #8137, opened 2019-07-03
Clean FAT test to better wait for app stop

andymc12:
FAT test currently stops an app by directly moving the app archive from the dropins directory, then waiting a second and a half.  It doesn't check that the app is completely stopped/removed before proceeding.  This change fixes that by using a server framework method that does what we want.


andymc12:
#build

PR #8142, opened 2019-07-04
Ensure tracked kafka partitions are cleaned up on shutdown

Azquelt:
Cleanup code already existed for closing a PartitionAckTracker when the
kafka broker unassigned a partition from us, we just need to ensure that
same code is called when the app is shut down.

This prevents us from trying to commit offsets after the app has closed.

This is a follow up to #8103 which improved the cleanup code, but didn't actually ensure it was called on shutdown.


Azquelt:
#build

Azquelt:
Personal build above did catch the latest commmit

PR #8145, opened 2019-07-08
Grant kafka test permission to connect to any hostname

Azquelt:
Initially I thought that testcontainers provided a local proxy and so
the kafka client would only need to connect to localhost but this is not
the case.

The names of our test systems are not guaranteed to stay the same so we
need to grant this test permission to connect out to any host.


Azquelt:
#build

PR #8166, opened 2019-07-09
Remove microProfile-3.0 from GA profiles

tevans78:


tevans78:
Not going to need this PR

PR #8167, opened 2019-07-09
Rework console log test

brenthdaniel:
We have a test that tries to verify that nothing gets written to the console log when console logging is disabled. Over time we have implemented so many workarounds for java messages being written to the log that the test is effectively not being run most of the time.


brenthdaniel:
#build

PR #8178, opened 2019-07-10
test config & validation REST endpoints for app-defined connection factory from an embedded resource adapter

njr-11:
Add an embedded resource adapter to one of our FAT buckets.
Define an app-defined connection factory pointing to it
and write a test that uses the /config/ REST endpoint to verify its configuration and available API.

Also write a test that uses the /validation/ REST endpoint to test the configured app-defined connection factory, and update product code to allow the /validation/ endpoint to access the connection factory for purposes of testing it.


njr-11:
#build

njr-11:
#build

PR #8182, opened 2019-07-10
Add small wait to make sure server.xml change is recognized

brenthdaniel:
The ActivationTypeTest replaces one server.xml with another that has an identical size. On some JDKs (particularly older Oracle versions) the timestamps don't have enough precision to recognize that the file was changed. So, if the test happens to update server.xml within 1 second of the last update, we will ignore the change. 

I'm going to add a small delay so that we will always pick up this change.


brenthdaniel:
#build

PR #8184, opened 2019-07-10
Porting persistant_multiple test bucket to open liberty 

KyleAure:
Commit History
Original port
Refactoring to work in OL


KyleAure:
#build

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
```diff
@@ -20,24 +20,42 @@
...
+  <dataSource id="DefaultDataSource" jndiName="jdbc/DefaultDataSource" containerAuthDataRef="authx">
     <jdbcDriver libraryRef="FATJDBCLib"/>
     <properties.derby.embedded createDatabase="create" databaseName="memory:persistmultidb"/>
   </dataSource>
-
-  <library id="FATJDBCLib">
-    <fileset dir="${shared.resource.dir}/derby" includes="derby.jar"/>
-  </library>
+  
+    <!-- anonymous database for scheduled tasks, also accessible to the application -->
```
aguibert: why is AnonymousDataSource and AnonymousDataSource_DB2 needed? We should be able to get by with just DefaultDataSource for everything

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
```diff
@@ -20,24 +20,42 @@
...
+    <fileset dir="${shared.resource.dir}/derby" includes="derby.jar"/>
+  </library>
+  
+  <library id="AnonymousJDBCLib">
+    <fileset dir="${shared.resource.dir}/jdbc" includes="driver.jar"/>
+  </library>
 
   <!-- database for scheduled tasks, also accessible to the application -->
-  <dataSource id="DefaultDataSource" fat.modify="true" containerAuthDataRef="authx">
+  <dataSource id="DefaultDataSource" jndiName="jdbc/DefaultDataSource" containerAuthDataRef="authx">
```
aguibert: it isn't necessary to add a jndiName to the DefaultDataSource -- it will automatically be available as `java:comp/DefaultDataSource`

dev/com.ibm.ws.concurrent.persistent_fat_multiple/test-applications/persistmultitest/resources/WEB-INF/ibm-web-bnd.xml
```diff
@@ -11,6 +11,8 @@
  -->
 <web-bnd xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://websphere.ibm.com/xml/ns/javaee"
     xsi:schemaLocation="http://websphere.ibm.com/xml/ns/javaee http://websphere.ibm.com/xml/ns/javaee/ibm-web-bnd_1_0.xsd" version="1.0">
-
+	<resource-ref name="jdbc/DefaultDataSource" binding-name="jdbc/DefaultDataSource"/>
```
aguibert: these resource-refs can be removed when we go back down to 1 datasource for the bucket

aguibert: same comment applies to the ibm-web-ext.xml and the web.xml too

dev/com.ibm.ws.concurrent.persistent_fat_multiple/test-applications/persistmultitest/src/web/DBUpdateTask.java
```diff
@@ -21,9 +21,13 @@
  * Task that update an entry in the database when it runs.
  */
 public class DBUpdateTask implements Runnable {
+	
+    //Get resource reference of either embedded or networked database for lookup.
+    private static final String resourceRef = System.getenv("DB_REF");
```
aguibert: lets try and get this mechanism working without any application changes necessary. I think the key here is that we use 1 datasource for everything, and not have to swap to a different datasource for Derby vs. other DBs

dev/com.ibm.ws.concurrent.persistent_fat_multiple/build.gradle
```diff
@@ -8,6 +8,71 @@
...
+               'org.apache.commons:commons-compress:1.18',
+               'org.rnorth.duct-tape:duct-tape:1.0.7',
+               'org.rnorth.visible-assertions:visible-assertions:2.1.2',
+               'org.rnorth:tcp-unix-socket-proxy:1.0.2',
+               'net.java.dev.jna:jna:5.2.0',
+               'org.slf4j:slf4j-api:1.7.7',
+               'org.slf4j:slf4j-jdk14:1.7.7'
+
+  //JDBC driver - depends on database type
+  switch(dbType){
```
aguibert: at the time we build the FAT (i.e. the time this code runs) we won't know what DB will be used yet. Furthermore, the same autoFVT.zip produced by this build may be re-used to run against every DB.
The easiest way around this would be to just package every (available) JDBC all the time into the autoFVT.zip, since we don't know which JDBC driver will be used at runtime.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/build.gradle
```diff
@@ -8,6 +8,71 @@
...
+      anonymousJDBC 'com.microsoft.sqlserver:mssql-jdbc:7.2.2.jre8'
+      break
+  }
+}
+
+task addAnonymousDriver(type: Copy) {
+  shouldRunAfter jar
+  from configurations.anonymousJDBC
+  into new File(autoFvtDir, 'publish/shared/resources/jdbc')
+  rename '.*.jar', 'driver.jar'
```
aguibert: lets copy all of the JDBC drivers into the jdbc folder, but instead of naming them to `driver.jar` we can just strip off the versions so we have things like `db2jcc.jar` and `sqlserver.jar` (or whatever the base names are)

dev/com.ibm.ws.concurrent.persistent_fat_multiple/build.gradle
```diff
@@ -8,6 +8,71 @@
...
+  switch(dbType){
+    case 'DB2':
+      anonymousJDBC 'com.ibm.db2:jcc:11.1.4.4'
+      break
+    case 'Postgre':
+      anonymousJDBC 'org.postgresql:postgresql:42.2.5'
+      break
+    case 'Oracle':
+      //If we are doing a personal build then prereq.dbtest will be available to get oracle jdbc driver
+      if (!isLocalBuild){
```
aguibert: lets still allow for the possibility that prereq.dbtest is available on a local build. Perhaps we could change this to:
```groovy
if (!isLocalBuild || <the Oracle JDBC jar file exists)
```

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,144 @@
...
+	 * 
+	 * If we are using a network database, a test container running
+	 * that database will be returned.  
+	 * [Postgre, DB2, Oracle, SQLServer, MySQL]
+	 * 
+	 * If we are using derby embedded or unsupported database
+	 * a no-op container will be returned.
+	 * [Derby, Informix, Sybase]
+	 * 
+	 * @param testClasses - Provide the name(s) of your test class(es). These will be passed to the test container which will then provide log output to the class(es) log.
```
aguibert: I don't think it's necessary to pass in the test classes just for the sake of logging. I recommend either:
A) Let the FAT bucket set up its own logging by calling `withLogConsumer()` on the returned `GenericContainer`
B) do the logging as `DatabaseContainerFactory.class` instead of as the test class
I have a slight preference towards (B) since it's less boiler plate for the FAT to worry about (once we move this all to fattest.simplicity)

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,144 @@
...
+	 * 
+	 * If we are using derby embedded or unsupported database
+	 * a no-op container will be returned.
+	 * [Derby, Informix, Sybase]
+	 * 
+	 * @param testClasses - Provide the name(s) of your test class(es). These will be passed to the test container which will then provide log output to the class(es) log.  
+	 * @return GenericContainer - Test test container.
+	 */
+	@SuppressWarnings("rawtypes")
+	public static GenericContainer<?> get(Class... testClasses) {
```
aguibert: Instead of returning a GenericContainer here, we should return a `JdbcDatabaseContainer` which is a common superclass of all of the various DB types

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,144 @@
...
+		case "Oracle":
+			//TODO does IBM/Websphere have an oracle docker container that could be used here instead?
+			return new OracleContainer("oracleinanutshell/oracle-xe-11g")
+					.withLogConsumer(DatabaseContainerFactory::log);
+		case "SQLServer":
+			return new MSSQLServerContainer<>()
+					.withLogConsumer(DatabaseContainerFactory::log);
+		case "Derby":
+		default:
+			return new MockServerContainer();
```
aguibert: why do we need to use MockServer here? As far as I can tell it's not used aside from having _some_ container to return in the case of Derby.

Instead, lets implement a "dummy" `DerbyContainer` class that extends `JdbcDatabaseContainer` (see `CustomPostgreSQLContainer` for an example) and we can just override/hardcode the necessary methods to so things like `container.start()` simply no-op for the Derby case

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,144 @@
...
+	
+	/**
+	 * Returns port to be used to access test container.
+	 * Each database uses a different internal port to accept incoming 
+	 * connection requests.  The test container automatically maps a 
+	 * random external port to the default internal port. 
+	 * 
+	 * @return int - port of test container to be used for database access.
+	 */
+	public static int getMappedPort() {
```
aguibert: I don't this this method is necessary, instead the test can just call `container.getFirstMappedPort()`

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
aguibert: also, lets rename this method to `create()` because we are returning a new container each time. IMO `get()` implies we will be returning the same singleton result on each call (which may matter if a FAT needs multiple DBs)

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,144 @@
...
+ * 
+ * The 'fat.bucket.db.type' property is set to different databases
+ * by our test infrastructure when a fat-suite has the property
+ * 'fat.test.databases' set to true
+ * 
+ * TODO: Since this factory class is general can it be moved to fattest simplicity?
+ *
+ */
+public class DatabaseContainerFactory {
+	private static String database = System.getProperty("fat.bucket.db.type", "Derby");
```
aguibert: instead of using a String here and in the switch blocks, perhaps we could make/reuse an enum for the DBs supported? This would help enforce more desirable behavior for an unknown/misspelled DB type raising an exception instead of silently falling back to using Derby

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/MultiplePersistentExecutorsTest.java
```diff
@@ -33,23 +36,72 @@
  * Tests for multiple persistent scheduled executor instances sharing the same database
  */
 @RunWith(FATRunner.class)
-@SkipIfSysProp(DB_Informix) // persistent executor is not support on Informix
+@SkipIfSysProp({DB_Informix, DB_Sybase, "fat.oracle.is.available=false"})
```
aguibert: checking `fat.oracle.is.available=false` will work when running locally, but not for remote builds because we build the FAT on one machine and then run it potentially multiple times on a different machine at a later time

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
```diff
@@ -20,24 +20,42 @@
...
-
-  <library id="FATJDBCLib">
-    <fileset dir="${shared.resource.dir}/derby" includes="derby.jar"/>
-  </library>
+  
+    <!-- anonymous database for scheduled tasks, also accessible to the application -->
+  <dataSource id="AnonymousDataSource" jndiName="jdbc/AnonymousDataSource">
+    <jdbcDriver libraryRef="AnonymousJDBCLib"/>
+    <properties serverName="${env.DB_HOST}" portNumber="${env.DB_PORT}" databaseName="${env.DB_NAME}" 
+    	user="${env.DB_USER}" password="${env.DB_PASS}" URL="${env.DB_URL}"/>
```
aguibert: typically configurations will configure as individual properties (serverName, portNumber, databasename, etc) OR using the URL, but not both.
Behavior is undefined if both individual props and URL is used. For easy portability across the DB types, I recommend just using the URL property, which is easily obtainable by calling `container.getJdbcUrl()`, and we can also override the dummy Derby container to return something like `dbc:derby:memory:persistmultidb;create=true`. This way we won't have to treat Derby Embedded any differently than the other DBs in the FAT setup code

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
```diff
@@ -20,24 +20,42 @@
   <include location="../fatTestPorts.xml"/>
 
   <persistentExecutor id="executor1" jndiName="concurrent/executor1" pollInterval="4s" retryLimit="1" pollSize="100"/>
-
   <persistentExecutor id="executor2" jndiName="concurrent/executor2" taskStoreRef="defaultDatabaseStore" pollInterval="4s" retryLimit="1"/>
 
   <databaseStore id="defaultDatabaseStore" keyGenerationStrategy="SEQUENCE"/>
+  
+  <library id="FATJDBCLib">
+    <fileset dir="${shared.resource.dir}/derby" includes="derby.jar"/>
```
aguibert: lets just use 1 `<library>` here, and we can control the `includes` attribute from the FAT setup with an env var

KyleAure:
#build requesting PB was having issues getting SQL Server to run locally.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Gets DataSource connection properties from JdbcDatabaseContainer, such as, username, password, and url.
+     * Then adds these properties to the server's environment variables.
+     * And, optionally, updates server config properties dependent on the type of database being used.
+     * 
+     * @param serv - LibertyServer server instance being used for this FAT suite.
+     * @param cont - JdbcDatabaseContainer instance being used for database connectivity.
+     * @param type - DBContainerType type information about this container.
+     */
+    public static void setupDataSourceProperties(LibertyServer serv, JdbcDatabaseContainer<?> cont, DatabaseContainerType type) throws Exception {
```
aguibert: I would prefer we didn't have this helper method, and instead just let the FAT's `@BeforeClass` set properties according to the env vars it has chosen. For example:
```java
server.addEnvVar("DB_DRIVER", type.getDriverName());
server.addEnvVar("DB_HOST", cont.getContainerIpAddress());
server.addEnvVar("DB_PORT", cont.getFirstMappedPort());
// etc...
```
this will be helpful because the FAT developer can declare whatever env vars they want, instead of needing to choose the specific variables names such as `DB_DRIVER` in their server.xml config.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
```diff
@@ -0,0 +1,138 @@
...
+			ds.getProperties_microsoft_sqlserver().add((Properties_microsoft_sqlserver) props);
+			break;
+		case SYBASE:
+			props = new Properties_sybase();
+			ds.clearDataSourceDBProperties();
+			ds.getProperties_sybase().add((Properties_sybase) props);
+			break;
+    	}
+    	
+    	props.setUser(username);
```
aguibert: I would prefer we don't use the server config API here, and instead let the FAT set environment variables on the server before it's started.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
```diff
@@ -0,0 +1,138 @@
...
+    	props.setServerName(serverName);
+    	props.setPortNumber(port);
+    	if(databaseName != null)
+    		props.setDatabaseName(databaseName);
+    	
+    	serv.updateServerConfiguration(cloneConfig);
+    }
+    
+    /**
+     * Private Utility Method:
```
aguibert: javadoc indicates private but the method is public

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/MultiplePersistentExecutorsTest.java
```diff
@@ -0,0 +1,280 @@
...
+    private static ServerConfiguration originalConfig;
+    
+    private static final LibertyServer server = FATSuite.server;
+    
+    private static final String TASK_ID_SEARCH_TEXT = "Task ids: ";
+    
+	@BeforeClass
+    public static void setUp() throws Exception {
+		//Skip if we do not have test container to run against
+		assumeTrue(dbContainerType.hasTestContainer());
```
aguibert: how does assumeTrue behave in a BeforeClass method? If the assumption is false, do all the tests skip? I would imagine we still want these tests to run even on Derby

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerType.java
```diff
@@ -0,0 +1,88 @@
...
+    SYBASE("Sybase", "jconn4.jar", false);
+
+    private String name;
+    private String driverName;
+    private boolean hasTestContainer;
+
+    DatabaseContainerType(final String name, final String driverName, final boolean inService) {
+        this.name = name;
+        this.driverName = driverName;
+        this.hasTestContainer = inService;
```
aguibert: this boolean and corresponding methods can be removed, since we can get rid of informix and sybase and just comment out the sqlserver code for now

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/NOOPJDBCContainer.java
```diff
@@ -0,0 +1,106 @@
...
+ * an unsupported database or to test against derby embedded.
+ * 
+ * This test container overrides the start and stop methods
+ * to prevent the creation of a docker container. 
+ * 
+ * Returns connection information for a derby embedded driver. 
+ *  
+ * TODO: Since this class is general it should be moved to fattest simplicity in the future
+ */
+class NOOPJDBCContainer<SELF extends NOOPJDBCContainer<SELF>> extends JdbcDatabaseContainer<SELF> {
```
aguibert: Lets rename this class to `DerbyNoopContainer`, since it only applies for Derby

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/NOOPJDBCContainer.java
```diff
@@ -0,0 +1,106 @@
...
+	}
+	
+	@Override
+	public SELF withDatabaseName(String name) {
+		this.databaseName = name;
+		return self();
+	}
+	@Override
+	public String getDatabaseName() {
+		return "memory:" + this.databaseName;
```
aguibert: lets not assume an in-memory database is desired. Instead, the caller can do `withDatabaseName("memory:foo")` if they want

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
```diff
@@ -0,0 +1,49 @@
...
+  <databaseStore id="defaultDatabaseStore" keyGenerationStrategy="SEQUENCE"/>
+  
+  <library id="AnonymousJDBCLib">
+    <fileset dir="${shared.resource.dir}/jdbc" includes="${env.DB_DRIVER}"/>
+  </library>
+  
+  <!-- Default database, properties changed by FATRunner during database rotation. -->
+  <dataSource id="DefaultDataSource">
+    <jdbcDriver libraryRef="AnonymousJDBCLib"/>
+  	<properties.derby.embedded createDatabase="create" databaseName="memory:persistmultidb" user="userx" password="passx"/>
```
aguibert: to go along with some of the earlier comments, lets use a generic properties element with variables like this:
```xml
<properties hostname="${DB_HOST}" portNumber="${DB_PORT}" databaseName="${DB_NAME}"
                     user="${DB_USER}" password="${DB_PASSWORD}" createDatabase="create"/>
```
This might give some warnings in logs because `portNumber` is not a valid attribute for Derby Embedded, and `create` is not a valid attribute for all DBs besides Derby Embedded, but we can ignore these on `stopServer()` and it will simplify the code because we won't need to deal with the server config API anymore

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
```diff
@@ -0,0 +1,49 @@
...
+  </dataSource>
+  
+  <javaPermission codebase="${shared.resource.dir}/jdbc/db2jcc.jar" className="java.security.AllPermission"/>
```
aguibert: instead of adding perms to all drivers, could we just have 1 `<javaPermission>` and re-use the `DB_DRIVER` variable to grant perms to the in-use driver like this?
```xml
<javaPermission codebase="${shared.resource.dir}/jdbc/${DB_DRIVER}" className="java.security.AllPermission"/>
```
this will help as drivers come/go or change names (e.g. ojdbc8_g.jar --> ojdbc9.jar)

dev/cnf/oss_dependencies.maven
```diff
@@ -31,6 +31,7 @@ com.graphql-java:graphql-java-servlet:6.1.3
 com.graphql-java:graphql-java:11.0
 com.graphql-java:java-dataloader:2.2.1
 com.ibm.db2:jcc:11.1.4.4
+com.ibm.db2:jcc:11.5.0.0
```
aguibert: instead of adding a second DB2 artifact, can we either:
A) re-use the existing one or
B) update the existing one to use 11.5.0.0
so that we only have 1 DB2 driver in the repo

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,127 @@
...
+	private static JdbcDatabaseContainer<?> initContainer() {
+		switch(dbContainerType) {
+		case DB2:
+			return new Db2Container().acceptLicense()
+				  // Use 5m timeout for local runs, 15m timeout for remote runs
+				  .withStartupTimeout(Duration.ofMinutes(FATRunner.FAT_TEST_LOCALRUN ? 5 : 15))
+				  .withLogConsumer(DatabaseContainerFactory::log);
+		case ORACLE:
+			//Check if we are running local and unable to get JDBC driver, if so warn user.
+			if(FATRunner.FAT_TEST_LOCALRUN && !DatabaseContainerUtil.isJdbcDriverAvailable(dbContainerType)) {
```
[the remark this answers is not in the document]
KyleAure: Instead of throwing an error we could also get rid of this check all together.  Then an error will be thrown from the test-containers code itself.  
```
[09/04/2019 10:08:54:858 CDT] 001 GenericContainer               tryStart                       S Could not start container
org.testcontainers.containers.JdbcDatabaseContainer$NoDriverFoundException: Could not get Driver
	at org.testcontainers.containers.JdbcDatabaseContainer.getJdbcDriverInstance(JdbcDatabaseContainer.java:169)
	at org.testcontainers.containers.JdbcDatabaseContainer.createConnection(JdbcDatabaseContainer.java:191)
	at org.testcontainers.containers.JdbcDatabaseContainer.waitUntilContainerStarted(JdbcDatabaseContainer.java:129)
	at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:332)
	at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:285)
```

Which seems pretty self explanatory

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/MultiplePersistentExecutorsTest.java
KyleAure: Yes, as soon as the assumption fails test execution ends.  The only thing that is run after that is cleanup code.  Derby is marked as having a test container since we have the noop container.  But I think I will remove the hasTestContainer attribute in favor of having the FAT fail if say "fat.bucket.db.type=Informix" or any other database that is not set up for the database rotation.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
KyleAure: I think that makes more sense for the case when we are testing against a single database test-container.  In that case you would want the flexibility to run tests using different data-source settings.  But, when doing database rotation I think this design works a better.  It eliminates the need for the person writing the test to worry about making sure the correct data-source properties are set for every database (since situations like DB2 arise where URL is not supported).  Instead they just worry about the test logic.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
KyleAure: This actually causes Exceptions to be thrown when attempting to create a connection.
`java.lang.AssertionError: Expected response 200, received 500 (Internal Server Error) while connecting to http://localhost:8010/persistmultitest?testMethod=testConnectionSharing.`
A 500 is returned since the warning about incorrect properties is thrown as a SQL Exception by the application 
`java.sql.SQLException: DSRA8020E: Warning: The property &#39;createDatabase&#39; does not exist on the DataSource class com.ibm.db2.jcc.DB2XADataSource.`

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
KyleAure: I would also be interested to know if we have FAT test suites that use database rotation with multiple datasources.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/bnd.bnd
```diff
@@ -0,0 +1,40 @@
...
+-buildpath: \
+	com.ibm.websphere.javaee.annotation.1.1;version=latest,\
+	com.ibm.websphere.javaee.concurrent.1.0;version=latest,\
+	com.ibm.websphere.javaee.servlet.3.1;version=latest,\
+	com.ibm.websphere.javaee.transaction.1.1;version=latest,\
+	com.ibm.ws.concurrent.persistent;version=latest,\
+	org.testcontainers:database-commons;version=1.12.0,\
+	org.testcontainers:jdbc;version=1.12.0,\
+	org.testcontainers:testcontainers;version=1.12.0,\
+	org.testcontainers:db2;version=1.12.0,\
```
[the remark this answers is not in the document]
KyleAure: We do need need these compile time dependencies since the factory class calls:
```java
return new Db2Container().acceptLicense()
.withStartupTimeout(Duration.ofMinutes(FATRunner.FAT_TEST_LOCALRUN ? 5 : 15))
.withLogConsumer(DatabaseContainerFactory::log);
```
which requires us to `import org.testcontainers.containers.Db2Container;`

When we move the factory class to `fattest.simplify`, then I agree we will just need the generic jdbc module

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
```diff
@@ -0,0 +1,149 @@
...
+	private static final String nl = System.getProperty("line.separator");
+	
+	/**
+	 * Initializes SQL Server database.  Creates a database with a given name, and
+	 * runs the sp_sqljdbc_xa_install stored procedure to enable XA transactions.
+	 * 
+	 * @param cont - SQLServer Test Container
+	 * @param databaseName - Name to be given to database
+	 */
+	public static void initSQLServer(JdbcDatabaseContainer<?> cont, String databaseName) {
```
aguibert: the `databaseName` param can likely be removed. I believe we can get the same info inside the method by calling `cont.getDatabaseName()`

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
```diff
@@ -0,0 +1,149 @@
...
+		// Create Database
+        try (Connection conn = cont.createConnection("")) {
+            Statement stmt = conn.createStatement();
+            stmt.execute("CREATE DATABASE [" + databaseName + "];");
+            stmt.close();
+        } catch (Exception e) {
+        	Log.error(c, "initContainer", e);
```
aguibert: instead of burying this exception and the next one, we should let the errors get thrown out of the method so we don't attempt to run the tests later on and have them fail in non-obvious ways

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
aguibert: here are the hits I got (includes OL and CL) for usages of datasource modification. There are quite a few with multiple datasources, and also many that don't use `id="DefaultDataSource"`. At minimum, this method will need to accept one or more Strings for the dataSource ID's that should be updated, rather than assuming only the DefaultDataSource needs to be updated or that one exists.

```
./com.ibm.ws.jpa_spec10_injection_fat/publish/servers/JPAServerDFI/database.xml:    <dataSource id="jdbc/JPA_DS" jndiName="jdbc/JPA_DS" fat.modify="true">
./com.ibm.ws.jpa_spec10_injection_fat/publish/servers/JPAServerDFI/database.xml:    <dataSource id="jdbc/JPA_NJTADS" jndiName="jdbc/JPA_NJTADS" fat.modify="true" transactional="false">
./com.ibm.ws.jpa_spec10_injection_fat/publish/servers/JPAServer/database.xml:    <dataSource id="jdbc/JPA_DS" jndiName="jdbc/JPA_DS" fat.modify="true">
./com.ibm.ws.jpa_spec10_injection_fat/publish/servers/JPAServer/database.xml:    <dataSource id="jdbc/JPA_NJTADS" jndiName="jdbc/JPA_NJTADS" fat.modify="true" transactional="false">
./com.ibm.ws.jpa_spec10_injection_fat/publish/servers/JPAServerDMI/database.xml:    <dataSource id="jdbc/JPA_DS" jndiName="jdbc/JPA_DS" fat.modify="true">
./com.ibm.ws.jpa_spec10_injection_fat/publish/servers/JPAServerDMI/database.xml:    <dataSource id="jdbc/JPA_NJTADS" jndiName="jdbc/JPA_NJTADS" fat.modify="true" transactional="false">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.anno.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.shortname.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.authdata.fat/server.xml:	<dataSource id="derby1" jndiName="jdbc/derby1fat" containerAuthDataRef="DBAuth"  fat.modify="true" >
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.dbidstore.deferred.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.dbidstore.deferred.fat/server.xml:    <dataSource id="NoDatabase" fat.modify="true" jndiName="jdbc/NoDatabase">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.dbidstore.immediate.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.dbidstore.immediate.fat/server.xml:    <dataSource id="NoDatabase" fat.modify="true" jndiName="jdbc/NoDatabase">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.authalias.fat/server.xml:	<dataSource id="derby1" jndiName="jdbc/derby1fat" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.nouser.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.hash.bean.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.security.javaeesec_fat/publish/servers/com.ibm.ws.security.javaeesec.db.hash.fat/server.xml:	<dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_spec10_fat/publish/servers/JPA10Server/database.xml:    <dataSource id="jdbc/JPA_DS" jndiName="jdbc/JPA_DS" fat.modify="true">
./com.ibm.ws.jpa_spec10_fat/publish/servers/JPA10Server/database.xml:    <dataSource id="jdbc/JPA_NJTADS" jndiName="jdbc/JPA_NJTADS" fat.modify="true" transactional="false">
./com.ibm.ws.jdbc_fat_v41/publish/servers/com.ibm.ws.jdbc.fat.v41/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jdbc_fat_v41/publish/servers/com.ibm.ws.jdbc.fat.v41/server.xml:    <dataSource id="ds1" jndiName="jdbc/${id}" fat.modify="true">
./com.ibm.ws.jdbc_fat_v41/publish/servers/com.ibm.ws.jdbc.fat.v41/server.xml:    <dataSource id="ds2" jndiName="jdbc/${id}" fat.modify="true">
./com.ibm.ws.jdbc_fat_v41/publish/servers/com.ibm.ws.jdbc.fat.v41/server.xml:    <dataSource id="ds3" jndiName="jdbc/${id}" fat.modify="true">
./com.ibm.ws.jdbc_fat_v41/publish/servers/com.ibm.ws.jdbc.fat.v41/server.xml:    <dataSource id="XAds" jndiName="jdbc/${id}" type="javax.sql.XADataSource" fat.modify="true">
./com.ibm.ws.jpa_spec20_fat/publish/servers/JPA20Server/database.xml:    <dataSource id="jdbc/JPA_DS" jndiName="jdbc/JPA_DS" fat.modify="true">
./com.ibm.ws.jpa_spec20_fat/publish/servers/JPA20Server/database.xml:    <dataSource id="jdbc/JPA_NJTADS" jndiName="jdbc/JPA_NJTADS" fat.modify="true" transactional="false">
./com.ibm.ws.jpa_22_fat/publish/servers/JPA22QueryServer/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_22_fat/publish/servers/JPA22InjectionServer/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_22_fat/publish/servers/JPA22InjectionServer/server.xml:    <dataSource id="jdbc/NonTransactionalDataSource" jndiName="jdbc/NonTransactionalDataSource" fat.modify="true" transactional="false">
./com.ibm.ws.jpa_22_fat/publish/servers/CDIFatServer/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_22_fat/publish/servers/JPA22TimeAPIServer/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_22_fat/publish/servers/JPABootstrapFATServer/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_22_fat/publish/servers/EJBPassivationServer/server.xml:    <dataSource id="jdbc/JPA_DS" jndiName="jdbc/JPA_DS" fat.modify="true">
./com.ibm.ws.jpa_22_fat/publish/servers/EJBPassivationServer/server.xml:    <dataSource id="jdbc/JPA_NJTADS" jndiName="jdbc/JPA_NJTADS" fat.modify="true" transactional="false">
./com.ibm.ws.jpa_22_fat/publish/servers/JPA22BeanValidationServer/server.xml:    <dataSource id="DefaultDataSource" jndiName="jdbc/TransactionalDataSource" fat.modify="true">
./com.ibm.ws.jpa_22_fat/publish/servers/JPA22BeanValidationServer/server.xml:    <dataSource id="jdbc/NonTransactionalDataSource" jndiName="jdbc/NonTransactionalDataSource" fat.modify="true" transactional="false">
./com.ibm.ws.jpa.container.ormdiagnostics_fat/publish/servers/JPADiagTestServer/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa.container.ormdiagnostics_fat/build/libs/autoFVT/publish/servers/JPADiagTestServer/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_spec10_query_fat/publish/servers/JPAQueryServer/database.xml:    <dataSource id="jdbc/JPA_DS" jndiName="jdbc/JPA_DS" fat.modify="true">
./com.ibm.ws.jpa_spec10_query_fat/publish/servers/JPAQueryServer/database.xml:    <dataSource id="jdbc/JPA_NJTADS" jndiName="jdbc/JPA_NJTADS" fat.modify="true" transactional="false">


./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.embeddable/server.xml:    <dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.embeddable/server.xml:    <dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.emlocking/server.xml:	<dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true" type="javax.sql.XADataSource">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.emlocking/server.xml:	<dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.dsoverride/server.xml:    <dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true" type="javax.sql.XADataSource">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.dsoverride/server.xml:    <dataSource jndiName="jdbc/AltJTADataSource" jdbcDriverRef="DerbyEmbedded" fat.modify="true" type="javax.sql.XADataSource">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.cache/server.xml:	<dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.cache/server.xml:	<dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.ejbpassivation/server.xml:    <dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true" type="javax.sql.XADataSource">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.entitymanager20/server.xml:	<dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true" type="javax.sql.XADataSource">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.entitymanager20/server.xml:	<dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.fetchgroup/server.xml:	<dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true" type="javax.sql.XADataSource">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.fetchgroup/server.xml:	<dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.ojexternalization/server.xml:	<dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true" type="javax.sql.XADataSource">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.ojexternalization/server.xml:	<dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.criteriaquery/server.xml:	<dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true">
./com.ibm.ws.jpa_fat/publish/servers/com.ibm.ws.jpa.fat.criteriaquery/server.xml:	<dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfatboot" jndiName="jdbc/dsfatboot" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfatboot1" jndiName="jdbc/dsfatboot1" fat.modify="true" jdbcDriverRef="FATJDBCDriver1" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfatboot2" jndiName="jdbc/dsfatboot2" fat.modify="true" jdbcDriverRef="FATJDBCDriver2" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfatboot3" jndiName="jdbc/dsfatboot3" fat.modify="true" jdbcDriverRef="FATJDBCDriver3" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfatboot4" jndiName="jdbc/dsfatboot4" fat.modify="true" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfatboot5" jndiName="jdbc/dsfatboot5" fat.modify="true" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfatboot6" jndiName="jdbc/dsfatboot6" fat.modify="true" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat0" jndiName="jdbc/dsfat0" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.DataSource" connectionSharing="MatchOriginalRequest">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat1" jndiName="jdbc/dsfat1" fat.modify="true" jdbcDriverRef="FATJDBCDriver" connectionManagerRef="conMgr1" connectionSharing="MatchCurrentState" transactional="true">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat2" jndiName="jdbc/dsfat2" fat.modify="true" jdbcDriverRef="FATJDBCDriver" connectionManagerRef="conMgr2" type="javax.sql.XADataSource" isolationLevel="TRANSACTION_READ_COMMITTED" recoveryAuthDataRef="recoveryAuth">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat3" jndiName="jdbc/dsfat3" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource jndiName="jdbc/dsfat4" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.XADataSource" transactional="true" connectionSharing="MatchCurrentState" recoveryAuthDataRef="recoveryAuth">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat20" fat.modify="true" jndiName="jdbc/dsfat20" jdbcDriverRef="FATJDBCDriver" connectionManagerRef="conMgr6" type="javax.sql.XADataSource" isolationLevel="TRANSACTION_READ_COMMITTED" recoveryAuthDataRef="recoveryAuth">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat21" fat.modify="true" jndiName="jdbc/dsfat21" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat22" fat.modify="true" jndiName="jdbc/dsfat22" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.jaas.fat/server.xml:    <dataSource id="dsfat23" fat.modify="true" jndiName="jdbc/dsfat23" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfatboot" jndiName="jdbc/dsfatboot" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfatboot1" jndiName="jdbc/dsfatboot1" fat.modify="true" jdbcDriverRef="FATJDBCDriver1" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfatboot2" jndiName="jdbc/dsfatboot2" fat.modify="true" jdbcDriverRef="FATJDBCDriver2" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfatboot3" jndiName="jdbc/dsfatboot3" fat.modify="true" jdbcDriverRef="FATJDBCDriver3" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfatboot4" jndiName="jdbc/dsfatboot4" fat.modify="true" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfatboot5" jndiName="jdbc/dsfatboot5" fat.modify="true" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfatboot6" jndiName="jdbc/dsfatboot6" fat.modify="true" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat0" jndiName="jdbc/dsfat0" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.DataSource" connectionSharing="MatchOriginalRequest">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat1" jndiName="jdbc/dsfat1" fat.modify="true" jdbcDriverRef="FATJDBCDriver" connectionManagerRef="conMgr1" connectionSharing="MatchCurrentState" transactional="true">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat2" jndiName="jdbc/dsfat2" fat.modify="true" jdbcDriverRef="FATJDBCDriver" connectionManagerRef="conMgr2" type="javax.sql.XADataSource" isolationLevel="TRANSACTION_READ_COMMITTED" recoveryAuthDataRef="recoveryAuth">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat3" jndiName="jdbc/dsfat3" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource" transactional="false">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource jndiName="jdbc/dsfat4" fat.modify="true" jdbcDriverRef="FATJDBCDriver" type="javax.sql.XADataSource" transactional="true" connectionSharing="MatchCurrentState" recoveryAuthDataRef="recoveryAuth">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat20" fat.modify="true" jndiName="jdbc/dsfat20" jdbcDriverRef="FATJDBCDriver" connectionManagerRef="conMgr6" type="javax.sql.XADataSource" isolationLevel="TRANSACTION_READ_COMMITTED" recoveryAuthDataRef="recoveryAuth">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat21" fat.modify="true" jndiName="jdbc/dsfat21" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat22" fat.modify="true" jndiName="jdbc/dsfat22" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource">
./com.ibm.ws.jdbc_fat/publish/servers/com.ibm.ws.jdbc.fat/server.xml:    <dataSource id="dsfat23" fat.modify="true" jndiName="jdbc/dsfat23" jdbcDriverRef="FATJDBCDriver" type="javax.sql.ConnectionPoolDataSource">
./com.ibm.ws.jdbc_fat_v42/publish/servers/jdbc.fat.v42/server.xml:    <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jpa_fat_java8/publish/servers/com.ibm.ws.jpa.fat.java8/server.xml:	<dataSource jndiName="jdbc/JPA_DS" jdbcDriverRef="DerbyEmbedded" fat.modify="true">
./com.ibm.ws.jpa_fat_java8/publish/servers/com.ibm.ws.jpa.fat.java8/server.xml:	<dataSource jndiName="jdbc/JPA_NJTADS" jdbcDriverRef="DerbyEmbedded" transactional="false" fat.modify="true">
./com.ibm.ws.concurrent.persistent_fat_timers/publish/servers/com.ibm.ws.concurrent.persistent.fat.timers/server.xml:  <dataSource id="DefaultDataSource" fat.modify="true">
./com.ibm.ws.jdbc_fat_fvt/publish/servers/com.ibm.ws.jdbc.fat.fvt/server.xml:    <dataSource id="basic-datasource" jndiName="jdbc/${id}" type="javax.sql.ConnectionPoolDataSource" fat.modify="true">
./com.ibm.ws.jdbc_fat_fvt/publish/servers/com.ibm.ws.jdbc.fat.fvt/server.xml:    <dataSource id="basic-datasource2" jndiName="jdbc/${id}" type="javax.sql.ConnectionPoolDataSource" fat.modify="true">
./com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml:  <dataSource id="DefaultDataSource" fat.modify="true" containerAuthDataRef="authx">
```

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerUtil.java
aguibert: also, lets just rename this method to `initDatabase` and have it no-op if the DB type is _not_ SQLServer. This way the FAT suite doesn't need to check what the DB type is and conditionally call this method

dev/com.ibm.ws.concurrent.persistent_fat_multiple/publish/servers/com.ibm.ws.concurrent.persistent.fat.multiple/server.xml
aguibert: oh, that's because we have `<variable name="onError" value="FAIL"/>` in the config. If you remove this line of config, then we ought to just get the warning which we can ignore on `stopServer()`

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,126 @@
...
+	private static final Class<DatabaseContainerFactory> c = DatabaseContainerFactory.class;
+	private static JdbcDatabaseContainer<?> container;
+	private static DatabaseContainerType dbContainerType;
+	
+	/**
+	 * @see #create(String databaseName)
+	 * 
+	 * Uses a default database name 'test'
+	 */
+	public static JdbcDatabaseContainer<?> create() {
```
aguibert: is this method ever used? I think we can just remove it

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
```diff
@@ -0,0 +1,126 @@
...
+	 * 
+	 * If {fat.bucket.db.type} is not set with a value, 
+	 * we will default to Derby Embedded.
+	 * 
+	 * @param databaseName - String passed to database container for those database containers 
+	 * that can use a custom database.
+	 *  
+	 * @return JdbcDatabaseContainer - The test container.
+	 */
+	public static JdbcDatabaseContainer<?> create(String databaseName) {
```
aguibert: I just realized that nothing is requiring `fat.test.databases: true` to be set when running locally, but this property needs to be set in order for the bucket to be picked up by the DB rotation build. So lets take the opportunity at the beginning of this method to make sure `fat.test.databases: true` is set and blow up with a message like:
`To use a generic database, the FAT must be opted into DB rotation by setting 'fat.test.databases: true' in the FAT project's bnd.bnd file`
if it's not set. This should guard against people getting everything working locally and then wondering why the bucket isn't running in DB rotation (or just not even realizing it's not running in rotation)

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DerbyNoopContainer.java
```diff
@@ -0,0 +1,106 @@
...
+		return self();
+	}
+	@Override
+	public String getDatabaseName() {
+		return databaseName;
+	}
+	
+	@Override
+	public String getJdbcUrl() {
+		return "jdbc:derby:" + databaseName + ";create=true";
```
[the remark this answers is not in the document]
KyleAure: Yes these are called when we use `setupDataSourceProperties`.  I have gone in and added  `UnsupportedOperationException` to the properties that do not make sense (ipaddress/port) and handled them where they are called using the super class.

dev/com.ibm.ws.concurrent.persistent_fat_multiple/fat/src/com/ibm/ws/concurrent/persistent/fat/multiple/DatabaseContainerFactory.java
KyleAure: I had left this here as a option later on if we have a test suite where we do not care what the name of the database is.  

And also as a general factory class I guess we shouldn't assume that it will be used specifically for database rotation.  I guess any test suite could use it to create a test-container if they wanted to.

KyleAure:
#build

PR #8188, opened 2019-07-11
FT 1.1: Do explicit counting of concurrent executions

Azquelt:
Switch from relying on the PolicyExecutor to indicate how many
executions are running to doing our own counting.

This ensures we can decrement the counter _before_ the Future is
completed, avoiding a race condition that was causing intermittent test
failures.


Azquelt:
#build

PR #8189, opened 2019-07-11
Fix timing issues in FutureShellTest

Azquelt:
Where possible we replace sleeps with manually completing futures in the
required order.

Where the test requires two actions to occur in parallel, use a shorter
"simulated work" time and a longer timeout time to reduce the risk of
false positive failures caused by timing issues.


Azquelt:
#build

PR #8199, opened 2019-07-14
Support putting encoded passwords in Oracle connectionProperties attr…

frowe:
…ibute of the properties.oracle config element


frowe:
#build

PR #8218, opened 2019-07-15
Fix server script missing ] message for debug launch

tjwatson:
There is a small syntax issue with the server script that doesn't seem to cause an actual issue, but does result in the following error message when launching with the `debug` option:

```
bin/server: line 1071: [: missing `]'
```
The issue is a missing space in the `if` statement at line 1071


tjwatson:
#build

PR #8234, opened 2019-07-16
Update OAuth client metatype API to return UTF-16

ayoho:
Ensures the metatype API response supports double byte character encoding and adds a dedicated FAT test to ensure it works.


ayoho:
#build

ayoho:
@brutif These are the same changes you had reviewed in Commercial Liberty. You had mentioned that some clients can't consume UTF-16 and only do UTF-8. However it doesn't look like the locales that require double byte encoding will render correctly with UTF-8; the only way to get both the single and double byte characters to play nice is to use UTF-16 (at least as far as I've tested).

PR #8237, opened 2019-07-16
BB 265329: Restrict fat_init_pooling to java 8+

KyleAure:
Fat bucket uses jdbc-4.2 which is only supported on java 8+


KyleAure:
#build

PR #8240, opened 2019-07-16
Pull request to remove tr.error from RegistrationEndpointServices

barbj:
ERROR_INVALID_CLIENT from RegistrationEndpointServices.processHeadOrGetSingleClient may not actually be an error. 
We can't remove the ffdc that is emitted from OAuth20EndpointServices, but the Tr.error that is emitted from this processHeadOrGetSingleClient can be removed.


barbj:
#build

barbj:
(Relying on lust build-lite since all I did was remove a tr.error)

PR #8244, opened 2019-07-16
Fix BundleInstallOriginBundleListener

tjwatson:
The BundleInstallOriginBundleListener has several issues.
What lead me to look at this class was the cost to load
the cache.  It uses an object stream which is slow.  But
then I noticed that the cache also was way too big,
that was because it incorrectly is tracking bundles
installed by feature manager.  Then I noticed how complicated
the purging code was and thought it was way to much to try
and save some String storage in memory.  This can easily be
Bundle ID longs instead.

Stop using object stream, use data stream instead
Stop tracking bundles installed by feature manager
Stop purging the cache Map
Use simple Bundle ID longs instead of location strings to track

Add tests to actually make sure this stuff keeps working


tjwatson:
#build

tjwatson:
#build

PR #8259, opened 2019-07-17
Improve init_pooling test failure logic

KyleAure:
This test checks to see if a task completes in the allotted time when pooling. 
The for-loop checks the status and resolves when the task should have completed.
When we check to see if the task has completed we first need to check to see if task is still scheduled. 
Otherwise, evaluation of `status.isDone()` will throw it's own exception instead of falling through to the test exception.

This change will ensure that this test will properly fail if a task does not completed in the expected amount of time.


KyleAure:
#build

PR #8270, opened 2019-07-18
Porting persistant_fat test bucket to open liberty 

KyleAure:
Commit History
Original port
Refactoring to work in OL

Wait to deliver this branch until we can implement testing against multiple databases.


KyleAure:
Three tests have been commented out since they do not work correctly when using Postgre database.  Product code changes are necessary and will be tracked using #9176 

We should get this test suite into OL now.  Then, the product code change  PR can re-enable these tests.

KyleAure:
#build

PR #8278, opened 2019-07-18
Add Kafka test using a custom data type

### LINKED ISSUE #8230: RM Kafka: FAT tests for Kafka message converters
Although reactive messaging doesn't provide an API for converting between types, the Kafka client does and providing a converter is mandatory. Since there's no "default" converter, we should check that this functionality works.

* test that several Kafka provided converters work
* test that a custom converter packaged in the app works
* test the behaviour when the type produced by a converter is not the type expected by the connected processor

tevans78:
fixes #8230 
#build

### DIFF
--- a/dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/delivery/KafkaAcknowledgementTestServlet.java
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/delivery/KafkaAcknowledgementTestServlet.java
@@ -59,7 +59,7 @@ public class KafkaAcknowledgementTestServlet extends AbstractKafkaTestServlet {
         CompletableFuture<Void> acknowledged = deliveryBean.sendMessage("test1");
 
         // Assert that message appears on topic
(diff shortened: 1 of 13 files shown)


dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/apps/kafka/MyDataDeserializer.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ *
+ */
+public class MyDataDeserializer implements Deserializer<MyData> {
+
+    /** {@inheritDoc} */
+    @Override
+    public MyData deserialize(String topic, byte[] data) {
+        String dataStr = new String(data);
```
Azquelt: I think you'll need to specify the character set here, otherwise the test may fail in cases where the platform encoding isn't compatible with UTF-8.

The Kafka String (de)serializer defaults to UTF-8 unless configured otherwise.

tevans78: done

tevans78:
#build

Issue #8277, opened 2019-07-18
login process is carried out for unprotected resources even TAI does not intercepts a request


toshiyamamoto:
When invokeUnprotectedURI is sett as true, if none of TAIs intercepts a request for accessing an unprotected resource, the application defined login process is carried out.
In this case, since the target resource is unprotected, the request should be processed without the login process.

PR #8279, opened 2019-07-18
fix unprotectedURI issue.

toshiyamamoto:
The code was modified not to invoke an application login process if invokeUnprotectedURI is set as true, but none of TAIs processed the request.


toshiyamamoto:
fixes #8277

toshiyamamoto:
#build

toshiyamamoto:
#build

Issue #8280, opened 2019-07-18
OIDC UI: value is undefined in utils.formatString


dmorgant:
When bringing up an OIDC UI endpoint (eg oidc/endpoint/OP/clientManagement), utils.formatString results in "TypeError: value is undefined" when a message is missing in a supported language.

If a message is missing in a supported language, it should default back to the message in the English locale. If the message is also missing in the English locale, it should return en empty string.

PR #8286, opened 2019-07-18
For non-english languages, merge with en message file

dmorgant:
If a message is missing in a supported language, it should default back to the message in the English locale. To do this, if we are retrieving messages from a non-English file, then we merge those messages with the messages from the English messages file so that any strings not in the translated file would be in the merged file from the strings in the English file.


dmorgant:
fixes #8280

dmorgant:
#build

dmorgant:
Testing....
Currently, the latest messages files have a few messages in them that have not yet been translated.  So, I brought up the new clientManagement, usersTokenManagement, and personalTokenManagement applications in another language (like French) and made sure that the strings that have not yet been translated appeared in English, thereby preventing an exception to occur on the UI due to a missing message.

PR #8300, opened 2019-07-22
Uncouple some execution/instance operations to prevent deadlocks

dmbelina:
This change is being made to prevent the possibility of two deadlocks:

1) In BaseStepControllerImpl.updateStepExecution, the current code attempts to update the step execution as well as the lastUpdated time in the associated job execution.  In updateStepExecution, there is: TranRequest<StepThreadExecutionEntity>(em) and that's where we call updateStepExecutionStatusTimeStampsUserDataAndMetrics).  In the reported issue, we see the indexes being locked for both tables.

It is not important that this be viewed atomically/consistently, so the lastUpdated time update can be done at the same stage in the lifecycle, just not in the same transaction.

2) During the use of the getJobExecutionsRunning API, the database query makes use of two tables for the execution and instance.  It will locate the instances with the specified jobName, and uses that to locate the executions with a "running" status (started, starting or stopping).  In the reported issue, we see the indexes being locked for both tables.

The proposed fix performs the two aformentioned operations in seperate transactions.

### DIFF
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/BaseStepControllerImpl.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/BaseStepControllerImpl.java
@@ -437,7 +437,8 @@ public abstract class BaseStepControllerImpl implements IExecutionElementControl
     // Will be overridden for partitioned steps by special method which
     // aggregates partition-level metrics
     protected void updateStepExecution() {
-        getPersistenceManagerService().updateStepExecution(runtimeStepExecution);
+        StepThreadExecutionEntity stepThreadExecutionEntity = getPersistenceManagerService().updateStepExecution(runtimeStepExecution);
+        stepThreadExecutionEntity.getJobExecution().setLastUpdatedTime(runtimeStepExecution.getLastUpdatedTime());
     }
 
     protected void markJobAndStepForFailure() {
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/persistence/jpa/JobInstanceEntity.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/persistence/jpa/JobInstanceEntity.java
@@ -41,6 +41,7 @@ import com.ibm.jbatch.container.ws.InstanceState;
 import com.ibm.jbatch.container.ws.WSJobInstance;
 
 @NamedQueries({
+                @NamedQuery(name = JobInstanceEntity.GET_JOBINSTANCEIDS_BY_NAME_AND_STATUSES_QUERY, query = "SELECT i.instanceId FROM JobInstanceEntity i WHERE i.batchStatus IN :status AND i.jobName = :name"),
                 @NamedQuery(name = JobInstanceEntity.GET_JOBINSTANCES_SORT_BY_CREATETIME_FIND_ALL_QUERY, query = "SELECT i FROM JobInstanceEntity i ORDER BY i.createTime DESC"),
                 @NamedQuery(name = JobInstanceEntity.GET_JOBINSTANCES_SORT_BY_CREATETIME_FIND_BY_SUBMITTER_QUERY, query = "SELECT i FROM JobInstanceEntity i WHERE i.submitter = :submitter ORDER BY i.createTime DESC"),
                 @NamedQuery(name = JobInstanceEntity.GET_JOB_NAMES_SET_QUERY, query = "SELECT DISTINCT i.jobName FROM JobInstanceEntity i"),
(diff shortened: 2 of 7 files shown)


scottkurz:
> This change is being made to prevent the possibility of two deadlocks:
> ...

Hmm... I can see how the changes "decouple" DB access differently from the previous access, but I don't really have any insight into what the original problem was.   Not sure how worth it it would be to write that up.. but it might help to have it "on record" if you can find the time.

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/BaseStepControllerImpl.java
```diff
@@ -437,7 +437,8 @@ protected void markStepForFailure() {
     // Will be overridden for partitioned steps by special method which
     // aggregates partition-level metrics
     protected void updateStepExecution() {
-        getPersistenceManagerService().updateStepExecution(runtimeStepExecution);
+        StepThreadExecutionEntity stepThreadExecutionEntity = getPersistenceManagerService().updateStepExecution(runtimeStepExecution);
+        stepThreadExecutionEntity.getJobExecution().setLastUpdatedTime(runtimeStepExecution.getLastUpdatedTime());
     }
```
scottkurz: Seems like we should still mark job lastUpdated for partition threads  (since the previous setLastUpdatedTime marked it for each.  But this comment seems to say it won't be..

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/persistence/jpa/JobInstanceEntity.java
```diff
@@ -41,6 +41,7 @@
 import com.ibm.jbatch.container.ws.WSJobInstance;
 
 @NamedQueries({
+                @NamedQuery(name = JobInstanceEntity.GET_JOBINSTANCEIDS_BY_NAME_AND_STATUSES_QUERY, query = "SELECT i.instanceId FROM JobInstanceEntity i WHERE i.batchStatus IN :status AND i.jobName = :name"),
                 @NamedQuery(name = JobInstanceEntity.GET_JOBINSTANCES_SORT_BY_CREATETIME_FIND_ALL_QUERY, query = "SELECT i FROM JobInstanceEntity i ORDER BY i.createTime DESC"),
```
scottkurz: So we switched from querying JobExecution batchStatus to  JobInstance batchStatus.   I think that's OK.. I think the one is just a redudantly-added extra column to avoid querying in certain cases.  Right?

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/services/impl/JPAPersistenceManagerImpl.java
```diff
@@ -1301,6 +1300,42 @@ public JobExecutionEntity call() throws Exception {
...
+
+                    List<Long> result = new ArrayList<Long>();
+
+                    for (int i = 0; i < runningInstances.size(); i++) {
+                        long instanceId = runningInstances.get(i);
+                        TypedQuery<Long> query = em.createNamedQuery(JobExecutionEntity.GET_JOB_EXECUTIONIDS_BY_JOB_INST_ID,
+                                                                     Long.class);
+                        query.setParameter("instanceId", instanceId);
+                        result.addAll(query.getResultList());
+                    }
```
scottkurz: Should we try to do this in one query... search for ids "IN" the set instead of 1 query per loop iteration?   
Also previously we were sorting by createTime `ORDER BY e.createTime DESC`... are we still getting that result or should we do a later sort to ensure this?

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/BaseStepControllerImpl.java
dmbelina: You're right here in that I missed the override for the partition case and I'll make that revision.  But even in that case, it looks like it is only marking the lastUpdated for the top level (see updateStepExecutionWithPartitionAggregate in JPAPersistenceManagerImpl)  The only thing it does with the partitions is to aggregate the metrics (which the comment indicates)

scottkurz: Hmm... not following your comment 100% but it made me see a mistake in my comment.   I was wrong to say your code doesn't update lastUpdated for partition _threads_... the missing gap is the top-level thread in partitioned steps.  That's the only gap that needs covering.

The `BaseStepControllerImpl.updateStepExecution()` method here runs on the partitioned threads (just not on the top-level thread of a partitioned step).. I believe.

Are we on the same page now?

dmbelina: I was referring to PartitionedStepControllerImpl.updateStepExecution() which I had missed.  It would need the same treatment as BaseStepControllerImpl. (They both use the same AbstractPersistenceManager.updateStepExecutionStatusTimeStampsUserDataAndMetrics where the set lastUpdated was moved out of)

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/persistence/jpa/JobInstanceEntity.java
dmbelina: That's correct

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/services/impl/JPAPersistenceManagerImpl.java
dmbelina: Revised this query per your suggestion in the latest commit

dmbelina:
Fixes #8532

PR #8301, opened 2019-07-22
Increase wait time on jobwaiter

dmbelina:
Update this test to increase the amount of time it is allowed to wait before failing


dmbelina:
Tested via local and personal build (was-liberty20190720-0425)

PR #8305, opened 2019-07-22
Batch remote partition restart

cgianfra:


dmbelina:
Noting that this needs the release bug label also

Issue #8289, opened 2019-07-18
personalTokenManagement endpoint: disable 'Regenerate' when flags not set


dmorgant:
For the OIDC personalTokenManagement endpoint, if the appPasswordAllowed or appTokenAllowed flags are not set on the OP, then grey out the 'Regenerate' button for app-passwords or app-tokens depending on which flags are not set.

PR #8313, opened 2019-07-23
Disable Regenerate button on personalTokenManagement when flag not set

dmorgant:
If the OAuth Provider does not have the appPasswordAllowed or appTokenAllowed flag set, then a user accessing this provider should not be able to create app-passwords or app-tokens (respectively).  On the personalTokenManagement application, also disable the 'Regenerate' button in the table since this action will also fail if the respective flag is not set on the OP.

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.server/resources/WEB-CONTENT/accountManager/js/table.js
+++ b/dev/com.ibm.ws.security.openidconnect.server/resources/WEB-CONTENT/accountManager/js/table.js
@@ -108,7 +108,15 @@ var table = (function() {
 
         // Action buttons
         var regenerateAriaLabel = utils.formatString(messages.REGENERATE_ARIA, [authData.authType, authData.name]);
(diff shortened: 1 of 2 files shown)


dmorgant:
Fixes #8289

dmorgant:
#build

dmorgant:
Tests:
- Set both appPasswordAllowed and appTokenAllowed flags on OP.  Logged into personalTokenManagement and created both app-passwords and app-tokens.   'Regenerate' was enabled for both types.

- Edited OP so that appPasswordAllowed was false.  Logged into personalTokenManagement and 'Regenerate' was greyed out for app-passwords in the table.

- Edited OP so that appTokenAllowed was false.  Logged into personalTokenManagement and 'Regenerate' was greyed out for app-passwords (because of step 2) and app-tokens.

- Edited OP so that appPasswordAllowed was true.  Logged into personalTokenManagement and 'Regenerate was only greyed out for app-tokens (because of step 3).  'Regenerate' was enabled for app-passwords.

- Edited OP so that appTokensAllowed was true.  Logged into personalTokenManagement and 'Regenerate' was enabled for all app-passwords and app-tokens.

Issue #8284, opened 2019-07-18
OIDC deleting an app password/token belonging to a different user not working


lavenal:
There is a problem with the following OIDC app password/token scenarios with two users, a testuser with clientManager role and a tokeuser with tokenManager role.
- login as testuser to personalTokenManagement endpoint
- create a token/password in personalTokenManagement tool, eg testpw1
- logout
- login as tokenuser to usersTokenManagement endpoint
- search for testuser
- delete the token/password previously created (testpw1) by clicking on the delete link for that particular token/password
  - confirm the delete
  - the delete restapi returned 200 if you inspect the network tab for the call
- search for testuser again
   - the deleted token/password (testpw1) is still there
- logout
- to confirm, go to personalTokenManagement endpoint and login as testuser
- the deleted token/password (testpw1) is still there
- perform delete by clicking on the delete link for that particular token/password (testpw1)
  - confirm the delete
  - this time the delete works

dmorgant:
The API used from usersTokenManagement that a user with tokenManager role uses to delete a single app-token belonging to another user is:

https://localhost:29443/oidc/endpoint/OP/app-tokens/V0HZTHfhsdIBOwl4jkaaWgK0kFKpJoT58XXCiKOM

ayoho:
Copying in part of the conversation from https://ibm-cloud.slack.com/archives/CGZTMT1LG/p1563482672168700 here for completeness:

It would seem this is actually working as designed, but the WAD is a little unclear about how this scenario is intended to work. The WAD documents two particular scenarios for an “admin-service scenario” (token manager scenario):
```
/oidc/endpoint/<provider_name>/app-passwords/<app_id>               DELETE: delete one app-password
/oidc/endpoint/<provider_name>/app-passwords/?user_id=<user name>   DELETE: delete passwords matched the query
```
The use case documented in this issue is falling under that first one - `/oidc/endpoint/<provider_name>/app-passwords/<app_id>` - which shows that the `user_id` parameter isn’t required. However, the WAD also states that **“Note: if user_id is missing, user_id defaults to the user identified in access_token.“** That means since the token manager’s access token was sent in the DELETE request, the runtime will look for an app password for the _token manager_ to delete. That user, of course, doesn’t have any app passwords/tokens, so effectively nothing happens.

PR #8319, opened 2019-07-23
Pass userid for single token deletes in usersTokenManagement

dmorgant:
When a user with tokenManager role is logged in to the usersTokenManagement application they need to be able to delete app-passwords and app-tokens for other users.  In order for this to be successful, pass the user_id of the owner of the app-password or app-token to be deleted.


dmorgant:
Fixes #8284

dmorgant:
#build

PR #8326, opened 2019-07-23
Don't use translated label for values in selection lists or radio button groups.

dmorgant:
In the add/edit OAuth client dialog in the OIDC clientManagement application, the OIDC client registration attributes that are formatted as multi-selection dropdown lists or radio button groups are displayed with api defined values to select from.   The values are retrieved from the oidc/endpoint/OP/clientMetatype where they are defined with a 'value' and 'label'.  The 'label' is a translated value.  However, we don't want to show the 'label' since the 'value' is what is known to the API and defined in the OAuth client registration API documentation (see https://www.ibm.com/support/knowledgecenter/en/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/twlp_client_registration.html).   Therefore, update the code to show the 'value' on the screen, not the 'label' for these widgets.

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.server/resources/WEB-CONTENT/clientAdmin/js/clientInputDialog.js
+++ b/dev/com.ibm.ws.security.openidconnect.server/resources/WEB-CONTENT/clientAdmin/js/clientInputDialog.js
@@ -498,7 +498,7 @@ var clientInputDialog = (function() {
                 "               <div class='tool_multiSelect_menuItem_option'>" +
                 "                   <div class='tool_checkbox_wrapper'>" +
                 "                       <input id='" + optionId + "' class='tool_checkbox' type='checkbox' value='" + option.value + "'>" +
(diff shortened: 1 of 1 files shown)


dmorgant:
Fixes #8326

dmorgant:
#build

dmorgant:
Test: 
Bring up the clientManagement application with your browser set to another language (like French).  Select 'Add New' to register a new OAuth client.  Scroll to the following fields and verify that the selections for those fields are the keywords used in the API for those fields:
Grant types:  'implicit' and 'password' should not be translated.
Response types: 'code' and 'token' should not be translated.
Token endpoint authentication method: 'none' should not be translated.
Application type: 'native' and 'web' should not be translated.

dmorgant:
#build

dmorgant:
This is a UI-only fix with wording that appears on a dialog and should not have failed any tests.  Ran another build to verify.

PR #8327, opened 2019-07-23
Add hash value to xml to avoid cached file re-hashing

cbridgha:
Sorry for reformatting - please "ignore whitespace" under the settings button when comparing changes.

2 Changes here:

1) Avoid a file not found exception during the cached file copy by moving a finally block to flush and close the OutputStream after the serialization is completed

2) Avoid re-hashing the cached file DOM entries by storing the hash value in the xml file itself, and retrieving this value to compare the current  config.


cbridgha:
#build

cbridgha:
Agree the hashing could be improved, but this has always been how its done, so no need to change at this point

PR #8347, opened 2019-07-24
Reestablish key traversal after Regenerate

dmorgant:
During 'Regenerate' processing, the contents of the authorization's row in the table is replaced with new contents.   These new cells need to have their event handlers set for keyboard navigation through the table.  Invoke tableUtils.initTableKeyTraversing() to set this up.


dmorgant:
Fixes #8347

dmorgant:
#build

PR #8351, opened 2019-07-24
Permit restricted PostgreSQL operations

### LINKED ISSUE #8346: Liberty 19.0.0.7 Blocks *all* Large Object API functions for Postgres
The Postgres JDBC Driver allows for accessing the LargeObjectAPI calls to deal with storing large files with the `oid` type in Postgres.

As of Websphere Liberty 19.0.0.7 these apis have been blocked for use by Websphere.

Attempting to use them results in errors such as:
```
java.sql.SQLFeatureNotSupportedException: DSRA9130E: Operation is not permitted by the application server: getLargeObjectAPI
    at com.ibm.ws.rsadapter.jdbc.WSJdbcWrapper.invoke(WSJdbcWrapper.java:224)
    at com.ibm.ws.rsadapter.jdbc.WSJdbcConnection.invoke(WSJdbcConnection.java:4124)
    at 
```
My understanding, based on conversations with devs and after seeing some Tests written in this issue: https://github.com/OpenLiberty/open-liberty/issues/8118  Is that  the LargeObjectAPI with postgres can do things to the underlying Connection outside of WebSphere's Transaction manager.  The end result is that some operation can result in data-integrity issues.

However, not all uses of the api are unsafe and some are safe.    Asking that safe operations not be blocked. Specifically things like:
`conn.largeObjectApi.createLO()`
`conn.largeObjectApi.open(int,int)`  which is
```
  public LargeObject open(long oid, int mode) throws SQLException {
    return open(oid, mode, false);
  }
```
LargeObject..getOutputStream()
```
@throws SQLException if a database-access error occurs.
   */
  public OutputStream getOutputStream() throws SQLException {
    if (os == null) {
      os = new BlobOutputStream(this, 4096);
    }
    return os;
  }
```

aguibert:
Fixes #8346

### DIFF
--- a/dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/PostgreSQLHelper.java
+++ b/dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/PostgreSQLHelper.java
@@ -31,23 +31,52 @@ public class PostgreSQLHelper extends DatabaseHelper {
     
     private static final String PROP_DEFAULT_FETCH_SIZE = "DefaultFetchSize";
     private static final String PROP_PREPARE_THRESHOLD = "PrepareThreshold";
+    private static final String PROP_AUTOSAVE = "AutoSave";
 
     @SuppressWarnings("deprecation")
     private transient com.ibm.ejs.ras.TraceComponent jdbcTC = com.ibm.ejs.ras.Tr.register("com.ibm.ws.postgresql.logwriter", "WAS.database", null);
 
     private transient PrintWriter jdbcTraceWriter;
+    
+    private Class<?> Autosave;
+    private Class<?> LargeObjectManager;
+    private Class<?> BaseConnection;
 
-    PostgreSQLHelper(WSManagedConnectionFactoryImpl mcf) {
+    PostgreSQLHelper(WSManagedConnectionFactoryImpl mcf) throws ClassNotFoundException {
         super(mcf);
+    }
+    
+    /**
+     *  Proxy method for org.postgresql.PGConnection.getLargeObjectAPI()
+     *  During 19.0.0.7 this method was blocked because it closed the underlying connection on some code paths.
+     *  However, we later realized that by overriding this method and constructing the LargeObjectManager using
+     *  a proxied instance of this class, we can maintain control of the connection that the LargeObjectAPI uses.
+     */
+    public Object getLargeObjectAPI(Connection con) throws SQLException {
+        try {
+            if (LargeObjectManager == null || BaseConnection == null) {
+                LargeObjectManager = WSManagedConnectionFactoryImpl.priv.loadClass(mcf.jdbcDriverLoader, "org.postgresql.largeobject.LargeObjectManager");
+                BaseConnection = WSManagedConnectionFactoryImpl.priv.loadClass(mcf.jdbcDriverLoader, "org.postgresql.core.BaseConnection");
+            }
+            
+            Object selfProxy = con.unwrap(BaseConnection);
+            return LargeObjectManager.getConstructor(BaseConnection).newInstance(selfProxy);
+        } catch (Exception e) {
+            throw new SQLException(e);
+        }
     }
 
     @Override
     public Map<String, Object> cacheVendorConnectionProps(Connection connImpl) throws SQLException {
         try {
+            if (Autosave == null)
+                Autosave = WSManagedConnectionFactoryImpl.priv.loadClass(mcf.jdbcDriverLoader, "org.postgresql.jdbc.AutoSave");
+            
             Class<?> PGConnection = connImpl.getClass();
             Map<String, Object> defaultProps = new HashMap<String, Object>(2);
             defaultProps.put(PROP_DEFAULT_FETCH_SIZE, PGConnection.getMethod("getDefaultFetchSize").invoke(connImpl));
             defaultProps.put(PROP_PREPARE_THRESHOLD, PGConnection.getMethod("getPrepareThreshold").invoke(connImpl));
+            defaultProps.put(PROP_AUTOSAVE, PGConnection.getMethod("getAutosave").invoke(connImpl));
             if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
                 Tr.debug(tc, "Got initial PGConnection vendor-specific property defaults: " + defaultProps);
             return defaultProps;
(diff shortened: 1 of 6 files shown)


aguibert:
#build

dev/com.ibm.ws.jdbc_fat_postgresql/test-applications/postgresqlApp/src/jdbc/fat/postgresql/web/PostgreSQLTestServlet.java
```diff
@@ -449,22 +450,51 @@ public void testPropertyCleanup_postgre_prepareThreshold() throws Exception {
...
+    public void testPropertyCleanup_postgre_autoSave() throws Exception {
+        // Use a DataSource with maxPoolSize=1 and shareable=false so that we always reuse the same underlying connection (if possible)
+        DataSource ds = InitialContext.doLookup("jdbc/postgres/maxPoolSize1");
+
+        try (Connection con = ds.getConnection()) {
+            PGConnection pgCon = con.unwrap(PGConnection.class);
+            assertEquals("Initial value of connection AutoSave was not correct", AutoSave.NEVER, pgCon.getAutosave());
+
+            pgCon.setAutosave(AutoSave.ALWAYS);
+            assertEquals("Connection did not honor setAutoSave(ALWYAS) call", AutoSave.ALWAYS, pgCon.getAutosave());
```
frowe: misspelled ALWAYS constant in error msg

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/PostgreSQLHelper.java
```diff
@@ -31,23 +31,52 @@
...
         super(mcf);
     }
+    
+    /**
+     *  Proxy method for org.postgresql.PGConnection.getLargeObjectAPI()
+     *  During 19.0.0.7 this method was blocked because it closed the underlying connection on some code paths.
+     *  However, we later realized that by overriding this method and constructing the LargeObjectManager using
+     *  a proxied instance of this class, we can maintain control of the connection that the LargeObjectAPI uses.
+     */
+    public Object getLargeObjectAPI(Connection con) throws SQLException {
```
frowe: The postgres doc says "For Postgres, you must access Large Objects within an SQL transaction. ", can we enforce that as part of our proxying.?

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/jdbc/WSJdbcProxyMethod.java
```diff
@@ -211,9 +211,10 @@
         //  oracle.jdbc.internal.OracleConnection.unwrap()
         
         //// PostgreSQL methods
-        unsafeMethods.add("getLargeObjectAPI"); // org.postgresql.PGConnection.getLargeObjectAPI() -- LargeObject API is not allowed because it has operations that can commit transactions on the underlying PG connection
-        unsafeMethods.add("getFastpathAPI"); // org.postgresql.PGConnection.getFastpathAPI() -- this is deprecated anyway on the PG API
-        unsafeMethods.add("setAutosave"); // org.postgresql.PGConnection.setAutosave(AutoSave)
+        // These methods were briefly blocked during 19.0.0.7
```
frowe: It's not necessary to create source control history using commented out code, remove.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/jdbc/WSJdbcProxyMethod.java
```diff
@@ -211,9 +211,10 @@
         //  oracle.jdbc.internal.OracleConnection.unwrap()
         
         //// PostgreSQL methods
-        unsafeMethods.add("getLargeObjectAPI"); // org.postgresql.PGConnection.getLargeObjectAPI() -- LargeObject API is not allowed because it has operations that can commit transactions on the underlying PG connection
-        unsafeMethods.add("getFastpathAPI"); // org.postgresql.PGConnection.getFastpathAPI() -- this is deprecated anyway on the PG API
```
frowe: was fastpath blocked because it was deprecated or unsafe?  If the latter, how did we address its safety?

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/PostgreSQLHelper.java
```diff
@@ -31,23 +31,52 @@
...
     private static final String PROP_DEFAULT_FETCH_SIZE = "DefaultFetchSize";
     private static final String PROP_PREPARE_THRESHOLD = "PrepareThreshold";
+    private static final String PROP_AUTOSAVE = "AutoSave";
 
     @SuppressWarnings("deprecation")
     private transient com.ibm.ejs.ras.TraceComponent jdbcTC = com.ibm.ejs.ras.Tr.register("com.ibm.ws.postgresql.logwriter", "WAS.database", null);
 
     private transient PrintWriter jdbcTraceWriter;
+    
+    private Class<?> Autosave;
```
njr-11: should this be volatile based on how it is used (lazy init)?

aguibert: it could be, yes. Since the cost of initializing it isn't very significant and double-computing the value isn't any difference in behavior I hadn't used it here.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/jdbc/WSJdbcProxyMethod.java
aguibert: it was previously blocked because some code paths can perform queries on the underlying connection. After discussing that setAutosave, my understanding on what is safe vs. unsafe has relaxed a bit, and this operation seems fine to permit.

dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/PostgreSQLHelper.java
aguibert: The implementation of PostgreSQL enforces this on its own. For example, in `LargeObjectManager.createLO()` it verifies that `autoCommit == false` and blows up otherwise

PR #8367, opened 2019-07-25
Added delay before modifying files in test case

ellen-lau:
Fixes: #263198

#build

Added a small 4s delay after the last server request before attempting to modify file paths within the test case to follow correct practices. Done in parallel with a Windows File Locking fix.


ellen-lau:
Closing because another fix (#8578) addresses issue.

PR #8374, opened 2019-07-29
Make TimeoutStateImplTest more reliable

Azquelt:
Change from using a real `ScheduledExecutorService` and `Thread.sleep()` to
using a mocked `ScheduledExecutorService` to avoid very occasional timing
issues.

This also makes the test slightly faster.

Additionally, hide some of the unhelpful warnings across the StateImpl tests.


Azquelt:
#build

Azquelt:
#build

Issue #8325, opened 2019-07-23
Investigate spurious character showing up in OAuth metatype API response


ayoho:
When updating the OAuth client metatype API to support UTF-16 (#8234) it was noticed that a non-visible character may or may not have been added at the beginning of the API response body. While browsers still generally seem to be able to parse the response as valid JSON, viewing the output a certain way shows some unknown character before the first `[` character that starts the API output.

PR #8383, opened 2019-07-29
Fix double byte character encoding for OAuth client metatype API

ayoho:
Fixes the OAuth client metatype API to set the response header character encoding to UTF-8, but to write the API string itself as a UTF-16 string.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.openidconnect.server_fat,com.ibm.ws.security.openidconnect.server_fat.endpoint.clientregistration

ayoho:
Resolves #8325

PR #8413, opened 2019-07-30
Unzip to temp directory to avoid multiple updates

brenthdaniel:
There's a timing issue in the app manager FAT where we can end up restarting an application twice because we are unzipping a file directly into the apps directory.


brenthdaniel:
#build

PR #8415, opened 2019-07-30
New test case for MP Rest Client with FT Fallback class specified

andymc12:
New test case that confirms that MP Rest Client methods annotated with `@Fallback` that specifies a `FallbackHandler` class (as opposed to a method on the same client interface) works as expected.

This is just a new test case and is not a release bug.


andymc12:
#build

PR #8421, opened 2019-07-31
Issue #8420: Update the WCServletContainerInitializerExceptionTest

### LINKED ISSUE #8420: Update testExceptionThrowInSCIOnStartup FAT
1) Ensure we look for the proper message ID in the logs.
2) Add a comment with an example of what the Warning message looks like.
3) Correct the package in the javax.servlet.ServletContainerInitializer
4) Use `waitForStringInLog` vs `findStringsInLogs`, since the SCI might not always have been invoked by the time the test starts, this should prevent intermittent test failures.
5) Use `ShrinkHelper.exportDropinAppToServer` vs `ShrinkHelperexportToServer`
6) Remove the negative test that checks if the translation key is in the logs, the simple fact that we find the message id in the logs is enough to verify the key isn't present.

This test was originally added as part of Issue #1770

pnicolucci:
Fixes #8420


pnicolucci:
#build

pnicolucci:
#build

PR #8426, opened 2019-07-31
Issue #8425: Fix where the required roles for auditing is filled when…

### LINKED ISSUE #8425: RESTHandlerContainerImpl should not set required roles
RESTHanlderContainerImpl is setting required roles on the response when that should be done in DefaultAuthorizationHelper. Additionally, there is an INFO message in place of auditing that should not be INFO. Eventually, the message will get replaced by actual auditing.

jvanhill:
… using the DefaultAuthorizationHelper.

fixes #8425


jvanhill:
#build

Issue #8427, opened 2019-07-31
SseEventSinks are leaked in Broadcasters after closed


andymc12:
Currently, when an SseEventSink is registered with a Broadcaster, there is no way via the spec to "unregister" it.  So, when a sink has gone bad or is closed, the sink is still in the list of subscribed clients in the broadcaster.  This is inefficient and can cause issues trying to sent events to a closed sink.

This problem occurs in 19.0.0.7, but has been a problem since JAX-RS 2.1 was released.

andymc12:
fixed by #8428

PR #8428, opened 2019-07-31
Issue 8427: Remove SSE sinks from broadcasters when closed

andymc12:
This fix removes closed sinks from broadcaster rather than attempting to send events to them.  It includes a test case that uses deep reflection on the Liberty-internal impl class in order to verify that the sink was removed.  While this is fine for testing, it is not a good practice for user applications as this field is internal, private, and subject to change without notice.

This resolves issue #8427.


andymc12:
#build

PR #8455, opened 2019-08-02
Reaper arq test

bberry1675:
PR to run personal builds for branch


bberry1675:
#build

bberry1675:
#build

bberry1675:
#build

tbitonti:
This is a partial fix for RTC 258010.  Two fix updates and one diagnostic update were made:
(1) The reaper run thread detects when it would be put incorrectly in an indefinitely wait and changes this to a definite wait.  An error is emitted if this happens, since this is an expected state.
(2) The reaper startup has a new latch to ensure the reaper thread acquires the reaper lock before any other activity occurs.
(3) Additions were made to the reaper state and to the reaper introspector so to display the history of lock and wait activity.

bberry1675:
#build

tbitonti:
The specific problem which was discovered and which is fixed by this update is that the reaper thread, while started, was not first to have activity against the reaper state.  That created the possibility of reaper activity opening two archives before the reaper thread was actually run.  The reaper had been coded to have an indefinite wait on its first loop, based on the expectation that no reaper activity had yet occurred.  The expectation was broken by the pattern of two early opens, causing the reaper to enter a permanent indefinite wait.

bberry1675:
#build

bberry1675:
#build

bberry1675:
#build

bberry1675:
#build

PR #8484, opened 2019-08-05
Add AllowedFFDC and ExpectedFFDC annotation scanning to servlets

olendvcook:


olendvcook:
#build

olendvcook:
#build

olendvcook:
#build

olendvcook:
build hit known problem

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -631,6 +631,13 @@ private String getFFDCHeader(RemoteFile ffdcLogFile) throws Exception {
                 annotationListPerClass.add(exceptionClass);
             }
         }
+        ExpectedFFDC servletFFDC = m.getMethod().getDeclaringClass().getAnnotation(ExpectedFFDC.class);
```
aguibert: I don't think `@ExpectedFFDC` will ever be used on the class level. There would have to be a scenario where every test method in a class MUST produce the same exception, which I think is extremely unlikely. Instead of adding this processing here, perhaps we could instead modify `@ExpectedFFDC` to only be applicable to methods, so that if anyone tries to put it on a non-method element in the future they would get a compile failure

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -645,6 +652,13 @@ private String getFFDCHeader(RemoteFile ffdcLogFile) throws Exception {
                 annotationListPerClass.add(exceptionClass);
             }
         }
+        AllowedFFDC servletFFDC = m.getMethod().getDeclaringClass().getAnnotation(AllowedFFDC.class);
+        if (servletFFDC != null) {
```
aguibert: in cases where `@TestServlet` is not used, `getTestClass().getJavaClass()` will be == to `m.getMethod().getDeclaringClass()`, I think we should check for that case to avoid double-adding the allowed FFDC classes

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -637,22 +637,34 @@ private String getFFDCHeader(RemoteFile ffdcLogFile) throws Exception {
 
     public List<String> getAllowedFFDCAnnotationFromTest(FrameworkMethod m) {
```
aguibert: minor nit: since this method is only ever called from 1 spot within this test class, lets change visibility to private, and make it return `Set<String>` instead of converting from Set -> List at the end

olendvcook:
#build

PR #8487, opened 2019-08-05
claim the right to run a late timer task

njr-11:
Experiment with reserving a database entry for an amount of time to enforce the right of one instance to attempt to run a late timer task.

ID review is not needed. This only includes a translatable metatype file because builds are backed up and prerequisite commits are unable to merge..


njr-11:
#build

PR #8527, opened 2019-08-07
Remove servlet-3.0

fritze2:


tevans78:
Merging without PB since we can't release without it. Taking the risk.

PR #8528, opened 2019-08-07
Make PermissionManager's dependency on WsJar service non-optional

andymc12:
This resolves a possible regression that may have been introduced with parallel bundle start where the WsJar protocol handler is not registered before the `PermissionManager` attempts to use it.

### DIFF
--- a/dev/com.ibm.ws.classloading/bnd.bnd
+++ b/dev/com.ibm.ws.classloading/bnd.bnd
@@ -60,10 +60,10 @@ Service-Component: \
     permission=com.ibm.ws.classloading.java2sec.JavaPermissionsConfiguration; \
     classLoadingService=com.ibm.wsspi.classloading.ClassLoadingService; \
     wsjarURLStreamHandler='org.osgi.service.url.URLStreamHandlerService(url.handler.protocol=wsjar)'; \
-    dynamic:='permission,wsjarURLStreamHandler'; \
+    dynamic:='permission'; \
     multiple:='permission'; \
     greedy:='permission'; \
-    optional:='permission,wsjarURLStreamHandler'; \
+    optional:='permission'; \
     properties:="service.vendor=IBM"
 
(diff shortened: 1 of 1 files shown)


andymc12:
#build

andymc12:
#build #spawn.fullfat.buckets=com.ibm.ws.jpa_22_fat

andymc12:
Also making the dependency non-dynamic - new build:
#build #spawn.fullfat.buckets=com.ibm.ws.jpa_22_fat

tjwatson:
#build #spawn.fullfat.buckets=com.ibm.ws.jpa_spec10_injection_fat

tevans78:
PB is not quite finished but we really need this PR and the JPA buckets have passed. Shouldn't break anything else, taking the risk. Andy merged it.

tjwatson:
I took a closer look at this.  The wsjar handler service component `com.ibm.ws.artifact.url.internal.WSJarURLStreamHandler` is provided by bundle `com.ibm.ws.artifact.url` which has a start-phase of `CONTAINER_EARLY`.  `com.ibm.ws.classloading.java2sec.PermissionManager` service component has a reference dependency on the wsjar handler and it is provided by the bundle `com.ibm.ws.classloading` which has the default start-phase of `CONTAINER`.  So `com.ibm.ws.classloading` does start AFTER `com.ibm.ws.artifact.url` and everything should always work even with parallel bundle activation ... but no...

`com.ibm.ws.artifact.url.internal.WSJarURLStreamHandler` has a dependency reference on `com.ibm.ws.artifact.zip.cache.ZipCachingService` which is provided by `com.ibm.ws.artifact.zip` and that is also using the default start-phase of `CONTAINER` so the two bundles `com.ibm.ws.artifact.zip` and `com.ibm.ws.classloading` can be started in parallel which can result in `com.ibm.ws.classloading.java2sec.PermissionManager` getting activated before the wsjar handler is available which ultimately will result in errors.  Making the dependency static and mandatory is the correct fix.

PR #8543, opened 2019-08-09
Use jar protocol if wsjar protocol handler is not available

andymc12:
This fixes a bug uncovered by a new feature (parallel bundle activation) - as such, it has not been made generally available yet, so it is not a release bug.

It works around a problem where the class loading PermissionManager attempts to use the wsjar protocol before the wsjar protocol handler has been registered.

### DIFF
--- a/dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/java2sec/PermissionManager.java
+++ b/dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/java2sec/PermissionManager.java
@@ -341,8 +341,13 @@ public class PermissionManager implements PermissionsCombiner {
         try {
             codeSource = new CodeSource(new URL("wsjar:file:/" + codeBase), certs);
         } catch (MalformedURLException e) {
-            if (tc.isDebugEnabled()) {
-                Tr.debug(tc, "Unable to create code source for protection domain");
+            try {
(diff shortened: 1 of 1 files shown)


andymc12:
#build - just for sanity - also will be running a Windows-specific Full-FAT build to re-test on a platform where we see the failure most often.

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/java2sec/PermissionManager.java
```diff
@@ -339,7 +339,11 @@ private CodeSource createCodeSource(String codeBase) {
         Certificate[] certs = null;
         CodeSource codeSource = null;
         try {
-            codeSource = new CodeSource(new URL("wsjar:file:/" + codeBase), certs);
+            if (wsjarUrlStreamHandlerAvailable) {
+                codeSource = new CodeSource(new URL("wsjar:file:/" + codeBase), certs);
+            } else {
+                codeSource = new CodeSource(new URL("jar:file:/" + codeBase), certs);
```
tjwatson: fallback should happen in the catch block for `MalformedURLException` I think without checking for the .`wsjarUrlStreamHandlerAvailable` flag.

andymc12:
#build

andymc12:
Closing without merging in favor of PR #8544

PR #8566, opened 2019-08-12
Exclude non-kafka properties

### LINKED ISSUE #8451: RM: Remove spurious Kafka warnings
```
[8/1/19 13:02:48:182 EDT] 0000004c SystemErr                                                    R [Default Executor-thread-16] WARN org.apache.kafka.clients.consumer.ConsumerConfig - The configuration 'channel-name' was supplied but isn't a known config.
[8/1/19 13:02:48:182 EDT] 0000004c SystemErr                                                    R [Default Executor-thread-16] WARN org.apache.kafka.clients.consumer.ConsumerConfig - The configuration 'connector' was supplied but isn't a known config.
[8/1/19 13:02:48:183 EDT] 0000004c SystemErr                                                    R [Default Executor-thread-16] WARN org.apache.kafka.clients.consumer.ConsumerConfig - The configuration 'topics' was supplied but isn't a known config.
```

The Kafka connector causes warnings like this to be emitted. This happens because we pass all config options through to the Kafka client, rather than stripping out the ones that are meant for the connector.

tevans78:
#build 
fixes #8451


tevans78:
GraphQL failures in the PB are unrelated to this change.

Issue #8559, opened 2019-08-12
invalidationTimeout unclear what unit of time will be used if none is specified


zhang-ca:
In the Knowledge Center we see the following:

HTTP Session (httpSession)
https://www.ibm.com/support/knowledgecenter/SSAW57_liberty/com.ibm.websphere.liberty.autogen.nd.doc/ae/rwlp_config_httpSession.html

Name: invalidationTimeout
Type: A period of time with second precision
Default: 30m
Description: Amount of time a session can go unused before it is no longer valid. Specify a positive integer followed by a unit of time, which can be hours (h), minutes (m), or seconds (s). For example, specify 30 seconds as 30s. You can include multiple values in a single entry. For example, 1m30s is equivalent to 90 seconds.

 It is unclear from the URL what unit of time will be used if none is specified (seconds).  Change be made to this page to reflect that seconds will be used if no time is specified

zhang-ca:
Change Description to:

Amount of time a session can go unused before it is no longer valid, **in seconds if unit of time isn't specified.** Specify a positive integer followed by a unit of time, which can be hours (h), minutes (m), or seconds (s). For example, specify 30 seconds as 30s. You can include multiple values in a single entry. For example, 1m30s is equivalent to 90 seconds.

PR #8569, opened 2019-08-12
Update invalidationTimeout Description

zhang-ca:
Name: invalidationTimeout
Type: A period of time with second precision
Default: 30m
Description: Amount of time a session can go unused before it is no longer valid**, in seconds if unit of time isn't specified**. Specify a positive integer followed by a unit of time, which can be hours (h), minutes (m), or seconds (s). For example, specify 30 seconds as 30s. You can include multiple values in a single entry. For example, 1m30s is equivalent to 90 seconds.


zhang-ca:
fixes #8559

zhang-ca:
for #8559

zhang-ca:
#build

PR #8617, opened 2019-08-15
Have profileManager check if the allowOpifRepoisDown field is set

### LINKED ISSUE #8169: ProfileManager.getImpl call ignores realm allowOpIfRepoDown setting
The ProfileManager.get(...) method is ignoring the allowOpIfRepoDown setting for the realm. It should also take into account the allowOpIfRepoDown setting in the realm configuration. This may lead to errors when repositories are down when the errors should be ignored.

agiridharan:
fixes #8169 

Currently the getImpl and searchImpl methods in profileManager are defaulting to false rather than checking if the allowOpifRepoIsDown field is set. This could lead to some problems is the repositories are unavailable


agiridharan:
#build

agiridharan:
#build

agiridharan:
#build

PR #8618, opened 2019-08-15
disabled test of scheduling timer to different instance

njr-11:
This test case will be disabled until a future point when product code is added to make it work.
The test involves scheduling a timer on an instance where execution is disabled and expecting it to run on another instance that is configured to be able to run its own timers and timers from other instances.


njr-11:
#build

njr-11:
#build

Issue #8152, opened 2019-07-08
TAI negotiateValidateandEstablishTrust called twice during authentication.


dburdine:
Customers TAI negotiateValidateandEstablishTrust() is called twice during authentication processing. There is not necessarily an error, just redundant.

PR #8626, opened 2019-08-16
8152-TAICalledTwice COMMIT1, 2, 3

dburdine:
PR for issue 8152. TAI is called twice for unprotected resource.


dburdine:
issue #8152

brad-gorman:
For what it's worth - I tried this in a CICS fix and it only called my TAIs once. :+1:

dburdine:
#build

PR #8633, opened 2019-08-16
Set com.ibm.ws.app.manager to CONTAINER_EARLY start-phase

tjwatson:
The component com.ibm.ws.app.manager.internal.ApplicationConfigurator
has a load of synchronization in its methods that are called by
SCR whild holding the SCR implementation is holding the component
state change lock.  This can lead to certain deadlock which is
eventually broken out of, but results in the OSGi service
registry eventually returning null when other components are
trying to get the services provided by the ApplicationConfigurator
component.  To reduce this posibility the start-phase of the
com.ibm.ws.app.manager is being moved up so that this component
is fully activated (it is immediate) before the CONTAINER
start-phase is entered.


tjwatson:
#build

tkburroughs:
@tjwatson any reason this hasn't been merged yet?  The problem continues to occur.

tjwatson:
IIRC it is because simply moving the bundle up start-phase did not really solve the issue because the component in question has references to other components that are at a higher start-phase so the component will not get  activated until later still.

tjwatson:
Abandoning this PR

PR #8658, opened 2019-08-19
Set max for parallel bundle activation to 4

tjwatson:
Performance measurements show no clear evidence that
using more than 4 threads to parallel activate bundles
gives any additional boost to the startup performance of
liberty.  In many cases it seems to increase the contention
on resources that cause the startup to use more CPU.

Limiting the max to 4 for parallel bundle activation


tjwatson:
#build

PR #8663, opened 2019-08-19
add a test (currently disabled) for cleaning up old entries

njr-11:
Write a test case that inserts some expired heartbeat entries and expects the detection logic (to be added under feature) to clean them up.


njr-11:
#build

PR #8671, opened 2019-08-20
Only allow single incoming topic and rename config property to topic

### LINKED ISSUE #8670: MP: Only allow connector to consume from one topic
So changing the param from `topics` to `topic`. Issue #8604 will also default the value to the channel name.

tevans78:
#build 
fixes #8670


Azquelt:
Conflicts resolved.

#build

PR #8675, opened 2019-08-20
remove character that breaks nls

brutif:
Removing at request of Chris P., this is causing the translation processing to fail.


brutif:
#build

PR #8677, opened 2019-08-20
Use extra single quote for nls

dmbelina:
When using replacement variables, single quotes need to be coded as 2 consecutive single quotes

Just need reviews as a formality.


andrea-pichler:
Requested message review completed by L2

PR #8678, opened 2019-08-20
Allow auto-commit to be enabled

### LINKED ISSUE #8654: RM: allow kafka connector to work when messages are not acknowledged
It's valid to configure `@Acknowledgement(NONE)` for a processor or subscriber in the reactive messaging application to indicate that any `Message` passed to the processor or subscriber will not be acknowledged.

~~The Kafka connector currently assumes that all messages will be acknowledged and that it should commit offsets. We need to provide a way to remove that assumption in the case that:
1) The user does not wish to use acknowledgements
2) The user wishes to have the kafka client acknowledge all records immediately using `auto.commit.enable = true`~~

The intention in the spec is that `@Acknowledgement(NONE)` should only be used when multiple subscribers are connected to a single publisher and all receive the same message (which we do not support).

We will still support the user setting `auto.commit.enable=true` and ignore acknowledgements if they do so.

Azquelt:
Allow enable.auto.commit to be set on the KafkaIncomingConnector.

This allows the user to get away with not acknowledging messages (although we think the intention of the spec is to require all messages to be acknowledged).

Also default the limit of in-flight unacked messages to a more sensible value.

Fixes #8654


Azquelt:
#build

PR #8681, opened 2019-08-20
Adding Fast Parse of existing xml

cbridgha:
Adding a fast parse of the plugin-config.xml to only read the root node containing the hash value needed to verify if anything has changed since last cached.

This should help reduce startup time of warm-start


cbridgha:
#build

cbridgha:
#build

cbridgha:
#build

PR #8687, opened 2019-08-20
Update line-height to include font size

dmorgant:
The line-height css setting for the .tool_modal_body_info_item entry in common/css/tool.css was not sufficient to hold the font-size.   Increase the line-height to hold the font size.


dmorgant:
This issue fixes TVT defect reported in closed liberty, 18421.

dmorgant:
#build

dmorgant:
To test:   The field was truncating Chinese characters on the dialog panel.   To test, I set the browser to display in Chinese and I opened the dialog to see that the characters were no longer truncated.

PR #8698, opened 2019-08-21
Update 'th' properties to word break on word boundary for OIDC application tables.

dmorgant:
In Polish and German, the column label "Issued On" is displayed with mid-word line break in German and Polish. Line breaks should not be on word boundaries only.

Update the css for 'th' (table headers) to set word-break to break-word instead of break-all.


dmorgant:
fixes CL issue: 18243

dmorgant:
#build

dmorgant:
Test:
The table column headings for the Personal Token Management application include two that have more than one word:  "Issued On" and "Expires On".   Manually tested that when you shrink the width of the browser window the column heading will wrap the word 'On' in these headings to a line below instead of wrapping the 'n' by itself.

Also tried this out in Polish and German where the issue was first reported.  The heading wrapped on the word boundary.

PR #8701, opened 2019-08-21
Add quotes around button values so multiple words don't get separated

dmorgant:
For the Input elements representing the buttons in the html for the tables on the OIDC applications set the value attribute to the translated string.  The value attribute value was not being quoted and therefore was not being added to the input element correctly.  So, added quotes to the value elements in the DOM.


dmorgant:
fixes CL issue: 18247

dmorgant:
#build

dmorgant:
Test:
Bringing up the OIDC applications in Spanish, we can see that the 'Regenerate' button which is translated to a phrase of multiple words correctly shows ALL the words in the column where the button is located instead of just the first one.

lavenal:
Denise, since you take care of the "delete" button value in usersTokenManagement, wondering whether you want to take care of the "delete" and "edit" button value in clientManagement.

dmorgant:
Great news......In the client management app we use \<button> elements, instead of \<input type=button> elements.   So the problem with the space does not exist.   Here's a picture....

image

PR #8713, opened 2019-08-21
Define activate(ComponentContext) methods

tjwatson:
Felix SCR has an optimization for searching activate
methods that take a single argument of type
ComponentContext.  All other activate methods require
a more exhaustive search of all the class methods to
find the appropriate activate method. Defining
activate(ComponentContext) methods avoids that.


tjwatson:
#build

tjwatson:
#build

PR #8717, opened 2019-08-22
Increase fault tolerance test timeout

Azquelt:
Increase the timeout values for these async tests. This does not affect
the correctness of any of the tests because they all use latches to
enforce ordering, it just increases the time that the test will wait for
a correct response before failing the test.

Increasing these timeouts should reduce the number of spurious failures
which occur in CI


Azquelt:
#build

PR #8725, opened 2019-08-22
Exclude more FT 1.0/1.1 tests with timing issues

Azquelt:
These tests usually pass but fail too regularly in the build.


Azquelt:
#build

Azquelt:
#build

PR #8727, opened 2019-08-22
Handle taskNames that may be options with arguments

fritze2:
When using the --tests option, the build takes a long time to configure all projects because it considers --tests and it's argument a project.

`./gradlew com.ibm.ws.security.oauth:test`
Duration: 68s

`./gradlew com.ibm.ws.security.oauth:test --tests "com.ibm.oauth.core.test.CustomConfigValueTest"`
Duration: 76s

After these changes --tests avoids configuration of unnecessary projects.
`./gradlew com.ibm.ws.security.oauth:test --tests "com.ibm.oauth.core.test.CustomConfigValueTest"`
Duration: 50s


fritze2:
Bypassing a PB since those never run project specific gradle calls and this change would not get tested there.

PR #8729, opened 2019-08-22
Fix bad version on export packages

neuwerk:
CXF API bundle has invalid syntax on the `Export-Package` header in `overrides.bnd` file.


neuwerk:
#build

PR #8730, opened 2019-08-22
Fix bad versioning of exports in CXF RT-Frontend-JAXWS

neuwerk:
CXF Frontend JAXWS bundle has invalid syntax on the Export-Package header in overrides.bnd file.


neuwerk:
#build

PR #8746, opened 2019-08-23
reject codepath where user omits the config element parameter but specifies uid

njr-11:
@frowe was doing some creative testing and discovered a bug where an erroneous result (list of single element) was being returned when the user invokes a configuration-based REST endpoint with a uid but without the config element path parameter (which is required when uid is present).

This pull adds detection and forces this situation to be a proper error path.


njr-11:
#build

njr-11:
#build

frowe:
#build
Although it appears the above failure is unrelated this PR, resubmitting to verify....

PR #8749, opened 2019-08-23
Prefer non-system bundle when finding perm classes

andymc12:
This should resolve an issue (not yet reported) when running the Jakarta EE CTS suite.  Java 2 security permission classes (specifically `javax.xml.ws.WebServicePermission`) are loaded from multiple locations in Java 8 - the JDK and a Liberty bundle.

The reason for this is that Oracle removed the JAX-WS APIs from Java 9+ JDKs, so the Liberty team packaged those APIs in a bundle so that Liberty would continue to operate with Java 8 and Java 9+ JDKs.

This issue occurs because the PermissionManager looks up the exporting bundle for the given permission - the OSGi framework returns a collection of bundles that exports the WebServicePermission, but the PermissionManager just takes the first bundle in the collection, which is the system bundle (that exports JDK classes) in the failing case.  Then when the permission check is made, it uses the WebServicePermission from the Liberty bundle and the AccessController fails because the permission class was loaded by a different classloader.

This fix changes the PermissionManager so that if multiple bundles export the same permission class, it will prefer bundles _other_ than the system bundle.


andymc12:
#build

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/java2sec/PermissionManager.java
```diff
@@ -491,24 +485,21 @@ public String getNamespace() {
...
-    private BundleWiring getBundleWiring(BundleCapability bundleCapability) {
-        BundleWiring providerBundleWiring = null;
-
-        if (bundleCapability != null) {
-            providerBundleWiring = bundleCapability.getRevision().getWiring();
+    private BundleWiring getBundleWiring(Collection<BundleCapability> bundleCapabilities) {
+        BundleCapability bundleCapability = null;
+        for (BundleCapability bc : bundleCapabilities) {
+            if (bundleCapability != null && 
+                Constants.SYSTEM_BUNDLE_LOCATION.equals(bc.getRevision().getBundle().getLocation())) {
```
tjwatson: Should not use getLocation here because it requires a permission check.  Instead use `getBundleId() == 0`

PR #8769, opened 2019-08-28
expose internal clientid and secret metatype attribs

brutif:


brutif:
The build had a timeout in openidConnect client tests, which can't be related to just exposing a metatype: 
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':com.ibm.ws.security.openidconnect.client:test'.
> Timeout has been exceeded
BUILD FAILED in 38m 42s

but I'm going to build again overnite to see if it clears. 

#build

brutif:
still seeing a timeout in an unrelated bucket but I think the pb is good.

PR #8789, opened 2019-08-28
Change event type names to to source names in jsonFields

Yushan-Lin:


Yushan-Lin:
Failing test is not related to changes

PR #8791, opened 2019-08-28
5914 expanded loose config

canterburymalcolm:
Packaging an app defined with a loose configuration now adds its resources in an expanded format as well as an archive. Initial tests can be run on multiple loose configurations.


dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/ProcessorUtils.java
```diff
@@ -261,7 +304,49 @@ private static File processArchive(String looseFileName, LooseConfig looseConfig
...
+
+        String entryPath = null;
+        if (archiveEntryPrefix.equalsIgnoreCase(PackageProcessor.PACKAGE_ARCHIVE_PREFIX) || !isUsr) {
+            entryPath = archiveEntryPrefix + BootstrapConstants.LOC_AREA_NAME_USR + looseFile.getAbsolutePath().substring(len);
+        } else {
+            entryPath = archiveEntryPrefix + looseFile.getAbsolutePath().substring(len);
+        }
+        entryPath = entryPath.replace("\\", "/");
+        entryPath = entryPath.replace("//", "/");
+        entryPath = entryPath.substring(0, entryPath.length() - 4); // trim the .xml
```
cbridgha: I know this was copied from existing code - but its ugly - what is being trimmed 4 chars?

canterburymalcolm:
The tests are now split into two files. 
- PackageLooseConfigParameterizedTest runs each test against a set of loose configs, so they're no longer run as one combined test. 

- PackageLooseConfigTest contains tests that would take too long to run multiple times, or just don't benefit from running against different loose configs.

canterburymalcolm:
#build

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/PackageProcessor.java
```diff
@@ -479,6 +479,8 @@ private void addLibExtractDir(List<ArchiveEntryConfig> entryConfigs) throws IOEx
          * Add the archived loose files
          */
         getReferencedResources(entryConfigs);
+        for (ArchiveEntryConfig e : entryConfigs)
+            System.out.println(e);
```
hlhoots: Assume this println was for debugging purposes. Should remove from the product code.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/ProcessorUtils.java
canterburymalcolm:
```suggestion
        return entryPath.replace("\\", "/").replaceAll("/(?=/)|.xml$", "");
```
This is the cleanest way I could think of. It also handles any size group of slashes now instead of just pairs of two and trims the .xml explicitly.

cbridgha:
Needs one more #build

Issue #8812, opened 2019-08-30
mark logging.jsonFields attribute as beta for 19.0.0.9


donbourne:
update metatype to mark jsonFields as beta for 19.0.0.9 (add ibm:beta="true" to the AD)

PR #8817, opened 2019-08-30
Update jsonFields in metatype.properties

Yushan-Lin:
#8812 
#build


Yushan-Lin:
Failed FAT test has nothing to do with changes. json.fields does not show up on schema

PR #8818, opened 2019-08-30
zWLM Classification for EJB Timers

mmerali96:
Issue #8855

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/TimerNpRunnable.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/TimerNpRunnable.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2018 IBM Corporation and others.
+ * Copyright (c) 2009, 2019 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -192,11 +192,25 @@ public abstract class TimerNpRunnable implements Runnable {
     }
 
     /**
+     * Getter for non-persistent EJB timer
+     */
+    public TimerNpImpl getNpTimer() {
+        return this.ivTimer;
+    }
+
+    /**
+     * Getter for method id
+     */
+    public int getMethodId() {
+        return ivMethodId;
+    }
+
+    /**
      * Schedules the first expiration for a non-persistent timer.
      *
      * @param expiration the point in time at which the first scheduled
-     *            expiration must occur; specified in milliseconds since
-     *            the standard base time (the epoch).
+     *                       expiration must occur; specified in milliseconds since
+     *                       the standard base time (the epoch).
      */
     protected abstract void schedule(long expiration);
 
@@ -204,8 +218,8 @@ public abstract class TimerNpRunnable implements Runnable {
      * Schedules recurring timers for the next scheduled expiration.
      *
      * @param nextExpiration the point in time at which the next scheduled
-     *            expiration must occur; specified in milliseconds since
-     *            the standard base time (the epoch).
+     *                           expiration must occur; specified in milliseconds since
+     *                           the standard base time (the epoch).
      */
     protected abstract void scheduleNext(long nextExpiration);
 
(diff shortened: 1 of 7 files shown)


andrea-pichler:

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/TimerNpRunnable.java
```diff
@@ -191,29 +191,47 @@ private void doWork() {
         }
     }
 
+    /**
```
tkburroughs: javadoc is wrong ... "Getter for bean id" is not correct; it is the timer implementation.

Also, if you have an @return there should be text after it to say what is returned.  For simple getters, the method description and @return description are often redundant, so you could have just one or the other.  An empty @return will produce a javadoc warning, so if present, there should be text after it.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/TimerNpRunnable.java
```diff
@@ -191,29 +191,47 @@ private void doWork() {
...
+     * Getter for bean id
+     *
+     * @return
+     */
+    public TimerNpImpl getNpTimer() {
+        return this.ivTimer;
+    }
+
+    /**
+     * Getter for method id
```
tkburroughs: if you have an @return there should be text after it to say what is returned. For simple getters, the method description and @return description are often redundant, so you could have just one or the other. An empty @return will produce a javadoc warning, so if present, there should be text after it.

dev/com.ibm.ws.ejbcontainer.timer/bnd.bnd
```diff
@@ -30,8 +30,11 @@ Include-Resource: \
...
 	com.ibm.ws.ejbcontainer;version=latest,\
 	com.ibm.ws.ejbcontainer.core;version=latest,\
-	com.ibm.ws.kernel.service,\
+	com.ibm.ws.kernel.service;version=latest,\
 	com.ibm.ws.logging,\
+	com.ibm.ws.container.service;version=latest,\
 	com.ibm.websphere.org.osgi.core,\
 	com.ibm.websphere.org.osgi.service.component,\
-	com.ibm.wsspi.org.osgi.service.component.annotations
+	com.ibm.wsspi.org.osgi.service.component.annotations,\
```
tkburroughs: indent is wrong here.... not sure if it is a tabs/blanks issue.... but this last line should line up.

Also... last line should NOT end with ,\   I think that could cause problems in the future if anything else were ever added after this that is not part of buildpath.

dev/com.ibm.ws.ejbcontainer.timer/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -30,6 +30,14 @@
         <AD name="%persistentExecutorRef" description="%persistentExecutorRef.desc"
             id="persistentExecutorRef" type="String" required="false" ibm:type="pid" ibm:reference="com.ibm.ws.concurrent.persistent.executor"/>
         <AD id="persistentExecutor.target" type="String"  ibm:final="true" default="(service.pid=${persistentExecutorRef})" name="internal" description="internal use only"/>
+
```
tkburroughs: Would be nice if the indent matched the above entries.... probably some tab vs blank problem.  If you just copy the white space from above to your new lines, things should match up.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/TimerNpRunnable.java
mmerali96: Got it, I updated the comment and removed @return here and in other places.

dev/com.ibm.ws.ejbcontainer.timer/bnd.bnd
mmerali96: Made change

dev/com.ibm.ws.ejbcontainer.timer/resources/OSGI-INF/metatype/metatype.xml
mmerali96: Made change

mmerali96:
Squashed. Running PB #build

mmerali96:
Build was green. Merging PR

PR #8828, opened 2019-09-03
Update metatype.xml file to include json.fields property for logging.osgi

Yushan-Lin:
#8812
#build


Yushan-Lin:
FAT test failures has nothing to do with changes

Yushan-Lin:
Fixes #8952

PR #8834, opened 2019-09-03
create nonship mpContextPropagation-1.1 feature

njr-11:
Create nonship feature, which to start out with, will point to the v1.0 binaries.
This will gate any future experimentation with proposed 1.1 spec changes.


njr-11:
#build

PR #8851, opened 2019-09-05
synchronize checkPeerServers

### LINKED ISSUE #8839: Synchronize Peer Server Transaction Recovery
A server is failing to start in the testColdStartLocalAndPeerServer test. From the trace what has happened is that peer recovery has been initiated at server startup but has yet to complete. Concurrently, peer recovery is started by the LeaseTimoutManager (the lease parameters are very aggressive in this test). The net effect is that peer recovery initiated at startup hangs in RecoveryDirectorImpl.directInitialization() at...

_outstandingInitializationRecords.wait();

Transaction recovery (and therefore server start) cannot complete.

I plan to synchronize LibertyRecoveryDirectorImpl.peerRecoverServers().

I'll make the change under this issue .

neilgsyoung:
Fixes #8839


neilgsyoung:
#build

PR #8865, opened 2019-09-05
enable some of the session invalidation/timeout tests for infinispan

njr-11:
Enable some of the tests in SessionCacheTimeoutTest after updating the server config for the test to switch over to Infinispan.


njr-11:
#build

PR #8869, opened 2019-09-06
remaining tests in SessionCacheTimeoutTest

njr-11:
Enable the remaining tests in SessionCacheTimeoutTest.
The tests need to be able to access the same caches that are used internally by HTTP sessions in order to check contents.  This should be possible for a test case via reflection.


njr-11:
#build

njr-11:
updated with code review comment fix
#build

PR #8871, opened 2019-09-06
enable parts of SessionCacheTwoServerTest that only need 1 server

njr-11:
It should be possible to enable all of the tests under SessionCacheTwoServerTest for Infinispan which only use one of the two servers.  Other tests will need to wait until the test is eventually made to configure both servers to use the same cache.


njr-11:
#build

PR #8875, opened 2019-09-06
copy from hazelcast session config bucket to create one for Infinispan

njr-11:
Copy from the http sessions config bucket that uses Hazelcast to create a corresponding bucket for Infinispan.  Enable an initial set of tests, but leave others for subsequent pulls.


njr-11:
#build

PR #8900, opened 2019-09-10
test case for updating configuration of infinispan library

njr-11:
Starting with the copied test case for updating configuration of hazelcast library for HTTP Session caching, convert the test case to be based on Infinispan, which, because it involves multiple JARs, will require updates to the library's fileset rather than file element.  The intent of this test is to ensure that a user who has misconfigured the library can observe the error and correct the server configuration to get HTTP Sessions working, without needing to restart the server.


njr-11:
#build

PR #8915, opened 2019-09-11
enable additional codepath for testing with infinispan

njr-11:
Several tests can be partially enabled to run with Infinispan, which will improve coverage.


njr-11:
#build

PR #8920, opened 2019-09-11
add HTTP call to the testLibertyWebConfigProvider method

### LINKED ISSUE #8919: jsf2.2_fat testLibertyWebConfigProvider fails in some builds
Bucket: com.ibm.ws.jsf.2.2_fat
Test Class: JSFServerTest
Method: testLibertyWebConfigProvider

An HTTP request needs to be called to log the "isErrorPagePresent ENTRY" message in the trace.  

Since the order of the junit tests is not guaranteed, some builds run the testServlet method (containing the HTTP call) after the testLibertyWebConfigProvider method.

volosied:
Fixes #8919 

Move the testServlet code into the testLibertyWebConfigProvider method.

The " isErrorPagePresent ENTRY" is logged during the restore view phase of the JSF life cycle. 
This means it needs a HTTP request for this particular test. 

However, sometimes, the testServlet runs after the testLibertyWebConfigProvider because the _order of the junit tests is not guaranteed._ 

No other tests need to be modified since the rest of the messages are logged (or not found) during normal app start up.

### DIFF
--- a/dev/com.ibm.ws.jsf.2.2_fat/fat/src/com/ibm/ws/jsf22/fat/tests/JSFServerTest.java
+++ b/dev/com.ibm.ws.jsf.2.2_fat/fat/src/com/ibm/ws/jsf22/fat/tests/JSFServerTest.java
@@ -114,6 +114,11 @@ public class JSFServerTest {
     @Test
     @SkipForRepeat("JSF-2.3")
     public void testLibertyWebConfigProvider() throws Exception {
+        WebClient webClient = new WebClient();
+        URL url = JSFUtils.createHttpUrl(jsfTestServer1, contextRoot, "");
+        // Ensure the isErrorPagePresent message is logged in the trace during the RESTORE_VIEW phase.
+        HtmlPage page = (HtmlPage) webClient.getPage(url);
+
         String msgToSearchFor = "isErrorPagePresent ENTRY";
 
         // Check the trace.log to see if the LibertyWebConfigProvider has any entry trace.


volosied:
#link 267096

volosied:
#build

dev/com.ibm.ws.jsf.2.2_fat/fat/src/com/ibm/ws/jsf22/fat/tests/JSFServerTest.java
```diff
@@ -114,6 +99,13 @@ public void testLibertyWebConfigProviderFactory() throws Exception {
...
+        WebClient webClient = new WebClient();
+        URL url = JSFUtils.createHttpUrl(jsfTestServer1, contextRoot, "");
+        // Ensure the isErrorPagePresent message is logged in the trace during the RESTORE_VIEW phase.
+        HtmlPage page = (HtmlPage) webClient.getPage(url);
+        //Basic check to confirm the app is working
+        assertTrue(page.asText().contains("Hello World"));
+
         String msgToSearchFor = "isErrorPagePresent ENTRY";
 
         // Check the trace.log to see if the LibertyWebConfigProvider has any entry trace.
```
pnicolucci: Can you give some explanation as to why we need to do this? Why did the test fail when that request was driven after we search for the message in the logs? Does the request trigger the message in the logs?

volosied:
#build

dev/com.ibm.ws.jsf.2.2_fat/fat/src/com/ibm/ws/jsf22/fat/tests/JSFServerTest.java
pnicolucci: Talked with Vlad, basically the test requires a request to be driven and usually we're lucky and the testServlet test runs first however order is not defined for  test execution so this change makes sense.

volosied:
#build

PR #8935, opened 2019-09-12
Make the FT/Metrics 2.0 feature ready for release

### LINKED ISSUE #8886: GA Fault Tolerance - Metrics 2.0 integration
The Fault Tolerance / Metrics 2.0 integration is a new auto-feature

We need to GA that auto-feature, and remove the internal test feature which was added to stop packaging trying to include it in the release. (Packaging would try to include it because it's an auto-feature and all of its dependencies are included in the release. Fortunately, a validation check picks up the discrepancy.)

Azquelt:
* Remove the guard feature
* Mark the FT/Metrics 2.0 as GA/core
* Remove references to the guard feature from tests

Fixes #8886


Azquelt:

PR #8945, opened 2019-09-13
Throw ConnectException in TransportURIResolver

neuwerk:
CXF throws a plan Exception masking the real issue, this PR adds a change to CXF so that it now throws the specific ConnectionException


WhiteCat22 (dismissed):

neuwerk:
#build

PR #8946, opened 2019-09-13
switch config bucket to Infinispan 10.0.0.CR1

njr-11:
Have the HTTP session cache infinispan config bucket run against Infinispan 10 to verify that it works on the newest version as well.


njr-11:
#build

njr-11:
#build

PR #8961, opened 2019-09-16
test bug: wait for ltpa key creation

brutif:
Address some intermittent test failures when ltpa key isn't created until after server is up for several seconds.


brutif:
#build

PR #8996, opened 2019-09-16
Add a couple tests to OAuth BrowserAndServerLogMessageTest

ayoho:
Adds a couple unit tests to ensure that null locales objects and `tc` variables that point to unknown bundles produce the expected results. Namely, a null locales object should fall back to using English and a `tc` variable that points to an unknown message bundle should just output the message key (and object array, if one was provided) as-is.


ayoho:
#build

PR #9000, opened 2019-09-16
Fix for when jks and p12 defaults exists

acdemyers:
Change code to make sure jks keystone file is used when default location is contains both jks and p12 default keystores.


acdemyers:
#build

acdemyers:
#build

acdemyers:
#build

PR #9023, opened 2019-09-17
configured cache for infinispan config bucket

njr-11:
Use infinispan.xml file to configure the caches that are used by HTTP session persistence.  This requires updating some of the prereqs and including some missing prereqs, so that the configuration can be parsed properly.  After doing this, enable a test case that requires infinispan config.


dev/com.ibm.ws.session.cache_fat_config_infinispan/fat/src/com/ibm/ws/session/cache/config/fat/infinispan/SessionCacheErrorPathsTest.java
```diff
@@ -52,7 +52,7 @@
     private static final String SERVLET_NAME = "SessionCacheConfigTestServlet";
 
     private static ServerConfiguration savedConfig;
-    //private static String hazelcastConfigFile = "hazelcast-localhost-only.xml";
+    private static String infinispanConfigFile = "infinispanxml";
```
nmittles: Should `infinispanxml` be `infinispan.xml` ?

njr-11:
#build

PR #9024, opened 2019-09-17
Issue #8899: Federated repositories is using non-participating base e…

### LINKED ISSUE #8899: federatedRegistry-1.0 group membership may use a repository that does not participate in the realm
When doing group membership lookups, federated repositories may use a repository that is not participating in the realm for the request. This can happen when the non-participating repository has a base entry that matches the group's distinguished name more closely than the participating repository's base entry. For example, if the participating base entry was **o=ibm.com**, and it contains a group named **cn=group,ou=org,o=ibm.com** it might use a non-participating repository whose base entry is **ou=org,o=ibm.com**.

This may result in incomplete group memberships or exceptions.

jvanhill:
…ntries during group lookups.

fixes #8899


jvanhill:
#build

jvanhill:
#build

jvanhill:
#build

PR #9034, opened 2019-09-18
erroneous CWWKL0058W warning

njr-11:
pull fixes #9033

PR #9037, opened 2019-09-18
configure uri on httpSessionCache for a bell

### LINKED ISSUE #8936: Configure Infinispan to join a cluster (non local cache) when loaded as a BELL
User story:
"As a user of Liberty, I want to provide an Infinispan config file when using Infinispan for HTTP Session persistence, so that I can further customize it to fit my usage patterns."

This is possible if configuring the Infinispan libraries along with uri on the httpSessionCache element, however, currently there is no way to specify Infinispan configuration xml when loading the libraries as a BELL. This means the caches default to local mode and won't ever join a cluster.  Compare the configuration in the Infinispan setup in the following server.xml with the corresponding Hazelcast test server:
`/com.ibm.ws.session.cache_fat_infinispan/publish/servers/com.ibm.ws.session.cache.fat.infinispan.serverB/server.xml` 

*Note that when Hazelcast has been configured as a BELL in tests, we provide the configuration via JVM options. Documentation on how to do this for Infinispan has not been found yet, and may not exist. There are programmatic equivalents to all Infinispan configuration xml, but this would require extensive use of reflection and Infinispan specific API's.

The approach to be taken for this story is to utilize the `httpSessionCache` config to specify the configuration URI and support not having the libraryRef specified.

*** When writing this story, we didn't realize that the above was already a supported behavior of httpSessionCache, and thought the capability still needed to be added.  However, it turns out that it is already working.  So the only work to do under this story amounts to updating tests to utilize it.

njr-11:
Pull fixes #8936

### DIFF
--- a/dev/com.ibm.ws.session.cache_fat_infinispan/fat/src/com/ibm/ws/session/cache/fat/infinispan/SessionCacheTwoServerTest.java
+++ b/dev/com.ibm.ws.session.cache_fat_infinispan/fat/src/com/ibm/ws/session/cache/fat/infinispan/SessionCacheTwoServerTest.java
@@ -47,20 +47,6 @@ public class SessionCacheTwoServerTest extends FATServletClient {
         appB = new SessionCacheApp(serverB, true, "session.cache.infinispan.web", "session.cache.infinispan.web.cdi", "session.cache.infinispan.web.listener1");
         serverB.useSecondaryHTTPPort();
 
-        //String hazelcastConfigFile = "hazelcast-localhost-only.xml";
-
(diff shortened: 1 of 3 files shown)


njr-11:
#build

njr-11:
Based on the build output, it appears this will require #9034 to go in first, then another build can be submitted.

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build completed successfully with no failing tests now that the change upon which it depends has gone into master.

PR #9042, opened 2019-09-18
Add @NonNull support

andymc12:
Note, this does not include a snapshot of MP GraphQL APIs that contain the `@NonNull` annotation.  So this won't actually work until the API is brought in to OpenLiberty - but it won't break anything either.

This implements the MP GraphQL issue [68](https://github.com/eclipse/microprofile-graphql/issues/68).  The spec is still in progress, and this functionality is part of currently non-GA code, so it is not a release bug.

These changes were tested with MP GraphQL TCK tests that are part of PR [71](https://github.com/eclipse/microprofile-graphql/pull/71).


andymc12:
#build

PR #9043, opened 2019-09-18
Fix incorrect security manager check

andymc12:
Fixes a Java 2 security manager check that was incorrect - was only using the doPriv if the security manager _wasn't_ set.


andymc12:
#build

PR #9050, opened 2019-09-19
Replacing original copyright years

KyleAure:
Moved FAT buckets from CL to OL, but did not bring over the original copyright years originally.  Also removing unused imports where necessary.


KyleAure:
#build

PR #9131, opened 2019-09-27
BNFHeader find header ignore case

### LINKED ISSUE #9130: Header Key retrieval fix for case sensitivity
There is a scenario where header key's might not match due to their casing if the channel's cache is full. This fix will ensure that case is ignored regardless of whether the cache is used or the internal connection's header memory is.

mrsaldana:
fixes #9130


mrsaldana:
#build

PR #9149, opened 2019-10-01
Build a new MP Config instance each time and close when done

### LINKED ISSUE #8958: Multiple instances of custom ConfigSource from MicroProfile Config 1.3 feature on WLP 19.0.0.8
The custom configSource implemented registered in _org.eclipse.microprofile.config.spi.ConfigSource_ file (using spi lifecycle) results in multiple instance if the application is hot redeployed.
The effect of the described issue is that the code below print the new objectId and the older objectid (not destroyed) after each app hot deploy.

`

    public class CustomConfigSource implements ConfigSource {

    Map<String, String> map = new HashMap<>();

    @Override
    public int getOrdinal() {
        return 400;
    }

    @Override
    public String getValue(String key) {
        return map.get(key);
    }

    @Override
    public String getName() {
        return "Custom Config Source";
    }

    @Override
    public Map<String, String> getProperties() {

        map.put("config_one", "config_one_val:" + new Date());
        System.out.println(">>>>> " + map.get("config_one") + this); // this is still alive after app redeploy
        return map;

    }
    }
`

tevans78:
fixes #8958 
#build

### DIFF
--- a/dev/com.ibm.ws.microprofile.openapi/src/com/ibm/ws/microprofile/openapi/ApplicationProcessor.java
+++ b/dev/com.ibm.ws.microprofile.openapi/src/com/ibm/ws/microprofile/openapi/ApplicationProcessor.java
@@ -119,169 +119,174 @@ public class ApplicationProcessor {
         ClassLoader appClassloader = moduleInfo.getClassLoader();
         boolean isOASApp = false;
 
+        OpenAPI newDocument = null;
+
         //read and process the MicroProfile config
         ConfigProcessor configProcessor = new ConfigProcessor(appClassloader);
-        if (OpenAPIUtils.isEventEnabled(tc)) {
-            Tr.event(tc, "Retrieved configuration values : " + configProcessor);
-        }
-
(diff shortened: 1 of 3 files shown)


tevans78:
#build

Azquelt:
If `ConfigProcessor` always needs to be closed, I would make it implement `Closable` (or `AutoClosable`) and use try-with-resources to close it.

tevans78:
> If `ConfigProcessor` always needs to be closed, I would make it implement `Closable` (or `AutoClosable`) and use try-with-resources to close it.

If I do that then the calling code needs to handle the Exception rather than it just being confined to the close method of ConfigProcessor.

tevans78:
> Not necessarily.
> 
> On the implementing class, you can drop (or more narrowly specify) the `throws` declaration. As long as the calling class knows that the `close()` method of the class it's dealing with doesn't throw checked exceptions, you won't need to handle the exception in the try-with-resources

Cheers, I made that change

tevans78:
#build

PR #9151, opened 2019-10-01
MicroProfile parallel bundle activation

### LINKED ISSUE #8854: Enable MicroProfile features for parallel bundle activation
Similar to how the Servlet features were enabled for parallel activation with #8106 the various microprofile features should also be enabled.  In order to do this you must detect the transitive closure of all features for the micro profile feature you want to enable and add the feature header `WLP-Activation-Type: parallel`

To get the feature list enable all the features you want to parallel activate in the `server.xml` and add the following trace option to your `bootstrap.properties`:
`com.ibm.ws.logging.trace.specification=com.ibm.ws.kernel.feature.internal.FeatureManager=FINEST`

This will add a message to the `trace.log` file with the text "all installed features" followed by all the installed features public/protected/private.  Capture the comma separated list of features.

After getting the list of features I want to enable I run something like the following script from the `dev/com.ibm.websphere.appserver.features/visibility` directory passing in the list of features.  You will then want to run several test builds and also enable full FAT execution for the buckets that exercise the features you are enabling parallel activation for.

```
#!/bin/bash

features="$1"

features="${features//,}"

addActivationType() {
  if grep -q WLP-Activation-Type "$1"; then
    echo "  PRESENT  WLP-Activation-Type already present in $1"
  else
    echo "  ADDED    WLP-Activation-Type added to           $1"
    echo "WLP-Activation-Type: parallel" >> $1
  fi
}

# Iterate the string variable using for loop
for val in $features; do
  if [ -f auto/$val.feature ]; then
    echo "AUTO $val"
    addActivationType auto/$val.feature
  else
    if [ -f private/$val.feature ]; then
      echo "PRIVATE $val"
      addActivationType private/$val.feature
    else
      if [ -f protected/$val.feature ]; then
        echo "PROTECTED $val"
        addActivationType protected/$val.feature
      else
        if [ -d public/$val ]; then
          echo "PUBLIC $val"
          for file in public/$val/*.feature
          do
            if [[ -f $file ]]; then
              addActivationType $file
            fi
          done
        else
          echo "UNKNOWN $val"
        fi
      fi
    fi
  fi
done
```

tevans78:
fixes #8854 
#build

### DIFF
--- a/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.cdi1.2-appSecurity1.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.cdi1.2-appSecurity1.0.feature
@@ -7,3 +7,4 @@ IBM-Provision-Capability: osgi.identity; filter:="(&(type=osgi.subsystem.feature
 IBM-Install-Policy: when-satisfied
 kind=ga
 edition=core
+WLP-Activation-Type: parallel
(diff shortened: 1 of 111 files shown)


tevans78:
Hopefully this is a PB with all tests in FULL mode
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_a1n5EORhEemM--BPo4KzaA

PR #9165, opened 2019-10-01
Implement and test `@Description` annotation

andymc12:
This implements the `@Description` handling in MP GraphQL.  It handles all cases where `@Description` can be placed: types, input types, queries, mutations, arguments, entity fields, entity setters (input only), entity getters (output only).

Additionally, when no `@Description` is provided, but the user does provide a `@JsonbDateFormat` annotation, the value of the date format is used as the description.

It includes test cases.


andymc12:
#build

PR #9168, opened 2019-10-02
Check for race condition when getting services

benjamin-confino:


dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/sources/OSGiConfigUtils.java
```diff
@@ -146,6 +146,7 @@ private static ConfigVariables getConfigVariables(BundleContext bundleContext) t
      * @throws InvalidFrameworkStateException if the server OSGi framework is being shutdown
      * @throws ServiceNotFoundException       if an instance of the requested service can not be found
      */
+    @FFDCIgnore(InvalidFrameworkStateException.class)
```
tevans78: why is this needed?

benjamin-confino: In all honesty I'm going by the fact it's on a lot of other methods. Though it seems reasonable that we don't want to throw an exception that only occurs during shutdown as it's unlikely to be an issue the customer needs to worry about.

tevans78: i think this will fail to build ... the build checks that all FFDCIgnore exceptions are actually thrown

tevans78: I think this will fail to build ... the build checks that all FFDCIgnore exceptions are _caught_ in the method.

benjamin-confino:
#build

PR #9173, opened 2019-10-02
Update ResponseUrlExpectationTest for different JDK behavior

ayoho:
A recent OpenJ9 OpenJDK update appears to have introduced a behavior change in the `java.net.URL(String)` constructor such that strings with LF characters now throw a `java.net.MalformedURLException`. The `test_validate_checkTypeMatches_matchSpansMultipleLines` test case creates a `URL` object with this exact scenario but currently doesn't expect any kind of exception to be thrown. This change now allows a `java.net.MalformedURLException` to be thrown.


ayoho:
#build

ayoho:
#build

PR #9176, opened 2019-10-02
Update createProperty logic

KyleAure:
Align createProperty exception checking to investigate SQL State when an eclipselink DatabaseException is thrown for situations when we can determine that property already exists.

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
+++ b/dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
@@ -12,6 +12,7 @@ package com.ibm.ws.concurrent.persistent.db;
 
 import java.security.AccessController;
 import java.sql.Connection;
+import java.sql.SQLException;
 import java.sql.SQLIntegrityConstraintViolationException;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -27,6 +28,8 @@ import javax.persistence.LockModeType;
 import javax.persistence.PersistenceException;
 import javax.persistence.Query;
 import javax.persistence.TypedQuery;
+
+import org.eclipse.persistence.exceptions.DatabaseException;
 
 import com.ibm.websphere.concurrent.persistent.PersistentExecutor;
 import com.ibm.websphere.concurrent.persistent.TaskState;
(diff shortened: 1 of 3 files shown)


dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
```diff
@@ -276,13 +279,46 @@ public boolean createProperty(String name, String value) throws Exception {
...
             em.flush();
         } catch (EntityExistsException x) {
-            return false;
+            /*
+             * FIXME JPA Spec says em.persist should throw EntityExistsException if entity already exists.
+             * EclipseLink implementation throws PersistanceException during the flush operation.
+             * If EclipseLink changes behavior, this exception will notify us that we need to reactor
+             * this catch to return false, and subsequent PersistenceException can remove extra checks.
+             */
+            throw new UnsupportedOperationException("EntityExistsException thrown during persist operation. "
```
KyleAure: Is it possible for a user to use a different JPA implementation for concurrent persistence or do we always use the eclipselink impl?

dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
```diff
@@ -276,13 +279,46 @@ public boolean createProperty(String name, String value) throws Exception {
...
+                if (sqlException != null) {
+
+                    //Investigate SQL State
+                    String message = sqlException.getLocalizedMessage();
+                    String sqlState = sqlException.getSQLState();
+                    int errorCode = sqlException.getErrorCode();
+
+                    //SQL State 23505 - Unique Constraint Violation
+                    //Suggests that we are trying to insert a row, but the primary key already exists.
+                    if (sqlState.equalsIgnoreCase("23505")) {
```
KyleAure: Postgre, Derby, and DB2 all use SQL State 23505 for unique constraint violation, which I think is specific enough that we do not need to investigate the individual error code.  Additionally, Postgre (in our tests) returns an error code of 0 - successful along with the 23505 state. 
Oracle and SQL Server use the more general SQL State 23000 for any integrity constraint violation.  Currently I only check for 23505 but we could also look into 23000 + error code.

dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
njr-11: Persistence service currently only allows the use of EclipseLink, but I'm glad you thought of that possibility, because we will want to take it into account.
Given that the error will be the first thing the JPA team sees when they make changes in this area -- either to enhance EclipseLink or to support a different JPA persistence provider -- let's be more explicit in the exception message.  Maybe something like: "EntityExistsException codepath indicates that additional function has been added. See comments in code for how to handle."

Other than that, some very minor typo fixes are:
PersistanceException -> PersistenceException
reactor --> refactor

dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
```diff
@@ -276,13 +279,46 @@ public boolean createProperty(String name, String value) throws Exception {
...
+             * If EclipseLink changes behavior, this exception will notify us that we need to reactor
+             * this catch to return false, and subsequent PersistenceException can remove extra checks.
+             */
+            throw new UnsupportedOperationException("EntityExistsException thrown during persist operation. "
+                                                    + "This behavior was not previously seen, and needs to be handled by the PersistanceExecutor.", x);
         } catch (PersistenceException x) {
             // Some JPA providers may throw PersistenceException for some scenarios where it can be
             // determined the row really does exist; handle those here
+
+            //TODO Remove later - Caused by Constraint Violation
```
njr-11: Let's remove it now because it's unreachable with EclipseLink.

dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
```diff
@@ -276,13 +279,46 @@ public boolean createProperty(String name, String value) throws Exception {
...
+
+            //TODO Remove later - Caused by Constraint Violation using generic SQLException
+            DatabaseException dbException = x.getCause() instanceof DatabaseException ? (DatabaseException) x.getCause() : null;
+            if (dbException != null) {
+                SQLException sqlException = dbException.getCause() instanceof SQLException ? (SQLException) dbException.getCause() : null;
+                if (sqlException != null) {
+
+                    //Investigate SQL State
+                    String message = sqlException.getLocalizedMessage();
+                    String sqlState = sqlException.getSQLState();
```
njr-11: Can defer operations to trace block which are only used within trace block.

dev/com.ibm.ws.concurrent.persistent/src/com/ibm/ws/concurrent/persistent/db/DatabaseTaskStore.java
njr-11: Completely agree.  Error codes are vendor-specific and there is no sense in consulting those (would need to know which vendor it is) when we can use the standard SQL States.  As far as we are concerned, the detection of any integrity constraint violation should be sufficient (code was previously intending to detect that via java.sql.SQLIntegrityConstraintViolationException, but did so erroneously due to the extra EclipseLink chained exceptions).  Let's make this as generic and standards-based as we possibly can, in order to have the best chance of working with all databases, even those we haven't tried out.  To accomplish that, I'd recommend a check for anything under 23xxx for constraint violations (need to confirm if this is accurate).  The other thing we can do to be even more generic is just step through every exception in the chain, regardless of EclipseLink-specific classes.  Something like this (I know it won't compile, but just to get the idea across),
```
for (Throwable cause = x.getCause(); cause != null; cause = cause.getCause())
    if (cause instanceof SQLIntegrityContstraintViolationException ||
        cause instanceof SQLException && sqlState != null && sqlState.length() == 5 && sqlState.startsWith("23")) ...
```

KyleAure:
#build

PR #9182, opened 2019-10-03
Issue 9178: Add eyecatcher on remote ldap failure/zOS run

### LINKED ISSUE #9178: Add eye catcher message if we can't retrieve remote Ldaps on z/OS run
On z/OS, we can't run the local Ldap clone of our remote Ldap servers. If we can't get remote Ldap servers at test time, we'll fail on those tests. This does not happen often and there are other valid tests in the buckets that can run.

Adding an extra message for LdapUtils when we're running on z/OS and hit this case.

kristip17:
Fixes #9178


kristip17:
#build

kristip17:
First PB had a single unrelated failure but it was only reported on my PB so launched another one.

PR #9194, opened 2019-10-03
initial tests of failover when app stops

njr-11:
Create the first few tests of scenarios where EJB persistent timers are expected to fail over when the application (but not the server) on which they are running stops.
Under this pull, coverage is added for scenarios with a singleton automatic timer and programmatic timer that runs outside of a global transaction.


njr-11:
#build

PR #9201, opened 2019-10-04
test where a failing timer fails over with retries remaining

njr-11:
Adds a test case where a timer that is failing on one server fails over to another server while there are still retries remaining.  It is not necessary to use up all of the retries before the failover occurs. It can happen as soon as the missedTaskThreshold is reached.


njr-11:
#build

PR #9208, opened 2019-10-07
Set TCCL before calling OASFilter

### LINKED ISSUE #9204: microprofile-config.properties is added to config sources too late
I did implementation of `OASFilter`, where I wanted to read config value, which is stored inside `microprofile-config.properties` file. The value hasn't been found.

Code I tried:
```
Config config = ConfigProvider.getConfig();
config.getValue("my.var", String.class)
```

After some investigation, I found out, that config file hadn't been added to config sources yet and `OASFilter` method is being executed.

I printed out all config sources:

> Server XML Application Properties Config Source(600)
Server XML Variables Config Source(500)
System Properties Config Source(400)
Environment Variables Config 1.3 Source
server.xml.variable.config.source(1)

When I tried to run the same code later over REST endpoint, everything worked as expected.

> Server XML Application Properties Config Source(600)
Server XML Variables Config Source(500)
System Properties Config Source(400)
Environment Variables Config 1.3 Source
Properties File Config Source: file:/opt/ol/wlp/usr/servers/defaultServer/apps/expanded/my.project.war/WEB-INF/classes/META-INF/microprofile-config.properties(100)
server.xml.variable.config.source(1)

It seems that `microprofile-config.properties` file is added too late to config sources.

tevans78:
fixes #9204

### DIFF
--- a/dev/com.ibm.ws.microprofile.openapi/src/com/ibm/ws/microprofile/openapi/ApplicationProcessor.java
+++ b/dev/com.ibm.ws.microprofile.openapi/src/com/ibm/ws/microprofile/openapi/ApplicationProcessor.java
@@ -10,6 +10,7 @@
  *******************************************************************************/
 package com.ibm.ws.microprofile.openapi;
 
+import java.security.AccessController;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -60,6 +61,7 @@ import com.ibm.ws.microprofile.openapi.impl.validation.ValidatorUtils;
 import com.ibm.ws.microprofile.openapi.utils.OpenAPIUtils;
 import com.ibm.ws.microprofile.openapi.utils.ProxySupportUtil;
 import com.ibm.ws.microprofile.openapi.utils.ServerInfo;
+import com.ibm.ws.util.ThreadContextAccessor;
 import com.ibm.wsspi.adaptable.module.Container;
 import com.ibm.wsspi.adaptable.module.Entry;
 import com.ibm.wsspi.adaptable.module.NonPersistentCache;
@@ -73,6 +75,8 @@ import com.ibm.wsspi.http.VirtualHost;
 public class ApplicationProcessor {
 
     private static final TraceComponent tc = Tr.register(ApplicationProcessor.class);
+
+    private static final ThreadContextAccessor THREAD_CONTEXT_ACCESSOR = AccessController.doPrivileged(ThreadContextAccessor.getPrivilegedAction());
 
     enum DocType {
         JSON,
@@ -254,6 +258,8 @@ public class ApplicationProcessor {
             OASFilter oasFilter = OpenAPIUtils.getOASFilter(appClassloader, configProcessor.getOpenAPIFilterClassName());
             if (oasFilter != null) {
                 final OpenAPIFilter filter = new OpenAPIFilter(oasFilter);
+
+                Object oldClassLoader = THREAD_CONTEXT_ACCESSOR.pushContextClassLoaderForUnprivileged(appClassloader);
                 try {
                     filter.filter(newDocument);
                     if (OpenAPIUtils.isEventEnabled(tc)) {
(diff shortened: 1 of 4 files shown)


tevans78:
#build

tevans78:
#build

dev/com.ibm.ws.kernel.service/src/com/ibm/ws/kernel/service/util/SecureAction.java
```diff
@@ -583,6 +584,20 @@ public ClassLoader getContextClassLoader() {
...
+            AccessController.doPrivileged(new PrivilegedAction<Void>() {
+                @Override
+                public Void run() {
+                    Thread.currentThread().setContextClassLoader(classLoader);
+                    return null;
+                }
+            }, controlContext);
+        }
+    }
+
```
aguibert: Instead of adding this method I would prefer we use the existing API `com.ibm.ejs.util.dopriv.SetContextClassLoaderPrivileged` which is a much more performant implementation of setting the TCCL. 

It would be nice to have this function consolidated into the `SecureAction` class, so perhaps you could keep this method and delegate off to the existing API

tevans78: Ok, I’ll take a look at that tomorrow... and we should probably advertise things like this as best practice

tevans78: I have switched to using `com.ibm.ejs.util.dopriv.SetContextClassLoaderPrivileged`. I would have liked to have made the code in `com.ibm.ws.kernel.service` common too but that will have to be another issue at another time.

tevans78:
#build

dev/com.ibm.ws.microprofile.openapi/src/com/ibm/ws/microprofile/openapi/ApplicationProcessor.java
```diff
@@ -74,6 +78,13 @@
 
     private static final TraceComponent tc = Tr.register(ApplicationProcessor.class);
 
+    private static final PrivilegedAction<ThreadContextAccessor> getThreadContextAccessorAction = new PrivilegedAction<ThreadContextAccessor>() {
```
Azquelt: If this is a constant, should it be named with CAPS_AND_UNDERSCORES?

Azquelt: Also, why do we declare and use this instead of using `com.ibm.ws.util.ThreadContextAccessor.getPrivilegedAction()`?

dev/com.ibm.ws.microprofile.openapi/src/com/ibm/ws/microprofile/openapi/ApplicationProcessor.java
```diff
@@ -254,6 +265,11 @@ private OpenAPI processWebModule(Container appContainer, WebModuleInfo moduleInf
             OASFilter oasFilter = OpenAPIUtils.getOASFilter(appClassloader, configProcessor.getOpenAPIFilterClassName());
             if (oasFilter != null) {
                 final OpenAPIFilter filter = new OpenAPIFilter(oasFilter);
+
+                ThreadContextAccessor tca = System.getSecurityManager() == null ? ThreadContextAccessor.getThreadContextAccessor() : AccessController.doPrivileged(getThreadContextAccessorAction);
```
Azquelt: Does this need to be done every time or could it be stored in a static?

Azquelt: Can we follow the examples in `ThreadContextAccessor`?

```
private static final ThreadContextAccessor threadContextAccessor = AccessController.doPrivileged(ThreadContextAccessor.getPrivilegedAction());
```

(but uppercased ;) )

and then

```
 Object origCL = threadContextAccessor.pushContextClassLoaderForUnprivileged(cl);
 try {
     ...
 } finally {
     threadContextAccessor.popContextClassLoaderForUnprivileged(origCL);
 }
```

tevans78: Done ;-)

PR #9216, opened 2019-10-07
clean up various TODOs

njr-11:
- Put SessionCacheErrorPathsTest test bucket back into TestMode.FULL
- Remove TODO around getCacheRemovals. It looks like other tests do use sessions and can run prior to this, so it should be okay for removals to be reported. Now the test will just check that the removal count is non-negative.
- Remove commented-out Hazelcast config from several classes
- Update SessionCacheTimeoutTest to tolerate the absence of multicast.
- Remove the copied HazelcastClientTest because a separate test suite exists for covering the Infinispan client/server scenarios. The test regarding WABs is for sessionCache-1.0 feature, not Inifispan, so it is fine to only have that test in the Hazelcast session cache suite.
- testWriteFrequency should not assume that a cache stays around when the only member is brought down and restarted.  Updated the test case to not make this assumption, but still test whether the actual config update being made to the sessionCache element is honored.


njr-11:
#build

PR #9218, opened 2019-10-07
Remove `@Deprecated` references and tests

andymc12:
The `@Deprecated` support in MP GraphQL has been pushed back to a later release.  It's still possible that it will be added back in to the 1.0 release, but unlikely at this point.  This change removes that support, and can be reverted if deprecation support is part of 1.0.


andymc12:
#build

PR #9223, opened 2019-10-07
Fixes 261572, checks if session is valid before invalidating

ellen-lau:
#build


ellen-lau:
Failures are unrelated to the changes made.

PR #9237, opened 2019-10-08
unnecessary failover attempt results in too much locking

njr-11:
If a timer takes too long to run (longer than the missedTaskThreshold, which ought to be increased to accommodate), Liberty makes an attempt to fail over, which it doesn't realize is unnecessary.  This attempt ends up with a lock collision on the task entry, which is still running (validly) on the original server.  This pull makes some updates to reduce the waiting for locks and avoid some of the impact when it happens.  The proper solution here is for the end user to increase the missedTaskThreshold, or otherwise debug why their timer has taken too long, but we would like Liberty to behave as gracefully as possible when this happens.


njr-11:
#build

PR #9251, opened 2019-10-09
Ensure name in `@InputField` is honored

andymc12:
The `@InputField("someOtherName")` annotation is used to specify a different field name for input type fields in the schema.  Prior to this change, it worked just fine for schema generation but failed when deserializing the input JSON (from a mutation) into the actual object, because the JSON would include a field name (i.e. "someOtherName") that doesn't exist in the Java POJO being deserialized.  This change fixes that by using a custom JSON-B `PropertyNamingStrategy` that maps the user-specified input field name back into the POJO field name.

This change also includes a FAT test case showing that this works.


andymc12:
#build

PR #9253, opened 2019-10-09
Add RESTHandler Auditing

### LINKED ISSUE #7265: Add authorization auditing to RESTHandler
As part of adding reader-role support, we should be auditing authorization results in RESTHandlerContainerImpl.

This requires addition of auditing logic as well as any auditing tests. The auditing FVT framework is currently in WS-CD-Open, so it may need to be pulled down into Open Liberty.

ryanesch:
fixes #7265


ryanesch:
#build

PR #9268, opened 2019-10-10
updated changelogs for rpm/deb packages

bensonlatibm:
Update changelog inside  openliberty rpm and debian packages


bensonlatibm:
#build

PR #9288, opened 2019-10-11
add heartbeatInterval as nonship

njr-11:
Add a configurable heartbeatInterval, initially as nonship/internal, so that we can experiment with its implementation.


njr-11:
#build

njr-11:
commit was delivered under another pull

PR #9297, opened 2019-10-11
Never set MaxPermSize and remove generation of WLP_SKIP_MAXPERMSIZE

### LINKED ISSUE #9197: Remove MaxPermSize settings since Java 7 is no longer supported
When we transitioned from Java 7 to Java 8 we added a setting related to MaxPermSize in all server.env files of newly created servers, and also added some code to handle this new boolean option in the server scripts.

Since Java 7 is no longer supported, we can clean up this option.

aguibert:
Fixes #9197


aguibert:
#build

aguibert:
#build

PR #9314, opened 2019-10-14
internal attribute to gate experimentation with infinispan config

njr-11:
Some experimentation will be done with augmenting/generating config for Infinsipan when used by HTTP session persistence.  In order to keep the experimentation on nonship (and later beta) codepath so that no one starts to depend on partially completed function, we need a configuration attribute to gate its enablement.


njr-11:
#build

PR #9326, opened 2019-10-15
experiment with temporary files for infinispan config

njr-11:
Experiment with writing a temporary file when Infinispan config is absent, gated by the property that was added for this purpose.  Findings thus far shown that doing this alters the MBean object names, which are based on the URI.  We will need to consider whether/how to cope with this later, as well as any changes to intelligently generate/modify the infinispan config.


njr-11:
#build

PR #9357, opened 2019-10-16
add heart beats per the heartbeatInterval

njr-11:
This pull adds period heart beats per the heartbeatInterval, along with checking for expired heartbeats and removal of the expired entry and transfer of its tasks. Having done this, a test case is enabled which now passes.


njr-11:
#build

njr-11:
#build

PR #9370, opened 2019-10-17
Support for OPENJ9_JAVA_OPTIONS

### LINKED ISSUE #9096: Support for OPENJ9_JAVA_OPTIONS
IBM_JAVA_OPTIONS is being deprecated from OpenJ9 builds
so we need to update the server and other scripts to make sure it works moving forward.

the new option is
OPENJ9_JAVA_OPTIONS
they will currently read both with OPENJ9_JAVA_OPTIONS preferred
so I think we should set both for now.

canterburymalcolm:
fixes #9096 

- IBM_JAVA_OPTIONS is still being set with the same Java options but all references to it in the server scripts now use OPENJ9_JAVA_OPTIONS.
- The added test checks that they are both set correctly and are equal to each other

### DIFF
--- a/dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/ServerStartJavaEnvironmentVariablesTest.java
+++ b/dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/ServerStartJavaEnvironmentVariablesTest.java
@@ -1,0 +1,138 @@
+package com.ibm.ws.kernel.boot;
+/*******************************************************************************
+ * Copyright (c) 2019 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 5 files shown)


canterburymalcolm:
#build

dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/ServerStartJavaEnvironmentVariablesTest.java
```diff
@@ -0,0 +1,138 @@
...
+                    }
+                }
+                assertTrue("IBM_JAVA_OPTIONS was not found", firstLine[0].equals("IBM_JAVA_OPTIONS"));
+
+                entryReader.close();
+                entryInputStream.close();
+            }
+        }
+
+        zipFile.close();
```
hlhoots: You may want to use a try block and have the zipFile.close() and dumofile.delete() in a finally clause so it always gets called.

canterburymalcolm:
#build

canterburymalcolm:
#build

PR #9388, opened 2019-10-18
3925 Propagate utility workarea to location service

### LINKED ISSUE #3925: Minfying and then starting a server can result in two singleton features being started
When you have a server which uses features with dependencies that have tolerates, you can end up with a minified server which includes two versions of a singleton feature.

E.g. in my development environment, I had
* mpFaultTolerance-1.1 -> mpConfig-1.1 (tolerates 1.2)
* mpMetrics-1.1 -> mpConfig-1.2

If I run `server package --include=minify` then the minified server includes both mpConfig-1.1 and mpConfig-1.2. (This fine and, I think, as expected)

However, if I then run `server run`, the server starts both mpConfig-1.1 and mpConfig-1.2 at the same time! (Predictably, this then causes my app not to start).
`[AUDIT   ] CWWKF0012I: The server installed the following features: [mpFaultTolerance-1.1, servlet-3.1, ssl-1.0, jndi-1.0, jsonp-1.0, appSecurity-2.0, mpConfig-1.1, jaxrs-2.0, mpConfig-1.2, jaxrsClient-2.0, concurrent-1.0, mpRestClient-1.0, mpMetrics-1.1, json-1.0, cdi-1.2, distributedMap-1.0].`

**Note 1**: I'm not sure how likely this is to happen in the real world with openliberty at the moment as I don't know how many tolerated feature dependencies we have. Some investigation is needed to see if it's possible to reproduce this on a released version. By the time it's ready for release, mpFaultTolerance will probably just require a specific version of mpConfig.

**Note 2**: I haven't checked if this occurs on Websphere Liberty (we have many more tolerated features there because we have things that work with Java EE6 and Java EE7)

**Note 3**: Packaging and then running a server might not seem a common scenario, but it's very likely to happen when using the liberty maven plugin if you've configured it to produce a minified package or runnable server jar. When testing for development you're likely to run `mvn package liberty:run-server` which will build the package and then start the server.

**Note 4**: Starting the server with `--clean` the first time after packaging fixes the issue. We probably just need to clean out the workarea directory after (and maybe before?) packaging the server. I don't know if it's possible to do this in the maven scenario.

dazavala:
fixes #3925 

This solution has two parts: 
1. Modify the bootstrap facility to set the package command's workarea as a new server config property, making it accessible at service activation
2. Modify the location service to use the property when creating the workarea directory resource and the corresponding variable registry value.  

These changes ensure the feature manager uses the workarea location specified by command utilities that launch an Embedded Server instance.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/BootstrapConfig.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/BootstrapConfig.java
@@ -91,6 +91,8 @@ public class BootstrapConfig {
      */
     protected File workarea = null;
 
+    protected String workareaDirStr = null;
+
     /** Name of server instance */
     protected String processName;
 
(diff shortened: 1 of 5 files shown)


dazavala:
#build

dazavala:
@tjwatson Would you take a look at commit "3925 Propagate utility workarea to location service".

dazavala:
#build

dazavala:
@tjwatson See new commit "3925 Verify minify does not corrupt server feature cache". Merge is pending any additional comment and build results.

tjwatson:
#build

tjwatson:
#build

Issue #9397, opened 2019-10-21
com.ibm.ws.injection_fat fails with NameNotFoundException


brideck:
ERROR: Caught exception attempting to call test method testServletCompDSD on servlet com.ibm.ws.injection.dsdann.web.BasicDSDAnnServlet 
javax.naming.NameNotFoundException: javax.naming.NameNotFoundException: java:module/ann_ServletCompLevelDS 
   at com.ibm.ws.jndi.url.contexts.javacolon.internal.JavaURLContext.lookup(JavaURLContext.java:355) 
   at com.ibm.ws.jndi.url.contexts.javacolon.internal.JavaURLContext.lookup(JavaURLContext.java:370) 
   at org.apache.aries.jndi.DelegateContext.lookup(DelegateContext.java:161) 
   at java.naming/javax.naming.InitialContext.lookup(InitialContext.java:409)
   at com.ibm.ws.injection.dsdann.web.BasicDSDAnnServlet.testServletCompDSD(BasicDSDAnnServlet.java:227)
   ...

PR #9401, opened 2019-10-21
Issue #9397: injection_fat fails with NameNotFoundException

brideck:
- Update OSGiInjectionScopeData to lookup in java:comp and java:module
- Update injection unittests
- Add new injection_fat


brideck:
fixes #9397

brideck:
#build

brideck:
The change will affect all types of resources, not just JavaMail definitions; however, @tkburroughs and I discussed this and we don't think it's a behavior change that should concern customers.  Basically, there's a certain category of lookups that used to throw NameNotFoundException and now they correctly resolve.  It doesn't impact any lookups that already would have been working.  We think it highly unlikely that a customer would be relying on a failed lookup for correct behavior in their application.

brideck:
The single failure is a UI failure that has to do with collectives initialization and is unrelated to the changes here.  I've consulted with @amyreit and she has said that it is okay to proceed with my merge.

PR #9359, opened 2019-10-16
default server file updates

acdemyers:

PR #9410, opened 2019-10-22
Revert "default server file updates"

tevans78:
Reverts OpenLiberty/open-liberty#9359

The parent feature is not being included in the next release and these changes are not beta guarded (may not be technically possible). Reverting until the following release.


acdemyers:
#build

PR #9432, opened 2019-10-22
[9412] Ensure mpGraphQL-1.0 works with mpMetrics-1.1 and 2.0

andymc12:
Resolves issue #9412 - makes use of dynamic imports for metrics packages, tweaks the metric names (adding "mp_graphql" prefix), and test cases to ensure that stats are collected accurately.

This feature has not yet shipped, so it is not a release bug.

PR #9435, opened 2019-10-22
Fix Java 2 security issue with getting thread context classloader

andymc12:
While working on pull request #9432, I ran into a failure with Java 2 security enabled where the mpMetrics-2.0 MetricRegistryImpl attempts to get the thread's context classloader without using a doPriv. (The mpMetrics-1.1 version works fine - either using a doPriv, or it does the same thing some other way).

This change checks for a security manager, and if one exists, then it gets the classloader using a doPriv block.


andymc12:
#build

PR #9442, opened 2019-10-23
Fix installing auto-features which depend on kernel features

### LINKED ISSUE #9441: auto-features which depend on kernel features do not get installed
If an auto-feature references a kernel feature in its Provision-Capability header, this requirement is incorrectly considered unmet during installation which results in the auto-feature not being installed.

I think this only affects some Websphere Liberty features specific to z/OS.

Azquelt:
Fixes #9441 

Kernel features are part of the platform and are always installed and started

Auto-features should be installed and started automatically when all of their dependencies are met

Previously, we weren't considering the kernel features when working out whether the dependencies of auto-features were met.

This was causing some z/OS specific Websphere Liberty auto-features not to be installed when their requirements were met.


Azquelt:
#build

Azquelt:
There's a Websphere Liberty test which needs to be updated before this can be merged.

Azquelt:
#build

Azquelt:

PR #9443, opened 2019-10-23
Enable Parallel Bundle Activation for MP3.1 (no security)

tevans78:
#build 
This is the same as https://github.com/OpenLiberty/open-liberty/pull/9346 but minus anything with security in the name, which may have to be done separately.


tevans78:
This PR is stale now, closing.

PR #9452, opened 2019-10-23
Fix Intermittent NullPointerException on TCP trace during shutdown

### LINKED ISSUE #9451: Fix Intermittent NullPointerException on TCP trace during shutdown
A java.lang.NullPointerException may happen while TCP channel tracing is enabled during SocketWRChannelSelector.checkForTimeouts if the server shutdown has already been issued.

mrsaldana:
fixes #9451


mrsaldana:
#build

Issue #9465, opened 2019-10-23
Change mockserver to use -keysize of 2048.


jkoehler22:

jkoehler22:
PR: https://github.com/OpenLiberty/open-liberty/pull/9466

PR #9466, opened 2019-10-23
Fix for Issue #9465: Change mockserver to use -keysize of 2048

jkoehler22:


jkoehler22:
To test locally I changed my java security config to only allow keySize of 2048

jkoehler22:
The code change I made was to change KeyStoreFactory.class (in mockserver-core-3.10.7.jar)

private static final int FAKE_KEYSIZE = 2048;

jkoehler22:
#build

PR #9477, opened 2019-10-24
partition info restored when missing

njr-11:
It is possible, when one server finds another that is past expiry and removes its entry, that the other server might still be active and just delayed. In this case, when the second server does attempt a heart beat, it discovers that its entry is gone. Under this pull, code is added to re-create its entry.  An automated test is added as well to simulate this scenario by directly removing the entry and then checking for it to reappear automatically.


njr-11:
#build

PR #9482, opened 2019-10-25
Issue #9479: Support RMI Remote throwing Exception

### LINKED ISSUE #9479: Support RMI Remote interfaces that throw Exception
For the following method on an interface that extends java.rmi.Remote:
```
void testMethodwithException(String exceptionToThrow) throws Exception;
```
The following error is logged and the application fails to start:
```
CNTR5104E: The testMethodwithException method on the com.ibm.ws.ejbcontainer.exception.ejb.SLRemoteEx interface must be defined with the java.rmi.RemoteException exception on the throws clause.
```

The RMI specification indicates that a method must throw RemoteException or a super class of RemoteException.... so Exception should be acceptable.  Perhaps not good practice, but it should work.

tkburroughs:
JITDeploy has been updated to allow an RMI Remote method to declare throwing Exception
rather than requiring it throw RemoteException.  Spec indicates it must be RemoteException
or a parent of RemoteException.

Generatioin of EJBWrapper also updated to add catch block for RemoteException so that
it is not treated as a checked exception, like other Exception subclasses....
except when both Exception and RemoteExctption have been declared. Both Exception
and RemoteException together were allowed previously, so don't want to change the
existing behavior.

fixes #9479

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/DeploymentUtil.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/DeploymentUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2016 IBM Corporation and others.
+ * Copyright (c) 1998, 2019 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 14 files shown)


tkburroughs:
#build

tkburroughs:
#build #spawn.fullfat.buckets=com.ibm.ws.ejbcontainer.remote_fat,com.ibm.ws.ejbcontainer.exception_fat

PR #9491, opened 2019-10-25
enable and add new tests for heartbeats

njr-11:
Enable a test that was waiting for heartbeats to be implemented.
Also write a test that confirms that heartbeats are sent repeatedly.


njr-11:
#build

Issue #9492, opened 2019-10-25
ConcurrentModificationException in JAX-RS on shutdown


andymc12:
When using JAX-RS and CDI, it is possible to see a ConcurrentModificationException on app shutdown that looks like this:

```
Stack Dump = javax.servlet.ServletException: java.util.ConcurrentModificationException
    at com.ibm.ws.webcontainer.servlet.ServletWrapper.doDestroy(ServletWrapper.java:1049)
    at com.ibm.ws.webcontainer.servlet.ServletWrapper.destroy(ServletWrapper.java:1126)
    at com.ibm.ws.webcontainer.osgi.servlet.ServletWrapper.destroy(ServletWrapper.java:91)
    at com.ibm.ws.webcontainer.webapp.WebApp.destroy(WebApp.java:3939)
    at com.ibm.ws.webcontainer.osgi.webapp.WebApp.destroy(WebApp.java:1477)
    at com.ibm.ws.container.AbstractContainer.destroy(AbstractContainer.java:79)
    at com.ibm.ws.webcontainer.webapp.WebGroup.destroy(WebGroup.java:215)
    at com.ibm.ws.webcontainer.webapp.WebGroup.removeWebApplication(WebGroup.java:256)
    at com.ibm.ws.webcontainer.VirtualHost.removeWebApplication(VirtualHost.java:264)
    at com.ibm.ws.webcontainer.VirtualHost.removeWebApplication(VirtualHost.java:241)
    at com.ibm.ws.webcontainer.WebContainer.removeWebApplication(WebContainer.java:596)
    at com.ibm.ws.webcontainer.osgi.WebContainer.removeModule(WebContainer.java:1223)
    at com.ibm.ws.webcontainer.osgi.WebContainer.stopModule(WebContainer.java:1187)
    at com.ibm.ws.app.manager.module.internal.ModuleHandlerBase.undeployModule(ModuleHandlerBase.java:133)
    at com.ibm.ws.app.manager.module.internal.DeployedModuleInfoImpl.uninstallModule(DeployedModuleInfoImpl.java:64)
    at com.ibm.ws.app.manager.module.internal.SimpleDeployedAppInfoBase.uninstallApp(SimpleDeployedAppInfoBase.java:638)
    at com.ibm.ws.app.manager.war.internal.WARApplicationHandlerImpl.uninstall(WARApplicationHandlerImpl.java:81)
[...]

PR #9493, opened 2019-10-25
[9492] Synchronize access to beanManagers map

andymc12:
This resolves issue #9492 by synchronizing access to the beanManagers hash map.


andymc12:
#build

PR #9502, opened 2019-10-28
Correct encoding setting

benjamin-confino:
This fixes RTC defect 268691 by correcting the encoding statements.


benjamin-confino:
#build

PR #9509, opened 2019-10-28
add http session cache name if absent from infinispan config

njr-11:
Try to detect if the infinispan config file already has an entry for the caches that are created by HTTP sessions.  If it doesn't, add one and write to a temporary file.  This will need some further enhancement to cover a variety of ways to match.  For now, it will just identify the presence of com.ibm.ws.session.cache.


njr-11:
#build

PR #8838, opened 2019-09-04
Change mvn install plugin and install utility to use new resolve as se…

TayyabDev:

PR #9526, opened 2019-10-29
Revert "Change mvn install plugin and install utility to use new resolve as se…"

tevans78:
Reverts OpenLiberty/open-liberty#8838

May have caused https://github.com/OpenLiberty/open-liberty/issues/9505

#build


alvinso:
I tested the PB manually and the problem of installing auto feature is gone.

PR #9530, opened 2019-10-29
Remove static value and kernel.boot.java7 project

tjwatson:
The StaticValue class is no longer needed and should be removed to get rid of the complication it introduces.  Also there is no longer a reason to separate out Java 7 classes into com.ibm.ws.kernel.boot.java7


tjwatson:
#build

PR #9548, opened 2019-10-29
better identification of http session cache name patterns

njr-11:
Add matching logic to match configured Infinispan cache names that include the wild card (*) character.


njr-11:
#build

PR #9552, opened 2019-10-29
9392 batch remote partition recovery

scottkurz:
### Mark RemotablePartition consumed 
(JPAPersistenceManagerImpl.java, MemoryPersistenceManagerImpl.java, BatchJmsEndpointListener.java)

* update to consumed, serverid, restURL when step thread exec created
* in BatchJmsEndpointListener.java,  abort `if (!rpState.equals(WSRemotablePartitionState.QUEUED)) {`
* we maintain memory persistence path for test purposes.. could be used by customer to test batch JMS config as well.. obviously limited usefulness of having a single-JVM-scoped repository
	
### Fix PSU creation, Oracle DB old tables
(JPAPersistenceManagerImpl.java, DatabaseStoreImpl
* createPSU   build the list of entities to pass to DBStoreImpl such that it only includes one for each entity type (instance, execution) and refactored the logic in DBStoreImpl to account for this.   This is in contrast with the recent idea of passing multiple entities in certain cases and having each trigger a different stanza within DBStoreImpl
* in DBStoreImpl also added comments clarifying the major logic branches
* Acknowledge  "ORA-00942" as a message which indicates new Oracle DB table hasn't been created

### Poll for recovered partitions

The problem with aborting a non-QUEUED RemotablePartition is that it prevents a redelivery from producing a failure msg which gets sent back as a reply msg to the top-level job.  (One problem with trying to consume such a message is that it could interfere with a partition currently executing, which is why we made the change in the first place.)

In **dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/PartitionedStepControllerImpl.java** we have changes to:
* switch from local vars to fields in a couple places (e.g.  currentPlan,  analyzerExceptions) to reduce passing objs around 
    * **NOTE**: This eliminates entry/exit logging so make sure we re-add anything good we'd lose
* throw JobStoppingException if job is stopped while a partition is starting (better flow of control instead of passing back return codes)
* keep an extra  finishedPartitionNumbers data structure for a Set view to compare against
* In waitForNextPartitionToFinish loop
    * changed to while(true) loop 
    * receivedLastMessageForPartition=> isFinalMessageForPartition
        * broke apart test for partition end vs. handling of this condition
    * add checkForRecoveredRemotePartitions() 

### Misc. refactoring

The rest is all fairly minor, deleting things no longer used, etc., removing some trivial trace.  

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/execution/impl/JobExecutionHelper.java**
* remove no-op  createPartitionExecution (seems confusing to have it)
* format

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/execution/impl/RuntimePartitionExecution.java**
* remove no-op  updatePartitionExecution (seems confusing to have it)

JPA entities 
* trivial trace

JPA entity extractors
* rename:  "table version" => "entity version"

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/persistence/jpa/JobExecutionEntityV3.java**
* remove duplicate @OneToMany field

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/persistence/jpa/RemotablePartitionEntity.java**
* remove redundant uniqueConstraints

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/services/IPersistenceManagerService.java**
* remove unsed methods, create/get by RPKey, 

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/services/impl/AbstractPersistenceManager.java**
* remove other illegal state values to emphasize this is only preventing move out of COMPLETED/ABANDONED
    * we **DO NOT** want to start disallowing other state transitions we think shouldn't happen since this effects the JMS executor redelivery path.

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/RemotablePartitionState.java → dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSRemotablePartitionState.java**
* rename to follow class naming convention

**dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/impl/WSJobRepositoryImpl.java**
* method name, PK refactoring
* fix of failure to use msg bundle in  "BATCH_SECURITY_NOT_ACTIVE"   (could lead to FAT failures though you'd think I'd have caught it if my test builds were broad enough)

**dev/com.ibm.ws.jbatch.jms/src/com/ibm/ws/jbatch/jms/internal/dispatcher/BatchJmsDispatcher.java**
* passing around an "exceptionMessage" as parm can be a bit misleading when looking at trace, say FAT failures... for someone on the L3 team somewhat new to the code.

### DIFF
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/PartitionedStepControllerImpl.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/PartitionedStepControllerImpl.java
@@ -464,7 +464,7 @@ public class PartitionedStepControllerImpl extends BaseStepControllerImpl {
         getPartitionReplyQueue().close();
 
         // Deal with the results.
-        checkFinishedPartitions();
+        checkFinishedPartitionsForFailures();
 
     }
 
@@ -667,7 +667,7 @@ public class PartitionedStepControllerImpl extends BaseStepControllerImpl {
     /**
      * Issue message for partition finished and add it to the finishedPartitions list.
      */
-    private void partitionFinished(PartitionReplyMsg msg) {
+    private void partitionFinishedReplyMessageReceived(PartitionReplyMsg msg) {
         JoblogUtil.logToJobLogAndTraceOnly(Level.FINE, "partition.ended", new Object[] {
                                                                                          msg.getPartitionPlanConfig().getPartitionNumber(),
                                                                                          msg.getBatchStatus(),
@@ -799,7 +799,7 @@ public class PartitionedStepControllerImpl extends BaseStepControllerImpl {
                             analyzerExceptions.add(t);
                         }
                         if (isFinalMessageForPartition(msg)) {
-                            partitionFinished(msg);
+                            partitionFinishedReplyMessageReceived(msg);
                         }
                     } else {
                         //If no messages left, break the loop
@@ -840,7 +840,7 @@ public class PartitionedStepControllerImpl extends BaseStepControllerImpl {
             }
 
             if (isFinalMessageForPartition(msg)) {
-                partitionFinished(msg);
+                partitionFinishedReplyMessageReceived(msg);
                 break;
             }
             // else keep looping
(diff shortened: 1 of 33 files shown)


scottkurz:
#run-libby-bot

scottkurz:
#build

scottkurz:
#build

PR #9580, opened 2019-10-31
enable main test bucket for for Infinispan 10.0.0.Final

njr-11:
Update our test dependencies to Infinispan 10.0.0.Final, which just recently became available.
Unfortunately, after doing so, things don't work so well.  One of the main issues was a new hard dependency on mpMetrics, which this pull covers for two test classes in the main bucket that I was able to get working.  There is complication with CDI annotations in the Infinispan JARs that prevent applications from being able to declare libraryRef to the Infinispan JARs when CDI is enabled, which is all of the time due to mpMetrics.


njr-11:
#build

PR #9623, opened 2019-11-04
Remove xtrace

benjamin-confino:


benjamin-confino:
This fixes RTC defect 269182

PR #9627, opened 2019-11-04
fix the NPE when the algorithm field is not provided in a Digest

### LINKED ISSUE #9607: NPE in the SIP Container when a Digest challenge does not contain the `algorithm` field
An NPE is thrown in the Liberty SIP Container when a Digest challenge does not contain the `algorithm` field. The field is optional and the SIP Container should default to `MD5` when it's not provided.

annaguri:
fixes #9607


annaguri:
#get-funky

annaguri:
tested the fix with `sipp`.

PR #9631, opened 2019-11-04
Don't reuse channels between tests

tevans78:
#build


tevans78:
There were some failures in the PB, however they are testcontainer issues that I have seen before and not related to this change. It may be something to do with running the same bucket 10 times in quick succession. So on balance I think it is better to merge this PR.

PR #9651, opened 2019-11-05
Add scope to app token

brutif:


brutif:
oops, picked up Adam's new stuff, need a clean pull, closing

Issue #9648, opened 2019-11-05
app-tokens created from personalTokenMangement UI have incorrect scope


brutif:
An access token generated from the OpenId Connect Server's personalTokenManagement UI endpoint has empty scopes.  It should contain the preauthorized scopes of the OAuth client specified by the internalClientId in server.xml

brutif:
#9652 has merged, so closing

PR #9652, opened 2019-11-05
add scope to app token

brutif:


brutif:
resolves #9648 

#build

PR #9655, opened 2019-11-06
Batch collector properties, zero-partition partitioned step, status passed to reducer - fixes #9645, #9707, #9708

### LINKED ISSUE #9645: PartitionCollector getting injected with the first partition's substitution val on all partitions, using partitionPlan substitution
For JSL (XML) like this (excerpt):

         <step> 
		<partition>
			<plan partitions="3" >
				<properties partition="0">
					<property name="idx" value="p0"/>
				</properties>
				<properties partition="1">
					<property name="idx" value="p1"/>
				</properties>
				<properties partition="2">
					<property name="idx" value="p2"/>
				</properties>
			</plan>
                        <collector ref="Collector">
				<properties>
					<property name="myProp" value="#{partitionPlan['idx']}" />
				</properties>
                       </collector>

the collector's "Item" property is supposed to be parameterizable with a different value on each partition, supplied via the partition plan properties corresponding to that partition.

That is, in the **Collector** artifact we should have:

        @Inject
        @BatchProperty
        String myProp;

injected with the values, `p0`, `p1`, `p2` on each of the three partitions, respectively.

Instead we are wrongly passing the value for the "first" partition (I put that in quotes since I'm not 100% sure if it's partition #0, e.g. the first one in the plan, or if it's the first one that happens to get executed, since of course each gets executed on a different thread)... so unsure precisely what I mean by "first" but the problem is clear enough.

### LINKED ISSUE #9707: Batch job partitioned step doesn't allow boundary case of steps with zero partitions
**Describe the bug**
A batch job with a partitioned step defined with '0' partitions will fail early in the step execution.  This happens whether a **PartitionMapper** is used to define the number of partitions dynamically or a **<plan>** element is used to define the number of partitions statically. 

One might get an error looking like:

> 0000008d id=         .ibm.jbatch.container.controller.impl.BaseStepControllerImpl : An exception occurred while running the step step1. -
> java.lang.IllegalArgumentException: Partition instances value must be 1 or greater in stepId: step1, with instances=0

**Steps to Reproduce**

Define a partitioned step with a `<mapper>` element pointing to a **PartitionMapper** implemented with something like:

        @Override
        public PartitionPlan mapPartitions() throws Exception {

            PartitionPlanImpl plan = new PartitionPlanImpl();
            plan.setPartitions(numPartitions);     //   numPartitions = 0

You probably wouldn't hard-code `numPartitions` above to '0' but it could be the degenerate or boundary case within your logic, and maybe the step-level artifacts like the StepListener or PartitionReducer are still useful to run in such a case.

You could also have a `<plan>` configured with:

    <partition>
	<plan partitions="0"/>

**Expected behavior**

Expect step to reach COMPLETED status and StepListener and PartitionReducer to run (if each configured) .

**Diagnostic information:**
 - OpenLiberty Version: Any in the past few years.
 - server.xml configuration:  **batch-1.0** and/or **batchManagement-1.0** features

### LINKED ISSUE #9708: For a batch job with partitioned step, the PartitionReducer's afterPartitionedStepCompletion gets ROLLBACK on normal completion.
**Describe the bug**

For a batch job with a partitioned step, with a **PartitionReducer** configured, during normal completion the reducer's `afterPartitionedStepCompletion` gets passed a **PartitionStatus.ROLLBACK** value even when a step completes normally.

**Steps to Reproduce**
Define a partitioned step with a reducer configured, like:

	<step id="step1">
		<chunk/>
		<partition>
			<plan partitions="3"/>
			<reducer ref="MyReducer">

and observe the value passed to **MyReducer**'s    `afterPartitionedStepCompletion(PartitionStatus status)` method upon normal completion.

**Expected behavior**

For a successful execution, the status passed to the reducer in `afterPartitionedStepCompletion` would be **COMMIT**, and for a failing execution it would be **ROLLBACK**

**Diagnostic information:**
* OpenLiberty Version: Any in the past few years.
* server.xml configuration: batch-1.0 and/or batchManagement-1.0 features

scottkurz:
This fixes #9645, fixes #9707, fixes #9708.

### DIFF
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/ExecutionElementControllerFactory.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/ExecutionElementControllerFactory.java
@@ -1,13 +1,13 @@
 /*
  * Copyright 2012 International Business Machines Corp.
- * 
+ *
  * See the NOTICE file distributed with this work for additional information
- * regarding copyright ownership. Licensed under the Apache License, 
+ * regarding copyright ownership. Licensed under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


scottkurz:
#build

scottkurz:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_MIjnsAGtEeqlbYcQx7mKkA

scottkurz:
#libbybot run

PR #9667, opened 2019-11-06
Add a check to see if the executor has been shutdown

tevans78:
#build 

On a previous test run there was an exception seen on shutdown
```
Caused by: java.util.concurrent.RejectedExecutionException: CWWKE1202E: A task cannot be submitted because the executor managedScheduledExecutorService[DefaultManagedScheduledExecutorService] (concurrencyPolicy[defaultConcurrencyPolicy]) has been shut down.
    at com.ibm.ws.threading.internal.PolicyExecutorImpl.enqueue(PolicyExecutorImpl.java:521)
    at com.ibm.ws.threading.internal.PolicyExecutorImpl.submit(PolicyExecutorImpl.java:1287)
    at com.ibm.ws.concurrent.internal.ManagedExecutorServiceImpl.submit(ManagedExecutorServiceImpl.java:559)
    at com.ibm.ws.concurrent.internal.ManagedExecutorServiceImpl.submit(ManagedExecutorServiceImpl.java:565)
    at com.ibm.ws.microprofile.reactive.messaging.kafka.KafkaInput.pollKafkaAsync(KafkaInput.java:212)
    at com.ibm.ws.microprofile.reactive.messaging.kafka.KafkaInput.lambda$createPublisher$1(KafkaInput.java:95)
    at com.ibm.ws.microprofile.reactive.messaging.kafka.KafkaInput$$Lambda$325.00000000300186D0.apply(Unknown Source)
    at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1106)
    ... 29 more
```

### DIFF
--- a/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaInput.java
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaInput.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
 
@@ -166,7 +167,7 @@ public class KafkaInput<K, V> {
         }
     }
 
-    @FFDCIgnore(WakeupException.class)
+    @FFDCIgnore({ WakeupException.class, RejectedExecutionException.class })
     private CompletionStage<PublisherBuilder<ConsumerRecord<K, V>>> pollKafkaAsync() {
         if (!this.subscribed) {
             this.lock.lock();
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaInput.java
```diff
@@ -209,9 +209,11 @@ public void shutdown() {
...
             result.complete(fromIterable(records));
         } else {
-            this.executor.submit(() -> {
-                executePollActions(result);
-            });
+            if (this.running && !this.executor.isShutdown()) {
+                this.executor.submit(() -> {
+                    executePollActions(result);
+                });
+            }
```
Azquelt: I would suggest a few changes here:
1. Catch the exception rather than checking `executor.isShutdown()` to eliminate the small window between that call and the call to `submit` where the executor could stop accepting new tasks.
2. Don't worry about checking `this.running`, that's already checked all over the place
3. Assume that a `RejectedExecutionException` means that the application is shutting down and treat it in the same way as the case on line 185 (trace and return an incomplete future because we're shutting down)

tevans78: done

tevans78:
#build

PR #9683, opened 2019-11-06
enable more tests for failover code path

njr-11:
add comment here


njr-11:
#build

dev/com.ibm.ws.concurrent.persistent_fat/fat/src/com/ibm/ws/concurrent/persistent/fat/PersistentExecutorWithFailoverEnabledTest.java
```diff
@@ -0,0 +1,114 @@
...
+    	DatabaseContainerUtil.setupDataSourceProperties(server, testContainer);
+    	
+		//Initialize database
+		DatabaseContainerUtil.initDatabase(testContainer, DB_NAME);
+
+		//Add application to server
+        ShrinkHelper.defaultDropinApp(server, APP_NAME, "web");
+
+        //Create Derby DB if using derby
+        if (dbContainerType == DatabaseContainerType.Derby)
```
KyleAure: I do not believe this is necessary.  I see that I had this step in the original `SchedulerTest` class, and I don't think it will hurt anything.  but I might go back and remove it.

Issue #9456, opened 2019-10-23
EJB Custom Bindings: Investigate 'ejblocal:' added to simple-binding-name in xml


olendvcook:
After starting to look at our implementation I discovered that the jndiName element in xmi and the simple-binding-name element in xml go to the same variable in BeanMetaData. I'm pretty sure this is true, be should confirm all the rules discovered in #8999 apply to simple-binding-name.

olendvcook:
We check module version before binding to local namespace:

```
if (local && priorToVersion3)
      {
         // A module version that is prior to EJB 3, so also bind the home into
         // the local: namespace.
         Context localCtx = getLocalEJBContext();
         if (localCtx != null)
         {
            String jndiName = bmd.ivLocalHomeJndiName;
            ivLastLocalSubcontextCreated = recursiveCreateSubcontexts(
                  localCtx, jndiName, 0, ivLastLocalSubcontextCreated);
            bindObject(localCtx, jndiName, bindingObject); // d660700
         }
      }
```
but should still investigate what happens when you stick `ejblocal:` and `local:` in front.

olendvcook:
Conclusion:

 1. If SimpleBindingName does not have a namespace written in, it will be bound to ejblocal: only, local: and local:ejb are for jndiName element normally.

 2. If simple-binding-name="ejblocal:<name> nothing changes

 3. If simple-binding-name="local:<name> **it will only be bound to local:** (even though these are 3X beans)
 - also, the binding will not have ejb stuck in front, the lookup is local:<name> not local:ejb/<name>
 - or with local: context it is <name> not ejb/<name>
 
 4. If simple-binding-name="local:ejb/<name> it will be bound in local: and local:ejb
  - the local: lookup will not have double ejb stuck in front
  - local:ejb/<name>
  - with local: context: ejb/<name>
  - with local:ejb context: <name>
 
 5. If you do simple-binding-name="<namespace>:<namespace>:<name> it will ignore up to the innermost namespace.
  - simple-binding-name="ejblocal:local:ejb/com/ibm/ejb3x/ejbinwar/webejb3x/Stateless2xLocalHome" would use the local:ejb rule above

 6. You can do a lookup and chain as many ejblocal: and local: in front as you want.

olendvcook:
Using this issue to deliver tests

olendvcook:
tests merged

Issue #8777, opened 2019-08-28
EJB Custom Bindings: Support simple-binding-name XML element


brideck:
- Implement support for simple-binding-name in ibm-ejb-jar-bnd.xml for both remote and ejblocal:
- Activate simple-binding-name tests in NoInterfaceBndWeb.war

Simple binding name (one name used for both local and remote) 
`<session name="AccountServiceBean" simple-binding-name="ejb/AccountService"/>`

nkuzman:
Any news/ETA on this?

olendvcook:
@nkuzman Hello, just saw your question. This should be working for **LOCAL EJBs** as of 20.0.0.2 (in a beta functionality fashion). Unfortunately, getting it to work for remote EJBs is taking a little longer, hopefully soon though.

Note that even if we get it to start working for the simple-binding-name for remote EJBs. the entire feature (#7918) Is still going to take a while and it would be considered beta until we're finished with that.

For now if you just need it for local EJBs and you have 20.0.0.2, you can enable the beta feature by adding `<ejbContainer bindToServerRoot="true"/>` to your server.xml and have the EJBHome feature added as well (`<feature>ejbHome-3.2</feature>`)

nkuzman:
Hi @olendvcook , thank you for your answer.
I mostly use remote EJBs, so apparently I'll just have to wait for it a bit longer :)

olendvcook:
@nkuzman I have delivered support for the remote simple-binding-name and it should be in the 2.0.0.5 beta.

PR #9686, opened 2019-11-07
EJB SimpleBindingName xml Namespace Tests

olendvcook:
Delivery of disabled tests that should be enabled and working when delivering #8777

for #9456


olendvcook:
#build

PR #9695, opened 2019-11-07
changed ActiveURL to be current URL

### LINKED ISSUE #9694: Active URL isn't being set correctly after change
**Describe the bug**
In ContextManager.java, in createContextPool, the provider URL which, is the current URL of the LDAP server, should also be the activeURL. However the active URL does not change when the provider URL falls back to the failover server.

**Steps to Reproduce**
1. Create an ldap server with failover servers.
2. Failover to the backup server.
3. Create a new context pool with the failover server as the current provider URL.
4. The active URL will still show that the initial server URL is the active url still.

**Expected behavior**
The activeURL variable should be updated to be the backup server URL if the priimary URL shuts down.

agiridharan:
fixes #9694


agiridharan:
#build

PR #9712, opened 2019-11-08
Revert "Smaller change to switch from Java 7 to Java 8 as minimum Java level"

mhldr:
Reverts OpenLiberty/open-liberty#8634 
I'm looking to re-enable Java 7 support briefly


mhldr:
As requested I'm not going to deliver this.

Issue #9717, opened 2019-11-08
SessionStatistics :  change the default value of ModifyActiveCountOnInvalidatedSession


zhang-ca:
Session active counter may not reduce due to session become invalid by other application, change the default value of ModifyActiveCountOnInvalidatedSession to true.

When sessions are shared with multiple applications, ModifyActiveCountOnInvalidatedSession should default to false. However, most applications are not shared session id, therefore default should set to true.

zhang-ca:
Ellen, can you make the change (in SessionManagerConfig.java) for me. Thanks.

PR #9727, opened 2019-11-11
Changing default value of ModifyActiveCountOnInvalidatedSession to be…

ellen-lau:
… true

fixes: #9717

#build


ellen-lau:
Failures don't look related to the changes made.

PR #9738, opened 2019-11-12
externalize string for not found page for OL splash page

aknguyen7:
1. externalize string - translate message for notFound page
2. fix DAP error


aknguyen7:
#build

PR #9747, opened 2019-11-12
Rename `@SchemaName` to `@Name` - as changed in the MP GraphQL spec

andymc12:
The MP GraphQL community decided to rename the `@SchemaName` annotation to `@Name` - this PR implements that change.

At the same time, the MP GraphQL TCK added tests that allow the `@Name` annotation to be used to name a query method if no name is specified in the `@Query` annotation.  This change is also reflected in this PR.


andymc12:
#build

PR #9748, opened 2019-11-12
Generate a server dump if the service fails to activate

brenthdaniel:
Changes a test case in the persistent_delayexec FAT to generate a server dump when the TestService fails to activate. This is intended to help resolve an extremely rare but longstanding intermittent test failure.


brenthdaniel:
#build

PR #9782, opened 2019-11-14
update fattest.simplicy to allow specifying multiple test classes/met…

WhiteCat22:
…hods when running specific tests as well as allowing wildcard matching and removing the need to fully qualify the package of the test class.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/custom/junit/runner/TestNameFilter.java
+++ b/dev/fattest.simplicity/src/componenttest/custom/junit/runner/TestNameFilter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013 IBM Corporation and others.
+ * Copyright (c) 2013, 2019 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -10,14 +10,13 @@
  *******************************************************************************/
 package componenttest.custom.junit.runner;
 
+import java.util.regex.Pattern;
+
 import org.junit.runner.Description;
 import org.junit.runner.manipulation.Filter;
 
 import com.ibm.websphere.simplicity.log.Log;
 
-/**
- *
- */
 public class TestNameFilter extends Filter {
 
     private static final String FAT_TEST_QNAME;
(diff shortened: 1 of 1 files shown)


dev/fattest.simplicity/src/componenttest/custom/junit/runner/TestNameFilter.java
```diff
@@ -30,9 +34,42 @@
...
+            String methodList = new String();
+            for (String method : methods) {
+                if (methodList.length() > 0) {
+                    methodList += "," + method;
+                } else {
+                    methodList = method;
+                }
+            }
+
+            FAT_TEST_CLASS = classList;
```
aguibert: I don't think it's necessary to check for comma-separated list for `FAT_TEST_QNAME`, since the property seems to imply that you must supply the fully qulaified name of the test class and method, doesn't seem likely that people will want to do multiples here

dev/fattest.simplicity/src/componenttest/custom/junit/runner/TestNameFilter.java
```diff
@@ -55,13 +92,47 @@ public String describe() {
...
+        System.out.println(arg + " " + match);
+        return match;
+    }
+
+    private boolean compare(String pattern, String arg) {
+        if (pattern.contains("*")) {
+            return wildcardMatch(pattern, arg);
+        } else {
+            return arg.equals(pattern);
+        }
```
aguibert: instead of branching here we could just call `wildCardMatch` in all cases, and when there are no `*` it would be the same as .equals()

dev/fattest.simplicity/src/componenttest/custom/junit/runner/TestNameFilter.java
```diff
@@ -55,13 +92,47 @@ public String describe() {
...
+                    int indx = arg.lastIndexOf(".");
+                    arg = arg.substring(indx + 1);
+                }
+                if (compare(s, arg)) {
+                    match = true;
+                    break;
+                }
+            }
+        }
+        System.out.println(arg + " " + match);
```
aguibert: this print statement can be removed

dev/fattest.simplicity/src/componenttest/custom/junit/runner/TestNameFilter.java
```diff
@@ -55,13 +92,47 @@ public String describe() {
...
         return true;
     }
 
+    private boolean match(String list, String arg) {
+        boolean match = false;
+        for (String s : list.split(",")) {
+            if (s.contains(".")) {
+                if (compare(s, arg)) {
+                    match = true;
+                    break;
```
aguibert: instead of having a boolean and break here you can just `return true`

dev/fattest.simplicity/src/componenttest/custom/junit/runner/TestNameFilter.java
```diff
@@ -55,13 +92,47 @@ public String describe() {
...
+                }
+            } else {
+                if (arg.contains(".")) {
+                    int indx = arg.lastIndexOf(".");
+                    arg = arg.substring(indx + 1);
+                }
+                if (compare(s, arg)) {
+                    match = true;
+                    break;
```
aguibert: same comment here about doing a `return true`

WhiteCat22:
#build

WhiteCat22:
updated copyright header date since I made a big change to this file.

I tested this manually by running tests locally with comma separated lists of test classes and test methods. I also tested fully qualified test classes, test class shortnames, wildcarding test classes and wildcarding test names.

WhiteCat22:
#run-libby-bot

PR #9802, opened 2019-11-15
9214 Support alternate symbolic and short names for public features

### LINKED ISSUE #9214: Ability to rename future versions of features
Technology names change which sometimes map onto our public Liberty feature names.  For example the following liberty public features:

jsf-2.2
jsf-2.3
jsfContainer-2.2
jsfContainer-2.3

JSF is likely to be renamed to just Faces.  So for version 2.4 we would have public features:

faces-2.4
facesContainer-2.4

The new faces named features must still be singletons and they cannot be allowed to be configured with the previous jsf named features.

The following must be allowed to work WRT to tolerates when another feature includes one of these features:

```
com.ibm.websphere.appserver.jsf-2.3; ibm.tolerates:="2.2,2.4"
```
and
```
com.ibm.websphere.appserver.faces-2.4; ibm.tolerates:="2.2,2.3"
```

In other words a feature must be able to tolerate versions across the required feature rename.  In the first example a feature prefers jsf-2.3 but can tolerate jsf-2.2 and faces-2.4.  In the second example a feature prefers faces-2.4 but can tolerate jsf-2.2 and jsf-2.3.

The solution that comes to mind is one that involves adding some kind of alternate name metadata to each feature that has a previous and/or future names.  Here the jsf features would have to specify com.ibm.websphere.appserver.faces as their alternate name and the faces feature would have to specify com.ibm.websphere.appserver.jsf as its alternate name.

Both seem needed in cases where a server is minified to only include one of the versions of the renamed feature.  In that case we need to know the alternate name for each incase the previous named feature is only available or if only the future named feature is available.

This will involve enhancing the feature meta-data as well as the Liberty feature resolver.  The feature resolver is used by others outside of the Liberty runtime and such users will need to be aware of the new meta-data if they want to model the feature resolution of the renamed features accurately like the runtime does.

dazavala:
Fixes #9214

The solution to date enhances the feature manager to resolve and provision public features using alternate symbolic and short names, i.e. aliases, defined in feature metadata.

Here are the details and issues found:

**The solution assumes renamed feature metadata declare a new symbolicName and shortName, and declare the previous symbolicName and shortName as the symblicAlias and shortAlias.**

ISSUE: We could rename features by declaring new names as aliases, rather than modify the symbolic and short names. The current solution assumes that symbolic and short names are preferred for the release, which is implicit information. Preferred names will appear in messages because they are specific to a release, and it's not clear whether logs may contain the previous feature names (e.g. jsf, jakarta-ee agreements.)

ISSUE: Will we support aliases for any non-public features?  We'll need to support protected features, seeing that stack product features probably declare dependencies on renamed features. The solution requires a few tweaks to support protected features.

ISSUE: How to handle equivalent feature names and aliases in server.xml. The solution ignores them by quietly installing the preferred name.  E.G. <featureManager>jsf-2.3, faces-2.3</featureManager>

Feature metadata now may include attributes `WLP-SymbolicName-Alias` and `WLP-ShortName-Alias`, where each may declare 0 or more aliases for Subsystem-SymbolicName and IBM-ShortName, respectively. For brevity, let's refer to these as symbolicAlias, shortAlias. symbolicName, shortName.

ProvisioningFeatureDefinition (intf), SubsystemFeatureDefinitionImpl, and FeatureDefinitionUtils now collect the symbolicAlias and shortAlias from feature manifests, and provide access to both aliases. Aliases are supported for public features, only.  New unit tests are in FeatureDefinitionUtilsTest.

Support to match aliases declared in auto-feature capabilities was added to SubsystemFeatureDefinitionImpl, see method isCapabilitySatisfied().

The solution does not modify the FeatureResource intf.

FeatureRepository now populates the publicFeatureNameToSymbolicName map with elements that map a symbolicAlias and shortAlias to its feature definition, in addition to symbolicName and shortName. 

This enables its implementation of method `FeatureResolver.Respository.getFeature(byName)` to return features by alias in addition to name. Method getFeature() is now documented with the new behavior, as similar support will be required by external repositories that implement the Repository interface. Lastly, the class now reads/writes alias metadata from/to the feature cache.

FeatureResolverImpl now supports symbolic names and aliases declared in Subsystem-Content (i.e. dependencies,) where the dependency may tolerate a version of the feature that was renamed.  The solution requires renamed features have an updated symbolicName, and the previous name is declared as an alias.

Unit tests for toleration support were added to FeatureResolverInterfacesTest. A new FAT case, AliasTest, was added to the kernel feature FAT bucket. AliasTest verifies alias support for auto-feature capabilities, non-support of non-public and user features, static feature updates (restart server) over clean vs dirty workareas; dynamic updates to add and remove features, and server configurations containing equivalent feature name and alias.

### DIFF
--- a/dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureResolverImpl.java
+++ b/dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureResolverImpl.java
@@ -366,7 +366,11 @@ public class FeatureResolverImpl implements FeatureResolver {
             for (FeatureResource included : includes) {
                 String symbolicName = included.getSymbolicName();
                 if (symbolicName != null) {
-                    String[] nameAndVersion = parseNameAndVersion(included.getSymbolicName());
+                    ProvisioningFeatureDefinition featureDef = selectionContext.getRepository().getFeature(symbolicName);
+                    if (featureDef != null && !!!symbolicName.equals(featureDef.getSymbolicName())) {
+                        symbolicName = featureDef.getSymbolicName(); // Use actual symbolic name, not alias
+                    }
+                    String[] nameAndVersion = parseNameAndVersion(symbolicName); //included.getSymbolicName());
                     String baseName = nameAndVersion[0];
                     includedBaseFeatureNames.add(baseName);
                 }
@@ -396,7 +400,8 @@ public class FeatureResolverImpl implements FeatureResolver {
         }
     }
 
-    private void processIncluded(ProvisioningFeatureDefinition includingFeature, FeatureResource included, Set<String> allowedTolerations, Deque<String> chain, Set<String> result,
+    private void processIncluded(ProvisioningFeatureDefinition includingFeature, FeatureResource included, Set<String> allowedTolerations,
+                                 Deque<String> chain, Set<String> result,
                                  SelectionContext selectionContext) {
         String symbolicName = included.getSymbolicName();
         if (symbolicName == null) {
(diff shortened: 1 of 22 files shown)


dazavala:
#build

dazavala:
#build (revision)

dazavala:
#build

dazavala:
Pushed changes to ActivationType test; awaiting overall review feedback for delivery.

dazavala:
Was holding onto to this in the event we decide to adopt feature aliasing.  That will likely never happen as the feature model has proven capable thus far.

PR #9822, opened 2019-11-18
Update the FT TCK trace spec

Azquelt:
Remove `APPCONFIG` because it's very noisy and rarely helpful

Add `Threading=event` so that we can see if pauses in fault tolerance
tests are caused by running out of threads. Will also give us visibility
of whether the expected number of threads are running at any point in
the test run.


Azquelt:
#build

PR #9826, opened 2019-11-18
Fix activation type test

dazavala:
Add assertions to the ActivationType test in the feature manager FAT.


dazavala:
Opened in error

PR #9829, opened 2019-11-18
testing for new failover approach via the test bucket that simulates failover with multiple executors on one server

njr-11:
Update the test bucket to also run tests when using the new failover approach.  The two tests for heartbeating are excluded in this case because the new failover approach does not involve heartbeating.


njr-11:
#build

njr-11:
#build

PR #9853, opened 2019-11-19
ensure cache is used before shutdown in tests

njr-11:
There is some asynchronous logic after processing a configuration update where http sessions code is accessing the CacheManager.  When a test case performs server shutdown after this point without any subsequent operations on a session, there is a timing window for this CacheManager initialization to overlap the shutdown and cause errors/warnings to appear in the logs, which causes the test case to report a failure.  This pull updates the test case to work around this by forcing an operation to access a session prior to invoking shutdown.


njr-11:
#build

PR #9859, opened 2019-11-19
Upgrade MP GraphQL TCK to get added diagnostics

andymc12:
Upgrading to a newer snapshot of MP GraphQL API and TCK - only the TCK has changed, but upgrading the API to keep them in sync.  The updated TCK contains diagnostics for a problem occurring in the z/OS builds.

Not a release bug.


andymc12:
#build

PR #9882, opened 2019-11-20
Add runtime update trace by default

brenthdaniel:
Enable runtime update trace by default so that we have information when some random test bucket gets a quiesce failure. This is a trivial amount of trace when there are no quiesce failures.


brenthdaniel:
#build

Issue #9506, opened 2019-10-28
EJB Custom Bindings: Investigate 'ejblocal:' added to binding-name in xml


olendvcook:
After starting to look at our implementation I discovered that the jndiName element in xmi and the simple-binding-name element in xml go to the same variable in BeanMetaData. I'm pretty sure this is true, be should confirm all the rules discovered in #8999 apply to binding-name.

And then possibly creating a new issue to add tests like #9403

olendvcook:
discovered that like local|remote-home-binding-name, binding-name validates that local interfaces start with `ejblocal:` and that remote do not.

olendvcook:
Tests merged

PR #9894, opened 2019-11-20
EJB Custom Bindings binding-name element tests

olendvcook:
Delivery of disabled tests for testing namespace combinations added to binding-name element in ejb-jar-bnd.xml

for #9506


olendvcook:
#build

PR #9899, opened 2019-11-20
Removed unnecessary derby call

gjwatts:
Removed the `DerbyEmbeddedUtilities.createDB` call.  This is unnecessary since we are setting the `createDatabase="create"` property on the datasource in server.xml.


gjwatts:
#build

PR #9908, opened 2019-11-21
test where transaction blocks after cancelling task

njr-11:
Write a test of the scenario where a user cancels a task within a transaction but doesn't commit right away, so that the task entry remains locked.  It turns out this blocks a number of other operations (even those for different tasks).  For now, I'm delivering the test case with portions commented out where we would like to see what improvements can be made, such that more the test can be enabled.


njr-11:
#build

njr-11:
I accidentally pushed another commit onto this pull even though it had a good build (it was held out by the buildcon), so need to request another build and run of the copyright checks:
#build
#run-libby-bot

njr-11:
#run-libby-bot

njr-11:
Build has 4 test failures, which are pre-existing issues in unrelated componentry (ui)

PR #9934, opened 2019-11-25
Improve test-container docker reporting

KyleAure:
Improve the way our test infrastructure reports an unavailable docker environment


KyleAure:
Closing PR.  This was not the correct solution for improving reporting.

PR #9962, opened 2019-11-26
add blocking tests for finder operations and more

njr-11:
Add tests for blocking a transaction after it perform a find-by-name or find-by-taskID operation and determining if other operations to the task are permitted.  Also, includes versions of the cancel/remove test that do cancel/remove by task ID instead of by name.  Adds a couple of tests for operations that ought to be avoided by proper configuration of the missed task threshold, but where we would like to have it time out quickly (code doesn't do this yet) if a task takes too long and causes this code path to be reached.  This last two tests are commented out, for future enablement.


njr-11:
#build

PR #9971, opened 2019-11-27
test that schedules, runs, cancels, and removes a task while another task blocks during execution

njr-11:
Add test coverage to ensure that one blocked task does not hold up scheduling, execution, cancellation, and removal of another task.


njr-11:
#build

PR #9974, opened 2019-11-28
Remove reinstateHADBLock optimisation

### LINKED ISSUE #9973: reinstateHADBLock() should not be called
I introduced the reinstateHADBLock() method in SQLMultiScopeRecoveryLog under issue #8047 to reinstate the home server name in the recovery log control row as a performance optimisation. Unfortunately there are situations where a peer server may have stalled rather than died, where the home server MUST know that a peer has performed recovery for it. I will remove this method under this issue.

neilgsyoung:
Fixes #9973


neilgsyoung:
#build

Issue #9878, opened 2019-11-20
Ensure `@Id` on MP GraphQL entities are honored


andymc12:
For the upcoming MP GraphQL epic issue #7956 - the `@Id` annotation on fields must be honored - i.e.:
```
public class TaxPayer {
    @Id
    private String taxpayerID;
    private long salary;
    // ... include getters and setters
}
```

should produce a GraphQL schema like:
```
type TaxPayer {
   taxpayerID: ID
   salary: Long
    ...
}
```

But the current implementation ignores the `@Id` annotation on fields (and setters/getters).

PR #10001, opened 2019-12-03
[9878] Handle `@Id` on entity fields

andymc12:
This allows `@Id` to be specified on fields/methods and be properly
reflected as ID types in the generated schema.

Because this change pulls in the latest version of the API/TCK, this change also includes the renaming of `@InputType` to `@Input` in the runtime and FAT test case code.


andymc12:
This resolves issue #9878 - #build

PR #10026, opened 2019-12-04
warnings in testMultipleInstancesCompeteToRunManyLateTasks

### LINKED ISSUE #10013: transaction timeouts during testMultipleInstancesCompeteToRunManyLateTasks
When tasks in the test case testMultipleInstancesCompeteToRunManyLateTasks are slow to run, their transaction can time out, resulting in warnings in the logs, which are detected as test failures.  The persistent executor code is built to retry and the test otherwise runs cleanly.  These rollback warnings needs to be tolerated by the test instead of being flagged as errors.

njr-11:
pull fixes #10013


njr-11:
#build

njr-11:
#build

PR #10057, opened 2019-12-06
10030: On connection errors, free up the awaiting thread

### LINKED ISSUE #10030: H2 connection error causes server timeout
When an H2 connection fails to initialize, a resource is left behind which makes the server unable to stop.

loriadi:
fixes #10030


loriadi:
#build

loriadi:
Two test failures in other components, not related.

PR #10071, opened 2019-12-08
Revert "Revert "9392 batch remote partition recovery""

### LINKED ISSUE #9392: Java Batch: Remote JVM partitions not restartable after executor shutdown when a second executor was active. (cont. #8761)
**Symptom:**  This symptom is basically the same as in issue #8761 as the fix there was incomplete.   The same symptom exists here: an inability to restart a job containing "remote partitions" (partitions dispatched to a JMS queue via the `batchJmsDispatcher` configuration element) after a failure in an "executor" server running one of these remote partitions.   There remains a timing window in the case that a second executor is active and configured (via message selector, etc.) so that it is also able to consume and execute the same partition dispatch message that had been running with the first executor failed.

**Versions affected:** All versions up to 19.0.0.10

**Instructions to reproduce:** Define a batch job with a partitioned step, with the `batchJmsDispatcher` element configured to dispatch remote JVM partitions.  Configure two "executor" servers with the `batchJmsExecutor` element configured to consume these partition dispatch messages.    During execution of one of the partitions, kill the executor processing it.   It will be easier to isolate the current issue if this is not also running the "top-level" thread of the job this partition is part of.     Though the job will reach a failed state, a restart of this job instance/execution via the Batch REST API will lead to a 409 (Conflict) response status code.   The batch runtime will not allow restart of this job, because the partition is still marked with a status indicating it is currently running.

Workaround: While not usually recommended, a potential workaround could be to manually update the relevant values in the job repository database, to mark the relevant rows as failed, making them restartable.   We do not currently document how to do this (except to say contact IBM support).

### LINKED ISSUE #9913: Re-introduce change reverted from 9552
Re-introduce change reverted from https://github.com/OpenLiberty/open-liberty/pull/9552

scottkurz:
Reverts #9912 (which reverted #9552, which fixed #9392).

This also fixes #9913, opened during the earlier revert.

### DIFF
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/PartitionedStepControllerImpl.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/controller/impl/PartitionedStepControllerImpl.java
@@ -21,9 +21,11 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.Future;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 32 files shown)


scottkurz:
#run libbybot

scottkurz:
Build here https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult

Pretty good results... the failure in the jbatch.jms_fat (deadlock in Derby around table init) is a known problem we've been content to blame on Derby and ignore.    The z/OS test build didn't run though.   Looking at that.

scottkurz:
#run libbybot

scottkurz:
#run libbybot

scottkurz:
#run-liberty-bot

scottkurz:
#run-libby-bot

Issue #10070, opened 2019-12-08
Process fork failed exception is logged occasionally while running FAT tests.


toshiyamamoto:
On Windows test system, occasonally, the following error was reported.

Exception occurred: C:\...\build\dev\com.ibm.ws.security.javaeesec_fat.3\autoFVT\src\ant\launch.xml:318: Process fork failed.

PR #10072, opened 2019-12-08
reduce the length of classpath for windows.

toshiyamamoto:
In order to avoid exceeding maximum command line length of windows (32767 characters), use a jar file of which manifiest file contains class path instead of using -cp option.


toshiyamamoto:
fixes #10070 
#build

Issue #9976, opened 2019-11-29
@Inject return null in @RequestScoped of ContainerRequestFilter


kazadamid:
I deal with the issued that I can't use **@Inject** with **@RequestScope**d in a class that implements **ContainerRequestFilter**. When I remove **@RequestScoped**  and replace with **@Dependent** it can be injected into concrete class. It's only happened in **Openliberty**, I've already tried with **Wildfly** and **Payara** It can work.
My code:

```
import java.util.HashSet;
import java.util.Set;

import javax.enterprise.context.ApplicationScoped;

@ApplicationScoped
public class CodeCache{

    private Set<String> codes= new HashSet<>();

    public synchronized void add(String code) {
        this.codes.add(code);
    }

    public synchronized boolean isExisted(String code) {
        return this.codes.contains(code);
    }

    public synchronized void remove(String code) {
        this.codes.remove(code);
   }
}
```

```
import javax.annotation.Priority;
import javax.enterprise.context.RequestScoped;
import javax.inject.Inject;
import javax.ws.rs.Priorities;
import javax.ws.rs.container.ContainerRequestContext;
import javax.ws.rs.container.ContainerRequestFilter;
import javax.ws.rs.ext.Provider;

@Provider
@RequestScoped
@Priority(Priorities.USER)
public class CodeRequestFilterimplements ContainerRequestFilter {
	
	@Inject
	private CodeCache codeCache;
	
	
	@Override
    public void filter(ContainerRequestContext requestContext) {
		//handle filter here
	}
}
```

How can I get rid of this issue? Please help

kazadamid:
Does anyone know what is root cause of it?

jkoehler22:
As I worked to fix this issue I learned that the injection for RequestScope is not allowed as defined by the JAX-RS spec.  https://download.oracle.com/otndocs/jcp/jaxrs-2_0-fr-eval-spec/index.html  See:

**10.2.3 Context and Dependency Injection (CDI)**
**In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource classes, providers and Application subclasses. Providers and Application subclasses MUST be singletons or use application scope.**

The message or trace log will have a warning message indicating the error.
**W CWWKW1002W: The CDI scope of JAXRS-2.0 Provider StartupFilter1 is javax.enterprise.context.RequestScoped. Liberty gets the provider instance from JAXRS.**

[...]

PR #10081, opened 2019-12-09
Fix for Issue #9976: Allow Injection when RequestScope is used with a ContainerRequestFilter

jkoehler22:


jkoehler22:
#build

dev/com.ibm.ws.jaxrs.2.0.cdi.1.2_fat/test-applications/loadonstartup/src/com/ibm/ws/jaxrs20/cdi12/fat/loadonstartup/StartupFilter1.java
```diff
@@ -13,12 +13,13 @@
...
 import javax.annotation.PostConstruct;
-import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.RequestScoped;
 import javax.inject.Inject;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
 
-@ApplicationScoped
+// Change from ApplicationScoped to RequestScoped to verify the fix for issue 9976
+@RequestScoped
```
andymc12: Instead of changing this test, could we duplicate the test - and use one with App scope and the other with Request scope?

dev/com.ibm.ws.jaxrs.2.0.cdi/src/com/ibm/ws/jaxrs20/cdi/component/JaxRsFactoryImplicitBeanCDICustomizer.java
```diff
@@ -418,7 +418,10 @@ public void onPrepareProviderResource(BeanCustomizerContext context) {
                     //else report warning, keep using provider from rs: change to use RuntimeType.POJO
                     else {
                         p.setRuntimeType(RuntimeType.POJO);
-                        resourcesManagedbyCDI.remove(p.getProviderResourceClass());
+                        // if CDI Scope is REQUEST_SCOPE, report warning and no action: get provider from JAXRS
+                        if (!(scopeName.contains(JAXRSCDIConstants.REQUEST_SCOPE))) {
+                            resourcesManagedbyCDI.remove(p.getProviderResourceClass());
```
andymc12: I think this is not allowed by the spec - see section 11.2.3:
```
In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root re- source classes, providers and Application subclasses. Providers and Application subclasses MUST be singletons or use application scope.
```

That second sentence makes me think that we should not be allowing providers to be set to Request scope - maybe instead of enabling this, we should be logging an error message...  In any case, the logProviderMismatch call should be included in the if block, but I think the logic is not quite right.

What we may need to do is add a behavior switch that enables providers to use request scope - and then use your new code path if that switch is enabled (it should be disabled by default to keep the current behavior).

dev/com.ibm.ws.jaxrs.2.0.cdi/src/com/ibm/ws/jaxrs20/cdi/component/JaxRsFactoryImplicitBeanCDICustomizer.java
```diff
@@ -499,7 +502,9 @@ public void onPrepareProviderResource(BeanCustomizerContext context) {
             }
 
         }
-
+        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+            Tr.debug(tc, "Map of Managed Objects " + resourcesManagedbyCDI);
```
andymc12: this is nice - it should help with issues like this in the future.

PR #10082, opened 2019-12-09
errorpaths tests with failover enabled

njr-11:
This pull arranges for the error paths tests for persistent executor to run when fail over is enabled.
Variants of the tests run with vs without polling enabled.
Also, while running these tests, I noticed that the combination of failover without polling, which is currently a permitted configuration, neglects to run tasks locally, and have updated the code to fix that.  We should consider if that combination should even be allowed at allow, but for the time being, I've fixed it to behave as would be expected.


njr-11:
#build

PR #10121, opened 2019-12-11
No timestamp update by peer

### LINKED ISSUE #10120: No timestamp update on peer recovery for DB Logs
The timestamping at the heart of the recovery log locking scheme can be asymmetric. The home server (which owns the logs) will be solely responsible for updating the timestamp. A peer will recover the servers logs if (a) instructed to do so by peer recovery logic and (b) if the logs are stale. The peer will not update the timestamp. This means that, in a split brain scenario, a home server may take over ownership of its own logs while a peer is driving recovery of those logs. This is safe in an RDBMS scenario as it will not update further log records.

neilgsyoung:
Fixes #10120


neilgsyoung:
#build

neilgsyoung:
#build

neilgsyoung:
#build

neilgsyoung:
#run-libby-bot

PR #10127, opened 2019-12-11
EJB Timer Service needs a way to check if fail over is enabled

njr-11:
Add an isFailOverEnabled method to the existing TimersPersistentExecutor interface which the EJB Timer Service has access to.


njr-11:
#build

PR #10131, opened 2019-12-11
query and lock timeout on various databases

njr-11:
Implement code around how query and lock timeout are established via the JPA persistence service.  Update tests for coverage with the set of databases that we typically test with.

During this effort, it was discovered that the mechanism for detecting the Oracle JDBC isn't working, and needed to be replaced.

Observed default behavior for various databases:
Oracle and PostgreSQL - timeout is unnecessary because they return from the query based on the committed state rather than being blocked by pending updates.
SQL Server - can get lock timeout immediately when EclipseLink generates SQL like: SELECT ID FROM WLPTASK WITH (UPDLOCK, NOWAIT) WHERE ((ID = ?) AND (VERSION = ?))
DB2 - blocked until query timeout occurs
Derby - does not appear to support query timeout or a more granular lock timeout.  The global 1 minute lock timeout applies, which causes a long wait.

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
@@ -87,7 +87,7 @@ public class SchedulerFATServlet extends HttpServlet {
     /**
      * Indicates if the database for the persistent executor is Derby.
      */
-    private boolean isDerby;
+    private static final boolean isDerby = System.getProperty("fat.bucket.db.type", "Derby").compareToIgnoreCase("Derby") == 0;
 
     @Resource(lookup = "concurrent/myScheduler")
     private PersistentExecutor scheduler;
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
```diff
@@ -80,6 +84,11 @@
      */
     private static final String SUCCESS_MESSAGE = "COMPLETED SUCCESSFULLY";
 
+    /**
+     * Indicates if the database for the persistent executor is Derby.
+     */
+    private boolean isDerby;
```
KyleAure: You could instead use: 
```
private final boolean isDerby = System.getProperty("fat.bucket.db.type", "Derby").compareToIgnoreCase("Derby") == 0;
```
And avoid making a connection in the `init` method

njr-11: That's a nice improvement. I'll update it.

njr-11:
#run-libby-bot
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build has 1 test failure, which is a pre-existing issue in ui tests

PR #10140, opened 2019-12-12
Check SharedMetricRegistries for MP Metrics

andymc12:
Use `SharedMetricRegistries` to determine if MP Metrics is enabled.


andymc12:
#build

PR #10146, opened 2019-12-12
Handle forthcoming package rename of MetricRegistryFactory

andymc12:
The MP Metrics 2.X implementation is planning to change the package of the MetricRegistryFactory, so this change is planning to handle it by using both packages (just not relying on the new one until it is committed.


andymc12:
#build

PR #10154, opened 2019-12-13
tests for task that cancels or removes itself and blocks

njr-11:
Add tests where a task execution cancels or removes itself and then blocks.
Ensure that other tasks can still be scheduled, run, and themselves canceled or removed.


njr-11:
#build

njr-11:
#build

PR #10162, opened 2019-12-13
choose fail over approach with time-based claim

njr-11:
This pull removes the fail over approach that is based on assignment of tasks to different members and chooses instead the fail over approach that uses time-based claims, updating all associated test cases.


njr-11:
#build

njr-11:
#build

PR #10163, opened 2019-12-14
use PBKDF2WithHmacSHA256 for password hash

### LINKED ISSUE #10141: Stronger hashing algorithm needs to be used for the authentication cache
The current authentication cache is using SHA-1 algorithm for hashing the password. Currently, SHA-1 is no longer recognized as a adequate hashing algorithm. Although the hashed data only exists on memory and the data won't be written in a persistent storage or over the wire, but It is preferable to use stronger hashing algorithm for safe.

toshiyamamoto:
fixes #10141


toshiyamamoto:
#build

toshiyamamoto:
#build

toshiyamamoto:
#build

PR #10173, opened 2019-12-16
Update Mongo FAT for IBMCertPathBuilderException FFDC

ayoho:
It appears that the IBM JDK now emits an `IBMCertPathBuilderException` when an untrusted signer certificate is recieved from a remote host. This change updates a FAT test for this scenario to allow the new FFDC.


ayoho:
#build

PR #10194, opened 2019-12-17
Introduce startup script for sql server

KyleAure:
SQLServer needs to have the database created at startup and a stored procedure to run to enable XA connections.  Previously this was being done after the server by making a connection and running sql commands.  Instead, we can provide testcontainers with a .sql file and it will be run just after container creation.  This eliminates the need to create extra connections and should speed up our startup time.


KyleAure:
#build 
#libbybot

Issue #10178, opened 2019-12-16
Resource leak when installing features through Gradle on Windows


ericglau:
**Describe the bug**
After installing features through Gradle, the Liberty directory cannot be deleted.  It looks like this is caused by InstallMap.java using a URLClassLoader with the Liberty jars, and the URLClassLoader is never closed.  See this part of the code https://github.com/OpenLiberty/open-liberty/blob/b06d50f7a356766796ccdd08be19cee940b5a626/dev/com.ibm.ws.install.map/src/com/ibm/ws/install/map/InstallMap.java#L543

**Steps to Reproduce**
1. Clone the `gradle` branch of https://github.com/OpenLiberty/demo-devmode
2. Run `gradle installFeature`
3. Run `gradle clean`

The following error occurs:
```
Execution failed for task ':clean'.
> java.io.IOException: Unable to delete directory 'C:\Users\Eric\git\demo-devmode\build'
    Failed to delete some children. This might happen because a process has files open or has its working directory set in the target directory.
    - C:\Users\Eric\git\demo-devmode\build\wlp\dev\api\spec\com.ibm.websphere.javaee.jsonp.1.0_1.0.35.jar
    - C:\Users\Eric\git\demo-devmode\build\wlp\dev\api\spec
    - C:\Users\Eric\git\demo-devmode\build\wlp\dev\api
    - C:\Users\Eric\git\demo-devmode\build\wlp\dev
    - C:\Users\Eric\git\demo-devmode\build\wlp\lib\com.ibm.ws.config_1.0.35.jar
    - C:\Users\Eric\git\demo-devmode\build\wlp\lib\com.ibm.ws.crypto.passwordutil_1.0.35.jar
    - C:\Users\Eric\git\demo-devmode\build\wlp\lib\com.ibm.ws.install.map_1.0.35.jar
    - C:\Users\Eric\git\demo-devmode\build\wlp\lib\com.ibm.ws.install_1.0.35.jar
    - C:\Users\Eric\git\demo-devmode\build\wlp\lib\com.ibm.ws.kernel.boot_1.0.35.jar
    - C:\Users\Eric\git\demo-devmode\build\wlp\lib\com.ibm.ws.kernel.feature_1.0.35.jar
    - C:\Users\Eric\git\demo-devmode\build\wlp\lib\com.ibm.ws.kernel.service_1.3.35.jar
[...]

PR #10200, opened 2019-12-17
Close URLClassLoader to fix resource leak

ericglau:
Fixes https://github.com/OpenLiberty/open-liberty/issues/10178

When installing features, InstallMap.java uses a URLClassLoader with all of the Liberty jar files.  However, this class loader is never closed.  Since Gradle uses a long running daemon and doesn't exit the JVM after the task is finished, installing features through Gradle will cause the resources to be held.  This causes both a memory leak and prevents the Liberty directory from being deleted.

The fix is to close the URLClassLoader in the map's `clear()` method, which must be called by the Gradle task.  This method call is needed because otherwise the caller just uses `put()` and `get()` methods, with clear no indication on when the resources can be closed.

### DIFF
--- a/dev/com.ibm.ws.install.map/src/com/ibm/ws/install/map/InstallMap.java
+++ b/dev/com.ibm.ws.install.map/src/com/ibm/ws/install/map/InstallMap.java
@@ -216,6 +216,8 @@ public class InstallMap implements Map {
      * Clears the map and closes all resources.
      *
      * <b>MUST</b> be used when the map is no longer needed.
+     * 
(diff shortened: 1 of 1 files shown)


ericglau:
#build

ericglau:
Build test failures are not related to install map.

Manually tested with the openliberty-kernel on Windows using the personal build, modified version to 19.0.0.12 so that it can get features from Maven Central, then ran `gradle installFeature` then `gradle clean`.

ericglau:
#run-libby-bot

PR #10206, opened 2019-12-17
restrict mbean test to public API to determine if the mbean is usable in Liberty

njr-11:
I was hoping to remove the undocumented PersistentExecutorMBean from Liberty given the different direction that we are taking for fail over.  However, it appears that existing main path code mistakenly registers PersistentExecutorMBean without restricting it to non-ship/beta code path.  To prove that it is truly accessible/usable, I updated the com.ibm.ws.concurrent.persistent_fat_mbean test bucket such that it only uses GA features rather than non-ship features/code, and that confirms it.  For now, I have also added a test case to enforce that the PersistentExecutorMBean interface never gets exposed as API.  As to whether or not the mbean can be completely removed, there will need to be further discussion.  At a minimum, we should aim to reject it when fail over is enabled (this can be done under a separate pull).


njr-11:
#build

njr-11:
#build

PR #10220, opened 2019-12-18
Moving loginmodule to separate package

KyleAure:
Login module code was previously in the jdbc.fat.tests package.  As a result we were packaging the entire fat/src directory and making a copy of it so that the app could access the login module code.

Here, I have moved this code it it's own package, and export it to a jar at runtime.


KyleAure:
#build 
#libbytool

PR #10225, opened 2019-12-18
Batch: Prevent noise on job dispatch message redelivery

### LINKED ISSUE #10337: Java Batch: Error reported when JMS job dispatch message is redelivered
**Describe the bug**
A JMS message for a Batch job dispatch may in some cases be redelivered to an endpoint. The Batch runtime correctly recognizes that the message has already been seen and does not start the job again, but in the process it produces an error message and FFDC. This is unnecessary noise given that there is no real impact and no action to be taken to further handle the redelivered message.

**Expected behavior**
Since there is no action needed to resolve the message redelivery, it should be processed silently.

**Versions affected**
All Liberty fix packs through 19.0.0.12.

cgianfra:
fixes #10337

### DIFF
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014 IBM Corporation and others.
+ * Copyright (c) 2014, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
@@ -269,12 +269,10 @@ public interface WSJobRepository {
     public abstract WSRemotablePartitionExecution createRemotablePartition(RemotablePartitionKey remotablePartitionKey);
 
     /**
-     * Updates an entry for this remote partition in the RemotablePartition table with the given internalStatus
-     * If the remotable partition is not found, then it just returns
+     * Gets the internal status of a remotable partition, does nothing if RemotablePartition table does not exist
      *
      * @param remotablePartitionKey
      * @return
-     * @throws IllegalArgumentException if partition isn't found via key
      */
     public abstract WSRemotablePartitionState getRemotablePartitionInternalState(RemotablePartitionKey remotablePartitionKey);
 
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
@@ -169,8 +169,9 @@ public interface WSJobRepository {
      * Update the instanceState to JMS_CONSUMED if it's a valid status transition
      *
      * @param instanceId
+     * @throws JobInstanceNotQueuedException
      */
-    public abstract WSJobInstance updateJobInstanceStateOnConsumed(long instanceId) throws BatchIllegalJobStatusTransitionException;
+    public abstract WSJobInstance updateJobInstanceStateOnConsumed(long instanceId) throws BatchIllegalJobStatusTransitionException, JobInstanceNotQueuedException;
 
     /**
      * Update the instanceState to JMS_QUEUED if it's a valid status transition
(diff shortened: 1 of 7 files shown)


dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
```diff
@@ -268,6 +269,9 @@ public abstract WSStepThreadExecutionAggregate getStepExecutionAggregateFromJobE
     public abstract WSRemotablePartitionExecution createRemotablePartition(RemotablePartitionKey remotablePartitionKey);
 
     /**
+     * Updates an entry for this remote partition in the RemotablePartition table with the given internalStatus
+     * If the remotable partition is not found, then it just returns
+     *
      * @param remotablePartitionKey
      * @return
      * @throws IllegalArgumentException if partition isn't found via key
```
scottkurz: So if the above comment is right (I think it is), the `@throws` clause should be removed.

dev/com.ibm.ws.jbatch.jms/src/com/ibm/ws/jbatch/jms/internal/listener/BatchJmsEndpointListener.java
```diff
@@ -569,10 +577,15 @@ private void handleStartRequest(long instanceId, long executionId, byte[] securi
...
-            if (jobInstance.getInstanceState() != InstanceState.JMS_CONSUMED) {
+            WSJobInstance jobInstance = null; 
+
+            try {
+            	jobInstance = jobRepositoryProxy.updateJobInstanceStateOnConsumed(instanceId);
+            } catch (JobInstanceNotQueuedException exc) {
+            	// trace and swallow this 
                 if(tc.isDebugEnabled()) {
-                    Tr.debug(BatchJmsEndpointListener.this, tc, "Exiting since instanceState isn't equal to JMS_CONSUMED, instead is: " + jobInstance.getInstanceState());
+                	// Might be misleading to trace the non-JMS_CONSUMED `state we found without locking it down
```
scottkurz: It'd be better to change "JMS_CONSUMED" here to "JMS_QUEUED", even though it's just a comment (it could help someone later to be more clear).

When we wrote this before, we were looking for a _return value_ with state of **JMS_CONSUMED**.   Now that we more simply throw an exception to indicate the wrong (non-JMS_QUEUED) state, it'd be clearer to mention the issue is that we didn't find this in the state of **JMS_QUEUED**

Now that we've simplified the

dev/com.ibm.ws.jbatch.jms/src/com/ibm/ws/jbatch/jms/internal/listener/BatchJmsEndpointListener.java
```diff
@@ -569,10 +577,15 @@ private void handleStartRequest(long instanceId, long executionId, byte[] securi
...
+                    Tr.debug(BatchJmsEndpointListener.this, tc, "Exiting since instanceState isn't equal to JMS_CONSUMED");
                 }
```
scottkurz: Similar comment for the trace string here.

dev/com.ibm.ws.jbatch.jms/src/com/ibm/ws/jbatch/jms/internal/listener/BatchJmsEndpointListener.java
```diff
@@ -490,13 +492,19 @@ private void handleRestartRequest(long jobInstanceId, long executionId, byte[] s
...
+            WSJobInstance jobInstance = null; 
+
+            try {
+            	jobInstance = jobRepositoryProxy.updateJobInstanceStateOnConsumed(instanceId);
+            } catch (JobInstanceNotQueuedException exc) {
+            	// trace and swallow this 
                 if(tc.isDebugEnabled()) {
-                    Tr.debug(BatchJmsEndpointListener.this, tc, "Exiting since instanceState isn't equal to JMS_CONSUMED, instead is: " + jobInstance.getInstanceState());
+                	// Might be misleading to trace the non-JMS_CONSUMED `state we found without locking it down
+                    Tr.debug(BatchJmsEndpointListener.this, tc, "Exiting since instanceState isn't equal to JMS_CONSUMED");
```
scottkurz: See comments below for comment and trace string.

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
cgianfra: Looks like there's a potential IllegalArgumentException on the EntityManager find method, if the entity class supplied isn't actually an entity, or if the key to look up is null or not applicable.. not sure if that realistically will ever trigger for us, but I guess it's technically possible? At the very least, the wording here is inaccurate, since not finding a match won't throw.

cgianfra:
fixes #10337

dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/ws/WSJobRepository.java
scottkurz: Yeah..better now.. we don't always document every very-unexpected exc but should document somewhat-expected ones accurately.

cgianfra:
#build

cgianfra:
#run-libby-bot

PR #10227, opened 2019-12-18
Handles "unassociated" types/enums

andymc12:
These types/enums are indirectly associated with top level queries or mutations via an interface of input type that is directly associated with a top level query/mutation.

Implements feature https://github.com/eclipse/microprofile-graphql/issues/125


andymc12:
#build

PR #10255, opened 2019-12-20
update to MP Context Propagation 1.0.2 for TCK dependency on Jakarta EE 8

njr-11:
MicroProfile Context Propagation is releasing a 1.0.2 micro version so that its TCK dependencies can be updated from Java EE to Jakarta EE.  This pull is to update OpenLiberty to start running against the 1.0.2 TCK, which ought to be identical to the 1.0.1 TCK, but this way, OpenLiberty will be running against the very latest version in automated test.


njr-11:
#build

PR #10264, opened 2019-12-30
tests that enable failover for existing instances

njr-11:
Write tests that start out with fail over disabled and schedule some tasks, then enable fail over on those same instances and verify that both the original tasks, as well as any new tasks, run successfully.


njr-11:
#build

PR #10270, opened 2019-12-31
Fix test suite reporting when using TestContainers

KyleAure:
Test suites are failing to report errors correctly when a docker host cannot be found on z/OS.
This seems to only happen in test suites that create a test container in the FATSuite.java file.


KyleAure:
#build 
#libbybot

KyleAure:
#build 
#libbybot

KyleAure:
#libbybot

PR #10273, opened 2020-01-02
tests that replace single-member instance with failover-enabled instance

njr-11:
Write tests for scenarios that involve replacing an instance that is not enabled for fail over with one that is.  This covers scenarios where the server is stopped as well as left running.  It also covers scenarios where the two instances overlap and where they do not.


njr-11:
#build

PR #10275, opened 2020-01-02
tests for disabling fail over

njr-11:
Write some tests to experiment with code paths where fail over is enabled and then disabled. We don't recommend doing that because it will leave existing tasks orphaned. The tests can experiment with ways of transferring orphaned tasks back to an available instance/member by directly accessing the database. However, there will not be any further investment in that area because it is something that users should not be doing.  If someone does come up with a need for it at some point, these tests will at least verify it is possible to transfer the tasks back and get them executing again.

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent_fat_failover1serv/test-applications/failover1servApp/src/failover1serv/web/Failover1ServerTestServlet.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat_failover1serv/test-applications/failover1servApp/src/failover1serv/web/Failover1ServerTestServlet.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2019,2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


njr-11:
#build

dev/com.ibm.ws.concurrent.persistent_fat_failover1serv/test-applications/failover1servApp/src/failover1serv/web/Failover1ServerTestServlet.java
```diff
@@ -247,4 +252,82 @@ public void testTaskStarted(HttpServletRequest request, HttpServletResponse resp
...
+        String[] taskIds = request.getParameterValues("taskId");
+
+        PersistentExecutor executor = InitialContext.doLookup(jndiName);
+        DataSource ds = InitialContext.doLookup("java:comp/DefaultDataSource");
+
+        for (String taskIdString : taskIds) {
+            long taskId = Long.valueOf(taskIdString);
+            // The only way to find the value stored in a task's PARTN column is to query the database
+            try (Connection con = ds.getConnection()) {
+                // querying only EXECUTOR and ignoring HOSTANME, USERDIR, LSERVER columns because there is only one instance
```
nmittles: HOSTANME typo.

Issue #10228, opened 2019-12-19
Rest Client for MicroProfile loses entity on POST requests with status code 202 response


SPSpunkt:
**Describe the bug**
If using Rest Client for MicroProfile (Apache CXF impl in case of OLP/WLP!) together with a remote service that responds a status code 202 and an entity, the response filter (chain) is called twice but the entity is null for the final call. (InputStream#read == -1)

* Logging filters may fail with NPT like issues or  
* even simple unmarshalling fails for the missing entity

Sample stack traces (please take care, it is very use case dependent)
```
[ERROR   ] SRVE0315E: Es ist eine Ausnahme eingetreten: java.lang.Throwable: javax.ws.rs.client.ResponseProcessingException: Problem with reading the data, class java.sql.Date, ContentType: application/json.
	at com.ibm.ws.webcontainer.webapp.WebApp.handleRequest(WebApp.java:5075)
	at [internal classes]
Caused by: javax.ws.rs.client.ResponseProcessingException: Problem with reading the data, class java.sql.Date, ContentType: application/json.
	at org.apache.cxf.jaxrs.impl.ResponseImpl.reportMessageHandlerProblem(ResponseImpl.java:563)
	at [internal classes]
	at com.sun.proxy.$Proxy86.dTOJson(Unknown Source)
	at servicea.Resources.dtoJson(Resources.java:36)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at com.ibm.ws.jaxrs20.cdi.component.JaxRsFactoryImplicitBeanCDICustomizer.serviceInvoke(JaxRsFactoryImplicitBeanCDICustomizer.java:304)
	... 1 more
Caused by: javax.json.bind.JsonbException: Internal error: null
	at org.eclipse.yasson.internal.Unmarshaller.deserializeItem(Unmarshaller.java:74)
	... 9 more
Caused by: java.util.NoSuchElementException
	at org.glassfish.json.JsonParserImpl.next(JsonParserImpl.java:361)
	... 9 more
```
**Steps to Reproduce**
[...]

PR #10276, opened 2020-01-02
Avoid closing entityStream in JsonBProvider

andymc12:
This is required by the spec (see MBW's javadoc) that we must not close the entity stream in the writeTo method. The toJson(obj, stream) method will always close the stream, so this cannot be used.

This commit includes a test case that verifies that the client response filter is only called once per response.  The prior behavior invoked the filter twice - once as a result of the toJson(obj, stream) method closing the stream and again when the response was actually processed.

This resolves issue #10228, and should be considered a release bug.


andymc12:
#build

PR #10278, opened 2020-01-02
New test case for MP Rest Client custom headers factory

andymc12:
Tests that MP Rest Client headers factories are used when specified in the `@RegisterClientHeaders` annotation on the client interface.


andymc12:
#build

PR #10297, opened 2020-01-06
Issue #10296: Validate server start even if script timeout occurs

### LINKED ISSUE #10296: Validate server start even if script timeout occurs
In the test framework, if the server script times out but the server's status shows that it started, we currently don't run the normal wait/validation to make sure that test applications and the server have fully started.

This obviously doesn't happen very often, but when it does entire buckets can fail as they plow ahead trying to run tests against a server that hasn't initialized yet.

brideck:
- Updated to not return right away, but continue on to validation step

fixes #10296


brideck:
fixes #10296

brideck:
#build

brideck:
Tested updated code path locally by forcing output to null.

brideck:
#build

brideck:
The only failure in the latest personal build was an unrelated UI issue.  Merging.

PR #10301, opened 2020-01-06
enforce limits on configured duration values

njr-11:
Enforce upper/lower limits on various configured duration values (missedTaskThreshold, pollInterval, retryInterval) when fail over is enabled to help ensure the user is running with reasonable values.  Translatable message will be added in a separate pull, once this part is reviewed and working.


njr-11:
An id review is not needed because that was done under 10284.  This pull has a couple of the same commits because 10284 hasn't been able to deliver yet.

njr-11:
#build

njr-11:
#build
#run-libby-bot

skasund:

PR #10336, opened 2020-01-08
10321 jax cdi managedobject

benjamin-confino:


dev/com.ibm.ws.jaxrs.2.0.cdi/src/com/ibm/ws/jaxrs20/cdi/component/JaxRsFactoryImplicitBeanCDICustomizer.java
```diff
@@ -273,6 +273,34 @@ private Object getClassFromCDI(Class<?> clazz) {
...
+        }
+
+        ManagedObjectFactory<Object> managedObjectFactory = (ManagedObjectFactory<Object>) getManagedObjectFactory(clazz);
+
+        ManagedObject<Object> bean = null;
+        try {
+            bean = managedObjectFactory.createManagedObject(serviceObject, null);
+        } catch (Exception e) {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "Couldn't create object instance from ManagedObjectFactory for : " + clazz.getName() + ", but ignore the FFDC: ", e);
```
jim-krueger: I'm thinking you might need warning or error messages here (and above) rather than just Tr.debugs

dev/com.ibm.ws.jaxrs.2.0.cdi/src/com/ibm/ws/jaxrs20/cdi/component/JaxRsFactoryImplicitBeanCDICustomizer.java
```diff
@@ -273,6 +273,34 @@ private Object getClassFromCDI(Class<?> clazz) {
...
     }
 
+    /**
+     * @param clazz
+     * @return
+     */
+    @FFDCIgnore(value = { Exception.class })
+    private ManagedObject<?> getClassFromServiceObject(Class<?> clazz, Object serviceObject) {
+
+        if (! clazz.equals(serviceObject.getClass())) {
```
andymc12: This block will never be executed, right?  From line 239: `Class<?> clazz = serviceObject.getClass();`

So, `clazz` will will also be equal to `serviceObject.getClass()`. 

Is the intent to determine if the serviceObject is already a Weld proxy?  If so, it might be better to just check for that directly - like maybe make a public `isWeldProxy(Object o)` method in CDI code that we could call from here, and you could call from the `CDIManagedObjectFactoryImpl` above?

dev/com.ibm.ws.jaxrs.2.0.cdi/src/com/ibm/ws/jaxrs20/cdi/component/JaxRsFactoryImplicitBeanCDICustomizer.java
andymc12: I'm fine with leaving this as a debug - it is the same as what we currently have in `getClassFromManagedObject`.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIManagedObjectFactoryImpl.java
```diff
@@ -116,4 +123,10 @@ public ManagedObjectContext createContext(boolean nonContextual) throws ManagedO
     public String toString() {
         return "CDI Managed Object Factory for class: " + getManagedObjectClass().getName();
     }
+
+    private void isWeldProxy() {
+        if (getManagedObjectClass().getSimpleName().contains("Proxy$_$$_WeldSubclass")) {
```
andymc12: does Weld only use `"Proxy$_$$_WeldSubclass"` as the extension?  In at least one test case that we've run across, we saw:  `com.ibm.ws.jaxrs20.cdi12.fat.loadonstartup.StartupFilter1$Proxy$_$$_WeldClientProxy`.

Maybe this should be:
```suggestion
        if (getManagedObjectClass().getSimpleName().contains("Proxy$_$$_Weld")) {
```

benjamin-confino: Good catch.

dev/com.ibm.ws.jaxrs.2.0.cdi/src/com/ibm/ws/jaxrs20/cdi/component/JaxRsFactoryImplicitBeanCDICustomizer.java
benjamin-confino: This is just a better safe than sorry thing. This method requires serviceObject to be instance of clazz. So why not explicitly check for it? 

The practical difference is that if this is here we'll return null if someone else calls with a mismatched clazz and serviceObject instead of a java error when AbstractManagedObjectFactory<T>.createManagedObject(T instance... is called with something that's not T.

benjamin-confino:
#build

benjamin-confino:
Build was good, only one bucket failed and it doesn't *look* related. Lets try again to be safe.

benjamin-confino:
#build

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIManagedObjectFactoryImpl.java
```diff
@@ -57,9 +57,20 @@ public CDIManagedObjectFactoryImpl(Class<T> classToManage, CDIRuntime cdiRuntime
 
     @Override
     public ManagedObject<T> createManagedObject() throws ManagedObjectException {
+        isWeldProxy();
```
tevans78: Why check the class here and not at the point where the MOF is constructed (where the class is passed in).

tevans78: And I am pretty sure this would apply across all flavours of the MOF so the check can be done in the constructor of AbstractManagedObjectFactory

PR #10338, opened 2020-01-08
Adding programmatic timer

KyleAure:
Enable application to create a programmatic timer.  This timer is being created by an automatic timer that runs once an hour.  The programmatic timer that is created runs every 5 minutes for a half an hour. 

This is to simulate the customer case where their app responds to some user action (button click, or http request) which starts a timer action.

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/fat/src/com/ibm/ws/concurrent/persistent/fat/demo/timers/DemoTimerTest.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/fat/src/com/ibm/ws/concurrent/persistent/fat/demo/timers/DemoTimerTest.java
@@ -56,6 +56,15 @@ public class DemoTimerTest extends FATServletClient {
         //Get driver name
         server.addEnvVar("DB_DRIVER", DatabaseContainerType.valueOf(testContainer).getDriverName());
 
+        if (DatabaseContainerType.valueOf(testContainer) == DatabaseContainerType.Oracle) {
+            //Type=javax.sql.ConnectionPoolDataSource to avoid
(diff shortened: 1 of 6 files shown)


dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/fat/src/com/ibm/ws/concurrent/persistent/fat/demo/timers/DemoTimerTest.java
```diff
@@ -56,6 +56,15 @@ public static void setUp() throws Exception {
         //Get driver name
         server.addEnvVar("DB_DRIVER", DatabaseContainerType.valueOf(testContainer).getDriverName());
 
+        if (DatabaseContainerType.valueOf(testContainer) == DatabaseContainerType.Oracle) {
+            //Type=javax.sql.ConnectionPoolDataSource to avoid
+            //  ORA-02089: COMMIT is not allowed in a subordinate session
+            //  when application tries to create a new table using oracle.
+            server.addEnvVar("DS_TYPE", "javax.sql.ConnectionPoolDataSource");
```
njr-11: The fact that it is possible to run the test with a one-phase only data source tells me that we need to add/update a scenario to cover multiple transactional resources.  It's fine to deliver this for now, be we need a follow-on work item to cover that.  It should be an easy update, just tweak the `@Resource` for the DataSource in AutomaticDatabase timer.  But then the create table code will need to be moved so that XADataSource can be used.

KyleAure: Yes, I will come back to this in another PR.

KyleAure:
#build 
#libbybot

PR #10340, opened 2020-01-08
renames to align with dual use of partition column

njr-11:
Refactoring/renames of fields/methods/constants within the code to be consistent with the repurposed partition column, which is used to store the expiry timestamps for task executions when fail over is enabled.


njr-11:
#build

PR #10348, opened 2020-01-09
testUnmanagedThreadSubmitsTask waits for what never happens

### LINKED ISSUE #10347: 4 minute delay in tests from testUnmanagedThreadSubmitsTask
SchedulerFATServlet.testUnmanagedThreadSubmitsTask, which runs twice (under different configurations) in the com.ibm.ws.concurrent.persistent_fat bucket uses a utility method, pollForResult, to wait for a task execution (3) that is never scheduled to occur.  This was a typo or mistake in the test - it should have been waiting for 1, because the test only schedules a single execution to happen.  This would have been noticed right away if the utility method raised an error after timing out, but it wasn't written to have that expectation.  So the invalidly written test passes, at the price of 2 minutes useless delays, which happen twice for the bucket, doubling the time it takes the bucket to run.  I looked around at other uses of the pollForResult utility method, and the handful that are present have some varying expectations of what it actually does.  Not all of them are using it correctly based on what it actually does.  This is a good example of why not to use utility methods like this in test cases.  It is better for the test logic to flow clearly from end to end, which aids debugging as well as correctly developing the test logic in the first place.  I'm going to remove this method and put the correct logic into each of the handful of places that were previously delegating to it.

njr-11:
pull fixes #10347

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
@@ -2070,6 +2070,9 @@ public class SchedulerFATServlet extends HttpServlet {
                 && !(statusC.hasResult() && Integer.valueOf(1).equals(statusC.getResult())); )
             Thread.sleep(POLL_INTERVAL);
 
+        if (statusC == null)
+            throw new Exception("Task C should not have been auto-purged.");
+
         if (!statusC.hasResult() || !Integer.valueOf(1).equals(statusC.getResult()))
             throw new Exception("Task C did not complete " + statusC);
 
@@ -2077,6 +2080,9 @@ public class SchedulerFATServlet extends HttpServlet {
                 && (statusD = scheduler.getStatus(statusD.getTaskId())) != null
                 && !(statusD.hasResult() && Integer.valueOf(1).equals(statusD.getResult())); )
             Thread.sleep(POLL_INTERVAL);
+
+        if (statusD == null)
+            throw new Exception("Task D should not have been auto-purged.");
 
         if (!statusD.hasResult() || !Integer.valueOf(1).equals(statusD.getResult()))
             throw new Exception("Task D did not complete " + statusD);
--- a/dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2019 IBM Corporation and others.
+ * Copyright (c) 2014, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -231,27 +231,6 @@ public class SchedulerFATServlet extends HttpServlet {
         } catch (SQLException x) {
             throw new ServletException(x);
         }
-    }
-
-    /**
-     * Utility method that waits for a task status with the specified result.
-     *
-     * @return task status if found within the allotted interval, otherwise null.
-     */
-    private <T> TaskStatus<T> pollForResult(long taskId, T result) throws Exception {
-        TaskStatus<T> status = null;
-        for (long start = System.nanoTime(); System.nanoTime() - start < TIMEOUT_NS; Thread.sleep(POLL_INTERVAL)) {
-            TaskStatus<T> newStatus = scheduler.getStatus(taskId);
-            status = newStatus;
-            try {
-                if (status == null || status.hasResult() && result.equals(status.getResult()))
-                    return status;
-            } catch (SkippedException x) {
-                // allow for skips
-            }
-        } ;
-
-        return status;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


njr-11:
#build

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
```diff
@@ -2086,8 +2065,21 @@ public void testFindByName(PrintWriter out) throws Exception {
             throw new Exception("Should not be able to see tasks from outside of application. " + resultsByTaskId);
 
         // Wait for tasks C and D to complete their only execution.
-        pollForResult(statusC.getTaskId(), 1);
-        pollForResult(statusD.getTaskId(), 1);
+        for (long start = System.nanoTime(); System.nanoTime() - start < TIMEOUT_NS
```
gjwatts: Instead of adding this logic inline, why not correct the logic in pollForResult?

njr-11: Because pollForResult was interpreted to mean different things to different tests that used it.  Its behavior was only incorrect if you are looking at one of the tests that didn't want it to behave the way it did, otherwise it was doing exactly the right thing, and altering it would break that.  It's a good example of why this sort of utility method is a bad practice in tests.  Developers writing a test think they know what it does based on its name and start to unknowingly misuse it.  The solution here is to remove it and let each test perform precisely the steps that it needs inline, which will be ideal for future debugging as well as for ensuring correctness.

gjwatts: I understand that now, what you are doing makes sense to me.  Thank you for taking the time to explain that to me!

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
```diff
@@ -3830,9 +3853,13 @@ public Integer call() throws Exception {
                     if (status.hasResult())
                         throw new Exception("Task status initial snapshot should not have a result. " + status);
 
-                    status = pollForResult(status.getTaskId(), 3);
+                    long taskId = status.getTaskId();
+                    for (long start = System.nanoTime(); System.nanoTime() - start < TIMEOUT_NS
+                            && (status = scheduler.getStatus(taskId)) != null
+                            && !(status.hasResult() && Integer.valueOf(1).equals(status.getResult())); )
```
gjwatts: Should Integer.valueOf(1) here be Integer.valueOf(3) instead?

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
```diff
@@ -2086,8 +2065,21 @@ public void testFindByName(PrintWriter out) throws Exception {
...
 
         // Wait for tasks C and D to complete their only execution.
-        pollForResult(statusC.getTaskId(), 1);
-        pollForResult(statusD.getTaskId(), 1);
+        for (long start = System.nanoTime(); System.nanoTime() - start < TIMEOUT_NS
+                && (statusC = scheduler.getStatus(statusC.getTaskId())) != null
+                && !(statusC.hasResult() && Integer.valueOf(1).equals(statusC.getResult())); )
+            Thread.sleep(POLL_INTERVAL);
+
+        if (!statusC.hasResult() || !Integer.valueOf(1).equals(statusC.getResult()))
```
gjwatts: Should we first check if statusC == null before we check statusC.hasResults()?

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
```diff
@@ -2086,8 +2065,21 @@ public void testFindByName(PrintWriter out) throws Exception {
...
+            throw new Exception("Task C did not complete " + statusC);
+
+        for (long start = System.nanoTime(); System.nanoTime() - start < TIMEOUT_NS
+                && (statusD = scheduler.getStatus(statusD.getTaskId())) != null
+                && !(statusD.hasResult() && Integer.valueOf(1).equals(statusD.getResult())); )
+            Thread.sleep(POLL_INTERVAL);
+
+        if (!statusD.hasResult() || !Integer.valueOf(1).equals(statusD.getResult()))
```
gjwatts: Should we first check if statusD == null before we check statusD.hasResults()?

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
```diff
@@ -3779,13 +3794,21 @@ public void testTwoTasksInOneTransaction(PrintWriter out) throws Exception {
...
+
+        for (long start = System.nanoTime(); System.nanoTime() - start < TIMEOUT_NS
+                && (statusB = scheduler.getStatus(statusB.getTaskId())) != null
+                && !(statusB.hasResult() && Integer.valueOf(1).equals(statusB.getResult())); )
+            Thread.sleep(POLL_INTERVAL);
 
-        statusB = pollForResult(statusB.getTaskId(), 1);
-        if (statusB == null || !statusB.isDone())
-            throw new Exception("TaskB not completed in allotted interval. Status: " + statusB);
+        if (statusB == null || !statusB.isDone() || !Integer.valueOf(1).equals(statusB.getResult()))
```
gjwatts: Is statusB.isDone() the same thing as statusB.hasResult()?

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
njr-11: The task is only scheduled to run once, which returns a value of 1.  Waiting for 3, which was in the previous code, was erroneous, and I'm fixing it here.

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
njr-11: It isn't necessary because an erroneously null statusC will cause a NullPointerException and fail the test on this line, but I can add it if you prefer.

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
njr-11: Similar to above, it isn't necessary because an erroneously null statusD will cause a NullPointerException and fail the test on this line, but I can add it if you prefer.

dev/com.ibm.ws.concurrent.persistent_fat/test-applications/schedtest/src/web/SchedulerFATServlet.java
njr-11: They are similar, but not the same.  For repeating tasks (which doesn't apply here) a task can have multiple results - one for each execution.  The hasResult method tells you if at least one execution of the task has happened, such that you can get a current result.  The isDone method tells you that all executions have completed.  For one-time tasks, such as is the case here, they are both telling you the same thing.

njr-11:
#build

njr-11:
#run-libby-bot

PR #10354, opened 2020-01-09
Enable all databases to use xa datasources

KyleAure:
Was previously changing datasource to not use an XA datasource when testing against oracle. This was due to a limitation in oracle JDBC driver to create a table when using an XA datasource.  

But, the DefaultDatasource is also used to persist timers, and we want all databases to use the same type of datasource.  Now, when testing against oracle I create the table prior to running the application.  The default datasource once again uses two phase commit for all databases.

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/test-application/demotimers/src/ejb/timers/AutomaticDatabase.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/test-application/demotimers/src/ejb/timers/AutomaticDatabase.java
@@ -70,15 +70,16 @@ public class AutomaticDatabase {
     }
 
     private void initTable() {
-        final String createTable = "CREATE TABLE AutomaticDatabase (name VARCHAR(64) NOT NULL PRIMARY KEY, count INT)";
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/test-application/demotimers/src/ejb/timers/AutomaticDatabase.java
```diff
@@ -70,15 +70,16 @@ public void run(Timer timer) {
...
 
     private void initTable() {
-        final String createTable = "CREATE TABLE AutomaticDatabase (name VARCHAR(64) NOT NULL PRIMARY KEY, count INT)";
+        final String createTable = "CREATE TABLE AUTOMATICDATABASE (name VARCHAR(64) NOT NULL PRIMARY KEY, count INT)";
 
         try (Connection conn = ds.getConnection()) {
             //See if table was created by another server
             DatabaseMetaData md = conn.getMetaData();
-            ResultSet rs = md.getTables(null, null, "AutomaticDatabase", null);
+            ResultSet rs = md.getTables(null, null, "AUTOMATICDATABASE".toUpperCase(), null);
```
njr-11: the toUpperCase here is redundant

KyleAure:
#build 
#libbybot

PR #10367, opened 2020-01-09
Close the original input stream in `copyAndGetEntityStream`

andymc12:
Suggestion proposed by the CXF community to close the copied input stream.


andymc12:
#build

PR #10403, opened 2020-01-13
test introspector when failover enabled

njr-11:
Test the persistent executor introspector when fail over is enabled.  Also, fix the introspector so that invalidly configured instances don't cause it to fail and abort.  When an instance cannot be obtained, instead log information from the server reference.


njr-11:
#build

njr-11:
Build has 1 test failure, which is a pre-existing issue.

PR #10404, opened 2020-01-13
Enabling all ConfigTests

KyleAure:
This PR fixes some bugs we had in our tests as a result of bringing the JDBC fat tests from CL to OL. 

1. JDBC drivers in ${shared.resource.dir} instead of ${server.config.dir}
2. loginTimeout prop not supported on java.sql.Driver
3. Need to set server env variables after each server shutdown. 

Also made some general performance / syntax improvements.

### DIFF
--- a/dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
+++ b/dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
@@ -10,22 +10,14 @@
  *******************************************************************************/
 package com.ibm.ws.jdbc.fat.tests;
 
-import static componenttest.annotation.SkipIfSysProp.DB_Oracle;
-import static componenttest.annotation.SkipIfSysProp.DB_Postgres;
-import static componenttest.annotation.SkipIfSysProp.DB_SQLServer;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 
-import java.io.BufferedReader;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.net.HttpURLConnection;
-import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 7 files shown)


KyleAure:
#build

dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/src/basicfat/DataSourceTestServlet.java
```diff
@@ -634,7 +634,7 @@ public void testConfigChangeLoginTimeout320() throws Throwable {
 
         int loginTimeout;
         try {
-            loginTimeout = ds1.getLoginTimeout();
+            loginTimeout = ds2.getLoginTimeout();
```
gjwatts: Why the switch to ds2?

dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
```diff
@@ -294,7 +247,7 @@ public void testConfigChangeAuthData() throws Throwable {
         String originalUserName = testContainer.getUsername();
 
         // First check that authData matches with what was provided by the TestContainer
-        runInServlet(basicfat, "DataSourceTestServlet", "testConfigChangeAuthDataOriginalValue", "originalUsername=" + originalUserName);
+        runTestWithResponse(server, basicfat, "testConfigChangeAuthDataOriginalValue&originalUsername=" + originalUserName);
```
gjwatts: What was behind the decision to the switch from runInServlet to runTestWithResponse?  Were you just wanting to eliminate redundant testing methods?

dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/src/basicfat/DataSourceTestServlet.java
KyleAure: ds1 is a java.sql.DataSource type.  Not all database drivers support a loginTimeout property when using that type of DataSource.  Therefore, I switched to use ds2 which is a javax.sql.XADataSource type which should support loginTimeout on all database drivers.

dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
KyleAure: Yes!  runTestWithResponse was available through the `componenttest.topology.utils` package.  I hadn't realized that before.

dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/src/basicfat/DataSourceTestServlet.java
gjwatts: Thanks!

PR #10406, opened 2020-01-13
Fix server script handling of java options variables

brenthdaniel:
Users should be able to specify either IBM_JAVA_OPTIONS or OPENJ9_JAVA_OPTIONS (though the former is deprecated, so the latter is recommended.) We will prefer OPENJ9_JAVA_OPTIONS, but will set both so that older JDKs will still pick up the value. 

We also shouldn't add Xshareclasses if it is already specified.


brenthdaniel:
#build

brenthdaniel:
#build

brenthdaniel:
#run-libby-bot

brenthdaniel:
PR for #9662

PR #10413, opened 2020-01-14
Remove intrinsic lock for access _appTypeSupport

tjwatson:
Currently all access to _appTypeSupport is protected by
the intrinsic lock on ApplicationConfigurator.  This causes
lock contention with the rest of the methods on
ApplicationConfigurator that also use the same lock.

This change removes that lock for _appTypeSupport and
replaces it with a ConcurrentMap to help break the contention


tjwatson:
#build

tjwatson:
#build

tjwatson:
Abandoning this old PR.  I don't recall the details of why this was needed.  I suspect there was a rare deadlock that occurred ... but doesn't seem to be a very likely one and the risk of this change is not warranted right now.

PR #10420, opened 2020-01-14
Update equinox framework to 3.15.200.v20200113-1015 to disable log entry location capture

### LINKED ISSUE #10412: Update Equinox Framework to disable log entry location capture
With tracing enabled for the logservice in Liberty the overhead of capturing the StackTraceElement is expensive.  In Liberty we do not use the LogEntry.getLocation method in our own logging of the entries so the work is wasted and ends up really hurting performance while gathering trace.

In Equinox a new option has been added to disable the capture of log locations with https://bugs.eclipse.org/bugs/show_bug.cgi?id=559036

Setting equinox.log.capture.entry.location=false in the framework configuration will now disable this.  We should update to the latest I-Build of the Equinox Framework and set equinox.log.capture.entry.location=false by default in Liberty

anjumfatima90:
fixes #10412


anjumfatima90:
#build

tjwatson:
The first build does look fine.  Seems to have already known issues.  Just running again to make sure.

#build

PR #10423, opened 2020-01-14
Try to make use of OSGi Services more robust

tevans78:
#build 

Still getting some occasional exceptions on shutdown :-(


tevans78:
#run-libby-bot

PR #10427, opened 2020-01-14
Introspection addition

ezhou365:


ezhou365:
#build

ezhou365:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat,com.ibm.ws.security.social_fat.LibertyOP,com.ibm.ws.security.social_fat.delegated,com.ibm.ws.security.social_fat.multiProvider,com.ibm.ws.security.social_fat.oidcCertification

dev/com.ibm.ws.security.social/src/com/ibm/ws/security/social/internal/Oauth2LoginConfigImpl.java
```diff
@@ -268,6 +272,19 @@ boolean isConfiguredForProxyFlow(Map<String, Object> props) {
...
+    boolean isIntrospectConfiguration(Map<String, Object> props) {
+        String userApiType = configUtils.getConfigAttribute(props, KEY_userApiType);
+        if (userApiType != null && "introspect".equals(userApiType)) {
+            return true;
+        }
+        return false;
+    }
+    
+    protected void checkForRequiredConfigAttributesForIntrospect(Map<String, Object> props) {
+        configUtils.getRequiredConfigAttribute(props, KEY_clientId);
```
ayoho: I'd switch this to use the `getRequiredConfigAttribute()` method that's defined in this class. If you keep this call as you have it now, this will emit a message that says
> CWWKS6104W: The configuration attribute [clientId] that is required is missing or empty and a default value is not provided. Verify that the attribute is configured, that it is not empty, and that it does not consist of only white space characters.

That's technically correct, but the message doesn't tell us _which element_ is missing that attribute. Using the other method will emit this error message:
>CWWKS5479E: The configuration attribute [clientId] that is required in the social login configuration [someSpecificSocialRP] is missing or empty. Verify that the attribute is configured, that it is not empty, and that it does not consist of only white space characters.

That helps a server admin because then they know that the `someSpecificSocialRP` element was the one with the problem. A server could have a bunch of social login configurations, so not knowing which specific one caused the problem would be a pain.

dev/com.ibm.ws.security.social/src/com/ibm/ws/security/social/internal/utils/IntrospectUserApiUtils.java
```diff
@@ -0,0 +1,120 @@
...
+import com.ibm.ws.security.social.TraceConstants;
+import com.ibm.ws.security.social.error.SocialLoginException;
+import com.ibm.ws.security.social.internal.Oauth2LoginConfigImpl;
+import com.ibm.ws.common.internal.encoder.Base64Coder;
+
+public class IntrospectUserApiUtils {
+
+    public static final TraceComponent tc = Tr.register(IntrospectUserApiUtils.class, TraceConstants.TRACE_GROUP, TraceConstants.MESSAGE_BUNDLE);
+
+    Oauth2LoginConfigImpl config = null;
```
ayoho: Strictly speaking, this just needs to be a `SocialLoginConfig` object and not a `Oauth2LoginConfigImpl` object. Making the change would avoid having to do a cast when this class' constructor is called by the runtime. I'd recommend switching it.

dev/com.ibm.ws.security.social/src/com/ibm/ws/security/social/internal/utils/IntrospectUserApiUtils.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private JsonObject getJsonResponseIfValid(String response) throws SocialLoginException {
+        if (response == null || response.isEmpty()) {
+            throw new SocialLoginException("INTROSPECT_USER_API_RESPONSE_NULL_EMPTY", null, null);
+        }
+        try {
+            return Json.createReader(new StringReader(response)).readObject();
+        } catch (JsonParsingException e) {
+            throw new SocialLoginException("INTROSPECT_USER_API_RESPONSE_NOT_JSON", null, new Object[] { response, e });
```
ayoho: I just noticed the code that this was based on had a minor mistake. The second argument to this `SocialLoginException` constructor is an `Exception` that's the cause of this new exception being thrown. That second argument should be the variable `e` instead of `null`.

ayoho:
#run-libby-bot

PR #10433, opened 2020-01-14
test case for rollback due to execution time exceeding missedTaskThreshold

njr-11:
Add test coverage for transaction timeout/rollback that occurs due to execution time exceeding the missed task threshold.


njr-11:
#build

Issue #10293, opened 2020-01-06
Test Failure: com.ibm.ws.testing.opentracing.test.FATOpentracing.testImmediate


fmhwong:
RTC 270522

```
testImmediate:junit.framework.AssertionFailedError: EQ [ Operation ] expected:<[GET:com.ibm.ws.testing.opentracing.service.FATOpentracingService.]getTracerState> but was:<[http://localhost:8010/serviceApp/rest/testService/]getTracerState>
at com.ibm.ws.testing.opentracing.test.FATOpentracing.assertEq(FATOpentracing.java:211)
at com.ibm.ws.testing.opentracing.test.FATOpentracing.verifyTracerStateEvent(FATOpentracing.java:615)
at com.ibm.ws.testing.opentracing.test.FATOpentracing.verifyTracerStateEvent(FATOpentracing.java:597)
at com.ibm.ws.testing.opentracing.test.FATOpentracing.verifyTracerStateEvent(FATOpentracing.java:591)
at com.ibm.ws.testing.opentracing.test.FATOpentracing.testImmediate(FATOpentracing.java:657)
at componenttest.rules.repeater.RepeatTests$CompositeRepeatTestActionStatement.evaluate(RepeatTests.java:115)

```

lamkavan:
After careful analysis by Felix  Wong, the root cause of this issue is caused by a synchronization  problem found within the implementation of opentracing1.x  PR to resolve issue is here https://github.com/OpenLiberty/open-liberty/pull/10442

In the implementation there is 
`protected void activate(ComponentContext context) {
        setContainerFilter();
        setClientFilter();
    }`
and 
`protected void setOpentracingFilterHelper(OpentracingFilterHelper helper) {
        this.helper = helper;
        if (containerFilter != null) {
            containerFilter.setFilterHelper(helper);
        }
        if (clientFilter != null) {
            clientFilter.setFilterHelper(helper);
        }
    }`

These two methods needs to be synchronized such that they do not both run at the same time. If these methods run at the same time it can result in the helper for either the Container or Client filter to be NULL when using MicroProfile OpenTracing.

PR #10442, opened 2020-01-15
Add synchronizing so Container/Client Filters don't have null Helpers when using MPOpentracing

lamkavan:
In relation to https://github.com/OpenLiberty/open-liberty/issues/10293

### DIFF
--- a/dev/com.ibm.ws.opentracing.1.1/src/com/ibm/ws/opentracing/OpentracingJaxRsProviderRegister.java
+++ b/dev/com.ibm.ws.opentracing.1.1/src/com/ibm/ws/opentracing/OpentracingJaxRsProviderRegister.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 IBM Corporation and others.
+ * Copyright (c) 2017, 2020 IBM Corporation and others.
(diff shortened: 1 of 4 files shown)


lamkavan:
#build

lamkavan:
#build

lamkavan:
Build failure not related to changes

lamkavan:
#build

lamkavan:
**Problem**
In the implementation for MPOpentracing 1.0 to 1.3 there is
protected void activate(ComponentContext context)
and
protected void setOpentracingFilterHelper(OpentracingFilterHelper helper)
in OpentracingJaxRsProviderRegister.java
If these two methods get executed at the same time on different threads then it can lead to either the container or client filter being set to null which leads to the reported issue at  https://github.com/OpenLiberty/open-liberty/issues/10293. In short, this is a simple synchronization problem.

**Solution**
These two methods needs to be synchronized such that they do not both run at the same time. If these methods run at the same time it can result in the helper for either the Container or Client filter to be NULL when using MicroProfile OpenTracing. The order of execution of these two methods does not matter so simply adding the `synchronized ` keyword to the methods to ensure the two don't get run at the same time is sufficient to resolve the issue.

PR #10446, opened 2020-01-15
Isolating table creation to run in a separate transaction

KyleAure:
I had previously changed the AutomaticDatabase bean to manage it's own transactions as the initTable method was failing to execute since the container was suspending global transactions.  

Instead, I have re-enabled transactions to be managed by the container.  And to ensure that initTable runs I have added the `@PostConstruct` annotation so that it runs before any other business methods.  
And `@TransactionAttribute(TransactionAttributeType.NOT_SUPPORTED)` to ensure that the initTable method is run outside of the container managed transactions.


KyleAure:
#build

PR #10455, opened 2020-01-15
Issue #10453: Auto-link EJB ref in java global with ejb-link

### LINKED ISSUE #10453: Auto-link EJB references in java:global with beanName or ejb-link
Currently, EJB references defined in java:global are required to have a binding in ibm-*-bnd.xml, to link the ejb reference (`@EJB`, `ejb-ref`, or `ejb-local-ref`) to an actual EJB.  Primarily this is because ejb-link only works within an application, whereas it has been assumed a ref in java:global will bind to an EJB in a different application.

However, if an EJB definition would like to define a reference to itself in java:global, so that it may make itself available at a non-standard location, then this isn't possible without a binding.

For example:

```
@Stateless
@Remote(ClientRemote.class)
@EJB(name = "java:global/env/ejb/ClientBeanRef", beanInterface = ClientRemote.class, beanName = "ClientBean")
public class ClientBean {
```

Would declare an EJB which may be looked up or injected from two locations in JNDI:

Spec defined location : ```java:global/<app>/<module>/ClientBean!xxx.xxx.ClientRemote```
EJB ref location : ```java:global/env/ejb/ClientBeanRef```

Currently, the above definition will fail, unless the ibm-ejb-jar-bnd.xml file is also provided and defines a binding as:

```
  <session name="ClientBean">
    <ejb-ref name="java:global/env/ejb/ClientBeanRef" binding-name="java:global/InjectionMiscTestApp/InjectionMiscBean/ClientBean"/>
  </session>

```

This should be supported without the binding, since the `beanName` of the `@EJB` annotation references a bean in the same application.

tkburroughs:
Relax the checking in EJBProcessor to allow ejb references
in java:global to use auto-link when the application name
is known and ejb-link has been specified (i.e. ejb-link or beanName).

This will allow EJB references in java:global to auto-link to
an EJB specified by ejb-link that exists in the same application.

If the EJB is in a different application, a binding will still be
required.

Also removed a binding from a test that was previously added to
demonstrate the current behavior. The test will now properly link
to the EJB without the need for a binding.

Please excuse the reformatting.... the only changes were the additions
of lines 233-235 in the new version of EJBProcessor.

fixes #10453

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ws/ejb/injection/processor/EJBProcessor.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ws/ejb/injection/processor/EJBProcessor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2013 IBM Corporation and others.
+ * Copyright (c) 2006, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


tkburroughs:
#build

tkburroughs:
A FAT failures in build have linked defects

PR #10473, opened 2020-01-16
Issue #10461: Fix issue where basic registry throws a PatternSyntaxEx…

### LINKED ISSUE #10461: Basic registry throws PatternSyntaxException when search filter contains paren
When the search filter passed in to a `UserRegistry.getUsers(...)` or `UserRegistry.getGroups(...)` contains a paren (i.e. "\*(\*" of "\*)\*") the basic registry will throw a `java.util.regex.PatternSyntaxException` with a message similar to `Unclosed group near index` or `Unmatched closing ')' near index`.

If using federated repositories, and `allowOpsIfReposDown` is set on the federated realm, the exception may be consumed and instead no results will be returned to the user. In this case, to determine cause, trace for federated repositories will need to be enabled (`com.ibm.ws.security.wim.\*=all:com.ibm.websphere.security.wim.\*=all`) while reproducing the issue. The exception may then be found in the server's trace logs.

The basic registry was not properly escaping the parens in the regular expression generated from the search filter, and therefore Java's regular expression pattern matching was assuming the paren was intended to be a group capture instead of part of the pattern to search for.

jvanhill:
…ception when a paren is included in a search filter.

fixes #10461

### DIFF
--- a/dev/com.ibm.ws.security.registry.basic/src/com/ibm/ws/security/registry/basic/internal/BasicRegistry.java
+++ b/dev/com.ibm.ws.security.registry.basic/src/com/ibm/ws/security/registry/basic/internal/BasicRegistry.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011,2018 IBM Corporation and others.
+ * Copyright (c) 2011,2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


jvanhill:
#build

jvanhill:
#runlibby

jvanhill:
#run-libby-bot

jvanhill:
The fats that failed are unrelated to my changes and have been triaged to existing build breaks.

PR #10480, opened 2020-01-17
move long running tests in errorpaths bucket to full mode

njr-11:
There are a handful of tests in the errorpaths test bucket, such as the tests that capture a dump of the server and test the introspector output, that take a long time to run, and therefore belong under full mode instead of lite.


njr-11:
#build

PR #10484, opened 2020-01-17
EJB Fat wait for security

tkburroughs:
Updated EJB fat servers that use appSecurity-2.0 to wait for
security and LTPA to start before running tests.


tkburroughs:
#build

PR #10499, opened 2020-01-20
Allow postgres to engage in two phase commit transactions

KyleAure:
Postgres was failing with the following error:

```
[01/20/2020 15:46:30:044 CST] 046 PostgreSQLContainer            output                         I ERROR:  prepared transactions are disabled
[01/20/2020 15:46:30:045 CST] 046 PostgreSQLContainer            output                         I HINT:  Set max_prepared_transactions to a nonzero value.
[01/20/2020 15:46:30:045 CST] 046 PostgreSQLContainer            output                         I STATEMENT:  PREPARE TRANSACTION '1463898948_AAABb8Tr3vIAAAABE64luOLe+k3znXUtazKvPrWOA7w3dYE9_AAABb8Tr3vIAAAABE64luOLe+k3znXUtazKvPrWOA7w3dYE9AAAAAQAAAAAAAAAAAAAAAAAC'
```

This is because by default Postgres is unable to prepare transactions. Therefore, by default cannot engage in a two phase commit.  I have set the max number of prepared transactions to 2 to allow Postgres to engage in two phase commits.  This number will likely need to be increased if this application is going to be run under heavier load.


KyleAure:
#build

PR #10506, opened 2020-01-21
Fixing config change test for auth data

KyleAure:
A previous change that had been made to this test case changed the original call to use the Test Containers username.  However, this test should only be running against derby.  I have switched it back to ignore test containers and to make sure we are using a derby only datasource. 

Another (minor) issue with this test case is that on the server side we are using dsfat6 to get a connection between config changes.  However, dsfat6 is defined using `@DatasourceDefinition`, but we do not document anywhere on the client side code that the authentication configuration is being overridden by ibm-web-bnd.xml.  So it looks like the test is changing authdata and using a datasource to confirm the change that doesn't itself use that authdata.


KyleAure:
#build

gjwatts:
#libbybot

KyleAure:
#build

PR #10522, opened 2020-01-21
Ensure duplicate key exception is avoided

KyleAure:
We were running into issues on our test systems where each time the AutomaticDatabase timer ran it would attempt to create the row where we are trying to keep track of executions.  This resulted in a duplicate key exception being thrown by the database. Likely due to a timing issue.

Here I have changed the test to ensure that the row is created before the timer, and there is only one attempt to create the row.  I did this by inserting the down during the `PostConstruct` method, after table creation.


KyleAure:
#build

KyleAure:
#build 
#libbybot

PR #10554, opened 2020-01-23
Use service reference

brenthdaniel:
Use a ServiceReference for the ApplicationRecycleCoordinator to avoid lock contention


brenthdaniel:
#build

PR #10557, opened 2020-01-23
disabled test for login module in standalone resource adapter

njr-11:
Copy the existing tests for JAAS login module loaded from a library into a test that will, when implementation is ready, be switched over to loading the JAAS login module from a standalone resource adapter.  In order to confirm the tests are otherwise correct, I have temporarily pointed a libraryRef at the RAR file.  This can switch to classProviderRef when that feature is implemented.


njr-11:
#build

PR #10564, opened 2020-01-23
Handle defunct processes

brenthdaniel:
When we check to see if a server is possibly still running we need to take `<defunct>` processes into consideration.


brenthdaniel:
#build

Issue #10529, opened 2020-01-22
Login module in an embedded resource adapter


njr-11:
User story:
"As a user of Liberty, I want to configure a JAAS login module that is embedded within a resource adapter which is itself embedded within an application so that the login module can be used for authenticating to resources that are created by the resource adapter."

PR #10567, opened 2020-01-23
port JCA enterpriseApp test bucket to open liberty

njr-11:
It will be helpful to port some of the com.ibm.ws.jca_fat_enterpriseApp bucket to Open Liberty because it includes a resource adapter embedded in an application that would be convenient to add a login module to in order to test #10529


njr-11:
#build

PR #10584, opened 2020-01-24
create embedded resource adapter with embedded login module

njr-11:
Create an embedded resource adapter which provides its own login module, which will be useful for testing the classProviderRef attribute of jaasLoginContextEntry, once implemented.


njr-11:
#build

PR #10596, opened 2020-01-27
Ensure duplicate key exception is avoided attempt 2

KyleAure:
Continued from PR: #10522 
We have continued to see Duplicate Key exceptions popping up in our SOE builds from this test bucket.  Without JDBC trace I was unable to locate exactly when we are originally creating the row.  My best guess is that by having this timer bean as `@Stateless` has somehow resulted in more than one bean being created, and therefore more then one attempt to create the same row in the database.

I wasn't able to recreate this error locally.  But to ensure that this is not possible I have changed the AutomaticDatabase bean to be a `@Singleton`.  This should ensure that the PostConstruct method is only ever called once.  And added in an extra check to see if the row has previously been created.

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/test-application/demotimers/src/ejb/timers/AutomaticDatabase.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/test-application/demotimers/src/ejb/timers/AutomaticDatabase.java
@@ -23,7 +23,7 @@ import javax.annotation.PostConstruct;
 import javax.annotation.Resource;
 import javax.ejb.Schedule;
 import javax.ejb.SessionContext;
-import javax.ejb.Singleton;
+import javax.ejb.Stateless;
 import javax.ejb.Timer;
 import javax.ejb.TransactionAttribute;
 import javax.ejb.TransactionAttributeType;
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/test-application/demotimers/src/ejb/timers/AutomaticDatabase.java
```diff
@@ -33,7 +33,7 @@
  * This class uses the @Schedule annotation.
  * Using this annotation will start the timer immediately on start and will run every 30 seconds.
  */
-@Stateless
+@Singleton //Ensure only one instance of this Timer is ever created.
```
njr-11: Singleton does seem convenient for restricting the table setup, but I think there may be other consequences for timers with respect to locking, so we need to ensure that is really what we want.

dev/com.ibm.ws.concurrent.persistent_fat_demo_timers/test-application/demotimers/src/ejb/timers/AutomaticDatabase.java
```diff
@@ -88,30 +88,54 @@ private void initTable() throws SQLException {
...
+                //If the count of rows is more than 0 then set this true.
+                //Otherwise, set to false when row is not found, or 0 is returned as a result.
+                boolean exists = rs.next() ? rs.getInt(1) > 0 : false;
+                if (exists) {
+                    System.out.println("Found row identified by " + name + " in table AUTOMATICDATABASE.  Skipping creation.");
+                    return;
+                }
+            } catch (SQLException sqle) {
+                System.out.println("Caught exception attempting to find row identified by " + name + " in table AUTOMATICDATABASE."
+                                   + " Continuing on to attempt to create row.");
```
njr-11: I don't see why we should suppress this error.  If something goes wrong, we should let the test fail and report the problem.

KyleAure:
#build 
#libbybot

PR #10601, opened 2020-01-27
disabled test for two embedded resource adapters with login modules

njr-11:
Enhance the tests for embedded resource adapters with login modules to add a login module into the second embedded RAR so that both have login modules.  This will help that we are able to load the right one even if there are multiple.  But we cannot enable the test until the function is added, so for now, it will also package the login modules in a separate library, which we can remove when the real function is ready to test.

### DIFF
--- a/dev/com.ibm.ws.jca_fat_enterpriseApp/test-resourceadapters/enterpriseAppRA/src/com/ibm/test/jca/enterprisera/EALoginModule.java
+++ b/dev/com.ibm.ws.jca_fat_enterpriseApp/test-resourceadapters/enterpriseAppRA/src/com/ibm/test/jca/enterprisera/EALoginModule.java
@@ -46,7 +46,6 @@ public class EALoginModule implements LoginModule {
 
             Map<?, ?> properties = mpropsCallback.getProperties();
             final String loginName = (String) properties.get("loginName");
-            System.out.println("### LOGINMODULE PROP " + loginName);
 
             AccessController.doPrivileged(new PrivilegedAction<Void>() {
                 @Override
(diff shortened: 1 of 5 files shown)


dev/com.ibm.ws.jca_fat_enterpriseApp/test-resourceadapters/enterpriseAppRA/src/com/ibm/test/jca/enterprisera/EALoginModule.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public boolean login() throws LoginException {
+        try {
+            final WSManagedConnectionFactoryCallback mcfCallback = new WSManagedConnectionFactoryCallback("Target ManagedConnectionFactory: ");
+            WSMappingPropertiesCallback mpropsCallback = new WSMappingPropertiesCallback("Mapping Properties (HashMap): ");
+            callbackHandler.handle(new Callback[] { mcfCallback, mpropsCallback });
+
+            Map<?, ?> properties = mpropsCallback.getProperties();
+            final String loginName = (String) properties.get("loginName");
+            System.out.println("### LOGINMODULE PROP " + loginName);
```
KyleAure: Not sure if you meant to leave this print statement here.  If so, maybe something like, `Found login module property key=loginName value=`

njr-11: Good catch. That was a mistake to leave the println, and I have removed it under https://github.com/OpenLiberty/open-liberty/pull/10601/commits/34432f1fea78f20fc714e96be0596399d5d210d0

njr-11:
#build
#run-libby-bot

PR #10634, opened 2020-01-29
Added new async request context test for when Future.cancel() occurs

Joseph-Cass:

### DIFF
--- a/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
+++ b/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
@@ -1562,6 +1562,11 @@
       <version>2.0.1</version>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+      <artifactId>microprofile-fault-tolerance-api</artifactId>
+      <version>2.1-RC2</version>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.microprofile.graphql</groupId>
       <artifactId>microprofile-graphql-api</artifactId>
       <version>1.0-M3</version>
--- a/dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AsyncTest.java
+++ b/dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AsyncTest.java
@@ -15,6 +15,7 @@ import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.lang.reflect.Method;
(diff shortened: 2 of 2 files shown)


dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AsyncTest.java
```diff
@@ -204,6 +202,46 @@ public void testAsyncRequestContextDeactivatedOnInterruption() throws Interrupte
...
+        builder.setRequestContextController(asyncRequestContextController);
+
+        Executor<Future<String>> executor = builder.buildAsync(Future.class);
+        ExecutionContext context = executor.newExecutionContext("testAsyncCS", null);
+
+        CountDownLatch startSignal = new CountDownLatch(1); // Prevents the execution from being interrupted until request context is set
+
+        // Test an execution which is cancelled
+        Future<String> result = executor.execute(() -> {
+            assertEquals("Activate request context should be called when request is started", 1, asyncRequestContextController.getActivateContextCount());
```
Azquelt: This assert won't cause the test to fail correctly as it's inside the executor's callable and the result of that is thrown away because you cancel the execution.

However, if this assert ever did fail, you wouldn't call `countDown()` on line 222 and then your test would hang waiting for it on line 231.

I don't really think this check is necessary as you're already testing this elsewhere, so you can just take it out.

dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
```diff
@@ -1561,6 +1561,11 @@
       <artifactId>microprofile-fault-tolerance-api</artifactId>
       <version>2.0.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+      <artifactId>microprofile-fault-tolerance-api</artifactId>
+      <version>2.1-RC2</version>
+    </dependency>
```
Azquelt: Someone else has delivered this change, so you can take it out of your PR.

dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AsyncTest.java
```diff
@@ -204,6 +202,46 @@ public void testAsyncRequestContextDeactivatedOnInterruption() throws Interrupte
...
+            startSignal.countDown();
+
+            // Wait for up to 10 seconds for the execution to be interrupted
+            Thread.sleep(FUTURE_TIMEOUT);
+
+            return null;
+        }, context);
+
+        // Wait for the execution to have started before interrupting
+        startSignal.await();
```
Azquelt: Add a timeout here as we don't want to wait forever. Also then assert that it returned `true`.

dev/com.ibm.ws.microprofile.faulttolerance.2.0/test/src/com/ibm/ws/microprofile/faulttolerance/test/AsyncTest.java
Joseph-Cass: Okay that's great, thanks for the explanation

Azquelt:
#run-libby-bot

Joseph-Cass:
#build

Joseph-Cass:
#build

PR #10643, opened 2020-01-29
Fix bug where metrics monitor is not properly removing ending brackets for REST metrics

Channyboy:
Bug in metrics-23-monitor-1.0 code where it is not appropriately removing the ending bracket from the method signature retrieved from the REST stat JMX beans.

#build


Channyboy:
Two failures reported in the build related for ui.tool FATs which have no relation to the changes here.

The changes here only affect mpMetrics-2.3 (currently a new beta feature).
The change here removes a bracket from the metrics output for REST metrics.

Matching RTC Defect 269311

PR #10652, opened 2020-01-30
Changed tck_suite.xml DOCKTYPE certificates to use http rather than https

Joseph-Cass:
Some tck-suite files referenced the DTD with an https URL which was causing issues in sun JVMs which don't accept the certificate.


Joseph-Cass:
#build

PR #10657, opened 2020-01-30
update builder and consumer tests to wait for server initialization to complete

c00crane:
Update the builder and consumer FAT tests to wait for server initialization to complete before trying to run any of the tests.
Adding waits for apps to initialize and in some cases, SSL init to complete.


c00crane:
#build

PR #10661, opened 2020-01-30
Don't set interruptible flag in setAppHandler

brenthdaniel:
Change to remove setting the 'interruptible' flag in the app manager when setAppHandler(null) is called, issue to follow


brenthdaniel:
#build

brenthdaniel:
#10672

PR #10679, opened 2020-01-31
disabled test of login module in JARs within RAR

njr-11:
Add login modules to JAR files that are inside of a resource adapter RAR at various levels.  This will be the basis for a test JAAS login context entries can be defined in server config pointing to login modules within a RAR which are not at top level.  The test will be disabled now, (by pointing at at standalone JAR instead) given that the necessary function has not been added.  The tests can be turned on to verify later when it is added.


njr-11:
#build

PR #10683, opened 2020-02-01
Return InitialContext for empty name lookups on root level context

### LINKED ISSUE #9099: Jndi Empty lookup should return top context
In WebSphere Application Server you could do a lookup like this: `ivContext = (Context) new InitialContext().lookup("");` which had some benefit. It does not work in Open Liberty though, and even though it wouldn't provide any benefit it, it could be a migration issue for customer applications.

The reason it doesn’t is most likely that the first lookup is returning a specific Context rather than an InitialContext, we are binding to the default namespace.
After that, an attempt to look up something with a namespace prefix is just treated as a lookup within the default namespace.

The solution could be to detect the conditions are right from within the implementation, and then could always just return new InitialContext()

dhaggart:
fixes #9099


dhaggart:
#run-libby-bot

dhaggart:
#build

dhaggart:
Build failures unrelated to changes (can't contact MQ server & admin UI test).

PR #10698, opened 2020-02-03
Rework how kafka connector handles partition rebalancing

### LINKED ISSUE #10169: mpReactiveMessaging liberty-kafka connector should discard messages after rebalance
**Is your feature request related to a problem? Please describe.**
I have a reactive messaging application with a processor method, with input and output channels connected to Kafka topics.  I have multiple instances of this application as part of a consumer group. While my application is running, a consumer group rebalance occurs and partitions are revoked. My processor method still gets called with messages belonging to the revoked partition.  The partition is assigned to another consumer, which will also process these messages. 

**Describe the solution you'd like**
The connector should respond to the rebalance by discarding any messages from revoked partitions, so that they are not delivered to the processor method.  This makes it less likely that duplicate message processing occurs. 

**Describe alternatives you've considered**
If I were implementing a Kafka consumer by hand I could respond to the rebalance in my application code and discard any fetched messages that I should no longer process.  However, reactive messaging hides these details from me, so this feels like the sort of thing it should handle. 

**Additional context**

### LINKED ISSUE #10281: Kafka Connector can commit offset backwards
**Describe the bug**
If a kafka partition is unassigned and then reassigned, the kafka connector goes back to the most recently committed offset for that partition. Messages retrieved before the partition was unassigned can still be processed, leading to messages being processed twice. See also #10169 

When partitions are unassigned, the Kafka connector closes the PartitionAckTracker, so no messages which use that tracker should be committed.

However, when the same partition is assigned again, a new PartitionAckTracker is created and I _think_ that messages retrieved before the unassignment may still be waiting to be associated with a PartitionAckTracker and may then pick up the new PartitionAckTracker rather than the old one.

This means that the new ParititionAckTracker may receive a string of messages like this:
`[3, 4, 5, 1, 2, 3, 4, 5, 6, ...]`

The PartitionAckTracker isn't designed to handle this and can end up committing offsets 3, 4, 5 but then going back and committing the offset back to 1.

**Expected behavior**

We shouldn't ever commit the offset backwards. We could change the design of the system to confirm this, or we can ensure that any given `PartitionAckTracker` only receives messages in ascending order.

**Diagnostic information:**
 - OpenLiberty Version: 19.0.0.11
 - Java Version: [i.e. full output of `java -version`]

Azquelt:
Redesign the Kafka connector slightly to give a stronger pairing between
a message and the partition tracker associated with it. This prevents
confusion when a partition is unassigned and then reassigned.

Avoid emitting messages if the partition those messages came from has
since been unassigned (and therefore an attempt to commit the partition
offset would fail).

Fixes #10281
Fixes #10169


Azquelt:
#build

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
```diff
@@ -0,0 +1,304 @@
...
+    private final KafkaAdapterFactory factory;
+    private final ScheduledExecutorService executor;
+    private final KafkaInput<?, ?> kafkaInput;
+    private final int maxCommitBatchSize;
+    private final Duration maxCommitBatchInterval;
+
+    /**
+     * Set of CompletedWork which has either not been committed or has been committed but the commit has not completed yet.
+     * <p>
+     * Any access to this must be synchronized
```
tevans78: synchronized on what? assume it is always `completedWork` for all of these variables?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
```diff
@@ -0,0 +1,304 @@
...
+    private Future<?> pendingCommitTask = null;
+
+    /**
+     *
+     * @param topicPartition the partition to track
+     * @param factory the KafkaAdaptorFactory
+     * @param kafkaInput the KafkaInput
+     * @param initialCommittedOffset the position of the reader when the partition was assigned
+     * @param executor a ScheduledExecutorService
+     * @param maxCommitBatchSize the maximum number of records to wait for before committing the offset
```
tevans78: Does a value less than or equal to zero actually mean unlimited batch size?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
```diff
@@ -0,0 +1,304 @@
...
+                Tr.debug(this, tc, "Rejecting commit attempt because partition is closed", this);
+            }
+
+            CompletableFuture<Void> result = new CompletableFuture<>();
+            result.completeExceptionally(new Exception("Partition is closed"));
+            return result;
+        }
+
+        // Note work.offset + 1
+        // Committed offset is the _next_ message we expect to receive
```
tevans78: huh? `Committed` is past tense, `next message` is future??

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
```diff
@@ -0,0 +1,304 @@
...
+                    break;
+                }
+                committedWork.add(completedWork);
+                i.remove();
+            }
+        }
+
+        if (exception == null) {
+            for (CompletedWork completedWork : committedWork) {
+                completedWork.completion.complete(null);
```
tevans78: Maybe use a different local variable name? To avoid confusion with the instance TreeSet variable above.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
```diff
@@ -0,0 +1,304 @@
...
+        this.maxCommitBatchSize = maxCommitBatchSize;
+        this.maxCommitBatchInterval = maxCommitBatchInterval;
+    }
+
+    @Override
+    public CompletionStage<Void> recordDone(long offset, Optional<Integer> leaderEpoch) {
+
+        CompletableFuture<Void> result = new CompletableFuture<>();
+        synchronized (completedWork) {
+            completedWork.add(new CompletedWork(offset, leaderEpoch, result));
```
tevans78: It might never happen (or matter) but... you seem to have code that copes with the same offset being recorded more than once; if that happens should you really increment `outstandingUncommittedWork`?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
```diff
@@ -0,0 +1,304 @@
...
+        return result;
+    }
+
+    /**
+     * Request that done but uncommitted work is committed, either now or in the future
+     * <p>
+     * This method will either commit the partition offset now, or schedule it to be done in the future, depending on the values of {@link #maxCommitBatchInterval} and
+     * {@link #maxCommitBatchSize}.
+     * <p>
+     * Calls to this method must be synchronized.
```
tevans78: synchronized on what?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
Azquelt: Yes, I'll add doc for zero being a special value for both `maxCommitBatchSize` and `maxCommitBatchInterval`.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
Azquelt: Does this make more sense:
```java
        // Note work.offset + 1
        // In general the committed offset for a partition is the first message which should be received by a new consumer
        // which starts consuming from that partition.
        // Therefore, the offset which we are about to commit must be the offset _after_ the last message we have processed.
```

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaInput.java
```diff
@@ -12,11 +12,11 @@
 
 import static com.ibm.websphere.ras.TraceComponent.isAnyTracingEnabled;
 import static java.time.Duration.ZERO;
-import static org.eclipse.microprofile.reactive.streams.operators.ReactiveStreams.fromIterable;
 
 import java.time.Duration;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
```
tevans78: copyright date

tevans78: probably needs checking thoughout

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
Azquelt: We ensure that everything works if `commitCompletedWork()` is called a second time before the `CompletionStage` from the previous commit has completed, but we don't actually do anything to protected against `recordDone()` being called twice with the same offset.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTracker.java
```diff
@@ -0,0 +1,80 @@
...
+     * Record that processing of a record has been completed.
+     * <p>
+     * The result of this method can be used as the return value from {@link Message#ack()}
+     *
+     * @param offset the record offset
+     * @param leaderEpoch the record leaderEpoch
+     * @return a CompletionStage which completes when any associated processing has been completed (e.g. when the message offset has been committed)
+     */
+    public CompletionStage<Void> recordDone(long offset, Optional<Integer> leaderEpoch) {
+        return CompletableFuture.completedFuture(null);
```
tevans78: what if isClosed == true?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTrackerFactory.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ *
+ */
+public class PartitionTrackerFactory {
+
+    private KafkaAdapterFactory adapterFactory = null;
+    private ScheduledExecutorService executor = null;
+    private int commitBatchMaxElements = 500;
+    private Duration commitBatchMaxInterval = Duration.ofMillis(500);
+    private boolean manualCommit = false;
```
tevans78: Does `manualCommit = !autoCommit`? It's called autoCommit elsewhere, why the inversion?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/ThresholdCounter.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Returns a completion stage which completes when the counter is less than the threshold value
+     * <p>
+     * If the counter is already below the threshold value then a completed completion stage is returned
+     *
+     * @return completion stage which completes when the counter is less than the threshold value
+     */
+    CompletionStage<Void> waitForBelowThreshold();
+
+    ThresholdCounter DUMMY = new ThresholdCounter() {
```
tevans78: is there a better name than `DUMMY`?

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/flatmap/package-info.java
```diff
@@ -0,0 +1,14 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+/**
+ *
+ */
+package com.ibm.ws.microprofile.reactive.messaging.fat.kafka.flatmap;
```
tevans78: Why is this file needed? If not need then I think there is another one which can be removed too.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/KafkaPartitionTest.java
```diff
@@ -56,14 +62,20 @@ public static void setup() throws Exception {
         adminClientProps.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, PlaintextTests.kafkaContainer.getBootstrapServers());
         AdminClient adminClient = AdminClient.create(adminClientProps);
 
-        NewTopic newTopic = new NewTopic(PartitionTestReceptionBean.CHANNEL_NAME, 2, (short) 1);
-        adminClient.createTopics(Collections.singleton(newTopic)).all().get(KafkaTestConstants.DEFAULT_KAFKA_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
+        List<NewTopic> newTopics = new ArrayList<>();
+        newTopics.add(new NewTopic(PartitionTestReceptionBean.CHANNEL_NAME, 2, (short) 1));
```
tevans78: is 2 partitions enough to properly exercise this code?

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/flatmap/KafkaFlatMapServlet.java
```diff
@@ -0,0 +1,58 @@
...
+        SimpleKafkaReader<String> reader = kafkaTestClient.readerFor(OUT_TOPIC);
+
+        long offset = kafkaTestClient.getTopicOffset(IN_TOPIC, APP_GROUPID);
+
+        writer.sendMessage("abc");
+        writer.sendMessage("abcd");
+        writer.sendMessage("abcde");
+        writer.sendMessage("abcdef");
+
+        List<String> messages = reader.waitForMessages(2, Duration.of(10, SECONDS));
```
tevans78: can you use a constant for the timeouts?

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestBean.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+@ApplicationScoped
+public class LivePartitionTestBean {
+
+    public static final String CHANNEL_IN = "live-partition-test-in";
+    public static final int WORK_TIME = 100;
+    public static final int FINAL_MESSAGE_NUMBER = 9999;
+
+    ArrayList<ReceivedMessage> messages = new ArrayList<>();
+    private final CountDownLatch paritionsFinished = new CountDownLatch(2);
```
tevans78: Can you use a constant for the number of partitions?

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestBean.java
```diff
@@ -0,0 +1,99 @@
...
+                status.ackStatus.set(AckStatus.ACK_FAILED);
+            }
+            return null;
+        });
+
+        return CompletableFuture.completedFuture(null);
+    }
+
+    public void awaitFinish() throws InterruptedException {
+        assertTrue("Test bean did not process all messages within 30 seconds", paritionsFinished.await(30, TimeUnit.SECONDS));
```
tevans78: timeout constant ... based on the number of partitions??

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestConsumer.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public void run() {
+        try (KafkaConsumer<?, String> consumer = new KafkaConsumer<>(config)) {
+            consumer.subscribe(Collections.singleton(topic));
+            ConsumerRecord<?, String> lastRecord = null;
+
+            int messages = 0;
+            long startTime = System.nanoTime();
+            while (messages < 5 && System.nanoTime() - startTime < MAX_DURATION) {
```
tevans78: number of messages in a constant?

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestServlet.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        Map<String, Object> consumerConfig = new HashMap<>();
+        consumerConfig.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, getKafkaBootstrap());
+        consumerConfig.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+        consumerConfig.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+        consumerConfig.put(ConsumerConfig.GROUP_ID_CONFIG, APP_GROUPID);
+
+        // Sleep
+        Thread.sleep(700);
```
tevans78: constant?

tevans78: is there any other way than using a sleep?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/ThresholdCounterImpl.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
+/**
+ *
+ */
+public class ThresholdCounterImpl implements ThresholdCounter {
```
tevans78: Can this class be unit-tested?

Azquelt:
> Is it possible that when a partition is revoked, some messages that have processed have not yet been committed? I wonder if CommitingPartitionTracker should commit any outstanding messages as part of close().

Yes, I'm considering doing that but I wanted to make sure we had the solution working for when things don't get committed.

If we do a commit when the partition is revoked, we also need to ensure that our testing is still effective.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/ThresholdCounter.java
Azquelt: I like `DUMMY`, though I'm open to other suggestions.

Azquelt: Maybe `MAX_THRESHOLD` would be appropriate, we're essentially saying that there is no threshold we must stay below.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/flatmap/KafkaFlatMapServlet.java
Azquelt: Yes

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/flatmap/package-info.java
Azquelt: No

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/KafkaPartitionTest.java
Azquelt: I think so. We're really just looking to test the case where a partition is assigned, removed, then assigned again to test that messages are handled correctly. We could add more partitions though we might then need to process more messages and make the test take longer and I'm already a bit concerned about how long this test will take.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestBean.java
Azquelt: If anything it should be based on the number of messages.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestServlet.java
Azquelt: Not really, it's trying to simulate a live environment where other consumers can join and leave the group at random. I'm open to other suggestions.

The design of the test is just to provoke a bunch of things to happen and then check that the runtime handled whatever happened in a valid way.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/ThresholdCounter.java
tevans78: `UNLIMITED`?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
tevans78: If the committed offset is the first message which should be received by a new consumer ... does that mean it's counting from 1? There is no message "index" 0?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
tevans78: zero or less

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/CommittingPartitionTracker.java
Azquelt: No, the offset of the first message is `0`.

Imagine there are three messages in the partition, none have yet been consumed. The committed offset is `0`, the next message to be received has offset `0`.

The first message is read and processed, the consumer doesn't want to receive this message again. The consumer has processed the record with offset `0`, it sets the committed offset to `1`, the next message to be received has offset `1`.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTracker.java
Azquelt: I don't think we care, this class is only used when we don't need to do anything in response to `ack()`, so I don't think we should say that we "failed" to do nothing.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTrackerFactory.java
Azquelt: It made more sense when I was writing it, since the manual case is the one where we have to do stuff. I'll change it.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestServlet.java
Azquelt: I don't think it's actually helpful to turn this into a constant. The length of time here is pretty arbitrary. It's more important for someone reading the test to be able to see at a glance how long it's sleeping for.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/partitions/LivePartitionTestConsumer.java
```diff
@@ -49,17 +54,17 @@ public void run() {
 
             int messages = 0;
             long startTime = System.nanoTime();
-            while (messages < 5 && System.nanoTime() - startTime < MAX_DURATION) {
+            while (messages < MESSAGES_TO_CONSUME && System.nanoTime() - startTime < MAX_DURATION) {
```
tevans78: use default timeout

Azquelt:
Existing commits rebased on integration and conflicts resolved, added new commits to acknowledge messages asynchronously and commit any completed work when the partition is revoked.

#run-libby-bot
#build

Azquelt:
#run-libby-bot
#build

PR #10699, opened 2020-02-03
Fixing XA Recovery issue with Postgres and optimized creating table

KyleAure:
It was found that during XA recovery Postgres (even when using isolation level Serializable) will return results for any connections that were commited, while not returning results for those that are in-doubt (that is in recovery).  And that Postgres takes a little bit longer than other databases to perform recovery

Therefore, that test has been changed to poll that database once a second for 5 seconds.  This should give postgres enough time to perform XA recovery. 

Also suppressed logs of server config changes at the end of tests as this was bloating the output.txt file and not displaying anything useful. 

Finally, decided to create the database table during the `init` phase of the servlet so that we can guarantee the table will be available for the app.

### DIFF
--- a/dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/resources/WEB-INF/ibm-web-bnd.xml
+++ b/dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/resources/WEB-INF/ibm-web-bnd.xml
@@ -5,6 +5,10 @@
   <resource-ref name="jdbc/dsfat3" binding-name="jdbc/dsfat3">
     <authentication-alias name="auth1"/>
   </resource-ref>
+  
+  <resource-ref name="jdbc/dsfat5ref1" binding-name="jdbc/dsfat5">
+    <authentication-alias name="derbyAuth1"/>
+  </resource-ref>
 
   <resource-ref name="jdbc/dsfat6ref1" binding-name="java:comp/env/jdbc/dsfat6">
     <authentication-alias name="derbyAuth1"/>
--- a/dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/src/basicfat/DataSourceTestServlet.java
+++ b/dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/src/basicfat/DataSourceTestServlet.java
@@ -33,8 +33,10 @@ import java.sql.SQLNonTransientException;
 import java.sql.SQLSyntaxErrorException;
 import java.sql.SQLTransientConnectionException;
 import java.sql.Statement;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.List;
 import java.util.Set;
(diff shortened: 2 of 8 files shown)


dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/resources/WEB-INF/ibm-web-bnd.xml
```diff
@@ -5,6 +5,10 @@
   <resource-ref name="jdbc/dsfat3" binding-name="jdbc/dsfat3">
     <authentication-alias name="auth1"/>
   </resource-ref>
+  
+  <resource-ref name="jdbc/dsfat5ref1" binding-name="jdbc/dsfat5">
+    <authentication-alias name="derbyAuth1"/>
+  </resource-ref>
```
gjwatts: What was behind moving derbyAuth1 from the datasource in server.xml to the resource reference in ibm-web-bnd.xml?  Is that part of the move to creating the table during init?

dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/src/basicfat/DataSourceTestServlet.java
```diff
@@ -155,6 +160,9 @@
     @Resource(lookup = "jdbc/dsfat5", shareable = false)
     DataSource ds5u; // one-phase, default isolation, unsharable, Derby only
 
+    @Resource(name = "jdbc/dsfat5ref1", lookup = "jdbc/dsfat5")
+    DataSource ds5_1; //one-phase, defualt isolation, Derby Only, derbyAuth1 set as authentication alias
```
gjwatts: typo: default

dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/resources/WEB-INF/ibm-web-bnd.xml
KyleAure: So the reason I did this was so that I could use `dsfat5` in the `testConfigChangeAuthData` test instead of dsfat6.  The main difference between the two, and why I wanted to change, was that dsfat6 uses connection pooling while dsfat5 does not.  Our continuous tests were experiencing an intermittent failure where the server was attempting to shutdown the connection pool just as dsfat6 was trying to get a connection for this test.  Since this test has nothing to do with connection pooling, I figured we could avoid this intermittent failure by using a datasource that does not use connection pooling. Defect: 263440

KyleAure:
#build #libbybot

dev/com.ibm.ws.jdbc_fat/test-applications/basicfat/resources/WEB-INF/ibm-web-bnd.xml
gjwatts: Thanks and good thinking!

PR #10700, opened 2020-02-03
Issue #10697: LDAP and URBridge need to unescape escaped quotation an…

### LINKED ISSUE #10697: LDAP registry and URBridge are not un-escaping double quotation and apostrophes from the XPATH search expression
When a WIM search operation with a search expression that includes escaped apostrophes or escaped quotation marks, the LDAP registry and UR bridge (which calls UserRegistries such as Basic Registry and custom UserRegistries) are not un-escaping the escaped characters in the XPath expression. This can lead to expressions that result in incorrect search results.

For example the search expression:

`@xsi:type='Group' and (cn = "*""*")`

may result in the LDAP search filter:

`((&objectclass=groupofnames)(cn=*""*))`

and/or the UserRegistry search expression:

`*""*`

Note that these errors can only be encountered when doing searches with the 'filter' parameter for SCIM searches.

jvanhill:
…d apostrophes in the XPath search expression.

fixes #10697


jvanhill:
#build

PR #10703, opened 2020-02-03
Sleep for 600ms so that Windows boxes will report sleep was >= 500.

andymc12:
Working around a bug in Windows JDKs where Thread.sleep(500) sleeps for less than 500 milliseconds (or that System.nanoTime() called before and after the sleep reports less than 500ms elapsed).


andymc12:
#build

PR #10706, opened 2020-02-03
Issue #10704: Fix EJB timer getNextTimeout

### LINKED ISSUE #10704: EJB timer service incorrectly reports timer "created with no scheduled timeouts"
When a non-persistent timer with no future timeouts is running (so either a single action timer or a calendar based timer)... then if a concurrent thread calls Timer.getNextTimeout, the timer service incorrectly throws a NoSuchObjectLocalException indicating that the timer was created with no scheduled timeout :

javax.ejb.NoSuchObjectLocalException: Timer with ID 2_NP_0BE6CFC0-0170-4000-E000-73660950C677 was created with no scheduled timeouts.
at com.ibm.ejs.container.TimerNpImpl.getNextTimeout(TimerNpImpl.java:771)
at suite.timer.persistent.ejb.NoMoreTimeoutsBean.test(NoMoreTimeoutsBean.java:147)

The message is incorrect, as the timer did have a scheduled timeout when created.  The exception is also incorrect, in that the timer does continue to exist until the timeout method completes.  Once the timeout method completes and the transaction is committed, then the timer will cease to exist, but until that time, it still exists.

The timer service should return the currently running timer expiration when called from a concurrent thread.  If called from within the timer thread, it should throw NoMoreTimeoutsException.

tkburroughs:
Timer.getNextTimeout may return NoSuchObjectLocalException indicating that
the timer was created with no scheduled timeout if there are no future
timouts and the timer is currently running.

Update getNextTimeout so if the timer is currently running, then other
concurrent threads will see the currently scheduled timeout, since it
hasn't actually completed yet.

**Note:** Please ignore the formatting changes - only real changes are on lines 719-727 of new version.

fixes #10704

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/TimerNpImpl.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/TimerNpImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2019 IBM Corporation and others.
+ * Copyright (c) 2009, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


tkburroughs:
#build

PR #10710, opened 2020-02-03
Mp jwt java2 updates

c00crane:
Update the mp jwt FAT to run with Java2 enabled on Windows.  
Regular builds are running just fine - I'm having trouble running locally on Windows.


c00crane:
#build

PR #10718, opened 2020-02-04
FT 1.0: Disable BulkheadSyncRetryTest.testNoRetriesBulkhead

Azquelt:
Test is broken in 1.0 TCK because it uses an async bulkhead and assumes
it can immediately fill both the concurrent executions and the waiting
task queue, which isn't true for our FT 1.x bulkhead.


Azquelt:
#build

PR #10719, opened 2020-02-04
Increase timeout multiplier for FT 2.0 TCK

Azquelt:
This should prevent some of the false positives which appear
occasionally in the build.


Azquelt:
#build

Azquelt:
Looks like there's an issue in the TCK when the timeout multiplier is increased to 6 because one of the values in TimeoutUninterruptableTest isn't scaled. This will require a fix to the TCK.

PR #10726, opened 2020-02-04
initial dbtaskstore methods to create, read, and update the partition entry for polling

njr-11:
Put the basic infrastructure in place to initialize a partition entry to be used for coordination of polling.  Fit this in so that it can start computing the delay until the next poll, once implemented.  For now, the function is guarded with a hidden property to be used only by tests.

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent/src/com/ibm/wsspi/concurrent/persistent/TaskStore.java
+++ b/dev/com.ibm.ws.concurrent.persistent/src/com/ibm/wsspi/concurrent/persistent/TaskStore.java
@@ -163,10 +163,10 @@ public interface TaskStore {
      * Reads information from the partition entry for polling coordination and obtains a write lock on it.
      *
      * @param partitionId unique identifier of the partition entry for polling coordination.
-     * @return TODO specify precise type for return value and document here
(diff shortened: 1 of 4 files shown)


njr-11:
#build

dev/com.ibm.ws.concurrent.persistent/src/com/ibm/wsspi/concurrent/persistent/TaskStore.java
```diff
@@ -370,4 +388,15 @@
      * @throws Exception if an error occurs when attempting to update the persistent task store.
      */
     int transfer(Long maxTaskId, long oldPartitionId, long newPartitionId) throws Exception;
+
+    /**
+     * Reads information from the partition entry for polling coordination.
```
nmittles: Javadoc should say "Writes..." or "Updates..." instead of "Reads...".

njr-11: Good catch, I'm fixing this and also cleaning up the parameters to align with the designed solution now that there is some more clarity

njr-11:
#build
#run-libby-bot

PR #10739, opened 2020-02-05
ConnectionManagerMBean.getJndiName JavaDoc is not accurate

### LINKED ISSUE #10730: JavaDoc of ConnectionManagerMBean.getJndiName is not accurate
It appears that a getJndiName method was added to ConnectionManagerMBean, which is confusing because connectionManager doesn't have a JNDI name.  Data sources and connection factories can optionally have a JNDI name, and these can use a connectionManager.  It looks like the implementation of this method is such that the first data source or connection factory to use the connectionManager lends its JNDI name, if it has one, to the ConnectionManagerMBean for this method.

However, the method documents itself as saying it "Displays the JndiName of the connection pool associated with this Connection Manager", and also making no mention of the possibility that there might not be a JNDI name at all.

Ideally, we would get rid of the method, but that would be a breaking change.  I have opened this method to at least correct the JavaDoc.

njr-11:
pull fixes #10730


njr-11:
#build

njr-11:
#build

njr-11:
#build

Issue #10744, opened 2020-02-05
Increase the timeout in BasicSseResource to account for slow build machines


jkoehler22:
See in the logs:

BasicSseResource.sendPlainEventsFromRX: completableFuture failed because it took to long

jkoehler22:
pr https://github.com/OpenLiberty/open-liberty/pull/10746

PR #10746, opened 2020-02-05
Fix for Issue #10744: Increase the timeout in BasicSseResource to account for slow build machines

jkoehler22:


jkoehler22:
#build

PR #10749, opened 2020-02-05
Enable kafka client trace

Azquelt:
Turn on tracing for the kafka client library to try to track down the
cause of an occasional test failure.


Azquelt:
#build

PR #10778, opened 2020-02-06
Remove Informix logic

gjwatts:
There was some obsolete Informix logic remaining in the JDBC FAT test, which has now been removed.


gjwatts:
#build #libbybot

PR #10802, opened 2020-02-07
disabled test of login module loaded from the application

njr-11:
Write a test (currently disabled until the function can be added) of a login module that is packaged within a web module within an application.  For now, we have the test configured to access the login module out of a separate JAR and can switch it over later.  Also, this has some login modules packaged within EJB modules for future experimentation.


njr-11:
#build

PR #10818, opened 2020-02-10
Fix timestamp bug in ConsumerRecordTest

Azquelt:
The ConsumerRecordTest manually set the timestamp to 100 (i.e. long ago
in the past). After an indeterminate period, the server would notice
that this was before the retention period and delete it, causing the
test to intermittantly fail.

Replace this with a more realistic timestamp, set the same data into a
header and confirm that they match in the received message.


Azquelt:
#build

wraschke:
Sorry, Andrew, I had to revert. It caused a hard break (compilation failure). See https://ibm-cloud.slack.com/archives/C30NGUQ9E/p1581543519222000 for details.

PR #10824, opened 2020-02-10
disabled tests of login modules loaded from webapp and library within EAR

njr-11:
Package a login module within a web application, associate it with a data source via a resource reference, attempt to access the data source from the web application and verify that the user name is chosen by the login module.  This test is disabled for now (by also including the login module within a separate library) and can be enabled once the necessary function is added.

Package a login module within a library at the top level within an EAR, associate it with a data source via a resource reference, attempt to access the data source from the web module and EJB modules and verify that the user name is chosen by the login module.  This test is disabled for now (by also including the login module within a separate library) and can be enabled once the necessary function is added.


njr-11:
#build

PR #10863, opened 2020-02-12
tie service account cache entry to config using it

c00crane:
Include the config id in the cache id when caching the service account.  Multiple configs could be using the same service account.  They may each specify different cache lifetimes - which one would be used if there is only one service account cache entry???


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat.LibertyOP

c00crane:
Had to merge in revert that was causing build failure.
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat.LibertyOP

c00crane:
#run-libby-bot

Issue #10821, opened 2020-02-10
Update message CWWKS5382E for serviceability


clarkek123:
The new message below has been added for the OpenShift service account token UFO for epic 10022.

ERROR_INTROSPECTING_SERVICE_ACCOUNT=CWWKS5382E: The response from the service account user API cannot be processed: {0}
ERROR_INTROSPECTING_SERVICE_ACCOUNT.explanation=Information about the service account cannot be obtained because another error occurred.
ERROR_INTROSPECTING_SERVICE_ACCOUNT.useraction=If the message contains another error message, see the user action for that message. Otherwise, check the server logs for more error messages that might indicate where the other error occurred.

The message is misleading because it refers to the "service account token user API" when this is just the "user API" endpoint in OpenShift.  I recommend re-wording this message as: 

 The response from the user API cannot be processed because of the error: {0}

PR #9539, opened 2019-10-29
OpenShift login: Initial NLS message drop

ayoho:
Initial drop of NLS messages for the `openshiftLogin` work being done in the social login feature.

PR #10864, opened 2020-02-12
small update to the new social service account handling error message

arunavemulapalli:
message update

### DIFF
--- a/dev/com.ibm.ws.security.social/resources/com/ibm/ws/security/social/resources/SocialMessages.nlsprops
+++ b/dev/com.ibm.ws.security.social/resources/com/ibm/ws/security/social/resources/SocialMessages.nlsprops
@@ -95,7 +95,7 @@ KUBERNETES_USER_API_RESPONSE_DEFAULT_USER_ATTR_NOT_FOUND.explanation=The user AP
 KUBERNETES_USER_API_RESPONSE_DEFAULT_USER_ATTR_NOT_FOUND.useraction=Verify that the configuration attribute that is specified in the message is configured correctly. Ensure that the value of the configuration attribute corresponds to the entry in the user API response that should be used as the username.
 
 # 0=Exception message
-ERROR_INTROSPECTING_SERVICE_ACCOUNT=CWWKS5382E: The response from the service account user API cannot be processed: {0}
+ERROR_INTROSPECTING_SERVICE_ACCOUNT=CWWKS5382E: The response from the user API cannot be processed: {0}
 ERROR_INTROSPECTING_SERVICE_ACCOUNT.explanation=Information about the service account cannot be obtained because another error occurred.
 ERROR_INTROSPECTING_SERVICE_ACCOUNT.useraction=If the message contains another error message, see the user action for that message. Otherwise, check the server logs for more error messages that might indicate where the other error occurred.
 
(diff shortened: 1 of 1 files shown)


arunavemulapalli:
#run-libby-bot

arunavemulapalli:
#build

arunavemulapalli:
fixes #10821

arunavemulapalli:
#build

dev/com.ibm.ws.security.social/resources/com/ibm/ws/security/social/resources/SocialMessages.nlsprops
```diff
@@ -95,7 +95,7 @@ KUBERNETES_USER_API_RESPONSE_DEFAULT_USER_ATTR_NOT_FOUND.explanation=The user AP
 KUBERNETES_USER_API_RESPONSE_DEFAULT_USER_ATTR_NOT_FOUND.useraction=Verify that the configuration attribute that is specified in the message is configured correctly. Ensure that the value of the configuration attribute corresponds to the entry in the user API response that should be used as the username.
 
 # 0=Exception message
-ERROR_INTROSPECTING_SERVICE_ACCOUNT=CWWKS5382E: The response from the service account user API cannot be processed: {0}
+ERROR_INTROSPECTING_SERVICE_ACCOUNT=CWWKS5382E: The response from the user API cannot be processed: {0}
```
dmuelle: Per IBM style, variable should precede descriptor

In this message, does {0} represent the response itself or the API?

skasund:
L2 message review completed

dev/com.ibm.ws.security.social/resources/com/ibm/ws/security/social/resources/SocialMessages.nlsprops
arunavemulapalli: response

dmuelle: per IBM style, it should read

The {0} response from the user API cannot be processed.

However, if the API response is expected to be a long string or similar, making the message illegible, it can be

The following response from the user API cannot be processed: {0}

arunavemulapalli: response can be very long and some times with an error exception included. I think it is better to do "The following response ........"

arunavemulapalli:
#run-libby-bot

ayoho:
FYI @dmuelle and @arunavemulapalli : The insert in that NLS message is another exception/error message; it's not necessarily the API response itself. So saying "the **following** response from the user API" isn't quite accurate. I structured the message that way based partly on https://github.com/OpenLiberty/open-liberty/pull/9539#discussion_r340839555. It's very possible that the insert could be a long string, such as another NLS message, that would make the message illegible as you said.

For example, it's entirely possible that an error will occur such that the insert in the CWWKS5371E message would be the CWWKS5381W message. The CWWKS5381W message is pretty long and isn't, nor does it include, the response from the user API. With the current message form, that would look something like:

> CWWKS5382E: The following response from the user API cannot be processed: CWWKS5381W: The social login configuration [myConfigId] specifies [someKey] as the value for the [userNameAttribute] attribute, but the user API response does not contain a [someKey] key. The [username] key, if present in the user API response, is used to determine the username instead.

Like I said, the insert isn't actually coming from the user API, so saying "the following response from the user API" is misleading.

dmuelle:
@ayoho - thanks for the context. Even without the "following", the proposed message update appears to indicate that the response is from the user API

`The response from the user API cannot be processed: {0}`

Should the "service account" be restored to the message? I assume that indicates a degree of remove that could allow for more variation in the type of content that might show up in the insert?

ayoho:
@dmuelle I think taking out "service account" was still the right move since technically there is no "service account user API" - there's simply a "user API" that can return information about service accounts as well as users. You're right, though, that the message could be emitted but not necessarily caused by an error in the response. An error also could have occurred before submitting or while submitting the request to the endpoint. I suppose a more accurate phrasing would be along the lines of

`The request to the user API cannot be sent, or the response from the user API cannot be processed: {0}`

dmuelle:
@ayoho i think that gives more complete information about a possible error- my only question is if the user action would be the same in either case. If the problem was that the request couldnot be sent, would a user still take the same steps?

`If the message contains another error message, see the user action for that message. Otherwise, check the server logs for more error messages that might indicate where the other error occurred.`

ezhou365:
Fixes #10821

PR #10865, opened 2020-02-12
load login module from standalone resource adapter

njr-11:
Code updates to load a login module from resource adapter that is specified by a classProviderRef.  Declarative services isn't used to inject the service due to a lot of unfortunate eager initialization in the security code which would require the resource adapter to be available far earlier on than is possible.  Instead, we lazily locate it a later point where it is guaranteed that the resource adapter is already available (per spec, resource adapters must be started before applications).


njr-11:
#build

njr-11:
Build failed in a unit test that I wasn't aware of which had an implementation of the metatype config class. A pull has been added to update accordingly.
#build
#run-libby-bot

njr-11:
The previous build appears to have ignored the second commit with the unit test fix.  Although the original changes are still under discussion with security team, I'm going to try the build again, just to see if the unit test fix can get picked up properly.  Maybe I submitted the build too soon previously?
#build

njr-11:
Security team preferred that we not take the approach used in this pull.
Instead, we plan to use the approach under https://github.com/OpenLiberty/open-liberty/pull/10908

PR #10895, opened 2020-02-13
Fix Eclipse building error

wraschke:
#build


wraschke:
#run-libby-bot

wraschke:
Defects 271651 and 264572 are already existing for the unittest and FAT failure that occurred in the PB mentioned above. This change has no impact on those tests and the functions they test.

Issue #10762, opened 2020-02-06
Missing warning when a server element is not present


brenthdaniel:
Normally, config processing will throw an error when server.xml or an include file does not contain a `<server>` or `<client>` element at the root. However, this will not be the case if the file being parsed only contains a single element.

PR #10897, opened 2020-02-14
Report an error when an included file doesn't have server or client a…

brenthdaniel:
PR for #10762

### DIFF
--- a/dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
+++ b/dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
@@ -469,12 +469,12 @@ warning.supplied.config.not.valid.explanation= A configuration attribute marked 
 warning.supplied.config.not.valid.useraction=Update the server configuration to remove the incorrect attribute value.
 
 variables.in.id.not.supported=CWWKG0104W: The ID attribute with value {1} for the configuration element {0} must have a fixed value. Variables in the ID attribute will not be replaced.
-variables.in.id.not.supported.explanation: Variable values are not allowed in ID attributes. All variables in ID attributes will be treated as literal strings.  
-variables.in.id.not.supported.useraction: Remove the variable from the value of the ID attribute.
+variables.in.id.not.supported.explanation=Variable values are not allowed in ID attributes. All variables in ID attributes will be treated as literal strings.  
+variables.in.id.not.supported.useraction=Remove the variable from the value of the ID attribute.
 
 error.root.must.be.server=CWWKG0105E: The {0} configuration document must begin with the [{1}] element.
 error.root.must.be.server.explanation=The configuration must be contained in either a server element or a client element. 
-error.root.must.be.server.useraction: Add a server or client element as the root of the configuration file. 
+error.root.must.be.server.useraction=Add a server or client element as the root of the configuration file. 
 
 # Non-TR message
 error.invalidOCDRef=ERROR: Metatype PID [{0}] specifies non-existent object class definition ID [{1}]
--- a/dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
+++ b/dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
@@ -472,8 +472,8 @@ variables.in.id.not.supported=CWWKG0104W: The ID attribute with value {1} for th
 variables.in.id.not.supported.explanation: Variable values are not allowed in ID attributes. All variables in ID attributes will be treated as literal strings.  
 variables.in.id.not.supported.useraction: Remove the variable from the value of the ID attribute.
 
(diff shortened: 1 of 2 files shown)


brenthdaniel:
#build

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
```diff
@@ -472,5 +472,9 @@ variables.in.id.not.supported=CWWKG0104W: The ID attribute with value {1} for th
 variables.in.id.not.supported.explanation: Variable values are not allowed in ID attributes. All variables in ID attributes will be treated as literal strings.  
 variables.in.id.not.supported.useraction: Remove the variable from the value of the ID attribute.
 
+error.root.must.be.server=CWWKG0105E: The configuration document {0} must begin with the element '{1}'.
```
dmuelle: **change**
The configuration document {0} must begin with the element '{1}'.
**to**
The {0} configuration document must begin with the '{1}' element.

why is one offset with brackets and straight  quote? '{1}' Are both needed?

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
```diff
@@ -472,5 +472,9 @@ variables.in.id.not.supported=CWWKG0104W: The ID attribute with value {1} for th
...
+error.root.must.be.server.explanation=Configuration must be contained in either a server element or a client element.
```
dmuelle: how will the user know whether a client or server element is required?

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
brenthdaniel: The only reason for the quote on {1} is that without it the phrase sounds like a concept. It's better with the phrasing you suggested, but with either "server element" or "element server" it sounds like a particular server or type of server. I can remove it if needed, though.

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
brenthdaniel: It's in the error message ( the {1} substitution )

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
dmuelle: Ok- I see your point. The concern is that using quotes can flag an accessibility violation in some sources. I'm not sure if that would apply in this case but it's something we watch out for. I don't know where the output is expected but in a markdown source backtics (`) around `server` could do the job. [Brackets] are also a possibility.

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
dmuelle: ok- if server or client is directly specified in the variable I think this is ok. Add "the"

The configuration must be contained in either a server element or a client element.

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
brenthdaniel: We use brackets elsewhere for config messages, so I'll go with that.

dev/com.ibm.ws.config/resources/com/ibm/ws/config/internal/resources/ConfigMessages.nlsprops
```diff
@@ -472,5 +472,9 @@ variables.in.id.not.supported=CWWKG0104W: The ID attribute with value {1} for th
 variables.in.id.not.supported.explanation: Variable values are not allowed in ID attributes. All variables in ID attributes will be treated as literal strings.  
 variables.in.id.not.supported.useraction: Remove the variable from the value of the ID attribute.
 
+error.root.must.be.server=CWWKG0105E: The {0} configuration document must begin with the [{1}] element.
+error.root.must.be.server.explanation=The configuration must be contained in either a server element or a client element.
```
dmuelle: should this message have a colon (:) like the other two or is it supposed to have the (=) ?

brenthdaniel: I'm surprised the build didn't fail for that.. but it looks like the previous message ID in the file has the same issue. I fixed both of them.

brenthdaniel:
#build

brenthdaniel:
#run-libby-bot

PR #10902, opened 2020-02-14
Add test variations for header("Expect", "100-continue")

jim-krueger:


jim-krueger:
#build

jim-krueger:
@andymc12 sure

dev/com.ibm.ws.jaxrs.2.x_fat_clientProps/test-applications/cxfClientPropsApp/src/jaxrs2x/cxfClientProps/CxfClientPropsTestServlet.java
```diff
@@ -253,6 +253,37 @@ public void testChunkingThreshold(HttpServletRequest req, HttpServletResponse re
...
+                        .property("client.ChunkingThreshold", "40000")
+                        .build();
+        response = client.target("http://localhost:" + req.getServerPort() + "/cxfClientPropsApp/resource/chunking")
+                       .request().header("Expect", "100-continue")
+                       .post(Entity.text(sb.toString()));
+        status = response.getStatus();
+        result = response.readEntity(String.class);
+
+        // If a 100 response is received then no data will be sent.
+        assertEquals(100,status);
```
andymc12: This does not seem right to me.  It is certainly inconsistent.  I don't think that we want a test case to verify that we behave incorrectly/inconsistently.  Instead, we should probably work with the CXF community to determine the right next step - possibly something like always enabling chunking when the `Expect: 100-continue` header is specified.

jim-krueger:
#build

jim-krueger:
#run-liberty-bot

dev/com.ibm.ws.jaxrs.2.x_fat_clientProps/test-applications/cxfClientPropsApp/src/jaxrs2x/cxfClientProps/CxfClientPropsTestServlet.java
jim-krueger: I agree in principle.   However, this is our behavior now so I think I will leave it.   However I agree this should be addressed so I'll create an issue to work with CXF to resolve, indicating that this testcase will need to be adjusted when that happens.

jim-krueger: And it will be interesting to see if this behavior is the same for all platforms/jdks.

jim-krueger:
#run-libby-bot

PR #10908, opened 2020-02-14
use proxy to redirect from app pid to class provider

njr-11:
This pull covers the approach of enforcing a declarative services dependency on the appmanager pid that is specified as the classProviderRef.  However, because the app instance becomes available in the service registry long before the corresponding resource adapter and its class loader, if we are to use this approach, we need an intermediate proxy that becomes registered in response to the real ClassProvider becoming available and which redirects to it.


njr-11:
#build
#run-libby-bot

njr-11:
#build
#run-libby-bot

njr-11:
#build
#run-libby-bot

njr-11:
Missed an implicit dependency on an app manager package from the bnd file.  The commit is updated with an exclude for it. Retrying the build:
#build
#run-libby-bot

njr-11:
#build

njr-11:
build has 1 test failure, which is a duplicate of a preexisting issue

PR #10932, opened 2020-02-17
Update minimum config jwt builder tests

c00crane:
The Minimum Config test class in the JWT Builder fat will start off with almost nothing configured.  For each test it would add and remove the test app, the builder configs and SSL.  Consistently trying to get all of those changes add/removed/added/removed/... is not feasible.  I've split the tests into 4 individual test class - each will start the server with the config that it needs.  Hopefully, we can ensure that everything that each test need is ready before we run the test.  (at a minimum, it'll be easier to debug)


c00crane:
#build

c00crane:
z/OS build was good too:  https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_csQYEFHREeqNE65GG-cLhw

PR #10953, opened 2020-02-19
load login module from embedded RAR

njr-11:
In order to make it possible to access the class loader from the application which contains the embedded RAR, and to do so via an OSGi service component which must be active if and only if the application that provides the embedded RAR (and the login module classes within it) is loaded, I have added a new service component registration (of existing ApplicationInfo) to the application management code, around the same places where ApplicationStateListener is notified.  ApplicationStateListener is itself not a solution here because it only has awareness of application life cycle AFTER the registration of the ApplicationStateListener is made, with no awareness of the applications that might have started before that point.  Also, it will be noted from the code under this pull, that the addition of ApplicationInfo as a service component, allows us to get rid of an ugly solution of intercepting and re-registering proxy services for the standalone resource adapter path.  The new solution fits seamlessly into both standalone and embedded RAR paths, and should similarly be extendable to later implement the additional support from loading from a non-resource adapter application library.


njr-11:
#build

njr-11:
#build
#run-libby-bot

njr-11:
Build has 2 test failures, which are duplicates of preexisting issues

Issue #10945, opened 2020-02-18
Proxy encryption test fix


herman-kailey:
fix for the proxy encryption test failures

PR #10966, opened 2020-02-19
changing log levels

herman-kailey:
changing log levels for message regarding missing proxy password encryption


herman-kailey:
fixes #10945

Issue #10563, opened 2020-01-23
Improve EJB missed persistent timer action (catch-up) when failover enabled.


tkburroughs:
Need to improve the EJB missed persistent timer action (catch-up) behavior when persistent timer failover support is enabled.

Currently, all missed persistent timer expirations that occur when the server is down (or application is not running) are performed immediately when the server (or application) starts.  This can be problematic when a timer would expire many times when the server is down.  For example, if a persistent EJB timer is scheduled to expire every 30 seconds, and the server is down for 20 minutes, then 40 expirations will be missed during the 20 minutes and so the timer will run 40 times, immediately, on server start to catch up on all the missed expirations.  The problem is compounded when there are more than one timer.  Just 5 similar timers would result in 200 expirations to run immediately on server start.

With the addition of persistent timer failover support, the problem of handling missed timer expirations can be propagated from one server to another.  For example, if there are 5 persistent timers, each set to expire every 30 seconds.... then if the server running them is having difficulty keeping up or goes down, after 5 minutes (configurable) those timers may then failover to a different server.  During those 5 minutes, each timer would miss 10 expirations, so once failover occurs, there would be 5 timers x 10 expirations = 50 timer expirations that would run immediately on the failover server.

This behavior is generally not expected by the majority of EJB application developers.  The EJB specification only requires that a missed persistent timer run exactly once on server start, and then may resume with the next timeout in the future; skipping all other missed expirations that occurred in the past.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Improved Missed Persistent Timer Action Behavior
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
In order to preserve the existing behavior, but enable better support for the majority of applications, a new configuration option will be provided, which will default to the new behavior when persistent timer failover is enabled.  The new missed persistent timer action and configuration option must support the EJB specification requirements, as described in EJB 3.2, spection 13.4.3 :

_In the event of a container crash or container shutdown, any single-event persistent timers that have expired during the intervening time before container restart must cause the corresponding timeout callback method to be invoked upon restart. Any interval persistent timers or schedule based persistent timers that have expired during the intervening time must cause the corresponding timeout callback method to be invoked at least once upon restart._

Specifically, the new configuration option will be defined as follows:

```
        <AD id="missedPersistentTimerAction" type="String" ibm:beta="true" required="false"
[...]

PR #10970, opened 2020-02-19
Issue #10563: Improve missedPersistentTimerAction behavior

tkburroughs:
Removed the "NONE" options, since it is only a slight variation of "ONCE"
add does not work well when failover is enabled. Could also result in a timer never running.

Added additional testing that covers scenarios where the server is stopped and
restarted.

for #10563


tkburroughs:
Part of a beta feature ID was removed, so there is no ID to review.

tkburroughs:
#build

PR #10976, opened 2020-02-20
enable tests of loading login module from top level jar of app

njr-11:
Enable tests that load a login module class from a top level JAR within a specified application.
These test verify that a data source which relies on the login module can be used from within the application's web module and EJB modules.  Also add a test to verify that the login module can be used by a data source within a completely different application.  For now, the tests will enforce that login modules are not loaded from other locations within the app (besides resource adapters which are tested elsewhere).  We can update those tests if later adding that function.

Also, this enables some additional tests for login module in a standalone RAR that were already working.


njr-11:
#build

PR #11033, opened 2020-02-24
make oauth scope optional, to match reality

brutif:
remove config checking that oauth scope must be specified, many providers have a default no matter what the client sends them


brutif:
#run-libby-bot
#build

rbtcwt:
pb is good. 

 I see if scopes are missing, error message or not, we put  an empty scope parameter in the url to the authorization endpoint.  I thought that might be a bug, but we actually have code to do that (always emit that param, even if scope is omitted).  I tested github with no scope and it works, despite the empty param in the url. We just don't need this error message since scopes aren't always required.

Issue #11018, opened 2020-02-24
clientManagement page buttons no longer clickable with newer browsers.


brutif:
The Oauth Client management ui (https://9.41.109.202:29443/oidc/endpoint/OP/clientManagement) contains radio buttons to select confiuration options (token endpoint authentication method, regular expression redirects, etc.)

These continue to work with Microsoft Edge,  but on Mac Mojave with firefox 73.0.1, they're not selectable. With chrome 80.0.3987.116, clicking on the button causes the entire dialog box to go blank.

The buttons should continue to be selectable.

brutif:
resolved by #11063

PR #11063, opened 2020-02-26
Allow radio buttons to behave correctly

dmorgant:
Radio button on the OIDC client admin's **Register New OAUTH Client** dialog appeared to not be selected when clicked.

Also, in chrome, selecting one of the radio buttons would cause the dialog to go blank.

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.server/resources/WEB-CONTENT/clientAdmin/js/clientInputDialog.js
+++ b/dev/com.ibm.ws.security.openidconnect.server/resources/WEB-CONTENT/clientAdmin/js/clientInputDialog.js
@@ -359,7 +359,7 @@ var clientInputDialog = (function() {
                             return;
                         }
                     } 
-                    // Something else happed with the request.  Put up an error message...
+                    // Something else happened with the request.  Put up an error message...
                     var errTitle = messages.GENERIC_UPDATE_FAIL;
                     var errDescription = "";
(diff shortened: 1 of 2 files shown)


dmorgant:
One DAP revision requested that \<input> elements be contained within their associated \<label>s.  However, this broke the OIDC radio button selection styling defined in ..\commom\tool.css.  Therefore, while the radio button selection was occurring, visually you could not see that the button was being selected.   To fix, I switched the \<input> element back to its original position above ( so it is now a sibling) of its associated \<label> element.  Now when you select the radio button you can visually see that it is selected.

The Chrome issue where the dialog box went blank when a radio button was selected was fixed by removing `overflow: hidden` from the` .tool_modal` styling, which is the overall styling of the dialog's outermost \<div>

dmorgant:
fixes #11018

dmorgant:
#build

dmorgant:
To test:
I logged into the OIDC Client Administration application and created a new client in multiple browsers (FF, Chrome, and Safari) to see that the default values for the radio button fields were visibly selected.   Then I changed the values and selected 'Register' to create a new client.  Then I edited the client I just created and verified that the values I selected were correctly shown as selected.  I changed the values in the edit dialog and when I re-edited the entry the new values were shown as selected.

In chrome I could easily verify that selecting the values did not cause the dialog to go blank.

dmorgant:
#build

dmorgant:
#build

Issue #10988, opened 2020-02-20
SRVE8115W and SRVE8094W warnings at every request to an Krazo application


mthmulders:
**Describe the bug**
A sample application built with [Krazo](https://github.com/eclipse-ee4j/krazo) logs SRVE8115W and SRVE8094W warnings at every request. Krazo is built on top of JAX-RS.

Stacktrace:

    [INFO] [WARNING ] SRVE8115W: WARNING: Cannot set status. Response already committed.
    [INFO] [WARNING ] SRVE8094W: WARNING: Cannot set header. Response already committed.  Stack trace of errant attempt to set header: 
    [INFO]         at com.ibm.ws.webcontainer.srt.SRTServletResponse.setHeader(SRTServletResponse.java:1759)
    [INFO]         at [internal classes].(Unknown Source)
    [INFO]         at org.eclipse.krazo.core.ViewableWriter$DelegatingServletOutputStream.write(ViewableWriter.java:248)
    [INFO]         at java.base/java.io.OutputStream.write(OutputStream.java:157)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:233)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.implFlushBuffer(StreamEncoder.java:312)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.implFlush(StreamEncoder.java:316)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.flush(StreamEncoder.java:153)
    [INFO]         at java.base/java.io.OutputStreamWriter.flush(OutputStreamWriter.java:254)
    [INFO]         at java.base/java.io.PrintWriter.flush(PrintWriter.java:396)
    [INFO]         at org.eclipse.krazo.core.ViewableWriter.writeTo(ViewableWriter.java:169)
    [INFO]         at org.eclipse.krazo.core.ViewableWriter.writeTo(ViewableWriter.java:70)
    [INFO]         at org.apache.cxf.jaxrs.utils.JAXRSUtils$3.run(JAXRSUtils.java:1479)
    [INFO]         at [internal classes].(Unknown Source)

**Steps to Reproduce**
When I deploy a [sample Krazo application](https://github.com/mthmulders/openliberty-mvc) to OpenLiberty and I access it through my web browser, the OpenLiberty logs shows these messages. The sample app uses the Liberty Maven plugin so `mvn liberty:dev` should be enough to get started.

[...]

PR #11070, opened 2020-02-26
Avoid setting status/headers when response is already committed

andymc12:
Test fix for issue #10988 - many thanks to @wtlucy for suggesting this change.  This should avoid calling setStatus or copying headers to the response output stream when the response has already been committed.  This should prevent warning errors like those mentioned in issue #10988.


andymc12:
#build

andymc12:
Build failed due to infrastructure issues... new #build

PR #11087, opened 2020-02-27
Fix hide password logic for connProps

KyleAure:
Updated hide password logic, was accidentally checking the same property twice.  
Also added in the driver type to generic properties when connecting to oracle.  This was causing errors during our DB rotation testing.


KyleAure:
#build 
#libbybot

PR #11100, opened 2020-02-27
disabled HandleList tests for unshared JDBC connections

njr-11:
Add some disabled test (because they don't currently work) that cover some potential scenarios where HandleList might be needed for unshared JDBC connections.


njr-11:
#build

PR #11148, opened 2020-03-02
load login module from web module

njr-11:
This pull adds the ability to load JAAS custom login modules from a web module within an enterprise application.  Prior to this, it was already possible to load from the EJB modules, RAR modules, and libraries within an EAR.


njr-11:
#build

njr-11:
Build has 4 failures, which are duplicates of pre-existing issues.

PR #11155, opened 2020-03-03
Resist incorrect timer failure

### LINKED ISSUE #11153: Test defect persistent timers demo
A test defect has been found during a build where the AutomaticDatabase timer in com.ibm.ws.concurrent.persistent_fat_demo_timers had it's `@PostConstruct` method called a second time.  Likely indicating there may have been a race condition issue, or was prematurely destroyed.  As a stateless bean (and not singleton) this behavior would be acceptable?

To increase the rigidity of this test we should actually look at the database to see how many times the timer has run.

KyleAure:
fixes #11153


KyleAure:
#build

PR #11161, opened 2020-03-03
test unshared autocommit connection across servlet requests

njr-11:
Add a test case that obtains an unshared connection in one servlet method and uses it, but does not close it, instead allowing it to continue to execution operations asynchronously via a completion stage. After the servlet method ends and goes out of scope, invokes a second servlet method, which under a new scope, accesses the completion stage, joins it, and checks that the operations executed successfully.  This is likely only possible due to the lack of the HandleList implementation.


njr-11:
#build

njr-11:
#run-libby-bot

njr-11:
#build

PR #11163, opened 2020-03-03
manually revert PR 10967

gjdeval:
In PR 10967 we introduced a threshold for hang resolution activity in the thread pool controller autonomic sizing algorithm, to avoid taking hang resolution actions when the system CPU is busy or the java process is doing work. 
Testing has convinced us that this threshold is not readily generalizable, or at least we have not yet figured out how to do that, so we are going to remove it for this release.
This PR backs out all of PR 10967 except a change in the default value for the hang resolution interval, from 250 ms to 750 ms.


gjdeval:
#build

gjdeval:
#run-libby-bot

PR #11167, opened 2020-03-03
Issue #11044: Fix res-ref config not honored

### LINKED ISSUE #11044: custom-login-configuration not honored in java:comp/env bindings without binding-name
Bindings with a java:comp/env  explicitly specified in the name, and without a binding-name specified in the bindings file (it's sent in `@Resource` instead), do not other parts of the bindings honored.

To work around this, a test case had to use jdbc/dsref as the binding name instead of java:comp/env/jdbc/dsref,
https://github.com/OpenLiberty/open-liberty/pull/11015/commits/832eff3e7fd426599556b182b2aad5153feef5a6

An easy way to test this once is is working again would be to switch the bindings in the above test from jdbc/dsref back to java:comp/env/jdbc/dsref,

tkburroughs:
When a jndi-binding-name is present, the resource processor loops through
looking for bindings using both the normalized (short) and denormalized (full)
names (i.e. jdbc/myds & java:comp/env/jdbc/myds). Once found, the additional
res-ref config is located matching that name.

However, when no binding, res-ref config is only located using the normalized (short)
form of the resource ref jndi name.

Fix is added to also look using the denormalized (full) name for two scenarios:
1 - a lookup name was specifiec either on the annotation or xml
2 - no lookup; so using auto-link where DS name is exact match of res-ref short name

Code is added after looking for bindings, to ensure we don't combine bindings
found in the *-bnd.xml file using both short and long forms. Should use only
short form if located, then only long form.

fixes #11044

### DIFF
--- a/dev/com.ibm.ws.injection.core/src/com/ibm/ws/injectionengine/processor/ResourceProcessor.java
+++ b/dev/com.ibm.ws.injection.core/src/com/ibm/ws/injectionengine/processor/ResourceProcessor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2016 IBM Corporation and others.
+ * Copyright (c) 2006, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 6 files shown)


tkburroughs:
#build

tkburroughs:
#build

tkburroughs:
All failing FAT have been linked to open issues.

PR #11168, opened 2020-03-03
fix issue identified during serviceability review

njr-11:
During the serviceability review, it was discovered that a message update requested under
https://github.com/OpenLiberty/open-liberty/pull/10913#discussion_r380380705
had been improperly made.  This pull corrects the word order.


njr-11:
ID review already happened under 10913, so another is not needed.  The only purpose of this pull is to fix the wording to match what the ID reviewer previously requested.

PR #11176, opened 2020-03-04
Disable Quarantined test testHTTP2HTTPSRedirect()

bensonlatibm:
Test case depends on maven.org to redirect from http to https.  Since Maven has changed to only support https, we can no longer use maven to test http redirects.  Disabling this test.

Will use https://github.com/OpenLiberty/open-liberty/issues/11183 to implement this test case properly.


bensonlatibm:
#build

PR #11204, opened 2020-03-05
Part 1 of split package issue with com.ibm.ejs.j2c: stop importing from cm bundle

njr-11:
Minor refactoring so that the com.ibm.ws.jca.cm bundle can stop importing classes defined elsewhere from the same com.ibm.ejs.j2c package that it declares as a private package for the connection manager implementation.

It should be noted that the cm bundle cannot stop using the com.ibm.ejs.j2c package for its implementation because at least one of its classes in that package is serialized to the transaction recovery logs and cannot be renamed.

This pull is part 1 to at least clear up the issue for the cm bundle.
A second part 2 can be done after this to refactor more of the Liberty code and remove its com.ibm.ejs.j2c package classes from the compat bundle entirely, after which only the cm bundle will define classes in that package.


njr-11:
#build

njr-11:
#build

njr-11:
#run-libby-bot

njr-11:
#build

njr-11:
Build has 2 test failures, which are unrelated preexisting issues.

PR #11221, opened 2020-03-06
test is missing Async and accidentally runs inline

njr-11:
A test that I recently wrote accidentally uses thenApply where it meant to use thenApplyAsync, causing incorrect test logic which made it look like a local transaction remains open across servlet boundaries.  After fixing the test logic, the test needs some updates elsewhere per the expected behavior of having the local transaction rolled back per the default unresolved action.  This test was written to investigate what behavior we are getting absent a HandleList implementation, and is expected to be updated once HandleList is in place.


njr-11:
#build

njr-11:
#build

PR #11251, opened 2020-03-10
Ejb custom binding namespace error bug fix

olendvcook:
Originally for a build break caused by the server installing every application at startup because a previous run had added them to dropins, and even though they were "removed" from dropins, they were still in the list of installed applications and being cached in the server after it stopped.

I then noticed a bug in my namespace warning code. some of my message printouts had null for the binding-value, so I fixed that too, and updated relevant tests to catch it.


olendvcook:
#build

olendvcook:
#build

olendvcook:
#run-libby-bot

unrelated test failure in build

Issue #11317, opened 2020-03-13
JAXRS requests can hang when Expect header set to "100-continue".


jim-krueger:
When the Expect header on a client connection is set to "100-continue" a 100 response is initially sent from the server.   When streaming is enabled the JDK will process the 100 so only the following 200 response will be sent to the JAXRS client.   However, when streaming is not involved the JAXRS client will receive the 100, which it is not expecting, and may error out or hang.  Chunking and FixedLengthStreamingMode are the two streaming options, however fixedLengthSteaming is disabled in CXF due to a JDK performance hit (https://bugs.openjdk.java.net/browse/JDK-8044726), so when the payload size is beneath the chunking threshold a failure will occur.

jim-krueger:
This issue is resolved by setting the client.ChunkingThreshold to "0" whenever the Expect header is set to "100-continue".  Due to backward compatibility issues the JDK team is unwilling to address the performance issue with fixedLengthStreeaming and alternatives to address this within CXF/JAXRS code also hold performance impacts.   Setting the ChunkingThreshold is the most viable alternative at this time.

PR #11318, opened 2020-03-13
Enable FixedLengthStreaming when Expect header set to 100-continue

jim-krueger:
#11317

### DIFF
--- a/dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
+++ b/dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
@@ -33,6 +33,7 @@ import java.net.URLConnection;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
@@ -69,6 +70,9 @@ public class URLConnectionHTTPConduit extends HTTPConduit {
         DEFAULT_USE_REFLECTION = Boolean.valueOf(SystemPropertyAction.getProperty(HTTPURL_CONNECTION_METHOD_REFLECTION, "false"));
         SET_REASON_PHRASE = Boolean.valueOf(SystemPropertyAction.getProperty(SET_REASON_PHRASE_NOT_NULL, "false"));
     }
+    
+    private Boolean allowFixedLengthStreamMode = null;
+    private Message message = null;
 
     /**
      * This field holds the connection factory, which primarily is used to
@@ -129,7 +133,10 @@ public class URLConnectionHTTPConduit extends HTTPConduit {
     @FFDCIgnore(java.net.ProtocolException.class)
     @Override
     protected void setupConnection(Message message, Address address, HTTPClientPolicy csPolicy) throws IOException {
+        this.message = message;
+        
         HttpURLConnection connection = createConnection(message, address, csPolicy);
+        
         connection.setDoOutput(true);
 
         int ctimeout = determineConnectionTimeout(message, csPolicy);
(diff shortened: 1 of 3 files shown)


jim-krueger:
#build

dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
```diff
@@ -130,6 +132,20 @@ private HttpURLConnection createConnection(Message message, Address address, HTT
...
+        
+        // If the Expect header is set to "100-continue" then currently the JDK
+        // will not handle the returned 100 and will pass that message on the JAXRS (that is expecting
+        // the 200, not the 100).   This causes issues and potentially hangs.   To work around this
+        // issue we will allow URLConnectionHttpConduit.setFixedLengthStreamingMode() to be used.  
+        // Normally FixedLengthStreamMode is disabled due to a JDK performance issue [CXF-6227].
+        Map<String, List<String>> headers = CastUtils.cast((Map<?, ?>)message.get(Message.PROTOCOL_HEADERS));
+        List<String> expectList = headers.get("Expect");
+        
+        if ((expectList != null) && (expectList.contains("100-continue"))) {
```
andymc12: According to the HTTP spec, the value of the `Expect` header is case-insensitive.  I think you'll need to iterate over the header values and check for for `value.equalsIgnoreCase("100-continue")`.

andymc12: Also, this seems absurd to me...  we're already handicapping CXF because of a JDK performance issue - so now we're making performance worse, on _every request_ to workaround another JDK bug.  The JDK needs to fix this, not us.

jim-krueger: @andymc12  I agree,  however switching the JDK may not be a viable solution since it would require us to add potentially performance impacting code to detect the JDK and do something different depending.

jim-krueger:
#run-libby-bot

jim-krueger:
#build

dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
```diff
@@ -463,6 +470,30 @@ protected boolean usingProxy() {
         protected void setFixedLengthStreamingMode(int i) {
             // [CXF-6227] do not call connection.setFixedLengthStreamingMode(i)
             // to prevent https://bugs.openjdk.java.net/browse/JDK-8044726
+            
+            //Allow this mode to be set only if the Expect header has been set to "100-continue"           
+            if ((message != null) && (allowFixedLengthStreamMode == null)) {
```
andymc12: if you use `outMessage` instead of `message` then you will get the Message instance associated with this output stream.

dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
```diff
@@ -69,6 +70,9 @@
         DEFAULT_USE_REFLECTION = Boolean.valueOf(SystemPropertyAction.getProperty(HTTPURL_CONNECTION_METHOD_REFLECTION, "false"));
         SET_REASON_PHRASE = Boolean.valueOf(SystemPropertyAction.getProperty(SET_REASON_PHRASE_NOT_NULL, "false"));
     }
+    
+    private Boolean allowFixedLengthStreamMode = null;
+    private Message message = null;
```
andymc12: This seems unsafe.  What if you have multiple messages go out on the same conduit - iiuc, the more recent will overwrite the less recent.

dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
```diff
@@ -463,6 +470,30 @@ protected boolean usingProxy() {
...
             // to prevent https://bugs.openjdk.java.net/browse/JDK-8044726
+            
+            //Allow this mode to be set only if the Expect header has been set to "100-continue"           
+            if ((message != null) && (allowFixedLengthStreamMode == null)) {
+                // If the Expect header is set to "100-continue" then currently the JDK
+                // will not handle the returned 100 and will pass that message on the JAXRS (that is expecting
+                // the 200, not the 100).   This causes issues and potentially hangs.   To work around this
+                // issue we will allow URLConnectionHttpConduit.setFixedLengthStreamingMode() to be used.  
+                // Normally FixedLengthStreamMode is disabled due to a JDK performance issue [CXF-6227].
+                Map<String, List<String>> headers = CastUtils.cast((Map<?, ?>)message.get(Message.PROTOCOL_HEADERS));
```
andymc12: Same here - use `outMessage` instead of `message`

dev/com.ibm.ws.jaxrs.2.0.common/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
```diff
@@ -463,6 +470,30 @@ protected boolean usingProxy() {
...
+                    Iterator<String> expects = expectList.iterator();
+                    while (expects.hasNext()) {
+                        if (expects.next().equalsIgnoreCase("100-continue")) {
+                            allowFixedLengthStreamMode = Boolean.TRUE;                            
+                        }
+                    }
+                }               
+
+            }
+            if ((allowFixedLengthStreamMode != null) && (allowFixedLengthStreamMode.booleanValue())) {
```
andymc12: All of this method (except possibly iterating over the `Expect` header values) still gets called on every request - the only difference is that once a 100-Continue header is found, then it avoids checking for the header on future requests.  Most users are probably not using `Expect: 100-Continue`, so it will slow them down on every request.  If the performance team is available to compare the results of a run with and without this change, it would be nice to see how much of an impact this is to most users.

dev/com.ibm.ws.org.apache.cxf.cxf.rt.transports.http.3.2/src/org/apache/cxf/transport/http/URLConnectionHTTPConduit.java
```diff
@@ -33,6 +33,9 @@
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Iterator;
```
andymc12: Same comments as 2.0.common change.

PR #11319, opened 2020-03-13
remove extra level of inheritance for resourceAdapter

njr-11:
The extra level of inheritance between application and resourceAdapter might be why schema for classProviderRef is not generating correctly.  This pull will see if we can remove the intermediate internal config element between the two.


njr-11:
#build

Issue #11275, opened 2020-03-11
JakartaEE: JSP 3.0 Feature Support


tbitonti:
This issue covers the feature file changes required for liberty to run jsp-3.0.

volosied:
Completed, but now I am enabling the jsp 2.3 fat to run with the jsp-3.0 feature.

pnicolucci:
I'm going to close this issue in favor of using https://github.com/OpenLiberty/open-liberty/issues/11717 to track the remaining work.

PR #11325, opened 2020-03-16
update jsp 3.0 bnd.bnd buildpath

volosied:
Previous commit did not include this change.  Since resources are overlay-ed, there is no need for this package.  (I don't believe it's even picked up in the jakarta bundle).


volosied:
for issue #11275

PR #11328, opened 2020-03-16
MP Config 1.4 raw value cache configuration

### LINKED ISSUE #11257: MP Config-1.4: Allow configuration of cache invalidation times
Two timed caches were added to MP Config 1.4 to address performance
* cache for raw values retrieved from config in `Config14Impl`
* cache for the map of app-specific properties retrieved from server.xml in `AppPropertyConfig14Source` (may be removed if #11231 is completed)

Both of these caches have a hard-coded 500ms limit on the period that a value is cached for. We should make these values configurable.

We should consider whether to document and support changing these values in production, or whether they would only be documented for support purposes and problem determination.

When making these values configurable, we may want to support disabling caching altogether.

We should also ensure that we won't cause a memory leak if someone sets the cache time to very large value (e.g. several days).

Azquelt:
Allow the time that values are kept in the cache to be configured.

Fixes #11257

### DIFF
--- a/dev/com.ibm.ws.microprofile.config.1.4/src/com/ibm/ws/microprofile/config14/impl/TimedCache.java
+++ b/dev/com.ibm.ws.microprofile.config.1.4/src/com/ibm/ws/microprofile/config14/impl/TimedCache.java
@@ -90,7 +90,7 @@ public class TimedCache<K, V> {
     public V get(K key, Function<K, V> lookupFunction) {
 
         // If no expiry time is set, do no caching
-        if (delay == 0) {
+        if (delay <= 0) {
             return lookupFunction.apply(key);
         }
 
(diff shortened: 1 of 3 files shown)


dev/com.ibm.ws.microprofile.config.1.4/src/com/ibm/ws/microprofile/config14/impl/TimedCache.java
```diff
@@ -87,6 +87,10 @@ public TimedCache(ScheduledExecutorService executor, long delay, TimeUnit unit)
      */
     public V get(K key, Function<K, V> lookupFunction) {
 
+        if (delay == 0) {
```
tevans78: Please add a comment

tevans78:
#build

dev/com.ibm.ws.microprofile.config.1.4/src/com/ibm/ws/microprofile/config14/impl/TimedCache.java
tevans78: What happens if the system property is set to a negative value?? I don’t think it rounds up to 0? Maybe it should?? See here https://github.com/OpenLiberty/open-liberty/blob/67eeadf6e6ba211a338a1b3e518c1d031f256d72/dev/com.ibm.ws.microprofile.config.1.1/src/com/ibm/ws/microprofile/config/impl/AbstractConfigBuilder.java#L270

Azquelt:
I think this is the wrong point to be checking this, but it matches what <1.3 does and it's easier than changing the common code now.

tevans78:
#build

Azquelt:
#run-libby-bot

PR #11342, opened 2020-03-16
remove dependencies on JCA bundles from jca.management.j2ee

njr-11:
Supporting both javax and jakarta package names will be simpler if less bundles rely on these packages.  We can remove usage from the com.ibm.ws.jca.management.j2ee bundle with some minor refactoring, and then we won't need to worry about updating it for the package rename at all.


njr-11:
#build

njr-11:
#build

njr-11:
#build

PR #11361, opened 2020-03-17
Include Jakarta Connector 2.0 RC1

KyleAure:
Fixes 11330


dev/com.ibm.websphere.appserver.features/visibility/public/connector-2.0/com.ibm.websphere.appserver.connector-2.0.feature
```diff
@@ -0,0 +1,15 @@
...
+visibility=public
+singleton=true
+IBM-API-Package: com.ibm.ws.jca.service; type="internal"
+IBM-ShortName: connector-2.0
+Subsystem-Name: Jakarta Connector 2.0
+-features=\
+ com.ibm.websphere.appserver.jakarta.connector-2.0,\
+ com.ibm.websphere.appserver.transaction-1.2
+-bundles=com.ibm.ws.app.manager.rar, \
+ com.ibm.ws.jca.1.7
```
KyleAure: This section of the feature file I am unsure about.  I know that connector has an API level dependency on transaction.  But not sure if this is the place to specify that.  And whether or not these bundles are still necessary?

njr-11: Given the dependency, we do need include the transaction feature (which will later be swapped out to a Jakarta transaction feature).  However, I think it would be better placed within the private Jakarta Connectors feature, which would need it as well, and then this public feature would inherit it from there.

dev/com.ibm.websphere.appserver.features/visibility/public/connector-2.0/resources/l10n/com.ibm.websphere.appserver.connector-2.0.properties
```diff
@@ -0,0 +1,19 @@
...
+###############################################################################
+#
+#ISMESSAGEFILE FALSE
+#NLS_ENCODING=UNICODE
+#NLS_MESSAGEFORMAT_NONE
+#
+description=This feature enables the configuration of resource adapters to access Enterprise Information \
+ Systems (EIS) from applications. The configuration of a resource adapter also includes the configuration \
+ of connection factories, administered objects and activation specifications. Any resource adapter that \
+ complies with the Jakarta Connectors 2.0 specification or lower can be used. High performance connection pooling is also provided.
```
njr-11: I don't know yet if the "or lower" statement will be true.  For the time being, I would recommend omitting that part until we know whether that is the case.

KyleAure:
#build 
#libbybot

dev/com.ibm.websphere.appserver.features/visibility/public/connectors-2.0/resources/l10n/com.ibm.websphere.appserver.connectors-2.0.properties
```diff
@@ -0,0 +1,19 @@
...
+###############################################################################
+#
+#ISMESSAGEFILE FALSE
+#NLS_ENCODING=UNICODE
+#NLS_MESSAGEFORMAT_NONE
+#
+description=This feature enables the configuration of resource adapters to access Enterprise Information \
+ Systems (EIS) from applications. The configuration of a resource adapter also includes the configuration \
+ of connection factories, administered objects and activation specifications. Any resource adapter that \
+ complies with the Jakarta Connectors 2.0 specification. High performance connection pooling is also provided.
```
Charlotte-Holt: Add a comma between `administered objects` and `and activation specifications` in  `The configuration of a resource adapter also includes the configuration of connection factories, administered objects and activation specifications.`

Also, add a hyphen in `High performance` (i.e., `High-performance`).

andrea-pichler:

PR #11409, opened 2020-03-22
7182 AutoExpand Only When Modified

### LINKED ISSUE #7182: AutoExpand Only When Modified
Current open-liberty "auto-expand" behavior is to always expand applications on server startup.  That has two issues:
1. That has a significant performance cost, with auto-expand costing (SU: 8% FP: 5% CPU: 7%) for a CDI application with many JAR files (see below).
2. Updates to application files which were made when the server was running are copied over on the next server restart.

tomb-anno-cache-190211-cdi-fat-autoExpand-true-j9-828-sr6-190103_01-cpus-4-runs-10-fp-try-1.sufp.results:SU: 6808 FP: 217 CPU: 11.88 app: cdi-fat
tomb-anno-cache-190211-cdi-fat-autoExpand-false-j9-828-sr6-190103_01-cpus-4-runs-10-fp-try-1.sufp.results:SU: 6321 FP: 206 CPU: 11.09 app: cdi-fat

As an alternative, auto-expand could be modified to only expand applications which have been updated.  Change detection would be by application time stamp.  (Change detection could be by other means, such as file size and/or SHA value.  Both of these options would require more code updates than detection of changes to the update time, which would be directly stored on as the update time of the expanded application.)

If made as a change of the default behavior, both current unit and FAT tests would see failures, as would automated updates to customer environments, in cases where application updates did not ensure updates to the application timestamp, or which relied on the restoration of application files which is performed during server startup.

Alternatively, auto-expand could be modified to include a third option, or could be modified by the addition of a new attribute.  That is:

Adding a new attribute value:
    <autoExpand="true|false|whenChanged"/>

Adding a new attribute:
    <autoExpand="true|false"/>
    <autoExpandWhen="always|changed"/>

A new attribute, if added, would be used only when auto-expand was enabled.

Related to this function are dynamic expansions which are made by the Liberty artifact file system, where-in nested archives are automatically expanded to the server work area).

Prototype code of the extraction modification (but not of new attribute handling) is at this commit:
https://github.com/tbitonti/open-liberty/commit/c9eabcee0d3ffb983d6cf97bebc5b93ba7ba8c6a

Of branch: 
https://github.com/tbitonti/open-liberty/tree/AutoExpand

That is, the changes to the expansion code are straightforward.  The main design issue is the change of behavior and how to configure to enable the new behavior.

NottyCode:
Store the app archive size and last modified date in a cache file in the
workarea. Then on restart check this file against the original archive
and only extract if they are different.  If anything goes wrong assume
we reextract.
Fixes #7182

### DIFF
--- a/dev/com.ibm.ws.app.manager.war/src/com/ibm/ws/app/manager/ear/internal/EARDeployedAppInfoFactoryImpl.java
+++ b/dev/com.ibm.ws.app.manager.war/src/com/ibm/ws/app/manager/ear/internal/EARDeployedAppInfoFactoryImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2019 IBM Corporation and others.
+ * Copyright (c) 2012, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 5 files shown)


NottyCode:
#build this please Libby

NottyCode:
#build

NottyCode:
#build

NottyCode:
#build

NottyCode:
The test failure is a known pre-existing intermittent failure that is being worked.

NottyCode:
#run-libby-bot

NottyCode:
Two tests were added that change the extracted war and ear files and then make a request to verify the change. The server is then restarted and the resource requested again. If the file is re-extracted the change will not be there and the tests fail.

PR #11425, opened 2020-03-23
Fix jsf.2.2_fat PI90507 tests

### LINKED ISSUE #11419: JSF 2.2 PI90507 FAT is incorrect
The test `com.ibm.ws.jsf22.fat.tests.JSF22AparTests.testPI90507NonBindingCase()` expects that it will be called before the next test method `testPI90507BindingCase()`; it fails with `PostConstruct was not called` if that order is changed.  Test invocation order isn't guaranteed, so this test needs to be updated.

wtlucy:
Fixes #11419


wtlucy:
#build

wtlucy:
#run-libby-bot

PR #11438, opened 2020-03-24
Fix JarFileClassLoader to ignore redefined package when parallel-enabled

dazavala:
Modify JarFileClassLoader.definePackage() to ignore the IllegalArgumentException that may be thrown by the call to java.lang.ClassLoader.definePackage whenever JarFileClassLoader is parallel enabled. The exception occurs when two threads attempt to concurrently define the same package via the JarFileClassLoader, where the first thread defines the package before the second, and the second fails with the exception.


dazavala:
#build

dazavala:
#build

dazavala:
fixes #11445

dazavala:
The build errors have already been reported and are not due to this change.

PR #11440, opened 2020-03-24
Introduce lightweight HADB retries and server stopping code

### LINKED ISSUE #11410: Lightweight Retries for certain operations in TX DB Recovery Log scenario
Where HADB support is available for the case where transaction recovery logs are being stored in an RDBMS certain operations should be retried quickly: namely the heartbeat and claimPeer operations.

neilgsyoung:
Fixes #11410


neilgsyoung:
#build

PR #11452, opened 2020-03-24
ability to identify jakarta version and update first project to use

njr-11:
Add platform feature for jakarta 9, and use it to identify whether Jakarta 9 vs earlier is needed in one of the several projects that will need this information.


njr-11:
#build

PR #11466, opened 2020-03-25
Timer Action Test Handle Slow Systems

tkburroughs:
Test is expecting timer to run every 1 second, which may not occur on "slow" systems.
When using "ONCE", slow systems may skip to the next future 1 second boundary,
so the test needs to account for that.


tkburroughs:
#build

PR #11477, opened 2020-03-26
Update oracle docker image

KyleAure:
Updating the oracle docker images from using Oracle 11g to Oracle 18c.

This is a necessary change to ensure that we continue to support the latest database versions. In addition, the previous docker image we were using was slimmed down and as a result removed support for distributed transactions (XA).


KyleAure:
#build

PR #11480, opened 2020-03-26
Repeat FAT tests with Jakarta EE

aguibert:
To repeat an entire FAT bucket, or a subset of FAT tests with JakartaEE, the following repeat action could be used:
```java
@ClassRule
public static RepeatTests r = RepeatTests.withoutModification() // EE7
                    .andWith(FeatureReplacementAction.EE8_FEATURES()) // EE8
                    .andWith(new JakartaEEAction()); // EE9
```

When a JakartaEE repeat action runs, the following steps occur:
1. The `JakartaTransformer.main()` method is invoked on each war/ear file found under `publish/servers/`. This may also be customized to include more/less files
2. The standard `FeatureReplacementAction` kicks in to update server.xml's from EE7/8 to EE9 features. For example, `servlet-4.0 --> servlet-5.0` and `componenttest-1.0 --> componenttest-2.0`

### How to enable this in a FAT bucket

1. Add a `RepeatTests` `@ClassRule` (if not already present). If you add it at the `FATSuite.java` level the entire FAT bucket will repeat. Or, if you add it only on a specific test class, the repeats will only apply to that class.
2. Add the transformer dependencies in the FATs build.gradle file: `addRequiredLibraries.dependsOn addJakartaTransformer`
3. (Optional) If certain tests don't apply to Jakarta EE 9, they may be skipped by adding the `@SkipForRepeat(EE9_FEATURES)` annotation to the `@Test` method


dev/com.ibm.ws.componenttest.2.0/bnd.bnd
```diff
@@ -0,0 +1,74 @@
+#*******************************************************************************
+# Copyright (c) 2017 IBM Corporation and others.
```
njr-11: Minor: Wrong year copyright for new file.  Some other files have the same issue.

dev/fattest.simplicity/src/componenttest/rules/repeater/JakartaEE9Action.java
```diff
@@ -0,0 +1,226 @@
...
+ * <li>Update all server.xml configs under the autoFVT/publish/ folder to use EE 9 features</li>
+ * </ol>
+ */
+public class JakartaEE9Action extends FeatureReplacementAction {
+
+    private static final Class<?> c = JakartaEE9Action.class;
+
+    public static final String ID = "EE9_FEATURES";
+
+    static final String[] EE9_FEATURES_ARRAY = { "jakartaee-9.0", "webProfile-9.0", "jakartaeeClient-9.0", "servlet-5.0", "jdbc-4.2", "javaMail-2.0", "cdi-3.0", "jpa-3.0",
```
njr-11: Why is jdbc-4.2 an EE9 feature?  Also, concurrent-2.0 is missing.  Are others missing as well - I don't see EJB (whatever the new name for it is) ?

aguibert: jdbc-4.2 doesn't have to be in here, what this implies is that when server.xml's are getting upgraded to EE9-level features, we will also upgrade to jdbc-4.2 if older jdbc version are specified like 4.0 or 4.1. While tests would technically work with any JDBC feature version, I think it would be good for us to test with the latest features when possible -- this list can also be customized at the FAT bucket level, so if a FAT wants to specifically test an older JDBC version it can still do so.

Regarding other features like concurrent-2.0, yes I will get those added.

PR #11483, opened 2020-03-26
test connectionManager pool size increase and app restart

njr-11:
Write tests that identify whether updates to nested and top level connectionManager elements causes an application restart, in the case of dataSource and connection factories.
The test is enabled for dataSource, where it enforces that the application does not restart.
Part of the test is commented out for connection factories, where observed behavior is that the application does restart. In that case, the connectionManager element handles its own modified, however, OSGi wants to sent a modified (leading to deactivate/activate) to the connection factory, triggering the app recycle logic.

### DIFF
--- a/dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
+++ b/dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
@@ -379,7 +379,7 @@ public class ConfigTest extends FATServletClient {
 
         try {
             updateServerConfig(config, EMPTY_EXPR_LIST);
-            // Behavior should now reflect the new setting of 1 for maxPoolSize
+            // Behavior should now reflect the new setting of 2 for maxPoolSize
             runTest(basicfat, "testMaxPoolSize2");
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
```diff
@@ -372,6 +372,25 @@ public void testConfigChangeConnectionManager() throws Throwable {
...
         }
 
+        runTest(basicfat, "setServletInstanceStillActive");
+
+        // Increase maxPoolSize to 2
+        conMgr2.setMaxPoolSize("2");
+
+        try {
+            updateServerConfig(config, EMPTY_EXPR_LIST);
+            // Behavior should now reflect the new setting of 1 for maxPoolSize
```
gjwatts: Change 1 -> 2 for maxPoolSize comment

njr-11: Thanks for catching that. I added a commit to fix the comment.

njr-11:
#build
#run-libby-bot

PR #11490, opened 2020-03-26
Minor Fix to LDAPHelper

### LINKED ISSUE #11478: Minor code issue in LdapHelper.getRDN in com.ibm.ws.security.wim.adapter.ldap
**Describe the bug**
See com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/LdapHelper.java around line 280.

Inside method getRDN, there is fallback code when the LdapName code fails. That code is wrong:
```java
            int pos1 = DN.indexOf(',');
            if (DN.charAt(pos1 - 1) == '\\') {
                pos1 = DN.indexOf(pos1, ',');
            }
```

should be:
```java
            int pos1 = DN.indexOf(',');
            while (DN.charAt(pos1 - 1) == '\\') {
                pos1 = DN.indexOf(',', pos1+1);
            }
```

1. String.indexOf(int, int) - the first parameter is the search character, the second parameter is the start position, not the opposite.
2. Better use a while loop, since there is no reason why a DN should have only one escaped comma.
3. pos1 must increase, otherwise it starts searching at a position where the comma is already found.

The issue is very minor since in most cases, the fallback code is not executed. The code is only reachable when you have a buggy LDAP that does not follow the distinguished name format rules.

**Steps to Reproduce**
N/A

**Expected behavior**
If the fallback code is executed, it should skip all escaped comma (\,) and find the first comma that is not escaped.

**Diagnostic information:**
N/A

**Additional context**
N/A

agiridharan:
fixes #11478

### DIFF
--- a/dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/LdapHelper.java
+++ b/dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/LdapHelper.java
@@ -279,8 +279,8 @@ public class LdapHelper {
             e.getMessage();
             DN = DN.trim();
             int pos1 = DN.indexOf(',');
-            if (DN.charAt(pos1 - 1) == '\\') {
-                pos1 = DN.indexOf(pos1, ',');
+            while (DN.charAt(pos1 - 1) == '\\') {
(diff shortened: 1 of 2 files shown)


jvanhill:
#libby

agiridharan:
#build

agiridharan:
#run-libby-bot

PR #11509, opened 2020-03-27
Fix availability of fat.test.localrun

KyleAure:
Previously we were performing the following sequence of steps to set fat.test.localrun to true:

```java
  System.setProperty("fat.test.localrun", "true");
  environment System.getProperties()
```

This is done during a gradle task's configuration step. Which happens regardless of whether the task runs. I had moved it to the execution step, to ensure this property is not set when running remote builds. However, this resulted in the property not being set locally during test runtime.

Instead, let's just add fat.test.localrun to the environment directly. No need to set it as a system property. Since, our build infrastructure runs tests on child build machines this environment variable won't be available at run time in those cases.


KyleAure:
#build

PR #11511, opened 2020-03-27
Update Jakarta Servlet 4 FAT Enablement 

volosied:
updating bnd.bnd and skipping problematic WCSameSiteCookieAttributeTests class


volosied:
Will need to update bnd.bnd again (for tested.features) when the jakarta appsecurity feature is created.

PR #11539, opened 2020-03-30
Remove old, unused version of jquery that was unnecessarily getting packaged

dmorgant:
The com.ibm.ws.security.openidconnect.server packaged both 

> lib/jquery/jquery-3.2.0/jquery.min.js   (v3.2.0)

and 

> resources/WEB-CONTENT/common/lib/jquery/jquery.min.js    (v3.4.1)

The jquery.min.js  in lib/jquery is unnecessarily getting packaged and is not needed and should be removed.


dmorgant:
fixes #11540

dmorgant:
#build

PR #11541, opened 2020-03-30
Change processing of feature bundles

brenthdaniel:
PR to rely on a list of bundles from the feature manager rather than bundleChanged events.

### DIFF
--- a/dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/statemachine/StartAction.java
+++ b/dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/statemachine/StartAction.java
@@ -43,7 +43,7 @@ class StartAction implements Action {
     private final AtomicLong _startTime = new AtomicLong();
     private final ApplicationMonitor _appMonitor;
     private final boolean _update;
-    private boolean cancelled = false;
+    private volatile boolean cancelled = false;
     private final AtomicReference<Future<?>> _slowMessageAction = new AtomicReference<Future<?>>();
     private final CompletionListener<Boolean> _listener = new CompletionListener<Boolean>() {
         @SuppressWarnings("deprecation")
@@ -166,7 +166,6 @@ class StartAction implements Action {
     @Override
     public void cancel() {
         this.cancelled = true;
-        //      _callback.set(null);
         stopSlowStartMessage();
     }
 }
--- a/dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/statemachine/StartAction.java
+++ b/dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/statemachine/StartAction.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2014 IBM Corporation and others.
+ * Copyright (c) 2012, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -43,6 +43,7 @@ class StartAction implements Action {
     private final AtomicLong _startTime = new AtomicLong();
     private final ApplicationMonitor _appMonitor;
     private final boolean _update;
+    private boolean cancelled = false;
     private final AtomicReference<Future<?>> _slowMessageAction = new AtomicReference<Future<?>>();
     private final CompletionListener<Boolean> _listener = new CompletionListener<Boolean>() {
         @SuppressWarnings("deprecation")
@@ -59,10 +60,12 @@ class StartAction implements Action {
                     AppMessageHelper.get(_aii.getHandler()).audit(key, _config.getName(), TimestampUtils.getElapsedTime(_startTime.get()));
                     callback.changed();
                 } else {
-                    String key = _update ? "APPLICATION_NOT_UPDATED" : "APPLICATION_NOT_STARTED";
-                    NotificationHelper.broadcastChange(_config.getMBeanNotifier(), _config.getMBeanName(), _update ? "application.update" : "application.start", Boolean.FALSE,
-                                                       AppMessageHelper.get(_aii.getHandler()).formatMessage(key, _config.getName()));
-                    AppMessageHelper.get(_aii.getHandler()).audit(key, _config.getName());
+                    if (!cancelled) {
+                        String key = _update ? "APPLICATION_NOT_UPDATED" : "APPLICATION_NOT_STARTED";
+                        NotificationHelper.broadcastChange(_config.getMBeanNotifier(), _config.getMBeanName(), _update ? "application.update" : "application.start", Boolean.FALSE,
+                                                           AppMessageHelper.get(_aii.getHandler()).formatMessage(key, _config.getName()));
+                        AppMessageHelper.get(_aii.getHandler()).audit(key, _config.getName());
+                    }
                     callback.failed(null);
                 }
             } else {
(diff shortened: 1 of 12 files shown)


brenthdaniel:
#build

brenthdaniel:
#build

brenthdaniel:
#build

brenthdaniel:
#build

brenthdaniel:
#build

brenthdaniel:
One of the child builds had issues, so 
#build 
again

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/Provisioner.java
```diff
@@ -305,6 +304,7 @@ public boolean handle(FeatureResource fr) {
                             }
                             level = newLevel;
                             bsl.setStartLevel(level);
+                            installStatus.addBundleAddedDelta(bundle);
```
tjwatson, line 307: This seems like a very limited condition to add to the bundle added delta.  I would have expected this to add all bundles that got installed that did not exist before.  But here we only add bundles that are not fragments and that we need to change the start-level for?

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureManager.java
```diff
@@ -1400,6 +1408,10 @@ protected boolean updateFeatures(WsLocationAdmin locService,
         provisioner.resolveBundles(bundleContext, installedBundles);
 
         if (featureChange.featureBundlesResolved != null) {
+            Map<String, Object> props = new HashMap<String, Object>(1);
```
tjwatson, line 1411: Does this map need to be thread safe?  It seems we populate it here and then leave it open to other threads to get/put stuff on it when the notification is fired.  Should setProperties just take the values and wrap them in an unmodifiable Map in the notification impl? This way they are only read-only on notification firing when they call getProperties.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/Provisioner.java
```diff
@@ -608,7 +588,9 @@ public boolean handle(FeatureResource fr) {
         for (Bundle bundleToUninstall : bundlesToUninstall) {
             try {
                 bundleToUninstall.uninstall();
+                installStatus.addBundleRemovedDelta(bundleToUninstall);
             } catch (IllegalStateException e) {
+                e.printStackTrace();
```
tjwatson, line 593: Did you intend to do a printStackTrace here?

tjwatson: It would seem that you should add to the status the intention to uninstall this bundle before calling uninstall().  The BundleProcessor no longer will get an uninstalled event so it would miss it if we somehow had another thread uninstall the bundle before we could here.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureManager.java
tjwatson: This has been addressed

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/Provisioner.java
brenthdaniel: Moved to the installFeatureBundle method

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/BundleProcessor.java
```diff
@@ -254,12 +248,14 @@ public void bundleChanged(BundleEvent event) {
...
             if (type == BundleEvent.RESOLVED) {
+
                 if (isFeatureBundle) {
+
                     if (tc.isDebugEnabled()) {
                         Tr.debug(tc, "Adding feature bundle {0}", b);
                     }
                     //add it to the queue for processing later
-                    addFeatureBundles.add(b);
+//                    addFeatureBundles.add(b);
```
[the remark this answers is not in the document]
brenthdaniel: Issues with comments/printStacktrace were fixed by restoring old commits

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/Provisioner.java
brenthdaniel: Changed so the installstatus method is called before the uninstall

brenthdaniel:
#build

dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/statemachine/StartAction.java
```diff
@@ -43,6 +43,7 @@
     private final AtomicLong _startTime = new AtomicLong();
     private final ApplicationMonitor _appMonitor;
     private final boolean _update;
+    private boolean cancelled = false;
```
tjwatson, line 46: Should this be volatile? I'm pretty sure it should be, it looks like it would be read from one thread and written with another.

dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/statemachine/StartAction.java
```diff
@@ -160,7 +165,8 @@ private void stopSlowStartMessage() {
     /** {@inheritDoc} */
     @Override
     public void cancel() {
-        _callback.set(null);
+        this.cancelled = true;
+        //      _callback.set(null);
```
tjwatson, line 169: Did you intend to leave this commented out?  It is unclear to me if _callback will always get set from the CompletionListener when cancel is called and it is safe to remove this call.  But if that is all good then this commented out code will always make the next set of eyes have the same questions I do (I think).

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/BundleInstallStatus.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009 IBM Corporation and others.
+ * Copyright (c) 2009-2020 IBM Corporation and others.
```
tjwatson, line 2: I thought the convention was to use a comma here `2009, 2020`.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureManager.java
```diff
@@ -744,10 +750,12 @@ protected void update(FeatureChange featureChange) throws IllegalStateException
 
                     //register a service that can be looked up for server start.
                     // Need a two phase approach, since ports will be opened for listening on the first phase
-                    bundleContext.registerService(ServerStarted.class, new ServerStarted() {}, new Hashtable<String, Object>());
+                    bundleContext.registerService(ServerStarted.class, new ServerStarted() {
+                    }, new Hashtable<String, Object>());
```
tjwatson, line 754: Not your change, just showing up because of line differences.  Creating a new empty Hashtable instead of passing null is just creating garbage.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureManager.java
```diff
@@ -744,10 +750,12 @@ protected void update(FeatureChange featureChange) throws IllegalStateException
...
 
                     // components which needed to wait till ports were opened for listening need to wait till Phase2
-                    bundleContext.registerService(ServerStartedPhase2.class, new ServerStartedPhase2() {}, new Hashtable<String, Object>());
+                    bundleContext.registerService(ServerStartedPhase2.class, new ServerStartedPhase2() {
+                    }, new Hashtable<String, Object>());
```
tjwatson, line 758: Same hashtable object garbage comment.

dev/com.ibm.ws.runtime.update/src/com/ibm/ws/runtime/update/RuntimeUpdateNotification.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014 IBM Corporation and others.
+ * Copyright (c) 2014-2020 IBM Corporation and others.
```
tjwatson, line 2: use of dash in copyright dates.

brenthdaniel:
artifactory failure... 
#build again

tjwatson:
#run-libby-bot

PR #11555, opened 2020-03-31
Fix unset token

utle:
Fixed unset tokenManager and there is no need for the DeleteAuthCache service to start immediately set to true.


utle:
#build 
#run-libby-bot

utle:
#build

tevans78:
The previous PB did not run properly. However, it is possible that we want this change in the GM so I am going to make an exception and merge it anyway. The change is small and look safe. They were tested locally.

PR #11565, opened 2020-03-31
11479: Enhance RP tracking cookie creation/update logic

ayoho:
- Updates the tracking cookie name (base name, and adds OAuth provider ID into the name)
- Sets the path of the cookie to the path of the request
- Sets the "secure" flag on the created cookie


ayoho:
#build

PR #11575, opened 2020-03-31
Issue #11572: Fix Calendar EJB timer with no expiration

### LINKED ISSUE #11572: Calendar based EJB Timer created with no expirations returned by getTimers
If a calendar based EJB timer is created with no expirations, it is considered to not exist, and should
not be returned when calling getTimers().  Only active timers should be returned by getTimers().
However, if the calendar based timer with no expirations is created in a global transaction and
getTimers() is called in that same transaction, then the timer is returned, despite having
no epirations and classified as not existing.

Same problem does not occur if the timer is created in a different global transaction or the unspecified transaction context (NOT_SUPPORTED).

tkburroughs:
A calendar EJB timer with no expiration may be returned by getTimers()
if getTimers() is called in the same transaction that creates the timer.
A timer with no expiration is not active and should not be returned by
getTimers().

Updated TimerNpImpl to expose the expiration (i.e 0 means none) and
then updated ContainerTx to not queue a timer for starting if it in
fact has no expirations and will thus never expire. Since the timer
is not queued to start on tx commit, the getTimers() will not find it.

Please ignore formatting changes in ContainerTx; only real changes start at line 1033.

fixes #11572

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/ContainerTx.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/ContainerTx.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1997, 2014 IBM Corporation and others.
+ * Copyright (c) 1997, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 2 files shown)


tkburroughs:
#build #spawn.fullfat.buckets=com.ibm.ws.ejbcontainer.timer_fat

tkburroughs:
#build #spawn.fullfat.buckets=com.ibm.ws.ejbcontainer.timer_fat

PR #11605, opened 2020-04-02
Fix processing of windows paths that start with a variable

brenthdaniel:
Fix processing of windows paths that start with a variable. We used to interpret any path that starts with an environment variable as absolute, but when we added support for variable resolution in include strings it changed the timing so that the environment variable was resolved before the 'absolute' check.


brenthdaniel:
#build

wraschke:
#help

wraschke:
#help

brenthdaniel:
#11475

PR #11647, opened 2020-04-06
remove EE Concurrency dependency from transaction context

njr-11:
Minor updates to Transaction context propagation and the MicroProfile wrapper around it, in order to remove the dependency of both on EE Concurrency API.  This will aid in the update to support Jakarta by making it possible to convert fewer bundles. (The transaction context provider bundle would still need updates for other EE specs, but this is a start and will allow work to proceed with the Concurrency spec)


njr-11:
#build

njr-11:
Build has 5 test failures in 3 test buckets which are pre-existing issues.

Issue #11584, opened 2020-04-01
EJB Custom Bindings: Append interface to simple-binding-name for beans with multiple interfaces


olendvcook:
When trying to enable the No Interface tests in the ejbcontainer.binding_fat I got this warning and the simple-binding-name was not bound for some local beans:

`SIMPLE_BINDING_NAME_MISSUSED_CNTR0168W=CNTR0168W: The {0} enterprise bean in the {1} module within the {2} application is configured with a simple binding name, but has multiple local or remote interfaces.  The naming locations that are used to bind these interfaces will differ from the simple binding name that is specified.`
`SIMPLE_BINDING_NAME_MISSUSED_CNTR0168W.explanation=Because the enterprise bean has multiple interfaces, each binding name must be extended to make it unique.`
`SIMPLE_BINDING_NAME_MISSUSED_CNTR0168W.useraction=Complete one of the following actions: /n Look up the interfaces for this enterprise bean using the extended names. /n Change the configuration data for this bean to specify unique binding locations for each interface.`

~~Probably something with how the test is set up~~

Edit: It turns out when a bean has a simple-binding-name binding specified but has multiple interfaces, it will append the interface to the end of the simple-binding-name. So this is now an implementation delivery as the test expected to look them up with the extended name.

olendvcook:
Complete

PR #11688, opened 2020-04-07
Append interface to simple-binding-name for beans with multiple interfaces

olendvcook:
#11584

It turns out when a bean has a simple-binding-name binding specified but has multiple interfaces, it will append the interface to the end of the simple-binding-name.

We already print this warning out:
`SIMPLE_BINDING_NAME_MISSUSED_CNTR0168W=CNTR0168W: The {0} enterprise bean in the {1} module within the {2} application is configured with a simple binding name, but has multiple local or remote interfaces.  The naming locations that are used to bind these interfaces will differ from the simple binding name that is specified.`
`SIMPLE_BINDING_NAME_MISSUSED_CNTR0168W.explanation=Because the enterprise bean has multiple interfaces, each binding name must be extended to make it unique.`
`SIMPLE_BINDING_NAME_MISSUSED_CNTR0168W.useraction=Complete one of the following actions: /n Look up the interfaces for this enterprise bean using the extended names. /n Change the configuration data for this bean to specify unique binding locations for each interface.`


olendvcook:
#build

olendvcook:
#build

PR #11712, opened 2020-04-08
Improve ReferenceContext trace

tkburroughs:
trace when items are found in the processed classes and injection targets caches,
and the sizes of those caches


tkburroughs:
#build

tkburroughs:
One test failure has a known issue unrelated to this change; ready to merge

Issue #11627, opened 2020-04-03
Deadlock between ServletContainerInitializer and Future.get from application code


njr-11:
There is a deadlock in webcontainer startup where a ServletContainerInitializer or ServletContextListener schedules a task using a managed executor and awaits completion of the task from the same method (onStartup/contextInitialized). After reproducing it in a test case (see commented TODO lines indicating disabled test code in https://github.com/OpenLiberty/open-liberty/pull/11608 ), I’ve narrowed it down to a deadlock between a webcontainer object, com/ibm/ws/webcontainer/webapp/WebApp$1 and the propagation of webcontainer metadata context to the thread where the task is attempting to start.

Here is that stack of the initialize thread, which obtains the lock, schedules a task and awaits the task result with Future.get(),

```
3XMTHREADINFO      "Default Executor-thread-14" J9VMThread:0x0000000007F15700, omrthread_t:0x00007FA26DAD15D8, java/lang/Thread:0x0000000701915498, state:P, prio=5
3XMJAVALTHREAD            (java/lang/Thread getId:0x48, isDaemon:true)
3XMTHREADINFO1            (native thread ID:0x4C153B, native priority:0x5, native policy:UNKNOWN, vmstate:P, vm thread flags:0x00020001)
3XMTHREADINFO2            (native stack address range from:0x00007000107DE000, to:0x000070001081E000, size:0x40000)
3XMCPUTIME               CPU usage total: 0.167381000 secs, current category="Application"
3XMTHREADBLOCK     Parked on: com/ibm/ws/threading/internal/PolicyTaskFutureImpl$State@0x000000070165F4A8 Owned by: <unknown>
3XMHEAPALLOC             Heap bytes allocated since last GC cycle=0 (0x0)
3XMTHREADINFO3           Java callstack:
4XESTACKTRACE                at jdk/internal/misc/Unsafe.park(Native Method)
4XESTACKTRACE                at java/util/concurrent/locks/LockSupport.park(LockSupport.java:194)
4XESTACKTRACE                at java/util/concurrent/locks/AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:885)
4XESTACKTRACE                at java/util/concurrent/locks/AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1039)
[...]

PR #11719, opened 2020-04-09
Deadlock during app startup

pmd1nh:
https://github.com/OpenLiberty/open-liberty/issues/11627


pmd1nh:
com/ibm/ws/context/service/serializable/ThreadContextDescriptorImpl.taskStarting(ThreadContextDescriptorImpl.java:348(Compiled Code))

This method is in the path of trigger the WeApp.init which is in-fly.  The comment in this method says: "java.lang.IllegalStateException exception if the application component is not started or deployed."

Fix here will be for WebApp.init() process to throw an IllegalStateException because it is not started.

public ArrayList<ThreadContext> taskStarting() throws RejectedExecutionException {
        final boolean trace = TraceComponent.isAnyTracingEnabled();

        // EE Concurrency 3.3.4: All invocations to any of the proxied interface methods will fail with a
        // java.lang.IllegalStateException exception if the application component is not started or deployed.  <<<<<<<<<SEE THIS COMMENT>>>>

PR #11762, opened 2020-04-14
handle  openliberty.io can't be reached to retrieve https://openliber…

aknguyen7:
If openliberty.io can't be reached to retrieve "https://openliberty.io/latestVersion.js" to do the version comparison, then the nls text is not displayed because of the js undefined error. 
The check whether "latestReleasedVersion" has been defined before calling doVersionCheck(latestReleasedVersion) won't fix the problem, instead we would need to do a try catch to handle the Reference error when unable to load latestVersion.js


aknguyen7:
#build

aknguyen7:
#build

aknguyen7:
With the latest submitted build, there are known defects on tests failures from the build

PR #11768, opened 2020-04-15
DO NOT MERGE - 9842 ttl not working

### LINKED ISSUE #9842: Message send with explicit timeToLive not working
Using MessageProducer.send() to send a message with an explicit TTL value doesn't respect this and uses the queue defined TTL (default 0: no time-out) instead.

This is missed in the WS-CD com.ibm.ws.messaging.clientcontainer_fat 'testTimetoLiveVariation' FAT test (the result isn't reported). I'm in the process of porting this over to open-liberty which is where/why I've encountered it.

Simple reproduction:

```
QueueConnection connection = ((QueueConnectionFactory)jndi.lookup("java:comp/env/jndi_JMS_BASE_QCF")).createQueueConnection();
Queue queue = (Queue) jndi.lookup("java:comp/env/jndi_QUEUE");
QueueSession session = connection.createQueueSession(false,javax.jms.Session.AUTO_ACKNOWLEDGE);
MessageProducer producer = session.createProducer(null);
MessageConsumer consumer = session.createConsumer(queue);

producer.send(queue,session.createMessage(),DeliveryMode.NON_PERSISTENT,0,500);  // 500 ms life
Thread.sleep(2000); // much longer than message life
if (null!=consumer.receiveNoWait()) System.out.println("failed");

```
Printing the message will show:

 JMSExpiration:    0
 JMSDestination:   queue://QUEUE

If TTL is set to a fixed value in the client container's server.xml, e.g.

    <properties.wasJms queueName="QUEUE" timeToLive="750"/>                                                                     

The message expires but if read in time and printed it shows, e.g.

 JMSExpiration:    1574177165300
 JMSDestination:   queue://QUEUE_ONE?timeToLive=750

It is expected that the TTL value in the send() call be used and that the message expires before it can be consumed in the example.

dhaggart:
fixes #9842

### DIFF
--- a/dev/com.ibm.ws.messaging.jms.2.0/src/com/ibm/websphere/sib/api/jms/JmsDestination.java
+++ b/dev/com.ibm.ws.messaging.jms.2.0/src/com/ibm/websphere/sib/api/jms/JmsDestination.java
@@ -65,6 +65,7 @@ public interface JmsDestination extends Destination, Serializable, Referenceable
    * @see javax.jms.MessageProducer
    */
   public void setDeliveryMode(String deliveryMode) throws JMSException;
+  public void setDeliveryModeDefault(String deliveryMode) throws JMSException;
 
   /**
    * Get the deliveryMode.<p>
@@ -72,6 +73,7 @@ public interface JmsDestination extends Destination, Serializable, Referenceable
    * @return a String representing the deliveryMode.
    */
   public String getDeliveryMode();
+  public String getDeliveryModeDefault();
 
   /**
    * Set the timeToLive (in milliseconds) to be used for all messages sent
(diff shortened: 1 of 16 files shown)


dhaggart:
#build

dhaggart:
#run-libby-bot

dhaggart:
Rebuild with changes to JMS 1.1 and FAT tests (to enable the attributes) in CL allowed clean completion.

dhaggart:
Reviewers,
I used 5 commits to break the code changes up.

The first is just some changes to fix-up tracing (I always try correct it when I see it).
The second is just some additional changes to changes Jared Anderson made to the utility class used in the test cases I ported.

The third is turning on and enhancing the JMS1 TTL tests & implementing simple JMS2 TTL testing. (I included changes to the TTL values to make them distinct and therefore easier to track in the tracing.)

The forth and fifth should really be merged as one; they cover the real change: making the override work as it does in tWAS (i.e. only override if set - i.e. makes them optional) and (for kicks) the addition of "default" attributes that behave as documented in that they provide only default (not override) values.  Should we do something like this we'd then have surely the best of both worlds? ;-)  (And minimal side-effects on existing installations; the only that would need any chaning are those that rely on the OpenLiberty-only enforced override of TTL=0 where they'd simply have to add 'timeToLive="0"' to their XML to continue as before.)

PR #11780, opened 2020-04-16
enable Jakarta in concurrent db, ejb, config tests

njr-11:
Run 3 additional test buckets using the Jakarta packages for Concurrency.
Some other features used by these tests will temporarily remain at Java EE until Jakarta versions of other features are available.  It is only possible to mix & match Jakarta/Java EE like this while the features are in beta.  Use of one vs the other will later be enforced.


njr-11:
#build

njr-11:
#build

PR #11787, opened 2020-04-16
11785: Intermittent h2 timing test failure

### LINKED ISSUE #11785: intermittent h2 timing test failure
The request is sent by the client with the http/2 upgrade header h2c.
The server accepts the upgrade and replies with the 101 response.
The connection preface is sent by the client with an empty settings frame, all good so far.
The server has stored away the initial request, waiting for the connection to complete setup.
The connection is completed, the server starts to process the initial request.
The server is interrupted, and another settings frame comes in, this one has a bad size.
The server processes this frame, sees the error, sends a goaway frame, and cleans up everything - streams, links, etc.

The server then returns to process the initial servlet request, but all the resources for the request were released, and then it gets the NPE, The actual NPE occurs on when the servlet tries to get the response, and it's null.

loriadi:
fixes #11785


loriadi:
#build

loriadi:
#build

loriadi:
#run-libby-bot

loriadi:
#run-libby-bot

PR #11309, opened 2020-03-13
Improve log messages from the kafka connector

Azquelt:
Ensure that we correctly scoop out the exceptions from the cause chain
for any kafka exceptions so that they get displayed in the console.

Example output after this change:
```
[ERROR   ] Unable to create the publisher or subscriber during initialization
CWMRX1007E: Could not create a Kafka incoming connector for the input channel. The error is org.apache.kafka.common.KafkaException: Failed to construct kafka consumer: org.apache.kafka.common.config.ConfigException: No resolvable bootstrap urls given in bootstrap.servers
```

Fixes #11274

Issue #11722, opened 2020-04-09
mpHealth - readiness check reports UP when application fails to start


matthewpwilson:
**Describe the bug**
My application fails to start up, but the /health/ready endpoint still replies with UP.

**Steps to Reproduce**
1. Configure server with mpHealth 2.0
2. Define a readiness check in my applicaiton.
3. Start the server under conditions where the application will fail to start. In my case, the application should connect to Kafka with mpReactiveMessaging, but Kafka is not available. 
4. GET /health/ready and receive the response UP. 

**Expected behavior**
The readiness check should report DOWN as my application is not available.  This partitcularly important in Kubernetes, because my application appears ready and therefore will get traffic routed to it, but it's actually dead. 

**Diagnostic information:**
 - OpenLiberty Version: 20.0.0.3
 - Java Version: OpenJDK Runtime Environment (1.8.0_242-b08)
 - server.xml configuration:
```
<server description="Sample Liberty server">

  <featureManager>
    <feature>jaxrs-2.1</feature>
    <feature>jsonb-1.0</feature>
    <feature>mpHealth-2.0</feature>
    <feature>mpRestClient-1.3</feature>
    <feature>mpReactiveMessaging-1.0</feature>
  </featureManager>

  <httpEndpoint httpPort="9080" id="defaultHttpEndpoint" host="*" />

  <webApplication location="coffeeshop-ui.war" contextRoot="/">
  </webApplication>

</server>
```

 - [messages.log](https://github.com/OpenLiberty/open-liberty/files/4458174/startup-fail.log)
- App is available on DockerHub: matthewpwilson/coffeeshop-ui:lib-20-03

matthewpwilson:
Trace with HEALTH=all and accompanying messages.log
[trace.log](https://github.com/OpenLiberty/open-liberty/files/4482401/trace.log)
[messages.log](https://github.com/OpenLiberty/open-liberty/files/4482404/messages.log)

emyes:
Hi All,

 Please note that issue exist in Open liberty version 20.0.0.2 as well.

kabicin:
@matthewpwilson I was able to reproduce this on my end, and got the same error.

Our AppTracker class is skipping states when it follows the lifecycle of your app due to the mpReactiveMessaging exception. Based on our implementation, this skipped state ignores that the app ever existed, when the health check is performed, which is problematic.

[...]

PR #11800, opened 2020-04-17
Allow the creation of the KafkaConsumer to be retried

### LINKED ISSUE #10937: mpReactiveMessaging Kafka connector does not retry connection
**Describe the bug**

The OpenLiberty Kafka connector for MP Reactive Messaging requires that the Kafka connection succeeds on startup. If it does not succeed (for example, if the Kafka is not yet available) it does not retry.

This means that your Open Liberty application is left "up" but "sick", with no way to recover.

**Steps to Reproduce**

0. Do not start Kafka
1. Start an OpenLiberty application with mpReactiveMessaging, attempting to connect to Kafka
2. Observe the connection to Kafka has failed
3. Start up Kafka
4. Observe that OpenLiberty does not retry the connection so it cannot recover from the failure state

**Expected behavior**

Open Liberty should retry the connection, enabling the application to recover.

**Other notes**

Unless I am mistaken, the Kafka connector in Quarkus does support this.  See the `retry` option:

https://smallrye.io/smallrye-reactive-messaging/#_retrieving_messages_from_kafka

Azquelt:
Fixes #10937 

This has been built on top of #11309 since it changes the same code but will be rebased once that PR has been merged.

Adds a new config property to the kafka connector `creation.retry.seconds` which configures how long liberty should retry the creation of the `KafkaConsumer`.

### DIFF
--- a/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
@@ -63,17 +63,17 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
 # A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
 kafka.create.incoming.error.CWMRX1007E=CWMRX1007E: A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
 kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
-kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
+kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
-kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
+kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # {0} = channel name, {1} = error message
 kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel could not be created but this operation will be retried. The error is {1}
 kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel but the set up will be retried.
-kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, server message.log, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector set up.
+kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector set up.
 
 
 #-----------------------------------------------------------------------------------------------------------------------------
--- a/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
@@ -70,6 +70,12 @@ kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector fo
 kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
 kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
 
+# {0} = channel name, {1} = error message
+kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel could not be created but this operation will be retried. The error is {1}
+kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel but the set up will be retried.
+kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, server message.log, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector set up.
+
+
 #-----------------------------------------------------------------------------------------------------------------------------
 # Emergency Reactive Messaging error message
 #-----------------------------------------------------------------------------------------------------------------------------
--- a/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
@@ -62,17 +62,17 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
 
 # A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
 kafka.create.incoming.error.CWMRX1007E=CWMRX1007E: A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
-kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
+kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This might be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
 kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
-kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
+kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This might be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
 kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # {0} = channel name, {1} = error message
 kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel cannot be created but this operation is being retried. The error is {1}
-kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
+kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel. This might be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
 kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector setup.
 
 
--- a/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
@@ -62,18 +62,18 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
 
 # A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
 kafka.create.incoming.error.CWMRX1007E=CWMRX1007E: A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
-kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
+kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
 kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
-kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
+kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
 kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # {0} = channel name, {1} = error message
-kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel could not be created but this operation will be retried. The error is {1}
-kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel but the set up will be retried.
-kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector set up.
+kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel cannot be created but this operation is being retried. The error is {1}
+kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
+kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector setup.
 
 
 #-----------------------------------------------------------------------------------------------------------------------------
--- a/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
@@ -10,6 +10,7 @@
  *******************************************************************************/
 package com.ibm.ws.microprofile.reactive.messaging.kafka;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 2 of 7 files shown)


Azquelt:
It's very difficult to test the actual use case because that requires an unresolvable hostname which becomes resolvable as the app is starting, but we should be able to configure a retry, cause some other invalid config failure and check that the correct warning is printed.

Azquelt:
#build
#run-libby-bot

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -70,6 +70,12 @@ kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector fo
 kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
 kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
 
+# {0} = channel name, {1} = error message
+kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel could not be created but this operation will be retried. The error is {1}
+kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel but the set up will be retried.
+kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, server message.log, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector set up.
```
jantley-ibm, line 76: change
server message.log
to
the message.log file on the server

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
```diff
@@ -135,4 +138,25 @@ private void shutdown() {
...
+        while (true) {
+            try {
+                return this.kafkaAdapterFactory.newKafkaConsumer(consumerConfig);
+            } catch (KafkaAdapterException e) {
+                if ((System.nanoTime() - startTime) > retryNs) {
+                    throw e;
+                }
+                Tr.warning(tc, "kafka.create.incoming.retry.CWMRX1009W", channelName, e.getMessage());
+            }
+            Thread.sleep(1000);
```
tevans78, line 158: Please extract this sleep time to a constant. Any thoughts on if it needs to be configurable?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
```diff
@@ -96,6 +98,7 @@ private void shutdown() {
             String topic = config.getOptionalValue(KafkaConnectorConstants.TOPIC, String.class).orElse(channelName);
             int maxPollRecords = config.getOptionalValue(KafkaConnectorConstants.MAX_POLL_RECORDS, Integer.class).orElse(500);
             int unackedLimit = config.getOptionalValue(KafkaConnectorConstants.UNACKED_LIMIT, Integer.class).orElse(maxPollRecords);
+            int retrySeconds = config.getOptionalValue(KafkaConnectorConstants.CREATION_RETRY_SECONDS, Integer.class).orElse(0);
```
Emily-Jiang, line 101: In SmallRye, the wait between the retry is 30s. https://smallrye.io/smallrye-reactive-messaging/smallrye-reactive-messaging/2/kafka/kafka.html I think the default should be 30s instead of 0.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
```diff
@@ -135,4 +138,25 @@ private void shutdown() {
...
+        if (retrySeconds == 0) {
+            return this.kafkaAdapterFactory.newKafkaConsumer(consumerConfig);
+        }
+
+        long retryNs = Duration.ofSeconds(retrySeconds).toNanos();
+        long startTime = System.nanoTime();
+
+        while (true) {
+            try {
+                return this.kafkaAdapterFactory.newKafkaConsumer(consumerConfig);
```
Emily-Jiang, line 151: Can we sync up with SmallRye for the `retry-max-wait`, `retry-attempts`?

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
Azquelt: I disagree.

We can't tell the difference between a failure caused by the hostname not being resolvable and a failure caused by bad configuration.

If the user has configured the application wrongly, we want to app to fail to start as soon as possible, not pointlessly retry for 30s.

As far as we've seen so far, having the broker hostname not be resolvable is the only failure that Kafka treats this way, in most cases network failures are already retried automatically by the kafka client.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
Azquelt: I'm not sure what the point of those options are in smallrye, they don't seem to be the same as this one.

They talk about retrying the connection if it goes down, which is not what this change is doing.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
Azquelt: No-one should have to use this unless they're deploying in an environment where the kafka broker and the app are being brought up at the same time. (e.g. during a kubernetes deployment in a test environment)

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
Azquelt: It's a fairly limited use case (broker hostname is not resolvable at startup) so I went for a simpler single option configuration, rather than two options (number of times to retry + delay between retries).

Azquelt: I'm not sure that extracting it into a constant actually helps anything here.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/KafkaIncomingConnector.java
Azquelt: I named the property `creation.retry.seconds` to emphasise that it's only the _creation_ of the kafka client which is retried.

Once we have set up the kafka client, we leave it to do any retries itself.

skasund:
L2 message review completed

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -63,12 +63,18 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
...
+kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
-kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
+kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
+
+# {0} = channel name, {1} = error message
+kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel could not be created but this operation will be retried. The error is {1}
```
helyarp, line 74: Change `could not be created` to `cannot be created` or to `was not created`

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -63,12 +63,18 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
...
+kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel but the set up will be retried.
```
helyarp, line 75: Change the second "set up" to "setup" because it is a noun. Thus, change to: `was set up for an incoming reactive messaging channel but the setup will be retried.`

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -63,12 +63,18 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
...
+kafka.create.incoming.retry.CWMRX1009W.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify whether the problem was transient and was resolved by retrying the connector set up.
```
helyarp, line 76: Change ` set up` to `setup` because it is a noun: `the connector setup.`

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -63,12 +63,18 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
 # A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
 kafka.create.incoming.error.CWMRX1007E=CWMRX1007E: A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
 kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
```
jantley-ibm, line 65: There's been additional ID QA review on this PR. For CWMRX1007E.explanation, we need to not say "An error ... occurred". The reviewer suggests that if there are multiple possible errors that could have occurred, we should provide a couple of the likely possible ones.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
jantley-ibm: Per QA ID review, provide a couple of possible or likely errors that might have occurred to cause this error. The explanation should have more information than just a restating of the error message.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -63,12 +63,18 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
...
-kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
+kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
```
jantley-ibm, line 70: Per QA ID review, provide a couple of possible or likely errors that might have occurred to cause this error.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
jantley-ibm: Also, avoid future tense if possible. You can achieve this by stating "when X, the setup is retried." instead of "the setup will be retried."

Azquelt: > provide a couple of possible or likely errors that might have occurred to cause this error

I think I can do this. We're calling a third party API and it's throwing an exception. We don't know all the reasons that this error can occur but we can make a guess at what it might be.

The most important part of this message is informing the user that the error occurred during channel setup and which channel it occurred for. The error message included as placeholder `{1}` is the exception thrown by the API which hopefully says what's gone wrong.

> avoid future tense if possible. You can achieve this by stating "when X, the setup is retried." instead of "the setup will be retried."

Hmm, we emit this message after catching an exception from a third-party API. We wait for a short period and then retry.

I guess I could say "the setup is being retried", even though we're actually going to wait a second or so before we actually retry it? Otherwise I'm not sure how to avoid using the future tense here.

Azquelt: @jantley-ibm I've made these changes, could you take a look at the latest commit on this PR? Thanks.

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -62,13 +62,19 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
 
 # A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
 kafka.create.incoming.error.CWMRX1007E=CWMRX1007E: A Kafka incoming connector for the {0} channel cannot be created. The error is {1}
-kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
-kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
+kafka.create.incoming.error.CWMRX1007E.explanation=An error that occurred when the Kafka connector was set up for an incoming reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
```
jantley-ibm, line 65: change "This may be" to "This might be" 

("may" can indicate permission being granted, which can confuse translators)

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -62,13 +62,19 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
...
+kafka.create.incoming.error.CWMRX1007E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
 
 # A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
 kafka.create.outgoing.error.CWMRX1008E=CWMRX1008E: A Kafka outgoing connector for the {0} channel cannot be created. The error is {1}
-kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging.
-kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, server message.log, and FFDC logs to identify the problem.
+kafka.create.outgoing.error.CWMRX1008E.explanation=An error that occurred when the Kafka connector was set up for an outgoing reactive messaging channel prevents normal operation of MicroProfile Reactive Messaging. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
```
jantley-ibm, line 70: may / might

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/resources/com/ibm/ws/microprofile/reactive/messaging/kafka/resources/ReactiveMessaging.nlsprops
```diff
@@ -62,13 +62,19 @@ kafka.library.not.present.CWMRX1006E.useraction=Package the kafka-clients JAR fi
...
+kafka.create.outgoing.error.CWMRX1008E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
+
+# {0} = channel name, {1} = error message
+kafka.create.incoming.retry.CWMRX1009W=CWMRX1009W: A Kafka incoming connector for the {0} channel cannot be created but this operation is being retried. The error is {1}
+kafka.create.incoming.retry.CWMRX1009W.explanation=An error occurred when the Kafka connector was set up for an incoming reactive messaging channel. This may be caused by a problem with the connector configuration, or by a problem resolving the Kafka broker hostname.
```
jantley-ibm, line 75: may / might

Azquelt:
#run-libby-bot
#build

wraschke:
libby

wraschke:
#libby

Azquelt:
#build

ianpartridge:
> `int retrySeconds = config.getOptionalValue(KafkaConnectorConstants.CREATION_RETRY_SECONDS, Integer.class).orElse(0);`

@Azquelt - what does `0` mean?  Will Liberty not retry at all unless `CREATION_RETRY_SECONDS` is passed, or will it retry indefinitely?

Azquelt:
It won't retry at all.

ianpartridge:
Is there an option for indefinite retry? `MAX_INT`?

Azquelt:
There is no option for indefinite retry and you almost certainly don't want that.

The use case for this option is very limited, it's only useful if the broker hostname isn't resolvable at startup but becomes resolvable after that.

This option also has downsides as it will retry for any configuration failure. (Unfortuantely the Kafka client treats an unresolvable hostname as a problem with your configuration, rather than a possible network issue).

This option really just gets around the very niche use case where you start the broker and the app in kubernetes all at once, and it's only a slight optimisation to avoid your app starting, failing and being restarted.

ianpartridge:
> This option really just gets around the very niche use case where you start the broker and the app in kubernetes all at once, and it's only a slight optimisation to avoid your app starting, failing and being restarted.

This isn't "very niche" though - this is what we expect people to do in a GitOps scenario where they're deploying a complete app (multiple microservices + database + Kafka + whatever else) via a deployment pipeline.  All the resources will be applied to Kubernetes simultaneously and as an eventually-consistent system microservices need to handle things coming up at any point.

Azquelt:
> This isn't "very niche" though - this is what we expect people to do in a GitOps scenario where they're deploying a complete app

You expect to redeploy the broker on every update?

Sounds like this might need more testing, I'm not sure that the kafka client would cope with the IP address of the broker changing either.

If this option in it's current form doesn't work for your use case, I'd like to get it reverted ASAP.

ianpartridge:
> You expect to redeploy the broker on every update?

No I'm talking about the initial deploy of a complete application. Subsequent updates to a microservice (code change etc.) would only redeploy that microservice.

Re: testing, the key test is deploying a scenario to Kubernetes involving an mpReactiveMessaging Open Liberty application that connects to a Kafka that's deployed at the same time but becomes available after Open Liberty has started up.

Azquelt:
> Re: testing, the key test is deploying a scenario to Kubernetes involving an mpReactiveMessaging Open Liberty application that connects to a Kafka that's deployed at the same time but becomes available after Open Liberty has started up.

Ok.

1. That should work anyway. If the broker hostname isn't resolvable when the app container starts, the app should fail to start and the container should be destroyed. I'd then expect Kubernetes to restart it with some kind of backoff.
2. With this change and a short retry applied, (say 30 seconds), you get a little more grace time for the broker hostname to resolve before the app fails to start, which should fix this scenario. If the broker takes longer than that, then you fall back to the container failing and being restarted by Kubernetes.

I'm quite concerned that you were looking to configure an infinite retry. That would result in the app essentially hanging during startup if there was any kind of permanent configuration error in the kafka configuration. This is really a workaround for the fact that the kafka client treats the broker hostname being unresolvable as a configuration error rather than a network error.

ianpartridge:
> That should work anyway. If the broker hostname isn't resolvable when the app container starts, the app should fail to start and the container should be destroyed. I'd then expect Kubernetes to restart it with some kind of backoff.

@matthew-wilson IIRC that's not what you were seeing.. Can you help clarify?

Azquelt:
@ianpartridge was that meant to tag @matthewpwilson ?

ianpartridge:
Oops, yes. Thanks!

matthewpwilson:
> That should work anyway. If the broker hostname isn't resolvable when the app container starts, the app should fail to start and the container should be destroyed. I'd then expect Kubernetes to restart it with some kind of backoff.

No we don't see this - the liveness and readiness checks continue to work even when the app fails due to #11722

Azquelt:
Right, so we have #11722 for that issue (and I see you're chasing it up).

Is this change still desired?

From the [discussion on #10937](https://github.com/OpenLiberty/open-liberty/issues/10937#issuecomment-614829264), I thought that it was but if the caveats described mean that it's too confusing or just not worth using, then I'd rather get it reverted before it's included in a release.

I also note that I marked this as "fixing" #10937 which was probably incorrect as really we need #11722 resolved to fix the problems you were having.

Azquelt:
@matthewpwilson @ianpartridge 

So, what's the verdict, do we want this change or not?

If not I want to get it taken out again.

ianpartridge:
It sounds like we have 2 problems here... one fixed by this PR, and #11722 

Do we need both fixes to resolve the problem I reported in #10937 ?

In the absence of a fix for #11722 is this PR helpful on its own?

If we got a fix for #11722 would that fix #10937 on its own, making this PR unnecessary?

If you can answer those questions @Azquelt it will help us decide the best path forwards.

matthewpwilson:
I think this fix could help without a fix for #11722 - if we can delay a failure long enough for Kafka to come up, we won't need to be restarted.  But probably once #11722 is fixed a retry won't be needed in this scenario.

ianpartridge:
Thanks, in that case as #11722 is waiting for mpHealth 3.0 we should stick with this PR in the meantime.

I would like us to verify that it does improve the situation.  @Azquelt - how can we get an OpenLiberty Docker image that includes this fix, for verification?

Azquelt:
@ianpartridge I'm not too familiar with our docker images but https://hub.docker.com/r/openliberty/daily looks promising.

The other option you have to work around the mpHealth issue is to provide a custom liveness command which checks that the `/health` endpoint response includes the expected checks.

ianpartridge:
Thanks, we'll pick up a nightly Docker image and test that this PR does help us.

PR #11802, opened 2020-04-17
port context bvt bucket to open liberty

njr-11:
Port the bucket com.ibm.ws.context_bvt to Open Liberty.
Open Liberty doesn't have BVT, so it must be converted to a FAT bucket.
There is already a functional test bucket named com.ibm.ws.context_fat, so it needs a different name.  I'm naming it the following because it involves creating a number of custom thread context providers and performs tests using those:
com.ibm.ws.context_fat_customproviders


njr-11:
#build

PR #11881, opened 2020-04-22
Upgrade to SmallRye GraphQL 1.0.1

andymc12:
Upgrading to SmallRye GraphQL 1.0.1 - available in Maven Central.  Includes overrides that should fix issues https://github.com/smallrye/smallrye-graphql/issues/170 and https://github.com/smallrye/smallrye-graphql/issues/172 (which should hopefully be resolved in 1.0.2).


andymc12:
#build

andymc12:
> Should these smallrye classes have copyright headers on them?

I would think so... that is a good question for the SmallRye team...  I'll ask Phillip, and we can add them when they do (or ideally delete the overridden classes entirely now that they have incorporated these changes in the yet-to-published 1.0.2 release).

Thanks for the review!

PR #11906, opened 2020-04-23
Fix EJB IOR profiles

### LINKED ISSUE #11903: Liberty publishes incorrect profiles in IORs
When a remote EJB reference is retrieved from a Liberty server with security configured, e.g. via a JNDI `corbaname:` lookup, the resulting reference has additional profile information. This will only affect an IIOP client when it retries, because the **first** profile is correct.

chackoj:
fixes #11903
#build


chackoj:
Closing as duplicated by #11915

Issue #11156, opened 2020-03-03
Enhance ExpectedFFDC and AllowedFFDC to be applies to certain runs of Repeat Test


olendvcook:
Recently I added a custom repeat tests where FFDCs are only thrown on the 2nd run, but had to have `@AllowedFFDC` on the test method. I want to add a way to specify allowed or expected ffdc for each run, something similar to `@SkipForRepeat`

olendvcook:
Complete

PR #11910, opened 2020-04-23
Add Allowed/ExpectedFFDC annotations for RepeatedTests

olendvcook:
for #11156

Adding a way to specify expected or allowed FFDC per repeat test run

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/annotation/AllowedFFDC.java
+++ b/dev/fattest.simplicity/src/componenttest/annotation/AllowedFFDC.java
@@ -21,6 +21,7 @@ import java.lang.annotation.Retention;
 public @interface AllowedFFDC {
 
     public static final String ALL_FFDC = "ALL_FFDC";
+    public static final String ALL_REPEAT_ACTIONS = "ALL_REPEAT_ACTIONS";
 
     /**
      * A string array fully-qualified FFDC exception class names.
@@ -28,4 +29,9 @@ public @interface AllowedFFDC {
      */
     String[] value() default { ALL_FFDC };
 
+    /**
+     * A string array of RepeatTestAction IDs to check if the allowed list of FFDCs apply to
+     * that specific repeated test run
+     */
+    String[] repeatAction() default ALL_REPEAT_ACTIONS;
 }
--- a/dev/fattest.simplicity/src/componenttest/annotation/ExpectedFFDC.java
+++ b/dev/fattest.simplicity/src/componenttest/annotation/ExpectedFFDC.java
@@ -17,11 +17,22 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 /**
- *
+ * An FFDC produced for the specified exception will be expected; The test will fail if it is not present.
  */
 @Target({ ElementType.METHOD })
 @Retention(RUNTIME)
 public @interface ExpectedFFDC {
+
+    public static final String ALL_REPEAT_ACTIONS = "ALL_REPEAT_ACTIONS";
+
+    /**
+     * A string array fully-qualified FFDC exception class names.
+     */
     String[] value();
 
+    /**
+     * A string array of RepeatTestAction IDs to check if the expected list of FFDCs apply to
+     * that specific repeated test run
+     */
+    String[] repeatAction() default ALL_REPEAT_ACTIONS;
 }
--- a/dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
+++ b/dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
@@ -622,13 +622,26 @@ public class FATRunner extends BlockJUnit4ClassRunner {
     public List<String> getExpectedFFDCAnnotationFromTest(FrameworkMethod m) {
 
         ArrayList<String> annotationListPerClass = new ArrayList<String>();
+
         ExpectedFFDC ffdc = m.getAnnotation(ExpectedFFDC.class);
         if (ffdc != null) {
-            String[] exceptionClasses = ffdc.value();
-            for (String exceptionClass : exceptionClasses) {
-                annotationListPerClass.add(exceptionClass);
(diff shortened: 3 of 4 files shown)


dev/fattest.simplicity/src/componenttest/annotation/ExpectedFFDCs.java
```diff
@@ -0,0 +1,23 @@
...
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.METHOD })
+@Retention(RUNTIME)
+public @interface ExpectedFFDCs {
```
aguibert, line 21: instead of making a public `@ExpectedFFDCs` annotation, lets use a package-protected annotation as an inner class as described in this article:
https://dzone.com/articles/repeatable-annotations-in-java-8-1
That way users won't accidentally import/use `@ExpectedFFDCs`

aguibert: This will also simplify the annotation processing code because you can just use `getAnnotationsByType` everywhere and not have to explicitly check for `@ExpectedFFDC` and `@ExpectedFFDCs` separately

dev/fattest.simplicity/src/componenttest/annotation/AllowedFFDC.java
```diff
@@ -21,11 +21,13 @@
 public @interface AllowedFFDC {
 
     public static final String ALL_FFDC = "ALL_FFDC";
+    public static final String ALL_REPEAT_ACTIONS = "ALLREPEATACTIONS";
```
aguibert, line 24: can we make this constant value separate by underscores so it's a bit more readable?

dev/fattest.simplicity/src/componenttest/annotation/ExpectedFFDC.java
```diff
@@ -22,6 +22,10 @@
 @Target({ ElementType.METHOD })
 @Retention(RUNTIME)
 public @interface ExpectedFFDC {
+
+    public static final String ALL_REPEAT_ACTIONS = "ALLREPEATACTIONS";
```
aguibert, line 26: same comment here about underscores

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -622,46 +622,53 @@ private String getFFDCHeader(RemoteFile ffdcLogFile) throws Exception {
     public List<String> getExpectedFFDCAnnotationFromTest(FrameworkMethod m) {
 
         ArrayList<String> annotationListPerClass = new ArrayList<String>();
-        ExpectedFFDC ffdc = m.getAnnotation(ExpectedFFDC.class);
-        if (ffdc != null) {
-            String[] exceptionClasses = ffdc.value();
-            for (String exceptionClass : exceptionClasses) {
-                annotationListPerClass.add(exceptionClass);
+
+        Set<ExpectedFFDC> ffdcs = new HashSet<ExpectedFFDC>();
```
aguibert, line 626: I don't think a set is necessary here since there will only ever be one `ExpectedFFDC`

dev/fattest.simplicity/src/componenttest/annotation/AllowedFFDC.java
```diff
@@ -21,11 +21,13 @@
...
 
     /**
      * A string array fully-qualified FFDC exception class names.
      * If no value is specified, any FFDCs are allowed.
      */
     String[] value() default { ALL_FFDC };
 
+    String repeatAction() default ALL_REPEAT_ACTIONS;
```
aguibert, line 32: Please add some javadoc that describes how this attr works here and in `ExpectedFFDC`

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -622,46 +622,53 @@ private String getFFDCHeader(RemoteFile ffdcLogFile) throws Exception {
...
+        ffdcs.add(m.getAnnotation(ExpectedFFDC.class));
+        for (ExpectedFFDC ffdc : ffdcs) {
+            if (ffdc != null) {
+                if (ffdc.repeatAction().equals(ExpectedFFDC.ALL_REPEAT_ACTIONS) || RepeatTestFilter.CURRENT_REPEAT_ACTION.equals(ffdc.repeatAction())) {
```
aguibert, line 630: `CURRENT_REPEAT_ACTION` will be null when no `RepeatTests` are defined, so we need to account for null here

dev/fattest.simplicity/src/componenttest/annotation/AllowedFFDC.java
aguibert: Also, should this be an array of String instead of a single `String`?

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -622,46 +622,53 @@ private String getFFDCHeader(RemoteFile ffdcLogFile) throws Exception {
...
-        AllowedFFDC methodFFDC = m.getAnnotation(AllowedFFDC.class);
-        if (methodFFDC != null) {
-            String[] exceptionClasses = methodFFDC.value();
-            for (String exceptionClass : exceptionClasses) {
-                annotationListPerClass.add(exceptionClass);
-            }
-        }
 
+        // Method
+        Set<AllowedFFDC> ffdcs = new HashSet<AllowedFFDC>();
```
aguibert, line 648: Same set of comments here for `AllowedFFDC` as in `ExpectedFFDC`

olendvcook:
#build

olendvcook:
#run-libby-bot

PR #11920, opened 2020-04-23
Issue #11919: Unblock EJB remtoe app start on shutdown

### LINKED ISSUE #11919: Unblock starting remote EJB apps during server shutdown
When applications with remote EJB interfaces start, application start will wait for the necessary ORB support to finish starting before completing application start.  If the server is stopped while the application is waiting to start, the application start may continue waiting for some time, which is unnecessary since the server is stopping.

The start of remote EJB applications should be unblocked on server stop.

tkburroughs:
When EJB remote apps start, the start process waits for the ORB to
start before proceeding. This wait should be discontinued if the
server is then stopped.

Changed to discontinue the wait when EJB container is notified the
server is stopping.

Also, once the ORB starts, fixed an order problem in the container
that will bind all remote interfaces. The latch used for the wait
needs to be cleared first, otherwise the bind process will wait
for 30 seconds, despite the ORB already having started.

fixes #11919


tkburroughs:
#build #spawn.fullfat.buckets=com.ibm.ws.javaee.7.0_fat

PR #11923, opened 2020-04-24
Disable genChecksums and publish to Maven tasks for iFix builds

jennifer-c:
#build


jennifer-c:
Will actually be using #11924 instead - improved version of this.

PR #11928, opened 2020-04-24
port context.bvt.serialization tests to open liberty

njr-11:
Port the test bucket com.ibm.ws.context_bvt.serialization to OpenLiberty.
This test bucket includes scenarios where old versions of context are deserialized and used, and this will be useful coverage to have when adding support for Jakarta EE, which has different execution property constants.  It would be nice to allow any old serialized values to still be recognized (future work item), but these tests will be useful in covering that.


njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
#build
#run-libby-bot

PR #11954, opened 2020-04-27
Update FAT Framework for Jakarta WebServlet

tkburroughs:
For a Jakarta EE 9 only FAT, the FAT client will still be running
with componenttest-1.0, so if the servlet extends FATServlet, it
will extend javax.servlet.http.HttpServlet when loaded by the
client, rather than jakarta.servlet.http.HttpServlet.

TestServletProcessor needs to be updated to look for
javax.servlet.annotation.WebServlet and
jakarta.servlet.annotation.WebServlet
regardless what the servlet implements.

In the running server, jakarta will be used, since
componenttest-2.0 will be used.


tkburroughs:
#build

tkburroughs:
All FAT failures are intermittent UI failures that occur when Selenium Servers are unreachable.  Ready to merge with builds are open (bldcon 3).

PR #11974, opened 2020-04-29
Clean work area cache on app delete and server stop

brenthdaniel:
1) Clean up the workarea cache when an app is deleted
2) On server stop, delete workarea cache content that does not correspond to an existing application pid


brenthdaniel:
#build

brenthdaniel:
artifactory again...
#build

brenthdaniel:
#11897

brenthdaniel:
The build had networking issues again.. Also, the delete on app removal needed to be moved a bit later. 
#build again

brenthdaniel:
The latest commit addresses the case where an application fails to start. We were only removing app monitors during the StopAction, so they were still around for failed applications. Removing the configuration for the failed application would cause us to clear the cache, which lead to an error about removing binaries while the app was still configured. 

#build again

brenthdaniel:
#run-libby-bot

brenthdaniel:
hmm..
#build

PR #12018, opened 2020-05-01
Add capabilities that are not effective

tjwatson:
Some capabilities are informational only, but
they can provide some useful information.
For example, the equinox.module.data capability
is generated by the framework and contains
information such as the bundle's activator
class


tjwatson:
#build

tjwatson:
While implementing this I noticed an Equinox bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=562710

This bug does not impact Liberty directly.  It results in some non-effective caps to show up in the wirings from fragments.  But this does not really impact the behavior of the runtime in Liberty.  Regardless, I have a fix in Equinox for that.

tjwatson:
#build

tjwatson:
Closing stale PR.  Will open a new one to do this work.

PR #12071, opened 2020-05-06
Add the Jakartaee JWS-3.0 Bundle

neuwerk:


neuwerk:
#build

andymc12 (dismissed):
The code change look good to me.  IIRC, there used to be a requirement that new projects/bundles must be included (directly or indirectly) from public (or auto) features.  Maybe that restriction has been loosened - but if not, then this might not pass a personal build.  In that case, you may need to create a JAX-WS 3.0 "noship" feature and have it include this bundle.  Thanks @neuwerk!

neuwerk:
#build

PR #12080, opened 2020-05-06
enable test with jakarta package in web.xml

njr-11:
Enable some tests that have a web.xml deployment descriptor that can refer to a class in a Jakarta pacakge.


njr-11:
#build

PR #12092, opened 2020-05-07
Cleanup features and bundles in kernel feature FAT

brenthdaniel:
The kernel feature FAT isn't cleaning up a few bundles and a system feature. If these are left around it can cause other FAT buckets to fail.


brenthdaniel:
#build

Issue #7956, opened 2019-06-19
API-31: MP GraphQL


andymc12:
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## List of Steps to complete or get approvals / sign-offs for Onboarding to the Liberty release (GM date)
 
Instructions:
 - Do the actions below and mark them complete in the checklist when they are done.
 - Make sure all feature readiness approvers put the appropriate tag on the epic to indicate their approval.
 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## TARGET COMPLETION DATE **Before Development Starts or 8 weeks before Onboarding** 
- [x] POC Design / WAD Review Scheduled (David Chang) or N/A. AHA Feature marked.
- [x] POC Design / WAD Reviewed (Feature Owner) or N/A. AHA Feature marked.
- [x] Complete any To-Do's from the POC Review, and create a new TO-DO in the AHA Feature for the "Design Approval Request" and assign it to Ian Robinson. Leave the date blank. (Feature Owner)
- [x] Design / WAD Approval (Ian Robinson) or N/A. AHA Feature marked.
- [x] No Design / No WAD Approval (Arthur De Magalhaes - cloud / Alasdair Nottingham - server) or N/A. AHA Feature marked.
- [x] AHA Feature has link to Github epic or issue. Github epic has link back to AHA Feature. (Feature Owner or Backlog Subtribe PM).
- [x] SVT Requirements identified. (Epic owner / Feature owner with SVT focal point)
- [x] ID Requirements identified. (Epic owner / Feature owner with ID focal point)
- [x] Create a child task of the epic entitled "FAT Approval Test Summary".  Add and fill in the template as described here: https://github.ibm.com/was-liberty/WS-CD-Open/wiki/Feature-Review-(Feature-Test-Summary-Process)
 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## TARGET COMPLETION DATE **3 weeks before Onboarding** 
- [x] Identify all open source libraries that are changing or are new. Work with Legal Release Services (Cass Tucker or Release PM) to get open source cleared and approved. Or N/A. (Epic Owner).   New or changed open source impacts license and Certificate of Originality.
 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## TARGET COMPLETION DATE ** 3 weeks before Onboarding** 
[...]

PR #12100, opened 2020-05-08
MP GraphQL Authorization

andymc12:
Authorization checks for MP GraphQL.  Works similar to JAX-RS authorization where `@DenyAll`, `@PermitAll`, and `@RolesAllowed` annotations may be used on query/mutation methods (in classes annotated with `@GraphQLApi`).  This must be used in conjunction with the `appSecurity-2.0` or `appSecurity-3.0` feature, and requires configuration via the web container - usually with security constraints in the web.xml - see test case for example.

### DIFF
--- a/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
+++ b/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
@@ -643,6 +643,16 @@
     </dependency>
     <dependency>
       <groupId>io.smallrye</groupId>
+      <artifactId>smallrye-graphql-client</artifactId>
+      <version>1.0.2</version>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye</groupId>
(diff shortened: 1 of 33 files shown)


andymc12:
Part of Epic issue #7956.

#build

andymc12:
One failure is due to a weird test infrastructure issue - the FeatureReplacementAction class rule that is applied to the RolesAuthTest seems to also get applied to all subsequent tests in the bucket - basically adding the `appSecurity-2.0` feature to other servers for other tests.  This results in the AuthorizationFilter being unable to find the ExecutionServlet in the VoidQueryTest (which makes sense - that test verifies that the GraphQL ExecutionServlet is not installed if the app contains invalid queries/mutations). But that test fails because of the additional warning message.

Ideally, the FeatureReplacementAction would only apply to the test case it is added to, but this is simple enough to fix by ignoring/expecting the additional warning in the VoidQueryTest. Now that that is done, new #build

andymc12:
#run-libby-bot

PR #12112, opened 2020-05-08
Issue #12107: Suppress extraneous trace which is filling logs and off…

### LINKED ISSUE #12107: WIM produces a bunch of extraneous trace
WIM produces a ton of extraneous trace which isn't helpful and adds a large amount of time to searches that return many results.

- Mark classes in the com.ibm.ws.securit.wim.model package as `Trivial` to remove entry / exit trace injection. These are mainly just POJO methods and provide little if any value in trace, but they output a TON of it since they are continuously being accessed.
- Mark com.ibm.ws.security.wim.adapter.ldap.LdapAdapter.createEntityFromLdapEntry(...) as `Trivial` to remove entry / exit trace injection. This method is brutal when searches that return large numbers of results are done b/c the rootDO instance continually grows each time it is accessed as we add an entity to it on every call. On the next call we print every entity in the rootDO.

jvanhill:
…ering very little value.

fixes #12107


jvanhill:
#build
#libby

Issue #11732, opened 2020-04-10
@ExpectedFFDC to handle Javax & Jakarta Exceptions


volosied:
Current FatRunner is not designed to handle exceptions from 2 different namespaces.

PR #12121, opened 2020-05-11
Improve FATRunner &  EE9 Package Replacment Helper

volosied:
Improves #11732

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
+++ b/dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
@@ -57,6 +57,7 @@ import componenttest.exception.TopologyException;
 import componenttest.logging.ffdc.IgnoredFFDCs;
 import componenttest.logging.ffdc.IgnoredFFDCs.IgnoredFFDC;
 import componenttest.rules.repeater.EE9PackageReplacementHelper;
+import componenttest.rules.repeater.JakartaEE9Action;
 import componenttest.topology.impl.LibertyServer;
 import componenttest.topology.impl.LibertyServerFactory;
 import componenttest.topology.impl.LibertyServerWrapper;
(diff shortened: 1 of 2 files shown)


dev/fattest.simplicity/src/componenttest/rules/repeater/EE9PackageReplacementHelper.java
```diff
@@ -35,7 +35,8 @@
 
     private static final Logger Log = Logger.getLogger(EE9PackageReplacementHelper.class.getName());
 
-    public EE9PackageReplacementHelper(String rulesPath){
```
aguibert, line 38: you can still keep this old constructor as an option for people, I was just suggesting adding a new ctor with 0-args

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -83,7 +84,7 @@
                                                                       new JavaLevelFilter()
     };
 
-    private static EE9PackageReplacementHelper ee9Helper;
+    private static EE9PackageReplacementHelper ee9Helper = new EE9PackageReplacementHelper();
```
aguibert, line 87: We can't init quite this early, it implies that in order to run any FAT (whether or not an EE9 repeat is used) this properties file will need to be present. Instead, lets just lazy-init the field, but do so using the no-args ctor

dev/fattest.simplicity/src/componenttest/rules/repeater/EE9PackageReplacementHelper.java
volosied: Added it back in.

volosied:
#build

volosied:
#build

volosied:
#build

dev/fattest.simplicity/src/componenttest/rules/repeater/EE9PackageReplacementHelper.java
```diff
@@ -35,7 +35,19 @@
...
+            appProps.load(new FileInputStream(rulesPath));
+            packageRenameRules = (Map) appProps;
+        } catch (Exception e) {
+            Log.warning("Error occured when reading in " + rulesPath);
+            e.printStackTrace();
+        }
+    }
+
+    public EE9PackageReplacementHelper() {
+            String rulesPath = System.getProperty("user.dir") + "/autoFVT-templates/" + "jakarta-renames.properties";
```
aguibert, line 50: since the code here is common with the other constructor, you can just do:
```java
public EE9PackageReplacementHelper() {
  this(System.getProperty("user.dir") + "/autoFVT-templates/" + "jakarta-renames.properties");
}

volosied: Fixed, Thank you :)

volosied:
#run-libby-bot

PR #12142, opened 2020-05-12
Fix unreliable trace that tests depend on

jonhawkes:
Trace that recovery tests depend on was not thread safe.

#build


jonhawkes:
#build

jonhawkes:
#build

PR #12168, opened 2020-05-13
CompletionStageFactory for general consumption by JAX-RS

njr-11:
Adds a CompletionStageFactory for general consumption by JAX-RS, which does not require the concurrent bundle.  If a managed executor is found (which means the concurrent bundle is enabled), this CompletionStageFactory will delegate to it to get the context propagation.  But absent that, it ties into the mechanism provided by Java to create completion stages that run on the Liberty thread pool or another supplied executor.  This mechanism became available in Java 9's CompletableFuture, and so there are limitations if using Java 8.

This pull does not yet switch JAX-RS over to it, which means it also does not yet remove the CompletionStageFactory implementation that was provided by the concurrent bundle.  I'm first sending this to the JAX-RS team and performance team for review, and will let the JAX-RS team switch over their code if they are satisfied with it.


njr-11:
#build

njr-11:
Build has 3 test failures which are preexisting issues in unrelated buckets. This is ready for delivery if the reviewers approve.

dev/com.ibm.ws.threading/src/com/ibm/ws/threading/CompletionStageFactory.java
```diff
@@ -0,0 +1,201 @@
...
+        if (executor instanceof CompletionStageExecutor)
+            return ((CompletionStageExecutor) executor).supplyAsync(supplier);
+
+        executor = executor == null ? globalExecutor : executor;
+        if (JAVA8)
+            return CompletableFuture.supplyAsync(supplier, executor); // at least run the initial stage on the executor
+        else {
+            LibertyCompletableFuture<T> future = new LibertyCompletableFuture<T>(executor);
+            // TODO once Java 8 is dropped, replace reflection with:
+            // TODO future.completeAsync(supplier, executor)
```
andymc12, line 175: Would a multi-release jar work here? Usually, I think MR jars are more work (dual maintenance) than they're worth, but maybe this case might be more beneficial?

njr-11: Yes - good idea - that could be a way to further improve it, but I'll leave that sort thing up to the performance team for future consideration.

njr-11:
#build

njr-11:
@gjdeval Andy McCright reviewed this but asked that you have a look at it as well because it moves some function to Liberty threading.  JAX-RS is currently getting this function from our MicroProfile Context Propagation implementation, but they really shouldn't have a dependency like this, so we are moving the part that they need to a more general purpose service. This pull doesn't actually switch JAX-RS over yet.  It just puts the infrastructure in place for them to do so once ready.  If you would like to review it as well, just add yourself as a reviewer, along with any comments.

gjdeval:
Thanks Nate - no comments.

njr-11:
updated commit to resolve merge conflict
#build
#run-libby-bot

njr-11:
#build

njr-11:
#build

PR #12170, opened 2020-05-13
Avoid processing non-GraphQL apps

andymc12:
Avoid any additional processing of non-GraphQL apps (i.e. apps with no `@GraphQLApi`, `@Query`, or `@Mutation` annotations) when the `mpGraphQL-1.0` feature is enabled.  This is really intended for multi-app servers where only some of the apps use GraphQL APIs.


andymc12:
#build

andymc12:
Build failed with a socket read timeout before building/testing. Retrying with new #build

PR #12180, opened 2020-05-14
Remove AcmeBoulderSimpleTest

ryanesch:
This test was added to test the functionality of the BoulderContainer, however other tests now exist that use BoulderContainer, so we do not need this duplicated ACME testing.


ryanesch:
#build

Issue #12140, opened 2020-05-12
Server stop hangs with NPE after startup with startAfter dependency cycle


gjdeval:
Working in 20006 dev build (wlp-1.0.41.cl200620200508-0300) I deliberately created an app dependency cycle using the startAfter attribute in the webApplication element, like:

* app-A startAfter app-B
* app-B startAfter app-C
* app-C startAfter app-D
* app-D startAfter app-A

When the server was started, the app manager correctly detected and reported the dependency cycle as blocking startup for the involved apps, like:
```
CWWKZ0066E: The following applications cannot start because the value of their startAfter attributes result in a cycle: meetings2 meetings3 meetings4 meetings5 meetings6 meetings7 meetings8 meetings
```
When the server is stopped, shutdown hangs with an NPE, until the stopTimeout interval elapses, like this:
```
[5/12/20 8:20:50:335 EDT] 0000001d com.ibm.ws.kernel.launch.internal.ServerCommandListener      A CWWKE0055I: Server shutdown requested on Tuesday, May 12, 2020 at 8:20 AM. The server multi-cdi-fat-8-apps-serial is shutting down.
[5/12/20 8:20:50:339 EDT] 00000040 com.ibm.ws.runtime.update.internal.RuntimeUpdateManagerImpl  A CWWKE1100I: Waiting for up to 30 seconds for the server to quiesce.
[5/12/20 8:20:50:342 EDT] 0000002b com.ibm.ws.tcpchannel.internal.TCPChannel                    I CWWKO0220I: TCP Channel defaultHttpEndpoint has stopped listening for requests on host localhost  (IPv4: 127.0.0.1) port 9080.
[5/12/20 8:20:51:344 EDT] 00000040 com.ibm.ws.runtime.update.internal.RuntimeUpdateManagerImpl  I CWWKE1101I: Server quiesce complete.
[5/12/20 8:20:51:419 EDT] 00000030 com.ibm.ws.logging.internal.impl.IncidentImpl                I FFDC1015I: An FFDC Incident has been created: "java.lang.NullPointerException com.ibm.ws.app.manager.internal.statemachine.ApplicationStateMachineImpl 900" at ffdc_20.05.12_08.20.51.0.log
[5/12/20 8:20:51:421 EDT] 00000030 SystemErr                                                    R Exception in thread "Default Executor-thread-4"
[5/12/20 8:20:51:422 EDT] 00000030 SystemErr                                                    R java.lang.NullPointerException
[...]

Issue #12139, opened 2020-05-12
Removing startAfter dependency cycle does not start the apps


gjdeval:
Working in 20006 dev build (wlp-1.0.41.cl200620200508-0300) I deliberately created an app dependency cycle using the startAfter attribute in the webApplication element, like:

* app-A startAfter  app-B
* app-B startAfter  app-C
* app-C startAfter  app-D
* app-D startAfter  app-A

When the server was started, the app manager correctly detected and reported the dependency cycle as blocking startup for the involved apps, like:
```
CWWKZ0066E: The following applications cannot start because the value of their startAfter attributes result in a cycle: meetings2 meetings3 meetings4 meetings5 meetings6 meetings7 meetings8 meetings
```
With the server still running, I then edited the server.xml to break the dependency cycle by removing the startAfter attributed in one of the webApplication elements. 

The server recognized the config change and processed it, like:
```
CWWKG0016I: Starting server configuration update.
CWWKG0017I: The server configuration was successfully updated in 0.035 seconds.
```
But the app manager apparently did not reprocess the non-started apps, to discover that the startAfter cycle no longer existed, so the apps were not started.

brenthdaniel:
The strange thing here is that we have a test that explicitly tests this case. This function can work as expected when app updates happen in a certain order, but I'm really surprised we haven't seen a failure for this test yet. 

Anyway, the referenced PR will fix the issue and also will prevent the server from waiting for apps involved in a cycle before reporting that it has started.

PR #12190, opened 2020-05-15
Allow apps to start when a cycle is removed

brenthdaniel:
Applications are not automatically starting when a cycle is removed.


brenthdaniel:
#build

brenthdaniel:
#12139

brenthdaniel:
should also fix #12140

PR #12191, opened 2020-05-15
Improve resource injection trace

tkburroughs:
Since supporting Object Factories are injected into the injection
engine, it can be difficult to tell which component is injecting
an object factory where there is a failure.

Include the name of the object factory in trace and the exception
message when an error occurs in registration.

Also improve trace when processing class annotations.


tkburroughs:
#build

PR #12208, opened 2020-05-15
align autofeatures for jakarta enablement

njr-11:
The jdbc autofeatures were previously renamed to make it possible to provide for differences between jdbc spec levels. Since then, transaction context has followed the same pattern.  This pull updates two autofeatures that I delivered previously to align with this pattern as well, which will make it easier to maintain all of these features together because they will all show up in the same place.


njr-11:
#build

njr-11:
#build

PR #12210, opened 2020-05-15
Do not cache variable lookups from expressions

brenthdaniel:
We are currently caching variables looked up during the evaluation of expressions like servicePidOrFilter despite the fact that they always come from configuration. This can result in cached configuration not being used on restart even though nothing has changed.


brenthdaniel:
#build

brenthdaniel:
#12212

brenthdaniel:
network error in one of the child builds...
#build again

PR #12231, opened 2020-05-18
9214 Switch jpaContainer feature to eeCompatible

dazavala:
Switch jpaContainer feature from javaeeCompatible to eeCompatible


tevans78:
Merging without PB

PR #12250, opened 2020-05-19
Fix for jakarta feature name change

### LINKED ISSUE #12189: Repeat Test doesn't replace some features with Jakarta equivalent
Working on the FAT bucket update to jakarta we noticed that some of the features were not being updated correctly so use the appropriate jakarta equivalent. This is due to the feature name change and how the repeat tests work on changing the feature name as seen [here](https://github.com/OpenLiberty/open-liberty/blob/integration/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java#L374)

A quick example:
**Expected:**
javaee-7.0 -> jakartaee-9.0
**Got:**
javaee-7.0 -> *EMPTY*

We were working on a test which uses the javaee7 and on the jakarta repeat it didnt change the feature because the appropriate name doesn't match. It was removed but was not updated.

We will need to update that with an appropriate name change

isaacrivriv:
fixes #12189

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
+++ b/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
@@ -414,21 +414,44 @@ public class FeatureReplacementAction implements RepeatTestAction {
         String methodName = "getReplacementFeature";
         // Example: servlet-3.1 --> servlet-4.0
         int dashOffset = originalFeature.indexOf('-');
-        if ( dashOffset == -1 ) {
-            Log.info(c, methodName, "Remove feature [ " + originalFeature +  " ]: No '-' was found.");
-            return null;
-        }
+        if ( dashOffset == -1 )
+            throw new IllegalArgumentException("Remove feature [ " + originalFeature +  " ]: No '-' was found.");
         // "servlet-3.1" ==> "servlet"
-        String baseFeature = originalFeature.substring(0, dashOffset);
-        if ( JakartaEE9Action.isActive() ) {
-            String ee9BaseFeature = featuresWithNameChangeOnEE9.get(baseFeature);
-            if ( ee9BaseFeature != null ) {
-                Log.info(c, methodName, "Replace base feature [ " + baseFeature +  " ] with EE9 feature [ " + ee9BaseFeature + " ]");
-                baseFeature = ee9BaseFeature;
+        String baseFeature = originalFeature.substring(0, dashOffset + 1);
+        // "servlet-4.0".contains("servlet-")
+        for ( String replacementFeature : replacementFeatures ) {
+            if ( replacementFeature.contains(baseFeature) )  {
+                Log.info(c, methodName, "Replace feature [ " + originalFeature +  " ] with [ " + replacementFeature + " ]");
+                return replacementFeature;
             }
         }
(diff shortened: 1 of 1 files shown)


isaacrivriv:
#build

dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
```diff
@@ -378,14 +390,54 @@ public void setup() throws Exception {
...
         // Example: servlet-3.1 --> servlet-4.0
-        String featureBasename = originalFeature.substring(0, originalFeature.indexOf('-') + 1); // "servlet-"
-        for (String featureToAdd : featuresToAdd)
-            if (featureToAdd.contains(featureBasename)) // "servlet-4.0".contains("servlet-")
-                return featureToAdd;
+        int dashOffset = originalFeature.indexOf('-');
+        if ( dashOffset == -1 ) {
+            Log.info(c, methodName, "Remove feature [ " + originalFeature +  " ]: No '-' was found.");
+            return null;
+        }
```
aguibert, line 420: we can remove this case, or throw an exception here. All public features should have a version component

dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
```diff
@@ -378,14 +390,54 @@ public void setup() throws Exception {
...
+        // "servlet-3.1" ==> "servlet"
+        String baseFeature = originalFeature.substring(0, dashOffset);
+        if ( JakartaEE9Action.isActive() ) {
+            String ee9BaseFeature = featuresWithNameChangeOnEE9.get(baseFeature);
+            if ( ee9BaseFeature != null ) {
+                Log.info(c, methodName, "Replace base feature [ " + baseFeature +  " ] with EE9 feature [ " + ee9BaseFeature + " ]");
+                baseFeature = ee9BaseFeature;
+            }
+        }
```
aguibert, line 429: Lets move this block after the standard replacement block -- we should prefer standard replacements over the special case replacements generally.

Also, we need to support both directions: `EE7/8 --> EE9` but also `EE9 --> EE7/8`
For this we can simply iterate over `featuresWithNameChangeOnEE9.entrySet()` and check if `replacementFeature` is the key or the value, and if it is, check the opposite entry (value or key respectively) for the replacement.

dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
isaacrivriv: Ok sounds good Ill make the change to throw an exception since it could be possible that manually someone adds a feature and mistakenly leaves out the '-' so that would be a good debug as to why it failed

dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
isaacrivriv: Ok sounds good Ill make the change

isaacrivriv:
#build

isaacrivriv:
#build

isaacrivriv:
#run-libby-bot

PR #12257, opened 2020-05-19
synchronize checkForCycles method

brenthdaniel:
References to _appFromPid should be synchronized, so add synchronization to the checkForCycles method.


brenthdaniel:
#build

PR #12429, opened 2020-05-29
beta of JDBC support for Jakarta

njr-11:
Switch all jdbc-related Jakarta autofeatures (plus transactionContext which appears to have been missed) to beta.

### DIFF
--- a/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
@@ -6,5 +6,5 @@ IBM-Provision-Capability:\
 IBM-Install-Policy: when-satisfied
 -features=com.ibm.websphere.appserver.jakarta.connector-2.0
 -bundles=com.ibm.ws.jca.cm.jakarta
-kind=noship
-edition=full
+kind=beta
+edition=core
--- a/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-jcaSecurity-1.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-jcaSecurity-1.0.feature
@@ -5,5 +5,5 @@ IBM-Provision-Capability:\
  osgi.identity; filter:="(&(type=osgi.subsystem.feature)(osgi.identity=com.ibm.websphere.appserver.jcaSecurity-1.0))"
 IBM-Install-Policy: when-satisfied
 -bundles=com.ibm.ws.security.jca.jakarta
-kind=noship
-edition=full
+kind=beta
+edition=core
--- a/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-jdbc-4.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-jdbc-4.0.feature
@@ -5,5 +5,5 @@ IBM-Provision-Capability:\
  osgi.identity; filter:="(&(type=osgi.subsystem.feature)(osgi.identity=com.ibm.websphere.appserver.jdbc-4.0))"
 IBM-Install-Policy: when-satisfied
 -bundles=com.ibm.ws.jdbc.jakarta
-kind=noship
-edition=full
+kind=beta
+edition=core
--- a/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-jdbc-4.1.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-jdbc-4.1.feature
@@ -7,5 +7,5 @@ IBM-Install-Policy: when-satisfied
 -bundles=\
  com.ibm.ws.jdbc.jakarta,\
  com.ibm.ws.jdbc.4.1.jakarta
-kind=noship
-edition=full
+kind=beta
+edition=core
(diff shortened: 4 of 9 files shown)


njr-11:
#build

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
```diff
@@ -6,5 +6,5 @@ IBM-Provision-Capability:\
 IBM-Install-Policy: when-satisfied
 -features=com.ibm.websphere.appserver.jakarta.connector-2.0
 -bundles=com.ibm.ws.jca.cm.jakarta
-kind=noship
+kind=beta
 edition=full
```
aguibert, line 10: this should be moved to `edition=base`

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
```diff
@@ -6,5 +6,5 @@ IBM-Provision-Capability:\
 IBM-Install-Policy: when-satisfied
 -features=com.ibm.websphere.appserver.jakarta.connector-2.0
 -bundles=com.ibm.ws.jca.cm.jakarta
-kind=noship
+kind=beta
```
aguibert, line 9: auto features cannot be beta due to some quirks with the build process. Instead, move auto features to `kind=ga` (they won't be included in the GA image until the features requiring enablement are also GA)

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
aguibert: same comment applies to all other feature files too (jdbc features should go into `core` and jca features should go into `base)

njr-11: As long as the spec features upon which they depend remain edition=full, these need to remain edition=full.  It might be correct to change those as well, but I'm not going to do it under this pull.  That needs to be a separate discussion.

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
njr-11: That's odd, but I'll switch the pull so that it uses GA for the beta of the autofeatures as you point out.

njr-11:
#build
#run-libby-bot

njr-11:
#build
#run-libby-bot

njr-11:
#build
#run-libby-bot

njr-11:
build has 1 test failure, which is a preexisting issue in an unrelated ui bucket

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
```diff
@@ -4,7 +4,7 @@ IBM-Provision-Capability:\
  osgi.identity; filter:="(&(type=osgi.subsystem.feature)(|(osgi.identity=com.ibm.websphere.appserver.transaction-2.0)))", \
  osgi.identity; filter:="(&(type=osgi.subsystem.feature)(osgi.identity=com.ibm.websphere.appserver.connectionManagement-1.0))"
 IBM-Install-Policy: when-satisfied
--features=com.ibm.websphere.appserver.jakarta.connector-2.0
+-features=com.ibm.websphere.appserver.jakarta.connector.internal-2.0
 -bundles=com.ibm.ws.jca.cm.jakarta
-kind=noship
```
tevans78, line 9: that doesn't say beta!!

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
tevans78: @aguibert that doesn't sound right, i'm sure i've done beta auto features before

tevans78: There is an issue whereby you can't have a beta auto-feature where all the features it depends on are GA ... but that doesn't seem to be the case here. Even if it was, the usual solution is to introduce a dummy/temporary beta public feature and depend on that as well.

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
njr-11: It says noship because you commented on the old version of the file rather than the new one.  When you look at the new version, it says ga, which is due to Andy's request to switch autofeatures from beta to ga, so this can be lumped in to that discussion.

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
njr-11: In this case, the autofeatures depend on a transaction-2.0 feature which is beta.  I'll give this a try (in a separate commit, just in case we need to fall back to the ga approach) and see if it makes it through a build that way.

dev/com.ibm.websphere.appserver.features/visibility/auto/com.ibm.websphere.appserver.jakarta-connectionManagement-1.0.feature
tevans78: lol that was weird, when i wrote that comment, it said ga!

njr-11:
#build
#run-libby-bot
@tevans78 thanks for clarifying the ga vs beta requirement for autofeatures.  I'll be much more comfortable delivering these as beta. I replaced the commits with one that does this, which basically goes back to what I originally had, but also includes Andy's comment to get the jdbc-related features into core edition, and am submitting a build for it now.

njr-11:
#build

PR #12436, opened 2020-05-29
Fix the tx ee9 fat from jdbc-4.3 to 4.2

dazavala:
Change the transactions ee9 FAT to use jdbc-4.2 rather than jdbc-4.3, because jdbc-4.3 requires java11 and Jakarta EE 9 must work with Java 8.


dazavala:
#build

dazavala:
Tests passed.  Merging now.

PR #12454, opened 2020-06-01
Batch remote partition job log fixes 12450

### LINKED ISSUE #12450: Batch: Fixes for remote partition job logs
Logs from Batch job partitions that are run on remote endpoints are not accessible via the REST interface as other job logs are. We need a set of fixes that will ensure these logs can be accessed using the same interface as locally-run jobs.

cgianfra:
Fixes #12450

### DIFF
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/execution/impl/RuntimePartitionExecution.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/execution/impl/RuntimePartitionExecution.java
@@ -167,8 +167,10 @@ public class RuntimePartitionExecution extends RuntimeWorkUnitExecution {
 
     @Override
     public void updateExecutionJobLogDir(String logDirPath) {
-        RemotablePartitionKey key = new RemotablePartitionKey(getTopLevelExecutionId(), getStepName(), getPartitionNumber());
-        getPersistenceManagerService().updatePartitionExecutionLogDir(key, logDirPath);
+        if (isRemoteDispatch) {
+            RemotablePartitionKey key = new RemotablePartitionKey(getTopLevelExecutionId(), getStepName(), getPartitionNumber());
+            getPersistenceManagerService().updateRemotablePartitionLogDir(key, logDirPath);
+        }
     }
 
     /**
--- a/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/services/IPersistenceManagerService.java
+++ b/dev/com.ibm.jbatch.container/src/com/ibm/jbatch/container/services/IPersistenceManagerService.java
@@ -494,7 +494,7 @@ public interface IPersistenceManagerService extends IBatchServiceBase {
      * @param key
      * @param logDirPath
      */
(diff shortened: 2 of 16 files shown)


scottkurz:
@cgianfra do we loop through each execution, and then get each remotable partition for each execution, and then aggregate that all back up the chain?   Where do we make the inter-server REST calls in this loop?   Do we collect all the top-level plus partition job logs at once for a given execution?   Or separately?

Also it looks like there's a new external parm to the REST API right?   We should doc that since I don't know that we have non-external parms right? 

One low-level detail to be careful of is to note that just because we have "remotable" partitions (i.e. dispatched via JMS) doesn't mean we have "remote" partitions (i.e. partitions running in another server).   It's possible to land back where you started and even to do it all in one server.   Just be aware of that (not saying you're not.. but I haven't gotten that far).

cgianfra:
@scottkurz 
> do we loop through each execution, and then get each remotable partition for each execution, and then aggregate that all back up the chain? Where do we make the inter-server REST calls in this loop? Do we collect all the top-level plus partition job logs at once for a given execution? Or separately?

When we're in JobInstances the loop is basically:
- Look at first execution
  - Any remote partitions?
    - Yes: collect all REST URLs for partition hosts, filter out dupes
    - Call each URL with localOnly to get those endpoint logs
- Look at second execution
  - etc etc

Typing that out made me realize if there were two executions, there's a high likelihood they ran on the same remote endpoints in most environments, so it'd be better to grab all the URLs in one pass and then hit them at the very end. I'll start making that change.

> Also it looks like there's a new external parm to the REST API right? We should doc that since I don't know that we have non-external parms right?

Yes, I think we talked about this but I don't recall what we landed on. I guess there may be edge cases where a customer could find use for it, if their redirect and/or proxy request wasn't working? Let's bring this up tomorrow.

> One low-level detail to be careful of is to note that just because we have "remotable" partitions (i.e. dispatched via JMS) doesn't mean we have "remote" partitions (i.e. partitions running in another server). It's possible to land back where you started and even to do it all in one server. Just be aware of that (not saying you're not.. but I haven't gotten that far).

I actually didn't know that. I think we're okay because I added a line to filter out the local REST URL to prevent self-calls anyway. I'll have to take a glance over everything again with that in mind though.

dmbelina:
One nit, the Tr.debug statements probably need a guard, something like:

if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())

cgianfra:
Thanks @dmbelina , I added your change and Scott's to a new commit I just pushed. Going to re-run a PB since Scott's change was functional.

Nothing special to do with tests at this point, once this gets merged I'll open a new PR for the tests.

cgianfra:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_Zr2MQLo2EeqsDL_UN9umug&tab=com.ibm.team.build.web.ui.internal.editors.result.test.TestProvider

cgianfra:
#rub-libby-bot

cgianfra:
#run-libby-bot

PR #12462, opened 2020-06-02
Only destroy the TCCL if we created it

andymc12:
Fixes an issue where bean validation might inadvertantly call `destroyThreadContextClassLoader` when it never created one. This can happen if somebody has already created a TCCL and put it on the current thread.  Then BV would not create a new one, but would call destroy.  This will lead to the reference count being off by one, which could cause the classloading service to uninstall the TCCL's gateway bundle when it is still in use.

### DIFF
--- a/dev/com.ibm.ws.beanvalidation.core/src/com/ibm/ws/beanvalidation/ValidatorFactoryAccessor.java
+++ b/dev/com.ibm.ws.beanvalidation.core/src/com/ibm/ws/beanvalidation/ValidatorFactoryAccessor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2016 IBM Corporation and others.
+ * Copyright (c) 2010, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -23,6 +23,7 @@ import javax.validation.ValidatorFactory;
 import com.ibm.ejs.ras.Tr;
 import com.ibm.ejs.ras.TraceComponent;
 import com.ibm.ejs.util.dopriv.SetContextClassLoaderPrivileged;
+import com.ibm.ws.beanvalidation.AbstractBeanValidation.ClassLoaderTuple;
 import com.ibm.ws.beanvalidation.config.ValidationConfigurationInterface;
 import com.ibm.ws.beanvalidation.service.BeanValidationExtensionHelper;
 import com.ibm.ws.ffdc.FFDCFilter;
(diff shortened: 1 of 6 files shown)


dev/com.ibm.ws.beanvalidation.v20.cdi/src/com/ibm/ws/beanvalidation/v20/cdi/internal/LibertyHibernateValidatorExtension.java
```diff
@@ -109,8 +110,8 @@ private ValidationExtension delegate(String newClassloaderHint) {
                         Tr.debug(tc, "Set Class loader back to " + oldClassLoader);
                     }
                 }
-                if (setClassLoader != null && setClassLoader.wasChanged) {
-                    releaseLoader(tcclClassLoader);
+                if (tuple != null && tuple.wasCreatedViaClassLoadingService) {
```
tjwatson, line 113: To make sure I understand the nature of this fix.  I think the problem could only occur `if (newClassloaderHint !=null)` because in that case the code above would wrapper the class loader in a `Validation20ClassLoader` instance and that would always force a replacement of the TCCL even if the expected TCCL instance was already set.  So that would force us to release the existing (previous) TCCL even though we may not have created it.  Is my understanding correct?

tjwatson:
I think `com.ibm.ws.beanvalidation.v20.ValidatorFactoryBuilderImpl.buildValidatorFactory(ClassLoader, String)` has the same issue

dev/com.ibm.ws.beanvalidation.v20.cdi/src/com/ibm/ws/beanvalidation/v20/cdi/internal/LibertyHibernateValidatorExtension.java
andymc12: Yes, for sure that case, but it's also possible that the TCCL is inadvertently destroyed (without being created) in a case where the current thread's context classloader is set to anything other than `AppClassLoader` or `ThreadContextClassLoader` - so I think it could occur when the thread's context classloader is set to an OSGi loader like "ContextFinder", etc.

andymc12:
#build

andymc12:
Added some more changes to cover some other code flows.  New #build

andymc12:
@tjwatson @brenthdaniel - @nmittles found a couple of other places that needed the same pattern applied (only destroying the TCCL if it was explicitly created), so I did some refactoring so that the `ClassLoaderTuple` is only in one place and reused. When you have time, could you take another look at these changes?  Thanks!

tjwatson:
I did review the changes.  They seem fine to me.

andymc12:
The three failing unit tests (all timeouts) are unrelated to these changes, and passed when re-ran.

dev/com.ibm.ws.beanvalidation.core/src/com/ibm/ws/beanvalidation/ValidatorFactoryAccessor.java
```diff
@@ -240,8 +241,8 @@ public static ValidatorFactory getValidatorFactory(ClassLoader classLoader, bool
                     Tr.debug(tc, "Set Class loader back to " + oldClassLoader);
                 }
             }
-            if (setClassLoader != null && setClassLoader.wasChanged) {
-                AbstractBeanValidation.instance().releaseLoader(classLoader);
+            if (tuple != null && tuple.wasCreatedViaClassLoadingService) {
```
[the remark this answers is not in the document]
andymc12: We'll plan to fix the 1.1 classes in a followup PR - we can remove the extra check in that PR too.  Thanks for the reviews!

andymc12:
#run-libby-bot

Issue #12461, opened 2020-06-02
Remove injected trace from jaxws-2.2 CXF bundles


neuwerk:
When the CXF 2.6.2 IBM code was moved from the RTC repository to GitHub, an extra level of tracing was added by our build tools. This injected trace needs to be disabled by updating all of the com.ibm.ws.org.apache.cxf...2.6.2/bnd.bnd files with the following line:

instrument.classesExcludes: **/*.class

PR #12465, opened 2020-06-03
Disable trace injection in CXF Bundles

neuwerk:


neuwerk:
#build

neuwerk:
#12461

neuwerk:
#run-libby-bot

neuwerk:
Manually verified over 15000 line reduction in trace by comparing the results of a single test case run with the same trace string with and without my fix.

PR #12468, opened 2020-06-03
Changes to allow ssl session timeout configuration to be recognized b…

### LINKED ISSUE #12032: Configuration for sslSessionTimeout is ignored at runtime
This issue was found while investigating SSL handshake behaviour using Liberty 20.0.0.3. We were looking to determine why we never saw full handshakes after the initial handshake even after the default SSL session timeout had elapsed. We only ever saw the existing SSL session resumed. It appears this is because the value for the SSL session timeout is always 24 hours regardless of what is set in the configuration.

The SSLSessionTimeout is specified by the sslSessionTimeout attribute on the sslOptions element. In the metatype the attribute is ibm:type=duration with a default of 8640ms. This value is eventually configured in the SSLChannelData class using getIntProperty. This calls getStringProperty which expects the value to be a String. The value in the metatype is specified as a duration which means it is presented to the runtime as a long rather than a String. This means that getIntPropety gets a null value from the property lookup which causes it to use the default value of 86400 seconds (from the SSLChannelData class) which is 24hours. 

**Steps to recreate**

Configure a Liberty server with an HTTPS endpoint e.g.

<httpEndpoint host="*" httpPort="9080" httpsPort="9043" id="defaultHttpEndpoint"> 
    <sslOptions sslSessionTimeout="5s"/>
</httpEndpoint>

The following trace was useful to help diagnose the issue com.ibm.ws.channel.*=all:com.ibm.ws.channelfw.*=all and the following Java VM argument generated appropriate SSL trace to see what was happening at runtime -Djavax.net.debug=ssl:handshake The Java VM option can be specified on the client, server or both.

Send a request to the server from a client, wait for the session timeout to expire and then send a new request. The SSL trace will show that the session was resumed rather than a new full handshake being performed which shows that the configured timeout is not used. Additionally the server trace will show the following entries which show that the value from the configuration is not being used:

[5/4/20 11:43:26:575 BST] 00000025 id=00000000 com.ibm.ws.channel.ssl.internal.SSLChannelData               3 Property SSLSessionCacheSize using default 100
[5/4/20 11:43:26:575 BST] 00000025 id=00000000 com.ibm.ws.channel.ssl.internal.SSLChannelData               3 Property SSLSessionTimeout using default 86400

**Expected behavior**

The value from the configuration should be used in place of the default value.

At runtime the value is set on the SSLContext in the SSLChannel class. This value is set as an int in seconds and is inconsitent with the current value in the metatype which is specified in milliseconds defaulting to 8640ms. The ibm:type in the metatype should be changed to ibm:type="duration(s)" with an appropriate default so as to be compatible . The default in the code is 86400 seconds or 24 hours which matches the default in various JDKs so this might be a reasonable default. Given the current behaviour of the code this would not lead to a change in behaviour experienced by end users.

Additionally the SSLSessionCacheSize is configurable via the SSLSessionCacheSize property in SSLChannelData but this setting is not exposed via any configuration options (e.g. on the sslOptions element) so only ever uses the default value which is 100. This may be insufficient for production workloads.

**Diagnostic information:**
 - OpenLiberty Version: 20.0.0.3
 - Java Version: 
    java version "1.8.0_231"
    Java(TM) SE Runtime Environment (build 1.8.0_231-b11)
    Java HotSpot(TM) 64-Bit Server VM (build 25.231-b11, mixed mode)

mrsaldana:
fixes #12032

### DIFF
--- a/dev/com.ibm.ws.channel.ssl/resources/OSGI-INF/metatype/metatype.xml
+++ b/dev/com.ibm.ws.channel.ssl/resources/OSGI-INF/metatype/metatype.xml
@@ -37,7 +37,7 @@
             id="service.ranking" required="false" type="Integer" default="0"/>
             
             <AD id="sslSessionTimeout" name="%ssl.sslSessionTimeout" description="%ssl.sslSessionTimeout.desc"
-            required="false" type="String" ibm:type="duration" default="8640ms" />
+            required="false" type="String" ibm:type="duration" default="8640s" />
     </OCD>
     
     <Designate factoryPid="com.ibm.ws.sslchannel.options">
--- a/dev/com.ibm.ws.channel.ssl/src/com/ibm/ws/channel/ssl/internal/SSLChannelData.java
+++ b/dev/com.ibm.ws.channel.ssl/src/com/ibm/ws/channel/ssl/internal/SSLChannelData.java
@@ -12,8 +12,8 @@ package com.ibm.ws.channel.ssl.internal;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Properties;
 import java.util.Map.Entry;
+import java.util.Properties;
 
 import com.ibm.websphere.channelfw.ChannelData;
 import com.ibm.websphere.channelfw.FlowType;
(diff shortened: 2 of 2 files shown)


mrsaldana:
#build

mrsaldana:
#run-libby-bot

Issue #8048, opened 2019-06-26
Unable to write multipart data in Jax-Rs


rosczja:
I cannot find any way to write a multipart/mixed or multipart/form-data response with jax-rs. If I use a java.util.Map with DataSource entities I reach the writeTo method of the IBMMultipartProvider which in turn calls the writeTo method of the Apache CXF MultipartProvider. Then a NullPointer is raised because the MessageContext is not set.

I would like two things:

- Provide API tro create IMultiPartBody and / or IAttachment classes
- set the message context when constructing the Apache CXF MultiPartProvider

WhiteCat22:
I'm sorry if this is an obvious question, but are you adding 
```@Produces("multipart/form-data")```
to your resource method?

**Comment by andymc12** [comment]
Hi @rosczja - do you have a sample app that reproduces the problem that we could view?  Also, could you provide the log files that show the NPE and the version of Liberty in use?  Thanks!

**Comment by rosczja** [comment]
@WhiteCat22 , I think adding this annotation will not help with the nullpointer.

Here is the exception:
```
java.lang.NullPointerException: null
	at org.apache.cxf.jaxrs.provider.MultipartProvider.writeTo(MultipartProvider.java:265) ~[com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2_1.0.22.jar:3.2.6]
	at com.ibm.ws.jaxrs20.providers.multipart.IBMMultipartProvider.writeTo(IBMMultipartProvider.java:164) ~[com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2_1.0.22.jar:3.2.6]
	at org.apache.cxf.jaxrs.utils.JAXRSUtils$3.run(JAXRSUtils.java:1424) ~[com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2_1.0.22.jar:3.2.6]
	at org.apache.cxf.jaxrs.utils.JAXRSUtils$3.run(JAXRSUtils.java:1421) ~[com.ibm.ws.org.apache.cxf.cxf.rt.frontend.jaxrs.3.2_1.0.22.jar:3.2.6]
	at java.security.AccessController.doPrivileged(Native Method) ~[?:1.8.0_172]
[...]

PR #12488, opened 2020-06-04
set the context in IBMMultipartProvider.writeTo() for jaxrs multipart support

WhiteCat22:
https://github.com/OpenLiberty/open-liberty/issues/8048


WhiteCat22:
User confirmed that the change fixed their NPE

#build

PR #12493, opened 2020-06-04
Update invocations fallbackUsed tag to match spec

Azquelt:
In a recent spec update, this tag was updated
 * Old: `fallbackUsed = {true|false|none}`
 * New: `fallback = {applied|notApplied|notDefined}`


Azquelt:
#build (last build failed due to connectivity issues)

PR #12502, opened 2020-06-04
Ensure persistent executor uses the proper precedence for Jakarta/Java EE constants

njr-11:
And approach to hard code the precedence for Jakarta/Java EE constants doesn't work if the transformer converts them.  This pull switches to a different approach, which is already used by transaction context code.


njr-11:
#build

Issue #12393, opened 2020-05-27
acmeCA-2.0: Add test with the keyStore updateTrigger disabled


kristip17:
https://www.ibm.com/support/knowledgecenter/SSEQTP_liberty/com.ibm.websphere.liberty.autogen.base.doc/ae/rwlp_config_keyStore.html

Set the `updateTrigger` to disabled and make sure the ACME CA certificate is still updated.

For #9017

PR #12503, opened 2020-06-04
Issue 12328: Add new Acme Cert Checker tests

### LINKED ISSUE #12328: acmeCA-2.0: Additional certChecker tests
Follow up to #10519

- [x] Enable/disable the cert checker during runtime (set certCheckerSchedule to 0), make sure it stops/starts appropriately
- [x] Call renew while certchecker is going -- make sure extra ffdc/W/E don't pop up
- [ ] ETA: Not doing -- Cause Boulder failure during test (adding a bad DNS record seems to work), revoke the cert, let scheduler fail, fix DNS, check that we renew cert that's revoked (probably copy `AcmeValidityAndRenewTest.autoRenewWithErrorAndRecovery` since we'll need to run on Boulder/AcmeRevocation test to do the revoke).  --- Couldn't fine a way to temporarily disable Boulder without also messing up the isRevoke check. Did some manual testing.

#9017

kristip17:
Fixes #12328 and #12393

Added/updated 3 tests described in linked Issues

Updated the BoulderContainer again -- undid the changes to bluenet as they caused and NPE on retry.

Code change: I moved where we log the certchecker is disabled message so it would applied either to the certcheckerschedule being set to 0 or the renew and revoke both being disabled.

### DIFF
--- a/dev/com.ibm.ws.security.acme_fat/fat/src/com/ibm/ws/security/acme/fat/AcmeValidityAndRenewTest.java
+++ b/dev/com.ibm.ws.security.acme_fat/fat/src/com/ibm/ws/security/acme/fat/AcmeValidityAndRenewTest.java
@@ -23,6 +23,15 @@ import static org.junit.Assert.assertTrue;
 import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 
+import javax.xml.bind.DatatypeConverter;
+
+import org.apache.http.Header;
+import org.apache.http.StatusLine;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.StringEntity;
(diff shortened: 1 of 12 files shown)


kristip17:
#run-libby-bot

kristip17:
#build

dev/com.ibm.ws.security.acme_fat/fat/src/com/ibm/ws/security/acme/fat/AcmeValidityAndRenewTest.java
```diff
@@ -667,4 +908,51 @@ public void autoRenewWithErrorAndRecovery() throws Exception {
 		}
 	}
 
+	/**
```
jvanhill, line 911: Isn't this method already somewhere else? I figured we did a renew in AcmeCaRestHandlerTest? Can we use that method, or are they different functionality-wise?

PR #12525, opened 2020-06-08
Remove "other" from gRPC metatype 

wtlucy:
The `io.openliberty.grpc.1.0.internal` and `io.openliberty.grpc.1.0.internal.client` bundles have a metatype property named `other`, which isn't actually used and should be removed


wtlucy:
#build

PR #12538, opened 2020-06-08
Issue 12536: Remove extra messages check in Acme rest test

### LINKED ISSUE #12536: acmeCA-2.0: Remove extra CWWKO0219I checks
On the AcmeCARestHandler test, we an `CWWKO0219I` check that we don't need -- depending on the order the rests runs in, we can hit an error if the marks are reset in the logs. We only need it at the start of the test.

For #9017

kristip17:
Fixes #12536 

Remove extra `AcmeFatUtils.waitForSslEndpoint(server);` as it can cause an error if a test that resets the log marks runs first.


kristip17:
#build

kristip17:
Single acme_fat failure is our ongoing problem with running several Boulder tests in parallel. Test in question ran successfully x10.

PR #12539, opened 2020-06-08
12537 - NPE in flushHeaders

loriadi:


loriadi:
#build

loriadi:
The two build errors were for other unrelated components.

PR #12549, opened 2020-06-09
Removed EE9 from description

isaacrivriv:
Subspecs are the same for each EE version so having EE9 in the description is unnecessary


isaacrivriv:
#build

PR #12552, opened 2020-06-09
test precedence of Jakarta transaction context

njr-11:
When Java EE and Jakarta constant values are both specified, give precedence to the enabled spec.


njr-11:
#build

PR #12561, opened 2020-06-10
Rename Jakarta API features and move concurrent 2.0 API to lib

tevans78:
#build

Note about the 2nd commit:
Because the implementation is shared between both concurrent 1.0 and 2.0, the 2.0 API bundle is already being shipped as GA with the 1.0 feature. This was a little premature because Jakarta have not released the API yet,

We can't remove the 2.0 API bundle without drastically reworking the implmentation. However, we don't want to expose it to users just yet either. No one should be using the API yet so moving it out of the dev folder and into lib is safe enough.

When the 2.0 feature moves to beta then the API bundle should be moved back from lib to dev/api/spec.


tevans78:
#build
#run-libby-both

tevans78:
#build

tevans78:
PB had some failure but unrelated to this change. Merging.
#libby-bot

PR #12577, opened 2020-06-11
Issue 12576: Improve Boulder retries

### LINKED ISSUE #12576: acmeCA-2.0: Retry on Boulder creation failures
Continuing to see problems when fats run in parallel and create Boulder containers.

For #9017

kristip17:
Fixes #12576 

- Added an excessive number of retires on starting the DB container for Boulder.
- Set the `withStartupAttempts` on the DB container to 1, otherwise the retries just get a `no networkMode` type error.
- We have to recreate the network or we get a `no network Mode` type error
- When starting actual Boulder, on my local box, I get an intermittent `NoClassDefFoundError ` so I gave that less sleep time.
- I also move `renewCertificate` to AcmeFatUtils so it's a general helper method.


kristip17:
#build

kristip17:
This PR didn't pan out.

PR #12595, opened 2020-06-12
Issue #12594: Add additional server start debug

### LINKED ISSUE #12594: Add additional server startup debug
Server start returns rc=22 in several locations for different scenarios.  Need to add some debug code that will be written to the start.log if the server fails to launch.  Can be used to narrow down which location of the code returned the rc=22.

hlhoots:
#build 
fixes #12594


hlhoots:
Build had a few errors, but none look related to the debug code which was added.  Merging.

PR #12505, opened 2020-06-04
Issue 12451: Add cloning to WSKeyStore

kristip17:
Fixes #12451

Issue #9017, opened 2019-09-17
SEC 27 - Support for ACME certificate management and LetsEncrypt


atosak:
LetsEncrypt has become the most popular free CA signing site. There are now over 100 million certs signed by LetsEncrypt in the wild today. This makes it one of the largest CAs. New cloud native applications/microservices would be served well by adding support for the underlying protocol that LetsEncrypt uses - called ACME (Automated Certificate Management Environment) which is fast becoming the defacto standard. With this feature we plan to add ACME support to Open Liberty.

**Link to WAD/UFO**
https://ibm.box.com/s/y2cjatj56o00rlrx34ed4omr6xgxb0hu

https://ibm.ent.box.com/file/667631862611

**Direct Link to Feature Checklist (gets hidden in the comments)**
https://github.com/OpenLiberty/open-liberty/issues/9017#issuecomment-571420951

 **Issue Queries:**
[High priority open issues](https://github.com/OpenLiberty/open-liberty/issues?q=is%3Aopen+type%3Aissue+label%3Apriority%2Fhigh+acmeCA-2.0+in%3Atitle+) (Min viable product for a beta)
[Medium priority open issues](https://github.com/OpenLiberty/open-liberty/issues?q=is%3Aopen+type%3Aissue+label%3Apriority%2Fmedium+acmeCA-2.0+in%3Atitle+)
[Unlabeled open issuse](https://github.com/OpenLiberty/open-liberty/issues?q=is%3Aopen+type%3Aissue+acmeCA-2.0+in%3Atitle++-label%3Apriority%2Fmedium++-label%3Apriority%2Fhigh)
[All open issues](https://github.com/OpenLiberty/open-liberty/issues?q=is%3Aopen+type%3Aissue+acmeCA-2.0+in%3Atitle)

**Shortcut links**
Doc/Blog prep issue is #9869
Doc request is here https://github.com/OpenLiberty/docs/issues/1699
Feature Test Summary is #9850
SVT test prep is [here](https://github.ibm.com/was-liberty/WS-CD-Open/issues/20744)
STE in progress is https://github.ibm.com/kristip/site/blob/patch-1/_STE/WAS/acmeCA-2.0
Published STE is https://pages.github.ibm.com/WASL3/site/STE/WAS/acmeCA-2.0/
GA Blog issue https://github.com/OpenLiberty/open-liberty/issues/12860

[...]

PR #12626, opened 2020-06-16
Revert "Issue 12451: Add cloning to WSKeyStore"

kristip17:
Reverts OpenLiberty/open-liberty#12505

Due to a performance hit, we are reverting this change and will re-evaulate.

For #9017


kristip17:
#build

PR #12629, opened 2020-06-16
Use FATRunner with app manager tests

brenthdaniel:
FULL tests are running in LITE mode because the test classes are not annotated with `@RunWith(FATRunner)`


brenthdaniel:
#build

PR #12558, opened 2020-06-10
correct message id

benjamin-confino:

PR #12634, opened 2020-06-16
Revert "Revert "correct message id""

wraschke:
Reverts OpenLiberty/open-liberty#12631


wraschke:
I'm reverting because unbeknownst to me when I first reverted #12558, another change was merged under #12619. This isn't good practice. However, given that the build will again be broken with 12631 in it, I am reverting it.

Issue #12397, opened 2020-05-28
server package minify complains of illegal reflective access


gcharters:
**Describe the bug**

Server package minify outputs the following warnings

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.ibm.ws.kernel.service.location.internal.Activator (file:/Users/charters/Downloads/wlp/lib/com.ibm.ws.kernel.service_1.3.40.jar) to field javax.naming.spi.NamingManager.initctx_factory_builder
WARNING: Please consider reporting this to the maintainers of com.ibm.ws.kernel.service.location.internal.Activator
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

**Steps to Reproduce**
Download and unzip Liberty (e.g. 20.0.0.5)
Create a server 'bin/server create'
Package the server 'bin/server package --include=minify

**Expected behavior**
I don't expect to see the warnings

**Diagnostic information:**
 - OpenLiberty Version: 20.0.0.5
 - Java Version: java -version
openjdk version "11.0.6" 2020-01-14
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.6+10)
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.18.1, JRE 11 Mac OS X amd64-64-Bit Compressed References 20200122_450 (JIT enabled, AOT enabled)
OpenJ9   - 51a5857d2
OMR      - 7a1b0239a
JCL      - da35e0c380 based on jdk-11.0.6+10)

 - the default server xml from 'bin/sever create' will do.

**Additional context**

hlhoots:
There were some script changes made when we moved to Java 9 such that some additional java options (wlp\lib\platform\java\java9.options) are applied to removed these warnings.  These changes were not applied to the server package (with --include=minify - which starts the server to build the smaller image) command.

hlhoots:
Should now be resolved.

PR #12716, opened 2020-06-19
Issue #12397: Fix warnings in server package

hlhoots:


hlhoots:
#build  
fixes #12397

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server.bat
```diff
@@ -341,15 +341,17 @@ goto:eof
...
 
 :package
-  call:serverEnv
+  call:serverEnvAndJVMOptions
+  if not %RC% == 0 goto:eof
   call:serverExists true
   if %RC% == 2 goto:eof
 
-  !JAVA_CMD_QUOTED! !JAVA_PARAMS_QUOTED! --batch-file=--package !PARAMS_QUOTED!
+  !JAVA_CMD_QUOTED! !JAVA_AGENT_QUOTED! !JVM_OPTIONS! !JAVA_PARAMS_QUOTED! --batch-file=--package !PARAMS_QUOTED!
```
brenthdaniel, line 349: Is JAVA_AGENT really needed here? I would think that would only be necessary for a server

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/PackageProcessor.java
```diff
@@ -170,6 +172,14 @@ protected File buildManifestForIncludeHasRunnable(File installationManifest) thr
         mf.getMainAttributes().putValue("Main-Class", "wlp.lib.extract.SelfExtractRun");
         mf.getMainAttributes().putValue("Server-Name", processName);
 
+        // For Java 9, we need to apply the /wlp/lib/platform/java/java9.options to the manifest.
+        if (!System.getProperty("java.specification.version").startsWith("1.")) {
```
brenthdaniel, line 176: You might want to check for null here.. java.specification.version should always be there, but probably best to be safe

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server.bat
hlhoots: Whenever I run with Java 8 (without the JAVA_AGENT_QUOTED in the cmd), I get [WARNING ] TRAS0032W: Runtime injection of detailed method trace is not available. Adding the JAVA_AGENT_QUOTED seems to resolve the issue.

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/PackageProcessor.java
hlhoots: I will add that null check.

hlhoots:
windows personal build here: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_aEj3YLJYEeqCjdhsjSafyw

hlhoots:
eu.artifactory timeout on linux build.  

#build

hlhoots:
windows build had 100% success on the com.ibm.ws.kernel.boot_fat tests as did the personal build.

hlhoots:
#build

hlhoots:
Both linux and windows build are good at this point.  Holding on delivery until after week of july 4th.  Will rebase and re-run builds at that time.

hlhoots:
#run-libby-bot

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/PackageProcessor.java
```diff
@@ -170,6 +172,13 @@ protected File buildManifestForIncludeHasRunnable(File installationManifest) thr
         mf.getMainAttributes().putValue("Main-Class", "wlp.lib.extract.SelfExtractRun");
         mf.getMainAttributes().putValue("Server-Name", processName);
 
+        // For Java 9, we need to apply the /wlp/lib/platform/java/java9.options to the manifest.
```
aguibert, line 175: the args in java9.options are things we need to run all features available in liberty, but it is likely that the SelfExractRun just needs a subset of these because all it's doing is launching a separate JVM to do the extract and then it exits (I think).

So instead of copying things over from java9.options to the manifest at runtime, can we just add the flags necessary to do a clean SelfExtract to the jar's manifest directly?

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/PackageProcessor.java
```diff
@@ -788,4 +797,39 @@ private boolean java2SecurityEnabled() {
     private boolean isArchiveJar() {
         return packageFile.getName().endsWith(".jar");
     }
+
+    // Reads the java9.options file
+    private HashMap<String, String> readJava9Options() throws IOException {
+        HashMap<String, String> hm = new HashMap<String, String>();
+        StringBuffer exports = new StringBuffer();
+        StringBuffer opens = new StringBuffer();
+        BufferedReader r = new BufferedReader(new FileReader(installRoot.getAbsolutePath() + File.separator + "lib" + File.separator + "platform" + File.separator + "java"
```
aguibert, line 806: Perhaps we should make this a bit more tolerant and just log a warning if the java9.options cannot be found/read? Would be bad if server start blew up on something that was just trying to avoid a warning message.

hlhoots: I decided to leave as-is.  It will result in: 

CWWKE0056E: Unable to zip up directory due to an IO exception java.io.FileNotFoundException: ${wlp.install.root}\lib\platform\java\java9.options (The system cannot find the file specified).
Server test1 package failed. Check server logs for details.

Which in my mind tells that their install is not correct when running java 9.

Issue #12683, opened 2020-06-18
Re-balance the FaultTolerance FAT buckets


tevans78:
We would like FAT buckets to execute in 5 minutes or less in LITE mode. For example, on average;

com.ibm.ws.microprofile.faulttolerance.cdi_fat	        00:24:45
com.ibm.ws.microprofile.faulttolerance.2.1_fat_tck	00:17:26
com.ibm.ws.microprofile.faulttolerance.2.0_fat_tck	00:15:49
com.ibm.ws.microprofile.faulttolerance.1.1_fat_tck	00:13:33
com.ibm.ws.microprofile.faulttolerance_fat_tck	00:07:01

These buckets should be rebalanced to run more tests in FULL mode or to split them across multiple buckets.

Azquelt:
This work is done. For the FAT tests it was less of a rebalance and more of a refactoring to speed up executions.

FAT tests: ~25 minutes -> 3 minutes
2.1 TCK: ~17 minutes -> 9 minutes (other TCK projects saw comparable runtime reductions)

PR #12730, opened 2020-06-22
Speed up running Fault Tolerance FAT in LITE mode

### LINKED ISSUE #12751: FT 2.0+ Interceptor doesn't destroy dependent objects
**Describe the bug**

The FT 2.0+ CDI Interceptor Bean class doesn't call the CDI API correctly to create and destroy the interceptor instance which results in dependent beans not being destroyed.

This results in `close()` not being called on the Executors.

However, none of the FT 2.0 Executors actually do anything in their `close()` methods, so there's no direct impact in our code, but we do need to check whether not correctly destroying the dependent beans causes a memory leak.

Azquelt:
For #12683

Reduces runtime on my laptop from 12m to 4m. Average runtime in the build is currently 24m, so it's hard to gauge what the change will be there. **Update:** time in personal build was ~7m50

The test refactoring to avoid restarting the server as much should also improve runtime in FULL mode.

Several techniques are used to reduce the bucket runtime:
* Refactor the basic CDI Tests to have one test class with multiple test
  servlets to avoid restarting the server as much as possible
* Add a filter so that we can use RepeatTests on our new single test
  class but only run a subset of tests for older fault tolerance feature
  versions.
* Move some edge case and integration tests from LITE to FULL mode
  * Multi-module EAR tests
  * Transaction tests
  * AsyncReturnNullTest
  * Interceptor ordering tests

Also fixes #12751

Reducing the overall runtime made it much more apparent that the `testExecutorsClose` test was taking way too long in the build. It turns out that this test was not correctly asserting so it wasn't failing when it should be. This in turn revealed a bug in the interceptor cleanup for FT 2.0+.


Azquelt:
#build

Azquelt:
#build
#run-libby-bot

Azquelt:
#build
#run-libby-bot

Azquelt:
The failures in the build don't appear to be related to this change.

PR #12734, opened 2020-06-22
test precedence for longrunning hint

njr-11:
Test to verify that the LONGRUNNING_HINT execution property from Jakarta Concurrency takes precedence over the one from Java EE Concurrency if both are specified and Jakarta Concurrency is enabled (concurrent-2.0) rather than Java EE Concurrency (concurrent-1.0).


njr-11:
#build

PR #12739, opened 2020-06-22
adding additional json handling

### LINKED ISSUE #12616: modification for featureUtility additional json handling
Change featureUtility to handle additional jsons extracted from the featureUtility properties file

hermankailey:
fixes #12616


hermankailey:
#build

Azquelt:
We think this change is somehow causing `FeaturesToMavenArtifactTest` to consume around 15GB of disk space during the build which is causing later tests to fail because they run out of disk space.

PR #12750, opened 2020-06-23
Remove duplicate tests and move some tests to FULL mode

brenthdaniel:
The app manager FAT is taking too long to run. We have some duplicate tests that can be removed, and some error scenarios can be moved to FULL mode.


brenthdaniel:
#build

PR #12773, opened 2020-06-24
test precedence for identity name

njr-11:
Add test case to cover that ManagedTask.IDENTITY_NAME from Jakarta or Java EE is given precedence based on whether the concurrent-2.0 (Jakarta Concurrency) or concurrent-1.0 (Java EE Concurrency) feature is enabled.


njr-11:
#build

PR #12783, opened 2020-06-25
Orphaned transaction changes

dhaggart:
Initial take on changes to orphan transaction handling including additional diagnostics.


dhaggart:
#run-libby-bot

dhaggart:
#build

dhaggart:
Andrew, since TransactionReference will still try to complete Inactive transactions, I thought it best to only avoid the FFDC if/when there was no reserved transaction space.  As you say, if we have Inactive with a non-zero reservation we have a more serious issue and probably should bring things down.

andrea-pichler:
No L2 message review required for translations

AndrewDJBanks:
Good point. Yes things are better as they are.

Issue #12746, opened 2020-06-22
Consolidate the Resteasy bundles for the jaxrs-3.0 feature


WhiteCat22:
Resteasy doesn't support OSGI to the extent that CXF does and also has many more modules than CXF. Therefore, we are going to condense the multiple Resteasy modules into a single bundle.

PR #12785, opened 2020-06-25
update the resteasy.common bundle to handle the javaee apis removed from the jdk

WhiteCat22:
I missed this piece when combing the resteasy bundles.
#12746


WhiteCat22:
#build

PR #12788, opened 2020-06-25
Add containsKey to reduce chances of hitting sync blocks

dettmoney:
Throughput profiles showing lock contention during WebContainer.addToCache. If we add a non-locking containsKey call before the putIfAbsent, we eliminate the contention on the ConcurrentMap.


dettmoney:
#build

dettmoney:
#build

Issue #12815, opened 2020-06-29
Delay application startup until configured mechanisms are in place


chackoj:
While addressing #11151 it became evident that there was no way to delay application startup _generically_ for services on which applications might depend.

This problem has historically been dealt with by outputting log messages when services were available and explicitly waiting for all the relevant log messages within any test that used those services. This approach is flawed because it allows applications to be available before they can work correctly. 

It may be reasonable to tell an administrator to start a server and wait for the appropriate log messages, but it is quite unreasonable to expect automated admin tasks to wait for a set of log messages that are not easily discoverable and subject to change with changes to server configuration.

We propose a generic mechanism that would allow a configured feature to declare and eventually supply a number of DS components of type `ApplicationPrereq`. Each application handler would wait for the number of declared instances of `ApplicationPrereq` to become available before starting up.

This would be achieved using the `….cardinality.minimum` service property [defined in the OSGi spec](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-minimum.cardinality.property) along with OpenLiberty's reference counting metatype extensions.

chackoj:
@brenthdaniel notes that while the described approach is both viable and an improvement to the status quo, it leaves a timing window in which application handlers might be started before config processing has progressed far enough to discover all the declared `ApplicationPrereq` services.

In this case, once the relevant config is discovered, the application handlers would be destroyed and re-created. This is undesirable but still eventually correct. Ideally, we should find a solution that does not have such a window, but perhaps this proposal is only a step on the journey towards that ideal.

Here is a description of the timing window problem as I understand it:

- The WAR application handler is started because no `ApplicationPrereqs` are known to have been declared yet.
- Configuration for the messaging feature and its bundles is processed, and an `ApplicationPrereq` is discovered.
- The WAR application handler configuration is re-processed, and its `ApplicationPrereq` service reference now has a minimum cardinality of 1.
- The WAR application handler is stopped because it does not meet the service requirements.
- A new WAR application handler is started with the declared service reference.

This timing window is minimal, and I do not think it would progress to the application being started up before the handler is stopped and replaced.

@brenthdaniel: please correct me if I have misunderstood/misrepresented anything

tjwatson:
[...]

PR #11152, opened 2020-03-03
Make JsAdminServiceImpl , JsMainAdminServiceImpl , OSGI Services

AndrewDJBanks:
MakeJsAdminServiceImpl , JsMainAdminServiceImpl ,  RuntimesecurityServiceImpl into OSGI Services.

PR #12816, opened 2020-06-29
Create application start barrier

AndrewDJBanks:
See issue #12815


AndrewDJBanks:
#build

AndrewDJBanks:
#build

AndrewDJBanks:
Note that: 
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_bpnpsLn4EeqsDL_UN9umug
contains an example of why the function is necessary. com.ibm.ws.jdbc.fat.db2.JDBCKerberosTest has run before  LTPA2 initialisation has completed.

dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/lifecycle/ApplicationStartBarrierImpl.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Reference(cardinality = ReferenceCardinality.MULTIPLE)
+    void setApplicationPrereq(ApplicationPrereq applicationPrereq) {
+    }
+
+    /**
+     * Non configurable immediate component to track prereqs as they appear.
+     */
+    @Component(configurationPolicy = ConfigurationPolicy.IGNORE, immediate = true)
+    public static class PrereqWatcher {
```
tjwatson, line 53: It is not clear to me how this component does anything other than provide some additional trace points.

dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/lifecycle/ApplicationStartBarrierImpl.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            Arrays.sort(prereqClassNames);
+            return Arrays.toString(prereqClassNames);
+        }
+
+        @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
+        void setApplicationPrereq(ApplicationPrereq applicationPrereq) {
+            realisedPrereqs.add(applicationPrereq.getClass().getName());
+            Collections.sort(realisedPrereqs);
```
tjwatson, line 94: Why always sort even when trace is not enabled.  I only see realisedPrereqs getting used within trace blocks.

dev/com.ibm.ws.app.manager.lifecycle/src/com/ibm/wsspi/application/lifecycle/ApplicationPrereq.java
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.wsspi.application.lifecycle;
```
tevans78, line 11: @tjwatson makes a good point, you can't expose this as a SPI straight away. You need a Feature and UFO to expose a new SPI. Please change the package name.

tjwatson:
If we go with the approach to avoid calling application handlers until the ApplicationStartBarrier is available we could place the following reference on `com.ibm.ws.app.manager.internal.ApplicationConfigurator` and remove that reference from all Application handlers.

```
    @Reference
    private void setApplicationStartBarrier(ApplicationStartBarrier applicationStartBarrier) {
    }
```

We need to understand how this solution behaves when you dynamically enable a feature that adds a new `ApplicationPrereq`.  Does that take down `ApplicationStartBarrier` and do we gracefully shutdown existing applications?  Then successfully restart when the `ApplicationPrereq` is available again?

What if we unconfigure a feature providing a `ApplicationPrereq` and at the same time configure a different feature that provides another `ApplicationPrereq` such that the overall count of configured `ApplicationPrereq` remains the same?

AndrewDJBanks:
#build

AndrewDJBanks:
#build

AndrewDJBanks:
#build

AndrewDJBanks:
#build

dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/lifecycle/ApplicationStartBarrierImpl.java
chackoj: Very good question. Would it be acceptable to use a sorted set instead, or should we simply defer sorting until we know it needs to be used?

AndrewDJBanks:
#build

dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/lifecycle/ApplicationStartBarrierImpl.java
tjwatson: I suppose a sorted set would be the most simple.  The number of prereqs likely is not high so sorting as you add likely should not show any measurable performance difference.

AndrewDJBanks:
#build

AndrewDJBanks:
#run-libby-bot

AndrewDJBanks:
#build

dev/com.ibm.ws.app.manager/src/com/ibm/ws/app/manager/internal/lifecycle/ApplicationStartBarrierImpl.java
AndrewDJBanks: The test case and https://github.com/OpenLiberty/open-liberty/pull/11152
are (going to be)  applicationPrereqs

dev/com.ibm.ws.app.manager.lifecycle/src/com/ibm/wsspi/application/lifecycle/ApplicationPrereq.java
AndrewDJBanks: we're not making it an SPI. (Yet)

brenthdaniel (approved):

PR #12843, opened 2020-07-01
Issue 12576: Boulder with prune and improved cleanup

kristip17:
Fixes #12376

Added pruning
Extra stop clean up
Adjusted the restarts
Had to move Boulder creation as we were initializing it twice otherwise.

Then I also manually cleaned up a bunch of our docker containers where we had stale images still running.

I was able to run back to back on the same server once the docker container was "clean"


kristip17:
#build

kristip17:
I also ran a PB with 10 x full fat and 1 one failure. When I checked the machine, there was  6 hr old boulder/mariaDB running that is probably leftover from a regular run. I will have to continue to clean up machines until we get this into master.

PR #12844, opened 2020-07-01
Add support for the creation.retry.seconds option to the outgoing Kafka connector

Azquelt:
Follow on work for  #10937

This parameter was handled by the incoming connector, but the same issue exists on the outgoing connector as well.


Azquelt:
#build

Azquelt:
#run-libby-bot
#build

PR #12857, opened 2020-07-01
PR for Issue 12778: Fix JWK cache logic

ayoho:
A change delivered in https://github.com/OpenLiberty/open-liberty/pull/9597 can cause the logic to return `null` earlier than expected when trying to find a cached JWK. This can happen if Liberty falls back to using the key text to look up the key in the cache and fails to find a match. The logic is being updated to go back to falling back to checking the cache that's keyed off the JWK location.

Resolves #12778

PR #12863, opened 2020-07-02
Do not use wrong entry from cache

### LINKED ISSUE #9588: Fix JWKS behavior that returns cached JWK despite the JWK not having right KID
The `JWKSet.getJWKByKidInCollection()` method has a behavior that's probably not what we want. Taking a look [here](https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.security.common.jsonwebkey/src/com/ibm/ws/security/common/jwk/impl/JWKSet.java#L92), you'll see the method is implemented to:
1. Return the only cached JWK entry if a `kid` wasn't provided and there's only one JWK in the cache
1. Return the JWK in the cache whose `kid` matches the `kid` provided to the method
1. If the above don't apply, return the only cached JWK entry if there's only one JWK in the cache

That last point - returning the only cached JWK even if the `kid` doesn't match - probably isn't the implementation we want. The method should probably be changed to return `null` in this case, where a `kid` is provided to the method but there are no JWKs in the cache with a matching `kid`. That should be done even if there's only one JWK that's cached. That will force the runtime to do a remote call to attempt to fetch the correct JWK.

jonasrutishauser:
This fixes #9588 (see comment from @pellmont)

### DIFF
--- a/dev/com.ibm.ws.jsp.2.3_fat/fat/src/com/ibm/ws/jsp23/fat/FATSuite.java
+++ b/dev/com.ibm.ws.jsp.2.3_fat/fat/src/com/ibm/ws/jsp23/fat/FATSuite.java
@@ -25,6 +25,7 @@ import com.ibm.ws.jsp23.fat.tests.JSPExceptionTests;
 import com.ibm.ws.jsp23.fat.tests.JSPJava11Test;
 import com.ibm.ws.jsp23.fat.tests.JSPJava17Test;
 import com.ibm.ws.jsp23.fat.tests.JSPJava21Test;
+import com.ibm.ws.jsp23.fat.tests.JSPJava7Test;
 import com.ibm.ws.jsp23.fat.tests.JSPGlobalTLDTest;
 import com.ibm.ws.jsp23.fat.tests.JSPJava8Test;
 import com.ibm.ws.jsp23.fat.tests.JSPPrepareJSPThreadCountDefaultValueTests;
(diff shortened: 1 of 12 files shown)


ayoho:
Thanks for contributing, @jonasrutishauser! However I have a separate fix ready to go under https://github.com/OpenLiberty/open-liberty/pull/12857 that we're going to try instead. Let us know if the problem persists after my PR has been merged and included in one of the development builds available at https://openliberty.io/downloads/#development_builds.

jonasrutishauser:
> Thanks for contributing, @jonasrutishauser! However I have a separate fix ready to go under #12857 that we're going to try instead. Let us know if the problem persists after my PR has been merged and included in one of the development builds available at https://openliberty.io/downloads/#development_builds.

If I simply look at the change of #12857 it does not change much to the logic.
The only change is that there are more cases when the only cache entry is returned (no matter if the entry is the right one).


The problematic call comes from here (and the result is expected to be `null` when `kid` changed): https://github.com/OpenLiberty/open-liberty/blob/17a26220fa4647008b7c2a37eff26c11d8116c3e/dev/com.ibm.ws.security.common.jsonwebkey/src/com/ibm/ws/security/common/jwk/impl/JwKRetriever.java#L347
In the problem case we see in our environment the following holds at line 181:
- the `key` form the cache will still be `null` as the `kid` has changed
- `keytext` is `null` too
- the only entry in the cache is for an other `kid`

therefore no conditional block will match and it will wrongly use the only entry of the cache (which has an other `kid`).

PR #12877, opened 2020-07-02
update SVT focals

gecock:
I am no longer an SVT focal for the organization. Updated the template to reflect that.


tevans78:
#run-libby-bot

Issue #12029, opened 2020-05-01
Jakarta EE 9 : Managed Bean FAT


tkburroughs:
Provide Managed Beans FAT.

`com.ibm.ws.managedbeans*_fat*`

 - [x] ManagedBeanBindingsEJBTest
 - [x] ManagedBeansCdiTest
 - [x] ManagedBeansEjbTest
 - [x] ManagedBeansWebTest

tkburroughs:
All FAT for Jakarta EE 9 is now complete; closing

PR #12879, opened 2020-07-03
Issue 12029: Add additional Managed Beans FAT

tkburroughs:
Add managed beans FAT that covers integration with CDI and
EJB 3.2 level interceptors.

for #12029


tkburroughs:
#build

PR #12887, opened 2020-07-06
Reduce time spent attempting to connect to each remote docker-host svc

aguibert:
#build

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/utils/ExternalTestServiceDockerClientStrategy.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/utils/ExternalTestServiceDockerClientStrategy.java
@@ -123,7 +123,7 @@ public class ExternalTestServiceDockerClientStrategy extends DockerClientProvide
 
         public void test() throws InvalidConfigurationException {
             final String m = "test";
-            final int maxAttempts = FATRunner.FAT_TEST_LOCALRUN ? 1 : 7; // attempt up to 7 times for remote builds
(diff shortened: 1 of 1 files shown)


dev/fattest.simplicity/src/componenttest/topology/utils/ExternalTestServiceDockerClientStrategy.java
```diff
@@ -123,7 +123,7 @@ public boolean isMatched(ExternalTestService dockerService) {
 
         public void test() throws InvalidConfigurationException {
             final String m = "test";
-            final int maxAttempts = FATRunner.FAT_TEST_LOCALRUN ? 1 : 7; // attempt up to 7 times for remote builds
+            final int maxAttempts = FATRunner.FAT_TEST_LOCALRUN ? 1 : 4; // attempt up to 4 times for remote builds
```
websterc87, line 126: 1 min timeout feels like it might be a tad agressive to overcome infrastructure, happy to change this for now but would recommend revisiting if we get frequent timeouts

dev/fattest.simplicity/src/componenttest/topology/utils/ExternalTestServiceDockerClientStrategy.java
```diff
@@ -148,7 +148,7 @@ public void test() throws InvalidConfigurationException {
                     if (firstIssue == null)
                         firstIssue = t;
                     if (attempt < maxAttempts) {
-                        int sleepForSec = Math.min(10 * attempt, 45); // increase wait by 10s each attempt up to 45s max
+                        int sleepForSec = 15;
```
websterc87, line 151: Constant time looks sane for me

PR #12904, opened 2020-07-06
Don't use ThreadLocal for request data unless async

dettmoney:
Throughput profiles show SRTServletRequestThreadData using ThreadLocals during every request. Further examination shows the only reason we actually need thread-specific data is when doing async requests. Change the code to only do TLs during async requests to get 1-2% throughput increase.


dettmoney:
#build

PR #12917, opened 2020-07-07
Issue 12911: Skip tests on specific OS/JDK

### LINKED ISSUE #12911: acmeCA-2.0: Skip tests on specific Windows/Java level
There is a bug on a specific level of OpenJDK on Windows which causes some tests to fail when deleting directories. It fails on OpenJDK Runtime Environment (11.0.5+10), but passes on 11.0.7+10.

`The process cannot access the file because it is being used by another process.` occurs trying to do a file delete. The process is stopped and sleep/retries do not help.

For #9017

kristip17:
Fixes #12911 

Skipped AcmeSimpleTests on Windows/OpenJDK 11.0.5. Appears to work on later drivers (for sure 11.0.7) so don't want to be limited when the build systems upgrade.


kristip17:
#build

Issue #10094, opened 2019-12-09
gRPC FAT framework


wtlucy:
Create a FAT framework to test gRPC client and server scenarios

PR #12922, opened 2020-07-07
gRPC FAT improvements

wtlucy:
`com.ibm.ws.fat.grpc.StoreServicesTests.testGetAppInfo()` fails on some platforms due to a hardcoded port - I'll get rid of that.  Additionally, there is a potential for the tests in HelloWorldTest to run before the SSL port comes up - that test class needs to wait for the ssl channel init message.

#10094


wtlucy:
#build

Issue #12934, opened 2020-07-08
Jakarta EE 9: Provide ejbLite-4.0 feature implementation


tkburroughs:
Provide ejbLite-4.0 feature implementation

tkburroughs:
ejbLite-4.0, all known auto-features, and a minimal amount of testing have been delivered, and included in the betas, so closing out this story.  Additional testing will be delivered under the base EJB epic, #11450

PR #12935, opened 2020-07-08
Issue 12934: Initial ejbLite-4.0 implemenation

tkburroughs:
Add necessary transforms for jakarta ee and add required
features to support ejbLite-4.0

Enabled one basic test using repeat action.

for #12934


tkburroughs:
#build

dev/com.ibm.ws.ejbcontainer.session.passivation_fat/fat/src/com/ibm/ws/ejbcontainer/session/passivation/tests/StatefulTimeoutTest.java
```diff
@@ -48,7 +49,7 @@ public LibertyServer getServer() {
     }
 
     @ClassRule
-    public static RepeatTests r = RepeatTests.with(FeatureReplacementAction.EE7_FEATURES().forServers("com.ibm.ws.ejbcontainer.session.passivation.fat.sfTimeout")).andWith(FeatureReplacementAction.EE8_FEATURES().forServers("com.ibm.ws.ejbcontainer.session.passivation.fat.sfTimeout"));
+    public static RepeatTests r = RepeatTests.with(FeatureReplacementAction.EE7_FEATURES().forServers("com.ibm.ws.ejbcontainer.session.passivation.fat.sfTimeout")).andWith(FeatureReplacementAction.EE8_FEATURES().forServers("com.ibm.ws.ejbcontainer.session.passivation.fat.sfTimeout")).andWith(new JakartaEE9Action().forServers("com.ibm.ws.ejbcontainer.session.passivation.fat.sfTimeout"));
```
tevans78, line 52: I created an convenience method for this that you might wish to try...

https://github.com/OpenLiberty/open-liberty/blob/654502ab95d074975df65dd78baf9e3e0508d247/dev/fattest.simplicity/src/componenttest/rules/repeater/EERepeatTests.java#L61

dev/io.openliberty.ejbcontainer.v40/bnd.bnd
```diff
@@ -0,0 +1,36 @@
...
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
+#*******************************************************************************
+-include= ~../cnf/resources/bnd/bundle.props
+bVersion=1.0
+
+Bundle-Name: Jakarta Enterprise Beans 4.0 Runtime
+Bundle-SymbolicName: io.openliberty.ejbcontainer.v40
```
tevans78, line 15: I know that the ejb container has kinda been named this way for decades (?!) but would you consider calling this something like `io.openliberty.ejb.4.0.internal.container`? Something that follows the guidelines here: https://github.com/OpenLiberty/open-liberty/wiki/Open-Liberty-naming-guidelines#examples

dev/io.openliberty.ejbcontainer.v40/src/io/openliberty/ejbcontainer/v40/internal/EJBRuntimeVersionV40.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.ejbcontainer.v40.internal;
```
tevans78, line 11: The wiki doesn't have an example for a version specific package name but how about
`io.openliberty.ejb.internal.container.v40`?

dev/com.ibm.ws.ejbcontainer.session.passivation_fat/fat/src/com/ibm/ws/ejbcontainer/session/passivation/tests/StatefulTimeoutTest.java
tkburroughs: @tevans78 That is a nice convenience method, but I generally try to adhere to the FAT guidelines that indicate FAT should only run one repeat action in lite mode, the others in full mode.  I had overlooked that this test was repeating both 7 and 8 in lite mode already..... I've updated the test to repeat ee7 in full mode only.  For now, it will run both ee8 and ee9 in lite mode, as I don't want to lose the ee8 since it is the ga version, but would like to get some minimal ee9 testing.

Perhaps that convenience method could be updated to allow passing in something like ("myserver", EE7-FF, EE8, EE9-FF), to allow designating some to repeat full fat only?  We like to switch things up a bit, always have some FAT run lite mode for  EE7, some EE8, and eventually some EE9, but the majority run the latest ga in lite.

dev/io.openliberty.ejbcontainer.v40/bnd.bnd
tkburroughs: @tevans78 Ah, I thought I was following those guidelines; in "io.openliberty.xxx.internal", I interpreted "xxx" to be "component" not "feature".... and this is for the "ejbcontainer.v40" component.... so "io.openliberty.ejbcontainer.v40" and "io.openliberty.ejbcontainer.v40.internal" for the package name.  I think the guidelines need to be updated to also consider the concept of "component" features, which are used by multiple public features.

The problem is, this "private" feature is not directly tied to a single public feature, so replacing "xxx" with "ejb.4.0" is not accurate.  This will in fact be used by multiple public features (at the moment it just happens to be ejbLite-4.0):

- ejbLite-4.0
- ejbHome-4.0
- ejbPersistentTimer-4.0
- ejbRemote-4.0
- ejb-4.0
- mdb-4.0

and could also be applicable to "managedBeans-2.0" in the future, since that feature is also based on the "ejbcontainer" component.

Based on the guidelines, if I happened to do "mdb-4.0" first, then it would be called "io.openliberty.mdb.4.0.internal", but it really wouldn't then fit the guideline when I pull it into the others.  Similarly, even using "4.0" doesn't necessarily work, as the code it effects is also part of the the "managedBeans-2.0" public feature.

Instead, I chose to consider this like "io.openliberty.connector-2.0" or "com.ibm.websphere.appserver.transaction-2.0" (that would be io.openliberty.transaction-2.0 if we could rename it).  These are "component" features that are pulled in by multiple public features with different public names.  "transaction-2.0 may be used by both the ejb-4.0 and servlet-5.0 public features.

I consider this a "component" feature, used by multiple public features.  I consider the component name to be "ejbcontainer.v40" partly for legacy purposes, but also to differentiate it from "ejb", which is pretty ambiguous; there are many other components, like security or jax-rs, or cdi that might include "ejb" in a project, feature, or package name.

The "ejbcontainer" component is also broken down into many sub components that also don't directly relate to public features.  Many of these sub-component projects are just "transformed" now, so don't have "io.openliberty" project names.... but if I were to create new projects, they would be something like:

io.openliberty.ejbcontainer.async
io.openliberty.ejbcontainer.ejb2x (yes, part of multiple public features, including several 4.0 ones)
io.openliberty.ejbcontainer.jpa
io.openliberty.ejbcontainer.mdb
io.openliberty.ejbcontainer.remote
io.openliberty.ejbcontainer.session
io.openliberty.ejbcontainer.timer

Most of these will be pulled into multiple public features, for example "ejbcontainer.timer" will be in "ejbLite-4.0", "ejb-4.0", "ejbHome-4.0", "ejbRemote-4.0" and "mdb-4.0".

Also chose to use "ejbcontainer.v40" as the component, as it fits well with the above list, and makes it easier to identify as the follow-on to "ejbcontainer.v31/v32".  It also separates it from the "4.0" convention since it could be included by "managedBeans-2.0".

The use of "ejbcontainer" also makes things easier for support purposes; easier to direct to proper support queues.

My preference is still to stick with "io.openliberty.ejbcontainer.v40" as a component project. I thought the "internal" convention was only for package names, but I could see renaming the project to "io.openliberty.ejbcontainer.v40.internal" since it would not likely ever also include any API or SPI packages.

dev/io.openliberty.ejbcontainer.v40/src/io/openliberty/ejbcontainer/v40/internal/EJBRuntimeVersionV40.java
tkburroughs: @tevans78 see my above comments. I consider this a "component" project not directly related to a specific public "ejb" feature.  I view ".ejb." in the package name to be too ambiguous.  Using "ejbcotnainer" makes it easier to identify which aspec of EJB, Managed Beans, and MDB this relates to.... the EJB container, not EJB security, or EJB integration with CDI, JAX-RS, WebContainer, etc.

I prefer to include the full "component" (i.e. project name) prior to the "internal".  Seems to make it easier to avoid multiple project/bundles ever exporting the same package.  Also, with the other ejbcontainer projects, I assume this view would make more sense:

io.openliberty.ejbcontainer.async
io.openliberty.ejbcontainer.async.spi
io.openliberty.ejbcontainer.async.internal
io.openliberty.ejbcontainer.timer
io.openliberty.ejbcontainer.timer.spi
io.openliberty.ejbcontainer.timer.internal
io.openliberty.ejbcontainer.v40
io.openliberty.ejbcontainer.v40.spi
io.openliberty.ejbcontainer.v40.internal

Keep in mind that these are all related to a component, like "transactions", which are included by many features, some with "ejb" in the name, and some not, like "mdb" and "managedBeans".

tkburroughs:
#run-libby-bot

dev/io.openliberty.ejbcontainer.v40/bnd.bnd
tevans78: @tkburroughs you make a good argument there. I will try to add something to the wiki to capture this. I think that the project/bundle should still be suffixed with `internal` ... so your suggestion of `io.openliberty.ejbcontainer.v40.internal` was a good one, particularly since that it matches up with the package name below.

dev/io.openliberty.ejbcontainer.v40/src/io/openliberty/ejbcontainer/v40/internal/EJBRuntimeVersionV40.java
tevans78: Agreed. I like your examples so I may use those in the wiki. Just need to try and define what a "component" is??

dev/io.openliberty.ejbcontainer.v40/bnd.bnd
tkburroughs: @tevans78 I've renamed the project to "io.openliberty.ejbcontainer.v40.internal".

tkburroughs:
#run-libby-bot

tkburroughs:
#build

tkburroughs:
One failing FAT is a known intermittent issue.

PR #12950, opened 2020-07-08
Add buildID to liberty-versions.props

wraschke:
#build


wraschke:
#build

wraschke:
UI FAT failure has nothing to do with this change, which really isn't impactful to almost anything.

PR #12958, opened 2020-07-08
Improve performance when no tracer is found

### LINKED ISSUE #12613: Enabling openTracing with no tracer class configured impacts performance
A performance issue occurs when openTracing is configured in the server.xml (individually or as part of a microProfile feature bundle) and no tracer class is configured. What seems to happen is that OpentracingTracerManager keeps looking for a tracer, on every request. This can have a large impact on the throughput of the server if the workload is mostly HTTP requests/responses with little service logic in the app.

We discovered this issue in 20006, but the problem is not new - it can be seen at least as far back as 19003.

fmhwong:
For mpOpenTracing-1.0, mpOpenTracing-1.1 and mpOpenTracing-1.2 only.  For mpOpenTracing-1.3, see PR  https://github.com/OpenLiberty/open-liberty/pull/12763

With MicroProfile umbrella feature is enabled, mpOpenTracing-1.x is also enabled although users have no intention of using OpenTracing. Most likely the loading of a tracer will fail. Previously, mpOT code tried to load the tracer on every JAX-RS request.

This PR will make the following changes:-

- Load tracer only once. A dummy tracer is stored in the tracer map when the initial loading of the tracer is failed. Subsequent requests will not load the tracer again when the map has a dummy tracer.
- AppStateListener is added to listener when an application stopped. The corresponding entry in the tracer map will be removed.
- AppStateListener also listens to <Library /> changes. All entries in the tracer map will be cleared.

Fixes #12613


fmhwong:
#build

PR #12961, opened 2020-07-09
Monitor metric 3.0 fat

Channyboy:
#build
#run-libby-bot

Tests EDF1 and EF12 removed as they are no longer relevant with the mpMetrics-3.0 feature.

Test EDF1 tests the mpMetric feature without the monitor feature and expects no vendor metrics to be available. however, has hard dependency on monitor-1.0 and  thus this test is not useful at all.

Test EDF12 tests the removal of monitor-1.0 feature and expected the removal of vendor metrics.  mpMetrics-3.0, however, has hard dependency on monitor-1.0 and  thus this test is not useful at all.

The relevant files that were designated for tests EDF1 and EDF12 were removed.

Monitor-Metrics bundle had to modify MonitorMetrics.nlsprops to update appropriate message code.


andrea-pichler:
No L2 message review is needed for translations

Channyboy:
Commits squashed. Building one more time.
#build 
#run-libby-bot

PR #13057, opened 2020-07-16
Update Apache commons-codec to latest version (1.14)

ayoho:


ayoho:
#build

dev/cnf/oss_dependencies.maven
```diff
@@ -90,6 +90,7 @@ commons-beanutils:commons-beanutils:1.9.4
 commons-cli:commons-cli:1.4
 commons-codec:commons-codec:1.10
 commons-codec:commons-codec:1.11
+commons-codec:commons-codec:1.14
```
tevans78, line 93: Why do we need to keep the old versions as well?

ayoho: I don't know everyone who might be using some of the old versions, but I know some FATs use some of the older libraries. I don't want to dive into making sure everyone's happy and working with the latest version just to get this upgrade delivered.

ayoho:
#build

ayoho:
#build

Issue #11803, opened 2020-04-17
Jakarta RESTful Web Services 3.0


andymc12:
Describe the high level feature, including any external spec links.
This will be the epic issue for Jakarta RESTful Web Services (aka JAX-RS) 3.0, part of the EE 9 effort covered by Epic issue #9876.

The 3.0 spec adds little more than some behavior clarifications over the 2.1 spec, but we may be changing the implementation of the 3.0 spec.

When ready, add links to the Upcoming Feature Overview document and Feature Test Summary issue:
- UFO: https://ibm.box.com/shared/static/6tfnxuor042psbtu76k85p0qcd0pzpq0.pptx
- FTS: https://github.com/OpenLiberty/open-liberty/issues/15552

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## List of Steps to complete or get approvals / sign-offs for Onboarding to the Liberty release (GM date)

Instructions:
- Do the actions below and mark them complete in the checklist when they are done.
- Make sure all feature readiness approvers put the appropriate tag on the epic to indicate their approval.

## **Design**
Before Development Starts or 8 weeks before Onboarding
- [x] POC Design / UFO Review Scheduled (David Chang) or N/A.
- [x] POC Design / UFO Reviewed (Feature Owner) or N/A.
- [x] Complete any follow-ons from the POC Review.
- [x] Design / UFO Approval (Alasdair Nottingham) or N/A.
- [x] No Design / No UFO Approval (Arthur De Magalhaes - cloud / Alasdair Nottingham - server) or N/A.
- [x] SVT Requirements identified. (Epic owner / Feature owner with SVT focal point)
- [x] ID Requirements identified. (Epic owner / Feature owner with ID focal point)
- [x] Create a child task of this epic entitled "FAT Approval Test Summary". Add the link in above.

## **Legal**
3 weeks before Onboarding
- [ ] Identify all open source libraries that are changing or are new. Work with Legal Release Services (Cass Tucker or Release PM) to get open source cleared and approved. Or N/A. (Epic Owner).   New or changed open source impacts license and Certificate of Originality.

## **Translation**
3 weeks before Onboarding
- [x] All new or changed PII messages are checked into the integration branch, before the last translation shipment out. (Epic Owner)

## **Feature Complete**
2 weeks before Onboarding
- [x] Implementation complete. (Epic owner / Feature owner)
[...]

PR #13058, opened 2020-07-16
Ensure a class can be both an Application subclass and a resource

andymc12:
- runtime change (that should no longer be required assuming
https://issues.redhat.com/browse/RESTEASY-2650 is resolved & pulled in)
- new FAT test bucket specific to RESTEasy runtime with App+Resource FAT
- updating Providers service to ensure all are included
- moving NLS file to proper "resources" directory

Part of epic issue #11803


andymc12:
@OpenLiberty/message-reviewer Message file was just moved to new location - no changes to message text.

#build

PR #13072, opened 2020-07-16
Rename lra project to io.openliberty

idlewis:
As per project guidelines


dev/io.openliberty.mpLRA.1.0/bnd.bnd
```diff
@@ -13,12 +13,12 @@
 bVersion=1.0
 
 Bundle-Name: Long Running Actions
-Bundle-SymbolicName: com.ibm.ws.lra
+Bundle-SymbolicName: io.openliberty.mpLRA
```
tevans78, line 16: I think this bundle name should have `internal` on the end like the package it exports... neither are APIs

PR #13075, opened 2020-07-16
Add FT 3.0 - MP Context Propagation integration

Azquelt:
Make FT 3.0 use MP Context Propagation to propagate context between threads if available.


Azquelt:
In the end, this wasn't added to the spec.

PR #13086, opened 2020-07-17
Removed general catch from Shared Server

### LINKED ISSUE #13106: Update buckets with missing tested.features
We noticed that there was an issue were some of the checks weren't appearing as a failure when running FAT buckets. Taking a deeper look it appears that the issue is with the Shared Server were there is a general catch block which swallows the errors. This shouldn't be this way since if an error occurs when the server is coming up it should fail the tests.

isaacrivriv:
fixes #13106


isaacrivriv:
#build

isaacrivriv:
#build

isaacrivriv:
Closing and doing updated work in another PR

PR #13102, opened 2020-07-20
Dynamically import com.sun JAXB impl classes in JAX-RS 3.0 / RESTEasy prototype

andymc12:
The `jaxrs-3.0` feature's io.openliberty.org.jboss.resteasy.common bundle fails to resolve when using Java 8.  This is because the RESTEasy implementation code directly depends on some com.sun JAXB implementation classes that are part of the JDK in Java 8 (and thus not exported as an OSGi package), but are not part of the JDK in Java 11 (where we provide the JAXB implementation via separate bundle that _does_ export the package).

This change uses dynamic imports to import the package if it is exported, but will use the JDK-supplied package if it is not exported.


andymc12:
#build

andymc12:
Added some more JAXB packages - new #build

andymc12:
#run-libby-bot

PR #13111, opened 2020-07-20
Renamed features to io.openliberty.X

### LINKED ISSUE #12916: Update feature file names (and symbolic names) for jsp-3.0, and el-4.0
These features still use the com.ibm.websphere.appserver.XXX.feature name if you look in the om.ibm.websphere.appserver.features/visibility/public/ directories. 

They should be updated to io.openliberty.feature-X.X per the naming guidelines. 

servlet-5.0 should be left alone because it needs to support the tolerates attribute.

isaacrivriv:
fixes #12916


isaacrivriv:
#build

PR #13115, opened 2020-07-21
launch wab module uninstall async

sebratton:


sebratton:
#build

sebratton:
#build

dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WABInstaller.java
```diff
@@ -568,9 +577,33 @@ protected boolean uninstallFromWebContainer(WAB wab) {
                             wabGroupsLock.unlock();
                         }
                         if (wab.getCreatedApplicationInfo()) {
+                            // system WABs get uninstalled here
                             deployedApp.uninstallApp(wab);
                         } else {
-                            deployedApp.uninstallModule(wab);
+                            final CountDownLatch latch = new CountDownLatch(1);
+                            if (ctx.getBundle(0).getState() == Bundle.STOPPING) {
```
tjwatson, line 584: should use `ctx.getBundle(org.osgi.framework.Constants.SYSTEM_BUNDLE_LOCATION)` instead of bundle ID `0` here to avoid calling of region framework hooks.

dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WABInstaller.java
```diff
@@ -568,9 +577,33 @@ protected boolean uninstallFromWebContainer(WAB wab) {
                             wabGroupsLock.unlock();
                         }
                         if (wab.getCreatedApplicationInfo()) {
+                            // system WABs get uninstalled here
                             deployedApp.uninstallApp(wab);
                         } else {
-                            deployedApp.uninstallModule(wab);
+                            final CountDownLatch latch = new CountDownLatch(1);
```
tjwatson, line 583: The `latch` is only used inside the following if.  Would be better to only construct it inside the if.

dev/com.ibm.ws.app.manager.wab/src/com/ibm/ws/app/manager/wab/internal/WABInstaller.java
```diff
@@ -568,9 +577,33 @@ protected boolean uninstallFromWebContainer(WAB wab) {
...
+                                }, Boolean.TRUE);
+                                try {
+                                    // If a quiesce has not been initiated then a force stop was called so
+                                    // continue without waiting. Otherwise give the uninstallModule thread
+                                    // time to unwind before proceeding with WAB removal.
+                                    if (quiesceStarted.get()) {
+                                        latch.await(30, TimeUnit.SECONDS);
+                                    }
+                                } catch (InterruptedException e) {
+                                    // Continue with wab uninstall on this thread.
```
tjwatson, line 602: Typically the best practice is to do the following when you catch the InterruptedException and don't want to propagate any errors:

```
Thread.currentThread().interrupt();
```
This is to keep the thread interrupted so that anything else doing work has to exit quickly also.

sebratton:
#build

sebratton:
#libbybot

sebratton:
fixes #21106

Issue #13091, opened 2020-07-17
grpc-1.0 java.lang.IllegalStateException: Past end of stream 


wtlucy:
[One of the gRPC tests](https://github.com/OpenLiberty/open-liberty/blob/master/dev/com.ibm.ws.grpc_fat/test-applications/StoreProducerApp.war/src/com/ibm/testapp/g3store/restProducer/client/ProducerEndpointFATServlet.java#L170) causes a `java.lang.IllegalStateException: Past end of stream` in some environments.  I suspect there is a race condition in the gRPC-specific end of stream signalling, namely around https://github.com/OpenLiberty/open-liberty/blob/master/dev/com.ibm.ws.transport.http/src/com/ibm/wsspi/http/ee7/HttpInputStreamEE7.java#L121 and https://github.com/OpenLiberty/open-liberty/blob/master/dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/internal/HttpServiceContextImpl.java#L5678
```
com.ibm.testapp.g3store.grpcservice.StoreProducerService     I createApps ----- request received by StoreProducer grpcService to create the app names  -----

io.grpc.internal.SerializingExecutor                         E Exception while executing runnable io.grpc.servlet.ServletAdapter$GrpcReadListener$$Lambda$159/0000000000000000@dc724a34
java.lang.IllegalStateException: Past end of stream
at com.google.common.base.Preconditions.checkState(Preconditions.java:508)
at io.grpc.internal.AbstractServerStream$TransportState.inboundDataReceived(AbstractServerStream.java:264)

PR #13117, opened 2020-07-21
Resolve incorrect async isFinished

wtlucy:
In a specific async case, the input stream will return true from isFinished() prematurely.  This PR will fix that case.  For #13091


wtlucy:
#build

wtlucy:
#build

wtlucy:
The build failures are unrelated, I'll merge.  #run-libby-bot

Issue #12504, opened 2020-06-04
Update OpenID Connect client API to retrieve raw id_token


chunlongliang-ibm:
The openidConnectClient-1.0 feature has API com.ibm.websphere.security.openidconnect.token.IdToken, which has getter to retrieve claims. However there is no way to get the original raw id_token. This request is to add a method to retrieve the raw id_token

chunlongliang-ibm:
see https://github.com/OpenLiberty/open-liberty/issues/10091

chunlongliang-ibm:
Here is the sample code:

    public String getRawIdToken() throws WSSecurityException {
        Subject subj = WSSubject.getRunAsSubject();
        Set<Object> privCredentials = subj.getPrivateCredentials();
        for (Object credentialObj : privCredentials) {
            if (credentialObj instanceof Map) {
                Object value = ((Map<?, ?>) credentialObj).get("id_token");
                if (value != null)
                    return (String) value;
            }
        }
        
        return null;
    }

ayoho:
https://github.com/OpenLiberty/open-liberty/pull/13126 is the PR that Julien opened for the product code change. The next step is to develop FATs to ensure the new API works before we deliver it.

PR #13126, opened 2020-07-21
Issue 12504: Implementing getRawIdToken()

julien2424:
Implementing a getRawIdToken() for the com.ibm.websphere.security.openidconnect.token.IdToken API.
for #12504

### DIFF
--- a/dev/com.ibm.ws.security.oauth/src/com/ibm/websphere/security/openidconnect/token/IdToken.java
+++ b/dev/com.ibm.ws.security.oauth/src/com/ibm/websphere/security/openidconnect/token/IdToken.java
@@ -13,34 +13,44 @@ package com.ibm.websphere.security.openidconnect.token;
 import java.util.List;
 import java.util.Map;
 
+import com.ibm.websphere.security.WSSecurityException;
+
 public interface IdToken {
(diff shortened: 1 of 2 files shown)


dev/com.ibm.ws.security.oauth/src/com/ibm/websphere/security/openidconnect/token/IdToken.java
```diff
@@ -13,34 +13,44 @@
...
+     * Case sensitive
      * (key-id jti)
-     * 
+     *
      * @return JWT ID
      */
     public String getJwtId();
 
+    /**
+     * Returns the raw Id of the IdToken
```
ayoho, line 31:
```suggestion
     * Returns the raw ID token
```

lannamal:
@ayoho @barbj 

As per our discussion in this slack thread https://ibm-cloud.slack.com/archives/C30NGTBFS/p1672925469997279, as part of IBM OpenPages team, we are blocked by not having the raw `id_token` string. This is blocking us from integrating with other applications which needs `id_token` from the OIDC provider.

Can we have this issue prioritized and merged. Thanks

Issue #13138, opened 2020-07-22
<include> tag not being considered when installing server.xml


TayyabDev:

TayyabDev:
This PR fixed the bug where the include tag in server.xml's was being ignored by installUtility and featureUtility. The include element can be used in a server.xml in order to consolidate a configuration from a separate server.xml file. The bug was fixed by handling the parsing with a more appropriate XML method. The fix was merged on July 23 and is available in 20.0.0.9.

PR #13141, opened 2020-07-22
Fix bug causing include tag not to be parsed with installUtility and featureUtility

TayyabDev:
Fixes https://github.com/OpenLiberty/open-liberty/issues/13138


TayyabDev:
#build

TayyabDev:
#run-libby-bot

TayyabDev:
This PR fixed the bug where the <include> tag in server.xml's was being ignored by installUtility and featureUtility. The include element can be used in a server.xml in order to consolidate a configuration from a separate server.xml file. The bug was fixed by handling the parsing with a more appropriate XML method. The fix was merged on July 23 and is available in 20.0.0.9.

PR #13171, opened 2020-07-25
Issue #13170: Allow MDB methods to be non-public final

### LINKED ISSUE #13170: MDB method restricted from being private final for no methods listener
When an MDB has a "no methods" message listener interface, the following error occurs if one of the MDB implementation class methods is declared non-public and final:

EJBConfigurationException: Method xxxx on class xx.xxx.Xxxx must not be declared as final for the no-interface local view of bean  App#Mod.jar#‌Bean.

However, the EJB specification indicates that non-public methods on the bean class may be final when using a "no methods" message listener interface.

A change will be made to allow non-public methods on message-driven bean class to be final.

tkburroughs:
Message-driven beans using a "no methods" message listener interface
may have non-public methods that are final; per the EJB specification.

Stop calling the method validation rules for session no-interface view,
as the same limitations do not apply to message-driven beans.

fixes #13170


tkburroughs:
#build

olendvcook:
@tkburroughs merged

PR #13172, opened 2020-07-25
test deserialization after switching between Jakarta and Java EE

njr-11:
Capture context with Jakarta enabled and deserialize when Java EE is enabled, and the other way around.


njr-11:
#build

PR #13186, opened 2020-07-27
More error message  tidy

idlewis:
and better exception handling

### DIFF
--- a/dev/io.openliberty.microprofile.lra.1.0.internal/resources/io/openliberty/microprofile/lra/internal/resources/Lra.nlsprops
+++ b/dev/io.openliberty.microprofile.lra.1.0.internal/resources/io/openliberty/microprofile/lra/internal/resources/Lra.nlsprops
@@ -17,11 +17,11 @@
 # LRA is currently using the message prefix CWMRX in the range 5000-5999
 # This is temporary and should be fixed with a proper prefix before GA.
 
-LRA_INVALID_COORDINATOR_URI=CWMRX5000E: The LRA coordinator URI is invalid
-LRA_INVALID_COORDINATOR_URI.explanation=LRA participants must be able to contact the coordinator
-LRA_INVALID_COORDINATOR_URI.useraction=Correct the LRA coordinator configuration in server.xml
+LRA_INVALID_COORDINATOR_URI.CWMRX5000E=CWMRX5000E: The LRA coordinator URI is invalid
+LRA_INVALID_COORDINATOR_URI.CWMRX5000E.explanation=LRA participants must be able to contact the coordinator
+LRA_INVALID_COORDINATOR_URI.CWMRX5000E.useraction=Correct the LRA coordinator configuration in server.xml
 
 
-LRA_CANT_REGISTER_FILTERS=CWMRX5001E: An error occurred during initialization of the mpLRA-1.0 feature. The error is {0}
-LRA_CANT_REGISTER_FILTERS.explanation=An LRA error has occured that prevents normal operation.
-LRA_CANT_REGISTER_FILTERS.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
+LRA_CANT_REGISTER_FILTERS.CWMRX5001E=CWMRX5001E: An error occurred during initialization of the mpLRA-1.0 feature. The error is {0}
+LRA_CANT_REGISTER_FILTERS.CWMRX5001E.explanation=An LRA error has occured that prevents normal operation.
+LRA_CANT_REGISTER_FILTERS.CWMRX5001E.useraction=Review the error message, the messages.log file on the server, and FFDC logs to identify the problem.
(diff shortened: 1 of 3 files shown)


dev/io.openliberty.microprofile.lra.1.0.internal/src/io/openliberty/microprofile/lra/internal/LraFilterComponent.java
```diff
@@ -63,8 +63,9 @@ protected void activate(Map<String, Object> properties) throws LraException {
             URI coord = new URI(coordString);
             NarayanaLRAClient.setDefaultCoordinatorEndpoint(coord);
         } catch (URISyntaxException e) {
-            // TODO Not sure how to handle config problems yet
-            Tr.error(tc, "Things went wrong");
+            if (TraceComponent.isAnyTracingEnabled() && tc.isErrorEnabled()) {
+                Tr.error(tc, "Invalid LRA coordinator URI: '" + coordString + "'");
+            }
```
Azquelt, line 68: Don't check `isAnyTracingEnabled()` when logging errors.

Checking `isErrorEnabled()` is not wrong but usually unnecessary, you can assume that errors and warnings are always enabled. The check here would only be for performance anyway.

dev/io.openliberty.microprofile.lra.1.0.internal/resources/io/openliberty/microprofile/lra/internal/resources/Lra.nlsprops
```diff
@@ -20,3 +20,8 @@
 LRA_INVALID_COORDINATOR_URI=CWMRX5000E: The LRA coordinator URI is invalid
 LRA_INVALID_COORDINATOR_URI.explanation=LRA participants must be able to contact the coordinator
 LRA_INVALID_COORDINATOR_URI.useraction=Correct the LRA coordinator configuration in server.xml
+
+
+LRA_CANT_REGISTER_FILTERS=CWMRX5001E: An error occurred during initialization of the mpLRA-1.0 feature. The error is {0}
```
Azquelt, line 25: It's not required but often helpful to include the message code (`CWMRX5001E`) in the message key (currently `LRA_CANT_REGISTER_FILTERS`), so that you can search the workspace for the message code and find out where it's produced.

dev/io.openliberty.microprofile.lra.1.0.internal/src/io/openliberty/microprofile/lra/internal/LraFilterComponent.java
```diff
@@ -86,38 +87,33 @@ protected void deactivate(int reason) {
...
-
-                // TODO Auto-generated catch block
-                // Do you need FFDC here? Remember FFDC instrumentation and @FFDCIgnore
-                // https://websphere.pok.ibm.com/~alpine/secure/docs/dev/API/com.ibm.ws.ras/com/ibm/ws/ffdc/annotation/FFDCIgnore.html
-                Tr.error(tc, "Couldn't register the Server filter", e);
-                return;
+                if (TraceComponent.isAnyTracingEnabled() && tc.isErrorEnabled()) {
+                    Tr.error(tc, "Exception during registration of LRA filters", e);
+                    throw new LraRuntimeException(Tr.formatMessage(tc, "LRA_CANT_REGISTER_FILTERS", e), e);
+                }
```
Azquelt, line 114: Don't check `isAnyTrackingEnabled` before logging errors or throwing exceptions.

Don't need to check `isErrorEnabled()` either.

Azquelt:
FYI: `isAnyTracingEnabled()` is supposed to be a very fast check to see if any logging is enabled for EVENT or below which allows the JIT to optimise out the whole block if the user hasn't turned that logging on.

You must only use it if you're logging at EVENT or DEBUG.

dev/io.openliberty.microprofile.lra.1.0.internal/src/io/openliberty/microprofile/lra/internal/LraFilterComponent.java
```diff
@@ -63,10 +63,8 @@ protected void activate(Map<String, Object> properties) throws LraException {
             URI coord = new URI(coordString);
             NarayanaLRAClient.setDefaultCoordinatorEndpoint(coord);
         } catch (URISyntaxException e) {
-            if (TraceComponent.isAnyTracingEnabled() && tc.isErrorEnabled()) {
-                Tr.error(tc, "Invalid LRA coordinator URI: '" + coordString + "'");
-            }
-            throw new LraException(Tr.formatMessage(tc, "LRA_INVALID_COORDINATOR_URI"), e);
+            Tr.error(tc, "Invalid LRA coordinator URI: '" + coordString + "'");
```
Azquelt, line 66: Calls to `Tr.error` should also be using the translated strings.

dev/io.openliberty.microprofile.lra.1.0.internal/src/io/openliberty/microprofile/lra/internal/LraFilterComponent.java
```diff
@@ -108,10 +106,8 @@ public void installProvider(boolean clientSide, List<Object> providers, Set<Stri
                 // the exception
                 providers.add(new ServerLRAFilter());
             } catch (Exception e) {
-                if (TraceComponent.isAnyTracingEnabled() && tc.isErrorEnabled()) {
-                    Tr.error(tc, "Exception during registration of LRA filters", e);
-                    throw new LraRuntimeException(Tr.formatMessage(tc, "LRA_CANT_REGISTER_FILTERS", e), e);
-                }
+                Tr.error(tc, "Exception during registration of LRA filters", e);
```
Azquelt, line 109: Calls to `Tr.error` should also be using the translated strings.

Issue #7887, opened 2019-06-14
Add GraphiQL UI to MP GraphQL feature


andymc12:
It would be nice to include a "switch" that when enabled would automatically add the GraphiQL HTML page to the user application.  The switch itself should either be a Liberty configuration option in the server.xml (i.e. `<graphql graphiqlURI="/graphiql.html"/>`) or a MicroProfile Config property (i.e. `io.openliberty.includeGraphiQL=true` or `io.openliberty.graphiql.uri=/graphiql.html`, etc.).  For starters, we could even use a system property - though we would probably need to change it to either Liberty or MP config before releasing.

The GraphiQL HTML file is located in some of the existing function tests - like: https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.microprofile.graphql.1.0_fat/test-applications/ifaceApp/resources/graphiql.html

In order to add this to applications, it would need to be added as a virtual servlet, similar to what we do with the base GraphQL servlet here: https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.microprofile.graphql.1.0/src/com/ibm/ws/microprofile/graphql/component/GraphQLServletContainerInitializer.java

andymc12:
Resolved by #13202 - using MP Config property, `io.openliberty.enableGraphQLUI` which can be set to true to enable GraphiQL at <context-root>/graphql-ui - the default value of this property is false (disabled).

PR #13202, opened 2020-07-27
Graphiql

### LINKED ISSUE #13201: Allow users to automatically use GraphiQL UI for testing MP GraphQL feature
We've used the [GraphiQL](https://github.com/graphql/graphiql/tree/main/packages/graphiql) web UI in order to demonstrate the power of MP GraphQL, but in order for users to use it, they must package the GraphiQL HTML files with their app. It would be much more convenient for users if we packaged it for them and made it available via a MP Config property.

With this enhancement, user can set a MP Config property, `io.openliberty.enableGraphQLUI` to `true` which will automatically add the GraphiQL web UI to the `/graphql-ui` path in their web application.

For example, suppose the user has a GraphQL WAR file (e.g. "myGraphQLApp.war") in their dropins directory and the following server.xml:
```
<server>
  <featureManager>
    <feature>mpGraphQL-1.0</feature>
  </featureManager>

  <variable name="io.openliberty.enableGraphQLUI" value="true" />
</server>
```

Now they would be able to test their GraphQL app via GraphiQL at: http://localhost:9080/myGraphQLApp/graphql-ui

andymc12:
Fixes #13201

### DIFF
--- a/dev/com.ibm.ws.io.smallrye.graphql/resources/META-INF/ui/graphiql.html
+++ b/dev/com.ibm.ws.io.smallrye.graphql/resources/META-INF/ui/graphiql.html
@@ -1,0 +1,142 @@
+<!--
+ *  Copyright (c) Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the license found in the
+ *  LICENSE file in the root directory of this source tree.
+-->
+<!DOCTYPE html>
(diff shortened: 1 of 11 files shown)


andymc12:
#build

andymc12:
also fixes #7887 - I forgot I opened that when I opened #13201...

PR #13212, opened 2020-07-28
Move expose ports to constructor to avoid duplicate key exception

KyleAure:
Issue originally resolved in Test Containers repo under PR https://github.com/testcontainers/testcontainers-java/pull/2473

But using a custom SQLServerContainer in Liberty to avoid having to provide an accept license file.  Applying fix to this custom SQLServerContainer.


KyleAure:
#build

PR #13213, opened 2020-07-28
Close connection to avoid leak

### LINKED ISSUE #11556: Connection leak when XAResource.recover fails
An Oracle XAResource failed with XAException XAER_RMERR on transaction recovery at server restart.  This was due to lack of permissions, however it was observed that in this scenario a connection leak occurred.

The issue is caused by the early return in XARecoveryData.recover when handling the exception from xarm.recover.  This should still call xarm.closeConnection() in order to tidy up the existing connection (it will request a new XAResource on retry).  NOTE the code performed the xarm.closeConnection() on failure in earlier, pre-open source versions.

neilgsyoung:
fixes #11556


neilgsyoung:
#build

PR #13221, opened 2020-07-28
Update JAX-B feature kind

neuwerk:


neuwerk:
#build

dev/com.ibm.websphere.appserver.features/visibility/public/jaxb-3.0/io.openliberty.jaxb-3.0.feature
```diff
@@ -22,5 +22,5 @@ Subsystem-Name: Jakarta XML Bindings 3.0
   io.openliberty.jakarta.jaxb-3.0
 -bundles=\
   io.openliberty.jaxb.3.0.internal.tools
-kind=noship
+kind=beta
 edition=base
```
tevans78, line 26: jaxb-3.0 is in the JEE 9 full profile right? Not web profile? You should probably add it to `com.ibm.websphere.appserver.jakartaee-9.0`

tevans78: At the moment, cdi-3.0 drags the API feature into the web profile ... we should check if that is correct or not.

neuwerk: @tevans78 thanks for checking this! That's correct: jaxb-3.0 isn't a part of the web profile? So I've added the jaxb-3.0 to the `com.ibm.websphere.appserver.jakartaee-9.0` feature as you suggested, as far as the private feature being dragged into the webProfile, that kind of matches what was done with web profile 8. Since it's a private feature that really only contains the spec API, is that necessarily a problem?

tevans78: I don't think it is a problem ... just worth checking because I can't remember why CDI pulls it in

neuwerk:
#build

neuwerk:
#run-libby-bot

PR #13225, opened 2020-07-28
Issue #13170: Skip MDB No Methods Interface final methods

### LINKED ISSUE #13170: MDB method restricted from being private final for no methods listener

tkburroughs:
When using a "no methods" message listener, the only methods on the
MDB class that should be final are private methods. If there are
any other methods that are final, they should be skipped (ignored)
and not added to the MessageEndpong proxy.

Although not allowed per the specification, in order to provide
backward compatibility with applications written prior to the
intorduction of "no methods" support, these methods will be
tolerated.

fixes #13170


tkburroughs:
#build

olendvcook:
@tkburroughs merged

PR #13241, opened 2020-07-29
Implement subject cache for KerberosService

aguibert:


aguibert:
#build

aguibert:
#run-libby

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
```diff
@@ -0,0 +1,91 @@
...
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.websphere.ras.annotation.Trivial;
+
+@Trivial
+public class LRUCache {
+
+    private static final TraceComponent tc = Tr.register(LRUCache.class);
+
+    private final Map<KerberosPrincipal, Subject> cache = new LinkedHashMap<KerberosPrincipal, Subject>() {
```
utle, line 33: We need to make sure the KerberosPrinicpal is a fully qualify Kerberos principal name userName@KerberosRealmName

aguibert: @utle it will automatically add the default realm name if it is not specified in the principal string, this is the main reason I'm using `KerberosPrinicpal` here instead of just `String`. I ran this test to show that the default realm name is automatically added:
```java
KerberosPrincipal p1 = new KerberosPrincipal("dbuser");
KerberosPrincipal p2 = new KerberosPrincipal("dbuser@EXAMPLE.COM");

System.out.println("equal? " + p1.equals(p2)); // shows both principals are equal
```

aguibert:
#build

aguibert:
#run-libby

aguibert:
#build

PR #13253, opened 2020-07-30
Minor fixes for wlp.lib.extract_fat project

tjwatson:
- Use @RunWith(FATRunner.class) annotation.  This allows things like the
option to run a single test to work with options like
-Dfat.test.class.name
- Fix NPE on extractLoc when foundWatchFor returns fast.  This is
probably only realistic to happen when stepping through with a debugger.


tjwatson:
#build

PR #13293, opened 2020-08-03
Fix usage of ConnectionManagerMBean attributes

aguibert:
If the ConnectionManagerMBean is used via proxy, currently most of the operations do not work, such as:
- getSize
- getMaxSize
- getAvailable
- getJndiName

#build


aguibert:
#build

aguibert:
#run-libby-bot

PR #13300, opened 2020-08-04
Add test for handling arrays as QueryParams parameters

andymc12:
This came about from a conversation in the MP Metrics spec - which had outlined how arrays should be handled when used as query parameters.  The spec does not require arrays to be supported, but that is likely to change in JAX-RS 3.1. CXF already supports arrays as parameters, and this test ensures that RESTEasy does too.


andymc12:
#build

PR #13303, opened 2020-08-04
Increasing sleeps for testParallelMultipleRequest

lamkavan:


lamkavan:
#build

lamkavan:
### Problem 
The test case called testParallelMultipleRequests in com.ibm.ws.request.timing.fat.HungRequestTiming fails due to not finding at least 2 hung request detection warning/messages in messages.log. To be specific, it only finds one. There is no problem with finding the hung detection messages themselves and from further log analysis it truly is the case that only one request is detected to be hung when there is suppose to be two. This issue intermittently occurs.

### Solution
In the test code we are creating two request in parallel running on separate threads. Each request will basically sleep for a given amount of time. The first thread sleeps for 4 sec and the second sleeps for 6 secs. The test failure can occur when the sleeps are too small and appears to be resolved by increasing the sleeps to 10 and 14 seconds. Something that may explain this behavior is that testParallelMultipleRequests is usually the first test case that get executed and because the sleeps are too small, we do not give enough time for the request timing feature to be ready to start detecting hung request. No exact cause could be found.

PR #13305, opened 2020-08-04
Export SmallRye GraphQL packages at the same version that we pull in

andymc12:
Updating package export version to match the same version of SmallRye that we pull in. The new version has not shipped yet, so this is not a release bug.


andymc12:
#build

PR #13319, opened 2020-08-05
Rename internal connectors features

KyleAure:
Renaming the internal connectors-2.0 features to follow the naming standard here: https://github.com/OpenLiberty/open-liberty/wiki/Open-Liberty-naming-guidelines

### DIFF
--- a/dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.connectors-2.0.internal.api.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.connectors-2.0.internal.api.feature
@@ -1,5 +1,5 @@
 -include= ~${workspace}/cnf/resources/bnd/feature.props
-symbolicName=io.openliberty.jakarta.internal.api.connectors-2.0
+symbolicName=io.openliberty.jakarta.connectors-2.0.internal.api
 singleton=true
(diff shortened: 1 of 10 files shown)


KyleAure:
#build

dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.connectors-2.0.internal.api.feature
```diff
@@ -1,5 +1,5 @@
 -include= ~${workspace}/cnf/resources/bnd/feature.props
-symbolicName=io.openliberty.jakarta.internal.api.connectors-2.0
```
tevans78, line 2: I am slightly confused by the name of this feature ... how can it be both internal and an API? The pattern used by most components is that a private feature to expose only a (Jakarta) API would simply be called `io.openliberty.jakarta.connectors-2.0`.

dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.connectors-2.0.internal.feature
```diff
@@ -1,5 +1,5 @@
 -include= ~${workspace}/cnf/resources/bnd/feature.props
-symbolicName=io.openliberty.jakarta.internal.connectors-2.0
```
tevans78, line 2: ...and then if this feature is an internal implementation feature (not API), I would drop `jakarta` from the name... `io.openliberty.connectors-2.0.internal`.

KyleAure:
#build 
#libbybot

PR #13324, opened 2020-08-05
Append _fat to build.featureStart

wraschke:
#build


wraschke:
#build

wraschke:
#build

wraschke:
#build

wraschke:
#run-libby

fritze2 (approved):

wraschke:
The PB above runs only `build.featureStart_fat` and as expected, runs it only in child build with label `Open Liberty feature simple start test with Open Liberty image`. There are two spawned child builds because automation specified that this bucket run in full mode, too, but there isn't a full mode for this bucket.

The PB I manually requested so that all FATs are run is still running and is delayed and impacted by the libertyfs maintenance. There are some child builds failed due to `com.jcraft.jsch.JSchException: connection is closed by foreign host` exception. My change does not cause that. So, I am merging this PR given that the PB requested by automation is good and I see from manually requested PB that this change results in expected behavior (runs in an OL child build using OL image).

PR #13353, opened 2020-08-06
Kerberos datasource bindings and password support

aguibert:


aguibert:
#build

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/KerberosService.java
```diff
@@ -110,28 +123,48 @@ public Path getKeytab() {
...
         return createdSubject;
     }
 
-    private Subject doKerberosLogin(String principal) throws LoginException {
+    private Subject doKerberosLogin(String principal, SerializableProtectedString pass) throws LoginException {
         Subject subject = new Subject();
         Krb5LoginModuleWrapper krb5 = new Krb5LoginModuleWrapper();
         Map<String, String> options = new HashMap<String, String>();
         Map<String, Object> sharedState = new HashMap<String, Object>();
+        CallbackHandler callback = pass == null ? null : new Krb5CallbackHandler(pass);
```
utle, line 158: I just wonder why we do not use WSCallbackHandlerImpl.java instead of Krb4CallbackHandler?

aguibert: I didn't know we had `WSCallbackHandlerImpl`, I could use that but it looks like we'd need to also supply a username for that callback, and here I just want to supply a password specifically for `PasswordCallback`s

aguibert:
build is good and ready to merge once review is approved

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/KerberosService.java
aguibert: will change this to `WSCallbackHandlerImpl` in a follow-on item

PR #13364, opened 2020-08-06
Use multiple startup attempts trying to start DB2 containers

aguibert:
#build


aguibert:
cancelling this PR as it did not fix the issue

PR #13367, opened 2020-08-06
Use 3 startup attempts for starting DB2 container

aguibert:
#build


aguibert:
cancelling this PR as it did not fix the issue

PR #13377, opened 2020-08-07
Improve testReadinessEndpointOnServerStart test case to allow for retry

lamkavan:
**Issue**
The current implementation of the testReadinessEndpointOnServerStart test case will fail the test if the connection is not refused the first time we hit the readiness endpoint. The point of the test is to ensure that the behavior of the readiness endpoint is correct while the server is still starting up. This issue only occurs in the rare case where the server starts faster than we can hit the readiness endpoint (Note: we are starting the server and hitting the readiness endpoint in parallel). 

**Solution**
The structure of the test is changed so that instead of failing the test we execute a retry with a max of 5 retries. After 5 tries the test will then be skipped. This change ensures that we do not fail the test when it is non-conclusive. 

The latest green build is here https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_dN8_UNvbEeqigNlDxrUK9A


lamkavan:
#build

lamkavan:
#build

lamkavan:
#build

lamkavan:
#run-libby-bot

PR #13405, opened 2020-08-10
Handle retryable kafka exceptions gracefully

### LINKED ISSUE #13404: Kafka connector can report failure for acknowledgements which eventually succeed
**Describe the bug**

When `Message.ack()` is called, the kafka connector returns a result as a `CompletionStage` which completes when the kafka commit operation has completed successfully.

In some cases, the kafka commit operation can fail and report a retriable exception. Currently, the Kafka connector fails the `CompletionStage` with this retriable exception, but then goes on to run a commit operation for a later offset, causing the earlier messages to be succesfully commit.

**Steps to Reproduce**
Cause the kafka connector to report a retriable exception

**Expected behavior**
The `ack()` result should not complete until the commit operation has been retried and either fails with a fatal exception or succeeds.

**Diagnostic information:**
 - OpenLiberty Version: current master

**Additional context**
Found through automated testing

Azquelt:
If a Kafka commit operation fails with a retryable exception, we should
retry the commit operation if necessary and should not report a failure
to Message.ack().

Fixes #13404


Azquelt:
#build

Azquelt:
#run-libby-bot
#build

PR #13406, opened 2020-08-10
Support additional signature algorithms in FAT tests

c00crane:
Add tests to support additional signature algorithms in the builder, consumer and mp.jwt 1.1..


c00crane:
going to close this once builder/consumer are delivered via #13410 and mpJwt are delivered via a yet to be named pull.

Issue #13411, opened 2020-08-10
testContinuationFrameAfterDataFrame incorrectly waits for header response


loriadi:
Sequence of events at the server:
1) Received request with upgrade header
2) Sent 101 upgrade
3) Read magic preface
4) Finished the rest of the settings frames to complete upgrade
5) Started working on the response to 1
6) Interrupted with reading a data frame
7) Interrupted again with a continuation frame
8) Send GOAWAY, protocol error
9) Back to the initial response, try to send it, but connection is closed

Need to update the testcase to not expect the HEADERS frame from the initial response as it may never come.

PR #13414, opened 2020-08-10
13411: update TestContFrameAfterDataFrame test

loriadi:
Sequence of events at the server:

Received request with upgrade header
Sent 101 upgrade
Read magic preface
Finished the rest of the settings frames to complete upgrade
Started working on the response to 1
Interrupted with reading a data frame
Interrupted again with a continuation frame
Send GOAWAY, protocol error
Back to the initial response, try to send it, but connection is closed
Need to update the testcase to not expect the HEADERS frame from the initial response as it may never come.


loriadi:
fixes #13411

loriadi:
#build

loriadi:
#build

Issue #13391, opened 2020-08-10
com.ibm.ws.javamail.fat.SMTPTest.testJNDIMail failure


dhaggart:
Failure in testJNDIMail as non-default SMTP port is not being used.

PR #13420, opened 2020-08-11
Revisions to address FAT test failure.

dhaggart:
Added javaPermission elements to resolve Java 2 security errors.

Revised lookup of ports to abort if there is a failure making it clearer
when it is the lookup that fails first.  Added explicit reporting of
looked-up values to aid diagnosis should similar problems occur again.

Changed exception stack reporting to go to System.out in an effort to
ensure failures are not missed in the testing framework logs. (Should be
unnecessary but test failure seems to be missing an exception stack in
order to explain what was seen.)

Removed all the automatically generated TODO comments.


dhaggart:
Testing involved locally forcing SMTP port to 3026 by editing fattest.simplicity/autoFVT-defaults/testports.properties and running the tests - confirming the port is correctly reported by the additional print statements etc.

dhaggart:
#build

dhaggart:
#run-libby-bot

dhaggart:
fixes #13391

PR #13442, opened 2020-08-11
Add clear error message for IBM JDK 8 with Oracle JDBC using kerberos

aguibert:
When Kerberos authentication is used in the Oracle JDBC driver, it goes down a code path that tries to load some JDK internal Kerberos APIs which do not exist on IBM JDK 8, which has its own set of security APIs.

To make this situation more clear to users, detect this combination and emit a clear error message indicating that the combination of IBM JDK 8 and Oracle JDBC w/ Kerberos is not supported.

Also, add support+tests for getting a Kerberos connection with the javax.sql.DataSource interface (previously only ConnectionPoolDataSource and XADataSource worked)


aguibert:
#build

andrea-pichler:
Completed the requested L2 message review

Issue #13447, opened 2020-08-12
Http/2 -clean up connection on error


loriadi:
A previous attempt at fixing this threw an ioException when this condition occurred.  This cause an FFDC in the server log.
Instead of throwing an exception, a log message is issued, and the code should just return so the connection, request, and response are cleaned up.

PR #13448, opened 2020-08-12
13447: Clean up H2 connection

loriadi:
This problem is a http/2 timing issue where different threads are handling requests on a connection.  An error on one thread causes the connection to be cleaned up, and then a second thread wakes up only to find all it's resources are gone.

A previous attempt at fixing this threw an ioException when this condition occurred. This causes an FFDC in the server log.
Instead of throwing an exception, a log message is issued, and the code should just return so the connection, request, and response are cleaned up.


loriadi:
fixes #13447

loriadi:
#build

loriadi:
The build break in the test is not related to this change.

PR #13401, opened 2020-08-10
Issue 13400: Adjust message for other JDKs Acme config test

kristip17:
Fixes #13400 

If the test is running on OpenJDK, sometimes the messages don't include the SocketTimeoutException. For example,

```
com.ibm.ws.security.acme.internal.AcmeConfigService          E CWPKI2016E: The ACME service request for an existing account from the ACME certificate authority at the https://localhost:32785/dir URI failed. The error is 'sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target'.
com.ibm.ws.security.acme.internal.AcmeConfigService          E CWPKI2016E: The ACME service request for an existing account from the ACME certificate authority at the https://localhost:32785/dir URI failed. The error is 'java.net.SocketTimeoutException: Read timed out'.
```

I thought this just happened on OpenJDK, but it seems more widespread. I'm going to make this more generic.

PR #13461, opened 2020-08-12
Issue 13459: Boulder container improvements

### LINKED ISSUE #13459: acmeCA-2.0: BoulderContainer improvements
We can ignore this `com.ibm.ws.security.acme.fat.AcmeRevocationTest:com.github.dockerjava.api.exception.ConflictException: {'message':'a prune operation is already running'}` on BoulderContainer

Also, our wait for a concurrent process to finish isn't long enough.

For #9017

kristip17:
Fixes #13459 

- After running with the last Boulder changes, I saw some tests correctly do the "is anyone else running check" and then hit the 6 minute retry timeout while the existing Boulder was still running (having started a few minutes before). Adding a longer retry
- Also added a longer timeout on the final Boulder container startup, just in case
- Also hit an error where the prune request threw a parallel execution exception, which I think we can just ignore.


kristip17:
#build

kristip17:
#run-libby-bot

kristip17:
Build failure was `AcmeSimpleTest` that is fixed in #13401. The 10xfull_fat that I was started ran partially and was stopped early due to an infrastructure problem, but the Boulder tests that ran were successful.

PR #13463, opened 2020-08-12
13453: Increase default H2 test timeout

### LINKED ISSUE #13453: Adjust test timeout value for Http/2
The current test timeout value is 10 seconds, which proves to be too small for some slow machines and environments.  Adjust the timeout value to 29 seconds to allow for these slower environments.

loriadi:
fixes #13453


loriadi:
#build

Issue #13488, opened 2020-08-13
Reorder the testUnknownFrameType H2 testcase


loriadi:
This purpose of this test is to send a frame which has an unknown type to the server, where the server should ignore it.

Currently the client side log shows the following:
1) Connection and upgrade happen normally
2) Ping frame sent
3) Ping frame received
4) Client determines that the ping was the last frame and sends a goaway which causes the connection to close.
5) The test continues and the bad frame is sent
6) Client gets IOException, connection aborted by host

loriadi:
I also found that this test was being run multiple times, let's just run it once in full mode.

loriadi:
The fix will be to send the bad frame type frame before the ping, then when the expected ping response comes back, it will be the last expected frame and it will be OK for the client to end the connection.

PR #13492, opened 2020-08-13
13488: Reorder the unknown frame type test

loriadi:
The purpose of this test is to send a frame which has an unknown type to the server, where the server should ignore it.

Currently the client side log shows the following:

1) Connection and upgrade happens normally
2) 200 response received
3) Ping frame sent
4) Ping frame received
5) Client determines that the ping was the last expected frame and sends a goaway which causes the connection to close
6) The test continues and the bad frame is sent
7) Client gets IOException, connection aborted by host

This fix is to reorder the testcase to send the frame with the bad type before the ping.
Then when the ping response is received, it is actually the last expected frame, the client can then close things up.


loriadi:
fixes #13488

loriadi:
#build

Issue #13422, opened 2020-08-11
Doc - remove "version" from the subsystem names of the springBoot features.


tjwatson:
The springBoot features have the following Subsystem-Name headers specified:

`Spring Boot Support version 1.5`
`Spring Boot Support version 2.0`

The produces an awkward TOC and headers for the Open Liberty documentation:

https://openliberty.io/docs/20.0.0.8/reference/feature/springBoot-2.0.html

The "version" string included in the Subsystem-Name should be removed to follow the pattern other Open Liberty features used for their names.

PR #13497, opened 2020-08-13
Correct springboot feature subsystem names

sebratton:
The springboot feature Subsystem-Names are not standard. While
they are not referenced elsewhere in the code, they are used by
tooling to generate supporting documentation and result in incorrect
output.


sebratton:
fixes #13422

sebratton:
#build

Issue #13500, opened 2020-08-13
testConnectMethodError test end prematurely


loriadi:
The testConnectMethodError test sends a CONNECT request that has an error.  We expect a reset on that stream..  It's the only frame that the test expected. 

When the test client receives the last expected frame, it sends a goaway.  This test needs to continue after receiving the reset.  To accomplish this, also add a ping as an expected frame and send the ping as the last frame in the test.

PR #13501, opened 2020-08-13
13500: Update the connectMethod tests to avoid ending prematurely

loriadi:
The client framework decides to end a test either when it gets a goaway or gets the last expected frame.  In this case, the last expected frame comes early, and the test framework sends a goaway frame when that happens.  That leaves the rest of the test hanging.

To prevent this, add in a ping frame as expected, and send a ping at the last minute.


loriadi:
fixes #13500

loriadi:
#build

PR #13523, opened 2020-08-17
Make searching END line more specific to avoid intermittent FAT failures

Yushan-Lin:
The test fails intermittently because it expects no logs containing END in message log. Sometimes the build directory contains END which causes the FAT to fail intermittently.
e.g. `/home/jazz_build/_ut7mENDLEeqmIcsdsdUKbw-EBC.PROD.WASRTC-6I0-000-00-00/`
https://github.ibm.com/was-liberty/WS-CD-Open/issues/14988


Yushan-Lin:
#build

PR #13533, opened 2020-08-17
13532: Turn on trace for problematic H2 tests

loriadi:
Turning on trace for 5 tests that fail intermittently.


loriadi:
fixes #13532

loriadi:
#build

PR #13570, opened 2020-08-19
Issue #13596: URBridge needs to return the security name so that it a…

### LINKED ISSUE #13569: Federated basicRegistry returns inconsistent results for case insensitive direct user lookups in scim-1.0
SCIM returns inconsistent results when the `ignoreCaseForAuthentication` attribute for `basicRegistry` is set to `true`.

For reference, the following basic registry configuration:

```
<basicRegistry realm="BasicRealm" ignoreCaseForAuthentication="true">
  <user name="scimUser" password="password"/>
  <group name="scimGroup">
    <member name="scimUser"/>
  </group>
</basicRegistry>
```

Take two similar requests below, the first searching `scimUser` (matching the case in the configuration) and the second searching for `scimuser`. Both calls return results that match the case sent on the search (i.e. id, userName, displayName, location). Additionally, the second search does not return the groups.

```
bash$ curl -sk -u "scimuser:password" https://localhost:9443/ibm/api/scim/Users/scimUser?attributes=id,userName,displayName,emails,groups | json_reformat 
{
    "id": "scimUser",
    "schemas": [
        "urn:scim:schemas:core:1.0"
    ],
    "userName": "scimUser",
    "displayName": "scimUser",
    "externalId": null,
    "groups": [
        {
            "value": "scimGroup",
            "display": "scimGroup"
        }
    ],
    "meta": {
        "location": "https://localhost:9443/ibm/api/scim/Users/scimUser"
    }
}

bash$ curl -sk -u "scimuser:password" https://localhost:9443/ibm/api/scim/Users/scimuser?attributes=id,userName,displayName,emails,groups | json_reformat 
{
    "id": "scimuser",
    "schemas": [
        "urn:scim:schemas:core:1.0"
    ],
    "userName": "scimuser",
    "displayName": "scimuser",
    "externalId": null,
    "meta": {
        "location": "https://localhost:9443/ibm/api/scim/Users/scimuser"
    }
}
```

The error is due to URBridge logic in federated repositories. It searches for the user in the federated basic registry, and since ignore case is turned on for authentication, the user is found. However; federated repositories URBridge doesn't use the security name passed back by the registry and instead uses the security name passed in to populate the results. It also uses the security name passed in to search for groups, which is not effected by the `ignoreCaseForAuthentication` configuration attribute.

The URBridge logic was updated to populate the entity from the registry. It will also use the security name returned from the registry to perform the group search so that groups will be returned.

jvanhill:
…lways matches the user and group in the registry.

fixes #13569

### DIFF
--- a/dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/adapter/urbridge/URBridge.java
+++ b/dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/adapter/urbridge/URBridge.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2019 IBM Corporation and others.
+ * Copyright (c) 2012, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -222,7 +222,7 @@ public class URBridge implements Repository {
      *
      * @return
      */
-
+    @SuppressWarnings("unchecked")
     public void initialize(Map<String, Object> configProps) throws WIMException {
         try {
             reposId = (String) configProps.get(KEY_ID);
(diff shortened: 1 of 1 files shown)


jvanhill:
#build

kristip17:
My comment to add is doing an SCIM test, but that has do be done in a separate PR.

PR #13588, opened 2020-08-21
release connection after checking http status code

benjamin-confino:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=275920 affected a WS-CD bucket, but the bug is in OL but hidden so this PR ports the fix.


benjamin-confino:
The last update after the build is just updating the copyright header.

benjamin-confino:
#run-libby-bot

PR #13614, opened 2020-08-24
Update Yoko binaries with support for IIOP transmission of Supplemental Multilingual Plane characters (such as emoji) in (wide) Strings

### LINKED ISSUE #13613: Support IIOP transmission of Supplemental Multilingual Plane characters (such as emoji) in (wide) Strings
Transmission of String containing Supplemental Multilingual Plane characters (those whose codepoints cannot be encoded in a single 2-byte value in UTF-16) is not currently supported by the Yoko ORB used by Liberty.

This change will update the version of Yoko used by Liberty to one that does support the transmission of these characters.

ngmr:
Fixes #13613


ngmr:
#build

dhaggart (approved):

Issue #13464, opened 2020-08-12
Unregister service listeners added by jndi feature when feature is deactivated


sebratton:
In dev/com.ibm.ws.jndi/src/com/ibm/ws/jndi/internal/WSContextFactory.java a new service listener is added:

`userContext.getBundle(Constants.SYSTEM_BUNDLE_LOCATION).getBundleContext().addServiceListener(this, JNDIServiceBinder.OSGI_JNDI_SERVICE_FILTER);`

Code needs to be added to unregister this listener when the jndi feature is removed.

PR #13666, opened 2020-08-25
Unregister jndi global service listener

sebratton:
A global system listener was added to the jndi component. It needs to be removed when the jndi feature is removed

OpenLiberty Pull Requester,


sebratton:
#build

sebratton:
fixes #13464

PR #13667, opened 2020-08-25
prepare EJB timers failover test for Jakarta EE

njr-11:
It will be necessary for EJB timers that are created in Java EE and persisted in a database to continue to be usable after the application is transformed to Jakarta EE.  A convenient way to test this is to take advantage of an existing EJB timer failover bucket with multiple servers, and make one of the servers remain at Java EE and the other server switch to Jakarta EE.  If timers are able to fail over between the two servers, it will help show that the persisted timers are compatible with both.  At this point, the support for Jakarta EE isn't there yet for EJB timers, but we can update the test bucket to prepare for it.


njr-11:
#build

PR #13688, opened 2020-08-26
Disable dynamic basic user registry tests

utle:
Disable dynamic basic UR tests to avoid intermittently the UR service is not ready.


utle:
#build

PR #13697, opened 2020-08-27
Fix await timing issue in EJB timer FAT

tkburroughs:
Test creates 2 timers, then waits for one to run, which cancels itself.....
then checks for 1 remaining timer.  Problem is that the cancel hasn't
completed yet by the time the main thread checks for timers.

Added some time delay to allow for cancel to complete.


tkburroughs:
#build

Issue #13572, opened 2020-08-19
MP JWT 1.2: Support using MP Config to specify the name of the cookie that contains the JWT token


teddyjtorres:
As a developer, I want to use MP Config to specify the name of the cookie that contains the JWT token.

PR #13788, opened 2020-09-03
Mp jwt 1.2 header and cookie name FAT

c00crane:
Add FAT test for mpJwt 1.2 Header and Cookie attributes.
Test with the new attributes in/as:
server.xml
microprofile-config.properties in META-INF
microprofile-config properties under WEB-INF
environment variables
system properties
Also test that server.xml values over-ride others.


c00crane:
#13571
#13572

c00crane:
#build

c00crane:
using a different pull to deliver these changes and additional changes.

PR #13808, opened 2020-09-04
Issue 12807: AcmeHistory updates and debug

### LINKED ISSUE #13807: acmeCA-2.0: AcmeHistory updates and startup debug
Based on a build break failure, add some debug in case we recreate the same failure. While I was reviewing code, found some try/finally/close changes to add to AcmeHistory. Also added some additional trace.

kristip17:
Fixes #13807

Adding debug for a build break that we encountered that looked like the `acmeFile = wslocation.getServerWorkareaResource("acmeca/" + acmeFileName).asFile();` call might be blocked. I walked the path for that call and no blocking calls jumped out at me. I was unable to recreate after running server starts/stops several times on my local box.

I added debug to the AcmeFatUtils that will take a javacore if we hit the same failure spot as the build break.

While reviewing AcmeHistory, I also added some debug and moved some closes to try/finally blocks.


kristip17:
#build

kristip17:
Verified that we ran lite and full fat on acme_fat successfully.

jvanhill (approved):

PR #13811, opened 2020-09-05
Connectors 2.0 Jakarta Generic Testing

KyleAure:

### DIFF
--- a/dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/FATSuite.java
+++ b/dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/FATSuite.java
@@ -20,7 +20,10 @@ import com.ibm.ws.jca.fat.app.DependantApplicationTest;
 import com.ibm.ws.jca.fat.app.JCATest;
 
 import componenttest.rules.repeater.FeatureReplacementAction;
+import componenttest.rules.repeater.JakartaEE9Action;
 import componenttest.rules.repeater.RepeatTests;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.impl.LibertyServerFactory;
 
 @RunWith(Suite.class)
 @SuiteClasses({
@@ -29,7 +32,21 @@ import componenttest.rules.repeater.RepeatTests;
                 ConnectionManagerMBeanTest.class
 })
 public class FATSuite {
+
+    public static final String javaeeServer = "com.ibm.ws.jca.fat";
+    public static final String jakartaeeServer = "com.ibm.ws.jca.fat.jakarta";
+
     @ClassRule
     public static RepeatTests r = RepeatTests.withoutModification()
-                    .andWith(FeatureReplacementAction.EE8_FEATURES());
+                    .andWith(FeatureReplacementAction.EE8_FEATURES())
+                    .andWith(new JakartaEE9Action());
+
+    public static LibertyServer getServer() {
+        if (JakartaEE9Action.isActive()) {
+            return LibertyServerFactory.getLibertyServer(jakartaeeServer);
+        } else {
+            return LibertyServerFactory.getLibertyServer(javaeeServer);
+        }
+
+    }
 }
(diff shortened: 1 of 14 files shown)


dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/FATSuite.java
```diff
@@ -29,7 +32,21 @@
...
-}
\ No newline at end of file
+                    .andWith(FeatureReplacementAction.EE8_FEATURES())
+                    .andWith(new JakartaEE9Action());
+
+    public static LibertyServer getServer() {
+        if (JakartaEE9Action.isActive()) {
+            return LibertyServerFactory.getLibertyServer(jakartaeeServer);
+        } else {
+            return LibertyServerFactory.getLibertyServer(javaeeServer);
```
njr-11, line 48: Normally the feature replacement repeat action would update the server for you. Is there a reason that doesn't work in this case?

dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/app/JCATest.java
```diff
@@ -134,27 +101,28 @@ public static void tearDown() throws Exception {
...
     }
 
     @Test
     public void testDestinations() throws Exception {
-        runInServlet("testDestinations", fvtweb);
+        runTest();
     }
 
     @Test
+    @SkipForRepeat(EE9_FEATURES) //Transformer does not seem to transform jars inside of jars
```
njr-11, line 118: It's fine to skip the test, but this seems like something to report to the team that is working on the transformer if you haven't already.

njr-11 (approved):

dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/FATSuite.java
KyleAure: The feature replacement action just replaces features, but in this case the server.xml file has configurations like: 
```
    <jmsActivationSpec id="FVTMessageDrivenBeanBindingOverride">
      <properties.FAT1 destinationRef="topic1" destinationType="javax.jms.Topic"/>
    </jmsActivationSpec>
```

Instead of trying to run the transformer on the server.xml file, I figured it would be easier and quicker just to make a copy and change the configuration manually.

dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/app/JCATest.java
KyleAure: 👍 Will do!

KyleAure:
#build 
#libbybot

dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/FATSuite.java
njr-11: That's a good point.  I'm sure making the copy is easier for now, but it will be more difficult in the long term to maintain.  With two copies, we will inevitably end up with developers making a fix or improvement in only one of the files and they will get out of sync.  It seems like it ought to be something the EE9 replacement action can do to update config element values automatically with the simplicity config support in place.  Does that not apply here because this is generated metatype/config?  Otherwise it ought to be possible.

KyleAure: Opened this as a new issue here: https://github.com/OpenLiberty/open-liberty/issues/14374

KyleAure:
Personal build results never posted back to this PR.
Internal Link: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_DXIHQAf5EeuIgpJAZoWiKQ
Results: 1 Failure from an existing defect in a different component.

PR #13829, opened 2020-09-08
Implicit Converter Cache

tevans78:
MP Config 1.2, 1.3 and 1.4 automatically create converters for types that have appropriate methods. This PR aims to cache them in a way which will not cause a memory leak.

#build

### DIFF
--- a/dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ConstructorFunction.java
+++ b/dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ConstructorFunction.java
@@ -1,0 +1,64 @@
+/*******************************************************************************
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.microprofile.config12.converters;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.function.Function;
+
+import com.ibm.websphere.ras.annotation.Trivial;
+import com.ibm.ws.ffdc.annotation.FFDCIgnore;
+import com.ibm.ws.microprofile.config.interfaces.ConversionException;
+
+public class ConstructorFunction<X> implements Function<String, X> {
+    private final Constructor<X> constructor;
+
+    public ConstructorFunction(Constructor<X> constructor) {
+        this.constructor = constructor;
+    }
+
+    /** {@inheritDoc} */
+    @SuppressWarnings("unchecked")
+    @Override
+    public X apply(String value) {
+        X converted = null;
+        if (value != null) { //if the value is null then we always return null
+            try {
+                converted = constructor.newInstance(value);
+            } catch (InvocationTargetException e) {
+                Throwable cause = e.getCause();
+                if (cause instanceof IllegalArgumentException) {
+                    throw (IllegalArgumentException) cause;
+                } else {
+                    throw new ConversionException(cause);
+                }
+            } catch (IllegalAccessException | InstantiationException e) {
+                throw new ConversionException(e);
+            }
+        }
+        return converted;
+    }
+
+    @FFDCIgnore(NoSuchMethodException.class)
+    @Trivial
+    public static <X> Function<String, X> getConstructorFunction(Class<X> reflectionClass) {
+        Function<String, X> implicitFunction = null;
+        try {
+            Constructor<X> ctor = reflectionClass.getConstructor(String.class);
+            implicitFunction = new ConstructorFunction<X>(ctor);
+        } catch (NoSuchMethodException e) {
+            //No FFDC
+        }
+
+        return implicitFunction;
+    }
+
+}
--- a/dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ConstructorFunction.java
+++ b/dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ConstructorFunction.java
@@ -26,8 +26,8 @@ public class ConstructorFunction<X> implements Function<String, X> {
     }
 
     /** {@inheritDoc} */
-    @SuppressWarnings("unchecked")
     @Override
+    @FFDCIgnore(InvocationTargetException.class)
     public X apply(String value) {
         X converted = null;
         if (value != null) { //if the value is null then we always return null
--- a/dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
+++ b/dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
@@ -35,33 +35,45 @@ public class ImplicitConverter extends BuiltInConverter {
         this.implicitFunction = getImplicitFunction(converterType);
     }
 
+    /**
+     * <p>If no explicit Converter and no built-in Converter could be found for a certain type,
+     * the {@code Config} provides an <em>Implicit Converter</em>, if</p>
+     * <ul>
+     * <li>The target type {@code T} has a public Constructor with a String parameter, or</li>
+     * <li>the target type {@code T} has a {@code public static T valueOf(String)} method, or</li>
+     * <li>the target type {@code T} has a {@code public static T parse(CharSequence)} method</li>
+     * </ul>
+     *
+     * @param converterType The class to convert using
+     */
     @Trivial
     protected <X> Function<String, X> getImplicitFunction(Class<X> converterType) {
         Function<String, X> implicitFunction = ConstructorFunction.getConstructorFunction(converterType);
 
+        if (implicitFunction != null && TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+            Tr.debug(tc, "Automatic converter for " + converterType + " using \"constructor\"");
+        }
+
         if (implicitFunction == null) {
             implicitFunction = MethodFunction.getValueOfFunction(converterType);
 
-            if (implicitFunction == null) {
-                implicitFunction = MethodFunction.getParseFunction(converterType);
-
-                if (implicitFunction == null) {
-                    throw new IllegalArgumentException(Tr.formatMessage(tc, "implicit.string.constructor.method.not.found.CWMCG0017E", converterType));
-                } else {
-                    if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                        Tr.debug(tc, "Automatic converter for " + converterType + " using \"parse\"");
-                    }
-                }
-            } else {
-                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                    Tr.debug(tc, "Automatic converter for " + converterType + " using \"valueOf\"");
-                }
-            }
-        } else {
-            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                Tr.debug(tc, "Automatic converter for " + converterType + " using \"constructor\"");
+            if (implicitFunction != null && TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "Automatic converter for " + converterType + " using \"valueOf\"");
             }
         }
+
+        if (implicitFunction == null) {
+            implicitFunction = MethodFunction.getParseFunction(converterType);
+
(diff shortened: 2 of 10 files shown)


dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
```diff
@@ -28,117 +33,196 @@
 public class ImplicitConverter extends BuiltInConverter {
```
jhanders34, line 33: I assume some of these base classes like BuiltinConverter and ImplicitConverter could move to the common bundle at some point?

jhanders34 (changes requested):

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
```diff
@@ -28,117 +33,196 @@
...
     }
 
     /** {@inheritDoc} */
     @Override
     public String toString() {
         return "Implicit Converter for type " + getType();
     }
+
+    protected static class ConstructorFunction<X> implements Function<String, X> {
+        private final static Map<Class<?>, Constructor<?>> constructorCache = Collections.synchronizedMap(new WeakHashMap<>());
```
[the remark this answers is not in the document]
jhanders34: I guess this comment is more important in the Manager if we don't keep the constructor / method ones.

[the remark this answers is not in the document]
tevans78: I don't think there is such a thing as a ConcurrentWeakHashMap??

You told me that if the ImplicitConverter had a reference to the Constructor or Methods then the WeakHashMap cache in the Manager would not be enough. The Method would hold a reference to the Class and the WeakReference in the cache would never get removed?

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
```diff
@@ -28,117 +33,196 @@
...
+        private final WeakReference<Class<?>> classRef;
+
+        protected ConstructorFunction(Class<X> conversionType, Constructor<X> ctor) {
+            constructorCache.put(conversionType, ctor);
```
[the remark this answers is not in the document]
tevans78: hmmm yep, fair point ... any suggestions?

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
tevans78: Do you mean `io.openliberty.microprofile.config.internal.common`? That is for the common classes between our original impl and the SmallyRye impl. These changes don't affect the SmallRye impl.

tevans78: I don't remember why that bundle is separate from `io.openliberty.microprofile.config.internal.serverxml`

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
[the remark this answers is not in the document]
jhanders34: There is not a ConcurrentWeakHashMap.  You can easily use a ConcurrentHashMap with a WeakReference impl key.  I showed examples of how to do that via slack earlier.  

Right now I see you look up ever time and didn't have a reference to the Method / Constructor in the function which is what I thought it was doing because that would make sense.  Instead I would suggest that the Function have a WeakReference to the Method / Constructor and you just do a get on the WeakReference to get the Method / Constructor instead of using yet another WeakHashMap which wouldn't work anyway because the value is the Constructor / Method which will keep a hard reference on the Class.  The only way to use the WeakHashMap would be to create a WeakReference on the Constructor / Method to use it as the value so it is best just to not use the WeakHashMap.

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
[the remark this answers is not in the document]
jhanders34: I answered in the review comment above.  Hopefully it will be sufficient.

tevans78:
#build

tevans78:
#build

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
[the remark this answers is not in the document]
tevans78: The WeakHashMap stuff was too hard to make work efficiently. So I gave up on the static caches and went with a simple cache on the `Config12ConversionManager`. This is scoped by the Config itself so when the application is stopped, the Config is removed and this cache will go too.

tevans78:
#run-libby-bot

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/MethodFunction.java
```diff
@@ -0,0 +1,86 @@
...
+                    throw (IllegalArgumentException) cause;
+                } else {
+                    throw new ConversionException(cause);
+                }
+            } catch (IllegalAccessException e) {
+                throw new ConversionException(e);
+            }
+        }
+        return converted;
+    }
```
Azquelt, line 49: Should this have `@FFDCIgnore` since we're throwing the exception up to the user anyway?

An exception here presumably means they had a method or constructor to create an object from a String, but calling it threw an exception?

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/converters/Config13ImplicitConverter.java
```diff
@@ -43,138 +42,60 @@
...
+                implicitFunction = ConstructorFunction.getConstructorFunction(converterType);
+                if (implicitFunction == null) {
+                    implicitFunction = MethodFunction.getParseFunction(converterType);
+                    if (implicitFunction == null) {
+                        throw new IllegalArgumentException(Tr.formatMessage(tc, "implicit.string.constructor.method.not.found.CWMCG0017E", converterType));
+                    } else {
+                        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                            Tr.debug(tc, Tr.formatMessage(tc, "Automatic converter for " + converterType + " using \"parse\""));
+                        }
+                    }
```
Azquelt, line 65: Don't much like these nested ifs either.

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/converters/Config13ImplicitConverter.java
```diff
@@ -43,138 +42,60 @@
...
+            return super.convert(value);
+        } catch (ConversionException e) { //The Config 1.3 spec clarified that the convert method should throw IllegalArgumentException
+                                          //if the value cannot be converted to the specified type
+            Throwable cause = e.getCause();
+            if (cause instanceof IllegalArgumentException) {
+                throw (IllegalArgumentException) cause;
+            } else {
+                throw new IllegalArgumentException(cause);
             }
         }
```
Azquelt, line 98: Should this method have `@FFDCIgnore`? At the moment we would get an FFDC from the `MethodFunction` and then another one here. We shouldn't need both, right?

dev/com.ibm.ws.microprofile.config.1.4/src/com/ibm/ws/microprofile/config14/converters/Config14ImplicitConverter.java
```diff
@@ -46,135 +44,41 @@
...
+    protected <X> Function<String, X> getImplicitFunction(Class<X> converterType) {
+        Function<String, X> implicitFunction = null;
+
+        implicitFunction = MethodFunction.getOfMethod(converterType);
+        if (implicitFunction == null) {
+            implicitFunction = MethodFunction.getValueOfFunction(converterType);
+            if (implicitFunction == null) {
+                implicitFunction = MethodFunction.getParseFunction(converterType);
+                if (implicitFunction == null) {
+                    implicitFunction = ConstructorFunction.getConstructorFunction(converterType);
```
Azquelt, line 58: More nested ifs

Azquelt (commented):
Raised some questions.

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
```diff
@@ -10,130 +10,65 @@
...
+            }
+        } else {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "Automatic converter for " + converterType + " using \"constructor\"");
             }
-        } catch (NoSuchMethodException e) {
-            //No FFDC
         }
-        return method;
+        return implicitFunction;
```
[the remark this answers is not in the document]
tevans78: That certainly looks better but I will do it in a follow up.

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/MethodFunction.java
tevans78: We should probably ignore `InvocationTargetException` but still FFDC for `IllegalAccessException`

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/converters/Config13ImplicitConverter.java
tevans78: Should probably ignore this one ... FFDCs should occur at the point of first failure; this is just re-wrapping the exception, it doesn't add anything to FFDC here.

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ConstructorFunction.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class ConstructorFunction<X> implements Function<String, X> {
+    private final Constructor<X> constructor;
+
+    public ConstructorFunction(Constructor<X> constructor) {
+        this.constructor = constructor;
+    }
+
+    /** {@inheritDoc} */
+    @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
tevans78: removed

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/ImplicitConverter.java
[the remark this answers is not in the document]
tevans78: done in 2nd commit

dev/com.ibm.ws.microprofile.config.1.4/src/com/ibm/ws/microprofile/config14/converters/Config14ImplicitConverter.java
tevans78: done

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/converters/Config13ImplicitConverter.java
tevans78: done

dev/com.ibm.ws.microprofile.config.1.3/src/com/ibm/ws/microprofile/config13/converters/Config13ImplicitConverter.java
tevans78: done

dev/com.ibm.ws.microprofile.config.1.2/src/com/ibm/ws/microprofile/config12/converters/MethodFunction.java
tevans78: done

tevans78:
#build

Azquelt (approved):

tevans78:
Failures in PB were caused by network/infra problems ... the config FAT buckets all passed.
#run-libby-bot

jdmcclur:
This fix looks good.Time spent in my primitive went from 11.5% -> 0.31%.

PR #13875, opened 2020-09-10
[13874] FAT test case for `ResourceInfo`

### LINKED ISSUE #13874: Test case for JAX-RS ResourceInfo
Add a test case that verifies that `ResourceInfo`'s `getResourceClass()` and `getResourceMethod()` returns the correct target resource class/method in a few different scenarios:
1) For a simple resource class
2) For a abstract resource class case where the `@GET` annotation is on the abstract class/method, but the target method is in the extending sub-class.
3) For an interface-based resource class case where the `@GET` annotation is on the interface/method, but the target method is in the implementing class.

The test should verify that the same results are returned whether the `ResourceInfo` object is injected into a request filter, response filter, or the resource class itself.  It should also verify that it returns null when injected into a pre-matching request filter.

andymc12:
Fixes #13874 - test case for `ResourceInfo`


andymc12:
#build

andymc12:
Added one more test - #build

jim-krueger (approved):
LGTM

andymc12:
#run-libby-bot

andymc12:
#run-libby-bot run!

Issue #10089, opened 2019-12-09
gRPC Client Support


wtlucy:
Implement gRPC client (outbound) functionality

Issue #10088, opened 2019-12-09
gRPC Server Support


wtlucy:
Implement gRPC server (inbound) functionality

PR #13902, opened 2020-09-11
Update gRPC bundles

wtlucy:
* fix javadoc generation for new API jar
* create stable API jar for stable API packages
* change Netty APIs to third party and create third-party jar

for #10088 #10089

### DIFF
--- a/dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.internal.grpc-1.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.internal.grpc-1.0.feature
@@ -2,17 +2,18 @@
 symbolicName=io.openliberty.internal.grpc-1.0
 singleton=true
 IBM-API-Package: \
-  com.google.common.collect; type="stable",\
-  com.google.common.util.concurrent; type="stable",\
-  com.google.protobuf;  type="stable",\
+  com.google.common.collect; type="internal",\
+  com.google.common.util.concurrent; type="internal",\
+  com.google.protobuf;  type="internal",\
   io.grpc;  type="stable", \
-  io.grpc.protobuf;  type="stable", \
-  io.grpc.stub;  type="stable", \
-  io.grpc.stub.annotations;  type="stable", \
+  io.grpc.protobuf;  type="internal", \
+  io.grpc.stub;  type="internal", \
+  io.grpc.stub.annotations;  type="internal", \
   io.grpc.internal; type="internal"
 Subsystem-Name: gRPC internal 1.0
 -bundles=\
-  io.openliberty.grpc.1.0.internal.common
+  io.openliberty.grpc.1.0.internal.common, \
+  io.openliberty.io.grpc.1.0; location:="dev/api/stable/,lib/"; mavenCoordinates="io.grpc:grpc-api:1.31.1"
 -features=\
   com.ibm.websphere.appserver.artifact-1.0, \
   com.ibm.websphere.appserver.classloading-1.0, \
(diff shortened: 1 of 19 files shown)


wtlucy:
#build

wtlucy:
#build

dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.internal.grpc-1.0.feature
```diff
@@ -2,16 +2,18 @@
 symbolicName=io.openliberty.internal.grpc-1.0
 singleton=true
 IBM-API-Package: \
-  com.google.common.util.concurrent; type="stable",\
-  com.google.protobuf;  type="stable",\
+  com.google.common.collect; type="internal",\
+  com.google.common.util.concurrent; type="internal",\
+  com.google.protobuf;  type="internal",\
   io.grpc;  type="stable", \
```
Azquelt, line 8: I thought that usually the IBM-API-Package was listed on the public API feature (so that the packages are only exposed to users if the public feature is enabled, and not if the private feature is pulled in by another feature).

I'm assuming in this case, this feature is only used by `grpc` and `grpcClient`?

dev/io.openliberty.grpc.1.0/bnd.bnd
```diff
@@ -11,19 +11,20 @@
...
+Bundle-SymbolicName: io.openliberty.grpc.1.0
 
-Import-Package: io.openliberty.grpc.annotation
+Import-Package: \
+  io.grpc, \
+  io.openliberty.grpc.annotation
 
-Export-Package: io.openliberty.grpc.annotation
+Export-Package: \
+  io.openliberty.grpc.annotation
```
Azquelt, line 23: This project seems slightly odd. It looks like it's not used at all as a bundle and just copies in the `io.openliberty.frpc.annotation` package so that it can be put in the `dev/api` directory as a .jar?

I think that's ok, just seemed slightly odd.

dev/io.openliberty.grpc.1.0.internal.common/bnd.bnd
```diff
@@ -41,7 +47,7 @@ Export-Package: \
   io.openliberty.grpc.internal.GrpcManagedObjectProvider
 
 Include-Resource: \
-  @${repo;io.grpc:grpc-core;1.31.1;EXACT}!/META-INF/**, \
+  @${repo;io.grpc:grpc-core;${grpcVersion};EXACT}!/META-INF/**, \
```
Azquelt, line 50: Trailing `, \` on the end of this line

Azquelt (commented):
I've left a couple of questions and there's one trailing comma which will I think will just give you a warning in your bnd file. Otherwise I think it looks good.

dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.internal.grpc-1.0.feature
wtlucy: You're right, this private feature is only used by the public features `grpc-1.0` and `grpcClient-1.0`.  My intention here was to provide the common APIs for those features from a single internal feature.

dev/io.openliberty.grpc.1.0/bnd.bnd
wtlucy: The only purpose of this package is to provide the IBM API for `grpc-1.0`, which is just a single package (and class!)  It is a little odd, but from the examples I looked at this is the correct way to create the `dev/api` jar.

dev/io.openliberty.grpc.1.0.internal.common/bnd.bnd
wtlucy: good catch!

Azquelt (approved):

billw1024 (approved):

wtlucy:
#build

wtlucy:
#run-libby-bot

PR #13935, opened 2020-09-14
Add do-priv blocks for the restful web services 3.0 implementation

WhiteCat22:
@tevans78 can you review dev/com.ibm.ws.microprofile.config.1.1.cdi/src/com/ibm/ws/microprofile/config/cdi/ConfigCDIExtension.java please?

### DIFF
--- a/dev/com.ibm.ws.microprofile.config.1.1.cdi/src/com/ibm/ws/microprofile/config/cdi/ConfigCDIExtension.java
+++ b/dev/com.ibm.ws.microprofile.config.1.1.cdi/src/com/ibm/ws/microprofile/config/cdi/ConfigCDIExtension.java
@@ -13,6 +13,8 @@ package com.ibm.ws.microprofile.config.cdi;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Optional;
@@ -60,7 +62,13 @@ public class ConfigCDIExtension implements Extension, WebSphereCDIExtension {
 
     void processInjectionTarget(@Observes ProcessInjectionTarget<?> pit) {
         Class<?> targetClass = pit.getAnnotatedType().getJavaClass();
-        ClassLoader classLoader = targetClass.getClassLoader();
+
+        ClassLoader classLoader = AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
+            @Override
+            public ClassLoader run() {
+                return targetClass.getClassLoader();
+            }
+        });
 
         for (InjectionPoint injectionPoint : pit.getInjectionTarget().getInjectionPoints()) {
             ConfigProperty configProperty = ConfigProducer.getConfigPropertyAnnotation(injectionPoint);
--- a/dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
+++ b/dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
@@ -24,6 +24,8 @@ public class MediaTypeHeaderDelegate implements RuntimeDelegate.HeaderDelegate
    private static Map<MediaType, String> reverseMap = new ConcurrentHashMap<MediaType, String>();
    private static final int MAX_MT_CACHE_SIZE =
                    // Liberty Change
+                   // This has been fixed in Resteasy 4.5.7.Final. We can remove this class
+                   // when we upgrade Resteasy to 4.5.7.Final or newer.
                    AccessController.doPrivileged(new PrivilegedAction<Integer>() {
                        @Override
                        public Integer run() {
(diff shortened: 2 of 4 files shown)


WhiteCat22:
#build

dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
```diff
@@ -0,0 +1,190 @@
...
+ * @version $Revision: 1 $
+ */
+public class MediaTypeHeaderDelegate implements RuntimeDelegate.HeaderDelegate
+{
+   public static final MediaTypeHeaderDelegate INSTANCE = new MediaTypeHeaderDelegate();
+
+   private static Map<String, MediaType> map = new ConcurrentHashMap<String, MediaType>();
+   private static Map<MediaType, String> reverseMap = new ConcurrentHashMap<MediaType, String>();
+   private static final int MAX_MT_CACHE_SIZE =
+                   // Liberty Change
```
andymc12, line 26: this has been fixed in RESTEasy, but not yet GA - should be in 4.5.7 when it comes out. maybe mention that in the comment so we know that we can delete this overlay once we pull in 4.5.7 or newer?

dev/com.ibm.ws.microprofile.config.1.1.cdi/src/com/ibm/ws/microprofile/config/cdi/ConfigCDIExtension.java
```diff
@@ -60,7 +62,13 @@ protected void addBadInjectionType(Type type) {
 
     void processInjectionTarget(@Observes ProcessInjectionTarget<?> pit) {
         Class<?> targetClass = pit.getAnnotatedType().getJavaClass();
-        ClassLoader classLoader = targetClass.getClassLoader();
+
+        ClassLoader classLoader = AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
+            @Override
+            public ClassLoader run() {
+                return targetClass.getClassLoader();
```
andymc12, line 69: I think I saw this same call to get classloader in other versions of MP Config (besides just 1.1).  It might be good to update them all - when I looked, I saw it in 1.4, but it's a good guess that it needs to be updated in 1.2 and 1.3 too.

andymc12 (commented):
Couple comments - both minor.  Thanks for catching these Adam!

dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
tevans78: If this has been fixed in the resteasy master branch, would it be better to build a snapshot version from there and put it in DHE ... rather than creating a new overlay!?

dev/com.ibm.ws.microprofile.config.1.1.cdi/src/com/ibm/ws/microprofile/config/cdi/ConfigCDIExtension.java
tevans78: I think Andy is right, there is another getClassLoader in `Config14CDIExtension`. The 1.2 and 1.3 versions don't have much in since they just inherit from this class.

WhiteCat22: I've updated ```Config14CDIExtension``` as well, thanks for pointing it out.

dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
WhiteCat22: Updated the comment. @tevans78 We plan to upgrade to the newer version of Resteasy before release, these overlay changes are are to enable us to do rapid development. This is currently how we override our CXF implementation for ```jaxrs-2.0``` and ```jaxrs-2.1```

We discussed having our own CXF fork and pulling that into OpenLiberty on today's design issues call. I assume we'll have a similar discussion about Resteasy in the future.

WhiteCat22:
#run-libby-bot
#build

dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
tevans78: Rapid development now but at what service cost if the overlays are never removed ... which is exactly what has happened with CXF. Your aim should be to avoid a fork or overlay unless there is no other choice. In this case there is a choice, the change is already in RESTEasy.

WhiteCat22: I don't disagree, but I think this conversation is beyond the scope of this PR.

andymc12 (approved):
LGTM - thanks @WhiteCat22

PR #13944, opened 2020-09-15
Fix MP Config 20

tevans78:
Fix build break due to incorrect annotations version when only feature running.


benjamin-confino (approved):

tevans78:
Merging without PB

PR #13966, opened 2020-09-16
Writing Events and Files for new way of publishing results

websterc87:
Adding support into the gradle build to product events and files for when we run tests. This allows us develop a new way of collecting the results rather than having the RTC launcher guessing at information


websterc87:
#build

SylvainGarcon (approved):
code works having run through prototype external contributor build

PR #14029, opened 2020-09-18
Fix Max version and enable more client tests

cbridgha:
Removed the repeat action JAKARTAEECLIENT-9.0 because the Normal EE9 feature repeat is now doing the right swap...  

Enabled rest of app client tests - and fixed max version to 9


cbridgha:
#build

jimblye (approved):

cbridgha:
#build

cbridgha:
#build

Issue #14035, opened 2020-09-18
SpringBootThinUtil.execute() can cause zip file exceptions if multiple threads try to thin at the same time


tjwatson:
I saw the error message:

`[ERROR   ] CWWKM0101E: The system could not process archive data for /home/jazz_build/Build/jbe/build/dev/image/output/wlp/usr/servers/com.ibm.ws.springboot.support.fat.SpringBootTests/workarea/spring/lib.index.cache/fb/53f8539e7fcb8f093a56e138112056ec1dc809ebb020b59d8a36a5ebac37e0/logback-classic-1.2.3.jar.`

And the FFDC:

```
Stack Dump = com.ibm.wsspi.adaptable.module.UnableToAdaptException: Unable to get the container for the entry /BOOT-INF/lib/logback-classic-1.2.3.jar
	at com.ibm.ws.app.manager.springboot.internal.LibIndexCache.getLibraryContainer(LibIndexCache.java:105)
	at com.ibm.ws.app.manager.springboot.internal.SpringBootApplicationImpl.getStoredIndexClassesInfos(SpringBootApplicationImpl.java:787)
	at com.ibm.ws.app.manager.springboot.internal.SpringBootApplicationImpl.getContainerInfos(SpringBootApplicationImpl.java:743)
	at com.ibm.ws.app.manager.springboot.internal.SpringBootApplicationImpl.<init>(SpringBootApplicationImpl.java:529)
	at com.ibm.ws.app.manager.springboot.internal.SpringBootApplicationFactory.createDeployedAppInfo(SpringBootApplicationFactory.java:100)
	at com.ibm.ws.app.manager.springboot.internal.SpringBootHandler.setUpApplicationMonitoring(SpringBootHandler.java:163)
	at com.ibm.ws.app.manager.internal.statemachine.StartAction.execute(StartAction.java:139)
	at com.ibm.ws.app.manager.internal.statemachine.ApplicationStateMachineImpl.enterState(ApplicationStateMachineImpl.java:1317)
	at com.ibm.ws.app.manager.internal.statemachine.ApplicationStateMachineImpl.run(ApplicationStateMachineImpl.java:897)
	at com.ibm.ws.threading.internal.ExecutorServiceImpl$RunnableWrapper.run(ExecutorServiceImpl.java:239)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #14046, opened 2020-09-21
Only allow one thread to thin at a time

tjwatson:
When multiple Spring Boot applications are configured
(an invalid config) it is possible that the library cache
where the JARs are persisted from a thin operation can
collide causing invalid ZIP exceptions to occur.

This change ensures that this cannot happen.


tjwatson:
#build

Fixes #14035

tjwatson:
#build

tjwatson:
#run-libby-bot

sebratton (approved):

PR #14049, opened 2020-09-21
Add variable to configure coordinator context root

idlewis:
This can be overriden by the user, as per normal liberty config variables


idlewis:
For the moment, decision taken that this does not need to be configurable

PR #14062, opened 2020-09-21
[14050] Allow RESTEasy bundle to be used as a client in a Java SE env

### LINKED ISSUE #14050: RESTful WS TCK: NPE when creating client instances
Brian posted the following in slack while trying to run the RESTful WS 3.0 TCK:
> Also seeing some tests that NPE, like so:
```
Caused by: java.lang.NullPointerException
       at io.openliberty.org.jboss.resteasy.common.client.LibertyResteasyClientBuilderImpl.getBundleContext(LibertyResteasyClientBuilderImpl.java:56)
       at io.openliberty.org.jboss.resteasy.common.client.LibertyResteasyClientBuilderImpl.build(LibertyResteasyClientBuilderImpl.java:35)
       at io.openliberty.org.jboss.resteasy.common.client.LibertyResteasyClientBuilderImpl.build(LibertyResteasyClientBuilderImpl.java:28)
       at jakarta.ws.rs.client.ClientBuilder.newClient(ClientBuilder.java:88)
       at com.sun.ts.tests.jaxrs.common.client.JaxrsWebTestCase.getClientWithRegisteredProviders(JaxrsWebTestCase.java:254)
       ...
```
> I had to add the common jar to the test client's classpath in order to get that far, so it's entirely possible that I need to add additional stuff and it's just not failing in an obvious way.

I believe that the error is occurring because we added some code to inject Liberty's SSL config into the client creation process, and that depends on OSGI Declarative Services, which means it will only work when running inside the Liberty server.  We need to make that support optional so that it will work in a Java SE environment (with no OSGI / DS).

andymc12:
Fixes #14050 

Fixes the NPE (which occurs because there is no OSGi environment to provide Bundles/BundleContexts/etc.).
Adds a test case to verify that a client can be created in a Java SE environment.
Updates the bnd file to handle unit tests and add to the test path to include everything needed to build a client in a Java SE env.
Adds minimal MP Config implementation required in order to work with RESTEasy config.

### DIFF
--- a/dev/io.openliberty.org.jboss.resteasy.common/resources/META-INF/services/org.eclipse.microprofile.config.spi.ConfigProviderResolver
+++ b/dev/io.openliberty.org.jboss.resteasy.common/resources/META-INF/services/org.eclipse.microprofile.config.spi.ConfigProviderResolver
@@ -1,0 +1,2 @@
+# Intended only for Java SE environments - should NOT be exposed by the ResourceProvider in the bnd file
+io.openliberty.restfulWS.client.ConfigProviderResolverImpl
--- a/dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/client/ConfigProviderResolverImpl.java
+++ b/dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/client/ConfigProviderResolverImpl.java
@@ -1,0 +1,95 @@
+/*******************************************************************************
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.restfulWS.client;
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Optional;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigBuilder;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+/**
+ * Only intended to be used in Java SE client environments.
+ */
+public class ConfigProviderResolverImpl extends ConfigProviderResolver {
+    
+    private final boolean java2SecurityEnabled = System.getSecurityManager() != null;
+
+    private final Config config = new Config() {
+
+        @Override
+        public Iterable<ConfigSource> getConfigSources() {
+            // TODO Auto-generated method stub
+            return null;
+        }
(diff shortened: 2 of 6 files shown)


andymc12:
#build

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/client/ConfigProviderResolverImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+        @Override
+        public <T> Optional<T> getOptionalValue(String propName, Class<T> type) {
+            return Optional.ofNullable(getValue(propName, type));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public Iterable<String> getPropertyNames() {
+            return (Iterable<String>) System.getProperties().keySet().iterator();
```
WhiteCat22, line 45: This is more of a question. I know this is an SE environment only, but do we need do-privs when getting system properties? If this wasn't SE would we need them?

dev/io.openliberty.org.jboss.resteasy.common/resources/META-INF/services/org.eclipse.microprofile.config.spi.ConfigProviderResolver
```diff
@@ -0,0 +1,2 @@
+# Intended only for Java SE environments - should NOT be exposed by the ResourceProvider in the bnd file
```
WhiteCat22, line 1: do we need to explicitly NOT export this package in bnd?

andymc12: The package is fine to export.  We just not explicitly NOT add this file to the `app_resources` list of META-INF files that we "publicize" via the ResourceProvider service - i.e.
```
app-resources= \
  META-INF/services/javax.ws.rs.client.ClientBuilder | \
  META-INF/services/javax.ws.rs.sse.SseEventSource\$Builder | \
  META-INF/services/javax.ws.rs.sse.SseEventSource.Builder | \
  META-INF/services/javax.ws.rs.ext.RuntimeDelegate | \
  META-INF/services/javax.ws.rs.ext.Providers
```

If we do, then things might go awry when users try to use the client inside Liberty servers.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/client/ConfigProviderResolverImpl.java
andymc12: Good catch.  Our FATs and the TCK most likely wouldn't test with a Java 2 security manager in a Java SE environment (and this code should never run outside of a Java SE env), but it is good practice to wrap it in a doPriv block.  I'll update.

andymc12:
Added a doPriv - new #build

WhiteCat22 (approved):

andymc12:
#run-libby-bot

andymc12:
1 failure, in an unrelated bucket

PR #14068, opened 2020-09-21
Make builder wait for SSL init to complete

c00crane:
Update MP Jwt Fat so that it waits for SSL to complete its initialization before running any tests.

#build


ayoho (approved):

c00crane:
Added a wait for the ssl ready message - all affected tests ran 100% successfully.

Issue #13965, opened 2020-09-16
LRA Coordinator and Participant default Endpoints do not match 


yongja79:
Currently the default endpoint of the coordinator is: ../lrac/lra-coordinator
where as the participant is sending to: ../lra-coordinator
These need to match so that they will talk together by default

idlewis:
Fixed by #14083

PR #14083, opened 2020-09-22
Change default coordinator location to match

idlewis:
For an Open Liberty coordinator with default config

This should mean that a coordinator and participant with default config should talk to each other


yongja79 (approved):

idlewis:
#build

idlewis:
fixes #13965

PR #14087, opened 2020-09-22
14086: Update testPriorityFrameAfterHeaderFrameNoEndHeaders

loriadi:
This test was creating an error which caused the server to send a goaway frame.

After that, the test tried to send another unnecessary frame, which caused an intermittent problem where on occasion, the frame could not be sent because the connection was already closed.  The PR removes the extra send frame.

This is the error seen in the client log when the frame cannot be sent.
IOException while doing IO requested on local: 127.0.0.1/127.0.0.1:62462 remote: localhost/127.0.0.1:8030
[9/15/20, 8:06:02:144 BST] 0000004f id=00000000 com.ibm.ws.tcpchannel.internal.WorkQueueManager              1 Exception is: java.io.IOException: An established connection was aborted by the software in your host machine


loriadi:
#build

loriadi:
The failures in the build are unrelated to the http/2 test change.

wtlucy (approved):
LGTM, thanks!

PR #14099, opened 2020-09-22
Issue #14098: Convert the com.ibm.ws.security.authentication.filter b…

jvanhill:
…undle for Jakarta EE9 as there is some work going on to add authentication filter to some features that have already been transformed, so this bundle needs to be transformed as well.

FYI @utle


jvanhill:
#build

kristip17 (approved):

jvanhill:
The RTC build had a single failure which was due to an unrelated FAT bucket over running the allowed log sizes. This is a prexisting issue.

PR #14117, opened 2020-09-23
Issue #13830: Replace String.valueOf usage with a new method that wil…

### LINKED ISSUE #13830: Federated repositories returns the string "null" instead of the value null for several methods
The following Bridge class methods may return the String "null" instead of the value null. These can lead to UserRegistry calls of the same name returning the same "null" String.

- DisplayNameBridge.getGroupDisplayName(String)
- DisplayNameBridge.getUserDisplayName(String)
- LoginBridge.checkPassword(String,String)
- LoginBridge.mapCertificate(X509Certificate[])
- MembershipBridge.getUniqueGroupIds(String)
- SearchBridge.getGroups(String, int)
- SearchBridge.getUsers(String, int)
- SecurityNameBridge.getGroupSecurityName(String)
- SecurityNameBridge.getUserSecurityName(String)
- UniqueIdBridge.getUniqueGroupId(String)
- UniqueIdBridge.getUniqueUserId(String)

The fix should be to replace any `String.valueOf(Object)` calls with a call to a new method in BridgeUtils that returns a null value if the passed in object is null, instead of the String "null". For example:

```
public static String getStringValueOfOrNull(Object o) {
   return (o == null) ? null : String.valueOf(o);
}
```

jvanhill:
…l return the null value instead of the string null.

fixes #13830

### DIFF
--- a/dev/com.ibm.ws.security.wim.registry/src/com/ibm/ws/security/wim/registry/util/BridgeUtils.java
+++ b/dev/com.ibm.ws.security.wim.registry/src/com/ibm/ws/security/wim/registry/util/BridgeUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2019 IBM Corporation and others.
+ * Copyright (c) 2012, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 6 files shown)


jvanhill:
#build

jvanhill:
#build

jvanhill:
Pushed copyright date changes. They were pushed before the build was out of the pending state in the build queue, so the previous build should include them.

jvanhill:
#libby

ryanesch (approved):

PR #14122, opened 2020-09-23
Rework KerberosTicket refresh mechanism

aguibert:
#build

### DIFF
--- a/dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
+++ b/dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
@@ -10,14 +10,14 @@
  *******************************************************************************/
 package com.ibm.ws.security.kerberos.auth.internal;
 
-import java.net.InetAddress;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Collections;
-import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 
+import javax.security.auth.RefreshFailedException;
 import javax.security.auth.Subject;
 import javax.security.auth.kerberos.KerberosPrincipal;
 import javax.security.auth.kerberos.KerberosTicket;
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
```diff
@@ -84,37 +86,38 @@ public Subject get(KerberosPrincipal principal) {
...
+                        break;
                     }
+                } else {
+                    if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
+                        Tr.debug(tc, "no match because ticket is not current and not renewable");
+                    return null;
                 }
             }
-            if (oldTicket != null && renewedTicket != null) {
-                kerbTickets.remove(oldTicket);
```
aguibert, line 115: This remove/add code is what needed to be replaced, since `sub.getPrivateCredentials(KerberosTicket.class)` only returns a copy of the credential set, not the actual set of credentials

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
```diff
@@ -84,37 +86,38 @@ public Subject get(KerberosPrincipal principal) {
...
-                if (!kerbTicket.isCurrent()) {
-                    if (kerbTicket.isRenewable()) {
-                        // If the ticket is not current but is renewable, try to renew the ticket if
-                        // there is at least 10m remaining on the renewTill time
-                        long renewTillMs = kerbTicket.getRenewTill().getTime();
-                        long timeLeft = renewTillMs - System.currentTimeMillis() - TimeUnit.MINUTES.toMillis(10);
-                        if (timeLeft < 0) {
+                if (kerbTicket.isCurrent()) {
+                    // we have a valid ticket
+                    break;
```
utle, line 92: If the kerberos ticket is current but almost expired, should we try to refresh it?

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
utle: ok, so we need to clone the Kerberos ticket and refresh/renew as needed.

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
aguibert: yes good point, I'll fix this

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/internal/LRUCache.java
aguibert: If we cloned the ticket and refreshed the cloned ticket then we would need to mutate the set of private credentials on the subject, which I hesitate to do because that could impact existing resources that are using this Subject.

Instead, I think we should just call `kerbTicket.refresh()` and not clone/replace it. I verified in the JDK implementation of `KerberosTicket.refresh()` that the ticket does properly synchronize on itself and clone and re-init its internal state during the refresh operation.

aguibert:
#build

utle (approved):

aguibert:
#run-libby

PR #14027, opened 2020-09-18
Rename private servlet

anjumfatima90:
for issue #13242

PR #14172, opened 2020-09-25
Revert "Rename private servlet"

cbridgha:
Reverts OpenLiberty/open-liberty#14027


mhldr (approved):
looks like a revert to me.

PR #14180, opened 2020-09-26
Mark SPNEGO krb5Config and krb5Keytab attributes internal

aguibert:
Mark the `krb5Config` and `krb5Keytab` attributes as internal on the `<spnego>` config element. The attributes will still be usable by existing configurations, but will no longer be visible in documentation.

Instead of these attributes, users should use the `configFile` and `keyTab` attributes on the new `<kerberos>` element.

#build


aguibert:
discussed this with Ut and we agreed it would be better to leave the attributes in place and update the description to be deprecated

PR #14204, opened 2020-09-28
None as a configurable mpcontext type

njr-11:
MP Config 2.0 is getting rid of type ability to specify an empty list/array/String as a value for config properties. This will now indicate unspecified.  MP Context Propagation is adding a new way to specify an empty context list -- the value "None" -- under https://github.com/eclipse/microprofile-context-propagation/issues/160 / https://github.com/eclipse/microprofile-context-propagation/pull/198 .

This pull is open to align OpenLiberty with that by supporting the value None.  Make sure not to merge this pull until the corresponding MP Context Propagation spec pull is merged first.


jmstephensgit (approved):

njr-11:
#build

Issue #10090, opened 2019-12-09
gRPC Configuration


wtlucy:
Implement server.xml configuration for client and server cases

PR #14279, opened 2020-10-01
grpcClient config update

wtlucy:
The `host` parameter in `<grpcClient/>` should be required so that users are required to know the scope of a given `grpcClient` config. This will be consistent with the `target` configuration in `<grpc/>`. For #10090


wtlucy:

billw1024 (approved):

PR #14254, opened 2020-09-30
Update JPA 3.0 API to RC2 and Eclipselink 3.0 to 879de6b.

jgrassel:
Update the Jakarta Persistence APIs to Release Candidate 2 and Eclipselink 3.0 to master-879de6b.  Add com.ibm.ws.jpa_30_fat to test using v3.0 persistence.xml schema.

PR #14314, opened 2020-10-05
Revert "Update JPA 3.0 API to RC2 and Eclipselink 3.0 to 879de6b."

mhldr:
Reverts OpenLiberty/open-liberty#14254 I think it's causing Defect 278491
com.ibm.websphere.appserver.thirdparty.eclipselink.3.0:compileJava FAILED


jgrassel:
Was not actually reverted from master, as the binaries are on DHE, so closing.

PR #14339, opened 2020-10-06
Move ThirdPartyServer to in memory DB

benjamin-confino:
For https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=278454

I am hopeful that this test failed because something went wrong with the HD and so moving the DB to memory will prevent it happening again. 

Even if I'm wrong memory DBs are the best practice for CDI fats.


benjamin-confino:
#build

scottcurtis2605 (approved):

Issue #11271, opened 2020-03-11
create Store test microservice application(s) for grpc feature


anupag:
The test application will

1) Http request flow from REST Endpoint -> grpc client  -> grpc server and vice versa.
2) request flow from one grpc service to other grpc service
3) Have security enabled at REST Endpoint as well as grpc server

anupag:
#12587

PR #14357, opened 2020-10-06
gRPC test app improvements

wtlucy:
* change the scope of shared objects in the test Servlets to fix request concurrency issues
* package additional packages in the test apps so that they can be run in a standalone environment
* fix a case where a countdown latch was never hit

for #11271


wtlucy:
#build

billw1024 (approved):
look's good

wtlucy:
#run-libby-bot

PR #14378, opened 2020-10-07
Implement config_ordinal consistently for built in sources

### LINKED ISSUE #14377: Server.xml config sources do not respect config_ordinal
**Describe the bug**
The spec allows the ordinal of a built in config source to be changed by defining a property named `config_ordinal` which is read from that config source
> Note that a special property config_ordinal can be set within any built-in ConfigSource implementation. The default implementation of getOrdinal() will attempt to read this value. If found and a valid integer, the value will be used. Otherwise the respective default value will be used.

However, our built in config sources for server.xml variables and `<appProperties>` do not change their ordinal if a property with `config_ordinal` is set.

Azquelt:
Currently some of our config sources respect `config_ordinal` and some don't and each source has its own logic to look up this variable.

Make the code that does this common and use it for all built in config sources.

Fixes #14377

### DIFF
--- a/dev/com.ibm.ws.microprofile.config.1.1/src/com/ibm/ws/microprofile/config/sources/EnvConfigSource.java
+++ b/dev/com.ibm.ws.microprofile.config.1.1/src/com/ibm/ws/microprofile/config/sources/EnvConfigSource.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 IBM Corporation and others.
+ * Copyright (c) 2017, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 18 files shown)


Azquelt:
#build
#run-libby-bot

dev/com.ibm.ws.microprofile.config.1.1/src/com/ibm/ws/microprofile/config/sources/EnvConfigSource.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 IBM Corporation and others.
+ * Copyright (c) 2017, 2020 IBM Corporation and others.
```
Emily-Jiang, line 2: 2018, 2020

Emily-Jiang: ignore my last comment. I missed the 2017, bit when I first reviewed.

Emily-Jiang (approved):

dev/com.ibm.ws.microprofile.config.1.1/src/com/ibm/ws/microprofile/config/sources/EnvConfigSource.java
Azquelt: I think this change is correct, 2017 is the original year (which stays the same) while 2020 is the year of the most recent change.

PR #14384, opened 2020-10-07
allow components to add additional features when using EE9 repeat action

andymc12:
Allow FAT projects to use repeat actions with EE9 features and add additional features - i.e. jaxrs-2.1 -> restfulWS-3.0 + jsonb-2.0

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
+++ b/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
@@ -101,6 +101,7 @@ public class FeatureReplacementAction implements RepeatTestAction {
     private final Set<String> clients = new HashSet<>(Arrays.asList(ALL_CLIENTS));
     private final Set<String> removeFeatures = new HashSet<>();
     private final Set<String> addFeatures = new HashSet<>();
+    private final Set<String> alwaysAddFeatures = new HashSet<>();
     private TestMode testRunMode = TestMode.LITE;
 
     public FeatureReplacementAction() {}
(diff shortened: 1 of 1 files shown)


andymc12:
#build

jim-krueger (approved):
LGTM:  Thanks Andy!

dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
```diff
@@ -417,6 +383,45 @@ public void setup() throws Exception {
             LibertyClientFactory.getLibertyClient(clientName);
     }
 
+    protected void swizzleFeatures(Set<String> features) {
```
aguibert, line 386: Looks like no change in the code here, just breaking it out into its own method right?

aguibert: Suggested name change: `swizzelFeatures` --> `replaceFeatures`

dev/fattest.simplicity/src/componenttest/rules/repeater/JakartaEE9Action.java
```diff
@@ -133,6 +135,11 @@ public JakartaEE9Action addFeatures(Set<String> addFeatures) {
         return (JakartaEE9Action) super.addFeatures(addFeatures);
     }
 
+    public JakartaEE9Action addAdditionalFeatures(Set<String> additionalFeatures) {
```
aguibert, line 138: If we are going to add this new API I think it makes sense to add it in `FeatureReplacementAction` so it can apply to all subclasses, rather than only putting it on the `JakartaEE9Action`

aguibert: Could we do a different name for this method? Without knowing the background it's not clear what "additional features" mean. The purpose of this new method is to "add a feature that will always be added to configs" (also referred to as "force adding"). So perhaps name change `addAdditionalFeatures` --> `forceAddFeature` or `alwaysAddFeature`. Another option could be overloading `addFeatures` with a boolean param for force add like: `addFeatures(Set<String>, boolean forciblyAdd)`

Whatever method signature we go with, javadoc please!

dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java
andymc12: Right - I had originally moved that to its own method in order to extend only that functionality.  If we make the change to the base class as you suggested later, then we no longer to split it out.

aguibert (approved):

andymc12:
#build

andymc12:
#run-libby-bot

PR #14397, opened 2020-10-08
Introduce transformation of server.xml

### LINKED ISSUE #14374: Transform server.xml for jakarta testing
Introduce a method to allow developers to transform their server.xml when repeating tests using jakarta.

Currently, the [FeatureReplacementAction](https://github.com/OpenLiberty/open-liberty/blob/integration/dev/fattest.simplicity/src/componenttest/rules/repeater/FeatureReplacementAction.java) swaps out features.  But in the case of the `JakartaEE9Action` a helper method to replace `javax.*` string in server.xml to `jakarta.*` would be helpful for configurations like the one here:

https://github.com/OpenLiberty/open-liberty/blob/80e12eb6c090e9c53de6488d0eb3a51ac4e3fa48/dev/com.ibm.ws.jca_fat/publish/servers/com.ibm.ws.jca.fat/server.xml#L54-L56

Where we are providing class names in server config.

KyleAure:
Initial prototype of a method we could use to transform `server.xml` files when a string literal has a `javax` class name. 

Fixes #14374


dev/fattest.simplicity/src/componenttest/rules/repeater/JakartaEE9Action.java
```diff
@@ -177,8 +190,11 @@ public void setup() throws Exception {
         }
         ShrinkHelper.cleanAllExportedArchives();
 
-        // Transform server.xml's
         super.setup();
+
+        if (transformServerConfig)
+            for (File serverXml : super.getReplacementServers())
```
KyleAure, line 196: This may be a bit inefficient as we need to call the transformer multiple times in cases when there are multiple server.xml's.  I wonder if we can provide a folder with all the server.xml's instead.

dev/wlp-jakartaee-transform/rules/jakarta-xml-master.properties
```diff
@@ -5,6 +5,7 @@ basic-constraints.xml=jakarta-xml-dd.properties
 ejb-jar.xml=jakarta-xml-dd.properties
 permissions.xml=jakarta-xml-dd.properties
 ra.xml=jakarta-xml-dd.properties
+server.xml=jakarta-xml-server.properties
```
KyleAure, line 8: Will this transform files that end with server.xml?  For example there are test suites that user <method-name>-server.xml

dev/fattest.simplicity/src/componenttest/rules/repeater/JakartaEE9Action.java
dazavala: Adding a folder would unnecessarily impart structure on the FAT project or require gradle changes.  I'm finding that few of the server and client configurations (let's just say `server.xml` files) in a FAT suite need transformation of config elements other than `<featureManager>`.  Invoking the transformer is expensive.  Rather than compute the set `serverXmls` I think the user should specify this set as an argument of method `transformServerConfig()` [or not -- see (*) below.]  The repeat action would also need to know which transform properties to apply to each file in the set.  The user would have to input this information as well, which IMHO adds unnecessary complication.  I think it's better that the user adds the server.xml files to the `jakartaee-dd-master` properties file, and whenever a server.xml file is specified, the repeat action simply invokes existing method `transformApp()` on the file.  This approach ensures the transformer is invoked uniformly across all product and test artifacts -- e.g. product configuration type metatype.xml and any corresponding config elements in server.xml -- as expected per each jakarta technology.  The approach alleviates the user from adding code to transform a server.xml while otherwise maintaining the usage pattern of how all artifacts are transformed. 

(*) There's another benefit to requiring the user to add server.xml files to the jakartaee-dd-master properties file: the need for method `transformServerConfig()` is obviated.  The repeat action can automagically invoke transformApp on any server.xml file it finds in the jakartaee-dd-master properties file. -- i.e. the serverXml set is implicitly configured by the user.

dev/wlp-jakartaee-transform/rules/jakarta-xml-master.properties
dazavala: The new line enables the transformation of files named `server.xml`.

dev/com.ibm.ws.jca_fat/fat/src/com/ibm/ws/jca/fat/app/JCATest.java
```diff
@@ -91,25 +86,6 @@ public static void setUp() throws Exception {
         ShrinkHelper.addDirectory(fvtapp_ear, "lib/LibertyFATTestFiles/fvtapp");
         ShrinkHelper.exportToServer(server, "apps", fvtapp_ear);
 
-        if (JakartaEE9Action.isActive()) {
-            /*
-             * Need to update the destination type of the topic to ensure it matches the Jakarta FQN.
-             */
-            ServerConfiguration clone = server.getServerConfiguration().clone();
```
dazavala, line 98: IMHO adding this kind of transformation to the repeat action implementation -- just after the features are swapped, before the updated configuration is written to file -- is a blazing fast, intuitive solution.  But solution could require specific information like config IDs in order to work correctly, so I didn't pursue it and don't have a strong opinion whether we should.

dev/fattest.simplicity/src/componenttest/rules/repeater/JakartaEE9Action.java
KyleAure: The problem is that the `transformApp()` method is called when an application is shrink-wrapped and deployed to server at runtime. The server.xml file is not a part of that process so adding those server.xml files to `jakartaee-dd-master` would never be found/transformed.

So we still need a different method of collecting a list of server.xml files that need to be transformed, there isn't a way (that I know of) to include this in the current app transformation process. 

Additionally users would still need to ensure the transformations they want are somewhere in the `wlp-jakartaee-transform/rules` folder.

KyleAure:
Abandoning this change as it doesn't seem worthwhile to make.

PR #14406, opened 2020-10-09
Fix testLogstashEvents()

fmhwong:
- Remove GC test in `testLogstashEvents()` as it is already covered in `testLogstashForGCEvent()`


fmhwong:
#build

Yushan-Lin (approved):
LGTM

Issue #14433, opened 2020-10-12
Do not fail Expirer test of no items are garbage collected.


AndrewDJBanks:
Test machines may not run the garbage collector aggressively enough to cause unreferenced items to be garbage collected
before the end of the test.

PR #14435, opened 2020-10-12
If the garbage collector does not collect unreferenced items, record a soft error only

AndrewDJBanks:
In the message store Expiry unit tests, some items are usually dropped, 
however, this is not guaranteed because it depends on the behaviour
of the garbage collector. Don't make it a hard error if no items are dropped.


AndrewDJBanks:
fixes #14433

AndrewDJBanks:
#build

dhaggart (approved):

PR #14441, opened 2020-10-12
Port JCA-1.7 fat

KyleAure:


KyleAure:
So far this is a direct port from what we have in CL.
There are a few "external resource adapters" aka archives that have been checked into git.  Would it be beneficial at all to covert these back to source code in this project for maintenance or are we okay with checking in these archives as is?

KyleAure:
#libbybot

KyleAure:
#build

PR #14495, opened 2020-10-14
fix gRPC service app start and server.xml update window

wtlucy:
There is a very small timing window during app startup which can cause the wrong `<grpc/>` server.xml config to get used. I'll close that window here.


wtlucy:

billw1024 (approved):

PR #14529, opened 2020-10-16
Change concurrent.persistent_fat to use in-memory derby by default

aguibert:
#build

### DIFF
--- a/dev/com.ibm.ws.concurrent.persistent_fat/fat/src/com/ibm/ws/concurrent/persistent/fat/PersistentExecutorWithFailoverEnabledTest.java
+++ b/dev/com.ibm.ws.concurrent.persistent_fat/fat/src/com/ibm/ws/concurrent/persistent/fat/PersistentExecutorWithFailoverEnabledTest.java
@@ -19,9 +19,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.testcontainers.containers.JdbcDatabaseContainer;
 
-import com.ibm.websphere.simplicity.Machine;
 import com.ibm.websphere.simplicity.ShrinkHelper;
-import com.ibm.websphere.simplicity.config.PersistentExecutor;
 import com.ibm.websphere.simplicity.config.ServerConfiguration;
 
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.concurrent.persistent_fat/fat/src/com/ibm/ws/concurrent/persistent/fat/PersistentExecutorWithFailoverEnabledTest.java
```diff
@@ -33,168 +31,135 @@
...
+		ShrinkHelper.defaultDropinApp(server, APP_NAME, "web");
+
+		// Use the Jakarta version of test features if Jakarta is being used.
+		if (JakartaEE9Action.isActive()) {
+			ServerConfiguration config = server.getServerConfiguration();
+			Set<String> features = config.getFeatureManager().getFeatures();
+			features.remove("persistentexecutor-1.0");
+			features.remove("timerinterfacestestfeature-1.0");
+			features.add("persistentExecutor-2.0");
+			features.add("timerInterfacesTestFeature-2.0");
```
KyleAure, line 74: Shouldn't the JakartaEE9Action do this automatically?  Or is it because it the features switched from all lowercase to camel case?

KyleAure (approved):
Looks good, just one question but doesn't actually affect what this PR is trying to accomplish. 😄

dev/com.ibm.ws.concurrent.persistent_fat/fat/src/com/ibm/ws/concurrent/persistent/fat/PersistentExecutorWithFailoverEnabledTest.java
aguibert: Partially yes -- we could update the `JakartaEE9Action` to auto-update the `persistentExecutor` feature, but we shouldn't add the `timerInterfacesTestFeature` to the EE9 action since that is a test feature only used in this bucket.

Issue #11851, opened 2020-04-20
JakartaEE: JMS 3.0 (messaging support) 


atosak:
Child story for epic #9876 

Support the [Jakarta Messaging 3.0 specification](https://jakarta.ee/specifications/messaging/3.0/).

 - [x] Provide new API (`jakarta.messaging-3.0`). 
 - [x] Enable new messaging features for EE 9 -- `messaging-3.0`, `messagingClient-3.0`, `messagingServer-3.0`, `messagingSecurity-3.0` -- along with the supporting services and runtime.
 - [x] Deliver in EE 9 beta 21.0.0.1.
 - [x] Enable EE 9 tests covering javax.jms-2.0, app client, and security requirements (see FAT summary at #12969)
 - [x] Announce in 21.0.0.2 EE 9 beta blog, note feature additions in 21.0.0.1 EE 9 beta blog.

malincoln:
@dazavala  pls add the FAT summary here:  https://github.com/OpenLiberty/open-liberty/issues/12969

dazavala:
Closing the issue as fat-complete. The is issue may reopen depending on FAT summary review or new milestones.

dazavala:
@chackoj I changed the title according to the adopted name convention.

PR #14535, opened 2020-10-16
11851 Fix Open JMS20 FAT MessageEndpoint Tests

dazavala:
Ref: #11851

Configure all server/client activation specs and application MDBs to declare (bind/enable) the message endpoints within all test applications in fat com.ibm.ws.messaging.open_jms20_fat.  Along with a few additional improvements, the FAT now provides a simple pattern to transform all server configurations which require the `javax.` packages to be replaced in JMS resource configuration elements.  The server transformation (server.xml) is the same as that applied to JMS configuration types metadata (metadata.xml).


dazavala:
#build

dev/com.ibm.ws.messaging.open_jms20_fat/fat/src/com/ibm/ws/messaging/JMS20/fat/SharedSubscription/SharedSubscriptionWithMsgSelTest_129623.java
```diff
@@ -85,27 +79,33 @@ private static String getLocalAddress() {
...
-        clientServer.setServerConfigurationFile("SharedSubscriptionDurClient.xml");
+        final String clientXml = "SharedSubscriptionDurClient.xml";
+        if (JakartaEE9Action.isActive()) {
+            Path clientXmlFile = Paths.get(
+                                           "lib/LibertyFATTestFiles",
+                                           clientXml);
+            JakartaEE9Action.transformApp(clientXmlFile);
+            Log.info(c, "setUp", "Transformed server " + clientXmlFile);
+        }
+        clientServer.setServerConfigurationFile(clientXml);
```
tbitonti, line 105: Doesn't the usual server transformation steps handle this file?  Or is it not transformed because it is not under the 'servers' folder?  Also, there are some helper methods, either on LibertyServer or on ShrinkWrapHelper which may do something similar.

tbitonti (approved):

tbitonti:
#libby

dev/com.ibm.ws.messaging.open_jms20_fat/fat/src/com/ibm/ws/messaging/JMS20/fat/SharedSubscription/SharedSubscriptionWithMsgSelTest_129623.java
dazavala: Simplicity does not invoke the transformer on server.xml files, at least not yet.  The most direct, simple solution is to invoke the transformer directly on the server configuration file and include the file name and transformer properties in the transformer property set.  That's what I've done here.  Simplicity can be modified to transform server and client configuration files, and there a couple ways this can be done.  Simplicity could scan the publish/ and publish/server/ directories for xml files and transform all files that match file names declared in a repeat action builder method, where the transformer is called with arg1=properties file containing the file name(s) and arg2=properties file containig match-replace values for all config metatypes.  Alternatively, Simplicity could directly replace matching strings in the ServerConfiguration instance that is instantiated in order to replace features.  When the instance is streamed to file all necessary transformation has been done -- i.e., there's no need to invoke the transformer , which is expensive.

dev/com.ibm.ws.messaging.open_jms20_fat/publish/files/EJBMDB_server.xml
```diff
@@ -20,16 +20,17 @@
   <variable name="onError" value="FAIL"/>
 
   <logging
```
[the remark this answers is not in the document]
dazavala: Will comment out the logging element.  Thanks.

dev/com.ibm.ws.messaging.open_jms20_fat/fat/src/com/ibm/ws/messaging/JMS20/fat/SharedSubscription/SharedSubscriptionWithMsgSelTest_129623.java
KyleAure: Thanks for the heads up!  I have a PR that does option 1 (calls the transformer on server.xml) here: https://github.com/OpenLiberty/open-liberty/pull/14397  but as you mentioned this is an expensive operation to do for each server.xml.  However, it does offer us more control in what we are replacing since not ever instance of `javax` should be replaced (i.e. javax.sql.XADataSource)

dazavala: Yes, that the idea.  Augment the ee9 repeat action to replace strings in config elements other than <FeatureManager> according to, say a new builder method, before the ServerConfiguration is streamed to file.  Any solution will have full control over what is transformed.  Even if the repeat action is modified to perform separate operations to replace features and other config, the two operations are still much less costly than invoking the transformer.  Something to consider.  What I like about using the transformer is the consistency of using it for both config metatype metadata and the server/client configuration metadata.  Thanks!

tbitonti (approved):

tbitonti:
#libby

dazavala:
#build

dazavala:
#build

dazavala:
#libby

dazavala:
The open_jms20_fat passed in the last personal build, which is about to complete.  The transformer property change applies to JMS, only; and all other changes are specific to open_jms20_fat.  Merging now.

PR #14536, opened 2020-10-16
FAT app restart utility should tolerate app updated message

wtlucy:
In a rare case during FAT testing, a `CWWKZ0003I` will be printed instead of a `CWWKZ0001I` after an app is removed from dropins and put back. Assuming the rest of app startup goes correctly, an error shouldn't be thrown for the `CWWKZ0003I` case.


wtlucy:
#build #spawn.fullfat.buckets=com.ibm.ws.jsf.2.2_fat

brenthdaniel (approved):

wtlucy:
The one failing test in the previous build doesn't use the method I've changed, so we can disregard it.

volosied (approved):
look good  to  me

wtlucy:
#run-libby-bot

PR #14547, opened 2020-10-19
Clear container output in the beginning of each test class

fmhwong:
Clear container output in the beginning of each test class so that the output from previous test class will not be read in the following test class.


fmhwong:
#build

Yushan-Lin (approved):
LGTM

PR #14557, opened 2020-10-19
Use ProcessBuilder to redirect and capture standard error

tjwatson:
This test is still failing on some platforms.  Using process
builder to see if we are missing something from the standard error
when invoking java.

The existing logging during the test indicate there is nothing
at all printed to the standard output when the test fails.
This seems to indicate that we may have a problem invoking Java
altogether.


tjwatson:
#build

hlhoots (approved):
looks good to me.

PR #14564, opened 2020-10-19
Ignore CWWKC1503W in EJB Timer FAT

tkburroughs:
The test creates persistent timers and then restarts the server to
verify the behavior; it is normal for CWWKC1503W to occur if an attempt
is made to run a timer while the server is shutting down; ignore.


tkburroughs:
#build

olendvcook (approved):

PR #14576, opened 2020-10-20
Respect config_ordinal for default server.xml vars

### LINKED ISSUE #14388: Respect config_ordinal for server.xml default variables
In #14378, I implemented config_ordinal for server.xml variables and appProperties but deliberately didn't do so for server.xml default variables.

**Background**

[As described in the OpenLiberty docs][ol-doc-variables], variables in liberty can have a `value` and a `defaultValue`. When doing variable substitution in server.xml, a `defaultValue` can be overridden by a value from server.env or bootstrap.properties whereas a `value` cannot.

In MicroProfile Config 1.3+, we have two config sources for server.xml variables, one for the `value` with ordinal 500 and a separate one for the `defaultValue` with ordinal 1.

Our built-in config sources and their ordinals are:
* Server.xml default variables (1)
* META-INF/microprofile-config.properties (100)
* Environment variables (300)
* System properties (400)
* server.xml variables (500)
* server.xml `<appProperties>` (600)

**Reasoning against supporting config_ordinal for server.xml default variables**

server.xml default variables really only make sense if they can be overridden by other config sources, so assigning them the lowest ordinal makes sense. Allowing the user to change this ordinal didn't make sense to me as it would undermine their purpose, particularly as the `config_ordinal` for default variables could be configured to be _higher_ than regular variables!

In addition, I think the syntax for configuring the ordinal is quite unintuitive. Setting `<variable name="config_ordinal" value="250"/>` has quite a different effect to setting `<variable name="config_ordinal" defaultValue="250"/>`.

**Reasoning in favour of supporting config_ordinal for server.xml default variables**

While the bad things listed above are enabled by allowing the config ordinal for default variables to be configured, there are also some valid use cases.

For example, the user might want to configure the ordinal for default variables to `150` so that it overrides `microprofile-config.properties` but can be overridden by everything else.

Since there are valid use cases and a strict reading of the spec suggests that `config_ordinal` should be supported here, it makes sense to implement it and trust that users will set it to a value which is consistent with the purpose of default variables.

[ol-doc-variables]:https://openliberty.io/docs/20.0.0.9/reference/config/server-configuration-overview.html#_variable_substitution

Azquelt:
Fixes #14388

### DIFF
--- a/dev/com.ibm.ws.microprofile.config.1.3/resources/com/ibm/ws/microprofile/config13/resources/Config13.nlsprops
+++ b/dev/com.ibm.ws.microprofile.config.1.3/resources/com/ibm/ws/microprofile/config13/resources/Config13.nlsprops
@@ -47,15 +47,6 @@ service.not.found.CWMCG0204E=CWMCG0204E: An instance of the {0} service could no
 service.not.found.CWMCG0204E.explanation=An internal error occurred and a service could not be found by MicroProfile Config.
 service.not.found.CWMCG0204E.useraction=Review the server message.log file and FFDC logs to identify the problem.
 
-# Server XML Application Properties Config Source
-server.xml.appproperties.config.source=Server XML Application Properties Config Source
-
-# Server XML Variables Config Source
-server.xml.variables.config.source=Server XML Variables Config Source
-
-# Server XML Default Variables Config Source
-server.xml.default.variables.config.source=Server XML Default Variables Config Source
-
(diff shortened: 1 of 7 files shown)


Azquelt:
Messages were in the wrong bundle and have been moved but not changed, no review should be required

Azquelt:
#build
#run-libby-bot

benjamin-confino (approved):

Issue #14462, opened 2020-10-13
Add package org.eclipse.microprofile.lra to participant feature as api


idlewis:
This is due to a new class in M2 of the the spec which has been added to this package, i.e. LRAResponse

PR #14579, opened 2020-10-20
Add new LRA package as API

idlewis:
The M2 release of the LRA spec added a new class in a new package which
needs to be exposed as API


idlewis:
fixes #14462

tevans78 (approved):

idlewis:
#build

PR #14656, opened 2020-10-22
Issue 14655: Only check on baseEntries after config processed

### LINKED ISSUE #14655: Move participatingBaseEntry check to avoid inaccurate logging of CWIMK0004E message
After configuration updates for a `federatedRepository` and user registries (such as a BasicRegistry or LdapRegistry), an incorrect `CWIMK0004E` message is logged indicating that the participating base entry cannot be found even though the registries are defined in the server.xml

For example:
`CWIMK0004E: The user registry operation could not be completed. The participating base entry definition o=SampleBasicRealm is not valid. Correct the participating base entry definition in the server.xml file. For example the syntax is <participatingBaseEntry name='...'/>.` message can be logged incorrectly. 

In some timing situations, the `participatingBaseEntry` list on the `federatedRepository` configuration is checked prior to loading all of the configured user registries and the `CWIMK0004E` message is logged. A fix was made to ensure checking is done after registries are loaded.

kristip17:
Fixes #14655 

The previous behavior was that the ConfigManager sent out a notification to Listeners only after receiving a Modify. When VMMService (the only implemented listener) received the event, it checks that the participatingBaseEntries listed match a registry. If a registry isn't found, an error message is issued (`CWIMK0004E`).

The startup timing has changed such that sometimes we process the config as an empty initialize then call modify when we may or may not have finished processing the registries (such as BasicRegistry or LdapRegistry). This caused a false negative on the error messages.

To match previous behavior and avoid the baseEntry check happening too early. I did the following:
- Add a listener for checking when server config modifications are complete
- Toggle when we're past init phase and modify is being called (otherwise, `URAPIs_RealmPropertyMappingTest` failed because we received a config complete check before initializing the CustomRegistry/CustomRepository features in that test -- I think this matches more closely to the original behavior).
- Added a call to deregister in VMMService or we ended up with a new and old VMMService to call from ConfigManager
- Added a new FAT test that recreates the failing scenario and intentional creates a scenario that should log the `CWIMK0004E`

### DIFF
--- a/dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
+++ b/dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2019 IBM Corporation and others.
+ * Copyright (c) 2012, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -19,6 +19,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Future;
 
 import org.osgi.service.component.ComponentContext;
 import org.osgi.service.component.annotations.Activate;
@@ -26,6 +27,7 @@ import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.ConfigurationPolicy;
 import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Modified;
+import org.osgi.service.component.annotations.Reference;
 
 import com.ibm.websphere.ras.Tr;
 import com.ibm.websphere.ras.TraceComponent;
@@ -33,8 +35,13 @@ import com.ibm.websphere.ras.annotation.Trivial;
 import com.ibm.websphere.security.wim.ras.WIMMessageHelper;
 import com.ibm.websphere.security.wim.ras.WIMMessageKey;
 import com.ibm.ws.config.xml.internal.nester.Nester;
+import com.ibm.ws.runtime.update.RuntimeUpdateListener;
+import com.ibm.ws.runtime.update.RuntimeUpdateManager;
+import com.ibm.ws.runtime.update.RuntimeUpdateNotification;
 import com.ibm.ws.security.wim.util.StringUtil;
 import com.ibm.ws.security.wim.util.UniqueNameHelper;
+import com.ibm.ws.threading.FutureMonitor;
+import com.ibm.ws.threading.listeners.CompletionListener;
 import com.ibm.wsspi.security.wim.SchemaConstants;
 import com.ibm.wsspi.security.wim.exception.InvalidArgumentException;
 import com.ibm.wsspi.security.wim.exception.WIMException;
@@ -47,9 +54,9 @@ import com.ibm.wsspi.security.wim.model.PersonAccount;
 
 @Component(configurationPid = "com.ibm.ws.security.wim.core.config",
            configurationPolicy = ConfigurationPolicy.OPTIONAL,
-           service = { ConfigManager.class },
+           service = { ConfigManager.class, RuntimeUpdateListener.class },
            property = { "service.vendor=IBM", "com.ibm.ws.security.registry.type=WIM", "config.id=default-WIM" })
-public class ConfigManager {
+public class ConfigManager implements RuntimeUpdateListener {
 
     private static final TraceComponent tc = Tr.register(ConfigManager.class);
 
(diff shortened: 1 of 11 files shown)


kristip17:
#build

kristip17:
There was one failure in the testing -- it was a "too much trace" error. I pulled the logs and reviewed. While it is a security test, it wasn't logging any of the new (small) trace that I added. I also searched and found an existing bug where this test is popping too much trace, so I didn't push it over.

kristip17:
#build

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
```diff
@@ -424,10 +434,14 @@ public void registerRealmConfigChangeListener(RealmConfigChangeListener listener
             listeners.add(listener);
     }
 
+    public void deregisterRealmConfigChangeListener(RealmConfigChangeListener listener) {
```
jvanhill, line 437: Add javadoc.

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
```diff
@@ -496,4 +510,56 @@ public long getPageCacheTimeOut() {
     void setRepositoryManager(RepositoryManager repositoryManager) {
         this.repositoryManager = repositoryManager;
     }
+
+    @Trivial
```
jvanhill, line 514: Remove trivial? Add javadoc.

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
```diff
@@ -496,4 +510,56 @@ public long getPageCacheTimeOut() {
...
+    @Reference(service = FutureMonitor.class)
+    protected void setFutureMonitor(FutureMonitor futureMonitor) {
+        _futureMonitor = futureMonitor;
+    }
+
+    @Trivial
```
jvanhill, line 520: Ditto.

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
```diff
@@ -496,4 +510,56 @@ public long getPageCacheTimeOut() {
...
+    /**
+     * After a configuration modification, run a RealConfigChange check which will verify the participatingBaseEntries.
+     *
+     * We used to do this directly from the modify command, but timing change during startup/modify and we would process the check
+     * before the user registries activated, leading to misleading error messages being printed.
+     */
+    @Trivial
+    @Override
+    public void notificationCreated(RuntimeUpdateManager updateManager, RuntimeUpdateNotification notification) {
+        if (modified && notification.getName().equals(RuntimeUpdateNotification.CONFIG_UPDATES_DELIVERED)) {
```
jvanhill, line 534: Reverse the string equality check so the constant is first to protect against possible NPE.

jvanhill (changes requested):

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
kristip17: Done

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
kristip17: Done

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
kristip17: Done

dev/com.ibm.ws.security.wim.core/src/com/ibm/ws/security/wim/ConfigManager.java
kristip17: Done

kristip17:
#build

dev/com.ibm.ws.security.wim.core_fat/publish/servers/com.ibm.ws.security.wim.core.fat.mod/.gitignore
```diff
@@ -0,0 +1 @@
+/dropins
```
jvanhill, line 1: We should be building in the autofvt build directory, so not sure we should need this.

jvanhill (approved):
There is one comment about a gitignore file that I am not sure you need. You can delete it if it is not needed and you shouldn't have to rebuild since it would not effect the build.

dev/com.ibm.ws.security.wim.core_fat/publish/servers/com.ibm.ws.security.wim.core.fat.mod/.gitignore
kristip17: Removed

kristip17:
#libby

PR #14667, opened 2020-10-22
[14650] Ensure WEB-INF/lib JARs are processed for MP GraphQL Components

### LINKED ISSUE #14650: MP GraphQL does not scan JARs in WEB-INF/lib for GraphQL components
Classes annotated with things like `@GraphQLApi`, `@Type`, etc. or types referenced from root level queries that exist in JARs in the WEB-INF/lib directory are not processed.  This means that only classes in the WEB-INF/classes directory are processed by the MP GraphQL runtime.

andymc12:
Fixes #14650 

Can remove overlayed file, IndexInitializer.java once https://github.com/smallrye/smallrye-graphql/pull/487 is released.

### DIFF
--- a/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
+++ b/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
@@ -1782,16 +1782,6 @@
       <version>1.4</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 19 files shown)


andymc12:
#build

WhiteCat22 (approved):

PR #14669, opened 2020-10-23
Use App BDAs with MP GraphQL (allows removal of overlay GraphQLConfig)

andymc12:
One of the overlayed files for SmallRye GraphQL was GraphQLConfig because it used CDI annotations that didn't play well with OSGi.  Using the application.bdas.visible=true DS property makes it play nicer, allowing us to remove the overlay.


andymc12:
#build

tevans78:
If the config is injected then you probably need to change this code too?
https://github.com/OpenLiberty/open-liberty/blob/77e75af2301f216109e3570629ee7fce8d207ba9/dev/com.ibm.ws.io.smallrye.graphql/src/com/ibm/ws/io/smallrye/graphql/component/GraphQLServletContainerInitializer.java#L118

andymc12:
@tevans78 no, I think we still want this code in place.  We want our config object to be used (it is more liberty-centric) and that allows us to new-up the servlets and dynamically add them to the web app with the proper config.

WhiteCat22:
What does `application.bdas.visible=true` do?

andymc12:
> What does application.bdas.visible=true do?

Tom suggested it:
> I think that MP Reactive Messaging had the same issue … the smallrye impl wanted the Config instance to be injected. I think the trick is to set application.bdas.visible=true on the CDI extension DS class - https://github.com/OpenLiberty/open-liberty/blob/a36ba6172e2894d405d996f50e4c89b1c160b452/dev/com.ibm.ws.io.smallrye.reactive.messaging-provider/src/com/ibm/ws/microprofile/reactive/messaging/OLReactiveMessagingExtension.java#L38 (edited)

tevans78:
@WhiteCat22 as part of CDI initialization for each application, a graph of Bean Deployment Archives is built up. One of the purposes for this is to define which Beans are visible to which other Beans. The graph of BDAs includes all application archives (which meet certain criteria) plus extension archives as provided by the runtime. The visibility rules are pretty complicated but one rule is that, in general, an internal extension archive can not see the Beans provided by an application archive or other internal extension archives.

`application.bdas.visible=true` basically disables that rule and allows runtime extension archives (such as the one provided by mpGraphQL-1.0) to see application archives and other runtime extension archives (such as the one provided by mpConfig-1.4). Originally there was just one rather niche usage for this but it is becoming more common and it perhaps could do with redesigning.

tevans78 (approved):

tevans78:
@andymc12 approved so long as you understand the possible issues with calling `new GraphQLConfig()` on a class that is supposed to be a CDI bean. https://github.com/OpenLiberty/open-liberty/blob/77e75af2301f216109e3570629ee7fce8d207ba9/dev/com.ibm.ws.io.smallrye.graphql/src/com/ibm/ws/io/smallrye/graphql/component/GraphQLServletContainerInitializer.java#L118

PR #14695, opened 2020-10-26
Use parallel assemble

aguibert:
Non-parallel build times appear to be in the range of 16m30s to 19m. See if parallel is significantly faster


aguibert:
Using parallel build did not noticably reduce execution time for assemble -- cancelling this PR

PR #14703, opened 2020-10-26
Allow config timeout and skip addl OpenJDK for ACME fats

kristip17:
Fix two build breaks with FAT fixes.

- Hit a config timeout -- I can't change the warning time in the FAT tests even though we do complete the update -- the ACME call just takes longer sometimes. `CWWKG0027W`

- Hit another Win/JDK combo where we can't remove files programmatically


kristip17:
#build

jvanhill (approved):

PR #14704, opened 2020-10-27
Ignore Tx failure during server stop for EJB FAT

tkburroughs:
In flight EJB timer attempts to rollback during server stop
but transaction service has already stopped; ignore this error.


tkburroughs:
#build

olendvcook (approved):

PR #14707, opened 2020-10-27
Fix unit tests that fail in isolation

KyleAure:
com.ibm.websphere.javaee.management.j2ee.1.1 -> assumes jar of project already exists, but won't if you just run gradle :test 

com.ibm.ws.security.fat.common -> fails during assembly on my local system and on the GH actions build. Looks to have unnecessary import and library.

com.ibm.ws.security.openidconnect.client -> assumes that one of the other unit tests runs before another which isn't always the case.


KyleAure:
#build

aguibert (approved):

PR #14708, opened 2020-10-27
Consolidate usage of utility method

brenthdaniel:
There are currently two places where we're transforming a File object into a variable name. They should be in the same place.


brenthdaniel:
#build

hlhoots (approved):
looks ok to me.

PR #14717, opened 2020-10-27
direct lookups and add more basic tests

njr-11:
Make modifications to the extension pattern so that direct lookups can work.  Add a variety of tests around direct lookups and resource references specifying the WorkManager interface rather than an executor interface.


jmstephensgit (approved):

njr-11:
#build

PR #14738, opened 2020-10-28
Fix relative path for referencing public visibility feature folders.

crpotter:
The relative path in the `build.gradle` file for `com.ibm.websphere.appserver.features` used during the task which copies the PII files incorrectly references itself from the `visibility` folder and not the root of the project. This commit simply tweaks the relative path to make it correct, which should allow these feature PII files to show up in the `build.pii.package` folder during proceessing.


crpotter:
#build

crpotter:
#build

wraschke (approved):

PR #14754, opened 2020-10-29
add file option to tlsprofiler

ryanesch:
Adding --file option for the tlsProfiler securityUtility command.
If no file is specified, the output will go to console as before.
If full path is specified, output will be written to specified location. If no file name is specified, filename will be created in the form of tlsProfiler-date.txt. If only filename is specified, file will be created in liberty's server directory.
console output will include file location, eg. Writing file to: /Users/bob/test1/test2/output.txt


una-tapa (approved):
Looks great. Thanks Ryan!

ryanesch:
#build

Issue #14743, opened 2020-10-28
Variables in include files not recognized after config update


brenthdaniel:
This scenario requires a few things:
1) A configuration update after startup
2) An include file that contains a variable in its location (ie, `<include location="${my.directory}/myFile.xml"/>` )
3) An include file that contains a variable definition that is loaded after the include file from (2)

In this case, the variable from (3) will not be loaded. As a workaround, the include file from (3) should appear before the include file in (2)

PR #14767, opened 2020-10-29
Clean up usage of temporary variables

brenthdaniel:
The internal state of config variables can get a bit confused if we encounter an 'include' element that contains a variable and have to temporarily add variables to the registry.


brenthdaniel:
#build

cbridgha (approved):
LGTM

brenthdaniel:
PR for #14743

PR #14772, opened 2020-10-30
MicroProfile specs should not beta yet

njr-11:
Per discussion with Emily, we shouldn't beta MicroProfile specs yet and should wait to align with the others.  This pull takes MP Context Propagation 1.1 out of beta.


njr-11:
#build

Issue #14752, opened 2020-10-29
mpConfig-2.0: getOptionalValue performance


jdmcclur:
In my primitives with mpOpentracing enabled, I am seeing a regression with microProfile-4.0 compared to microProfile-3.3. 10% in the worst case.

This is showing up in the profile. Opentracing looks up an optional config variable.
```
  Parent      0   0.04   3.20           3         236    J:io/openliberty/opentracing/internal/OpentracingService.process(Ljava/net/URI;Ljava/lang/String;Lio/openliberty/opentracing/internal/filters/SpanFilterType;)Z
 
    Self      0   0.04   3.20           3         236    J:io/openliberty/opentracing/internal/OpentracingConfiguration.getServerSkipPattern()Ljava/lang/String;
 
   Child      0   0.04   2.97           3         219    J:io/smallrye/config/SmallRyeConfig.getOptionalValue(Ljava/lang/String;Ljava/lang/Class;)Ljava/util/Optional;
   Child      0   0.01   0.19           1          14    J:org/eclipse/microprofile/config/ConfigProvider.getConfig(Ljava/lang/ClassLoader;)Lorg/eclipse/microprofile/config/Config;
 ```

Looking farther down the profile, it looks like we spend a lot of time iterating over the server.xml default variables (and getting env variables). This gets worse the more these methods are called.

This looks like it was recently added:
```
  Parent      0   0.18   1.53          13         113    J:io/openliberty/microprofile/config/internal/serverxml/ServerXMLVariableConfigSource.getProperties()Ljava/util/Map;
 
    Self      0   0.18   1.53          13         113    J:io/openliberty/microprofile/config/internal/serverxml/ServerXMLDefaultVariableConfigSource.getServerXMLVariables()Ljava/util/Map;
 
   Child      0   0.01   0.42           1          31    J:java/util/stream/ReferencePipeline.collect(Ljava/util/stream/Collector;)Ljava/lang/Object;
   Child      0   0.11   0.39           8          29    J:com/ibm/ws/config/xml/internal/variables/ConfigVariableRegistry.getAllLibertyVariables()Ljava/util/Collection;
   Child      0   0.00   0.33           0          24    J:java/util/stream/Collectors.toMap(Ljava/util/function/Function;Ljava/util/function/Function;)Ljava/util/stream/Collector;
   Child      0   0.14   0.22          10          16    J:java/util/Collections$UnmodifiableCollection.stream()Ljava/util/stream/Stream;
``` 

[...]

PR #14783, opened 2020-11-02
Serverxml vars config source performance

Azquelt:
Improve performance by
* converting streams to for loop (this is the major performance improvement)
* removing unnecessary copying of maps (this is just easy and csimplifies the code)

If these improvements aren't sufficient then we'll need to request changes to `ConfigVariableRegistry` to allow us to get the default values of variables which have been overridden by an actual value more efficiently.

For #14752


benjamin-confino (approved):

Azquelt:
#build

Azquelt:
#build

PR #14786, opened 2020-11-02
Enable FATs in GH Actions build

aguibert:
Explicitly mentioned categories will be ordered first. Directly modified FATs will also be run first.

Run these categories first: JCA, SECURITY_1


aguibert:
#run-libby-bot

KyleAure (approved):

PR #14803, opened 2020-11-02
Improve GenerateCategories for handling FAT-only and no-FAT PRs

aguibert:
Currently the change detector will no-op buckets that aren't related to the code being modified, but this still takes a non-negligable amount of time to no-op an entire category.

There are 3 categories of PRs:
1. FAT-only changes
2. FAT + other changes
3. no FAT changes

Previously we were only optimized for (2), and this PR will allow us to heavily optimize for (1), as well as slightly optimizing for (3)


KyleAure (approved):

aguibert:
#run-libby-bot

PR #14805, opened 2020-11-03
Alternative server-xml transformer

### LINKED ISSUE #14374: Transform server.xml for jakarta testing

KyleAure:
Here is an alternative to using the Jakarta transformer to transform string literals in server.xml file that need to be transformed.

Fixes #14374


dev/wlp-jakartaee-transform/rules/jakarta-xml-server.properties
```diff
@@ -0,0 +1,3 @@
+# This properites file is used to transform javax package names in server.xml
```
dazavala, line 1: Running the transform for every server/client configuration is still overkill IMHO.  Also, we have all kinds of xml files transformed via jakarta-xml-master.  I believe we do need a special properties file like jakarta-xml-server to cover the case where the config files are actually named server.xml (client.xml).  This solution could perform the transform iff the server configuration file name appears in the LHS of a property in jakarta-xml-master.  The solution would also tease the transformer replacement properties from the RHS of the same property, which might very well be jakarta-xml-server.properties.

KyleAure:
Abandoning this change as it doesn't seem worthwhile to make.

PR #14810, opened 2020-11-03
SSL auto feature should work with restfulWSClient-3.0 (not just server)

andymc12:
Users should only need `restfulWSClient-3.0` and SSL to be able to use Liberty-configured SSL settings.

This is not yet released function.


andymc12:
#build

jim-krueger (approved):

andymc12:
All failures are unrelated and appear in other builds.

andymc12:
#libby

PR #14819, opened 2020-11-04
Ensure manifest file for Liberty packaged server archive

tjwatson:
The temporary manifest File for the Liberty server archive is not closed
after writing it.  It is still left open when trying to read it to add
it to the final archive.  It is possible this is causing some corruption
of the manifest in the final archive.

Also adding trace to the SpringBootThinUtil to help debug the inputs
during the Spring Boot FAT that use the Liberty uber JAR.


tjwatson:
#build

dazavala (approved):
Had to review try-with-resource and Objects semantics to ensure the FOS closes and no new paths result in NPE.  Changes look good.

PR #14827, opened 2020-11-04
Add arhive validation which allows prepended scripts

### LINKED ISSUE #14826: Allow Spring Boot app with embedded launcher script to  deploy
**Describe the bug**   
Spring Boot tools generate WARs with launcher scripts prepended to the front.  Such a WAR, which used to be deployable pre-19.0.0.2, no longer deploys.   In 19.0.0.2 the test for valid archives was updated.  In this update, a check for readability by JavaInputStream was added.  This check fails because of the script at the front of the archive.

**Steps to Reproduce**
Append a script, or any data at all to the front of a WAR and try to deploy it.

**Expected behavior**
The WAR should deploy, but it does not.

**Diagnostic information:**
 - OpenLiberty Version: [e.g. 19.0.0.2 and up]
 - Java Version: [ doesn't matter ]
 - server.xml configuration ( Nothing specific.  Test with dropins )

_-- **Removed regression label**  -- since we never claimed support for this type of WAR.  It just happened to work because we didn't do validation._

jimblye:
Fixes #14826

### DIFF
--- a/dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
+++ b/dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
@@ -47,8 +47,7 @@
  *   
  *   Note the offsetOfCentralDirectory is relative to the beginning of the actual archive (not an offset  
  *   from the beginning of the file.)  Subtracting away all of the known pieces gives us the unknown 
- *   piece which is the actual archive offset.
- *  
+ *   piece which is the actual archive offset.  
  */
 
 package com.ibm.ws.artifact.zip.internal;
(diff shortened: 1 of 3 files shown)


jimblye:
#build 
#libby

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+    		_endOfCentralDirectoryRecord = getEndOfCentralDirectoryRecord(file);
+    		
+    		if ( isZip64() ) {
+    		   _zip64EoCDRLocator = getZip64EoCDLocator(file);
+    		   _zip64EoCDRLocatorOffset = _fileLength - _eocdRecordSize - ZIP64_LOCATOR_SIZE;
+    		   _zip64EoCDR = getZip64EndOfCentralDirectoryRecord(file);
+    		   _zip64EoCDRecSize = getZip64EoCDRecordSize( _zip64EoCDR, 0 );
+    		}
+    		
+    		int archiveOffset = (int)getActualArchiveOffset();
```
tbitonti, line 121: Is this cast to (int) safe?  How is it know that the offset is less than MAX_INT?

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+
+    		if (archiveStartsWithSignature(file, archiveOffset)) {
+    			return true;
+    		} 
+
+    	} catch (Exception e) {
+    		//e.printStackTrace();
```
tbitonti, line 128: Do we want to display some information here in case of an exception?

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+     * Reads from the end of the file looking backwards for the EoCDR signature.
+     * 
+     * The implementation reads the file in BLOCK_SIZE blocks from the end of the file
+     * and searches each block.  Note that there is some overlap of the successive 
+     * blocks; EOCDR_MIN_SIZE bytes of overlap to be precise.  This is because the 
+     * record might start at the end of one block and finish in the next.
+     * 
+     * @param file
+     * @throws IOException
+     */
```
tbitonti, line 146: Either add documentation on the @param and @throws lines, or omit them.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+//		// We don't need the comment, which could be up to 64k.		
+//		_eocdRecordSize = size;
+//		return Arrays.copyOfRange(block, offset, offset + EOCDR_MIN_SIZE ); 
+//	}
+	
+	/**
+	 * Get the 20-byte Zip64 Locator.  It must sit 20 bytes before the End-of-Central-Directory-Record.
+	 * @param file the zip file
+	 * @return
+	 * @throws IOException
```
tbitonti, line 228: As above; @param, @return, and @throws can be omitted if they provide no actual documentation.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+	 */
+	private byte[] getZip64EndOfCentralDirectoryRecord(RandomAccessFile file ) throws IOException {
+		
+		if ( !isZip64() ) {
+			return null;
+		}
+
+		// The offset of the Zip64-End-of-Central-Directory-Record is found in the Zip64 locator at offset 8.
+		// That offset will not be correct if data (i.e. a script) has been attached in front of the archive.
+		long zip64EoCDOffset = this.getLittleEndianValue( _zip64EoCDRLocator, ZIP64_LOCATOR_EOCD_OFFSET_OFFSET, 4);
```
tbitonti, line 265: Stale 'this' reference.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+		
+		// Offset 16 contains the relative offset from the beginning of the actual archive to the Central Directory
+		long offsetOfCentralDirectory = getLittleEndianValue( _endOfCentralDirectoryRecord, EOCDR_CENTRAL_DIR_OFFSET, 4 );
+		
+		long zip64RecLengths = isZip64() ? ( _zip64EoCDRecSize + ZIP64_LOCATOR_SIZE ) : 0;
+		
+		return _fileLength - _eocdRecordSize - zip64RecLengths - lengthOfCentralDirectory - offsetOfCentralDirectory;
+	}
+	
+	public static void main(String[] args) {
```
tbitonti, line 530: This should probably be made into a unit test, with the test JARs constructed by the test, or stored remotely.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/ZipFileContainerFactory.java
```diff
@@ -352,7 +352,10 @@ private static boolean isZip(ArtifactEntry artifactEntry) {
             try {
                 ZipEntry entry = zipInputStream.getNextEntry();
                 if ( entry == null ) {
-                    Tr.error(tc, "bad.zip.data", getPhysicalPath(artifactEntry));
+                    JavaArchive javaArchive = new JavaArchive(getPhysicalPath(artifactEntry));
```
tbitonti, line 355: Does the JavaArchive need to be closed?

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/ZipFileContainerFactory.java
```diff
@@ -448,8 +451,11 @@ private static boolean isZip(File file) {
             try {
                 ZipEntry entry = zipInputStream.getNextEntry(); // throws IOException
                 if ( entry == null ) {
-                    Tr.error(tc, "bad.zip.data", file.getAbsolutePath());
-                    return false;
+                    JavaArchive javaArchive = new JavaArchive(file.getAbsolutePath());
```
tbitonti, line 454: Same as above: Does the JavaArchive need to be closed?

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+ *  
+ */
+
+package com.ibm.ws.artifact.zip.internal;
+
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.util.Arrays;
+
+public class JavaArchive {
```
tbitonti, line 60: Not sure about using the name "JavaArchive': The target archive is a PK archive, and can be a plain ZIP file, which has nothing to do with java.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+			// a large file would cause an unacceptable delay if we did not set a limit.
+			if (offsetFromStartingPoint > MAX_PREFIX_ALLOWED) {
+			     throw new IOException("Failed to find the Zip64-End-of-Central-Directory-Record after searching " + MAX_PREFIX_ALLOWED + " bytes.");	
+			}
+			
+			if ( offsetInBlock > block.length - 4 ) {  // Must allow 4 bytes at end of block to read signature
+
+				block = getBlockFromBeginningOfFile(file, offsetFromBeginningOfFile, FORWARD_BLOCK_SIZE);
+				if (block.length < 4) {
+					throw new IOException("Reached end of file while searching for to Zip64-End-of-Central-Directory-Record does not correct.");
```
tbitonti, line 295: Hmm; run on exception text?

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+	/**
+	 * 
+	 * @param file  a zip file
+	 * @param offsetFromEnd Offset from end of file where reading begins
+	 * @param size number of bytes of data to read
+	 * @return Returns an array of 'size' bytes that starts from 'offsetFromEnd' of file.
+	 * @throws IOException
+	 */
+	private byte[] getBlockFromEndOfFile(RandomAccessFile file, int offsetFromEnd, int size) throws IOException {
+
```
tbitonti, line 337: For better performance, the block can be allocated outside of the method, and reused when there are several calls.  But, since this code is a very rare case, that level of optimization doesn't seem necessary.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+	}
+	
+	/**
+	 * 
+	 * @return true if the signature parameter matches the bytes at offset
+	 */
+	private boolean isSignature(byte[] bytes, int offset, int signature) {
+
+		long value = getSignature(bytes, offset);
+
```
tbitonti, line 384: We are reading a long but comparing that against an int signature value?

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
```diff
@@ -0,0 +1,575 @@
...
+		return (value == signature);
+	}
+	
+	private boolean isValidEoCDRecSize(byte[] bytes, int offset, int size) {
+		// Size of central directory must be the structure size + comment
+		// First put length in little endian form (reverse the bytes).
+		long commentLength = getLittleEndianValue(bytes, offset, 2);
+		return size == (EOCDR_MIN_SIZE + commentLength);
```
tbitonti, line 392: 'size' is an int, but comment length is a long ... but the comment is restricted to 64K at most?  Should 'getLittleEndianValue' answer an int instead of a long?

tbitonti (approved):
Some concern about the int/long conversions, but otherwise looks good.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: In this particular call to getLittleEndianValue, we are asking it to read 2 bytes.  So the most it can return here is 0xFFFF.  But sometimes we ask getLittleEndianValue to read 4 bytes and in one place we ask it to read 8 bytes.   A cast to int might be appropriate here, but shouldn't be any harm in storing it in a long.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: done

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: Changed getSignature to return an int.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: yes

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: changed to ZipValidator

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: done

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: done

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: done

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: Added // ffdc.   I think the ffdc is enough

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: Removed cast.   Decided archiveOffset can be a long, at least when the zip is not in Zip64 format.   In Zip64 format, if the prepended script is bigger than 200MB, an exception will throw us out.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/ZipFileContainerFactory.java
jimblye: Well, not here.   The JavaArchive.isValid() method opens the archive file in a try with resources.   So it should be closed automagically.

jimblye: Well, not here.   The JavaArchive.isValid() method opens the archive file in a try with resources.   So it should be closed automagically.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/JavaArchive.java
jimblye: Removed main for now, but yes there should be unit tests

tbitonti (approved):

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/ZipValidator.java
```diff
@@ -0,0 +1,483 @@
...
+            return null;
+        }
+
+        // The offset of the Zip64-End-of-Central-Directory-Record is found in the Zip64 locator at offset 8.
+        // That offset will NOT be correct if data (i.e. a script) has been attached in front of the archive,
+        // but in that case, the offset will be our starting point for the search.
+        long zip64EoCDOffset = getLittleEndianValue( _zip64EoCDRLocator, ZIP64_LOCATOR_EOCD_OFFSET_OFFSET, 4);
+
+        byte [] block = getBlockFromBeginningOfFile(file, null, zip64EoCDOffset, ZIP64_EOCDR_MINIMUM_SIZE);
+        if (block.length < ZIP64_EOCDR_MINIMUM_SIZE) {
```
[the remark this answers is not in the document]
jimblye: file.read() actually will return less bytes than requested if there aren't enough bytes available before the end of file.
file.seek() lets you seek past the eof, but then when you call file.read() it returns -1.
Anyway, I think it is possible that you will get back less bytes than you asked for.

dev/com.ibm.ws.artifact.zip/src/com/ibm/ws/artifact/zip/internal/ZipValidator.java
```diff
@@ -0,0 +1,483 @@
...
+
+        byte[] block = getBlockFromEndOfFile(file, null, BLOCK_SIZE, BLOCK_SIZE);
+        int size = EOCDR_MIN_SIZE ;
+        int offset = block.length - size;
+        int offsetFromEnd = block.length - offset;   // offset from the end of the block (not the file)
+
+        if ( block.length < EOCDR_MIN_SIZE ) {
+            throw new IOException("Not a valid zip file.  Less than minimum required length.");
+        }
+
```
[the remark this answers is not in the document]
jimblye: It actually goes into that the "  if ( block.length < EOCDR_MIN_SIZE ) { "  and throws the exception.
I tried with an "archive" file of size 1.

jimblye:
#build
#libby

tbitonti (approved):

jimblye:
#build
#libby

jimblye:
#build

jimblye:
#libby

PR #14860, opened 2020-11-06
Fix modified FAT bucket detection for paths containing a second '_fat' token

aguibert:
Fixes an issue in the GH Actions build where we would incorrectly detect the modified FATs for the MODIFIED_LITE_MODE and MODIFIED_FULL_MODE categories:

```
Modified files are:
dev/io.openliberty.microprofile.config.2.0.internal_fat/fat/src/io/openliberty/microprofile/config/internal_fat/apps/defaultSources/DefaultSourcesTestServlet.java
Will be running buckets io.openliberty.microprofile.config.2.0.internal_fat/fat/src/io/openliberty/microprofile/config/internal_fat
```

Run this category first: MP_CONFIG

### DIFF
--- a/.github/workflow-scripts/check-fat-failures.sh
+++ b/.github/workflow-scripts/check-fat-failures.sh
@@ -8,7 +8,7 @@ if [[ $CATEGORY =~ MODIFIED_.*_MODE ]]; then
   git diff --name-only HEAD^...HEAD^2 >> modified_files.diff
   echo "Modified files are:"
   cat modified_files.diff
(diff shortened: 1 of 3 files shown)


Joseph-Cass (approved):
LGTM (as long as the test identifies the bucket has been "modified") :)

aguibert:
Looks like the issue is fixed now since the `MODIFIED_LITE_MODE` and `MODIFIED_FULL_MODE` categories each passed. Merging

PR #14866, opened 2020-11-06
HTTP/2 channel needs to check for gRPC request earlier

wtlucy:
In some cases, we need to check to see if a request is for a gRPC service earlier than we currently do.  We need to perform that check earlier in the processing of new HTTP/2 requests.


wtlucy:
#build

billw1024 (approved):
looks good

wtlucy:
#build

wtlucy:
#run-libby-bot

PR #14909, opened 2020-11-10
Use uniq token with Logout test

c00crane:
Update the the testLogoutReuseCache test to wait 1 second before completing.  The current time is part of what makes a token unique.  Our tests use the same user, client, ... so when we the a token is created for the logout test, we may end up creating the same token in the next test if the tests run too fast.  When cache cleanup runs, it ends up cleaning up the token created for the next test causing that test to fail.


c00crane:
#build

c00crane:
#build
Previous build ran out of space

ayoho (approved):

PR #14920, opened 2020-11-11
initialize InstallKernelMap before capturing SharedOutput

bensonlatibm:
This should fix the issue with Install Unit Tests printing 1 character per line on windows and some other systems.


bensonlatibm:
#build

hermankailey (approved):

PR #14929, opened 2020-11-11
Remove comment with link to Narayana code

andymc12:
Removing a confusing comment with a link to Narayana code.

Since this is only removing a comment, I don't want to waste build machine cycles on a personal build, so I plan to merge as soon as this has been reviewed/approved.

### DIFF
--- a/dev/com.ibm.ws.io.smallrye.graphql/src/io/smallrye/graphql/servlet/IndexInitializer.java
+++ b/dev/com.ibm.ws.io.smallrye.graphql/src/io/smallrye/graphql/servlet/IndexInitializer.java
@@ -30,8 +30,6 @@ import com.ibm.ws.ffdc.annotation.FFDCIgnore;
 
 /**
  * This creates an index from the classpath.
- * Based on similar class in LRA
- * (https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/ClassPathIndexer.java)
  * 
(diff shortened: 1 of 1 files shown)


andymc12:
This came about due to some confusion over whether the IndexInitializer class was exposing LGPL-licensed code vs re-licensing, etc.  Here is a conversation I had with Phillip Kruger, the author of this code:

> Andy McCright @andymc12 09:59
do you have a few minutes after the meeting to chat? a question came up about licensing in SmallRye GraphQL
Andy McCright @andymc12 11:04
sorry, was hoping to chat in person but the meeting went too long... we're concerned about this comment in IndexInitializer:
>
```
/**
 * This creates an index from the classpath.
 * Based on similar class in LRA
 * (https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/ClassPathIndexer.java)
```
> Since Narayana is licensed under LGPL, but SmallRye GraphQL is licensed under Apache, the concern is whether that code was copied from Narayana. And if so, was it then re-licensed under Apache?
From what I could tell, it didn't look like a copy in any case - more just like using the same patterns as LRA/Narayana, but the legal people are asking me to confirm that with you.

>Phillip Krüger @phillip-kruger 13:03
Sorry, only seeing this now. It's mostly based in the same pattern, not a copy, we should actually just remove the comment in the javadoc...

andymc12:
#build

NottyCode (approved):

PR #14932, opened 2020-11-11
Add Beta & Blog parts to feature template

mbroz2:
Update feature template with beta and blog post specific parts.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.md
+++ b/.github/ISSUE_TEMPLATE/feature.md
@@ -9,9 +9,11 @@ assignees: ''
 Describe the high level feature, including any external spec links.
 
 
-When ready, add links to the Upcoming Feature Overview document and Feature Test Summary issue:
+When ready, add links to the Upcoming Feature Overview document as well as Feature Test Summary and blog post issues:
 - UFO:
 - FTS:
+- Beta Blog Post (if applicable):
+- Blog Post: 
 
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## List of Steps to complete or get approvals / sign-offs for Onboarding to the Liberty release (GM date)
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -69,6 +79,9 @@ All features (both "Design Approved" and "No Design Approved")
 - [ ] Ship Readiness Review and Release Notes completed (Epic owner / Feature owner / Release PM)
 - [ ] Github Epic and Epic's issues are closed / complete. All PRs are committed to the master branch. (Epic owner / Feature owner / Backlog Subtribe PM)
 
+1 week before GA
+- [ ] Create, populate, and link to the [Blog post issue](https://github.com/OpenLiberty/open-liberty/issues/new?assignees=lauracowen%2C+jakub-pomykala&labels=&template=blog_post_ga_release.md&title=GA+BLOG+-+title_of_your_update)
```
tevans78, line 83: Do you really want Laura and Jakub hardcoded into this query?

tevans78:
> Read and understand this completely, then delete this entire template. If a reviewer or merger sees this template, they should fail the review or merge.
@mbroz2 So should I fail the review for this template update?!

.github/ISSUE_TEMPLATE/feature.md
lauracowen: If they need to be assigned to anyone, someone needs to be hardcoded in I think or they have to guess at who to assign it to. But probably Jakub and Austin (https://github.com/austin0) would be better now? Or just leave it blank and they'll work from Sam's query - feels a bit scary without assignees though - could just get lost in the ether....

lauracowen (approved):
LGTM

.github/ISSUE_TEMPLATE/feature.md
tevans78: from a developer's perspective, isn't the label enough? it is not up to them to ensure that the issue is assigned? and what happens if the team members change?

lauracowen: I completely agree. I just see them tend to assign the issues to someone they think they should go to. But maybe that doesn't matter. If they do assign it, fine. If they don't, also fine.

mbroz2: I don't mind one way or the other, but those links are the ones that are used by the blog post templates: https://github.com/OpenLiberty/open-liberty/issues/new/choose
which is why I used them.  If we don't want to have default assignees on the blog posts issues, then the templates should be updated as well.

mbroz2: Given the above, I'm going to merge this as-is, and if we want to remove default assignees from the blog post, we can do it in all 3 templates in a subsequent PR, but I'll leave that call for @lauracowen to make

Issue #14936, opened 2020-11-12
Issue when deploying Open Liberty application to Openshift


levipapp:
I'm trying to expose a SOAP service in open liberty that is connected to a db. For that I added a data source configuration in the server.xml file. It works locally(docker image run locally connected to remote DB) but when I try to run the application in Openshift I get this error:
[ERROR   ] SRVE0271E: Uncaught init() exception created by servlet [com.ibm.lom.inputservice.facade.IsPrintServiceWSmpl] in application [input-service]: java.lang.NoClassDefFoundError: javax.xml.soap.SOAPException
	at com.ibm.ws.jaxws.globalhandler.JaxwsGlobalHandlerServiceImpl$JaxWSGlobalHandlerBusListener.initComplete(JaxwsGlobalHandlerServiceImpl.java:111)
	at [internal classes]
Caused by: java.lang.ClassNotFoundException: javax.xml.soap.SOAPException cannot be found by com.ibm.ws.jaxws.common_1.0.45.cl201020200915-1100
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:464)
	... 1 more
[ERROR   ] SRVE0315E: An exception occurred: java.lang.Throwable: javax.servlet.ServletException: SRVE0207E: Uncaught initialization exception created by servlet
	at com.ibm.ws.webcontainer.webapp.WebApp.handleRequest(WebApp.java:5115)
	at [internal classes]
Caused by: javax.servlet.ServletException: SRVE0207E: Uncaught initialization exception created by servlet
	at com.ibm.ws.webcontainer.servlet.ServletWrapper.init(ServletWrapper.java:360)
	... 1 more
Caused by: java.lang.NoClassDefFoundError: javax.xml.soap.SOAPException
	at com.ibm.ws.jaxws.globalhandler.JaxwsGlobalHandlerServiceImpl$JaxWSGlobalHandlerBusListener.initComplete(JaxwsGlobalHandlerServiceImpl.java:111)
	... 1 more
Caused by: java.lang.ClassNotFoundException: javax.xml.soap.SOAPException cannot be found by com.ibm.ws.jaxws.common_1.0.45.cl201020200915-1100
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:464)
	... 1 more

- If I remove the data source configuration from the server.xml file the application works in Openshift. We are using the same data source configuration on several other micro services and they work in Openshift.

[...]

PR #14940, opened 2020-11-12
Disable osgi.checkConfiguration by default

tjwatson:
When enabled this setting cause Equinox
to check the timestamp of every bundle JAR
installed and it uninstall anything that
has changed since last start.

The framework then causes a refresh of the
uninstalled bundles.  In some cases this
will cause other bundles to re-resolve
which ends up resolving before the
region digraph has been registered with
the framework.  This means the bundles
will be resolved without the issolation
which the Feature Manager establishes.

It appears in some cloud environments and
container images false timestamp reporting
is happening which incorrectly causes some
kernel bundles to be tossed.  To avoid
this the osgi.checkConfiguration option
is disabled.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkConfigurator.java
+++ b/dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkConfigurator.java
@@ -95,11 +95,10 @@ public class FrameworkConfigurator {
         // equinox to do that.
         config.put("osgi.framework.activeThreadType", "none");
 
-        // Use an equinox property to ensure that equinox checks whether or not
-        // the jar files we're installing (for file:// URLs) have changed if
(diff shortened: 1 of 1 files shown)


tjwatson:
Fixes issue #14936

tjwatson:
#build

NottyCode (approved):

tjwatson:
This change needs to be considered for scenarios that may want to overlay a Liberty bundle JAR with a patched or updated JAR for some reason.  Certain updates to bundle JARs may not be detected until you run with --clean.

Changes that will be detected:

1. Changes to bundle classes.
1. Changes to resources that are not cached

Changes that will not be detected:
1. Changes to the bundle manifest.  For example, changes to exported packages, changes to imports, changes to versions etc.
1. Changes to Declarative Service component XML files
1. Metatype XML data
1. Other resources that may be cached for faster startup

Any time a JAR is overlaid in the Liberty installation the servers should be started with --clean to ensure any changes are picked up by the runtime.

PR #14944, opened 2020-11-12
Enable junit report

### LINKED ISSUE #14943: Replace testspace with junit report
Replace current reporting of junit test results to use the `action-junit-report` action instead. 
Documentation: https://github.com/marketplace/actions/junit-report

KyleAure:
Fixes #14943 and #15144 and #15051


KyleAure:
#build

.github/workflow-scripts/run-unit-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+mkdir $UNIT_DIR
+
+#Collect test results in central location
+#Check each project
+for project in */ ; do
+    # If unit test folder exists
+    if [[ -d "$project/build/libs/test-reports/test" ]]; then
+        pushd $project/build/libs/test-reports/test &> /dev/null
+        # Then copy unit test file(s) to central location
+        for f in TEST-com.*.xml TEST-io.*.xml ; do
```
aguibert, line 42: Are we trying to exclude something by doing `TEST-com.*.xml TEST-io.*.xml` as opposed to just `TEST-*.xml`?

.github/workflows/openliberty-ci.yml
```diff
@@ -20,14 +20,14 @@ on:
 env:
   LANG: en_US.UTF-8
 jobs:
-  build:
-    name: Build Open Liberty
-    runs-on: ubuntu-latest
-    outputs:
+  setup: #Validates copyright headers, selects test-os, selects java version, and sets up test categories.
```
aguibert, line 23: I'd rather leave this stuff in the core `Build` job because there is such a small amount of work occurring in `Setup`. Using public runners this completes in about 70 seconds it looks like, but most of that time is spent provisioning the new machine and cloning the repo, When we switch to self-hosted runners this will likely become more of an issue where provisioning new jobs are slightly slower.

.github/workflows/openliberty-ci.yml
```diff
@@ -82,27 +81,27 @@ jobs:
...
-    - uses: actions/checkout@v2
-#    - uses: testspace-com/setup-testspace@v1
-#      with:
-#        domain: ${{ github.repository_owner }}  
+    - uses: actions/checkout@v2 
     - name: Set up Java
       uses: joschi/setup-jdk@v2
       with:
-        java-version: 11
+        java-version: ${{ needs.setup.outputs.test-java }}
```
aguibert, line 104: for now lets leave unit tests hardcoded to java 11. In our existing RTC setup we only ever run unit tests on java 11, so for now it should be sufficient to leave that behavior in place.

.github/workflows/openliberty-ci.yml
```diff
@@ -126,41 +125,40 @@ jobs:
...
-      run: .github/workflow-scripts/upload-unit-tests.sh
-  fat_tests:
+    - name: Analyze junit output
+      if: steps.run-unit-tests.outputs.status == 'failure'
+      uses: KyleAure/junit-report-annotations-action@1.6
+      with:
+        name: Unit test report
+        path: "dev/unit-results/TEST-*.xml"
+        includeSummary: true
+        numFailures: 10 # Truncate number of failures so we don't run out of annoations.
```
aguibert, line 143: what's the reasoning behind limiting num failures? Do we have a limited amount of annotations on each build? Or is it just so we avoid flooding logs/results with failures?

dev/wlp-gradle/java.gradle
```diff
@@ -77,8 +77,9 @@ subprojects {
   test { 
     // Once testing is complete copy 
     finalizedBy junitReport
-  
-    ignoreFailures = true
+    
+    // This is unnecessary as finalizedBy will run regardless of subproject failure
+	//ignoreFailures = true
```
aguibert, line 82: lets just delete this instead of commenting it out

dev/wlp-gradle/subprojects/tasks.gradle
```diff
@@ -360,7 +360,8 @@ test {
   }
 
   timeout = Duration.ofMinutes(15)
-  ignoreFailures = Boolean.valueOf(rootProject.ext.userProps.getProperty("gradle.test.ignoreFailures", "false"))
+  //Set ignore failures using -Dgradle.test.ignoreFailures=true
+  ignoreFailures = Boolean.valueOf(System.getProperty("gradle.test.ignoreFailures", "false"))
```
aguibert, line 364: this looks fine, but lets be sure to test that it still works on an RTC build when there is a unit test failure

aguibert (commented):
Looks good overall, my main comment is to get rid of the extra `Setup` job and just leave it as part of the `Build` job like we had before

KyleAure:
#build

aguibert (approved):

KyleAure:
This build passed and is ready to merge when the streams are open again.

KyleAure:
#libbybot

PR #14956, opened 2020-11-12
14807 update weld inject

benjamin-confino:
jakarta.inject-api from 2.0.0-RC4 to 2.0.0
jakarta.enterprise.cdi-api from 3.0.0-M4 to 3.0.0
weld-se-core from 4.0.0.Alpha3 to 4.0.0.Beta5
weld-osgi-bundle 4.0.0.Alpha3 to 4.0.0.Beta5

### DIFF
--- a/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
+++ b/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
@@ -714,12 +714,12 @@
     <dependency>
       <groupId>jakarta.enterprise</groupId>
       <artifactId>jakarta.enterprise.cdi-api</artifactId>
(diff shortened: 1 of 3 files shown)


benjamin-confino:
#build

volosied (approved):
looks good, thanks!

wraschke:
After discussing with Benjamin, we think this will resurface https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemName/com.ibm.team.workitem.WorkItem/279573, which went away in a build that contained revert of 14813. The PB mentioned here doesn't run any FAT tests, so we think this change will cause continuous builds to fail again with `java.lang.IllegalStateException: Unable to locate CDIProvider`.

PR #14967, opened 2020-11-13
Cache the generated OpenAPI model

Azquelt:
Add the ability to cache the OpenAPI model when a web module is deployed as a single file.

Invalidate the cache if the app file or any relevant config changes.

Note that we don't do anything to delete the cached model file if the app is undeployed (though it will be deleted if the server is started with `--clean`). This is consistent with the way liberty handles the annotation scanning cache.

Relates to #14752

### DIFF
--- a/dev/fattest.simplicity/src/com/ibm/websphere/simplicity/config/ApplicationManagerElement.java
+++ b/dev/fattest.simplicity/src/com/ibm/websphere/simplicity/config/ApplicationManagerElement.java
@@ -1,0 +1,88 @@
+/*******************************************************************************
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.websphere.simplicity.config;
+
+import javax.xml.bind.annotation.XmlAttribute;
+
+/**
+ *
+ */
+public class ApplicationManagerElement extends ConfigElement {
+
+    private Boolean autoExpand;
+    private String startTimeout;
+    private String stopTimeout;
+    private Boolean useJandex;
+
+    public Boolean getAutoExpand() {
+        return autoExpand;
+    }
+
+    @XmlAttribute
+    public void setAutoExpand(Boolean autoExpand) {
+        this.autoExpand = autoExpand;
+    }
+
+    public String getStartTimeout() {
+        return startTimeout;
+    }
+
+    @XmlAttribute
+    public void setStartTimeout(String startTimeout) {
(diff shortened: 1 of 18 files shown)


Azquelt:
#build

msmiths (changes requested):
Looks pretty good...

You make multiple calls to getCacheDir() in the processWebModule method... you should simply reference the cacheDir variable that you reference in the if statement:

https://github.com/OpenLiberty/open-liberty/blob/6d880b2c3d45bbf49f648dd8be0e546f34da763c/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationProcessor.java#L202

https://github.com/OpenLiberty/open-liberty/blob/6d880b2c3d45bbf49f648dd8be0e546f34da763c/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationProcessor.java#L206

It appears that the list of dependent files (fileEntries) that is maintained in CacheEntry just contains the name & path to the WAR module that is processed... is this correct? The method  name CacheEntry::addDependentFile suggests that we should also be adding the names of the dependence JARs to this list.

The definition of the ConfigField enum has a trailing comma on the last entry:

https://github.com/OpenLiberty/open-liberty/blob/6d880b2c3d45bbf49f648dd8be0e546f34da763c/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/ConfigSerializer.java#L58

Azquelt:
> Looks pretty good...
> 
> You make multiple calls to getCacheDir() in the processWebModule method... you should simply reference the cacheDir variable that you reference in the if statement:

Good point, I'll fix that.

> It appears that the list of dependent files (fileEntries) that is maintained in CacheEntry just contains the name & path to the WAR module that is processed... is this correct? The method name CacheEntry::addDependentFile suggests that we should also be adding the names of the dependence JARs to this list.

Kind of. We should add every on-disk file that we scan (i.e. no need to add jars which are within .wars). At the moment, we're only caching at all if the application is a single .war file and our scanning code only scans within this.

If we change the scanning code, we will have to add any other on-disk files we scan using `addDependentFile`. If scanning becomes more complex, we might also need another method to indicate that we've scanned something that means we shouldn't cache the result at all.
 
> The definition of the ConfigField enum has a trailing comma on the last entry:

True, I tend to do that wherever it's allowed since it makes it easier to add new entries later. I can take it out if it offends.

msmiths (approved):

Azquelt:
#build

Azquelt:
* Squashed review comment changes since they've been re-reviewed and accepted.
* Added new tests for different deployment types.
I think this is now ready to deliver.

Azquelt:
#build

Azquelt:
#run-libby-bot

Azquelt:
@msmiths would you mind checking over the tests added in the last commit?

msmiths (approved):
LGTM

Azquelt:
The failures in the build do not appear to be related to this change.

PR #14970, opened 2020-11-13
test that handlelist closes leaked autocommit false connection

njr-11:
Connection with autocommit false and a database local transaction is kept open across LTC boundaries.  There is already code in place that detects crossing the LTC boundary and rolls back the transaction.  However, with the HandleList in place, the connection handle is also automatically closed, preventing further use.  Update a test case that was written for this scenario to use a data source with HandleList enabled and cover the expected behavior.


njr-11:
#build

jmstephensgit (approved):

PR #14997, opened 2020-11-16
Do not create a Java SCC with tools

jdmcclur:
If a SCC option exists on IBM_JAVA_OPTIONS or OPENJ9_JAVA_OPTIONS, the tools will create a SCC or add to an existing SCC, which we don't want. Adding -Xshareclasses:none.


jdmcclur:
#build

jhanders34 (approved):

Issue #14885, opened 2020-11-09
NullPointer in EJB bind/unbind if Naming Context Service Does not start


olendvcook:
```
Stack Dump = java.lang.IllegalStateException: Located service is null,AtomicServiceReference[name=orbRef,ctx=org.apache.felix.scr.impl.manager.ComponentContextImpl@ea95011d,ref={com.ibm.ws.transport.iiop.spi.ORBRef, com.ibm.ws.transport.iiop.spi.ClientORBRef, com.ibm.ws.transport.iiop.spi.ServerPolicySource}={orbSSLInitTimeout=10, service.scope=bundle, AdapterActivatorOp.target=(service.pid=com.ibm.ws.transport.iiop.internal.NamingServiceAdapterActivator_19), service.ranking=5, component.name=com.ibm.ws.transport.iiop.internal.ORBWrapperInternal, config.source=file, id=defaultOrb, service.pid=com.ibm.ws.transport.iiop.internal.ORBWrapper_4, AdapterActivatorOp.cardinality.minimum=1, nameService=corbaname::localhost:2809, service.id=360, IiopEndpoint.target=(service.pid=com.ibm.ws.transport.iiop.internal.IIOPEndpointImpl_11), service.bundleid=96, config.overrides=true, iiopEndpoint=[com.ibm.ws.transport.iiop.internal.IIOPEndpointImpl_11], ComponentFactory.target=(component.factory=com.ibm.ws.transport.iiop.internal.ORBWrapperInternal), component.id=294, SubsystemFactory.cardinality.minimum=4, config.id=com.ibm.ws.transport.iiop.internal.ORBWrapper[defaultOrb], service.factoryPid=com.ibm.ws.transport.iiop.internal.ORBWrapper, adapterActivator=[com.ibm.ws.transport.iiop.internal.NamingServiceAdapterActivator_19], service.vendor=IBM, subsystemFactory=[com.ibm.ws.transport.iiop.spi.SubsystemFactory_8,com.ibm.ws.transport.iiop.spi.SubsystemFactory_7,com.ibm.ws.transport.iiop.spi.SubsystemFactory_6,com.ibm.ws.transport.iiop.spi.SubsystemFactory_5], config.displayId=orb[defaultOrb]},svc=null]
	at com.ibm.wsspi.kernel.service.utils.AtomicServiceReference.getService(AtomicServiceReference.java:261)
[...]

PR #15004, opened 2020-11-17
Handle NPE during remote unbind if orb didn't come up

olendvcook:
#14885

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.remote/src/com/ibm/ws/ejbcontainer/remote/internal/EJBRemoteRuntimeImpl.java
+++ b/dev/com.ibm.ws.ejbcontainer.remote/src/com/ibm/ws/ejbcontainer/remote/internal/EJBRemoteRuntimeImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2017 IBM Corporation and others.
+ * Copyright (c) 2014, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -145,7 +145,8 @@ public class EJBRemoteRuntimeImpl implements EJBRemoteRuntime, RemoteObjectRepla
         @Override
         public String toString() {
             return super.toString() + '[' + beanMetaData.j2eeName +
-                   ", contexts=" + Arrays.asList(contextNames) +
+                   ", contextNames=" + Arrays.asList(contextNames) +
+                   ", context= " + Arrays.asList(contexts) +
                    ']';
         }
     }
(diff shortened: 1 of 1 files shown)


olendvcook:
#build

dev/com.ibm.ws.ejbcontainer.remote/src/com/ibm/ws/ejbcontainer/remote/internal/EJBRemoteRuntimeImpl.java
```diff
@@ -249,17 +250,23 @@ public synchronized void unbindAll(Object bindingDataObject) {
             }
         }
 
-        for (int i = bindingData.contextNames.length - 1; i >= 0 && !hasBindings(bindingData.contexts[i]); i--) {
+        // For contextNames, if completely empty of bindings, unbind the naming context from its parent context.
+        for (int i = bindingData.contextNames.length - 1; i >= 0 && isBindingContextEmpty(bindingData.contexts[i]); i--) {
```
tkburroughs, line 254: Seems we are starting at a leaf node, and if that context is empty, we unbind it from the parent, then back up one and check if that context is null, etc... all the way back to the root.  As soon as we find one that is not empty, we stop.

However, now we will stop as soon as we find a "null" context.... but what if the parent context exists and is empty, shouldn't we back up one and check the parent as well?  If it does exist and is empty, seems we should unbind it.

dev/com.ibm.ws.ejbcontainer.remote/src/com/ibm/ws/ejbcontainer/remote/internal/EJBRemoteRuntimeImpl.java
```diff
@@ -249,17 +250,25 @@ public synchronized void unbindAll(Object bindingDataObject) {
             }
         }
 
-        for (int i = bindingData.contextNames.length - 1; i >= 0 && !hasBindings(bindingData.contexts[i]); i--) {
-            NamingContext parentContext = i == 0 ? rootNamingContext : bindingData.contexts[i - 1];
-            unbind(parentContext, bindingData.contextNames[i]);
+        // For contextNames, if completely empty of bindings, unbind the naming context from its parent context.
+        for (int i = bindingData.contextNames.length - 1; i >= 0; i--) {
+            if (isBindingContextEmpty(bindingData.contexts[i])) {
```
tkburroughs, line 255: This is better, but inefficient.  The code used to stop at the first non-empty list, but will now keep going.  The logic here is skipped, but `isBindingContextEmpty` will continue to be called and creates several objects and creates a list.  Could we have `isBindingContextEmpty` return `Boolean`, so we could tell the difference between True, False, and null?

tkburroughs (approved):

olendvcook:
#build

olendvcook:
#run-libby-bot

PR #15016, opened 2020-11-17
Re-enable osgi.checkConfiguration setting

tjwatson:
In order to recover when the equinox.region bundle is
discarded by the framework the kernel boot provisioner
has changed to do the following:

1) initialize the framework
2) make sure all kernel bundles are installed
3) if the equinox region bundle needs installed then
force it to resolve without allow any other bundles
to resolve.
4) finally start the framework as usual

This will enable region digraph hooks before allowing
other bundles to resolve to ensure they are following
the rules setout from a previous run.  The equinox
region persists in the system bundle data area.  Even
when it is uninstalled the persistent digraph remains
and is discovered when equinox region is reinstalled.


tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

brenthdaniel (approved):

tjwatson:
Fixes issue #14936

tjwatson:
#build

tjwatson:
#libby

PR #15024, opened 2020-11-18
Ensure app path is decoded when matching resources.

andymc12:
This resolves an issue with restfulWS-3.0 regarding the encoding/decoding of the application path. Both `@ApplicationPath("app%21")` and `@ApplicationPath("app!")` are legit according to the spec/TCK and either one should be able to be accessed from requests `GET http://localhost:9080/context-root/app%21/...` and `GET http://localhost:9080/context-root/app!/...`.

When storing resource paths in RESTEasy, the decoded path is used (i.e. `/app!`). But when the HTTP request arrives, the app path portion of the URL may not always be decoded, preventing RESTEasy from matching the request to the proper resource (resulting in a 404).  This change resolves this by ensuring that the app portion of the URL is decoded prior to resource matching.

### DIFF
--- a/dev/com.ibm.ws.jaxrs.2.0_fat/fat/src/com/ibm/ws/jaxrs20/fat/helloworld/HelloWorldTest.java
+++ b/dev/com.ibm.ws.jaxrs.2.0_fat/fat/src/com/ibm/ws/jaxrs20/fat/helloworld/HelloWorldTest.java
@@ -28,12 +28,10 @@ import org.junit.runner.RunWith;
 import com.ibm.websphere.simplicity.ShrinkHelper;
 
 import componenttest.annotation.Server;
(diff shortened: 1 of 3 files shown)


andymc12:
#build

jim-krueger:
When this issue is resolved com.ibm.ws.jaxrs20.fat.helloworld.HelloWorldTest should execute successfully on EE9.  So part of this PR should be to remove the @SkipForRepeat("EE9_FEATURES") from this testcase.

andymc12:
Tweaked and added test case. New #build

jim-krueger (approved):

andymc12:
#libby Not sure why an unrelated unit test would fail in the GH Action, but the RTC build was 100% successful, so I'll merge.

@aguibert fyi.

PR #15048, opened 2020-11-19
Make OpenAPI 2.0 fully scan module directories

### LINKED ISSUE #15033: Libraries within a .war file are not scanned if the .war is expanded
**Describe the bug**
Libraries within a .war file are not scanned if the .war is expanded. Any annotations on classes within this .war will be picked up by JAX-RS, but won't be processed by OpenAPI, leading to content missing in the output document.

**Steps to Reproduce**
Deploy a .war with
```
<applicationManager autoExpand="true"/>
```

**Expected behavior**
Behaviour should be the same whether the .war is expanded or not.

**Diagnostic information:**

Azquelt:
When a web module corresponds to a directory on disk, scan the whole
directory, not just the WEB-INF/classes directory, so we don't miss the
contents of WEB-INF/lib.

This makes the behaviour consistent whether the module is an archive or
a directory. However, in both cases we're potentially scanning more than
we should.

Fixes #15033


Azquelt:
#build

Emily-Jiang (approved):
LGTM!

Azquelt:
#run-libby-bot

PR #15052, opened 2020-11-19
Upgrade GitHub Actions to use ubuntu-18.04

### LINKED ISSUE #15051: GitHub Action to use specific OS version
GitHub actions is rolling out new default OS version here: https://github.com/actions/virtual-environments/issues/1816

Which causes a warning in each build

For now we should specify `ubuntu-18.04` and later upgrade to `ubuntu-20.04`

KyleAure:
Fixes #15051


KyleAure:
No longer needed resolving in another PR.

PR #15059, opened 2020-11-19
message not ported to OpenLiberty

njr-11:
The key for a translatable message, PARK_OR_REASSOCIATE_FAILED_W_J2CA0083, is used within ported code within HCMDetails. However, the message itself was not yet ported to OpenLiberty.  When I looked over the message to port it, it appeared to refer to internals useful for the support team rather than something meaningful to the end user.  Under this pull, I'm replacing it with debug trace rather than porting the message over.


njr-11:
#build

jmstephensgit (approved):

PR #15062, opened 2020-11-19
Update and add description for REST stat metrics for MP Metris 3.0

Channyboy:
Updates the description for `REST.request` and adds a new description for `REST.request.unmappedException`
 #build


Channyboy:
The errors that occurred in the build were during gradle builds where 403s were encountered will querying artifactory.
There were no test issues. All tests passed.

pgunapal (approved):
LGTM

Channyboy:
L2 and Message review not needed.
The description modified here is the description for the affected metrics as directed by the  MP Metrics specification .

PR #15065, opened 2020-11-20
Ignore Quotes in legacy EJB msgs

tkburroughs:
Avoid warnings from globalization validation for legacy EJB messages
that cannot be changed. Other messages will continue validation.


tkburroughs:
#build

olendvcook (approved):

PR #15086, opened 2020-11-23
OpenAPI-2.0: write cache entries asynchronously

Azquelt:
Part of the startup process is to write the generated OpenAPI model to
the cache so that it can be used next time the application starts.

However, since the cache entry isn't needed for the current run of the
application, we can write it in the background to allow the app to start
faster.

For #14678

### DIFF
--- a/dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
+++ b/dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
@@ -64,4 +64,7 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
 OPENAPI_APPLICATION_PROCESSING_ERROR=CWWKO1661E: An error occured when processing application {0} and an OpenAPI document was not produced.
 OPENAPI_APPLICATION_PROCESSING_ERROR.explanation=A runtime error occured when processing the application.
 OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Gather logs and contact IBM service.
-  
+
+OPENAPI_CACHE_WRITE_ERROR=CWWKO1662W: An unexpected error occurred while writing the OpenAPI cache for application {0}. The error is {1}
+OPENAPI_CACHE_WRITE_ERROR.explanation=A runtime error occurred while caching the OpenAPI model for the application.
+OPENAPI_CACHE_WRITE_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
--- a/dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
+++ b/dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
@@ -65,6 +65,6 @@ OPENAPI_APPLICATION_PROCESSING_ERROR=CWWKO1661E: An error occured when processin
 OPENAPI_APPLICATION_PROCESSING_ERROR.explanation=A runtime error occured when processing the application.
 OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Gather logs and contact IBM service.
 
-OPENAPI_CACHE_WRITE_ERROR=CWWKO1662W: An unexpected error occurred while writing the OpenAPI cache for application {0}. The error is {1}
-OPENAPI_CACHE_WRITE_ERROR.explanation=A runtime error occurred while caching the OpenAPI model for the application.
+OPENAPI_CACHE_WRITE_ERROR=CWWKO1662W: An unexpected error occurred when the OpenAPI cache was written for the {0} application. The error is {1}
+OPENAPI_CACHE_WRITE_ERROR.explanation=A runtime error occurred when the OpenAPI model was cached for the application.
 OPENAPI_CACHE_WRITE_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
--- a/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationRegistry.java
+++ b/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationRegistry.java
@@ -14,6 +14,12 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+
 import com.ibm.websphere.ras.Tr;
 import com.ibm.websphere.ras.TraceComponent;
 import com.ibm.ws.container.service.app.deploy.ApplicationInfo;
@@ -31,14 +37,30 @@ import io.openliberty.microprofile.openapi20.utils.LoggingUtils;
  *       deployed that contains multiple web modules, an OpenAPI document will only be generated for the first Web
  *       Module that generates an OpenAPI document.
  */
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE, service = ApplicationRegistry.class)
 public class ApplicationRegistry {
 
     private static final TraceComponent tc = Tr.register(ApplicationRegistry.class);
     
-    private static final ApplicationRegistry INSTANCE = new ApplicationRegistry();
+    private static ApplicationRegistry INSTANCE;
     private Map<String, ApplicationInfo> applications = Collections.synchronizedMap(new HashMap<>());
     private ApplicationInfo currentApp = null;
     private OpenAPIProvider currentProvider = null;
+    
+    @Reference
+    protected ApplicationProcessor applicationProcessor;
+    
+    @Activate
+    protected void activate() {
+        INSTANCE = this;
+    }
+    
+    @Deactivate
+    protected void deactivate() {
+        if (INSTANCE == this) {
+            INSTANCE = null;
+        }
+    }
 
     /**
      * The getInstance method returns the singleton instance of the ApplicationRegistry
@@ -93,6 +115,9 @@ public class ApplicationRegistry {
         // Check to see if the application being remove is the currentApp
         if (currentApp != null && currentApp.getName().equals(removedAppInfo.getName())) {
             
+            // Current app is being removed, ensure it has finished writing its cache entry
+            currentProvider.waitForBackgroundTasks();
+            
             // The currentApp is being removed... see if any of the other applications implement a JAX-RS REST API
             currentApp = null;
             currentProvider = null;
@@ -132,14 +157,11 @@ public class ApplicationRegistry {
     private void processApplication(ApplicationInfo appInfo) {
         // Only scan the application if we do not have a current application
         if (currentApp == null) {
-            currentProvider = ApplicationProcessor.processApplication(appInfo);
+            currentProvider = applicationProcessor.processApplication(appInfo);
             if (currentProvider != null) {
                 currentApp = appInfo;
             }
         }
     }
 
-    private ApplicationRegistry() {
-        // This class is not meant to be instantiated.
-    }
 }
--- a/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
+++ b/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
@@ -3,17 +3,26 @@ package io.openliberty.microprofile.openapi20;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.eclipse.microprofile.openapi.models.OpenAPI;
 import org.eclipse.microprofile.openapi.models.Paths;
 import org.eclipse.microprofile.openapi.models.servers.Server;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
 import com.ibm.ws.container.service.app.deploy.WebModuleInfo;
 
+import io.openliberty.microprofile.openapi20.utils.MessageConstants;
 import io.openliberty.microprofile.openapi20.utils.OpenAPIUtils;
 import io.smallrye.openapi.runtime.io.Format;
 
 public class WebModuleOpenAPIProvider implements OpenAPIProvider {
+    
+    private static final TraceComponent tc = Tr.register(WebModuleOpenAPIProvider.class);
 
     // The WebModuleInfo for the application/module that the OpenAPI model was generated from
     private final WebModuleInfo webModuleInfo;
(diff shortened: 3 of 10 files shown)


Azquelt:
#build

dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
```diff
@@ -64,4 +64,7 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
 OPENAPI_APPLICATION_PROCESSING_ERROR=CWWKO1661E: An error occured when processing application {0} and an OpenAPI document was not produced.
 OPENAPI_APPLICATION_PROCESSING_ERROR.explanation=A runtime error occured when processing the application.
 OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Gather logs and contact IBM service.
-  
\ No newline at end of file
+
+OPENAPI_CACHE_WRITE_ERROR=CWWKO1662W: An unexpected error occurred while writing the OpenAPI cache for application {0}. The error is {1}
```
ManasiGandhi, line 68: Change,
" An unexpected error occurred while writing the OpenAPI cache for application {0}. The error is {1}"
to,
"An unexpected error occurred when the OpenAPI cache was written for {0} application. The error is {1}"

dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
```diff
@@ -64,4 +64,7 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_CACHE_WRITE_ERROR.explanation=A runtime error occurred while caching the OpenAPI model for the application.
```
ManasiGandhi, line 69: Change,
"A runtime error occurred while caching the OpenAPI model for the application."
to,
"A runtime error occurred when the OpenAPI model was cached for the application."

ManasiGandhi (changes requested):
Some edit suggestions.

dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
Azquelt: Done.

dev/io.openliberty.microprofile.openapi.2.0.internal/resources/io/openliberty/microprofile/openapi20/resources/OpenAPI.nlsprops
Azquelt: Done.

Azquelt:
#build

Azquelt:
#run-libby-bot

andrea-pichler:
Completed the requested L2 message review

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
```diff
@@ -144,4 +159,14 @@ public String getApplicationPath() {
     public boolean getServersDefined() {
         return serversDefined;
     }
+
+    @Override
+    public void waitForBackgroundTasks() {
```
hutchig, line 164: <bikeshedding>This is OK but the name might be better if it had a notion of there being timeout in it, or the timeout was passed in perhaps</bikeshedding>

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
```diff
@@ -144,4 +159,14 @@ public String getApplicationPath() {
...
+        try {
+            // Note: this will be called before app shutdown so we don't want to wait too long
+            asyncProcessingFuture.get(10, TimeUnit.SECONDS);
+        } catch (TimeoutException | ExecutionException | InterruptedException e) {
+            Tr.warning(tc, MessageConstants.OPENAPI_CACHE_WRITE_ERROR, webModuleInfo.getApplicationInfo().getDeploymentName(), e.toString());
```
hutchig, line 169: I think the message _"An unexpected error occurred when the OpenAPI cache was written for the {0} application. The error is"_ This works here because of the e.toString() having the notion of timeout within it but really that message only fits really well with "ExecutionException".  If e was an interrupted exception it might suggest the cache write was interrupted (which in some sense it might be if we are being deactivated on shutdown that occurs before the cache write's future completes) then we are down to the stacktrace and code knowledge to know that it was somehow the deactivation that was interrupted and not the cache write.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
```diff
@@ -240,42 +253,49 @@ public void write() throws IOException {
                 if (fileEntries.isEmpty())
                     Tr.debug(this, tc, "Not writing cache entry, fileEntries is empty");
```
hutchig, line 254: I don't see that CacheEntry redefines toString in order to get the appName, model etc in the Tr.debug s ?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
```diff
@@ -240,42 +253,49 @@ public void write() throws IOException {
...
+                    Tr.debug(this, tc, "Beginning cache entry write");
                 }
-            }
 
-            // create cache directory
-            Files.createDirectories(cacheDir);
+                // delete cache directory if present
+                if (Files.exists(cacheDir)) {
+                    if (Files.isDirectory(cacheDir)) {
+                        Files.walkFileTree(cacheDir, RECURSIVE_DELETER);
```
hutchig, line 272: Is there a safety issue here if someone could somehow (too) freely configure the cache directory location?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
```diff
@@ -302,7 +322,11 @@ private void readModel(Path input) throws IOException {
     }
 
     private void writeModel(Path output) throws IOException {
-        String jsonModel = OpenApiSerializer.serialize(model, Format.JSON);
+        String jsonModel;
+        // This method runs asynchronously, so synchronize to ensure we don't use the model at the same time as WebModuleOpenAPIProvider
+        synchronized (model) {
```
hutchig, line 327: This is an exclusive lock...but we really only need a Read lock?

Also when one adds a lock, one must always be careful if there is the possibility of trying
to acquire a second lock or waiting for something while holding it. In case another code
path might do the same in reverse - creating a hang. Can any other thread lock this
same model? If so the codepaths need to be checked and a common lock ordering firmly
established.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
```diff
@@ -144,4 +159,14 @@ public String getApplicationPath() {
     public boolean getServersDefined() {
         return serversDefined;
     }
+
+    @Override
+    public void waitForBackgroundTasks() {
+        try {
+            // Note: this will be called before app shutdown so we don't want to wait too long
+            asyncProcessingFuture.get(10, TimeUnit.SECONDS);
```
Emily-Jiang, line 167: Do you think 10s is enough?

hutchig (approved):
I had a few queries/comments but did not see anything that is wrong,
please address the comments inline and I will set them resolved as they are dealt with.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
Azquelt: It doesn't. `this` as the first argument is only used to provide the value for the `id=xxxxxxxxx` column of the trace log to identify the instance that logged the trace message.

We're not tracing out the app name or model here as we will have done so before, so in context of the surrounding trace messages it should be clear enough which app and model is being processed.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
Azquelt: Don't think so since we get the cache directory from OSGi which allocates a private working directory to each bundle. The class Javadoc should probably be clearer that `CacheEntry` will assume it has ownership of the `cacheDir` and is allowed to delete anything within it.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
Azquelt: Correct, we only need a read lock, but there's only two participants here and the other one needs a write lock.

I don't think any other locks are involved in either place that synchronises on the model but I will check.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
hutchig: OK

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
hutchig: OK

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java
hutchig: OK sounds good

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationRegistry.java
```diff
@@ -31,14 +37,30 @@
  *       deployed that contains multiple web modules, an OpenAPI document will only be generated for the first Web
  *       Module that generates an OpenAPI document.
  */
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE, service = ApplicationRegistry.class)
 public class ApplicationRegistry {
 
     private static final TraceComponent tc = Tr.register(ApplicationRegistry.class);
     
-    private static final ApplicationRegistry INSTANCE = new ApplicationRegistry();
+    private static ApplicationRegistry INSTANCE;
```
tevans78, line 45: Perhaps we could get rid of this static INSTANCE? Use something like this to look it up instead?
```
        Bundle bundle = FrameworkUtil.getBundle(ApplicationRegistry.class);
        ApplicationRegistry applicationRegistry = AccessController.doPrivileged((PrivilegedAction<ApplicationRegistry>) () -> {
            BundleContext bCtx = bundle.getBundleContext();
            ServiceReference<ApplicationRegistry> svcRef = bCtx.getServiceReference(ApplicationRegistry.class);
            return svcRef == null ? null : bCtx.getService(svcRef);
        });
```

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/utils/MessageConstants.java
```diff
@@ -12,17 +12,69 @@
 
 public class MessageConstants {
 
+    /**
+     * CWWKO1652W: The server cannot read the specified CSS document {0} due to {1} : {2}.
```
tevans78, line 16: The CWWKO prefix is already taken;

CWWKO | orb | WebSphere Application Server - Alpine project
https://github.ibm.com/was-liberty/WS-CD-Open/wiki/CWW-Message-Prefixes
(IBM internal url)

Azquelt: I think we need a separate issue for this as it's also used throughout the mpOpenAPI-1.0 and 1.1 features. https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.microprofile.openapi/resources/com/ibm/ws/microprofile/openapi/resources/OpenAPI.nlsprops#L18

tevans78: https://github.com/OpenLiberty/open-liberty/issues/15108

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationRegistry.java
Azquelt: We could, and we could also probably get the `BundleContext` in a static initializer, rather than looking it up every time, though we would still need the PrivilegedAction to get the instance and the code is more complicated.

Another option would be to remove the method altogether. The only place that uses the `getInstance` method should now be the servlet which serves the OpenAPI document, so we could have that get hold of the bundle context (which is made available in the servlet context) and look up the service itself.

However, I'm not sure it's worth it to look up a component which is a singleton.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
Azquelt: It's worth noting that we don't expect either of these to happen, and if they do, there's not really a lot that the user can do about it.

For the `TimeoutException` I guess we could have a more explicit message "Timed out waiting for the OpenAPI cache to be written", though I'm not sure what we should put as the useraction here. I thought that treating it as a general unexpected exception which occurred while writing the OpenAPI cache file was reasonable.

For the `InterruptedException` I'm less sure. Thread interruption is generally a signal to give up on whatever we're doing right now. I would never expect that to occur here, except possibly if the server is shutting down? Maybe we would be better wrapping it in a `RuntimeException` and rethrowing it?

Azquelt: I guess maybe for the `TimeoutException` we could instead have an informational message that we timed out waiting and are shutting the app down anyway? We could then say that no action was required.

hutchig: On server shutdown even if the background tasks is Interrupted it should complete the Future
exceptionally so we should not have to wait for an interrupted background task that got stopped
before we got deactivated. Given that, TimedoutException probably will not be seen due to waiting for a
shutdowned cache writer here (which is what I was wondering about) - so although I think perhaps what's logged could be more 'info' than 'warning' (as we are deactivating anyway) and indicates that the shutdown is perhaps not as orderly/designed as we'd expect or at least something is going on that we didn't ...it is a bit of a moot point though so I have no problem with the code staying as it is. If it is occurring unexpectedly, we will at least have something that draws our attention to it with the code as it is.

helyarp (commented):
Text looks good! I'll add the ID reviewed label.

Azquelt:
This didn't make any difference to performance so we didn't merge it.

PR #15090, opened 2020-11-23
Update mpMetrics to use the MP Metrics 3.0 artifacts from sonatype - via artifactory and DHE

Channyboy:
#build

PR #15103, opened 2020-11-25
Update mpMetrics to use the MP Metrics 3.0 artifacts from sonatype - via artifactory and DHE - v2

Channyboy:
#build
Old PR  at #15090 was improperly pulling artifacts from maven central instead and was using older versions (i.e not  3.0.0) with the absence of the  `EXACT` designation in the bundle reference.

If retrieving artifacts from `DHE`, the dependency should only be included in the `oss_ibm.maven` file.
Including it in `oss_dependencies.maven` will force it to look up the affected artifacts on maven and skip DHE lookup when artifactory credentials are not presented. Even if the DHE artifact has already been downloaded to the local `.ibmdhe` repository from a `cnf:initialize` call the build will ignore it.

### DIFF
--- a/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
+++ b/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
@@ -1797,21 +1797,6 @@
       <version>2.3</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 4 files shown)


Emily-Jiang (approved):
LGTM!

PR #15107, opened 2020-11-25
test that HandleList closes handles on EJB destroy

njr-11:
Write a test to confirm that when HandleList is enabled, Liberty automatically closes connection handles that are leaked beyond EJB destroy.
Also, write a test to confirm that when HandleList is disabled, the connection handle is not closed and can still be used.


njr-11:
#build

nmittles (approved):

PR #15171, opened 2020-12-03
HandleList test for Singleton EJB

njr-11:
Write a test of HandleList where a singleton EJB method that leaks a connection handle is invoked concurrently.  Verify that when the method invocation completes, the respective connection handle is automatically closed, while the connection handles from other method invocations remain open until those method invocations end.  This is possible because a separate HandleList is used per method for singleton EJBs.


tkburroughs (approved):

njr-11:
#build

PR #15192, opened 2020-12-04
fix HTTP/2 connection shutdown test problems

wtlucy:
For #15219 and #15220

* add back an IOException for case where response resources are cleaned up earlier than expected due to an error condition
* update most tests which rely on the h2c upgrade to wait for the first stream to complete before sending frames on new (often error-inducing) frames
* increase the http2 client first read timeout


loriadi (approved):
Looks good.

wtlucy:
#build #spawn.fullfat.buckets=com.ibm.ws.transport.http2_fat

wtlucy:
#run-libby-bot

PR #15214, opened 2020-12-08
EL Permission investigate

### LINKED ISSUE #12284: Investigate java 2 sec issue with EL fat bucket
We ran into an issue with java 2 sec while adding the jakarta repeat test to the el-3.0 fat bucket #12153. Locally we were hitting a java sec 2 issue which we had to add a permission for on the app

```xml
    <permission>  
        <class-name>java.lang.RuntimePermission</class-name>
        <name>accessClassInPackage.com.sun.beans.editors</name>  
    </permission>
```

Whats weird is that the issue occurred for both el-3.0 AND with el-4.0. We never encountered this issue before on the previous builds with el-3.0 so we would need to investigate why this happened. On our pre-investigation we noticed that the issue seems to be a java version bug. It doesn't happen with java 8 but it does with java 11 and we found this bug reported https://bugs.openjdk.java.net/browse/JDK-8175324 . The weird issue is that by looking at the server logs on the tests in the java2sec build machines we noticed that the java version it's using is java 11 so it's weird that no error popped up before.

isaacrivriv:
fixes #12284 
Removed previously added j2sec permission to verify Build Break doesn't occur and the failure was just locally. Ran two custom j2sec builds last week with the EL-3, Servlet-4 and Wsoc FAT buckets which  passed without the permission so can confirm this was a false positive and the permission should be removed.

### DIFF
--- a/dev/com.ibm.ws.el.3.0_fat/test-applications/TestEL3.0.war/resources/META-INF/permissions.xml
+++ b/dev/com.ibm.ws.el.3.0_fat/test-applications/TestEL3.0.war/resources/META-INF/permissions.xml
@@ -20,9 +20,4 @@
         <name>getClassLoader</name>  
     </permission>
 
-    <permission>  
-        <class-name>java.lang.RuntimePermission</class-name>
-        <name>accessClassInPackage.com.sun.beans.editors</name>  
-    </permission>
-
 </permissions>


isaacrivriv:
#build

pnicolucci (approved):

PR #15231, opened 2020-12-08
translatable message for autoCloseConnections

njr-11:
Add translatable message for autoCloseConnections that we will be able to use once the capability is put into beta.

### DIFF
--- a/dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
+++ b/dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
@@ -52,7 +52,7 @@ reapTime.desc=Amount of time between runs of the pool maintenance thread. A valu
 # Connection Manager advanced properties
 
 autoCloseCon=Automatically close connections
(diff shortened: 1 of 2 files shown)


dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -51,6 +51,9 @@ reapTime.desc=Amount of time between runs of the pool maintenance thread. A valu
 
 # Connection Manager advanced properties
 
+autoCloseCon=Automatically close connections
+autoCloseCon.desc=Attempts to clean up after applications that leave connections open across the end of a servlet request, enterprise bean instance, managed executor task, contextual task, or managed completion stage. When an unsharable connection is obtained within one of these application artifacts and remains open when it ends, the container attempts to close the connection handle. The container may also close sharable connections that do not support DissociatableManagedConnection. It is recommended that applications always close connections rather than relying on the container, even when this option is enabled.
```
jantley-ibm, line 55: In the first sentence, suggest changing the word "across" to "after".

In the last sentence: we should avoid "recommending". Instead, could we mention benefit(s) of closing connections via the app vs. relying on this setting, or any disadvantages of relying on the setting?

njr-11: Thanks for reviewing. I added a commit that switches across to after and replaces the recommendation with a statement that applications should always follow the specification in closing connections.  It isn't just our recommendation. It's the specification that outlines the correct usage of the programming model.  This is more about us clarifying that, while we provide this option to help cover for incorrect usage, we are not advocating that customers disregard the specification.

jmstephensgit (approved):

njr-11:
#build

njr-11:
#run-libby-bot

PR #15261, opened 2020-12-10
potential thread safety issue with HandleList

njr-11:
HandleList has some temporary code where it inherited from ArrayList as a convenience.  Various methods are synchronized to eliminate concurrent access.  This was useful for a proof of concept but is vulnerable to other code using the other non-synchronized methods of ArrayList to modify or iterate the list.  A simple way to address this is to switch to Vector.  Another possibility would be to stop inheriting from a list implementation and include an array field within the class, in which case we would need to handle resize ourselves.


jmstephensgit (approved):

njr-11:
#build
#run-libby-bot

njr-11:
#build

jmstephensgit (approved):

njr-11:
Build had 1 failure which is a duplicate of a pre-existing issue.

PR #15264, opened 2020-12-10
[15263] enableTraceRequests should be lowercase when in WCCustomerProps

### LINKED ISSUE #15263: HTTP TRACE method requests are rejected with a 403, and `enableTraceRequests="true"` doesn't help
When I attempt a TRACE request to a backend resource, I get:
```
$ curl http://localhost:9080/jaxrs_jaxrs21_ee_client_rxinvoker_web/resource/trace -X TRACE
Error 403: Illegal request. Default implementation of TRACE not allowed.
```

From my understanding, I should be able to disable the TRACE rejection with this config setting:
`<webContainer enableTraceRequests="true"/>`

But that doesn't help - the issue is that the config property `enableTraceRequests` is lowercased, but then when it is checked at runtime, it still uses camel case.  This should be a simple fix.

andymc12:
Fixes #15263


andymc12:
#build

jhanders34 (approved):

pmd1nh (approved):

wtlucy (approved):
good catch, thanks Andy

andymc12:
All failures occur in other builds and are unrelated to these changes.

PR #15265, opened 2020-12-10
Override `doTrace` to avoid automatic TRACE request rejection

andymc12:
The web container rejects TRACE requests by default - with or without the security features enabled.  When security is disabled, this can be worked around by setting `<webContainer enableTraceRequests="true"/>` in the server.xml, but that will not work when security is enabled.  To ensure that TRACE requests are not rejected when security is enabled, we must override the `doTrace` method in IBMRestServlet.


andymc12:
#build

andymc12:
also see related issue #15263

jim-krueger (approved):
LGTM

andymc12:
Single failures occurs in other builds and is not related to these changes.

PR #15275, opened 2020-12-11
Fixes bug where entity is not read correctly when @FormParms are used

andymc12:
Resolves a CTS issue when using `@FormParam` either on the resource method directly as a parameter or as part of a `@BeanParam` (where it is failing multiple CTS tests).  

The issue is that RESTEasy attempts to optimize it's read of the HTTP entity data and only do it once for both `@FormParam` annotations and for the entity content.

So consider the following cases:
```
    @POST
    @Path("formParamOnly")
    public String postFormParamData(@FormParam("foo") String foo) {
        return "fooParam=" + foo + "\n"; 
    }
    @POST
    @Path("entityOnly")
    public String postFormEntityData(String content) {
        return "content=" + content + "\n"; 
    }
    @POST
    @Path("formAndEntity")
    public String postFormParamAndEntityData(@FormParam("foo") String foo, String content) {
        return "fooParam=" + foo + " content=" + content + "\n"; 
    }
```

and three identical requests (with the exception of the path to ensure each method is invoked with the same data):
```
$ curl -X POST -d "bar&foo=foo" http://localhost:9080/rest/formParamOnly
fooParam=foo

$ curl -X POST -d "bar&foo=foo" http://localhost:9080/rest/entityOnly
content=bar&foo=foo

$ curl -X POST -d "bar&foo=foo" http://localhost:9080/rest/formAndEntity
fooParam=foo foo=foo
```

Note in the last request that "bar" is missing.  This is because RESTEasy effectively dropped it when it processed the `@FormParam` parameters. Then when it was time to process the `content` parameter, it didn't read the entity stream again, it just used the processing from the `@FormParam` which did not cache the full entity.

This fix may be just a temporary fix - basically it just prevents the optimization and forces RESTEasy to re-read the entity stream for the content data.  There may be a better way to cache the entity data for all of these paths that doesn't end up losing data (and thus failing CTS tests), but since we're on a deadline, we'll start with this hack.

### DIFF
--- a/dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/core/MessageBodyParameterInjector.java
+++ b/dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/core/MessageBodyParameterInjector.java
@@ -1,0 +1,260 @@
+package org.jboss.resteasy.core;
+
+import org.jboss.resteasy.core.interception.jaxrs.AbstractReaderInterceptorContext;
+import org.jboss.resteasy.core.interception.jaxrs.ServerReaderInterceptorContext;
+import org.jboss.resteasy.resteasy_jaxrs.i18n.LogMessages;
+import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
+import org.jboss.resteasy.spi.HttpRequest;
+import org.jboss.resteasy.spi.HttpResponse;
+import org.jboss.resteasy.spi.MarshalledEntity;
+import org.jboss.resteasy.spi.ReaderException;
+import org.jboss.resteasy.spi.ResteasyProviderFactory;
+import org.jboss.resteasy.spi.ValueInjector;
+import org.jboss.resteasy.spi.interception.JaxrsInterceptorRegistry;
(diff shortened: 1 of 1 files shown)


andymc12:
#build

jkoehler22 (approved):
LGTM

jim-krueger (approved):

andymc12:
Build passed all FAT tests, but failed the gradle audit - apparently I did too many audits? :)

```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':io.openliberty.jakarta.expressionLanguage.4.0:audit'.
> Too many requests (429): Use OSS Index credentials for increased rate limit.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':io.openliberty.org.apache.jasper.expressionLanguage.4.0:audit'.
> Too many requests (429): Use OSS Index credentials for increased rate limit.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

```

I'm assuming this is safe to merge.

PR #15285, opened 2020-12-11
Re-add jbatch_fat after revert

### LINKED ISSUE #15278: Re-introduce change reverted from 15245
Re-introduce change reverted from https://github.com/OpenLiberty/open-liberty/pull/15245

cgianfra:
The commit for the port of jbatch_fat was reverted due to the common fat tooling project being named _fat, causing it to be treated as a FAT project. This re-delivers the ported projects but with jbatch.common_fat renamed to jbatch.fat.common.

Fixes #15278


cgianfra:
Fixes #15278

cgianfra:
#build

jhanders34 (changes requested):
I notice that ojdbc6.jar is checked in twice.  We do not check in binary files usually.  We get them from open source and bring them in via gradle or some other way.  They should not be checked into github.  if it isn't needed it should be removed, but otherwise it should be done differently.

cgianfra:
#build

cgianfra:
#build

jhanders34 (approved):

cgianfra:
#libby

PR #15296, opened 2020-12-14
Use private mirror of Docker Hub

aguibert:
The basic items are:
 - upgrading to testcontainers 1.15.0
 - making it easy for FATs to use a maven central mirror (that way we don't need to manually upload test deps to Artifactory)
 - Make all testcontainers buckets use a Docker hub mirror (to fix rate limiting issues)
 - Split the testcontainers gradle deps into an addTestcontainers task so it becomes easier to update TC versions in the future

Prioritize: JDBC, JAXRS_1, JAXRS_2


aguibert:
#build

aguibert:
#build

aguibert:
#build

aguibert:
#build #run-libby-bot

aguibert:
#build #run-libby-bot

dev/com.ibm.ws.jaxrs.2.1.client_fat/build.gradle
```diff
@@ -9,17 +9,17 @@
...
-      credentials {
-        username userProps.getProperty("artifactory.download.user")
-        password userProps.getProperty("artifactory.download.token")
-      }
-      url ("https://" + gradle.userProps.getProperty("artifactory.download.server") + "/artifactory/wasliberty-maven-remote")
-    }
-  } else {
-    mavenCentral()
-  }
+  if (isUsingArtifactory) {
```
KyleAure, line 12: Looks like this one could be updated to use 
`apply from: '../wlp-gradle/subprojects/maven-central-mirror.gradle'`

dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
```diff
@@ -0,0 +1,84 @@
...
+import com.ibm.websphere.simplicity.log.Log;
+
+import componenttest.topology.utils.ExternalTestService;
+
+/**
+ * An image name substituter is configured in testcontainers.properties and will transform docker image names.
+ * Here we use it to apply a private registry prefix so that in remote builds we use an internal mirror
+ * of Docker Hub, instead of downloading from Docker Hub in each build which causes rate limiting issues.
+ */
+public class ArtifactoryImageNameSubstitutor extends ImageNameSubstitutor {
```
KyleAure, line 25: I do not really understand what this class is doing.  Or more accurately how our infrastructure is set up to support this.  

Do we have a system set up that caches docker images from docker hub and hosts them within our test environment? 

Will developers running tests locally still be able to run against our remote docker machines?

KyleAure (approved):

dev/com.ibm.ws.jaxrs.2.1.client_fat/build.gradle
aguibert: Originally I had this one using the common gradle file, but the FAT started failing with a non-obvious issue and I didn't want to delay these changes any further. The only difference between the two is that the common file does `repositories.clear()` and the original file did not, so that must be where the behavior difference was

dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
aguibert: The image name substitutor is an SPI that Testcontainers recently added specifically for addressing the docker hub rate limiting issue:
https://www.testcontainers.org/features/image_name_substitution/

Basically you can configure an `ImageNameSubstitutor` globally (as we do in this PR) and it will remap all of the docker image names used by Testcontainers. So if some FAT bucket does:
```java
public static GenericContainer myContainer = new GenericContainer("someImage:1.0");
```
When the substitutor is activated the code will _effectively_ be:
```java
public static GenericContainer myContainer = new GenericContainer("com.mycompany.registry/someImage:1.0");
```
Where `com.mycompany.registry` is a companies private docker registry (ideally a mirror of Docker Hub). I've set up a mirror registry within the IBM network to handle this. A registry mirror is basically a lazily-populated cache. If a bucket does a pull of a new image `someImage:1.0`, the mirror will check to see if it's available within the mirror, and if not it'll download the image from Docker Hub and cache it for future use. So that way the next time we do a pull of `someImage:1.0` it comes from the mirror registry and we don't need to touch Docker Hub, thus avoiding the rate limiting issue.

> Will developers running tests locally still be able to run against our remote docker machines?

Yes. If we detect that remote docker is being used, then the new `ExternalTestServiceDockerClient.setupTestcontainers` method will rewrite the `.testcontainers.properties` file to use the `ArtifactoryImageNameSubstitutor` and it will generate a `~/.docker/config.json` with a the private docker registry mirror auth token.

KyleAure: Okay thanks for that extra explanation! 👍

aguibert:
#build #run-libby-bot

aguibert:
#build #run-libby-bot

Issue #12387, opened 2020-05-27
Investigate testJspFeatureChange test for EE9 in the jsp.2.3_fat bucket


pnicolucci:
When this test is modified to run with jsp-3.0 and jsp-2.3 it fails because:

```
package com.ibm.ws.jsp23.webcontainerext;

import com.ibm.ws.jsp.webcontainerext.JspVersion;

public class JspVersionImpl implements JspVersion {

    public JspVersionImpl() {}

    @Override
    public String getJspVersionString() {
        return "2.3";
    }
}
```

We don't have a getJspVersionString() for 3.0.

1) Should we split the 2.3 factories out into their own project / bundle and only load with the 2.3 feature and then create a 3.0 factories bundle and update and then only load those with the jsp 3.0 feature?
2) Should we implement something like the ServletVersion or use the JavaEEVersion?
3) Can we do something with the transformer tool here?

volosied:
Merged #13173.

PR #15319, opened 2020-12-15
Change JSP Version String with Transformation, Update JSP 2.3 FAT

volosied:
Potential fix for #12387 that uses the transformer

Instead of 2.3=3.0, we have a rule for JSPVersion2.3=JSPVersion3.0 since "2.3" and "3.0" are too general.


volosied:
#build

volosied:
#build

PR #15331, opened 2020-12-16
Adds a service loader reference to CDIRuntimeImpl (which implements C…

benjamin-confino:
Adds a service file to work around the CDI.java SPI update clearing out the provider we set in AbstractCDIRuntime.start if CDI.current() is ever called in the context of a non-CDI app.

### DIFF
--- a/dev/io.openliberty.jakarta.cdi.3.0/bnd.bnd
+++ b/dev/io.openliberty.jakarta.cdi.3.0/bnd.bnd
@@ -40,12 +40,16 @@ Import-Package:  jakarta.el; version="[4.0.0,5.0.0)", \
  jakarta.enterprise.inject.spi.configurator; version=3.0, \
  jakarta.enterprise.util; version=3.0, \
  jakarta.inject; version=2.0, \
- jakarta.interceptor;version=2.0
+ jakarta.interceptor;version=2.0, \
+ org.osgi.framework
 
 Include-Resource: \
+  META-INF=resources/META-INF, \
   @${repo;jakarta.enterprise:jakarta.enterprise.cdi-api;[3.0,4)}!/beans_1_0.xsd, \
   @${repo;jakarta.enterprise:jakarta.enterprise.cdi-api;[3.0,4)}!/beans_1_1.xsd, \
   @${repo;jakarta.enterprise:jakarta.enterprise.cdi-api;[3.0,4)}!/beans_2_0.xsd
+
+Private-Package: com.ibm.ws.cdi.provider.service
 
 
 instrument.disabled: true
@@ -54,4 +58,6 @@ publish.wlp.jar.suffix: dev/api/spec
 
 -buildpath: \
 	jakarta.inject:jakarta.inject-api;version=2.0.0,\
-	jakarta.enterprise:jakarta.enterprise.cdi-api;version=3.0.0
+	jakarta.enterprise:jakarta.enterprise.cdi-api;version=3.0.0,\
+	com.ibm.wsspi.org.osgi.service.component.annotations,\
+	org.eclipse.osgi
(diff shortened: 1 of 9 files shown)


benjamin-confino:
#build

benjamin-confino:
#build

dev/com.ibm.websphere.javaee.cdi.1.2/bnd.bnd
```diff
@@ -34,18 +34,19 @@ Import-Package:  javax.el; version="[3.0.0,4.0.0)", \
...
  javax.enterprise.util;version="1.1", \
  javax.inject;version=1, \
- javax.interceptor;version="1.2"
+ javax.interceptor;version="1.2", \
+ com.ibm.ws.cdi.liberty
 
 Include-Resource: \
+  META-INF=resources/META-INF, \
   @${repo;javax.enterprise:cdi-api;1.2}!/beans_1_0.xsd, \
   @${repo;javax.enterprise:cdi-api;1.2}!/beans_1_1.xsd
```
Azquelt, line 43: I would strongly suggest not changing this for CDI 1.2 and 2.0 API bundles where it's not needed

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/CDIProviderService.java
```diff
@@ -0,0 +1,74 @@
...
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+
+import io.openliberty.cdi.spi.CDIExtensionMetadata;
+
+/**
+ * This class is to get hold all necessary services.
+ */
+@Component(name = "com.ibm.ws.cdi.liberty.CDIRuntimeImpl", service = {CDIProvider.class}, property = { "service.vendor=IBM" })
+public class CDIProviderService implements CDIProvider {
```
Azquelt, line 39: This class does not need to be a Component as it's managed by `ServiceLoader` rather than as an OSGi service.

The Javadoc should also be improved to say it's a thin layer which looks up the real `CDIProvider` as an OSGi service.

dev/io.openliberty.jakarta.cdi.3.0/bnd.bnd
```diff
@@ -40,9 +40,11 @@ Import-Package:  jakarta.el; version="[4.0.0,5.0.0)", \
  jakarta.enterprise.inject.spi.configurator; version=3.0, \
  jakarta.enterprise.util; version=3.0, \
  jakarta.inject; version=2.0, \
- jakarta.interceptor;version=2.0
+ jakarta.interceptor;version=2.0, \
+ com.ibm.ws.cdi.liberty
```
Azquelt, line 44: This import is problematic.

There are cases where another feature might want to include the CDI API bundle, (or the CDI API feature) so that it can work with or without CDI present. I think this import will prevent the API bundle from resolving if the rest of CDI is not present.

I think the easiest way to do this would be to:
 - Add `CDIProvider` to the list of services provided by the `CDIRuntimeImpl` component
 - Remove all references to `CDIRuntimeImpl` and `CDIService` from `CDIProviderService`, just look up `CDIProvider` as a service instead
 - Move `CDIProviderService` into the API bundle.

If we didn't want to add non-API things to the API bundle, you could
 - Put the services file and `CDIProviderService` into a fragment which attaches to the API bundle
 - This would still require making the `CDIRuntimeImpl` component a `CDIProvider` service

Azquelt (changes requested):
As a whole, this workaround isn't very clean from an OSGi point of view.

I've added comments to make it cleaner, and because it's not very nice, I'd recommend limiting it to only the 3.0 CDI API bundle where it's currently required.

Azquelt (approved):

benjamin-confino:
#build

benjamin-confino:
#run-libby-bot

PR #15340, opened 2020-12-18
11793 - return HTTP 415 when JAX-RS request Content-Type header contains an invalid charset

WhiteCat22:
The error occurs while we are setting up the CXF Message for processing by the PhaseInterceptorChain. The exception is thrown back to Webcontainer before we can process it ourselves. Since JAX-RS can't process the exception, Webcontainer has to process it, however, webcontainer doesn't know what to do, so it sets the error code to HTTP 500 in `WebAppErrorReport.constructErrorReport()`

I added the new Webcontainer Exception: `InvalidMediaTypeException` so that I could throw it from an exception from JAX-RS  that webcontainer would understand and map it to a 415 response code in `WebAppErrorReport.constructErrorReport()`.
By adding `InvalidMediaTypeException` to Webcontainer, I avoid any JAX-RS dependencies in Webcontainer and can send back a better response code than HTTP 500.

I also created 2 tests:
- One tests that a Content-Type with and invalid charset properly returns HTTP 415
- One tests that an invalid json payload returns an HTTP 400

I had previously gotten the invalid payload scenario to work, but wanted a test to compare why the JAX-RS exceptions were being mapped for an invalid payload and not in the case of an invalid Content-Type header.
The difference is that the invalid payload gets processed in the PhaseInterceptorChain, which handles sending the response, whereas invalid Content-Type throws an exception back to Webcontainer during setup and relies on Webcontainer to return the response.

### DIFF
--- a/dev/com.ibm.ws.jaxrs.2.0.common/bnd.bnd
+++ b/dev/com.ibm.ws.jaxrs.2.0.common/bnd.bnd
@@ -188,7 +188,8 @@ instrument.classesExcludes: com/ibm/ws/jaxrs20/internal/resources/*.class, \
 	com.ibm.websphere.appserver.spi.threading, \
 	com.ibm.ws.org.osgi.annotation.versioning;version=latest, \
 	com.ibm.ws.jaxrs.2.x.config;version=latest,\
-	javax.activation:activation;version=1.1
+	javax.activation:activation;version=1.1, \
+	com.ibm.ws.webcontainer;version=latest
 
 -testpath: \
 	org.hamcrest:hamcrest-all;version=1.3, \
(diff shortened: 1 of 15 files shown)


WhiteCat22:
#build

jim-krueger (approved):

dev/com.ibm.ws.webcontainer/src/com/ibm/ws/webcontainer/filter/WebAppFilterManager.java
```diff
@@ -1358,11 +1357,6 @@ else if (servletWrapper instanceof GenericServletWrapper){
 //            if (!invokedAsyncErrorHandling){
             ServletErrorReport errorReport = WebAppErrorReport.constructErrorReport(re, dispatchContext.getCurrentServletReference());
             dispatchContext.pushException(re);
```
pmd1nh, line 1359: Why are removing this FFDC?

WhiteCat22:
#build

WhiteCat22:
#run-libby-bot

pmd1nh (approved):

WhiteCat22:
green build, waiting to merge since we are currently at BuildCon 1

PR #15391, opened 2020-12-31
various threading bugs in the Liberty scheduled executor and get it to run tasks on a policy executor

### LINKED ISSUE #15390: various thread safety issues in the Liberty scheduled executor
While making some updates (and adding some tests) to get the Liberty scheduled executor to run scheduled tasks via a policy executor, I came across a couple of thread safety bugs in the Liberty scheduled executor.

1. Intermittent issue where canceling a repeating task sometimes doesn't actually cancel it.  This happens because the mechanism for swapping out the m_schedFuture in SchedulingRunnableFixedHelper isn't thread safe.  When requesting a cancel around the same time an execution is finishing and rescheduling, the cancel can be sent to the old, already-completed future rather than the future that represents the next scheduled execution.
2. Intermittent issue where various ScheduledFuture operations encounter NullPointerException because m_schedFuture is initialized _after_ the task is actually scheduled. There is no way to reverse the ordering because the scheduling produces the value to set. However, code can be updated to allow for the possibility of seeing a null m_schedFuture.

Another bug is that when a repeating task execution fails to submit to the executor, only the SchedulingHelper's latch is counted down, with the SchedulingRunnableFixedHelper remaining unaware.  From the user's perspective, the future for the repeating scheduled task never shows as begin done and get operations time out, even though the task has actually aborted and will never run again.  This issue may have been hidden in the past because prior to usage with a policy executor, submits directly to the Liberty global tend to not raise exceptions.

njr-11:
fixes #15390

### DIFF
--- a/dev/com.ibm.ws.threading/src/com/ibm/ws/threading/ScheduledPolicyExecutorTask.java
+++ b/dev/com.ibm.ws.threading/src/com/ibm/ws/threading/ScheduledPolicyExecutorTask.java
@@ -1,0 +1,28 @@
+/*******************************************************************************
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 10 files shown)


njr-11:
#build

KyleAure (approved):

PR #15427, opened 2021-01-06
tests for suspend and resume of scheduled executor via a policy executor

njr-11:
Policy executor appears to already have suspend/resume capability by reconfiguring the max concurrency to 0 (suspend) and reconfiguring it back to its original value (resume).  Add some tests for this in various scenarios to help demonstrate that it will work as a suspend/resume mechanism.  This includes:

- suspend before scheduling tasks
- suspend after task is scheduled
- suspend within a repeating task execution itself

and include resume in all of these scenarios as well.


jmstephensgit (approved):

njr-11:
#build

njr-11:
#build

PR #15437, opened 2021-01-06
Injecting app defined as server name in web xml

andymc12:
This should resolve at least one RESTful WS TCK issue - where no app is injected via `@Context Application app` into a resource method.


andymc12:
#build

jim-krueger (approved):

PR #15445, opened 2021-01-07
Ensure async servlet support is enabled when `@ApplicationPath` not used

andymc12:
When there is a web.xml that specifies the JAX-RS application to use (and defines the app path in lieu of the `@ApplicationPath("/somePath")` annotation), we need to ensure that async servlet support is enabled.


andymc12:
#build

jim-krueger (approved):

WhiteCat22 (approved):

PR #15453, opened 2021-01-08
Fix o auth java2

c00crane:
The OAuth FAT tests are failing when run with Java 2 enabled.
Add a permission and fix a jar name that should already have permissions.


c00crane:
#build

ayoho (approved):

PR #15457, opened 2021-01-08
Avoid calling WeldBootstrap when looking up a BeanManager to prevent …

benjamin-confino:
…thread deadlocks


benjamin-confino:
#build

tevans78 (approved):

PR #15482, opened 2021-01-11
Increase heap size for Maven surefire plugin for MP Rest Client TCKs

andymc12:
Should resolve intermittent failures where insufficient heap size results in Javacores.


jkoehler22 (approved):
LGTM

PR #15484, opened 2021-01-11
Upgrade to bnd 5.2.0

fritze2:
Upgrading to bnd 5.2.0 has performance improvements mentioned by https://github.com/bndtools/bnd/wiki/Changes-in-5.2.0 and https://github.com/bndtools/bnd/pull/4038.

### DIFF
--- a/dev/cnf/resources/bnd/bundle.props
+++ b/dev/cnf/resources/bnd/bundle.props
@@ -52,3 +52,6 @@ Import-Package: ${defaultPackageImport}
 #require explicitly specifying all ds and metatype annotated classes.
 -dsannotations:
 -metatypeannotations:
+
+# With bnd 5.2.0 The importing of java.* packages need to be disabled to maintain legacy behavior.
+-noimportjava: true
--- a/dev/gradle.properties
+++ b/dev/gradle.properties
@@ -13,7 +13,7 @@ bnd_cnf=cnf
 
 # bnd_plugin is the dependency declaration for the bnd gradle plugin
 #bnd_plugin=biz.aQute.bnd:biz.aQute.bnd.gradle:+
-bnd_plugin=biz.aQute.bnd:biz.aQute.bnd.gradle:5.0.1
+bnd_plugin=biz.aQute.bnd:biz.aQute.bnd.gradle:5.2.0
 
 # The URL to the bnd build repo.
 bnd_repourl=https://repo1.maven.org/maven2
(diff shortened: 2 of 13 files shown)


fritze2:
#build

fritze2:
#build

fritze2:
#build

fritze2:
#build

fritze2:
#build

fritze2:
#build

dev/cnf/resources/bnd/bundle.props
```diff
@@ -52,3 +52,6 @@ Import-Package: ${defaultPackageImport}
 #require explicitly specifying all ds and metatype annotated classes.
 -dsannotations:
 -metatypeannotations:
+
+# With bnd 5.2.0 The importing of java.* packages need to be disabled to maintain legacy behavior.
```
gjwatts, line 56: Why do we need to disable importing the java.* packages?

fritze2: I disabled importing java.* packages because they are not needed for the bundle to run. Some bundles use `*` for their Import-Package to find all packages from the `-buildpath`, which ends up including java.* now regardless of `-noimportjava: true` being set. So where bundles use `*`, I added a `!java.*` line to prevent those from importing java.* packages.

If I didn't remove the java.* imports, then the bundles would fail to start on server start. For example,
```
junit.framework.AssertionFailedError: Found issues in console.log after start: [[ERROR   ] CWWKE0702E: Could not resolve module: com.ibm.ws.org.apache.cxf.cxf.rt.transports.http.hc.3.2 [194]
```

dev/gradle.properties
```diff
@@ -13,7 +13,7 @@ bnd_cnf=cnf
 
 # bnd_plugin is the dependency declaration for the bnd gradle plugin
 #bnd_plugin=biz.aQute.bnd:biz.aQute.bnd.gradle:+
-bnd_plugin=biz.aQute.bnd:biz.aQute.bnd.gradle:5.0.1
+bnd_plugin=biz.aQute.bnd:biz.aQute.bnd.gradle:5.2.0
```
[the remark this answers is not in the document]
fritze2: Yeah I think we should, but it since it doesn't need to happen in the same PR, I plan to update those in another PR.

dev/cnf/resources/bnd/bundle.props
gjwatts: Thanks for taking the time to explain that Ethan.

dev/gradle.properties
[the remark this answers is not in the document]
gjwatts: Ok, we can use #16217 for that if you want.

gjwatts (approved):

fritze2:
#run-libby-bot

PR #15492, opened 2021-01-12
Handle beans.xml with discoveryMode="none" - no beanManager

andymc12:
Fixes cases where a user app has a beans.xml that declares bean-discovery-mode="none" - in this case there would be no bean manager for the app/module, and so we want to resort back to the non-CDI injection factory to handle JAX-RS `@Context` injection, but not CDI `@Inject` injection.


andymc12:
Manually submitted personal build at https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=__L0u8FRsEeuFI6_BZVYvOw - visible only to IBMers.

jim-krueger (approved):

jkoehler22 (approved):
LGTM

PR #15531, opened 2021-01-13
quarantine mislabeled projects

KyleAure:
com.ibm.ws.security.saml.sso_fat.common is not a fat suite and shouldn't have :buildandrun executed against it.


gjwatts (approved):

KyleAure:
#libbybot

Issue #14934, opened 2020-11-11
JAX-RS client creates a new SSLSocketFactory for every request


WhiteCat22:
I removed our caching of the SSLSocketFactories in #13027 because the underlying security service caches them.

Unfortunately, there is a missing piece in the layer between the security team's cache and JAX-RS. As a result, our JAX-RS client now creates a new SSLSocketFactory for each request.

WhiteCat22:
Issue was confirmed resolved by #15534 and should go in to 21.0.0.2

PR #15534, opened 2021-01-13
13027 - cache the SSLContext and SSLSocketFactory

WhiteCat22:
Cache the SSLContext and the SSLSocket Factory, then get the SSLContext from JSSEHelper and compare against our cached copy each request to determine if we need to refresh our cached SSLSocketFactory

We need to do this because JSSEHelper only caches the SSLContext. Calling SSLContext.getSocketFactory() creates a new SSLSocketFactory every time. So, we need to cache the socket factory for re-use, but also the SSLContext so that we can compare it against the current SSLContext. If they are not equal, then the SSL config has changed and we need to get the new SSLContext and cache the new SSLSocketFactory to support dynamic SSL updates.

#14934

### DIFF
--- a/dev/com.ibm.ws.jaxrs.2.0.security/src/com/ibm/ws/jaxrs20/appsecurity/security/JaxRsSSLManager.java
+++ b/dev/com.ibm.ws.jaxrs.2.0.security/src/com/ibm/ws/jaxrs20/appsecurity/security/JaxRsSSLManager.java
@@ -17,6 +17,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 
+import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
 
 import com.ibm.websphere.ras.Tr;
@@ -27,6 +28,9 @@ import com.ibm.websphere.ssl.SSLException;
 
 public class JaxRsSSLManager {
     private static final TraceComponent tc = Tr.register(JaxRsSSLManager.class);
+
+    private static final Map<String, SSLSocketFactory> socketFactories = new HashMap<>();
+    private static final Map<String, SSLContext> sslContexts = new HashMap<>();
 
     /**
      * Get the SSLSocketFactory by sslRef, if could not get the configuration, try use the server's default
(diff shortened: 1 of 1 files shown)


WhiteCat22:
#run-libby-bot
#build

acdemyers (approved):

dev/com.ibm.ws.jaxrs.2.0.security/src/com/ibm/ws/jaxrs20/appsecurity/security/JaxRsSSLManager.java
```diff
@@ -28,6 +29,9 @@
 public class JaxRsSSLManager {
     private static final TraceComponent tc = Tr.register(JaxRsSSLManager.class);
 
+    private static final Map<String, SSLSocketFactory> socketFactories = new HashMap<>();
+    private static SSLContext sslContext = null;
```
andymc12, line 33: Wouldn't we want this to be a Map too? (Map<String, SSLContext>) It seems like we might have more than one SSLContext configured for a given server, and if we just have one static field, then we might always skip the cache (i.e. app1 uses context1, then app2 uses context2 forcing a swap of the sslContext field, etc.).

dev/com.ibm.ws.jaxrs.2.0.security/src/com/ibm/ws/jaxrs20/appsecurity/security/JaxRsSSLManager.java
```diff
@@ -78,8 +82,23 @@ public Boolean run() throws SSLException {
...
+            SSLContext context = jsseHelper.getSSLContext(connectionInfo, sslProps);
+
+            boolean recache = false;
+            if (sslContext == null || !context.equals(sslContext)) {
+                // SSL config has changed, re-cache the SSLSocketFactory
+                synchronized (sslContext) {
+                    sslContext = context;
+                }
+                recache = true;
+            }
```
andymc12, line 94: see above comments about using a Map for sslContext.

andymc12 (changes requested):

WhiteCat22:
#run-libby-bot
#build

andymc12 (approved):
LGTM - thanks @WhiteCat22!

WhiteCat22:
Testing is complete, this solved the issue of persistent TLS connections breaking while also supporting dynamic ssl config changes without having to restart the server.

PR #15547, opened 2021-01-14
CL22175 single catch-up for fixed-rate execution

njr-11:
Add the ability to override the algorithm for computing the next fixed-rate execution.  By default, when a fixed-rate scheduled task is delayed, the Liberty Scheduled Executor is continually running all missed executions until caught up.  In some cases, this might be the desired behavior, however in other cases only a single catch-up execution is wanted before returning to the fixed-rate schedule.


mswatosh (approved):

njr-11:
#build

njr-11:
Build has 1 test failure which is a pre-existing issue

PR #15549, opened 2021-01-14
test that ManagedThreadFactory is still working after app restart if looked up again

njr-11:
Add test case that looks up a ManagedThreadFactory after application restart and verifies that it is still usable.


njr-11:
#build

jmstephensgit (approved):

PR #15562, opened 2021-01-15
Early exit when dir.listFiles() returns null

andymc12:
Resolves NPE and traces condition where directory's `listFiles()` method returns null - according to the Javadoc, this can occur when the directory does not exist or if some sort of IOException has occurred.


andymc12:
#build

jim-krueger (approved):

PR #15563, opened 2021-01-15
Synchronize access to configProperties map

andymc12:
Some test builds found a ConcurrentModificationException while accessing Liberty's RESTEasy config impl's config properties hashmap.  Synchronizing access to the map should prevent these exceptions.


andymc12:
#build

jim-krueger (approved):

Issue #15471, opened 2021-01-11
Control SelfExtract for *nix based containers


lborupj:
When packaging OpenLiberty into a self-extracting archive (JAR) and adding that to an image based upon the distroless ( https://github.com/GoogleContainerTools/distroless ) Java 11 image, you get an error saying that chmod is not available (as expected) BUT it would be really nice if I could set ENV vars to control whether chmod'ing to executable and other native tools should/can be used.

I have a work around (not pretty) to set "-Dos.name=windows" even though it is running on linux but atleast OL extracts archive and does start.

cbridgha:
Thanks - we'll take a look

PR #15565, opened 2021-01-17
Guard against environments where chmod doesn't exist

cbridgha:
Recognize the exceptions thrown when the "chmod" command doesn't exist. - Assuming the user runs as root, or already has permission so skipping this step won't have an effect.

i tested creating a runnable jar before and after this change and deploying on a image based on gcr.io/distroless/java:11


cbridgha:
#build

cbridgha:
fixes #15471

jimblye (approved):

PR #15573, opened 2021-01-19
15517 cdi 3.0 beansxml fats

benjamin-confino:
This enables java EE9 on the CDI beansxml and beansxml.implicit fat buckets.


benjamin-confino:
This was resolved in a different PR.

PR #15584, opened 2021-01-19
Enable MPJWT InvalidToken test

clarkek123:
This PR enables the EmptyTokenTest method invalidToken with its own server config and TAI property set.


clarkek123:
#build

clarkek123:
This PR is not needed now that a solution has been implemented in the code so that all tests can run under the same server config. I am closing this PR.

Issue #9809, opened 2019-11-18
Enable trustedSensitiveHeaderOrigin to be configured with a IP segment


pnicolucci:
Current description of the property: https://openliberty.io/docs/ref/config/#httpDispatcher.html#trustedSensitiveHeaderOrigin

 **<Current specification>**
_-trustedSensitiveHeaderOrigin_
the "trustedSensitiveHeaderOrigin" property should be configured with a comma-separated list of IP addresses corresponding to those of any WebSphere-aware proxy servers in front of the WebSphere server. 

**<Requests>**
_-trustedSensitiveHeaderOrigin_
"trustedSensitiveHeaderOrigin" property could be configured with a IP segment corresponding to those of any WebSphere-aware proxy servers in front of the WebSphere server. 

We'll use the wildcard syntax that's used by the [TCP channel include and exclide lists](https://www.ibm.com/support/knowledgecenter/SSEQTP_8.5.5/com.ibm.websphere.base.doc/ae/urun_chain_typetcp.html)
ex.) `trustedSensitiveHeaderOrigin="192.168.10.*"`

These IP segments should support IPv6 syntax was well.
ex.) `trustedSensitiveHeaderOrigin="0:0:0:0:0:0:*:*"`

**<Reason for Request>**
Maintaining a list of all valid IP addresses without having the ability to use an IP segment is a large cost to some users.

This property should additionally support hostnames, since some customers would prefer not to configure IPs.  Those hostnames will also allow leading wildcards, eg. `*.ibm.com`.

The improvements above should also be made for the _trustedHeaderOrigin_ property.

RFE Link: https://www.ibm.com/developerworks/rfe/execute?use_case=viewChangeRequest&CR_ID=134886

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## List of Steps to complete or get approvals / sign-offs for Onboarding to the Liberty release (GM date)
 
Instructions:
 - Do the actions below and mark them complete in the checklist when they are done.
 - Make sure all feature readiness approvers put the appropriate tag on the epic to indicate their approval.
 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## TARGET COMPLETION DATE **Before Development Starts or 8 weeks before Onboarding** 
- [x] POC Design / WAD Review Scheduled (David Chang) or N/A.
- [x] POC Design / WAD Reviewed (Feature Owner) or N/A.
[...]

PR #15601, opened 2021-01-20
trusted header improvements: rework IP address validity check

wtlucy:
for #9809 

The `FilterList` class was borrowed from the TCP channel, and it relies on exception handling to validate address validity. To avoid unnecessary FFDCs in the logs, I'd rather avoid the exception checking.


mrsaldana (approved):
Changes look good, thanks!

wtlucy:
#build

wtlucy:
#run-libby-bot

PR #15630, opened 2021-01-22
CL22175 clarify getNextFixedRateExecutionTime

njr-11:
One of the parameters to getNextFixedRateExecutionTime and its description are misleading.  This pull corrects this to clarify that it is provided with the nanosecond timestamp of the expected execution time of the task.  This method is not externalized to end users, but we should make its description and naming accurate in order to ensure proper usage internally.


njr-11:
#build

njr-11:
#build

mswatosh (approved):

njr-11:
Build has 11 failures which are preexisting issues in 3 unrelated test buckets.

PR #15631, opened 2021-01-22
Continue iterating if the ejbDescripter fetches the wrong bean

### LINKED ISSUE #15628: CDIEJBManagedObjectFactory can get the wrong EJBDescriptor depending on BDA ordering
CDIEJBManagedObjectFactory iterates through BDAs looking for a BDA that can return a match for EJB names. However it's possible for multiple bdas to contain the same EJB name so we need to check we get the right bean before ending the search.

benjamin-confino:
fixes #15628

### DIFF
--- a/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
+++ b/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package com.ibm.ws.cdi.impl.managedobject;
 
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Iterator;
 import java.util.Set;
 
@@ -52,9 +54,9 @@ public class CDIEJBManagedObjectFactoryImpl<T> extends AbstractManagedObjectFact
 
     @Override
     protected synchronized WeldManager getBeanManager() {
-        if (this.ejbBeanManager == null) {
+        if (!ejbBeanLookupComplete) {
             //getEjbDescriptor will initialize the bean manager with the one which really contains the ejbDescriptor
-            getEjbDescriptor();
+            doInitialization();
         }
         return this.ejbBeanManager;
     }
@@ -89,46 +91,23 @@ public class CDIEJBManagedObjectFactoryImpl<T> extends AbstractManagedObjectFact
         return managedObjectContext;
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     protected synchronized Bean<T> getBean() {
-        if (!this.ejbBeanLookupComplete) {
-
-            WeldManager beanManager = getBeanManager();
-
-            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                Tr.debug(tc, "Looking for EJB Bean: " + this.ejbName);
-            }
-
-            EjbDescriptor<?> ejbDescriptor = getEjbDescriptor();
-            if (ejbDescriptor != null) {
-                this.ejbBean = (Bean<T>) beanManager.getBean(ejbDescriptor);
-                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                    if (this.ejbBean != null) {
-                        Tr.debug(tc, "Found EJB Bean: " + this.ejbBean);
-                    }
-                }
-            }
-
-            this.ejbBeanLookupComplete = true;
-
-            if (this.ejbBean != null) {
-                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                    Tr.debug(tc, "Found a bean of class : " + this.ejbBean.getBeanClass());
-                }
-                if (!this.ejbBean.getBeanClass().equals(getManagedObjectClass())) {
-                    //this exception should never happen
-                    throw new IllegalStateException("Managed Class {" + getManagedObjectClass().getName() + "} does not match Bean Class {"
-                                                    + this.ejbBean.getBeanClass().getName() + "}");
-                }
-            }
-
+        if (!ejbBeanLookupComplete) {
+           doInitialization();
         }
         return this.ejbBean;
     }
 
-    private synchronized EjbDescriptor<T> getEjbDescriptor() {
-        if (this.ejbDescriptor == null) {
+    private synchronized void doInitialization() {
+        if (!ejbBeanLookupComplete) {
+            Bean<T> bean = null;
+
+            List<EjbDescriptor<T>> partialMatchDescriptors = new LinkedList<EjbDescriptor<T>>();
+
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "Looking for EJB Bean: " + this.ejbName);
+            }
 
             if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
                 Tr.debug(tc, "Looking for EJB Bean: " + this.ejbName);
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -89,28 +91,86 @@ public ManagedObjectContext createContext() {
         return managedObjectContext;
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings("unchecked")	
     @Override
     protected synchronized Bean<T> getBean() {
-        if (!this.ejbBeanLookupComplete) {
+        if (!ejbBeanLookupComplete) {
+
```
Azquelt, line 98: The logic in this method is now effectively initialization for the whole class which sets `ejbDescriptor`, `ejbBean`, `ejbBDA` and `ejbBeanManager`, so I think it would make sense to
1. Move the initialization code out to its own method, so that all your getters look the same
2. Refactor the initialization code so that those four fields are all set at once at the end of the method, ensuring that you don't get a half-inititialized object if an exception is thrown part way through.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -89,28 +91,86 @@ public ManagedObjectContext createContext() {
         return managedObjectContext;
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings("unchecked")
```
Azquelt, line 94: This suppression annotation isn't needed, and there's trailing whitespace suggesting the code hasn't been auto-formatted.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -129,53 +189,24 @@ public ManagedObjectContext createContext() {
...
         }
+        return this.ejbDescriptor;
+    }
 
-        if (this.ejbDescriptor == null) {
-            Tr.error(tc, "Could not find an EjbDescriptor for : " + this.ejbName);
+    private boolean getBeanFromDescriptor(EjbDescriptor<T> descriptor, WeldManager beanManager)
+    {
+        if (ejbDescriptor != null) {
+            Bean bean = (Bean<T>) beanManager.getBean(ejbDescriptor);
```
Azquelt, line 200: There's no need for the cast here, and you can use `Bean<T>` instead of `Bean`.
```suggestion
            Bean<T> bean = beanManager.getBean(ejbDescriptor);
```

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -89,28 +91,86 @@ public ManagedObjectContext createContext() {
...
                 }
+
             }
 
-            this.ejbBeanLookupComplete = true;
+            if (this.ejbDescriptor == null && partialMatchDescriptors.isEmpty()) {
+                Tr.error(tc, "Could not find an EjbDescriptor for : " + this.ejbName);
+            } else if (this.ejbDescriptor == null && ! partialMatchDescriptors.isEmpty()) {
+                String listStr = "[";
+                for (EjbDescriptor desc : partialMatchDescriptors) {
```
Azquelt, line 165:
```suggestion
                for (EjbDescriptor<T> desc : partialMatchDescriptors) {
```

Azquelt (changes requested):

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -89,28 +91,99 @@ public ManagedObjectContext createContext() {
...
+                listStr = listStr.substring(0, listStr.length() - 1);
+                listStr += "]";
+                Tr.error(tc, "Found the following EjbDescriptors but they produced the wrong bean " + listStr);
             }
 
-            this.ejbBeanLookupComplete = true;
+            ejbBeanLookupComplete = true;
 
             if (this.ejbBean != null) {
                 if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
```
Azquelt, line 189: Should this check not happen before we set `ejbBeanLookupComplete = true`? (and before we initialize the rest of the fields)

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -89,28 +91,99 @@ public ManagedObjectContext createContext() {
...
             WeldManager beanManager = getBeanManager();
 
             if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
                 Tr.debug(tc, "Looking for EJB Bean: " + this.ejbName);
             }
 
-            EjbDescriptor<?> ejbDescriptor = getEjbDescriptor();
-            if (ejbDescriptor != null) {
-                this.ejbBean = (Bean<T>) beanManager.getBean(ejbDescriptor);
+            if (this.ejbDescriptor == null) {
```
Azquelt, line 113: I think this `if` shouldn't be needed any more, this method should only be called if initialization hasn't been done yet and so `ejbDescriptor` will never be set at this point.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -89,28 +91,99 @@ public ManagedObjectContext createContext() {
...
+        }
+        return this.ejbBean;
+    }
+
+    private synchronized void doInitialization() {
+        if (!ejbBeanLookupComplete) {
+
+            List<EjbDescriptor<T>> partialMatchDescriptors = new LinkedList<EjbDescriptor<T>>();
 
             WeldManager beanManager = getBeanManager();
```
Azquelt, line 107: `beanManager` is set here, but will always be set to `null` on line 119.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -89,28 +91,99 @@ public ManagedObjectContext createContext() {
...
+                if (! bean.getBeanClass().equals(getManagedObjectClass())) {
+                    ejbDescriptor = null;
+                    bean = null;
+                    Set<WebSphereBeanDeploymentArchive> children = bda.getDescendantBdas();
+                    Iterator<WebSphereBeanDeploymentArchive> itr = children.iterator();
+                    while (ejbDescriptor == null && itr.hasNext()) {
+                        bda = itr.next();
+                        beanManager = (WeldManager) bda.getBeanManager();
+                        ejbDescriptor = beanManager.getEjbDescriptor(this.ejbName);
+                        bean = beanManager.getBean(ejbDescriptor);
```
Azquelt, line 139: Do we need to check `ejbDescriptor` for `null` before calling `getBean()`?

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
benjamin-confino: Good question. In the old version three of the four fields could be set from getEJBDescriptor without ever calling getBean

Do we want to preserve that old behaviour? 

But then we could set this.ejbBean to a bean, throw an exception because `this.ejbBean.getBeanClass().equals(getManagedObjectClass()` then the next time you call get bean ejbBeanLookupComplete is true so you get the bean with no exception which is _horrific_

So I think we do it the sane way, at least until we run some builds and see if anything breaks.

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -165,19 +186,32 @@ private synchronized void doInitialization() {
...
+                //The bean can be null if it's Vetoed. 
+                Tr.error(tc, "Found the following EjbDescriptor " + this.ejbDescriptor + " for ejbName " + this.ejbName + " but it did not produce a bean or produced the wrong bean");
+                if (! partialMatchDescriptors.isEmpty()) {
+                    String listStr = "[";
+                    for (EjbDescriptor<T> desc : partialMatchDescriptors) {
+                        listStr += desc.getEjbName() + ",";
+                    }
+                    listStr = listStr.substring(0, listStr.length() - 1);
+                    listStr += "]";
+                    Tr.error(tc, "Found the following EjbDescriptors but they produced the wrong bean " + listStr);
```
Azquelt, line 213: * If there are cases where this is expected, these log messages shouldn't be errors. Maybe a warning at most if it's usually wrong, but more likely an info message.
* Error/warning/info messages need to be translated
* It's a bit odd to have both `Tr.error` messages here. From the code I can see how they're saying different things but just reading the messages it's less clear. Maybe there's nothing we can do about this.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/managedobject/CDIEJBManagedObjectFactoryImpl.java
```diff
@@ -189,9 +223,6 @@ private synchronized void doInitialization() {
 	        } else {
```
Azquelt, line 223: This `else` is wrongly indented (it has a tab at the start of the line instead of eight spaces)

Azquelt (approved):

Emily-Jiang (approved):

benjamin-confino:
#build

benjamin-confino:
#build

Emily-Jiang (approved):
LGTM!

benjamin-confino:
#run-libby-bot

PR #15673, opened 2021-01-26
mpOpenAPI-2.0 Informative CWWKO1661E Error Message 

### LINKED ISSUE #15650: mpOpenAPI-2.0 - Informative Error Messages
Currently in some scenarios,`mpOpenAPI-2.0` does not give error messages which offer enough information to successfully debug without resorting to enabling trace strings. This has been documented within the following [serviceability investigation](https://github.com/OpenLiberty/open-liberty/issues/11020#issuecomment-765541767) under the **Common Problem Path - Filtering the OpenAPI tree elements** section.

In common cases of `mpOpenAPI-2.0`, the `CWWKO1661E` message is output. This simply states the OpenAPI document could not be generated, without reference to the cause of the error. If a developer was unable to work out what had caused the issue themselves, the Knowledge Center entry for `CWWKO1661E` instructs the user to '**gather logs and contact IBM service'**. In the example provided in the referenced issue for which the fix is simple, elevating the issue to a PMR should not be necessary and only detracts from the developer experience with `mpOpenAPI-2.0`.

I believe a potential solution going forward would be to have `CWWKO1661E` reading something like the following, giving information about the exception which has occurred, rather than just the application name:

**Current:**
```
[ERROR   ] CWWKO1661E: An error occured when processing application <application name> and an OpenAPI document was not produced.
```
**Proposed:**
```
[ERROR   ] CWWKO1661E: An error occured when processing application <application name> and an OpenAPI document was not produced. The error is <stack-trace>
```
This should at least help the user resolve failures related to common use cases and improve the overall developer experience of `mpOpenAPI-2.0`.

Also, **occured** should be changed to **occurred** within `CWWKO1661E`, both in the fix and the KC entry.

scottcurtis2605:
resolves #15650

- Updated `CWWKO1661E` within `mpOpenAPI-2.0` application processing code to include information about the cause of the error, rather than just 'there was an error'.
- Corrected typo in `CWWKO1661E`, from "occured" to "occurred"
- Updated `CWWKO1661E` user action.

#build
#spawn.fullfat.buckets=com.ibm.ws.microprofile.openapi_fat,com.ibm.ws.microprofile.openapi_fat_tck,com.ibm.ws.microprofile.openapi.1.1_fat_tck,io.openliberty.microprofile.openapi.2.0.internal_fat,io.openliberty.microprofile.openapi.2.0.internal_fat_tck

### DIFF
--- a/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationProcessor.java
+++ b/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationProcessor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020 IBM Corporation and others.
+ * Copyright (c) 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


Azquelt (approved):

skasund:
L2 message review completed

PR #15676, opened 2021-01-26
Try 2GB max heap size for framework and servers

andymc12:
A 4GB max heap size is too much for some of our build machines (VMs?) - resulting in errors like:
```
JVMJ9VM015W Initialization error for library j9gc29(2): Failed to instantiate heap; 4G requested
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

This PR reduces the max heap size to 2GB which hopefully is closer to the sweet spot of "enough heap to run the tests" vs "too much heap for our test VMs".


andymc12:
#build

jkoehler22 (approved):

Issue #13152, opened 2020-07-23
Rename com.ibm.websphere.appserver.thirdparty.eclipselink.3.0


tevans78:
`com.ibm.websphere.appserver.thirdparty.eclipselink.3.0` should have been named `io.openliberty…`

My preference would be `io.openliberty.jpa.3.0.thirdparty`. I prefer that pattern since it lets you update the eclipselink version internally (for upstream bug fixes etc) without having to rename projects and bundles.

jgrassel:
Delivered with https://github.com/OpenLiberty/open-liberty/pull/15683

PR #15683, opened 2021-01-26
OLGH13152: Rename com.ibm.websphere.appserver.thirdparty.eclipselink.…

jgrassel:
For #13152

com.ibm.websphere.appserver.thirdparty.eclipselink.3.0 should have been named io.openliberty…

My preference would be io.openliberty.jpa.3.0.thirdparty. I prefer that pattern since it lets you update the eclipselink version internally (for upstream bug fixes etc) without having to rename projects and bundles.


jgrassel:
#build

dazey3 (approved):
LGTM

PR #15686, opened 2021-01-26
Allow FT warning if app stops before timeout

andymc12:
If the app stops before a FT timer fires, it may print a warning in the server logs.  This is ok, and the test has already passed by this point.


andymc12:
#build

jkoehler22 (approved):

PR #15701, opened 2021-01-27
refactor useRemoteDocker logic

### LINKED ISSUE #15575: Attempt to run remote testcontainers
Improve how we determine if we are running against a local docker instance or if we should try to run against a remote docker instance. Priority list:

1. fat.test.use.remote.docker=true : Force remote
2. Github Actions build: Force local
3. RTC build: Force remote
4. Local Build / No docker instance: Force remote -> fail if not on IBM network
5. Default path (local build / docker): local

KyleAure:
Fixes #15575

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
+++ b/dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
@@ -352,7 +352,7 @@ public class ExternalTestServiceDockerClientStrategy extends DockerClientProvide
             // Assume a user on windows with no other preferences will want to use a remote host.
             if (System.getProperty("os.name", "unknown").toLowerCase().contains("windows")) {
                 result = true;
-                reason = "Local operating system is Windows. Default container support not guarenteed.";
+                reason = "Local operating system is Windows. Default container support not guaranteed.";
                 break;
             }
 
--- a/dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
+++ b/dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
@@ -53,12 +53,18 @@ public class ExternalTestServiceDockerClientStrategy extends DockerClientProvide
      * Used to specify a particular docker host machine to run with. For example: -Dfat.test.docker.host=some-docker-host.mycompany.com
      */
     private static final String USE_DOCKER_HOST = System.getProperty("fat.test.docker.host");
-    private static final boolean USE_REMOTE_DOCKER = Boolean.getBoolean("fat.test.use.remote.docker") || USE_DOCKER_HOST != null;
 
     private DefaultDockerClientConfig config;
     private TransportConfig transportConfig;
 
     private static boolean setupComplete = false;
+
+    /**
+     * Used to specify if we plan on running against a remote docker host, or a local docker host.
+     *
+     * @see #useRemoteDocker()
+     */
+    public static final boolean USE_REMOTE_DOCKER_HOST = useRemoteDocker();
 
     /**
      * <pre>
@@ -115,7 +121,7 @@ public class ExternalTestServiceDockerClientStrategy extends DockerClientProvide
      */
     private static void generateDockerConfig() {
         final String m = "generateDockerConfig";
-        if (!useRemoteDocker())
+        if (!USE_REMOTE_DOCKER_HOST)
             return;
 
         File configDir = new File(System.getProperty("user.home"), ".docker");
@@ -288,7 +294,7 @@ public class ExternalTestServiceDockerClientStrategy extends DockerClientProvide
 
     @Override
     protected int getPriority() {
-        return useRemoteDocker() ? 900 : 0;
+        return USE_REMOTE_DOCKER_HOST ? 900 : 0;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
```diff
@@ -302,15 +308,65 @@ protected boolean isApplicable() {
...
+
+            //State 2: User provided a remote docker host, assume they want to use the remote host
+            if (USE_DOCKER_HOST != null) {
+                result = true;
+                reason = "fat.test.docker.host set to " + USE_DOCKER_HOST;
+                break;
+            }
 
+            //State 3: Github actions build should always use local
+            if (Boolean.parseBoolean(System.getenv("GITHUB_ACTIONS"))) {
```
gjwatts, line 345: If Github actions should always be built local, shouldn't this be the first thing checked?  Or is there a possible scenario where you would want to use a remote docker with Github Actions?

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
```diff
@@ -302,15 +308,65 @@ protected boolean isApplicable() {
...
+                result = false;
+                reason = "GitHub Actions Build";
+                break;
+            }
+
+            //State 4: Earlier version of TestContainers didn't support docker for windows
+            // Assume a user on windows with no other preferences will want to use a remote host.
+            if (System.getProperty("os.name", "unknown").toLowerCase().contains("windows")) {
+                result = true;
+                reason = "Local operating system is Windows. Default container support not guarenteed.";
```
gjwatts, line 355: typo: guaranteed

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
```diff
@@ -302,15 +308,65 @@ protected boolean isApplicable() {
...
+            if (System.getProperty("fat.test.use.remote.docker") != null) {
+                result = Boolean.getBoolean("fat.test.use.remote.docker");
+                reason = "fat.test.use.remote.docker set to " + result;
+                break;
+            }
+
+            //State 2: User provided a remote docker host, assume they want to use the remote host
+            if (USE_DOCKER_HOST != null) {
+                result = true;
+                reason = "fat.test.docker.host set to " + USE_DOCKER_HOST;
```
gjwatts, line 340: Can "fat.test.docker.host" be set to the localhost?  Or is it pretty safe that a value in this field will only be the name of a remote host?

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
```diff
@@ -302,15 +308,65 @@ protected boolean isApplicable() {
...
+    private static boolean useRemoteDocker() {
+        boolean result;
+        String reason;
+
+        do {
+            //State 1: fat.test.use.remote.docker should always be honored first
+            if (System.getProperty("fat.test.use.remote.docker") != null) {
+                result = Boolean.getBoolean("fat.test.use.remote.docker");
+                reason = "fat.test.use.remote.docker set to " + result;
+                break;
```
gjwatts, line 334: What if a "fat.test.use.remote.docker" is true, but "fat.test.docker.host" is blank/null?

gjwatts (changes requested):

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
KyleAure: Yes and no.  In my opinion if someone sets a property explicitly, then they should expect it to be honored.  So if this came first, we would not honor fat.test.use.remote.docker or fat.test.docker.host in some cases. 

As a note both of these properties are user provided, not something we calculate at build time.

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
KyleAure: Fixed.

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
KyleAure: It could be set to localhost and the following would happen:
1. This would return true
2. Then the ExternalTestServiceDockerClientStategy (this class) would get used first as it has priority 900
3. None of the remote docker hosts set in user.build.properties would match "localhost"
4. We would try the next docker strategy with a priority
5. Either find a docker host locally and run, or fail to find a local docker host. 

But this value is only set by a user or a build system so it's pretty safe that we won't be doing the above process all the time.

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
gjwatts: Ok, I thought you were determining if it was a GHA from the environment somehow.  Thanks for explaining that to me.

dev/fattest.simplicity/src/componenttest/containers/ExternalTestServiceDockerClientStrategy.java
KyleAure: fat.test.use.remote.docker will make the priority of ExternalTestServiceDockerClientStategy (this class) high (900) so it will be the first strategy that docker-java (the underlying API) will use to try to connect to a docker host. 

This class uses user.build.properties on the machines home directory to get a list of possible hosts to try.  If fat.test.docker.host is set, it will ignore all but the host that matches that one. 

fat.test.docker.host is really meant so that if a certain docker host is flaky we can explicitly avoid or test against it.

gjwatts (approved):

KyleAure:
#build

KyleAure:
#build #libbybot

PR #14248, opened 2020-09-30
Update WC property suppressHtmlRecursiveErrorOutput

pmd1nh:
https://github.com/OpenLiberty/open-liberty/issues/14242

PR #15725, opened 2021-01-29
Revert "Update WC property suppressHtmlRecursiveErrorOutput"

Azquelt:
Reverts OpenLiberty/open-liberty#14248

The error in the personal build is complaining that `suppressHtmlRecursiveErrorOutput` has been removed. It looks like you have removed that attribute from the metatype in this PR, so I think this PR needs to be reverted.


Joseph-Cass (approved):
#14248 does seem to be the PR which is causing Defect 281378

PR #15729, opened 2021-01-29
Keep the order of subsystem content

tjwatson:
ManifestHeaderProcessor.parseImportString does
not maintain the order specified in the header
in the returned map.  This change first
tries to use ManifestElement, which preserves
order and falls back to ManifestHeaderProcessor
on exception to try and maintain order


tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

tjwatson:
This is to evaluate the impact of maintaining consistent order for design issue #15734

tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

dazavala (approved):
Hello @tjwatson .  The changes involving `ManifestHeaderProcessor.parseExportString()` look correct -- that is, the changes preserve the order of the feature content and ignore any duplicate entries possibly returned by parseExportString().  I did not read the source for `parseImportString()` to see if it handles duplicates in the same way.  The openidConnectServer-1.0 feature content looks correct, though I'm not sure about the order.  If you restored the content order to a previous version of the file, then fyi, I couldn't find a recent version with the exact same order.  httpCommons-1.0 always appears before servlet-3.0.  I approved, since others will address the content order if it's an issue.

tjwatson:
> Hello @tjwatson . The changes involving `ManifestHeaderProcessor.parseExportString()` look correct -- that is, the changes preserve the order of the feature content and ignore any duplicate entries possibly returned by parseExportString(). I did not read the source for `parseImportString()` to see if it handles duplicates in the same way.

That is a good observation!  The parseImportString will have the latest dup specified win where the current logic using parseExportString will have the first one found win.  I'm not really worried about such a change here.  A feature that has dups like that is in error and should be fixed IMO.  Otherwise we need a bit more code to go remove the existing resource from the list and replace it with the later dup.  But then should that dup replace the slot the previous one was or go at the end.  Such questions for something that I think is a feature error to begin with, not something I think we should worry about doing now.

> The openidConnectServer-1.0 feature content looks correct, though I'm not sure about the order. If you restored the content order to a previous version of the file, then fyi, I couldn't find a recent version with the exact same order. httpCommons-1.0 always appears before servlet-3.0. I approved, since others will address the content order if it's an issue.

I updated it to the hash order that I found using Java 11 on my system which made the component work in the feature.  It will not be the order of the previous version.  It turns out the tool to calculate the hashing order of the features is not 100% accurate because it did not consider the other content types.  Instead of revisiting all the features again I decided to fix up this feature that has a very brittle order dependency.  Ultimately this brittle order dependency needs to be fixed because I'm sure it will come to bite us again in the future.

tjwatson:
#libby

PR #15742, opened 2021-01-30
Properly handle exceptions caught by Default Exception Mapper in mpMetrics-3.0.

### LINKED ISSUE #15715: Exceptions lost from CDI Interceptors in Metrics 3.0 due to the use of default exception mapper.
In 3.0 we're using the `DefaultExceptionMapper` mechanism from JAXRS to allow us to intercept any unmapped exceptions which allows use to create and update the new REST metric. However, this seems to have the consequence of also capturing any exceptions thrown by any CDI interceptors (and/or any exceptions thrown by any code called by the CDI interceptors) as it is considered part of the application.

Attempting to rethrow the exception where we catch the exception will also be caught by the CXF code. It'll print out a stack trace but the exception/throwable is _lost_.

Channyboy:
Metrics Exception Mapper callback is to print the stack trace from exceptions caught by the JAXRS Default Exception Mapper.
In beta they use to just be consumed and were "lost". This is a bug. 

~~Current approach will build it into a single string and have it be print out to the System Error stream. This is chosen over the alternative which is to just `printStackTrace()` directly to the `System.err` stream. This causes each line of the stack trace to be an individual log record. This clutters the `message.log` (i.e one message per stack element). In an intense multi-thread environment the stack trace may be interleaved with other messages. The draw back of combining into a singular string before writing to `System.err` is that internal classes are not _combined_ with `at [internal classes]` in the console.~~

The intention in the future is to log the caught exception as a `WARNING`. This is similar to the mpOT approach.

Exceptions in Metrics CDI bundle to also be surrounded by try and catch to capture and generate an FFDC. The exceptions are re-thrown in the catch block to be _caught_ again by the Default Exception Mapper where the exception is then printed out.
#build 
fix #15715

### DIFF
--- a/dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
+++ b/dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020 IBM Corporation and others.
+ * Copyright (c) 2020, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -10,11 +10,15 @@
  *******************************************************************************/
 package io.openliberty.microprofile.metrics.internal.monitor;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
 import com.ibm.ws.jaxrs.defaultexceptionmapper.DefaultExceptionMapperCallback;
 import com.ibm.ws.microprofile.metrics.impl.SharedMetricRegistries;
 import com.ibm.ws.runtime.metadata.ComponentMetaData;
 import com.ibm.ws.threadContext.ComponentMetaDataAccessorImpl;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.lang.reflect.Method;
 import java.util.AbstractMap;
 import java.util.Collections;
@@ -37,9 +41,9 @@ import org.osgi.service.component.annotations.ConfigurationPolicy;
 				"service.vendor=IBM" })
 public class MetricsJaxRsEMCallbackImpl  implements DefaultExceptionMapperCallback {
 
-	
+    private static final TraceComponent tc = Tr.register(MetricsJaxRsEMCallbackImpl.class);
 	public static final String EXCEPTION_KEY = MetricsJaxRsEMCallbackImpl.class.getName() + ".Exception";
-	
+	private static final String[] metriCDIBundles = {"io.astefanutti.metrics.cdi30", "io.openliberty.microprofile.metrics.internal.cdi30.interceptors"};
 	
 	public synchronized static Counter registerOrRetrieveRESTUnmappedExceptionMetric(String fullyQualifiedClassName, String methodSignature) {
 		MetricRegistry baseMetricRegistry = sharedMetricRegistry.getOrCreate(MetricRegistry.Type.BASE.getName());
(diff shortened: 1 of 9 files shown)


dev/io.openliberty.microprofile.metrics.internal.cdi.3.0/src/io/astefanutti/metrics/cdi30/MetricName.java
```diff
@@ -104,8 +114,24 @@ private String getParameterName(AnnotatedParameter<?> parameter) {
...
+             * which does not result in FFDCs. Caught Exception generates an FFDC. The subsequent
+             * rethrow is caught by the DefaultExceptionMapper where it is printed out.
+             */
+            try {
+                throw new UnsupportedOperationException("Unable to retrieve name for parameter [" + parameter
+                                                        + "], @Metric annotation on injected parameter is required before Java 8");
+            } catch (UnsupportedOperationException exception) {
+                throw exception;
+            }
+        } catch (UnsupportedOperationException exception) {
```
Channyboy, line 128: This catches the one thrown at line 114

dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
```diff
@@ -37,9 +41,9 @@
 				"service.vendor=IBM" })
 public class MetricsJaxRsEMCallbackImpl  implements DefaultExceptionMapperCallback {
 
-	
+    private static final TraceComponent tc = Tr.register(MetricsJaxRsEMCallbackImpl.class);
```
donbourne, line 44: Does this get used anywhere?

dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
```diff
@@ -37,9 +41,9 @@
...
 	public static final String EXCEPTION_KEY = MetricsJaxRsEMCallbackImpl.class.getName() + ".Exception";
-	
+	private static final String[] metriCDIBundles = {"io.astefanutti.metrics.cdi30", "io.openliberty.microprofile.metrics.internal.cdi30.interceptors"};
```
donbourne, line 46: metriCDIBundles -> metricCDIBundles

dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
```diff
@@ -57,14 +61,39 @@ public synchronized static Counter registerOrRetrieveRESTUnmappedExceptionMetric
...
+
 	@Override
-	public Map<String, Object> onDefaultMappedException(Throwable t, int statusCode, ResourceInfo resourceInfo) {
-		Map.Entry<String, String> classXmethod = resolveSimpleTimerClassMethodTags(resourceInfo);
+	public Map<String, Object> onDefaultMappedException(Throwable throwable, int statusCode, ResourceInfo resourceInfo) {
+
+		StackTraceElement[] ste = throwable.getStackTrace();
+
+		/*
+		 * If the Exception originates from from the Metrics CDI Bundle we do not want to count
```
donbourne, line 71: from from -> from

dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
```diff
@@ -57,14 +61,39 @@ public synchronized static Counter registerOrRetrieveRESTUnmappedExceptionMetric
...
+		 * throwable.printStackTrace() which will stream
+		 * every stack element/line as a logging event making
+		 * messages.log cluttered. The trade-off being that
+		 * internal classes are not hidden in the console
+		 * output.
+		 */
+		StringWriter sw = new StringWriter();
+		PrintWriter pw = new PrintWriter(sw);
+		throwable.printStackTrace(pw);
+		System.err.print(sw.toString());
```
donbourne, line 94: suggest you just use throwable.printStackTrace -- we have another issue coming soon which will handle merging stack traces into a single event

dev/io.openliberty.microprofile.metrics.internal.cdi.3.0/src/io/astefanutti/metrics/cdi30/MetricResolver.java
```diff
@@ -359,8 +379,16 @@ else if (SimplyTimed.class.isInstance(annotation))
...
         if ((existingMetadata != null && !existingMetadata.equals(of.metadata()))) {
-            throw new IllegalArgumentException("Cannot reuse metric with MetricID " + metricID.toString() + "with Metadata " + of.metadata().toString()
-                                               + ". There already exists a Metadata fopr tthis metric name with different values: " + existingMetadata.toString());
+            /*
+             * Try and Catch to force an FFDC since the DefaultExceptionMapper is catching exceptions
+             * which does not result in FFDCs.
+             */
+            try {
+                throw new IllegalArgumentException("Cannot reuse metric with MetricID " + metricID.toString() + "with Metadata " + of.metadata().toString()
+                                                   + ". There already exists a Metadata fopr tthis metric name with different values: " + existingMetadata.toString());
```
donbourne, line 388: fopr tthis -> for this

donbourne (changes requested):

Channyboy:
#build
#run-libby-bot

dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
Channyboy: Ah nope. That was left over when it originally was printing an warning.

donbourne (approved):

PR #15778, opened 2021-02-02
update cxf security prereqs to 341 to match with the core api level

arunavemulapalli:
update security , security.saml, ws-security and ws-mex levels.


arunavemulapalli:
#run-libby-bot

arunavemulapalli:
#build

dmuelle:
looks like the nls changes are filename only. Adding the ID review not needed label- png me if that's wrong and I'll review. Thanks

arunavemulapalli:
#build

neuwerk (approved):
Changes look good to me! Going forward we should consider removing the micro versions from the project/bundle names so we don't have to up-tick them every time we pull in a new minor version of CXF.

arunavemulapalli:
#run-libby-bot

PR #15798, opened 2021-02-03
CL22321 add callback for policy executor shutdown

njr-11:
Add a shutdown callback to the policy executor to enable services that build on top of it which require additional shutdown logic.


gjwatts (approved):

njr-11:
#build

njr-11:
Build has 11 test failures across 4 test buckets, which are preexisting issues.

PR #15808, opened 2021-02-04
Design for SameSite=None known incompatible clients

mrsaldana:
This change takes into consideration the known incompatible clients in order to remove the SameSite value when set to None when applicable. This is modeled after the following Chromium resource: https://www.chromium.org/updates/same-site/incompatible-clients


dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/internal/HttpBaseMessageImpl.java
```diff
@@ -2905,6 +2906,14 @@ else if (getServiceContext().getHttpConfig().onlySameSiteStar()) {
                 }
             }
 
+            if (cookie.getAttribute("samesite") != null && cookie.getAttribute("samesite").equals(HttpConfigConstants.SameSite.NONE.getName())) {
+                String userAgent = getServiceContext().getRequest().getHeader(HttpHeaderKeys.HDR_USER_AGENT).asString();
+                if (SameSiteCookieUtils.isSameSiteNoneIncompatible(userAgent)) {
+                    //TODO: do we remove Secure, probably should be retained.
```
pnicolucci, line 2912: At this point in time there is probably no way to know if Session/WC/Security etc added the Secure attribute or if it was already set correct?

dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/internal/cookies/SameSiteCookieUtils.java
```diff
@@ -0,0 +1,277 @@
...
+    /**
+     * Compares the request's user agent against a regex of known incompatible
+     * clients for the SameSite=None cookie attribute. If the client is incompatible,
+     * this method will return true.
+     *
+     * @param userAgent
+     * @return
+     */
+    public static boolean isSameSiteNoneIncompatible(String userAgent) {
+        System.out.println("isSameSiteNoneIncompatible: enter");
```
pnicolucci, line 68: I assume we'll move this to an actual logger of some sort and not use System.out?

dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/internal/cookies/SameSiteCookieUtils.java
```diff
@@ -0,0 +1,277 @@
...
+        String normalizedUserAgent = userAgent.toLowerCase().trim();
```
pnicolucci, line 69: Is there any chance that the userAgent is null?

dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/internal/cookies/SameSiteCookieUtils.java
```diff
@@ -0,0 +1,277 @@
...
+     * of Chromium derived browsers, as well as Android WebView.
+     *
+     * Versions of UC browser on Android prior to version 12.13.2 reject cookies
+     * with SameSite=None.
+     *
+     * @param userAgent
+     * @return
+     */
+    private static boolean dropsUnrecognizedSameSiteAttribute(String userAgent) {
+        System.out.println("dropsUnrecognizedSameSiteAttribute: enter");
```
pnicolucci, line 166: System.out...

dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/internal/cookies/SameSiteCookieUtils.java
mrsaldana: It is checked for null by the calling logic up in HttpBaseMessageImpl.marshallCookies()

PR #15830, opened 2021-02-08
Add introspection of values stored in MP Config

### LINKED ISSUE #15804: Create a server dump introspector for MicroProfile Config
To do this, we need to provide a DS that implements `com.ibm.wsspi.logging.Introspector`. It may be that we want more than one version for the different MP Config implementations we have.

The introspector should dump out the contents of each Config instance in the classloader cache; the keys, values and the resolved Config Sources that provided them.

Azquelt:
Add an introspector to dump the values stored in MP Config when running `server dump`

Fixes #15804


dev/com.ibm.ws.microprofile.config.1.1_fat/fat/src/com/ibm/ws/microprofile/config/fat/suite/FATSuite.java
```diff
@@ -39,18 +29,18 @@
...
-                SimultaneousRequestsTest.class, //FULL
-                SharedLibTest.class, //FULL
-                StressTest.class, //FULL
-
-                // The following don't repeat against mpConfig > 1.4. See classes for why.
-                LibertySpecificConfigTests.class, //FULL
-                CDIBrokenInjectionTest.class, //FULL
-                ClassLoaderCacheTest.class, //FULL
-                DynamicSourcesTest.class //FULL
+//                ClassLoadersTest.class, //FULL
```
hutchig, line 32: Is this intentional or just there from testing?

dev/com.ibm.ws.microprofile.config.1.1_fat/fat/src/com/ibm/ws/microprofile/config/fat/tests/BasicConfigTests.java
```diff
@@ -118,4 +140,61 @@ public static void tearDown() throws Exception {
...
+     * @return map from app name to corresponding section of the introspection
+     */
+    private static Map<String, String> readIntrospectionSections(InputStream in) throws IOException {
+        Map<String, String> result = new HashMap<>();
+        BufferedReader reader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
+        String line = reader.readLine();
+        StringBuilder currentBuilder = new StringBuilder();
+        String currentApp = null;
+        while (line != null) {
+            Matcher m = Pattern.compile("Config for (.+)").matcher(line);
```
hutchig, line 182: can the compile go outside the loop if this is being run in every full fat run?

dev/io.openliberty.io.smallrye.config/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -20,31 +20,64 @@ org.eclipse.jdt.core.compiler.debug.sourceFile=generate
 org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
 org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
 org.eclipse.jdt.core.compiler.source=1.8
+org.eclipse.jdt.core.formatter.align_assignment_statements_on_columns=false
```
hutchig, line 23: did you intend to checkin changes to this file?

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
```diff
@@ -161,4 +164,28 @@ private static String getApplicationName() {
         return applicationName;
     }
 
+    @Override
+    public String getIntrospectorDescription() {
```
hutchig, line 168: Would it help to differentiate this description and name from the AbstractProviderResolver intropector (far)above?

dev/io.openliberty.microprofile.config.internal.common/src/io/openliberty/microprofile/config/internal/common/ConfigIntrospection.java
```diff
@@ -0,0 +1,74 @@
...
+            }
+        } catch (Exception e) {
+            pw.printf("Exception introspecting config: %s", e);
+        }
+    }
+
+    private static void introspectConfigSource(PrintWriter pw, ConfigSource source) {
+        try {
+            pw.printf(" - %s config source (ordinal %d)\n", source.getName(), source.getOrdinal());
+            for (Entry<String, String> entry : source.getProperties().entrySet()) {
```
hutchig, line 67: could getProperties() ever be null?

hutchig (changes requested):

hutchig:
Why was the new defect raised when https://ibm-cloud.slack.com/archives/C6EG3K3MK/p1612179980118900
?

dev/io.openliberty.io.smallrye.config/.settings/org.eclipse.jdt.core.prefs
Azquelt: Yes, I made changes to the formatter settings to allow it to wrap between chained calls (e.g. `a.getB().getC()`) and between attributes in an annotation.

Without this I was getting bad wrapping decisions like this one where the formatter was happy to break the line in the middle of the array value for the `service` attribute, but not between each attribute.
```
@Component(service = { ConfigProviderResolver.class,
                       ApplicationStateListener.class }, configurationPolicy = ConfigurationPolicy.IGNORE, property = { "service.vendor=IBM" }, immediate = true) 
```

The other changes in this file are a result of the newer version of eclipse adding values for options which didn't exist when the file was last written.

However, the changes to the other settings file look less good. It's added new templates for Java 9 features which includes things like `@author` tags which we don't use in OL.

dev/com.ibm.ws.microprofile.config.1.1_fat/fat/src/com/ibm/ws/microprofile/config/fat/tests/BasicConfigTests.java
Azquelt: Yes, will do.

dev/com.ibm.ws.microprofile.config.1.1_fat/fat/src/com/ibm/ws/microprofile/config/fat/suite/FATSuite.java
Azquelt: This was left in from testing by mistake.

dev/io.openliberty.io.smallrye.config/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -51,6 +51,8 @@ cleanup.use_this_for_non_static_method_access=false
 cleanup.use_this_for_non_static_method_access_only_if_necessary=true
 eclipse.preferences.version=1
 editor_save_participant_org.eclipse.jdt.ui.postsavelistener.cleanup=true
+formatter_profile=_config
```
hutchig, line 54: Did you want to change this?

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
Azquelt: I think it makes sense for both introspectors to have the same name since that decides the output filename in the dump. Both introspectors are essentially the same, just for mpConfig 1.x and 2.x and I think it makes more sense for the introspection output to be in the same place for both versions.

Potentially we could put which introspector was used in the description, though we couldn't easily include which specific version of the config feature is in use since the introspector doesn't know that information, so we could only put whether it was 1.x or 2.x.

dev/io.openliberty.microprofile.config.internal.common/src/io/openliberty/microprofile/config/internal/common/ConfigIntrospection.java
Azquelt: Hmm, it shouldn't be `null` but since it could be provided by the user and we're outputting diagnostic information it would make sense to check here before iterating over it so that we can put out a better message than would be given by the catch block below.

dev/io.openliberty.io.smallrye.config/.settings/org.eclipse.jdt.ui.prefs
Azquelt: No, it looks like this was a result of updating the formatter settings with a newer version of eclipse (that part was deliberate). 

I don't think the `formatter_profile` name is at all relevant when we're using project-specific formatting settings but I'll take this file out since the new templates which have been added include `@author` tags which we don't use in openliberty.

hutchig: Apologies I was just trying to draw attention to the file in general and not any specific line.

hutchig (approved):

Azquelt:
I wrote this as I thought it would be a fairly simple easy win to make diagnosing user mistakes with mpConfig easier, but concerns have since been raised around including sensitive config such as passwords (which can't be identified) and the risk of causing issues to a running server by dumping out a very large config source.

PR #15844, opened 2021-02-09
Update to use snapshot version of TCK with added timeout padding

andymc12:
This pulls in a snapshot build of the MP Rest Client 2.0 TCK that has some extra padding allowing us to work around JDK issues with time unit conversion rounding and inaccuracies with `System.nanoTime()`.


andymc12:
#build

jim-krueger (approved):

andymc12:
Updated timeout cushion to match previous value (as converted to ms) - will run new #build

andymc12:
#libby

PR #15850, opened 2021-02-09
Issue #15849: Increase timeout for createSpnAndKeytab SSH command.

### LINKED ISSUE #15849: Test Failure (20201118-0738): junit.framework.TestSuite.com.ibm.ws.security.spnego.fat.FATSuite
After making a fix in #15740, we still occasionally receive timeouts since in the previous fix I reduced the timeout to 60 seconds. During testing I saw this command take anywhere from 30-45 seconds, so I had set the timeout to 60 seconds, down from 300 seconds. I will increase the timeout to 120 seconds and that should resolve this timeout.

jvanhill:
fixes #15849


jvanhill:
#build
#libby

kristip17 (approved):

PR #15861, opened 2021-02-09
CL22321 policy executor shutdown by identifier match

njr-11:
Add a method to the policy executor provider to shutdown all matching policy executors based on the identifier prefix.


njr-11:
#build

njr-11:
The build has 1 test failure, which is a duplicate of a pre-existing issue in an unrelated test bucket.

mswatosh (approved):

PR #15877, opened 2021-02-10
Issue 15867: javamail fails under clientEE7 on windows, add debug

uberskigeek:
Update javamail client fat to add debug info when starting the greenmail mail server.


uberskigeek:
fixes #15877 15867

PR #15905, opened 2021-02-11
Update folder check

KyleAure:
```
if [ -f .github/test-categories/MODIFIED_* ]; then
```

worked on my local system and would return true/false depending on the existence of these files.

The github action's build instead returned a non failing error
```
.github/test-categories/MODIFIED_FULL_MODE_0: binary operator expected
```

Here the failure is due to the fact that bash substituted the GLOB with the actual file names meaning there was an extra argument since two files matched.

Instead, change it to a more shell agnostic check
```
if ls .github/test-categories/MODIFIED_* &> /dev/null ; then
```

### DIFF
--- a/.github/workflow-scripts/generate-params.sh
+++ b/.github/workflow-scripts/generate-params.sh
@@ -1,22 +1,29 @@
 #!/bin/bash
+
 # Save off PR Summary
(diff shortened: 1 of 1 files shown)


olendvcook (approved):

KyleAure:
#libbybot

PR #15912, opened 2021-02-12
Add jwtsso fat for new sig algs

c00crane:
Add tests using newly supported signature algorithms with jwtSso.
jwtSso will use the jwtBuilder and mpJwt and both of those fully test all supported signature algorithms.
We'll just test a subset of signature algorithms with jwtSso (just to show that there is no issue)
Newly supported algorithms (HS384, HS512, RS384, RS512, ES256, ES384, ES512) - adding tests for RS512 and ES384.


c00crane:
#build

ayoho (approved):

PR #15925, opened 2021-02-15
Prorotype integrating support for mpConfig-2.0 into MPRM 1.0 

Azquelt:
Prototype integrating a hypothetical future version of Smallrye MPRM
which has support for mpConfig-2.0.

Test changes are WIP and don't pass yet.

### DIFF
--- a/dev/cnf/oss_dependencies.maven
+++ b/dev/cnf/oss_dependencies.maven
@@ -125,6 +125,7 @@ io.smallrye.common:smallrye-common-function:1.5.0
 io.smallrye.config:smallrye-config-common:2.0.2
 io.smallrye.config:smallrye-config-core:2.0.2
 io.smallrye.config:smallrye-config:2.0.2
+io.smallrye.reactive:smallrye-reactive-messaging-provider:1.0.9-SNAPSHOT
 io.smallrye:smallrye-graphql-client-api:1.0.9
 io.smallrye:smallrye-graphql-client:1.0.9
 io.smallrye:smallrye-graphql-schema-model:1.0.9
--- a/dev/com.ibm.websphere.appserver.features/visibility/auto/io.openliberty.mpReactiveMessaging-mpConfig2.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/auto/io.openliberty.mpReactiveMessaging-mpConfig2.feature
@@ -1,0 +1,10 @@
+-include= ~${workspace}/cnf/resources/bnd/feature.props
+symbolicName=io.openliberty.mpReactiveMessaging-mpConfig2
+visibility=private
+IBM-Provision-Capability: osgi.identity; filter:="(&(type=osgi.subsystem.feature)(osgi.identity=com.ibm.websphere.appserver.mpReactiveMessaging-1.0))", \
+ osgi.identity; filter:="(&(type=osgi.subsystem.feature)(|(osgi.identity=com.ibm.websphere.appserver.mpConfig-2.0) \
+                                                          ))"
+-bundles=com.ibm.ws.io.smallrye.reactive.messaging-provider.config20
+IBM-Install-Policy: when-satisfied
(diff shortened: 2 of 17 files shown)


dev/cnf/oss_dependencies.maven
```diff
@@ -125,6 +125,7 @@ io.smallrye.common:smallrye-common-function:1.5.0
 io.smallrye.config:smallrye-config-common:2.0.2
 io.smallrye.config:smallrye-config-core:2.0.2
 io.smallrye.config:smallrye-config:2.0.2
+io.smallrye.reactive:smallrye-reactive-messaging-provider:1.0.9-SNAPSHOT
```
tevans78, line 128: Where is the other version? How come it wasn't listed to start with?

dev/com.ibm.websphere.appserver.features/visibility/auto/io.openliberty.mpReactiveMessaging-mpConfig2.feature
```diff
@@ -0,0 +1,10 @@
+-include= ~${workspace}/cnf/resources/bnd/feature.props
+symbolicName=io.openliberty.mpReactiveMessaging-mpConfig2
+visibility=private
+IBM-Provision-Capability: osgi.identity; filter:="(&(type=osgi.subsystem.feature)(osgi.identity=com.ibm.websphere.appserver.mpReactiveMessaging-1.0))", \
+ osgi.identity; filter:="(&(type=osgi.subsystem.feature)(|(osgi.identity=com.ibm.websphere.appserver.mpConfig-2.0) \
+                                                          ))"
+-bundles=com.ibm.ws.io.smallrye.reactive.messaging-provider.config20
+IBM-Install-Policy: when-satisfied
+kind=ga
```
tevans78, line 9: If this wasn't a draft, I would say set this to beta

dev/cnf/oss_dependencies.maven
Azquelt: it's in the `oss_source_dependencies.maven` file

dev/com.ibm.websphere.appserver.features/visibility/auto/io.openliberty.mpReactiveMessaging-mpConfig2.feature
Azquelt: Yeah, it should be.

Azquelt: Unfortunately, I don't think we can do that. mpRM-1.0 and mpConfig-2.0 are already available. If we make the changes to mpRM-1.0 to tolerate mpConfig-2.0, then we need this feature to be present, otherwise you have a server which starts but then the bundles won't resolve.

Azquelt: I think the only way we could beta this feature would be to have a separate feature which is identical to mpRM-1.0 except that it also tolerates mpConfig-2.0. To try out the functionality, users would have to enable the separate feature in place of mpRM-1.0.

tevans78: mpConfig-2.0 is still beta

Azquelt: ah, that's true, it should match mpConfig-2.0

Emily-Jiang: well, it might go to GA straightaway though, 21.0.0.3.

PR #15931, opened 2021-02-15
Reduce max heap size to 1GB for MP Rest Client 1.0/1.1 TCK FATs

andymc12:
- this is consistent with other MP Rest Client TCK FATs and avoids
issues on zOS test machines (VMs) that don't have much memory


andymc12:
#build

jkoehler22 (approved):

PR #15932, opened 2021-02-15
Add jwtsso fat for encryption (with tests disabled until runtime supports)

c00crane:
I'm adding a couple tests for jwtSso using encryption.
The runtime does not support this at the moment - I've created tests and will deliver them disabled.
The tests can be used during runtime development and can be enabled when the runtime code is delivered.
The jwtBuilder and mpJwt runtime are supporting JWE's, but, we still need some updates to the jwtSso runtime to make it work end to end...


c00crane:
#build

ayoho (approved):

PR #15940, opened 2021-02-15
Initial io.openliberty.netty with old WsByteBuffer impl and unit tests

hutchig:
A first version of the io.openliberty.netty feature containing just WsByteBuffer (OL's impl)


andrea-mh:
No L2 message review needed for translations

PR #15988, opened 2021-02-19
Ensure container can run startup script

KyleAure:
Sometimes when using a Generic Container we are getting a failure running the custom initDB.sql file:
```
/usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/initDB.sql
/docker-entrypoint-initdb.d/initDB.sql: Permission denied
```

This only happens on some systems and intermittently.  Being a docker environment I would expect this to work the same on all systems. 

I went in a changed the permission on the initDB.sql file to from 644 -> 755 to ensure it is executable on any system by any user.


KyleAure:
Trying RTC #build first to make sure fix works.

gjwatts (approved):

KyleAure:
Let's try another #build

PR #15998, opened 2021-02-22
MP Config Introspection

### LINKED ISSUE #15804: Create a server dump introspector for MicroProfile Config

Azquelt:
Second attempt at adding an Introspector for MP Config, addressing the concerns which were raised for #15830.

Fixes #15804


dev/io.openliberty.microprofile.config.internal.common/src/io/openliberty/microprofile/config/internal/common/ConfigIntrospector.java
```diff
@@ -0,0 +1,111 @@
...
+                pw.printf("   - %s = %s\n", entry.getKey(), obscuredValue);
+            }
+        } catch (Exception e) {
+            pw.printf("Exception introspecting config source: %s\n", e);
+        }
+    }
+
+    @Override
+    protected String getObscuredValue(String key, Object value) {
+        if (MAYBE_SENSITIVE.matcher(key).find()) {
```
tevans78, line 106: Can you add a test for this? A few variations of key name that should match the pattern?

Azquelt: Good point, added

tevans78 (approved):

Azquelt:
#build
#run-libby-bot

PR #16004, opened 2021-02-22
281853 - timing issue in jwtsso server startup

c00crane:
Fix timing issue in jwtsso fat - add a wait for the ltpa ready message


c00crane:
#build

ayoho (approved):

PR #16005, opened 2021-02-22
Issue 14977 and 15585: Clarify some OIDC client metatype descriptions

### LINKED ISSUE #14977: OIDC RP: Make description of allowCustomCacheKey attribute more clear
TS004203138
<s>
The openidClientConfig has a property called `allowCustomCacheKey` and a hidden property called `includeCustomCacheKeyInSubject`.

`allowCustomCacheKey` does not control if a cache key is put in the Subject or not, `includeCustomCacheKeyInSubject` does.  If a customer does not want a cache key in the subject at all, they can't do that because they do not know that the `includeCustomCacheKeyInSubject` property exists.

I propose that we expose a property and, instead of just exposing `includeCustomCacheKeyInSubject`, we create a property called `includeCacheKeyInSubject` and set `includeCustomCacheKeyInSubject` based on its value (if both are in the config, `includeCacheKeyInSubject` wins because it is the one that is exposed).

The reason why I propose `includeCacheKeyInSubject` instead of just exposing `includeCustomCacheKeyInSubject` is because `includeCustomCacheKeyInSubject` is confusing.  If you set `allowCustomCacheKey=false,` and then make them set `includeCustomCacheKeyInSubject=false`:

"_I just told you to not have a custom cache key and now you are telling me that I have to tell you to not put that custom cache key that you didn't create on the subject?_"  -- We know that's not what it means, but you see how that can be confusing to a customer</s>

It seems that `includeCustomCacheKeyInSubject` was already exposed.  It was exposed as `allowCustomCacheKey`.  The problem is that the `includeCustomCacheKeyInSubject` attribute wasn't working in the first place so then `allowCustomCacheKey` didn't either.  #14975 addresses that issue.

I want this issue to be used to perhaps fix up the description of the `allowCustomCacheKey` attribute.  This is what it says:

_Specifies if a custom cache key is used to store users in an authentication cache. Set this to false when using the jwtsso feature to allow the security subject to be constructed directly from the jwtsso cookie. The default value is true._

It also means that, when not using jwtsso, the Subject can be constructed directly from the registry.  If it is reconstructed, the result will be that there will be no SSO components on the Subject.  
You may want to set this property to false if your LTPA cookie is used by more than one server.

### LINKED ISSUE #15585: OIDC client userIdentifier and userIdentityToCreateSubject descriptions are confusing
An internal customer had a problem distinguishing between these properties in the OIDC client doc, https://www.ibm.com/support/knowledgecenter/SSAW57_liberty/com.ibm.websphere.liberty.autogen.nd.doc/ae/rwlp_config_openidConnectClient.html

**userIdentifier**  (no default)
_Specifies a JSON attribute in the ID token that is used as the user principal name in the subject. If no value is specified, the JSON attribute "sub" is used._

**userIdentityToCreateSubject** (default=sub)
_Specifies a user identity in the ID token used to create the user subject._

Chunlong said that **userIdentityToCreateSubject**  is an old property and that **userIdentifier**, **groupidentifier**, **realmIdentifier** were added together.

After some discussion, Chunlong verified the following:

**userIdentifier** and **userIdentityToCreateSubject** mean exactly the same thing.
They both default to 'sub'.
If you enter a value for both, the value for **userIdentifier** will only be used. 

This is what I propose:

<hr></hr>

**userIdentifier**  (default=sub)
_Specifies a JSON attribute in the ID token that is used as the user principal name in the subject. If no value is specified, the JSON attribute "sub" is used._

**userIdentityToCreateSubject** (default=sub)
_Specifies a JSON attribute in the ID token that is used as the user principal name in the subject. If no value is specified, the JSON attribute "sub" is used.  The value for this property is overridden by the value for userIdentifier, if specified._

ayoho:
Fixes #14977
Fixes #15585

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.client/resources/OSGI-INF/l10n/metatype.properties
+++ b/dev/com.ibm.ws.security.openidconnect.client/resources/OSGI-INF/l10n/metatype.properties
@@ -227,7 +227,7 @@ jwkParameter=Custom Parameters For JWK Endpoint
 jwkParameter.desc=Specifies custom parameters to send to JWK endpoint of the OpenID Connect provider.
 
 allowCustomCacheKey=Allow custom cache key
-allowCustomCacheKey.desc=Specifies if a custom cache key is used to store users in an authentication cache. If set to true, the security subject can be constructed directly from the user registry. If the security subject is constructed from the user registry, there will be no SSO components on the subject. Set this to false when using the jwtsso feature to allow the security subject to be constructed directly from the jwtsso cookie.
+allowCustomCacheKey.desc=Specifies whether a custom cache key is used to store users in an authentication cache. If this property is set to true, the security subject can be constructed directly from the user registry. If the security subject is constructed from the user registry, there will be no SSO components on the subject. Set this property to false when you use the jwtSso feature to allow the security subject to be constructed directly from the jwtSso cookie.
 
 oidcClientCustomRequestParam=Custom Parameters To Include In Requests
 oidcClientCustomRequestParam.desc=Specifies additional parameters to include in the requests to send to the provider.
--- a/dev/com.ibm.ws.security.openidconnect.client/resources/OSGI-INF/l10n/metatype.properties
+++ b/dev/com.ibm.ws.security.openidconnect.client/resources/OSGI-INF/l10n/metatype.properties
@@ -227,7 +227,7 @@ jwkParameter=Custom Parameters For JWK Endpoint
 jwkParameter.desc=Specifies custom parameters to send to JWK endpoint of the OpenID Connect provider.
 
 allowCustomCacheKey=Allow custom cache key
-allowCustomCacheKey.desc=Specifies whether a custom cache key is used to store users in an authentication cache. If this property is set to true, the security subject can be constructed directly from the user registry. If the security subject is constructed from the user registry, there will be no SSO components on the subject. Set this property to false when you use the jwtSso feature to allow the security subject to be constructed directly from the jwtSso cookie.
+allowCustomCacheKey.desc=Specifies whether a custom cache key is used to store users in an authentication cache. If this property is set to true and the cache key for a user is not found in the authentication cache, the user will be required to log in again. Set this property to false when you use the jwtSso feature to allow the security subject to be constructed directly from the jwtSso cookie. Set this property to false when you do not use the jwtSso feature to allow the security subject to be constructed directly from the user registry. If the security subject is reconstructed from the user registry, there will be no SSO components in the subject. If your LTPA cookie is used by more than one server, consider setting this property to false. If your application always requires the SSO components to be present in the subject, you must either set this property to true or use the jwtSso feature.
 
 oidcClientCustomRequestParam=Custom Parameters To Include In Requests
 oidcClientCustomRequestParam.desc=Specifies additional parameters to include in the requests to send to the provider.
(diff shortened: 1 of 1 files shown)


ayoho:
#build

dev/com.ibm.ws.security.openidconnect.client/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -227,7 +227,7 @@ jwkParameter=Custom Parameters For JWK Endpoint
 jwkParameter.desc=Specifies custom parameters to send to JWK endpoint of the OpenID Connect provider.
 
 allowCustomCacheKey=Allow custom cache key
-allowCustomCacheKey.desc=Specifies if a custom cache key is used to store users in an authentication cache. Set this to false when using the jwtsso feature to allow the security subject to be constructed directly from the jwtsso cookie.  The default value is true.
+allowCustomCacheKey.desc=Specifies if a custom cache key is used to store users in an authentication cache. If set to true, the security subject can be constructed directly from the user registry. If the security subject is constructed from the user registry, there will be no SSO components on the subject. Set this to false when using the jwtsso feature to allow the security subject to be constructed directly from the jwtsso cookie.
```
Charlotte-Holt, line 230: Minor suggested update:

>Specifies **whether** a custom cache key is used to store users in an authentication cache. If **this property is** set to true, the security subject can be constructed directly from the user registry. If the security subject is constructed from the user registry, there will be no SSO components on the subject. Set **this property** to false when you use the **jwtSso** feature to allow the security subject to be constructed directly from the **jwtSso** cookie.

Charlotte-Holt (commented):
@ayoho Just a small suggestion for one of these messages. Let me know if you have any questions.

ayoho:
#libby

barbj (commented):
I think that you got the values for allowCustomCacheKey backwards.  false means that it can be reconstructed from the registry.

andrea-mh:

barbj:
I know what happens with OIDC/LTPA.  I talked with Ut and Jackson and got more info about jwtsso.  I'll write up a new def and append it here when its ready.

barbj:
Specifies if a custom cache key is used to store users in an authentication cache.  When this value is set to true, if the cache key for a user is not found in the authentication cache, the user will be required to login again.  When using the jwtsso feature, set this property to false to allow the Subject to be constructed directly from the jwtsso cookie.  When not using the jwtsso feature, set this property to false to allow the Subject to be constructed directly from the user registry.  If the Subject is reconstructed from the user registry, there will be no SSO components on the Subject.  You may want to set this property to false if your LTPA cookie is used by more than one server.  If your application requires the SSO components always be present on the Subject, you must either set this property to true, or use the jwtsso feature.

barbj:
Specifies if a custom cache key is used to store users in an authentication cache.  You may want to set this property to false if your LTPA cookie is used by more than one server. If your application requires the SSO components always be present on the Subject, you must either set this property to true, or use the jwtsso feature.
**true**
When this value is set to true, if the cache key for a user is not found in the authentication cache, the user will be required to login again. 
**false**
When using the jwtsso feature, set this property to false to allow the Subject to be constructed directly from the jwtsso cookie. When not using the jwtsso feature, set this property to false to allow the Subject to be constructed directly from the user registry. If the Subject is reconstructed from the user registry, there will be no SSO components on the Subject.

ayoho:
#libby
#build

PR #16036, opened 2021-02-25
Update Docker image to use TLS1.2 for JDBC testDB2SecureConnection

gjwatts:
The `com.ibm.ws.jdbc_fat_db2` FAT test `testDB2SecureConnection` fails at Java 16 because the Docker image that it uses only supports TLS1.1 and TLS1.1 was removed in Java 16.  Updated the Docker image to use TLS1.2 to allow the test to pass, and point test to updated Docker image.


gjwatts:
#build

dev/com.ibm.ws.jdbc_fat_db2/fat/src/com/ibm/ws/jdbc/fat/db2/FATSuite.java
```diff
@@ -42,7 +42,8 @@
         };
     }
 
-    static final DockerImageName db2Image = DockerImageName.parse("aguibert/db2-ssl:1.0")
+    // Updated docker image to use TLS1.2 for secure communication
+    static final DockerImageName db2Image = DockerImageName.parse("gjwatts/db2-ssl-tls12:latest")
```
KyleAure, line 46: Can you tag this image with a version as well, and use that here. 
For example `gjwatts/db2-ssl-tls12:1.0`
Using latest could be unsafe since it is assumed the latest image could be changed and introduce a breaking change without any changes actually happening in the repo itself.

KyleAure (changes requested):
Also I searched dockerhub and did not find the `gjwatts/db2-ssl-tls12` image.  Can you link to it in this PR? Or was it only uploaded to artifactory?

dev/com.ibm.ws.jdbc_fat_db2/fat/src/com/ibm/ws/jdbc/fat/db2/FATSuite.java
gjwatts: Good point, thanks for catching that!  I have renamed the container and added a specific version.  It is now `gjwatts/db2-tls12:1.0` (the ssl part seemed redundant) and can be found [here](https://hub.docker.com/r/gjwatts/db2-tls12).  I also updated the code to point to the new name and reflected the fact that the container is now based on DB2 `11.5.5.0` (the old one was `11.5.0.0a`)

gjwatts:
#build

KyleAure (approved):

gjwatts:
#libbybot

PR #16059, opened 2021-02-26
OpenAPI 2.0: use fast mode when scanning archives

Azquelt:
Using fast mode keeps the archive open while we scan through its
classes, but we need to manually disable it after we've finished
scanning the archive.


Azquelt:
#build

Azquelt:
This didn't make any difference to performance so we didn't merge it.

PR #16065, opened 2021-02-26
Upgrade oracle image versions

KyleAure:
Old oracle docker image had a pre configured user `SYSTEM` whose password was set to expire 03/05/2021.  Upgraded container to have no password timeouts and pushed it here: https://hub.docker.com/r/kyleaure/oracle-18.4.0-xe-prebuilt

There was another image that was built on top of this one that I also updated and pushed here: https://hub.docker.com/r/kyleaure/krb5-oracle


KyleAure:
#build

mswatosh (approved):

PR #16066, opened 2021-02-26
NullPointerException on parked managed connection

njr-11:
Test resource adapter has a managed connection implementation that is unable to cope with a cleanup or destroy without an intermediate getConnection request.  This happens when the connection manager creates a managed connection for parked handles.


njr-11:
#build

jmstephensgit (approved):

PR #16084, opened 2021-03-01
Disable JPA Query tests affected by RTC282016

jgrassel:
I need to find a better way to handle timezone translation for these tests, so disabling them until I find a more acceptable method.


jgrassel:
#build

dazey3 (approved):
LGTM

PR #16094, opened 2021-03-02
Rework test validation

KyleAure:

### DIFF
--- a/.github/workflow-scripts/validate-tests.sh
+++ b/.github/workflow-scripts/validate-tests.sh
@@ -31,13 +31,14 @@ then
     testsRemoved=$(cat $TEMP_COMPAR_DIR/diff | grep "<" )
     if [ ! -z "$testsAdded" ]; then
         echo "::error::The following fat(s) need to be added to a category under .github/test-categories: $testsAdded"
+        echo "::warning::Not all fats were run, this build should not be considered complete."
+        exit 0; # Post error to annotation, but do not fail build outright
     fi
 
     if [ ! -z "$testsRemoved" ]; then
         echo "::error::The following fat(s) need to be removed from a category under .github/test-categories: $testsRemoved"
+        exit 1; # Post error to annotation and fail build, otherwise, we will try to run a fat that does not exist.
     fi
-    
-    exit 1;
 else 
     echo "List of tests in .github/test-categories matches the list of projects with '_fat' in the name"
 fi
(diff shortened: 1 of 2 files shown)


.github/workflow-scripts/validate-tests.sh
```diff
@@ -31,13 +31,14 @@ then
     testsRemoved=$(cat $TEMP_COMPAR_DIR/diff | grep "<" )
     if [ ! -z "$testsAdded" ]; then
         echo "::error::The following fat(s) need to be added to a category under .github/test-categories: $testsAdded"
```
tevans78, line 33: Shouldn't this say warning too?

.github/workflow-scripts/validate-tests.sh
```diff
@@ -31,13 +31,14 @@ then
...
+        echo "::warning::Not all fats were run, this build should not be considered complete."
+        exit 0; # Post error to annotation, but do not fail build outright
     fi
 
     if [ ! -z "$testsRemoved" ]; then
         echo "::error::The following fat(s) need to be removed from a category under .github/test-categories: $testsRemoved"
+        exit 1; # Post error to annotation and fail build, otherwise, we will try to run a fat that does not exist.
```
tevans78, line 40: I'm not convinced that a missing bucket should prevent all the others from running?

.github/workflow-scripts/validate-tests.sh
KyleAure: It could, but I think error fits better here.  The developer has an error they need to fix.  The error annotation does not stop the build from progressing it just produces the annotation.

KyleAure: What was stopping the build from progressing past this point was `exit 1`

.github/workflow-scripts/validate-tests.sh
KyleAure: The reason I put this validation steps as one of the first things we do is so that the developer can get quick feedback.  If I forget to add/remove a test from a category I want to know as soon as possible so I can correct it and push the change.  If all we do is warn users then they aren't going to realize the mistake until after a 3 hour build and have to rebuild everything again.  

Honestly, I think failing quick is the right way to go.  I'd rather have a build that is building everything I expect or not at all.

tevans78 (approved):

KyleAure:
#libbybot

PR #16096, opened 2021-03-02
Remove file dependency on build.example.testcontainers_fat

KyleAure:
During builds gradle copies files under `/publish/files/` to `/build/autoFVT/lib/LibertyFATTestFiles/`. On some build systems the permissions of these files are preserved, and on others the permissions are not preserved.
Work around this limitation by instead running startup SQL via connection.


KyleAure:
#build

gjwatts (approved):

KyleAure:
#libbybot

PR #16119, opened 2021-03-03
disable parking of managed connections

njr-11:
The WMQ resource adapter has a bug on the associateConnection path for parking JMSContext connection handles, so we cannot have that path be default and need to disable it in order to avoid a breaking change.


jmstephensgit (approved):

njr-11:
#build

njr-11:
#build

PR #16121, opened 2021-03-03
additional config option for PostgreSQL targetServerType

njr-11:
The PostgreSQL JDBC Driver added `primary` as a preferred config option for `targetServerType`, deprecating some of the existing values.

https://jdbc.postgresql.org/documentation/head/connect.html

This pull adds `primary` so that users can configure it.  A separate design issue #16122 will cover how to deal with the deprecated options.


njr-11:
#build

gjwatts (approved):

PR #16142, opened 2021-03-05
Add flush to use of printwriter in the jwt builder fat

c00crane:
When the builder test app runs, it builds a response with builder and token content - the response that is returned to the test app is truncated.
I'm going to add a flush to the routine that builds the response - I'm not convinced that it'll fix the issue (since the response has information from the beginning and end and is missing information from the middle - it shouldn't hurt...


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.jwt_fat.builder,com.ibm.ws.security.jwt_fat.consumer

ayoho (approved):

PR #16163, opened 2021-03-08
CDI FAT Bucket updates

tevans78:
This is a roll-up of some previous changes to the CDI 2.0 and Annotations buckets, with some additional review updates ... to use fixed beans.xml files rather than generating them.


tevans78:
#build

dev/fattest.simplicity/src/com/ibm/websphere/simplicity/beansxml/BeansAsset.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * ...or directly...
+ *
+ * <pre>
+ * <code>
+ * BeansAsset beans = new BeansAsset(Mode.ALL);
+ *
+ * WebArchive war = ShrinkWrap.create(WebArchive.class)
+ *              .addPackage(MyClass.class.getPackage())
+ *              .addAsWebInfResource(beans, "beans.xml");
```
Azquelt, line 38: A static constructor method may make this nicer if you were actually to use it this way. Seems odd to be calling a constructor to get what is effectively a constant.

```
 * WebArchive war = ShrinkWrap.create(WebArchive.class)
 *              .addPackage(MyClass.class.getPackage())
 *              .addAsWebInfResource(getBeansXml(ALL), "beans.xml");
```

dev/fattest.simplicity/src/com/ibm/websphere/simplicity/CDIArchiveHelper.java
```diff
@@ -69,6 +70,16 @@ public static WebArchive addBeansXML(WebArchive webArchive, Asset beans) {
...
 
+    /**
+     * Create a WEB-INF/beans.xml file in a war
+     *
+     * @param webArchive The WAR to create the beans.xml in
+     * @param srcPackage The package where the source beans.xml file can be found
+     */
+    public static WebArchive addBeansXML(WebArchive webArchive, Package srcPackage) {
+        return webArchive.addAsWebInfResource(srcPackage, BEANS_XML, BEANS_XML);
+    }
```
Azquelt, line 81: It's possibly nicer to pass the class in here rather than the package.

I'm actually tempted to suggest getting rid of this method and recommend that tests do
```
addBeansXml(war, MyTestServlet.class.getResource("beans.xml"));
```

This would make the method slightly less magic.

dev/io.openliberty.cdi.3.0.internal_fat/fat/src/io/openliberty/cdi30/internal/fat/tests/CDI30BeansXMLTest.java
```diff
@@ -52,7 +52,7 @@ public static void setUp() throws Exception {
                                               .addClass(CDI30BeansXMLTestServlet.class.getName())
                                               .addClass(SimpleBean.class.getName())
                                               .addClass(AnnotatedBean.class.getName());
-        cdiBeansXMLWar = CDIArchiveHelper.addBeansXML(cdiBeansXMLWar, Mode.ALL, Version.CDI30);
+        cdiBeansXMLWar = CDIArchiveHelper.addBeansXML(cdiBeansXMLWar, DiscoveryMode.ALL, CDIVersion.CDI30);
```
Azquelt, line 55: Given that all the helper methods modify the argument anyway, can please not do the assignment on this line?

The only reason to that all the shrinkwrap methods return `this` is so that you can chain method calls together for brevity and given that we're not doing that, I don't think it makes sense to pretend that you might need to assign the returned value.

dev/fattest.simplicity/src/com/ibm/websphere/simplicity/beansxml/BeansAsset.java
tevans78: OK I added a static constructor method and cached the possible results.

dev/fattest.simplicity/src/com/ibm/websphere/simplicity/CDIArchiveHelper.java
tevans78: These are convienence utility methods designed to reduce repetitious typing. I have added a couple of extra versions of the method; One which takes the "owning" class and the other which takes a resource URL ... so you are free do it in whichever way suits.

dev/io.openliberty.cdi.3.0.internal_fat/fat/src/io/openliberty/cdi30/internal/fat/tests/CDI30BeansXMLTest.java
tevans78: ok, don't much like method chaining anyway

tevans78:
#build

Joseph-Cass (approved):
LGTM :)

tevans78:
#run-libby-bot

Issue #16152, opened 2021-03-05
NPE in LibertyServer.stopServer when the server was never started.


tbitonti:
This is a FAT only logging bug.  No deliverable code is impacted.  The impact is to FAT logs, which will display an extraneous warning and an extraneous NullPointerException stack trace.

---

FAT cleanup, during cleanup, will emit a warning due to an NPE.

The NPE occurs when attempting to invoke the ACEScanner on any server which was never started.

The NPE is due to a missing assignment to 'consoleAbsPath' in 'setup'.

---

[03/05/2021 12:33:36:703 EST] 001 LibertyServer                  stopServer
  W Caught exception trying to scan for AccessControlExceptions

java.lang.NullPointerException
	at com.ibm.websphere.simplicity.RemoteFile.getParentPath(RemoteFile.java:944)
	at com.ibm.websphere.simplicity.RemoteFile.<init>(RemoteFile.java:194)
	at com.ibm.websphere.simplicity.RemoteFile.<init>(RemoteFile.java:177)
	at componenttest.topology.impl.LibertyServer.getConsoleLogFile(LibertyServer.java:4331)
	at com.ibm.ws.fat.util.ACEScanner.<init>(ACEScanner.java:52)
	at componenttest.topology.impl.LibertyServer.stopServer(LibertyServer.java:2447)
	at componenttest.topology.impl.LibertyServer.stopServer(LibertyServer.java:2237)
	at componenttest.topology.impl.LibertyServer.stopServer(LibertyServer.java:2212)
	at componenttest.topology.impl.LibertyServer.postTestTidy(LibertyServer.java:2593)
	at componenttest.topology.impl.LibertyServerFactory.tidyAllKnownServers(LibertyServerFactory.java:327)
	at componenttest.custom.junit.runner.FATRunner$2.evaluate(FATRunner.java:338)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:300)
...

PR #16168, opened 2021-03-08
Fix for 16152: NPE when LibertyServer invokes ACEScanner if the server was not started

tbitonti:
Fix for issue #16152.

Ensured assignment of 'consoleAbsPath', which removes an NPE when invoking ACEScanner if the server was never started.  Since that leads to a new FileNotFoundException, the ACEScanner is not run when the console log does not exist.

Removal of compiler warnings.
Conversion of closable resource use try-with-resources, including cases missing try/finally.
Correction of a few "methodName" values.
Ensured "entering" log statemenets have corresponding "exiting" statements.
Changed one System.out to Log.info.  Universally use the cached class and class name variables instead of using 'getClass()' or 'getClass().getName()'.
Reduced code width in a number of places.
Numerous JavaDoc cleanups.

OpenLiberty Pull Requester,

list of bugs fixed.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2020 IBM Corporation and others.
+ * Copyright (c) 2011, 2021 IBM Corporation and others.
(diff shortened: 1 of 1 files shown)


tbitonti:
#build

jimblye (approved):
Would be a lot easier to review if the unnecessary, but safe,  changes were in a separate commit.

PR #16178, opened 2021-03-09
Reduce JPA Query FAT's multi-line Exception cause string analysis to …

jgrassel:
It's possible for on Windows systems, for the multi-line String comparison for the Exception cause to fail because of differing new-line encodings.  Reduced the string comparison to a single line while still searching for the essential message parts.


jgrassel:
#build

dazey3 (approved):
LGTM

PR #16192, opened 2021-03-10
Use concurrent locks

AndrewDJBanks:


AndrewDJBanks:
Fixes #16175

sakerl (commented):
Comparing these changes to the ones in the tWAS codebase, I think there's a line missing here.
In CATAsynchReadAheadReader.consumeMessages(), there is a finally block in the tWAS code that isn't in the code here. It mostly contains Xct (monitoring?) operations, which isn't in the Liberty codebase, which is presumably why the block as a whole isn't there, but there's also the line:
	if (State.UNDEFINED!=fallback) consumerSession.setState(fallback);
which sets the state if a problem occurred while stopping the consumer, and this seems to be missing in the Liberty changes.

AndrewDJBanks:
#build

sakerl (approved):
Looks good

AndrewDJBanks:
#libbybot

PR #16209, opened 2021-03-11
Ignore error messages when the same application tries to to come up t…

jvanhill:
…wice. This can occur when we are doing dynamic server configuration updates.

This fix overrides that made in #15718.


jvanhill:
#build

jvanhill:
The last build appeared to have frozen. Respinning.

#build

utle (approved):

PR #16171, opened 2021-03-08
Issue 16167: Fixes and Kerberos Ldap FAT support

kristip17:
Fixes #16167

Two runtime changes, both discussed in slack:

1. Added a doPriv in the login module

2. Added a clear option for the KerberosLRU cache for LdapRegistry/ContextManager to call

Initial check in of FAT infrastructure and some FAT tests. I have some simple bind and ticketCache tests. Keytab is still in progress. Still have some to-dos to address -- follow up #16181

General structure:

-  the ApacheDS infrastracture is started in a beforeClass/afterClass globally for the bucket.

- There's a general helper class.

- There's a base UserLoginTest that the different bind type tests will expand on. It has a baseline "login" test where I do a good/bad password check of a user, isValidUser, getGroupsForUser and getUsersForGroup.

Function covered so far:
`SimpleBindTest` -- running baseline login
- BindAuthMech is simple
- BindAuthMech is none, Ldap allows anon bind and disallows
- BindAuthMech is not set
- `swapFromSimpleToKerberos` is in progress

`TicketCacheBindTest`
- Cannot contact KDC
- Empty, missing, or invalid `krb5PrincipalName`
- Empty, missing, or read-only  `krb5TicketCache`
- Start with expired cache, manually updated to valid
- Swap to keytab

`TicketCacheBindMultiRegistryTest`
- Use two Ldaps (ApacheDS and UnboundID) with allowOp = false. Confirm that neither registry can login when ApacheDS is down or cache is expired (and recovers when Apacheds is up, valid cache).
- Use two Ldaps (ApacheDS and UnboundID) with allowOp = true. Confirm that UnboundID can login when ApacheDS is down or cache is expired (and recovers when Apacheds is up, valid cache).

`TicketCacheBindLongRunTest`
- Restart ApacheDS server, confirm login recovers after LDAP restart
- Start with a valid certificate that is renewable and expires during runtime, verify ticket is renewed
- Start with a valid certificate that is not renewable and expires during runtime, ticket is not renewed, but manual renew works.

PR #16216, opened 2021-03-12
Issue 16181: Add tests and fixes for Kerberos Ldap

### LINKED ISSUE #16181: KRB5: Add Keytab and Config FAT test for Kerberos Ldap and other to-dos
For #12794

See PR for list of fixes and test updates.

kristip17:
Fixes #16181 

Runtime: Updated `KerberosService` to set the config to original JVM setting if modify is removing the config value. (added clear to helper class `SecureAction`)

Runtime: Clear attribute and search cache when the `<Kerberos/>` config is updated

Runtime: Adjusted some of the error handling and tracing in `ContextManager`

Addressed last batch of review comments from #16171

Added some LeakedPassword detection in the tests

Added several keytab and config tests

Add/expanded simple/ticketcache tests

Added more test comments

Removed unnecessary `setTraceMarkToEndOfDefaultTrace` --- the `updateConfigDynamically` resets log marks.

Swapped trace checks to `findStringsInLogsAndTraceUsingMark`

Customize stop messages allowed per test instead of global list in `CommonBindTest`


kristip17:
#build

jvanhill (changes requested):
I have a few small changes. I also wonder about the wrongUser.keytab. Is it always going to fail b/c it is the wrong user, or will it fail b/c the credential inside it expired at some point? Or does it not really matter?

dev/com.ibm.ws.security.wim.adapter.ldap/src/com/ibm/ws/security/wim/adapter/ldap/LdapConstants.java
```diff
@@ -227,4 +227,6 @@
     String SECUREWAY_LDAP_SERVER = "IBM SECUREWAY DIRECTORY SERVER";
 
     String CUSTOM_LDAP_SERVER = "CUSTOM";
+
+    String KERBEROS_UDPATE_MSG = "The Kerberos configuration was updated, refresh the context pool.";
```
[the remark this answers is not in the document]
kristip17: `recreateDirContext` takes an errorMessage to log so we don't lose the reason why we're running it. I set it the message in a constant so it'd be easier to keep consistent if we change it in the future.

kristip17:
> I have a few small changes. I also wonder about the wrongUser.keytab. Is it always going to fail b/c it is the wrong user, or will it fail b/c the credential inside it expired at some point? Or does it not really matter?

It is supposed to fail because the user can't be found in the keytab which is generally a valid keytab (versus an empty file or wrong file name) but it looks like I have an empty keytab instead of a populated one. I will fix it in my next round of fat updates -- #16226

dev/com.ibm.ws.security.wim.adapter.ldap_fat.krb5/fat/src/com/ibm/ws/security/wim/adapter/ldap/fat/krb5/TicketCacheBindTest.java
```diff
@@ -57,6 +65,25 @@ public void basicLoginChecks() throws Exception {
...
     }
 
+    /**
+     * Run golden path tests for bindAuthMech=GSSAPI, a valid krbPrincipalName and valid krb5TicketCache
+     *
+     * @throws Exception
+     */
+    @Test
+    @CheckForLeakedPasswords(LdapKerberosUtils.BIND_PASSWORD)
+    @AllowedFFDC("javax.naming.NamingException") // temporary, remove when Issue #16231 is fixed
```
jvanhill, line 75: Do you have a comment in 16231 to come back and remove this?

jvanhill (approved):

dev/com.ibm.ws.security.wim.adapter.ldap_fat.krb5/fat/src/com/ibm/ws/security/wim/adapter/ldap/fat/krb5/TicketCacheBindTest.java
kristip17: It does-- #16231

kristip17:
#libby

kristip17:
All krb5 ldap tests ran, test failure is a known issue.

PR #16236, opened 2021-03-15
Only add runlevel to server if repeattests is active

benjamin-confino:
Add an if statement so we only add runlevel to server log directory names if repeat tests is active.


benjamin-confino:
#build

jonhawkes (approved):

PR #16265, opened 2021-03-16
use default helper class if unspecified

njr-11:
If heritage behavior is switched on and a helper class isn't provided, we need to default it based on knowledge of which database/driver is being used.


njr-11:
#build

mswatosh (approved):

PR #16269, opened 2021-03-16
Issue 16226: Add more Kerberos Ldap FATs

### LINKED ISSUE #16226: KRB5: Additional tests variations for LDAP Kerberos
For #12794

Keytab is set in the config file
Realm name is defined by java property
No default realm in config file
More ticket cache with context pool enabled

kristip17:
Fixes #16226 

Added/changed
- Keytab is defined in the krb.conf file
- Realm is defined as a JVM property
- Move expiring cache tests to their own test
- Added a "with context pool" version to some tests.
- Added option to disable context pool and/or caches
- In order to do the realm JVM test, had to try and set the KDC to port 88. Then had to change the badPort test to use an actual bad port (instead of -1). This will also fix a BB on the Mac runs.


kristip17:
#libby

kristip17:
#build

jvanhill (approved):

kristip17:
#libby

kristip17:
#build

kristip17:
#libby

kristip17:
Per slack, the failure in the PB is a known issue.

PR #16291, opened 2021-03-18
Reduce timestamp precision of bundle and feature caches

brenthdaniel:
Docker container file systems can have inconsistent timestamps where at some point in the build process the file system has millisecond precision and in the final container it only has second precision. Neither the feature cache nor the bundle cache should care about millisecond level precision, so we can reduce the precision of these timestamps when we do comparisons.


brenthdaniel:
#build

brenthdaniel:
#build

brenthdaniel:
#build

brenthdaniel:
#run-libby-bot

PR #16297, opened 2021-03-19
14901 shell variable support in server.env

jimblye:
OpenLiberty Pull Requester,

list of bugs fixed.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/BootstrapConfig.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/BootstrapConfig.java
@@ -10,7 +10,6 @@
  *******************************************************************************/
 package com.ibm.ws.kernel.boot;
 
-import java.io.BufferedWriter;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
@@ -1164,7 +1163,7 @@ public class BootstrapConfig {
      * @return
      */
     protected ReturnCode generateServerEnv(boolean generatePassword) {
-        BufferedWriter bw = null;
+
         File serverEnv = getConfigFile("server.env");
         try {
             char[] keystorePass = PasswordGenerator.generateRandom();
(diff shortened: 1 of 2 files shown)


jimblye:
#build

jimblye:
#libby

covener:
Is the eval necessary? Just source/. seems to accomplish the same thing?

jimblye:
@covener 
Without the eval, I can't get it to execute "source" or "."

I tried:
  . \"$1\"
  and 
  " . \"$1\""
  
 Note that "source" workers on fewer shells.

jimblye:
Here's an alternative to "sourcing" the server.env.    This is a bit mind boggling and **only works in bash.** So it's probably not acceptable as is.
It is just a test program that will take a line with variables and resolve those variables.   So the proposal would be to call the replace_bash_variables() function for all of the values in server.env.
**It only works for ${varName} syntax.   Does not work for $varName.**   
I don't have a solution yet unless it's bash.  

#!/bin/**bash**

```
## safeEcho: Portable echo that can handle backslashes and leading dashes.
safeEcho()
{
  cat <<EOF
$*
EOF
  return $?
}

function replace_bash_variables () {
    local line="$1"
    local sed_args=()
    local varname
    while read varname ; do
        sed_args+=(-e "s/\${$varname\}/${!varname}/g")
    done < <(safeEcho $line | grep -o '[$][{][a-z_][a-z_]*[}]' | grep -o '[a-z_][a-z_]*' | sort -u)
    if [[ "${#sed_args[@]}" = 0 ]] ; then
        # if no variables to replace, just cat the line:
        safeEcho "$line" | cat 
    else
        safeEcho "$line" | sed "${sed_args[@]}" 
    fi
}

echo `replace_bash_variables '${cow}fred${myvar}dog${cat}horse'`
echo `replace_bash_variables 'freddoghorse'`
```
If you set the variables cow, myvar, and cat and then run the script, it should make sense.

jimblye:
#build

covener:
> Also, I know eval has a bad rep. But I think as long as you are not 'eval'ing some command that comes from an unknown source, it should be ok?

yeah that is fair.

PR #16312, opened 2021-03-22
Fix issues with initDatabase() method

KyleAure:
We had several places where we were setting up the database, table, and stored procedures for use by the test servlet. 
One of the setup steps was happening in the servlet itself, which was not good practice as some unrelated issue could affect the setup step and cause build failures unrelated to the root cause. 

Moving all setup to a single method that uses native JDBC driver connection to avoid any appserver issues from affecting setup.


jmstephensgit (approved):

KyleAure:
#build

KyleAure:
#libbybot

PR #16313, opened 2021-03-22
Output GenerateCategories stderr and save off stdout

KyleAure:
Recently seen errors during the GenerateCategories step.  Since we were using the stdout of GenerateCategories to set the category matrix we had no debug output to look at if there was an error at this step.  Therefore, I changed it to allow stderr to be output to GitHub actions, and stdout is saved off to a file.

### DIFF
--- a/.github/workflow-scripts/GenerateCategories.java
+++ b/.github/workflow-scripts/GenerateCategories.java
@@ -19,7 +19,7 @@ import java.util.stream.Collectors;
 public class GenerateCategories {
     
     //Constants
-    static final boolean DEBUG = false;
+    static final boolean DEBUG = true;
     static final String TEST_CATEGORY_DIR = ".github/test-categories/";
     static final String MODIFIED_FILES_DIFF = ".github/modified_files.diff";
 
@@ -200,12 +200,16 @@ public class GenerateCategories {
             if (modifiedFile.startsWith(TEST_CATEGORY_DIR))
                 continue;
             if (!modifiedFile.startsWith("dev/")) {
-                debug("WARN: Found modified file outside of 'dev/' tree.");
+                debug("WARN: Found modified file outside of 'dev/' tree: " + modifiedFile);
                 modifiedProjects.add("INFRA_OR_UNKNOWN");
             }
-            String projectName = modifiedFile.substring(4);
-            projectName = projectName.substring(0, projectName.indexOf('/'));
-            modifiedProjects.add(projectName);
+            try {
+                String projectName = modifiedFile.substring(4);
+                projectName = projectName.substring(0, projectName.indexOf('/'));
+                modifiedProjects.add(projectName);
+            } catch (StringIndexOutOfBoundsException e) {
+                debug("Could not parse modifiedFile=" + modifiedFile);
+            }
         }
 
         return modifiedProjects;
@@ -233,6 +237,6 @@ public class GenerateCategories {
     private static void debug(Object msg) {
         if (!DEBUG)
             return;
-        System.out.println(msg);
+        System.err.println(msg);
     }
 }
--- a/.github/workflow-scripts/generate-params.sh
+++ b/.github/workflow-scripts/generate-params.sh
@@ -14,6 +14,9 @@
 #   App: GenerateCategories.java        - calculates category matrix and outputs in json format
 ##############################
 
+# global variables
+OUTPUT_FILE=tmp/GenerateCategories.out && mkdir -p -- "$(dirname -- "$OUTPUT_FILE")" && touch -- "$OUTPUT_FILE"
+
 # Save off PR Summary
 echo $PRBODY >> .github/pull_request_body.txt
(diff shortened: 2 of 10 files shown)


.github/workflow-scripts/run-fat-build.sh
```diff
@@ -78,22 +78,23 @@ for FAT_BUCKET in $FAT_BUCKETS
...
-  TEMP_DIR=tmp/$FAT_BUCKET && mkdir -p $TEMP_DIR && touch $TEMP_DIR/gradle.log
+  OUTPUT_FILE=tmp/$FAT_BUCKET/gradle.log && mkdir -p -- "$(dirname -- "$OUTPUT_FILE")" && touch -- "$OUTPUT_FILE"
 
   #Run fat
-  ./gradlew :$FAT_BUCKET:buildandrun $FAT_ARGS &> $TEMP_DIR/gradle.log || BUCKET_PASSED=false
+  ./gradlew :$FAT_BUCKET:buildandrun $FAT_ARGS &> $OUTPUT_FILE || BUCKET_PASSED=false
 
   OUTPUT_DIR=$FAT_BUCKET/build/libs/autoFVT/output
   RESULTS_DIR=$FAT_BUCKET/build/libs/autoFVT/results
-  mkdir -p $OUTPUT_DIR && touch $OUTPUT_DIR/gradle.log && mv $TEMP_DIR/gradle.log $OUTPUT_DIR/gradle.log
```
gjwatts, line 88: Don't you still need to create the $OUTPUT_DIR directory first:
```
mkdir -p $OUTPUT_DIR
```

first so you can later, if needed, do
```
touch "$OUTPUT_DIR/passed.log";
```
or
```
touch "$OUTPUT_DIR/fail.log";
```
?

I got an error using the `touch` command like this:
```
touch a/b
```
when I tested it in Linux (WSL2) if directory `a` does not already exist.  But this might work different in GHA.

.github/workflow-scripts/run-unit-tests.sh
```diff
@@ -9,15 +9,18 @@
...
 ############################## 
 
+#Setup gradle
 cd dev
 chmod +x gradlew
 
-echo "Running gradle test and testReport tasks.  This will take approx. 30 minutes."
+# global variables
+OUTPUT_FILE=tmp/gradle.log && mkdir -p -- "$(dirname -- "$OUTPUT_FILE")" && touch -- "$OUTPUT_FILE"
+UNIT_DIR=unit-results/ && mkdir -p $UNIT_DIR
```
gjwatts: Previously, you had set
UNIT_DIR=`$PWD`/unit-results
now it is:
UNIT_DIR=unit-results/

so the `$PWD` prefix is no longer needed?

.github/workflow-scripts/run-fat-build.sh
KyleAure: the gradle command makes the output directory

.github/workflow-scripts/generate-params.sh
```diff
@@ -27,8 +30,12 @@ cat .github/modified_files.diff
 echo "::endgroup::"
 
 # Generate test categories
-MATRIX_RESULT=$(java .github/workflow-scripts/GenerateCategories.java .github/pull_request_body.txt)
+echo "::group::GenerateCategories Debug"
+java .github/workflow-scripts/GenerateCategories.java .github/pull_request_body.txt 1>> $OUTPUT_FILE
```
gjwatts, line 34: I am confused about something.  Please help me understand this correctly.  In `.github/workflow-scripts/GenerateCategories.java` you send debug information to System.err, but here it looks like you are only capturing `stdout` (1>>) to the $OUTPUT_FILE.  Where does the `stderr` output go?

.github/workflow-scripts/GenerateCategories.java
```diff
@@ -19,7 +19,7 @@
 public class GenerateCategories {
     
     //Constants
-    static final boolean DEBUG = false;
+    static final boolean DEBUG = true;
```
gjwatts, line 22: Just want to verify that you always want to collect debug information.

.github/workflow-scripts/generate-params.sh
KyleAure: stdout will be written to the output file and read into the `MATRIX_RESULT` variable. 
stderr will be written to the console so that it is viewable by the developer 
Screen Shot 2021-04-20 at 2 53 19 PM

.github/workflow-scripts/GenerateCategories.java
KyleAure: Yes, previously enabling debug would corrupt the variable `MATRIX_RESULT` in the build script.  Now debug goes to stderr while the actual output we want to go to MATRIX_RESULT is sent to stdout.

gjwatts (approved):

KyleAure:
#libbybot

PR #16319, opened 2021-03-22
isUnsupported legacy DataStoreHelper method

njr-11:
Add integration point for the isUnsupported DataStoreHelper method, such that the identifyException configuration overrides it if present.


njr-11:
#build

jmstephensgit (approved):

PR #16339, opened 2021-03-24
transaction branch coupling for Oracle

njr-11:
Support the branch-coupling resource reference extension when Oracle is used.
Because this will need to introduced as nonship and separately be promoted to beta/GA, we will need to guard it with a temporary internal dataSource attribute for now.  Expect the attribute to be removed upon GA.
Add tests for both tight (default) and loose coupling with Oracle.


KyleAure (approved):

njr-11:
#build

njr-11:
Build has 1 test failure which is a duplicate of a preexisting issue in an unrelated test bucket.

PR #16341, opened 2021-03-24
Remove unwanted trace and move empty handle list out of JCA trace group

njr-11:
Remove unnecessary tracing for the EmptyHandleListContextProvider and move tracing in general for the empty handle list to a separate trace group to keep it from cluttering up the JCA trace.


njr-11:
#build

gjwatts (approved):

PR #16350, opened 2021-03-24
Microsoft SQL Server tightly coupled transactions

njr-11:
Add support for tightly coupled transactions by the Microsoft SQL Server JDBC Driver to Liberty.  Unfortunately, the function in Microsoft SQL Server is broken and needs to be fixed.  I opened issue #16351 to track that, but this pull will be merged regardless so that the issue can be easily reproduced (just switch the rollback in the test case to a commit).


njr-11:
#build

jmstephensgit (approved):

PR #16358, opened 2021-03-25
fix gradle file logic

KyleAure:
When gradle runs it overwrites the output directories instead of appending.  
For some reason this didn't present itself as a build failure during testing.
Fixing this by using a temp folder during buildAndRun, and copying that output to the build directory afterwords.

### DIFF
--- a/.github/workflow-scripts/run-fat-build.sh
+++ b/.github/workflow-scripts/run-fat-build.sh
@@ -78,12 +78,14 @@ for FAT_BUCKET in $FAT_BUCKETS
 do
   echo "::group:: Run $FAT_BUCKET with FAT_ARGS=$FAT_ARGS"
   BUCKET_PASSED=true
-  OUTPUT_DIR=$FAT_BUCKET/build/libs/autoFVT/output
-  RESULTS_DIR=$FAT_BUCKET/build/libs/autoFVT/results
-  mkdir -p $OUTPUT_DIR && touch $OUTPUT_DIR/gradle.log
+  TEMP_DIR=tmp/$FAT_BUCKET && mkdir -p $TEMP_DIR && touch $TEMP_DIR/gradle.log
 
   #Run fat
-  ./gradlew :$FAT_BUCKET:buildandrun $FAT_ARGS &> $OUTPUT_DIR/gradle.log || BUCKET_PASSED=false
+  ./gradlew :$FAT_BUCKET:buildandrun $FAT_ARGS &> $TEMP_DIR/gradle.log || BUCKET_PASSED=false
+
+  OUTPUT_DIR=$FAT_BUCKET/build/libs/autoFVT/output
+  RESULTS_DIR=$FAT_BUCKET/build/libs/autoFVT/results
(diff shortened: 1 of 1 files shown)


.github/workflow-scripts/run-fat-build.sh
```diff
@@ -78,12 +78,14 @@ for FAT_BUCKET in $FAT_BUCKETS
...
   echo "::group:: Run $FAT_BUCKET with FAT_ARGS=$FAT_ARGS"
   BUCKET_PASSED=true
-  OUTPUT_DIR=$FAT_BUCKET/build/libs/autoFVT/output
-  RESULTS_DIR=$FAT_BUCKET/build/libs/autoFVT/results
-  mkdir -p $OUTPUT_DIR && touch $OUTPUT_DIR/gradle.log
+  TEMP_DIR=tmp/$FAT_BUCKET && mkdir -p $TEMP_DIR && touch $TEMP_DIR/gradle.log
 
   #Run fat
-  ./gradlew :$FAT_BUCKET:buildandrun $FAT_ARGS &> $OUTPUT_DIR/gradle.log || BUCKET_PASSED=false
+  ./gradlew :$FAT_BUCKET:buildandrun $FAT_ARGS &> $TEMP_DIR/gradle.log || BUCKET_PASSED=false
```
gjwatts, line 84: And here does the `||` work like `if` in Java?  If the build fails, then BUCKET_PASSES is set to `false`?

Another option to consider for the future would be if the operator `&>>` works to append like it does in *nix.  Not sure if it would work in this environment or not.

gjwatts (commented):
I noticed your `Unit tests` and `FAT Tests - GENERAL` had errors in them -> https://github.com/KyleAure/open-liberty/runs/2194670900?check_suite_focus=true.  

Were those expected?

.github/workflow-scripts/run-fat-build.sh
```diff
@@ -78,12 +78,14 @@ for FAT_BUCKET in $FAT_BUCKETS
 do
   echo "::group:: Run $FAT_BUCKET with FAT_ARGS=$FAT_ARGS"
   BUCKET_PASSED=true
-  OUTPUT_DIR=$FAT_BUCKET/build/libs/autoFVT/output
-  RESULTS_DIR=$FAT_BUCKET/build/libs/autoFVT/results
-  mkdir -p $OUTPUT_DIR && touch $OUTPUT_DIR/gradle.log
+  TEMP_DIR=tmp/$FAT_BUCKET && mkdir -p $TEMP_DIR && touch $TEMP_DIR/gradle.log
```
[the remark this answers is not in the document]
KyleAure: Yeah so && indicates what to run if the previous command was successful.  So I don't want to try to create $TEMP_DIR/gradle.log if the mkdir command failed.

.github/workflow-scripts/run-fat-build.sh
KyleAure: Correct, the command after || will only run if the previous command failed.

gjwatts (approved):

KyleAure:
#libbybot

PR #16360, opened 2021-03-25
Remove hard coded system-packages lists

tjwatson:
Depend on the framework to calculate the JVM system packages for us.
This allows better package calculation on Java 9+.


tjwatson:
#build

PR #16371, opened 2021-03-26
Minimise Eclipse settings for IIOP projects

ngmr:
#build


ngmr:
The test failure in the build clearly has nothing to do with the changes here, so I think this will be good to go once reviewed.

joe-chacko (approved):
Burn the core prefs and destroy them utterly.
Cast their ashes into the outer darkness.

PR #16383, opened 2021-03-26
Disable authentication cache for JWT SSO FATs with auth filter

ayoho:
The authentication cache needs to be disabled for the JWT SSO FATs to avoid having the core security functionality recreate a logged out SSO cookie from the auth cache.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.jwtsso_fat

Zech-Hein (approved):

PR #16403, opened 2021-03-29
Issue 16400: Fix message creation and FAT check for Krb5 Ldap

### LINKED ISSUE #16400: KRB5: Update badPrincipalName test to allow for other valid error messages
On some levels of Java 8, we hit this message, which is also a valid path to indicate an invalid PrincipalName:

```
CWIML4512E: The badPrincipalName Kerberos principal name is incorrectly formatted, or the realm name is missing, or a default realm name cannot be found. java.lang.IllegalArgumentException: null
```

Will also take the opportunity to tweak the error message in cases where the getMessage is null (`java.lang.IllegalArgumentException` vs `java.lang.IllegalArgumentException: null`)

kristip17:
Fixes #16400

Runtime fix: Avoid printing an Exception's getMessage() if it is null in `ContextManager.handleKerberos` -- example: `java.lang.IllegalArgumentException: null`

Test fix: Java 8 is more variable than I thought -- add second acceptable error message on a bad principalName.


jvanhill (approved):

kristip17:
#build
#libby

PR #16413, opened 2021-03-29
tests for app defined datasource with container auth for direct lookup

njr-11:
As requested during the feature test summary review, I'm adding test coverage for application-defined data sources that configure container authentication for direct lookups, either to disabled or enabled, and expect results consistent with the application-defined configuration.


njr-11:
#build

gjwatts (approved):

PR #16422, opened 2021-03-30
Revert "Create setHtmlContentTypeOnError Property"

volosied:
Reverts OpenLiberty/open-liberty#16120

Missed the translation deadline


volosied:
Marking as internal instead. See #16426

PR #16427, opened 2021-03-30
Fix JPA Spec 1.0 Part 2 FAT bucket with respect to Postgres issues.

jgrassel:
The `com.ibm.ws.jpa_spec10_part02_fat` bucket requires a couple of fixes with respect to running on postgresdb.


dev/com.ibm.ws.jpa_spec10_part02_fat/test-applications/jpa10/relationships/oneXmany/ddl/WEB-INF/classes/JPA10_ONEXMANY_DELETE_POSTGRES.ddl
```diff
@@ -1,10 +1,8 @@
 DELETE FROM EmbedIDOMEntityA;
 DELETE FROM RELOM_EIDOMENTA;
-DELETE FROM EmbedIDOMEntityA_EmbedIDOMEntityB;
```
dazey3, line 3: Interesting. There doesn't appear to be any references to these tables in any of the CREATE .ddl files for `com.ibm.ws.jpa_spec10_part02_fat`. It looks like some of the JoinTable relationships were commented out in the original delivery 2 years ago. For instance:

`com.ibm.ws.jpa_spec10_part02_fat\test-applications\jpa10\relationships\oneXmany\src\com\ibm\ws\jpa\fvt\relationships\oneXmany\entities\containertype\annotated\OMContainerTypeEntityA.java`

I'm fine with deleting these DDL statements from DELETE since there is no matching CREATE, but we should also remove them from the DROP .ddl as well, yes?
Also, can we just delete them from all the database ddls for `com.ibm.ws.jpa_spec10_part02_fat`? It appears that no database has a CREATE for these tables.

dazey3 (changes requested):

dazey3 (approved):
LGTM

PR #16456, opened 2021-04-01
Fix small timing issue with bundle resolution

brenthdaniel:


brenthdaniel:
#build

brenthdaniel:
In extraordinarily rare cases, it's possible for us to process a feature bundle before it has been resolved. This can cause metatype to not be processed, which results in missing configuration and possibly service activation.

hlhoots (approved):
LGTM.

PR #16473, opened 2021-04-02
make sure example.com is available before using in saml logout bucket

c00crane:
Had added code previously to make sure that example.com was available before trying to use it.
This change will modify where the checker method is called from - only tests using example.com as the post logout page will use it.
This change will also enhance the check.  It will now verify that we landed on the correct page.  In my testing, I noticed one instance where I landed on the domain sale page instead of the normal example.com page.


c00crane:
#build

ayoho (approved):
The new method is longer than I think is advisable, but it looks like it'll work.

PR #16480, opened 2021-04-05
Tests will now install features from local repo

chitwansharmaa:
For featureUtility_fat intermittent errors regarding connection to maven repository for installing features was seen. Made some changes in the test methods to now bypass installing features from maven instead install some of the features from local repository which will consist of the dependency files.


chitwansharmaa:
#build

bensonlatibm (approved):
Changes look ok

PR #16485, opened 2021-04-05
Increase Mongo FAT connection timeout

tkburroughs:
Avoid connection timeout in Mongo fat that is testing the
setting of timout options by setting value to something
greater than the default rather than smaller.


tkburroughs:
#build

olendvcook (approved):

PR #16517, opened 2021-04-07
Create atomic dispatcher state

ngmr:
Create atomic dispatcher state
    
Each instance of the new DispatcherState represents an immutable snapshot of a dispatcher's state.
    
As it is immutable, this can then be safely inspected without further locking (to check the dispatcher's health or describe its current state.
    
Transition from one state snapshot to the next is done using AtomicReference.compareAndSet() in the Dispatcher.

#build


ngmr:
Superceded by #16535

PR #16518, opened 2021-04-07
enable allowqueryparamwithnoequal by default in JAX-RS 3.0

andymc12:
This fix resolves a TCK issue where content send using content type of `application/x-www-form-urlencoded`.  The TCK sends content like "ABC", but when the JAX-RS resource method is invoked, the entity parameter is injected with "".  In our testing, if we send content of "ABC=", then the injected entity parameter is "ABC".  

The reason for this is that RESTEasy uses the `HttpServletRequest.getParameterMap()` to determine the content when handling `application/x-www-form-urlencoded` requests.  By default, the web container does not process parameters that do not contain an equals sign. Setting the `allowqueryparamwithnoequal` attribute to true fixes this.


andymc12:
The actual fix for this issue came from the web container by changing the default (in EE9) of this property to true - so now this PR just verifies that fix by adding a test case very similar to the TCK test case that was the original purpose of this PR/issue.

#build

WhiteCat22 (approved):

andymc12:
#libby

PR #16528, opened 2021-04-07
EJB Persistent Timer fat updates

olendvcook:
EJB Timer BB fixes, ignore database exceptions during server shutdown and move when a CDL is countdown


olendvcook:
#build

dev/com.ibm.ws.ejbcontainer.timer.persistent_fat/test-applications/PersistentTimerCoreEJB.jar/src/com/ibm/ws/ejbcontainer/timer/persistent/core/ejb/StatelessTimedBean.java
```diff
@@ -673,6 +671,7 @@ public void testTimerServicePhase2() throws Exception {
         // -------------------------------------------------------------------
         {
             //wait for first interval
+            timerFirstIntervalLatch = new CountDownLatch(1);
```
tkburroughs, line 674: Based on the error, the problem appears to be that the timer has run at least once, but the latest expected execution is late, -3430 ms late.  The checking below is verifying the timer is run within + or - 900 ms of expected time, and so fails because the timer is too far off.  If this is the scenario, then does resetting the latch and waiting for the timer to run again really mitigate the slowness of the system.  Would it be better to wait first, and if the value is too far off then reset the latch and wait again, perhaps up to 3 times?

olendvcook:
#run-libby-bot
#build

tkburroughs (approved):

PR #16530, opened 2021-04-07
split SAML jaxrs config FAT class to avoid reconfig timeouts

c00crane:
The RSSamlIDPInitiatedSSLConfigTests.java starts 3 servers.  One server has a large and complex configuration.  Most, but not all of the tests in that class use that default config.  We're running into problems on Windows with Java 8 where the reconfig back to the default config is taking longer than the configRefresher allows for.  It will only wait 1 minute before logging a msg that the refresh has timed out.  The config refresh will continue behind the scenes after the configRefresher logs the timeout msg and returns to the test code.  Our test code will then wait for the reconfig completed successfully msg.  We never get that msg because the refresh stopped and logged the timout).  We will actually wait long enough, but terminate our process (and fail the test) because we don't get the success message.
I"ll split the class into 2 classes
- 1 will use the default config for each test (therefore, not needing any reconfigs).
- 1 will start the server with one config and then reconfig for each test case - these configs are much simpler and will not run into the issue - we won't reconfig back to the original config  - this too will prevent the issue.


c00crane:
#run-libby-bot
#build

c00crane:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_7K7dgJflEeumwuCQm6DXsQ

teddyjtorres (approved):

PR #16549, opened 2021-04-08
test Subject from AccessControlContext after security context propagated

njr-11:
Write test to determine if OpenLiberty has the same issue with security context propagation as https://github.com/smallrye/smallrye-context-propagation/pull/260

It turns out that it does have the same issue, so some checks within the test are disabled and marked with comments.

I also tried with WSSubject in case the behavior might differ, but is the same with respect to getting the null Subject from the AccessControlContext.


njr-11:
#build

mswatosh (approved):

njr-11:
build has 1 intermittent failure in an unrelated test bucket that wasn't updated by these changes

PR #16559, opened 2021-04-09
permit IBM JDK for Oracle 21c JDBC driver kerberos

### LINKED ISSUE #16307: Update Liberty to not block use of Oracle 21c JDBC driver with IBM Java 8 and kerberos authentication.
When support for using kerberos authentication on datasources was added to Liberty, Oracle support docs stated that their JDBC drivers did not support kerberos authentication with any JDK other than Oracle.  To add in serviceability, code was added to Liberty to detect an attempt to use kerberos auth on an Oracle driver in conjunction with IBM Java 8 and to throw an exception preventing a connection attempt.  With the release of Oracle 21c, reportedly the Oracle limitation has been lifted.  We need to update Liberty code to allow connection attempts if the driver version is 21 or greater and add/update FAT tests to verify the behavior.

**Steps to Reproduce**
Start Liberty with IBM Java 8 and a datasource configured to use the 21c version of Oracle JDBC driver and kerberos auth.  Execute connection test.  Observe exception in logs and lack of connection attempt.

**Expected behavior**
If a datasource is configured with Oracle JDBC driver version 21c or greater, a connection attempt should be made.  If version is less than 21, the current behavior should be observed.

njr-11:
fixes #16307

### DIFF
--- a/dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/OracleHelper.java
+++ b/dev/com.ibm.ws.jdbc/src/com/ibm/ws/rsadapter/impl/OracleHelper.java
@@ -50,6 +50,7 @@ import javax.transaction.xa.XAResource;
 import com.ibm.websphere.ras.Tr;
 import com.ibm.websphere.ras.TraceComponent;
 import com.ibm.ws.ffdc.FFDCFilter;
+import com.ibm.ws.ffdc.annotation.FFDCIgnore;
 import com.ibm.ws.kernel.service.util.JavaInfo;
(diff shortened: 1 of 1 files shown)


njr-11:
#build

KyleAure (approved):

KyleAure (changes requested):
I think we need to do some more testing to be sure that the Oracle 21c driver did actually break their dependency on the Oracle JDK.

KyleAure (approved):
Tested and working on Windows, Linux and Mac machines.  This looks like a safe change, but going to look into this failing when forwardable tickets are enabled.

PR #16597, opened 2021-04-13
allow additional SSL handshake message

c00crane:
Add an additional ssl connection error message to the list of errors that can occur.
Adding CWWKO0801E


c00crane:
#build

ayoho (approved):

PR #16606, opened 2021-04-14
Add wait for ssl restart to sigAlg tests

c00crane:
The Signature Algorithm tests need to wait for the SSL channel restart successful message to be logged when doing their reconfig.
We hit a case where we got the reconfig complete message, but did not get the ssl ready message until after we made our first request.


c00crane:
#build

ayoho (approved):

PR #16607, opened 2021-04-14
Update bnd dependencies to consistent version (5.3.0)

tevans78:
#build
biz.aQute.bnd.annotation and biz.aQute.bnd.transform depend on biz.aQute.bnd so they should all be at a consistent version


cbridgha (approved):
LGTM - but really make sure you run a full "all test" build - because a change like this can change build behavior

tevans78:
LITE PB was green. This is a FULL PB, which is green and almost finished: https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_GM9BEJ03EeuEbPrbRku4xA

PR #16612, opened 2021-04-14
Move beta flag setting earlier in bootstrap

NottyCode:
Currently we set the flag that the server is in beta in an OSGi bundle activator.
This means it cannot be read early in startup by things like logging, or even things
that start sooner than the OSGi bundle that currently activates it. So moving it much
sooner in the startup routine. ProductInfo doesn't rely on anything setup in bootstrap
so can be called at any point safely.


NottyCode:
#build

jhanders34 (approved):

NottyCode:
#build

NottyCode:
#libby

PR #16623, opened 2021-04-15
Beta fence Request Timing Metrics

Channyboy:
#build


fmhwong (changes requested):
`-Dcom.ibm.ws.beta.edition=true` is a JVM option.  Should it be put in `jvm.options`?

Channyboy:
> `-Dcom.ibm.ws.beta.edition=true` is a JVM option. Should it be put in `jvm.options`?

Yup! I thought they would be picked up the same in bootstrap.options similar to what we allow it do in logging. But that should be two different mechanisms. Plus I forgot to remove the `-D` from two of the files even if it did.

Channyboy:
#build
#run-libby-bot

fmhwong (approved):
LGTM

PR #16624, opened 2021-04-15
Ensure JAX-RS JsonPProvider can work in Java SE environment

andymc12:
The JsonPProvider that we ship with JAX-RS 2.1 works fine in OSGi environments (Liberty), but fails when run outside of Liberty.  This is because it depends on the JSON-P provider to be set using OSGi services.  This change enables the JsonPProvider message body reader/writer to function outside of Liberty so long as the JSON-P 1.1 APIs and a valid implementation are present on the classpath.


andymc12:
#build

andymc12:
Wrapping new code in try/catch. New #build

jim-krueger (approved):

andymc12:
#libby - build had another connection timeout issue, but all tests passed.

PR #16708, opened 2021-04-21
Ensure MP Config properties from the application are visible when mpOpenApi-2.0 runs filters

### LINKED ISSUE #16661: microprofile-config.properties is not loaded in OASFilter
I upgraded to OpenLiberty 21.0.0.4, because I wanted to start using Microprofile 4.0.

It seems that bug described and fixed in https://github.com/OpenLiberty/open-liberty/issues/9204 is present again.

Azquelt:
Main changes:
* Ensure that both the service loader file and the implementation class for the Smallrye OpenAPI jaxrs extension are both available on the thread context class loader.
* Set the TCCL correctly during mpOpenAPI-2.0 setup (which is when filters are run)
* Add test using MP Config in MP OpenAPI filter

Fixes #16661


Azquelt:
#build
#run-libby-bot

Azquelt:
#build
#run-libby-bot

tevans78 (approved):

PR #16716, opened 2021-04-21
transition from inheritance model to direct use of legacy datastorehelper

njr-11:
Switch to direct usage (via reflection) of the legacy DataStoreHelper, instead of through an intermediate interface which caused problems.


njr-11:
#build

njr-11:
Build has 1 test failure which is a duplicate of a preexisting issue in an unrelated test bucket.

jmstephensgit (approved):

PR #16730, opened 2021-04-22
Adding trace for test failure in RTC 278913. Windows systems are stil…

jvanhill:
…l showing a lock when trying to copy files, but the trace does not include which file is failing. I am adding that trace.


jvanhill:
#build

ayoho (approved):

PR #16731, opened 2021-04-22
remove references to GenericDataStoreHelper that were missed

njr-11:
Missed a couple of references to GenericDataStoreHelper which still need to be removed in order to get the legacy code path to build cleanly.  The classes in the heritage package themselves will need to be removed as well, but in a later phase.


njr-11:
#build

jmstephensgit (approved):

njr-11:
Build has 1 test failure, which is an infrastructure issue

PR #16732, opened 2021-04-22
Dump the server if an app fails to start

brenthdaniel:
Debug for an app start failure


brenthdaniel:
#build

hlhoots (approved):
LGTM

tbitonti (commented):
Do you want to **not** re-throw the exception?

PR #16748, opened 2021-04-23
Make memory threshold configurable for multipart attachments in RESTEasy

andymc12:
In RESTEasy, the memory threshold for multipart attachments (basically the attachment size limit before the implementation writes to disk) is 1024 bytes - and it is currently hardcoded, so multipart parts that are >1024 are written to disk for later processing, which is an expensive operation.

This change allows users to change the threshold by setting the MP Config property, "org.jboss.resteasy.plugins.providers.multipart.memoryThreshold" to a non-negative number.


andymc12:
#build

jkoehler22 (approved):

andymc12:
1 unrelated failure.

PR #16762, opened 2021-04-26
Break apart JPA github test-categories into smaller groups.

jgrassel:
The github test-categories for JPA might have grown too large to run under the allowed execution times, and need to be broken into smaller blocks.


jgrassel:
#build

tkburroughs:
Added @KyleAure as a reviewer, since I believe this only affects the GitHub Actions.

KyleAure (approved):

KyleAure:
Correct, no need to do an RTC build for this change.  It only affects github actions builds and can be merged.

jgrassel:
This is old news, and buckets have been broken apart in later PRs.

PR #16806, opened 2021-04-27
Add another failover ldap server to ldap registry config

c00crane:
We've seen some communication issues using ldap4 and ldap8 during testing.  In the latest case (at least), ldap4 and ldap8 were set to the same ldap server.  This doesn't do much for failover.  I'm looking into how many ldap servers of this type we have, but, so far, it looks like only 2.  ldap4 and ldap8 are getting set to the same server, while ldap7 is set to another server.
I'm going to add ldap7 to the list of failover servers, so, this should give us a better chance of failing over.


c00crane:
#build

ayoho (approved):

PR #16818, opened 2021-04-28
Mark SPI update as beta

benjamin-confino:
Marks the new SPI method for bean defining annotations as beta as more FAT tests are needed.

### DIFF
--- a/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
+++ b/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
@@ -628,6 +628,8 @@ public class CDIContainerImpl implements CDIContainer, InjectionMetaDataListener
         return this.probeExtensionArchive;
     }
 
+    private static boolean issuedBetaMessage = false;
+
     private ExtensionArchive newSPIExtensionArchive(ServiceReference<CDIExtensionMetadata> sr,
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
```diff
@@ -636,6 +638,15 @@ private ExtensionArchive newSPIExtensionArchive(ServiceReference<CDIExtensionMet
         Set<Class<?>> beanClasses = webSphereCDIExtensionMetaData.getBeanClasses();
         Set<Class<? extends Annotation>> beanDefiningAnnotationClasses = webSphereCDIExtensionMetaData.getBeanDefiningAnnotationClasses();
 
+        if (! beanDefiningAnnotationClasses.isEmpty()) {
+            if (! com.ibm.ws.kernel.productinfo.ProductInfo.getBetaEdition()) {
+                throw new UnsupportedOperationException("This method is beta and is not available.");
```
tevans78, line 643: I think that in this case we probably don't want to throw an exception. We should just ignore the content... as if it is not there.

Otherwise, any components which are adding this will also need to have their own beta guards.

tevans78: we decided to leave this exception and force implementors to include beta guards ... this makes it more obvious when it does not behave as may be expected

tevans78 (approved):

benjamin-confino:
#build

benjamin-confino:
#run-libby-bot

Issue #16826, opened 2021-04-28
Test Failure: org.eclipse.microprofile.metrics.test.ReusableMetricsTest


fmhwong:
RTC 283474

```
java.lang.NoClassDefFoundError: Could not initialize class com.jayway.restassured.RestAssured
at org.eclipse.microprofile.metrics.test.ReusableMetricsTest.setup(ReusableMetricsTest.java:76)
```

RTC 283475

```
java.lang.NoClassDefFoundError: com.jayway.restassured.RestAssured (initialization failure)
at java.base/java.lang.J9VMInternals.initializationAlreadyFailed(J9VMInternals.java:134)
at org.eclipse.microprofile.metrics.test.ReusableMetricsTest.setup(ReusableMetricsTest.java:76)
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make protected java.lang.Object java.lang.Object.clone() throws java.lang.CloneNotSupportedException accessible: module java.base does not "opens java.lang" to unnamed module @9c321d4
at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:357)
at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
at java.base/java.lang.reflect.AccessibleObject.setAccessible(AccessibleObject.java:130)
at org.codehaus.groovy.reflection.CachedClass$3$1.run(CachedClass.java:86)
at java.base/java.security.AccessController.doPrivileged(AccessController.java:682)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:81)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:79)
at org.codehaus.groovy.util.LazyReference.getLocked(LazyReference.java:46)
at org.codehaus.groovy.util.LazyReference.get(LazyReference.java:33)
at org.codehaus.groovy.reflection.CachedClass.getMethods(CachedClass.java:250)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:116)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:79)
at org.codehaus.groovy.util.LazyReference.getLocked(LazyReference.java:46)
at org.codehaus.groovy.util.LazyReference.get(LazyReference.java:33)
[...]

PR #16715, opened 2021-04-21
Interim fix to use the released 2.0.5-RC1 and 1.1.4-RC1 to address Java 16

Channyboy:
Interim fix for #16162

Issue #16162, opened 2021-03-08
Java16 Test Failure: org.eclipse.microprofile.metrics.test.MpMetricTest.org.eclipse.microprofile.metrics.test.MpMetricTest


fmhwong:
RTC 282266, 282303

```
org.eclipse.microprofile.metrics.test.MpMetricTest:java.lang.ExceptionInInitializerError: null
at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:357)
at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
at java.base/java.lang.reflect.AccessibleObject.setAccessible(AccessibleObject.java:130)
at org.codehaus.groovy.reflection.CachedClass$3$1.run(CachedClass.java:86)
at java.base/java.security.AccessController.doPrivileged(AccessController.java:312)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:81)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:79)
at org.codehaus.groovy.util.LazyReference.getLocked(LazyReference.java:46)
at org.codehaus.groovy.util.LazyReference.get(LazyReference.java:33)
at org.codehaus.groovy.reflection.CachedClass.getMethods(CachedClass.java:250)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:116)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:79)
at org.codehaus.groovy.util.LazyReference.getLocked(LazyReference.java:46)
at org.codehaus.groovy.util.LazyReference.get(LazyReference.java:33)
at org.codehaus.groovy.reflection.CachedClass.getMethods(CachedClass.java:250)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:116)
at org.codehaus.groovy.reflection.CachedClass$3.initValue(CachedClass.java:79)
at org.codehaus.groovy.util.LazyReference.getLocked(LazyReference.java:46)
at org.codehaus.groovy.util.LazyReference.get(LazyReference.java:33)
at org.codehaus.groovy.reflection.CachedClass.getMethods(CachedClass.java:250)
[...]

PR #16836, opened 2021-04-28
Include metrics 1.1 no auth fat to use the RC to address java 16 issues

Channyboy:
#build
Addresses #16826
Which is an overall temporary fix for #16162
The original PR to fix this should've been: https://github.com/OpenLiberty/open-liberty/pull/16715


fmhwong (approved):
LGTM

Channyboy:
The error encountered was that one of the child fat builds failed to start. However the com.ibm.ws.microprofile.metrics.1.1.noauth_fat_tck have executed 11 times without failure.

PR #16848, opened 2021-04-28
Keep separate track of prime (root) feature conflicts

tjwatson:
Prime conflicts are ones that are direct conflicts from the initial set
of root features.  The resolver has always removed such conflicts from
the initial set of root features but would also block the feature from
resolving if it was pulled in from a transitive dependency.

This causes a miscalculation in the initial blocked features from the
first pass over the configured features.  The blocked prime features
added to the initial count which then later would cause a miscalculation
for what additional permutations to try.  This ultimately causes what
appears to be an endless retry of new permutations.

The solution is to instead track the prime conflicts separately.  There
is no choice that can be made that would solve these types of conflicts
so they should not contribute to the initial block count that we use as
the bar for letting in new permutations to try.


tjwatson:
#build

tjwatson:
#build

dazavala (approved):

tjwatson:
#libby

PR #16948, opened 2021-04-29
the wrapper login module needs to return true/false

utle:
The Kerberos login wrapper module needs to return the true/false result from the JDK since some of the JDK version do not throw an exception when login/commit failed.

The Kerberos service can check the result returns from the Kerberos login wrapper and issue the best possible error message.

### DIFF
--- a/dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/KerberosService.java
+++ b/dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/KerberosService.java
@@ -214,9 +214,9 @@ public class KerberosService {
      * the resulting Subject is cached.
      *
      * @param principal The principal to obtain a subject for
-     * @param pass      The password to be used via CallbackHandler. This will only be used if no password
-     *                      is found in the credential cache or keytab files first. May be null.
-     * @param ccache    The path to the credential cache to be used for this principal. May be null.
+     * @param pass The password to be used via CallbackHandler. This will only be used if no password
+     *            is found in the credential cache or keytab files first. May be null.
+     * @param ccache The path to the credential cache to be used for this principal. May be null.
      * @return A valid subject for the supplied principal
      */
     public Subject getOrCreateSubject(String principal, SerializableProtectedString pass, Path ccache) throws LoginException {
@@ -264,12 +264,17 @@ public class KerberosService {
         }
 
         krb5.initialize(subject, callback, sharedState, options);
-        krb5.login();
-        krb5.commit();
+        boolean goodResult = krb5.login();
+        if (goodResult)
+            goodResult = krb5.commit();
+
+        if (!goodResult) // login/commit return false but JDK does not throw an exception so we have to do it here
+            throwAnException(principal, ccache);
 
         // If the created Subject does not have a GSSCredential, then create one and
         // associate it with the Subject
         Set<GSSCredential> gssCreds = subject.getPrivateCredentials(GSSCredential.class);
+
         if (gssCreds == null || gssCreds.size() == 0) {
             GSSCredential gssCred = SubjectHelper.createGSSCredential(subject);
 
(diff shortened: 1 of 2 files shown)


utle:
#build

utle:
#build 
#spawn.fullfat.buckets=com.ibm.ws.security.spnego_fat,com.ibm.ws.security.oidc.client_fat.spnego,com.ibm.ws.security.oidc.server_fat.spnego,com.ibm.ws.security.wim.adapter.ldap_fat,com.ibm.ws.jdbc_fat_krb5,com.ibm.ws.jdbc_fat

kristip17 (approved):

utle:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.spnego_fat,com.ibm.ws.security.oidc.client_fat.spnego,com.ibm.ws.security.oidc.server_fat.spnego,com.ibm.ws.security.wim.adapter.ldap_fat,com.ibm.ws.jdbc_fat_krb5,com.ibm.ws.jdbc_fat

utle:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.spnego_fat,com.ibm.ws.security.oidc.client_fat.spnego,com.ibm.ws.security.oidc.server_fat.spnego,com.ibm.ws.security.wim.adapter.ldap_fat,com.ibm.ws.jdbc_fat_krb5,com.ibm.ws.jdbc_fat

mswatosh (changes requested):
Since the OpenJDK login module provides it's own LoginExceptions, we should be checking for false in the Krb5LoginModuleWrapper if `isIBMJdk8` is true and throwing a LoginException there.

utle:
> Since the OpenJDK login module provides it's own LoginExceptions, we should be checking for false in the Krb5LoginModuleWrapper if `isIBMJdk8` is true and throwing a LoginException there.

I handle that in the KerberosService where we have to throw an exception for other reasons.  Also, this will catch all other return false cases

mswatosh:
login and commit can't return false with the OpenJDK Krb5LoginModule so we don't need to do the check for openJDK, also it's fine for the Krb5LoginModuleWrapper to return `true`, since `false` would mean skip the LoginModule, which isn't an option here. So the only options are `true` if Authentication succeeds, or throw `LoginException` if the login fails. We really just need to correct the IBM JDK8 krb5LoginModule violating the LoginModule Interface by returning `false` when it should be throwing a `LoginException`

Once we address that, then we could catch any `LoginException` thrown by the LoginModule, and provide a more Liberty-specific (and translatable) message, with the original `LoginException` as the cause. There we could use knowledge of what the user has set in configuration to give a better hint (e.g. if the user only has a ccache set we can tell them to check that) while still keeping the original exception.

Let me know if you want me to clarify further

utle:
> login and commit can't return false with the OpenJDK Krb5LoginModule so we don't need to do the check for openJDK, also it's fine for the Krb5LoginModuleWrapper to return `true`, since `false` would mean skip the LoginModule, which isn't an option here. So the only options are `true` if Authentication succeeds, or throw `LoginException` if the login fails. We really just need to correct the IBM JDK8 krb5LoginModule violating the LoginModule Interface by returning `false` when it should be throwing a `LoginException`
> 
> Once we address that, then we could catch any `LoginException` thrown by the LoginModule, and provide a more Liberty-specific (and translatable) message, with the original `LoginException` as the cause. There we could use knowledge of what the user has set in configuration to give a better hint (e.g. if the user only has a ccache set we can tell them to check that) while still keeping the original exception.
> 
> Let me know if you want me to clarify further

IBM Krb5LoginModule return false for failure and clearly document it. However, what you suggest is the same thing what I have now. I choose to do it in the KerberosService instead since we already have the logic to issue the best possible error message. I don't want to call or duplicate that logic in the wrapper.

utle:
The more I think about it. We should open a JDK issue and let them fix it so we do not have to do anything here and make it consistent with OpenJDK. Close this PR.

PR #16953, opened 2021-04-29
Improve FeatureUtility performance 

### LINKED ISSUE #16700: Improve featureUtility performance with remote repository
FeatureUtility can take a long time to resolve features when using a remote repository with a slow network. 

Areas that could be improved: 
1.  Check and retrieve more than one checksums from maven repository
2. Check more than one artifact urls for missing features

jjiwooLim:
fixes #16700

### DIFF
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020 IBM Corporation and others.
+ * Copyright (c) 2020, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
@@ -58,29 +58,19 @@ public class ArtifactDownloader implements AutoCloseable {
 
     private final ProgressBar progressBar;
 
-    private static Map<String, Object> envMap;
-
-    private final String num_threads = System.getProperty("artifactThreads", DEFAULT_THREAD_NUM).toString();
+    private static Map<String, Object> envMap = null;
 
     private final ExecutorService executor;
-
-    private final static String DEFAULT_THREAD_NUM = "8";
 
     ArtifactDownloader() {
         this.downloadedFiles = new ArrayList<File>();
         this.progressBar = ProgressBar.getInstance();
-        this.envMap = null;
-        this.executor = Executors.newFixedThreadPool(Integer.parseInt(num_threads));
-
+        this.executor = Executors.newFixedThreadPool(ArtifactDownloaderUtils.getNumThreads());
     }
 
     private Future<String> submitDownloadRequest(String coords, String fileType, String dLocation, MavenRepository repository) {
         return executor.submit(() -> {
-            try {
-                synthesizeAndDownload(coords, fileType, dLocation, repository, false);
-            } catch (InstallException e) {
-                fine(e.getMessage());
-            }
+            synthesizeAndDownload(coords, fileType, dLocation, repository, false);
             return coords + "." + fileType;
         });
     }
@@ -146,7 +136,6 @@ public class ArtifactDownloader implements AutoCloseable {
                 Iterator<Future<?>> iter = futures.iterator();
                 try {
                     while (iter.hasNext()) {
-                        boolean updateProgress = false;
                         Future<?> future = iter.next();
                         if (future.isDone()) {
                             String downloadedCoords;
@@ -161,7 +150,7 @@ public class ArtifactDownloader implements AutoCloseable {
                         }
                     }
                     logger.fine("Remaining artifacts: " + futures.size());
-                    Thread.sleep(300);
+                    Thread.sleep(ArtifactDownloaderUtils.THREAD_SLEEP);
                 } catch (InterruptedException | ExecutionException e) {
                     throw new InstallException(e.getMessage());
                 }
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
@@ -124,7 +124,7 @@ public class ArtifactDownloader implements AutoCloseable {
             // we have downloaded mavenCoords.length * 2 (esa and pom) amount of features.
             double individualSize = progressBar.getMethodIncrement("downloadArtifacts") / (2 * mavenCoords.size());
             progressBar.updateMethodMap("downloadArtifact", individualSize);
-            logger.info(Messages.INSTALL_KERNEL_MESSAGES.getMessage("MSG_BEGINNING_DOWNLOAD_FEATURES"));
+            info(Messages.INSTALL_KERNEL_MESSAGES.getMessage("MSG_BEGINNING_DOWNLOAD_FEATURES"));
             for (String coords : mavenCoords) {
                 Future<?> future1 = submitDownloadRequest(coords, "esa", dLocation, repository);
                 futures.add(future1);
@@ -149,7 +149,7 @@ public class ArtifactDownloader implements AutoCloseable {
                             iter.remove();
                         }
                     }
-                    logger.fine("Remaining artifacts: " + futures.size());
+                    fine("Remaining artifacts: " + futures.size());
                     Thread.sleep(ArtifactDownloaderUtils.THREAD_SLEEP);
                 } catch (InterruptedException | ExecutionException e) {
                     throw new InstallException(e.getMessage());
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
@@ -29,8 +29,13 @@ import java.net.URLConnection;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.logging.Logger;
 
 import com.ibm.ws.install.InstallException;
@@ -52,6 +57,30 @@ public class ArtifactDownloader {
 
     private final ProgressBar progressBar = ProgressBar.getInstance();
     private static Map<String, Object> envMap = null;
+
+    final ExecutorService executor = Executors.newCachedThreadPool();
+
+    public Future<?> esaDownload(String coords, String dLocation, MavenRepository repository) {
+        return executor.submit(() -> {
+            try {
+                synthesizeAndDownload(coords, "esa", dLocation, repository, false);
+            } catch (InstallException e) {
+                e.printStackTrace();
+            }
+            return coords;
+        });
+    }
+
+    public Future<?> pomDownload(String coords, String dLocation, MavenRepository repository) {
+        return executor.submit(() -> {
+            try {
+                synthesizeAndDownload(coords, "pom", dLocation, repository, false);
+            } catch (InstallException e) {
+                e.printStackTrace();
+            }
+            return coords;
+        });
+    }
 
     public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLocation, MavenRepository repository) throws InstallException {
         info(Messages.INSTALL_KERNEL_MESSAGES.getLogMessage("STATE_CONTACTING_MAVEN_REPO"));
@@ -93,19 +122,40 @@ public class ArtifactDownloader {
             fine("The remote repository is missing the following artifacts: " + missingFeatureList.toString());
             throw ExceptionUtils.createByKey("ERROR_FAILED_TO_DOWNLOAD_ASSETS_FROM_REPO", "required", "feature(s)", repo);
         } else {
+            List<String> result = new ArrayList<String>();
+
+            final List<Future<?>> futures = new ArrayList<>();
             // we have downloaded mavenCoords.length amount of features.
             double individualSize = progressBar.getMethodIncrement("downloadArtifacts") / mavenCoords.size();
             progressBar.updateMethodMap("downloadArtifact", individualSize);
             logger.info(Messages.INSTALL_KERNEL_MESSAGES.getMessage("MSG_BEGINNING_DOWNLOAD_FEATURES"));
             for (String coords : mavenCoords) {
-                synthesizeAndDownload(coords, "esa", dLocation, repository, false);
-                synthesizeAndDownload(coords, "pom", dLocation, repository, false);
+                Future<?> future1 = esaDownload(coords, dLocation, repository);
+                futures.add(future1);
+                Future<?> future2 = pomDownload(coords, dLocation, repository);
+                futures.add(future2);
+//                synthesizeAndDownload(coords, "esa", dLocation, repository, false);
+//                synthesizeAndDownload(coords, "pom", dLocation, repository, false);
+            }
 
-                // update progress bar, drain the downloadArtifacts total size
-                updateProgress(individualSize);
-                progressBar.updateMethodMap("downloadArtifacts", progressBar.getMethodIncrement("downloadArtifacts") - individualSize);
-                progressBar.manuallyUpdate();
-                fine("Finished downloading artifact: " + coords);
+            for (int i = 0; i < futures.size(); i++) {
+                String downloadedCoords;
+                try {
+                    while (!futures.get(i).isDone()) {
+                        logger.info("Downloading index: " + i);
+                        Thread.sleep(300);
+                    }
+                    downloadedCoords = (String) futures.get(i).get();
+                    if (i % 2 == 1) {
+                        // update progress bar, drain the downloadArtifacts total size
+                        updateProgress(individualSize);
+                        progressBar.updateMethodMap("downloadArtifacts", progressBar.getMethodIncrement("downloadArtifacts") - individualSize);
+                        progressBar.manuallyUpdate();
+                        fine("Finished downloading artifact: " + downloadedCoords);
+                    }
+                } catch (InterruptedException | ExecutionException e) {
+                    e.printStackTrace();
+                }
             }
         }
     }
@@ -157,9 +207,10 @@ public class ArtifactDownloader {
             }
         }
         String[] checksumFormats = new String[3];
-        checksumFormats[0] = "SHA256";
-        checksumFormats[1] = "MD5";
-        checksumFormats[2] = "SHA1";
+
+        checksumFormats[0] = "MD5";
+        checksumFormats[1] = "SHA1";
+        checksumFormats[2] = "SHA256";
 
         try {
             if (individualDownload && ArtifactDownloaderUtils.fileIsMissing(urlLocation, envMap)) {
@@ -185,11 +236,12 @@ public class ArtifactDownloader {
             boolean someChecksumExists = false;
             boolean checksumFail = false;
             boolean checksumSuccess = false;
+            HashMap<String, String> checkSumCache = new HashMap<String, String>();
             for (String checksumFormat : checksumFormats) {
                 if (!checksumSuccess) {
-                    if (checksumIsAvailable(urlLocation, checksumFormat)) {
+                    if (checksumIsAvailable(urlLocation, checksumFormat, checkSumCache)) {
                         someChecksumExists = true;
-                        if (isIncorrectChecksum(fileLoc.getAbsolutePath(), urlLocation, checksumFormat)) {
+                        if (isIncorrectChecksum(fileLoc.getAbsolutePath(), urlLocation, checksumFormat, checkSumCache)) {
                             fine("Failed to validate " + checksumFormat + " checksum for file: " + filename);
                             checksumFail = true;
                         } else {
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -53,6 +58,30 @@
...
     private static Map<String, Object> envMap = null;
 
+    final ExecutorService executor = Executors.newCachedThreadPool();
+
+    public Future<?> esaDownload(String coords, String dLocation, MavenRepository repository) {
+        return executor.submit(() -> {
+            try {
+                synthesizeAndDownload(coords, "esa", dLocation, repository, false);
+            } catch (InstallException e) {
+                e.printStackTrace();
```
jacobwdv: Should we also error log occurrences where we have `e.printStackTrace()` or consider throwing them up the stack?

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -157,9 +207,10 @@ public void synthesizeAndDownload(String mavenCoords, String filetype, String dL
             }
         }
         String[] checksumFormats = new String[3];
-        checksumFormats[0] = "SHA256";
-        checksumFormats[1] = "MD5";
-        checksumFormats[2] = "SHA1";
+
+        checksumFormats[0] = "MD5";
```
jacobwdv, line 210: forget what i said in the scrum about these, for some reason I thought these hashes were related to encryption but these are just checksum hashes. I would vote we check the most popular checksum type first for best performance.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -93,19 +122,40 @@ public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLoca
...
+//                synthesizeAndDownload(coords, "esa", dLocation, repository, false);
+//                synthesizeAndDownload(coords, "pom", dLocation, repository, false);
+            }
 
-                // update progress bar, drain the downloadArtifacts total size
-                updateProgress(individualSize);
-                progressBar.updateMethodMap("downloadArtifacts", progressBar.getMethodIncrement("downloadArtifacts") - individualSize);
-                progressBar.manuallyUpdate();
-                fine("Finished downloading artifact: " + coords);
+            for (int i = 0; i < futures.size(); i++) {
```
jacobwdv: Unless we need to serially wait for each download i suggest going through the entire list before sleeping.
Something like:

```
while (!futures.isEmpty())
      {
         Iterator<Future<String>> iter = futures.iterator();

         while (iter.hasNext())
         {
            boolean updateProgress = false;
            Future<String> future = iter.next();
            if (future.isDone())
            {
               String downloadedCoords = future.get();
               if (updateProgress)
               {
                  // update progress bar, drain the downloadArtifacts total size
                  updateProgress(individualSize);
                  progressBar.updateMethodMap("downloadArtifacts",
                        progressBar.getMethodIncrement("downloadArtifacts") - individualSize);
                  progressBar.manuallyUpdate();
                  fine("Finished downloading artifact: " + downloadedCoords);
               }
               updateProgress = !updateProgress;
               iter.remove();
            }
         }

         logger.info("Downloading " + futures.size() + " artifacts");
         Thread.sleep(300);
      }
```

This way time will be spent checking the other futures in the list instead of sleeping.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -217,31 +269,29 @@ private void download(String urlLocation, String dLocation, String groupId, Stri
         }
     }
 
-    private boolean checksumIsAvailable(String urlLocation, String checksumFormat) {
+    private boolean checksumIsAvailable(String urlLocation, String checksumFormat, HashMap<String, String> checkSumCache) {
         boolean result = true;
         try {
-            ArtifactDownloaderUtils.getMasterChecksum(urlLocation, checksumFormat);
+            if (checkSumCache.get(checksumFormat) != null)
```
jacobwdv: assuming keys cannot be mapped to null, use `checkSumCache.containsKey()` instead

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloaderUtils.java
```diff
@@ -36,14 +41,39 @@
 public class ArtifactDownloaderUtils {
 
     private static boolean isFinished = false;
+    private final static Logger logger = InstallLogUtils.getInstallLogger();
 
     public static List<String> getMissingFiles(List<String> featureURLs, Map<String, Object> envMap) throws IOException {
         List<String> result = new ArrayList<String>();
```
jacobwdv: Change this to use `new Vector<String>`, `ArrayList` is not thread safe.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloaderUtils.java
```diff
@@ -36,14 +41,39 @@
...
+                    }
+                } catch (IOException ioe) {
+                    ioe.printStackTrace();
+                }
+            });
+            futures.add(future);
+        }
+
+        try {
+            for (Future<?> future : futures) {
```
jacobwdv: we could do a similar loop as suggested above here as well

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
jjiwooLim: thowing instead

jjiwooLim:
#libbybot

jjiwooLim:
#build

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -53,7 +59,32 @@
     private final ProgressBar progressBar = ProgressBar.getInstance();
     private static Map<String, Object> envMap = null;
 
-    public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLocation, MavenRepository repository) throws InstallException {
+    final ExecutorService executor = Executors.newCachedThreadPool();
+
+    public Future<?> esaDownload(String coords, String dLocation, MavenRepository repository) {
```
jacobwdv: I like the idea of having `esaDownload` and `pomDownload` as separate functions but we could do better with code reuse: 

```
public Future<?> submitDownloadRequest(String coords, String fileType, String dLocation, MavenRepository repository)
{
    return executor.submit(() -> {
            try {
                synthesizeAndDownload(coords, fileType, dLocation, repository, false);
            } catch (InstallException e) {
                e.printStackTrace(); // throw this exception or log it 
            }
            return coords + fileType;
        });
}

public Future<?> esaDownload(String coords, String dLocation, MavenRepository repository) { 
   return submitDownloadRequest(coords, "esa", dLocation, repository, false); 
}

public Future<?> pomDownload(String coords, String dLocation, MavenRepository repository) { 
   return submitDownloadRequest(coords, "pom", dLocation, repository, false); 
}
```

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -53,7 +59,32 @@
...
+            try {
+                synthesizeAndDownload(coords, "esa", dLocation, repository, false);
+            } catch (InstallException e) {
+                e.printStackTrace();
+            }
+            return coords + ".esa";
+        });
+    }
+
+    public Future<?> pomDownload(String coords, String dLocation, MavenRepository repository) {
```
jacobwdv: both esaDownload and pomDownload return `Future<String>` I would change this to show String is expected instead of `?`

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -53,7 +59,32 @@
...
+        return executor.submit(() -> {
```
jacobwdv, line 72: We can define this task as a `Callable<String>()` in order to get the hard typed Future<String> suggested above

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -53,7 +59,32 @@
     private final ProgressBar progressBar = ProgressBar.getInstance();
     private static Map<String, Object> envMap = null;
 
-    public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLocation, MavenRepository repository) throws InstallException {
+    final ExecutorService executor = Executors.newCachedThreadPool();
```
jacobwdv: This ExecutorService should be shutdown when users are done with the ArtifactDownloader object. 

A simple way to do this would be to implement `AutoCloseable`: https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html

Then, when users create the object they can do so within a try-with-resources block or will be expected to run the `close()` method themselves.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -75,11 +106,13 @@ public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLoca
         }
         updateProgress(progressBar.getMethodIncrement("establishConnection"));
         info(Messages.INSTALL_KERNEL_MESSAGES.getLogMessage("STATE_MAVEN_REPO_CONNECTION_SUCCESSFUL"));
+
         try {
             missingFeatures = ArtifactDownloaderUtils.getMissingFiles(featureURLs, envMap);
-        } catch (IOException e) {
+        } catch (IOException | InterruptedException | ExecutionException e) {
```
jacobwdv, line 104: what code changed here that requires us to catch these additional exceptions?

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -53,7 +59,32 @@
     private final ProgressBar progressBar = ProgressBar.getInstance();
```
jacobwdv: I know these aren't your changes here but I would prefer these variables to be set within a constructor. We could still have a 0-argument constructor

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -217,24 +277,31 @@ private void download(String urlLocation, String dLocation, String groupId, Stri
...
         String checksumLocal;
         try {
             checksumLocal = ArtifactDownloaderUtils.getChecksum(localFile, checksumFormat);
         } catch (IOException e) {
             return true;
         }
+      
+        String checksumOrigin = checkSumCache.get(checksumFormat);
+        if (checksumOrigin == null || !checksumLocal.equals(checksumOrigin)) {
         String checksumOrigin;
```
jacobwdv: checksumOrigin is defined twice, remove line 305

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -288,7 +355,7 @@ protected boolean testConnection(MavenRepository repository, List<String> mavenC
                 return true;
             }
         } catch (IOException e) {
-
+            logger.warning(repository.getRepositoryUrl() + " cannot be connected");
```
jacobwdv, line 340: log `IOException e` as well

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -309,7 +376,7 @@ protected boolean testConnection(MavenRepository repository) {
...
```
jacobwdv, line 362: log `IOException e` as well

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloaderUtils.java
```diff
@@ -36,14 +43,43 @@
...
-            if (!(exists(url, envMap) == HttpURLConnection.HTTP_OK)) {
-                result.add(url);
+            Future<?> future = executor.submit(() -> {
+
+                try {
+                    if (!(exists(url, envMap) == HttpURLConnection.HTTP_OK)) {
+                        result.add(url);
+                    }
+                } catch (IOException e) {
+                    e.printStackTrace();
```
jacobwdv: log stack trace / exception here.

jacobwdv (changes requested):
Looks good, some minor changes suggested

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
jjiwooLim: It's from the ExectuorService that gets thrown from ArtifactDownloaderUtils.getMissingFiles(featureURLs, envMap);

jjiwooLim:
#build

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -46,14 +52,41 @@
...
     private final static Logger logger = InstallLogUtils.getInstallLogger();
 
-    private final ProgressBar progressBar = ProgressBar.getInstance();
-    private static Map<String, Object> envMap = null;
+    private final ProgressBar progressBar;
 
-    public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLocation, MavenRepository repository) throws InstallException {
+    private static Map<String, Object> envMap;
+
+    private final String num_threads = System.getProperty("artifactThreads", DEFAULT_THREAD_NUM).toString();
```
jjiwooLim: Default number of threads is set to 8.
If user wants to increase/decrease number of threads, they can do so by setting the JVM_ARG.
ex) ENV JVM_ARGS="-DartifactThreads=5"

jacobwdv: Looks good, as discussed we will have to get someones blessing for the JVM argument name. Conventionally it would start with the package name that uses it, ie: `com.ibm.ws.install.artifactThreads`

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -46,14 +52,41 @@
...
+
+    private final ExecutorService executor;
+
+    private final static String DEFAULT_THREAD_NUM = "8";
+
+    ArtifactDownloader() {
+        this.downloadedFiles = new ArrayList<File>();
+        this.progressBar = ProgressBar.getInstance();
+        this.envMap = null;
+        this.executor = Executors.newFixedThreadPool(Integer.parseInt(num_threads));
```
jacobwdv: only risk here is using `parseInt()` within the constructor. It's very unlikely but if the user sets `artifactThreads=two` then this will throw a `NumberFormatException` and blow up. Instead of letting that happen we can catch the `NumberFormatException` and supply the default 8. (maybe log an error when that happens).

jacobwdv: a method for setting this would be a good place to log the value of it too if you go that route

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -46,14 +52,41 @@
...
+    }
+
+    private Future<String> submitDownloadRequest(String coords, String fileType, String dLocation, MavenRepository repository) {
+        return executor.submit(() -> {
+            try {
+                synthesizeAndDownload(coords, fileType, dLocation, repository, false);
+            } catch (InstallException e) {
+                fine(e.getMessage());
+            }
+            return coords + "." + fileType;
```
jacobwdv, line 74: this makes sense in the 'happy path' but if `synthesizeAndDownload` throws an exception, we still return a value? How would the caller know that this call actually failed?

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -93,20 +128,45 @@ public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLoca
...
+                futures.add(future1);
+                Future<?> future2 = submitDownloadRequest(coords, "pom", dLocation, repository);
+                futures.add(future2);
             }
+
+            while (!futures.isEmpty()) {
+                Iterator<Future<?>> iter = futures.iterator();
+                try {
+                    while (iter.hasNext()) {
+                        boolean updateProgress = false;
```
jacobwdv: I think this line can be removed

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -93,20 +128,45 @@ public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLoca
...
+                            // update progress bar, drain the downloadArtifacts total size
+                            updateProgress(individualSize);
+                            progressBar.updateMethodMap("downloadArtifacts",
+                                                        progressBar.getMethodIncrement("downloadArtifacts") - individualSize);
+                            fine("Finished downloading artifact: " + downloadedCoords);
+
+                            iter.remove();
+                        }
+                    }
+                    logger.fine("Remaining artifacts: " + futures.size());
```
jacobwdv: what is the difference between `fine` and `logger.fine` are they the same thing?

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
```diff
@@ -93,20 +128,45 @@ public void synthesizeAndDownloadFeatures(List<String> mavenCoords, String dLoca
...
+                    Thread.sleep(300);
```
jacobwdv: use a constant here for `300`.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloaderUtils.java
```diff
@@ -36,14 +43,47 @@
...
+                Future<?> future = iter.next();
+                if (future.isDone()) {
+                    future.get();
+                    iter.remove();
+                }
             }
+            logger.fine("Finding " + futures.size() + " maven artifacts.. ");
+            Thread.sleep(100);
         }
+        executor.shutdown();
```
jacobwdv, line 98: I think we want to remove this call since the `close()` method will do this for us

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -1322,32 +1324,33 @@ public File downloadSingleFeature() {
         data.put(ACTION_ERROR_MESSAGE, null);
         data.put(ACTION_EXCEPTION_STACKTRACE, null);
 
-        ArtifactDownloader artifactDownloader = new ArtifactDownloader();
-        String featureList = (String) data.get(DOWNLOAD_ARTIFACT_LIST);
-        String filetype = (String) data.get(DOWNLOAD_FILETYPE);
+        try (ArtifactDownloader artifactDownloader = new ArtifactDownloader()) {
```
jacobwdv, line 1329: Be sure to update the entire call hierarchy for this constructor to use try-with-resources. I only see the one being updated but in Eclipse I can see there are actually 4:
image

jacobwdv (changes requested):
Looks good, a few comments.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
jjiwooLim: It will not catch the exception here anymore.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
jjiwooLim: They are the same thing

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloaderUtils.java
jjiwooLim: This is a static method which will declare ExecutorService within the method (not same as ArtifactDownloader.java). I think it makes sense to shut it down here.

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
jjiwooLim: They are the same thing

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloaderUtils.java
jjiwooLim: This is a static method that declares ExecutorService within the method. I think it makes sense to shut it down here. (note: this is not the ArtifactDownloader class )

jjiwooLim:
**Testing steps:**
- Set up Nexus to host the Maven repository on remote server
- Simulate the network latency of 1000 ms
- On the local machine set a `FEATURE_REPO_URL` environment variable to point to the Nexus repository from above and run featureUtility to install 'microProfile-4.0'

Without these changes, it took about 343 ~ 411s.
| Number of threads   | Time (s) |
| ------------- | ------------- |
| 2  | 96 ~ 171  |
| 4  | 77 ~ 115  |
| 6  | 58 ~ 65  |
| 8  | 48 ~ 60  |
| 10  | 48 ~ 53  |

On my machine, the performance stopped improving with 10 threads or more. 
Therefore, the default number of threads is set to 8.
User can change the number of threads by setting the Java argument  `com.ibm.ws.install.featureUtility.artifactThreads`

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/ArtifactDownloader.java
jjiwooLim: I created a getNumThreads() method in ArtifactDownloaderUtil.java that catches the NumberFormatException

jacobwdv (approved):
LGTM! Thanks!

jjiwooLim:
#build

jdmcclur:
I did some testing with the personal build. Performance looks really good, but seeing these odd null messages when connecting to maven directly.

```
Initializing ...
Using 8 threads to download artifacts.
null cannot be connected
null cannot be connected
null cannot be connected
null cannot be connected
Resolving remote features. This process might take several minutes to complete.
Preparing assets for installation. This process might take several minutes to complete.
null cannot be connected
null cannot be connected
```

jjiwooLim:
#libbybot

jjiwooLim:
#build

jjiwooLim:
> I did some testing with the personal build. Performance looks really good, but seeing these odd null messages when connecting to maven directly.
> 
> ```
> Initializing ...
> Using 8 threads to download artifacts.
> null cannot be connected
> null cannot be connected
> null cannot be connected
> null cannot be connected
> Resolving remote features. This process might take several minutes to complete.
> Preparing assets for installation. This process might take several minutes to complete.
> null cannot be connected
> null cannot be connected
> ```

I added a null check to resolve this issue

jjiwooLim:
fat failures are not related to my changes

PR #16746, opened 2021-04-23
PR for Issue 15023: Fix WASReqURLOidc encoding problem

ayoho:
Fixes #15023

Instead of encoding the query string in the original request URL, we'll leave it as-is. That should ensure we ultimately use the "real" value of the query string. During the creation of the WASReqURLOidc cookie, the cookie value is encoded anyway to mitigate against malformed and potentially malicious cookie values. The `OidcRedirectServlet` code is also updated to use slightly different code to read the cookie so that the respective decoding is done.

Issue #15023, opened 2020-11-18
WASReqURLOidc cookie encodes the request url but doesn't decoded it upon successful redirection 


KingMackle:
**Describe the bug**
We are calling a protected URL including some query params like
 `/protected/resource/?name=aaaaaaaa%20aaaaaaa&description=bbbbbbb%20bbbbbbbbbb&type=CHECKLIST`
The server uses OIDC. The URL in the `WASReqURL` cookie appears to be incorrectly encoded:
`WASReqURLOidcn1674482248=https://server.com/protected/resource/?name=aaaaaaaa%252520aaaaaaa&description=bbbbbbb%252520bbbbbbbbbb&type=CHECKLIST; Expires=Mon, 16 Nov 2020 21:25:46 GMT; Path=/; Secure; HttpOnly`
Note that %20 become %252520

**Steps to Reproduce**
Accesse a protected resource on a WLP server that uses OIDC 
Request 1:
GET `/protected/resource/?name=aaaaaaaa%20aaaaaaa&description=bbbbbbb%20bbbbbbbbbb&type=CHECKLIST HTTP/1.1`

Response 1:
HTTP/1.1 302 Found
Location: `https://server.com/oidc/endpoint/ums/authorize?response_type=code&client_id=client&state=001605561526461ZdjsBdS3i&redirect_uri=https%3A%2F%2Fserver.com%3A9443%2Foidcclient%2Fredirect%2FumsClient&scope=openid+profile+email`
Set-Cookie: 
WASReqURLOidcn1674482248=`https://server.com/protected/resource/?name=aaaaaaaa%252520aaaaaaa&description=bbbbbbb%252520bbbbbbbbbb&type=CHECKLIST`; Expires=Mon, 16 Nov 2020 21:25:46 GMT; Path=/; Secure; HttpOnly

Request 2:
GET
`https://server.com/oidc/endpoint/ums/authorize?response_type=code&client_id=client&state=001605561526461ZdjsBdS3i&redirect_uri=https%3A%2F%2Fserver.com%3A9443%2Foidcclient%2Fredirect%2FumsClient&scope=openid+profile+email`

Response 2:
HTTP/1.1 302
Location: 
[...]

PR #16954, opened 2021-04-29
create new tests for parm handling with isClientSideRedirectSupported equals false

c00crane:
Add new tests to verify parameter and cookie handling with wasReqUrl set and isClientSideRedirectSupported=false.
Before fix #16746, parameters arrived at the servlet encoded.
The previously mentioned fix will fix this and those parameters will arrive in the original form.
This work item delivers tests to verify the proper behavior (and these tests fail without the fix)


c00crane:
#15023

c00crane:
#run-libby-bot

c00crane:
#build

ayoho (approved):

PR #16956, opened 2021-04-29
Add call to setMarkToEndOfLogs at start of config update 

c00crane:
The samesite tests use combinations of multiple settings.  Each test sets some primary config attributes and then will run a bunch of "sub" tests that reconfig other attrs and make requests.
We recently hit an issue where occasionally the wait for the "config update completed" found the message from the previous reconfig instead of the message for the just requested reconfig.  Because of this, the test made it's next request before the config that it was expecting to use was in place.  The test didn't get the behavior that it expected and the case fails.
I'm adding a call to server.setMarkToEndOfLogs() to move the mark to the end of each log.  That causes waitForConfigUpdateInLogUsingMark to search for the "config update completed" starting at point in the log right before the latest reconfig.  We'll only find the final update message.


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.oidc.client_fat,com.ibm.ws.security.social_fat.LibertyOP

ayoho (approved):

PR #17046, opened 2021-05-04
Fix for defect in RTC 283650. The test was not waiting for the config…

jvanhill:
…uration to be applied before checking for configuration error messages.


jvanhill:
#build

kristip17 (approved):

PR #17064, opened 2021-05-05
Add new featureManager attr package.server.conflict

tjwatson:
The package.server.conflict attribute is internal.  It is
intended to be used to produce the packages during the build.
In some cases the packages are built using a service
configuration that we know has conflicts on purpose.
For example, if the package is meant to include multiple
versions of a singleton feature.  The package.server.conflict
attribute is used doing the package optioneration.
When set to true the package operation will ignore
singlton rules for feature resolution while building
the server package


tjwatson:
#build

cbridgha (approved):
LGTM

tjwatson:
#build

tjwatson:
#libby

PR #17065, opened 2021-05-05
Correcting IBM JDK LoginModule behavior plus minor fixes

### LINKED ISSUE #16946: Throw LoginException when IBM JDK8 Kerberos Login Module fails to authenticate
If the IBM JDK 8 krb5LoginModule fails to authenticate it returns false instead of throwing a LoginException, which means it fails silently. We should notice when the krb5LoginModule returns false, and throw a LoginException.

mswatosh:
fixes #16946
- Catches the IBM Krb5LoginModule returning false and throws a LoginException to match the Sun Krb5LoginModule
- Adds test for Expired Credentials in the ccache
- Stops adding the invalid "oracle.net.kerberos5_mutual_authentications" property.

### DIFF
--- a/dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/Krb5LoginModuleWrapper.java
+++ b/dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/Krb5LoginModuleWrapper.java
@@ -172,18 +172,31 @@ public class Krb5LoginModuleWrapper implements LoginModule {
         krb5loginModule.initialize(subject, callbackHandler, sharedState, options);
     }
 
+    /**
+     * If the login module returns false, throws a LoginException. This is to make the IBM Krb5LoginModule behavior
+     * match the Sun Krb5LoginModule.<br><br>
+     *
+     * {@inheritDoc}
+     */
     @Override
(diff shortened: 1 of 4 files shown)


mswatosh:
#build

frowe (approved):

dev/com.ibm.ws.security.kerberos.auth/src/com/ibm/ws/security/kerberos/auth/Krb5LoginModuleWrapper.java
```diff
@@ -172,18 +172,31 @@ public void initialize(Subject subject, CallbackHandler callbackHandler, Map<Str
...
+     * If the login module returns false, throws a LoginException. This is to make the IBM Krb5LoginModule behavior
+     * match the Sun Krb5LoginModule.<br><br>
+     *
+     * {@inheritDoc}
+     */
     @Override
     public boolean login() throws LoginException {
-        krb5loginModule.login();
+        if (!krb5loginModule.login())
+            throw new LoginException("Kerberos login failed");
```
kristip17, line 184: Do we need this to be a translated message since we're throwing it?

kristip17 (approved):

mswatosh:
#run-libby-bot

PR #17081, opened 2021-05-06
Validate parsed scheme from remoteIp related headers

### LINKED ISSUE #17074: Improvement to remoteIp feature
The HTTP transport's processing of `Forwarded` and `X-Forwarded` headers, when the endpoint is configured to enable the `remoteIp` feature, can be improved. 

According to the Forwarded HTTP Extension [[RFC 7239](https://tools.ietf.org/html/rfc7239#section-5.2)], "the syntax of a `proto` value MUST conform to the URI scheme name as defined in [Section 3.1 in [RFC3986]](https://tools.ietf.org/html/rfc3986#section-3.1)". 

In turn, section 3.1 specifies that "when presented with a URI that violates one or more scheme-specific restrictions, the scheme-specific resolution process should flag the reference as an error rather than ignore the unused parts". 

When an unrecognized or malformed value is obtained from the proto value, the transport will now treat the request as malformed. This results in a `400 Bad Request` status code being set to the response, instead of the `500 Internal Server Error` status code that might occur prior to this change. Setting a status code of `400 Bad Request` on the response is more representative of the cause of the error.

 **_Note: this behavior is contingent on the endpoint being configured with `<remoteIp>` ._**

mrsaldana:
Fixes #17074

### DIFF
--- a/dev/com.ibm.ws.transport.http.remoteip_fat/fat/src/com/ibm/ws/transport/http/HttpXForwardedAndForwardedHeaderTests.java
+++ b/dev/com.ibm.ws.transport.http.remoteip_fat/fat/src/com/ibm/ws/transport/http/HttpXForwardedAndForwardedHeaderTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 IBM Corporation and others.
+ * Copyright (c) 2018, 2021 IBM Corporation and others.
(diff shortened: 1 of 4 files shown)


mrsaldana:
#build #spawn.fullfat.buckets=com.ibm.ws.transport.http.remoteip_fat

PR #17087, opened 2021-05-07
Allow deleted PIDs to be processed by the ApplicationPrereq component

brenthdaniel:
A missing PID is not necessarily an error as it may have been deleted as a result of removing an application.


brenthdaniel:
#build

jimblye (approved):

PR #17118, opened 2021-05-11
Isolate SSL testing from non SSL tests

KyleAure:
Recent tests failed with: 
```
Java.sql.SQLException: The driver could not establish a secure connection to SQL Server by using Secure Sockets Layer (SSL) encryption
Caused by: java.io.IOException: SQL Server did not return a response. The connection has been closed. ClientConnectionId:203a5f11-f3ac-49a3-8e44-4029240de139
```

This affected all datasources even those that did not enable `encrypt=true`

This is likely caused by one of two issues:
1. The database process running in the container was not ready to accept SSL connections and therefore rejected all connections
2. The JDK used on z/OS does not have TLS 1.2 enabled by default and all connections were trying to connect using TLS 1.0

I believe that (2) is the real issue here, but have tried to resolve several issues here:
1. Separate out TLS tests from other tests that do not require TLS
  - Normal tests will run against `mcr.microsoft.com/mssql/server:2019-CU10-ubuntu-16.04` base image.
  - TLS tests will run against `kyleaure/sqlserver-ssl:2019-CU10-ubuntu-16.04` custom image
2. Forces driver to use TLS 1.2 as long as JVM supports it even it it is not default.
3. Enables TLS trace (for future debugging)
4. getConnection requests with retry for better fat testing stability on slow infrastructure


jmstephensgit (approved):

KyleAure:
#build #libbybot

PR #17139, opened 2021-05-12
Fix for RTC 283750 test failure. The keystores were not previously ge…

jvanhill:
…nerated in JKS format, as intended.


jvanhill:
#build

acdemyers (approved):

PR #17141, opened 2021-05-12
Session cache npe check

randyclee:
Fixes: RTC 278870

#build


randyclee:
infrastructure server connection issues causing NPE sessions, added checks for this so that there are less cluttered errors

zhang-ca (approved):

PR #17167, opened 2021-05-14
Draft: Faces-3.0: Switch to CDIExtenionMetaData SPI 

### LINKED ISSUE #13933: Currently WebSphereCDIExtension is used by the faces-3.0 bnd.bnd investigate using CDIExtensionMetadata
We've used the WebSphereCDIExtension historically for JSF and we need to use it or something similar for faces-3.0.

We've currently ported the existing use of `com.ibm.ws.jsf.shared.ext.CdiExtension` for faces-3.0 but we should move away from it since it is now deprecated in favor of CDIExtensionMetadata.

volosied:
Fixes #13933

!!!-- DRAFT -- !!!


volosied:
Hoping this would have been simple. Doesn't actually appear to be... :(

volosied:
Closed in favor of  #17510

Issue #17163, opened 2021-05-14
Update feature in output when creating certificate using securityUtility


mtamboli:
**Describe the bug**
securityUtility lets you create a certificate and keystore. As part of the command, it also suggests to make updates to server.xml. Output of securityUtility as shown below,  suggests  `ssl-1.0` but it is better to replace with latest  `transportSecurity-1.0` feature. 
```
./bin/securityUtility createSSLCertificate --password=keypass --server=testServer1
Creating keystore /root/liberty/was/wlp/usr/servers/testServer1/resources/security/key.p12
Created SSL certificate for server testServer1. The certificate is created with CN=automst1.fyre.ibm.com,OU=testServer1,O=ibm,C=us as the SubjectDN.
Add the following lines to the server.xml to enable SSL:
    <featureManager>
        <feature>ssl-1.0</feature>
    </featureManager>
    <keyStore id="defaultKeyStore" password="{xor}NDomLz4sLA==" />
```

**Steps to Reproduce**
* Steps to reproduce the bug
* Install Open Liberty 
* Run this command `./bin/securityUtility createSSLCertificate --password=keypass --server=testServer1
Creating keystore /root/liberty/was/wlp/usr/servers/testServer1/resources/security/key.p12`

**Expected behavior**
Output should be updated with latest feature like `transportSecurity-1.0`

**Diagnostic information:**
 - OpenLiberty Version: [e.g. 19.0.0.8]
 - Java Version: [i.e. full output of `java -version`]
 - server.xml configuration (WITHOUT sensitive information like passwords)
 - If it would be useful, upload the messages.log file found in `$WLP_OUTPUT_DIR/messages.log`
 
**Additional context**

mtamboli:
I verified that this defect is fixed in latest version of Liberty. So closing it:
```
Created SSL certificate for server defaultServer. The certificate is created with CN= ...

Add the following lines to the server.xml to enable SSL:

    <featureManager>
        <feature>transportSecurity-1.0</feature>
    </featureManager>
    <keyStore id="defaultKeyStore" password="{xor}NDomLz4sLA==" />

```

PR #17170, opened 2021-05-14
Fix message return from cert utility to say transportSecurity-1.0

acdemyers:
Fix message return from the certificateUtility to say transportSecurity-1.0 instead of ssl-1.0.


acdemyers:
fixes #17163

acdemyers:
#build

emilytee (approved):

PR #17183, opened 2021-05-17
Use a longer timeout

benjamin-confino:
In https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=283911 I see the sysout `GREP MOE + " + CDI.current()` shortly after the test fails on `assertTrue("The thread with CDI.current never completed", false);`

A longer timeout should fix this and only calling `CDI.current()` once should help the second thread complete faster.

### DIFF
--- a/dev/com.ibm.ws.cdi.api_fat/fat/src/com/ibm/ws/cdi/api/fat/apps/current/CDICurrentTestServlet.java
+++ b/dev/com.ibm.ws.cdi.api_fat/fat/src/com/ibm/ws/cdi/api/fat/apps/current/CDICurrentTestServlet.java
@@ -29,7 +29,7 @@ public class CDICurrentTestServlet extends FATServlet {
     private static final long serialVersionUID = 1L;
 
     private static Boolean wasCDICurrentFound = null;
-    private static Boolean wasCDICurrentFoundViaMOE = null;
+    private static volatile Boolean wasCDICurrentFoundViaMES = null;
 
     @Resource
     ManagedExecutorService managedExecutorService;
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.cdi.api_fat/fat/src/com/ibm/ws/cdi/api/fat/apps/current/CDICurrentTestServlet.java
```diff
@@ -59,7 +59,7 @@ public void testCDICurrentViaMES() throws Exception {
 
         managedExecutorService.submit(new callCDICurrent());
 
-        int i = 10;
+        int i = 40;
```
tevans78, line 62: 1. so this change equates to going from 50ms to 200ms ... seems like a big step up but actually still 200ms isn't very long. Is that the right amount of time? Do you want to pull the actual sleep time out into a variable as well?
2. counting down rather than up is a bit odd?
3. Should `wasCDICurrentFoundViaMOE` be `wasCDICurrentFoundViaMES`?

tevans78: 4. Is `callCDICurrent` a class? Why doesn't it start with a capital?

tevans78: 5. Should `wasCDICurrentFoundViaMOE` be either volatile or access synchronized?

tevans78 (approved):

benjamin-confino:
#build

benjamin-confino:
#run-libby-bot

PR #17191, opened 2021-05-17
17150 Add endpoint icons to minified server content

### LINKED ISSUE #17150: oidc icons not being package inside the zip file for Open Liberty build
The problem is we currently have an issue with packaging due to the icons are not include in the OL zip.
For example, when we have multiple icons in the .feature file, these doesn't seem to be pick up to be include in certain .zip files. With OIDC icons they are using Subsystem-Endpoint-Icons in the `com.ibm.websphere.appserver.openidConnectServer-1.0.mf` feature file. The icons are under the `build.image/wlp/lib/features/icons/com.ibm.websphere.appserver.openidConnectServer-1.0`

I think this problem has always been there as it was not handle multiple icons. The `openliberty-dev-21.0.0.4-XXX.zip` includes the icons but not other zips like `openliberty-all-21.0.0.4-xxxxxx.zip`, Chuck Bridgham said if the zip doesn’t use feature packager then it includes the icons

image

I talked to Wendy Raschke then Ethan Fritz and he looked at the build.image/build.gradle file. Those other zips are created by this task https://github.com/OpenLiberty/open-liberty/blob/integration/dev/build.image/build.gradle#L106
Those other zips use minify to be created and `openliberty-dev-xxx.zip` is just a zip task so it includes everything in the directory structure. He pointed 
to the file under https://github.com/OpenLiberty/open-liberty/blob/2fd4a880754c37a988c5ed9ac4f1ea5988e465d6/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/PackageCommand.java

I have also talk to Dave Zavala for PackageCommand.java

dazavala:
Modify the Package command to include icon files declared within the Subsystem-Endpoint-Icons header.  

Here is an example from the openidConnectServer-1.0 feature manifest:
```
Subsystem-Endpoint-Icons: clientManagement=OSGI-INF/clientManagement_142.png,
 OSGI-INF/clientManagement_78.png; size=78,
 OSGI-INF/clientManagement_142.png; size=142,
 personalTokenManagement=OSGI-INF/personalTokenManagement_142.png,
 OSGI-INF/personalTokenManagement_78.png; size=78,
 OSGI-INF/personalTokenManagement_142.png; size=142,
 usersTokenManagement=OSGI-INF/usersTokenManagement_142.png,
 OSGI-INF/usersTokenManagement_78.png; size=78,
 OSGI-INF/usersTokenManagement_142.png; size=142
```

### DIFF
--- a/dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
+++ b/dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2015 IBM Corporation and others.
+ * Copyright (c) 2011, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -77,7 +77,7 @@ public class SubsystemFeatureDefinitionImpl implements ProvisioningFeatureDefini
      * Create a new subsystem definition with the specified immutable attributes.
      * Called when rebuilding from a cache.
      *
-     * @param attr Immutable attributes
+     * @param attr    Immutable attributes
      * @param details Provisioning details (will be cleared when provisioning operation is complete)
      * @see #load(String, SubsystemFeatureDefinitionImpl)
      * @see FeatureDefinitionUtils#loadAttributes(String, ImmutableAttributes)
@@ -99,8 +99,8 @@ public class SubsystemFeatureDefinitionImpl implements ProvisioningFeatureDefini
      * <p>
      * Some operations, like finding resource bundles, may not work.
      *
-     * @param repoType emtpy/null for core, "usr" for user extension, or the product
-     *            extension name
+     * @param repoType    emtpy/null for core, "usr" for user extension, or the product
+     *                        extension name
      * @param inputStream The input stream to read from
      * @see ExtensionConstants#CORE_EXTENSION
      * @see ExtensionConstants#USER_EXTENSION
@@ -121,8 +121,8 @@ public class SubsystemFeatureDefinitionImpl implements ProvisioningFeatureDefini
      * specified input stream.
      *
      * @param repoType emtpy/null for core, "usr" for user extension, or the product
-     *            extension name
-     * @param file Subsystem feature definition manifest file
+     *                     extension name
+     * @param file     Subsystem feature definition manifest file
      *
      * @see ExtensionConstants#CORE_EXTENSION
      * @see ExtensionConstants#USER_EXTENSION
@@ -467,21 +467,38 @@ public class SubsystemFeatureDefinitionImpl implements ProvisioningFeatureDefini
     /*
      * (non-Javadoc)
      *
-     * @see com.ibm.ws.kernel.feature.provisioning.ProvisioningFeatureDefinition#getIconFiles()
+     * @see com.ibm.ws.kernel.feature.provisioning.ProvisioningFeatureDefinition#getIcons()
      */
     @Override
     public Collection<String> getIcons() {
-        Collection<String> icons = new ArrayList<String>();
+        Collection<String> result = new ArrayList<String>();
         String iconHeader = getHeader("Subsystem-Icon");
         if (iconHeader != null) {
-            String[] iconHeaders = iconHeader.split(",");
-            for (String iconDirective : iconHeaders) {
-                String[] iconDirectives = iconDirective.split(";");
-                // Assume the first part is always the file URL
-                icons.add(iconDirectives[0].trim());
+            String[] icons = iconHeader.split(",");
+            for (String icon : icons) {
+                String[] iconAttrs = icon.split(";");
+                // icon has form "<iconUrl>[; size=n]"
+                result.add(iconAttrs[0].trim());
             }
         }
-        return icons;
+        String epIconsHeader = getHeader("Subsystem-Endpoint-Icons");
+        if (epIconsHeader != null) {
+            String epNamePath = "";
+            String[] epIcons = epIconsHeader.split(",");
+            for (String epIcon : epIcons) {
+                String[] epIconAttrs = epIcon.split(";");
+                if (epIconAttrs[0].indexOf("=") >= 0) {
+                    // epIcon has form "<epName>=<epIconUrl>[; size=N]
+                    String[] epNameAndIconUrl = epIconAttrs[0].split("=");
+                    epNamePath = epNameAndIconUrl[0].trim() + "/";
+                    result.add(epNamePath + epNameAndIconUrl[1].trim());
+                } else {
+                    // epIcon has form "<epIconUrl>[; size=n]"
+                    result.add(epNamePath + epIconAttrs[0].trim());
+                }
+            }
+        }
+        return result;
     }
 
     /*
--- a/dev/com.ibm.ws.kernel.boot_fat/bnd.bnd
+++ b/dev/com.ibm.ws.kernel.boot_fat/bnd.bnd
@@ -1,5 +1,5 @@
 #*******************************************************************************
-# Copyright (c) 2017 IBM Corporation and others.
+# Copyright (c) 2017, 2021 IBM Corporation and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License v1.0
 # which accompanies this distribution, and is available at
@@ -21,7 +21,8 @@ src: \
 fat.project: true
 test.classpath.wlp.include: api/spec/*.jar,api/ibm/*.jar
 
-tested.features: mpconfig-1.3, concurrent-1.0, mpfaulttolerance-1.1, mpmetrics-1.1
+tested.features: mpconfig-1.3, concurrent-1.0, mpfaulttolerance-1.1, mpmetrics-1.1, restconnector-2.0,\
+   federatedregistry-1.0, oauth-2.0, openidconnectserver-1.0, admincenter-1.0, ldapregistry-3.0
 
 -buildpath: \
 	com.ibm.websphere.javaee.annotation.1.1;version=latest,\
--- a/dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/internal/commands/PackageCommandTest.java
+++ b/dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/internal/commands/PackageCommandTest.java
@@ -567,6 +567,65 @@ public class PackageCommandTest {
 
     }
 
+    private static final List<String> IconFeatures;
+    static {
+        IconFeatures = new ArrayList<String>(2);
+        IconFeatures.add("openidConnectServer-1.0");
+        IconFeatures.add("adminCenter-1.0");
+    }
+
+    /**
+     * Verify minified server content includes all types of icons used by admin features.
+     */
+    @Test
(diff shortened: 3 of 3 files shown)


dazavala:
#build

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
```diff
@@ -484,6 +484,33 @@ public String toString() {
...
     }
 
+    /*
+     * (non-Javadoc)
+     *
+     * @see com.ibm.ws.kernel.feature.provisioning.ProvisioningFeatureDefinition#getEndpointIcons()
+     */
+    @Override
+    public Collection<String> getEndpointIcons() {
+        Collection<String> icons = new ArrayList<String>();
```
tjwatson: Because this is rare, this ArrayList should only be constructed if endpointIconsHeader  is non-null.  So assign `icons` to Collections.emptyList here and then construct the ArrayList inside the if NOT null.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
```diff
@@ -484,6 +484,33 @@ public String toString() {
...
+            for (String endpointIcon : endpointIcons) {
+                String[] endpointIconDirectives = endpointIcon.split(";");
+                if (endpointIconDirectives.length <= 1 && endpointIcon.indexOf("=") >= 0) {
+                    String[] endpointIconSplit = endpointIcon.split("=");
+                    endpointIconCategory = endpointIconSplit[0].trim();
+                    icons.add(endpointIconCategory + "/" + endpointIconSplit[1].trim());
+                    continue;
+                }
+                // Assume the first part is always the file URL
+                icons.add(endpointIconCategory + "/" + endpointIconDirectives[0].trim());
```
tjwatson: The example I see for Subsystem-Endpoint-Icons is this:
```
Subsystem-Endpoint-Icons: \
 clientManagement=OSGI-INF/clientManagement_142.png,
   OSGI-INF/clientManagement_78.png;size=78,
   OSGI-INF/clientManagement_142.png;size=142
```
So the category ends up being `clientManagement`.  We then add the path based of the category.  Will `clientManagement/OSGI-INF/clientManagement_142.png` be added twice to the result?

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
```diff
@@ -484,6 +484,33 @@ public String toString() {
...
+     */
+    @Override
+    public Collection<String> getEndpointIcons() {
+        Collection<String> icons = new ArrayList<String>();
+        String endpointIconsHeader = getHeader("Subsystem-Endpoint-Icons");
+        if (endpointIconsHeader != null) {
+            String[] endpointIcons = endpointIconsHeader.split(",");
+            String endpointIconCategory = "";
+            for (String endpointIcon : endpointIcons) {
+                String[] endpointIconDirectives = endpointIcon.split(";");
```
tjwatson: These are really attributes if we are sticking to OSGi manifest header terms.  Directives are specified with `:=` where attributes just use the single char `=` for assignment.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
```diff
@@ -484,6 +484,33 @@ public String toString() {
...
+                if (endpointIconDirectives.length <= 1 && endpointIcon.indexOf("=") >= 0) {
```
tjwatson: Would it be better to assume any first element of a semi-colon `;` separated list which contains an `=` specifies the category?

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
dazavala: Yes.  The redundancy is similar to what we see when processing the Subsystem-Icon header.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
dazavala: I actually changed the variable name from xxxAttrs to xxxDirectives for consistency with the code that processes the Subsystem-Icon header.  Easy change to make in both places.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
dazavala: Sure.  I had considered doing the same earlier, and using an if-else to better indicate which part of the icon metadata is processed.  Also, I figured out that "category" is really the "endpoint-name".  You'll that change as well.

dazavala:
#build

dazavala:
@aknguyen7 and @lavenal: While you review the changes please advise on the following...

First, the solution assumes endpoint icon metadata has the strict form:
 
` Subsystem-Endpoint-Icons: <endpointName>=<iconFile>[, <iconFile>[; size=N] ]*`

Is the assumption correct?  If not, then this code will fail whenever the first comma-delimited element contains an attribute.  For instance, the following endpoint icon metadata element, which contains a size attribute, can never exist.  If it can, then I need to revise the solution.
 
` Subsystem-Endpoint-Icons: clientManagement=OSGI-INF/clientManagement_142.png; size=142` 

Second, does the solution reflect the semantic intent of the endpoint icon metadata?

lavenal:
Thanks @dazavala for supporting the scenario. For the answers to your questions:

- the assumption on the first element of the Subsystem-Endpoint-Icons is correct. It is exactly the same format used in Subsystem-Icon listing the first element without the size attribute.
-  I don't know the answer for your second question yet as I've to go back to refresh myself with the existing admin center codes. I'll get back to you later with the answer.

tjwatson:
> Thanks @dazavala for supporting the scenario. For the answers to your questions:
> 
> * the assumption on the first element of the Subsystem-Endpoint-Icons is correct. It is exactly the same format used in Subsystem-Icon listing the first element without the size attribute.

Are these headers externals?  Does Liberty support product extensions having their own icons for admin center to use?  If the format is indeed the same then I think the parsing code used in minify needs to be updated for Subsystem-Icon, correct?  I would also be interested in seeing the documentation (if this is an external).  It is hard to imagine how the doc would adequately explain what is going on with the semantics of this syntax.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
```diff
@@ -467,21 +467,38 @@ public String toString() {
...
+            for (String epIcon : epIcons) {
+                String[] epIconAttrs = epIcon.split(";");
+                if (epIconAttrs[0].indexOf("=") >= 0) {
+                    // epIcon has form "<endpointName>=<iconFile>
+                    String[] epNameAndIconFile = epIcon.split("=");
+                    epName = epNameAndIconFile[0].trim();
+                    result.add(epName + "/" + epNameAndIconFile[1].trim());
+                } else {
+                    // epIcon has form "<iconFile>[; size=n]"
+                    result.add(epName + "/" + epIconAttrs[0].trim());
```
tjwatson: What if epName is never set (remains "")?  This will add a path with a proceeding "/".  Will that proceeding "/" cause issues?  Perhaps add the path separator  in the `if (epIconAttrs[0].indexOf("=") >= 0) {` branch:
```
    epName =  epNameAndIconFile[0].trim() + "/";
    result.add(epName + epNameAndIconFile[1].trim());
} else {
    // epIcon has form "<iconFile>[; size=n]"
    result.add(epName + epIconAttrs[0].trim());
```

lavenal:
@tjwatson, these headers are not external. They're created to support OIDC. I don't believe that admin center supports product extensions with their own icons at the moment. 

I may have misspoken when I said the format is the same as Subsystem-icon. What I mean is the value of the icon specified is the same format on both. Here's an example of Subsystem-Icon:

Subsystem-Icon: OSGI-INF/deploy_142x142.png,OSGI-INF/deploy_78x78.png;size=78,OSGI-INF/deploy_142x142.png;size=142

 Here's an example for clientManagement of Subsystem-Endpoint-icons:

clientManagement=OSGI-INF/clientManagement_142.png,
 OSGI-INF/clientManagement_78.png; size=78,
 OSGI-INF/clientManagement_142.png; size=142,

Sorry for causing the confusion.

lavenal:
@dazavala I reviewed the admin center codes to support OIDC and downloaded openliberty-all-21.0.0.6-202105171428.zip from the earlier completed PB for this PR. After updating a newly created server with OIDC configuration in it and including adminCenter-1.0 feature, the OIDC icons were displayed instead of the generic icon when bringing up the admin center. Thank you so much for the work to support the multiple icons scenario.

dazavala:
#build 
Review changes.  Thanks all.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
dazavala: We merged getEndpointIcons() behavior into getIcons(), which obviates this change.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
dazavala: Changes completed.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
dazavala: Changes completed.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/SubsystemFeatureDefinitionImpl.java
dazavala: Changes completed.

dazavala:
@lavenal Thank you for verifying these changes resolve the admin center issue.

aknguyen7:
the wlp-base-all-21.0.0.6.jar from Dave's PB build now also included the icons

tjwatson (approved):

tjwatson:
#libby

dazavala:
Merging now.  The PackageCommandTest passes, and the open-liberty image from the latest PB contains the endpoint icons: 

image

PR #17194, opened 2021-05-17
Use Logger to allow expected warnings to be suppressed

### LINKED ISSUE #17108: Trace specification does not correctly for jakarta 9 feature io.openliberty.org.jboss.resteasy.validator.provider
I'm executing the SPECjEnterprise2018 benchmark which contains bean validation tests that are expected to fail the validation. I want to suppress the warnings from org.jboss.resteasy.plugins.validation.GeneralValidatorImpl that are printed to the messages.log since they are expected. I've tried many different additions to the traceSpecification, including org.jboss.*=error, org.jboss.resteasy.plugins.validation.GeneralValidatorImpl=error, org.jboss.*=all=disabled, etc. But I still get the warnings in my log like this:
[5/10/21 13:10:30:694 CDT] 000000d9 org.jboss.resteasy.plugins.validation.GeneralValidatorImpl   W register.arg0.address.phone: numeric value out of bounds (<12 digits>.<0 digits> expected), register.arg0.address.phone: size must be between 8 and 12, register.arg0.address.city: size must be between 1 and 30

This is a relatively new feature that is part of the servlet-5.0 new jaxrs implementation (resteasy).

dettmoney:
There's a class from resteasy pulled into open liberty (org.jboss.resteasy.plugins.validation.GeneralValidatorImpl) that currently uses Tr to output warnings due to failed validations. The only way to suppress these warning from showing up in the messages.log is to use java.util.Logger. When running the SPECjEnterprise benchmark, there are expected bean validation failures that result in the logs filling up with these warning. Tr doesn't allow suppression of warnings to messages.log.

### DIFF
--- a/dev/io.openliberty.org.jboss.resteasy.validator.provider/src/org/jboss/resteasy/plugins/validation/GeneralValidatorImpl.java
+++ b/dev/io.openliberty.org.jboss.resteasy.validator.provider/src/org/jboss/resteasy/plugins/validation/GeneralValidatorImpl.java
@@ -26,6 +26,7 @@ import javax.validation.ValidatorFactory;
 import javax.validation.executable.ExecutableType;
 import javax.validation.executable.ValidateOnExecution;
 
+import org.jboss.logging.Logger;
 import org.jboss.resteasy.api.validation.ConstraintType;
 import org.jboss.resteasy.api.validation.ConstraintType.Type;
 import org.jboss.resteasy.api.validation.ResteasyConstraintViolation;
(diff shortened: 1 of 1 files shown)


dettmoney:
#build

jim-krueger (commented):
I don't believe this change is needed:  See hideMessage in the server config.  https://openliberty.io/docs/21.0.0.4/log-trace-configuration.html

dettmoney:
#build

jim-krueger (approved):

dettmoney:
#libbybot

PR #17207, opened 2021-05-18
omit unused Informix property

njr-11:
Omit informixLockModeWait from data store helper properties.  Liberty doesn't have a property by that name, and the data store helper doesn't need it because existing vendor property ifxIFX_LOCK_MODE_WAIT already covers the function.


njr-11:
#build

nmittles (approved):

PR #17211, opened 2021-05-18
Only run original and latest MP versions in LITE mode; all else in FULL

andymc12:
Trim down FAT run time for `com.ibm.ws.microprofile.rest.client_fat` by only running the original version of the test and the latest version (currently 2.0) in LITE mode.  All other versions will now only run in FULL mode.


andymc12:
#build

WhiteCat22 (approved):

PR #17217, opened 2021-05-18
Fix ConcurrentModificationException in ConfigImpl when iterating over System.getProperties().keySet()

jkoehler22:

### DIFF
--- a/dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/config/ConfigImpl.java
+++ b/dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/config/ConfigImpl.java
@@ -14,6 +14,7 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


jkoehler22:
#build

jkoehler22:
Note here that getProperties is not a copy.  So coping the Set should avoid the ConcurrentModificationException

```
/**
 * Answers the system properties. Note that this is
 * not a copy, so that changes made to the returned
 * Properties object will be reflected in subsequent
 * calls to getProperty and getProperties.
 * <p>
 * Security managers should restrict access to this
 * API if possible.
 *
 * @return		the system properties
 */
public static Properties getProperties() {
	if (!propertiesInitialized) throw new Error("bootstrap error, system property access before init"); //$NON-NLS-1$
	SecurityManager security = System.getSecurityManager();
	if (security != null)
		security.checkPropertiesAccess();
	
	return systemProperties;
}
```

jkoehler22:
The `SystemWAB_OneMember_AfterServerStart` failure is a known problem.

jim-krueger (approved):

PR #17259, opened 2021-05-20
SQLServer reset transaction timeout

KyleAure:
The test `testTransactionTimeout` sets a timeout of 8 seconds. 
However, this timeout is set on a UserTransaction and therefore will affect all subsequent tests. 
It is possible on a slow system that one of the other tests could throw a Rollback Exception if this timeout is reached due to a slow test. 

Therefore, re-set this timeout to 0 (disabled) after `testTransactionTimeout` is finished.


KyleAure:
#build

jmstephensgit (approved):

KyleAure:
#libbybot

PR #17295, opened 2021-05-25
better align the JDBC heritage interop test bucket

njr-11:
Updates the JDBC heritage interoperability test bucket to more closely match how heritage API will actually be added by a bundle fragment for the API and normal bundle for the metatype.


njr-11:
#run-libby-bot
#build

nmittles (approved):

PR #17307, opened 2021-05-26
Skip SQLServer SSL Secure test on IBM JDK

KyleAure:
Previously noticed that this test was failing on z/OS running IBM JDK.  
Can now confirm that this will fail on all OS's running IBM JDK due to: 

The Keystore used for this test is PKCS12 formatted and was created using openjdk.
The IBM JDK and will fail with `java.io.IOException: Invalid keystore format` since the Keystore provider is SUN instead of IBMJCE.

Skipping this tests on all systems when IBM JDK is used.


KyleAure:
#build

mswatosh (approved):

PR #17310, opened 2021-05-26
Prevent wsAtomicTransaction-1.2 from activating the jaxrs transport bundle

neuwerk:
When wsAtomicTransaction-1.2 and jaxrs-2.1 are started in the same enviornment, the CXFActivator class starts both versions of the transport layer in CXF.


neuwerk:
#build

BerksanAtes (approved):
Look good

neuwerk:
#build

PR #17315, opened 2021-05-26
http2_fat secure h2 client is not closing connections correctly

wtlucy:
Our `SecureHttp2Client` needs to invoke a blocking `close()` on its `HttpAsyncRequester`. The client is currently using the non-blocking shutdown request, and on very slow test systems it's not waiting long enough, which causes errors in subsequent tests.


wtlucy:
#build

loriadi (approved):
Hoping this works!

PR #17330, opened 2021-05-27
Access permit

cbridgha:


Thihup:
The --illegal-access flag will be a Noop in JDK 17, is this a good idea to add it to the java9.options?
https://openjdk.java.net/jeps/403

PR #17339, opened 2021-05-27
Fix defect from RTC 283928. Skip the test if running on Windows with …

jvanhill:
…EE9 transform action as the transform can cause file handle issues on windows.


jvanhill:
#build

ryanesch (approved):

PR #17340, opened 2021-05-28
Update SessionCacheTwoServerTimeoutTest cases

zhang-ca:
Fixing session test case issues:
sessionDestroyed() method only called on the server who invalidated the session.
There is no guarantee server A will timeout first.


zhang-ca:
#build

randyclee (approved):
LGTM

PR #17342, opened 2021-05-28
Include keystore for test case use instead of auto generating keystore

c00crane:
Occasionally, we're hitting an issue (when we creating a new server for basically every test case (1 test/test class)) where the keystore creation fails.  In automation, we can't really retry, so, we're going to try using a test provided keystore file.

I created a key/trust store pair and replaced the default keystore in server.xml with config that points to the new files.


c00crane:
#build

ayoho (approved):

PR #17363, opened 2021-06-01
Issue 16587: Update OIDC isClientSideRedirectSupported description

### LINKED ISSUE #16587: Update OIDC client isClientSideRedirectSupported description TS005403444 TS005562950
Update the description of **isClientSideRedirectSupported** to include what it does (use JavaScript) and why you may need to use it (to preserve request fragments).  

Many people do not understand what the property means and recommendations are being given to use this property for things for which it will have no affect.

Here are some examples that can be used to craft a new description:

The **useJavaScript** property for OIDC in tWAS:
_Set this property to false if you do not want to use JavaScript when you redirect to the OP for the initial authentication request. When you do not use JavaScript, any URI fragments that are present on the original inbound request are lost._

The backwards (**redirectToIdPonServerSide**) property for SAML SSO on tWAS:
_This property is used to indicate that the TAI should redirect to the IdP itself. Set this property to false to do a client-side redirect. Set this property to false if URL fragments are being lost on the redirect; the TAI will then do a client-side redirect._

ayoho:
Resolves #16587

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.client/resources/OSGI-INF/l10n/metatype.properties
+++ b/dev/com.ibm.ws.security.openidconnect.client/resources/OSGI-INF/l10n/metatype.properties
@@ -129,7 +129,7 @@ createSession=Create session
 createSession.desc=Specifies whether to create an HttpSession if the current HttpSession does not exist.
 
 isClientSideRedirectSupported=Client side redirect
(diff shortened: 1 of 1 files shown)


ayoho:
#build

barbj (approved):

skasund:

PR #17416, opened 2021-06-03
Changes to support launching from server script with checkpoint

sebratton:

### DIFF
--- a/dev/gradle.properties
+++ b/dev/gradle.properties
@@ -33,7 +33,7 @@ bnd_preCompileRefresh=false
 # By default Gradle will reserve 1GB of heap space.
 # Very large builds might need more memory to hold Gradle’s model and caches.
 # Set file encoding to override the system encoding.
-org.gradle.jvmargs=-Xmx2560M -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx4G -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
 
 # Fix for the TLS protocol version used to communicate with Maven Central when using the IBM JVM.
 systemProp.com.ibm.jsse2.overrideDefaultTLS=true
--- a/dev/gradle.properties
+++ b/dev/gradle.properties
@@ -33,7 +33,7 @@ bnd_preCompileRefresh=false
 # By default Gradle will reserve 1GB of heap space.
 # Very large builds might need more memory to hold Gradle’s model and caches.
 # Set file encoding to override the system encoding.
-org.gradle.jvmargs=-Xmx4G -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2560M -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
 
 # Fix for the TLS protocol version used to communicate with Maven Central when using the IBM JVM.
 systemProp.com.ibm.jsse2.overrideDefaultTLS=true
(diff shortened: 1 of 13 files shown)


dev/gradle.properties
```diff
@@ -33,7 +33,7 @@ bnd_preCompileRefresh=false
 # By default Gradle will reserve 1GB of heap space.
 # Very large builds might need more memory to hold Gradle’s model and caches.
 # Set file encoding to override the system encoding.
-org.gradle.jvmargs=-Xmx2560M -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx4G -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
tjwatson: Should we really do this?

dev/io.openliberty.checkpoint/bnd.bnd
```diff
@@ -17,7 +17,7 @@ Bundle-Description: Open Liberty Checkoint Support, version ${bVersion}
 Bundle-SymbolicName: io.openliberty.checkpoint
 Private-Package: io.openliberty.checkpoint.internal
 Export-Package: io.openliberty.checkpoint.internal.criu
-Import-Package: org.apache.felix.service.command; resolution:=optional, *
+Import-Package: org.apache.felix.service.command;version="[1.0,2)";resolution:=optional,*
```
tjwatson: Is the version not filled in by bnd during the build?

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -151,4 +172,9 @@ private void restore(Phase phase, List<SnapshotHook> snapshotHooks) throws Snaps
     private static SnapshotFailed failedRestore(Exception cause) {
         return new SnapshotFailed(Type.RESTORE_ABORT, "Failed to restore from snapshot.", cause);
     }
+
+    public File getSnapshotDir() {
+        return new File(cc.getBundleContext().getDataFile(""), SNAPSHOT_DIR_NAME);
```
tjwatson: How do you find this later during restore?

dev/io.openliberty.checkpoint/bnd.bnd
sebratton: Unintentionally added to commit from a temp fix I made before you fixed the issue yourself. I will remove.

dev/gradle.properties
sebratton: Thanks, I meant to take that out.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointApplications.java
```diff
@@ -33,14 +31,14 @@
...
+        this.doCheckpoint = "applications".equals(bc.getProperty(Checkpoint.CHECKPOINT_PROPERTY_NAME));
     }
 
     @Override
     public void check() {
         if (doCheckpoint) {
             try {
-                checkpoint.snapshot(Phase.APPLICATIONS, new File("snapshot"));
+                checkpoint.snapshot(Phase.APPLICATIONS);
             } catch (SnapshotFailed e) {
```
tjwatson, line 42: Not for this PR. But we need an issue to track what should happen when a snapshot fails on the prepare side vs the restore side.  On the prepare side it will likely be during a container image build, we need to ensure an image build can fail when this happens.  On the restore side It may be sufficient to FFDC here, but I suspect we will also want some translated messages for the logs on both prepare and restore here.  We should probably create a utility that can do the exception handling for both phases (features and applications).

tjwatson (approved):

PR #17420, opened 2021-06-04
Refactor JPA Relationship Buckets into multiple subprojects targeting…

jgrassel:
Refactor the JPA Relationship buckets into individual bucket projects that test only one JPA feature level at a time.  This is to reduce the time it takes to run each FAT on OEM runs involving remote databases, which can and do exceed the allowed execution time due to network latency.

This should be done with all multi-database JPA FATs, the relationship buckets are the first to be reworked to use this approach as they have been the most problematic recently.  Buckets which are not multi-db (like the injection tests) shouldn't need this change.


jgrassel:
#build

jgrassel:
The build looks good, all of the relationship tests passed.  The build's attempt to run the common tset projects is sort of a behavior of libbybot with its scanning of the file changes in the PR -- it should not result in any problems in a regular build.

jgrassel:
#run-libby-bot

dazey3 (approved):
LGTM

PR #17455, opened 2021-06-08
Break JPA Callback, EntityManager, Inheritance, and Query fats to ind…

jgrassel:
SImilar to the JPA relationship buckets, breaking out the callback, entitymanager, inheritance, and query fat buckets to individual spec level test projects.


jgrassel:
#build

dazey3 (approved):
LGTM

Issue #17489, opened 2021-06-09
IllegalStateException is thrown when Liberty tries to update a readOnly subject


una-tapa:
**Describe the bug**

The `java.lang.IllegalStateException` is thrown with the reason `Subject is read-only`.  The error is thrown when TAI authenticator tries to check and remove an internal property that is no longer used.   

```
Caused by: java.lang.IllegalStateException: Subject is read-only
        at javax.security.auth.Subject$SecureSet$1.remove(Subject.java:1297)
        at javax.security.auth.Subject$SecureSet.remove(Subject.java:1386)
        at java.util.Collections$SynchronizedCollection.remove(Collections.java:2051)
        at com.ibm.ws.webcontainer.security.internal.TAIAuthenticator.removeInternalProps(TAIAuthenticator.java:388)
        at com.ibm.ws.webcontainer.security.internal.TAIAuthenticator.authenticateWithSubject(TAIAuthenticator.java:374)
        at com.ibm.ws.webcontainer.security.internal.TAIAuthenticator.loginWithTAIUserName(TAIAuthenticator.java:351)
        at com.ibm.ws.webcontainer.security.internal.TAIAuthenticator.authenticateWithTAIResult(TAIAuthenticator.java:313)
        at com.ibm.ws.webcontainer.security.internal.TAIAuthenticator.handleTaiResult(TAIAuthenticator.java:252)
        at com.ibm.ws.webcontainer.security.internal.TAIAuthenticator.authenticate(TAIAuthenticator.java:167)
        at com.ibm.ws.webcontainer.security.WebProviderAuthenticatorProxy.handleTAI(WebProviderAuthenticatorProxy.java:337)
        at com.ibm.ws.webcontainer.security.WebProviderAuthenticatorProxy.authenticate(WebProviderAuthenticatorProxy.java:454)
        at com.ibm.ws.webcontainer.security.WebAuthenticatorProxy.authenticate(WebAuthenticatorProxy.java:68)
        at com.ibm.ws.webcontainer.security.WebAppSecurityCollaboratorImpl.authenticateRequest(WebAppSecurityCollaboratorImpl.java:1233)
        at com.ibm.ws.webcontainer.security.WebAppSecurityCollaboratorImpl.determineWebReply(WebAppSecurityCollaboratorImpl.java:989)
[...]

PR #17535, opened 2021-06-14
Change to avoid removing property when the Subject is already readOnly

una-tapa:
This PR is to address issue #17489


una-tapa:
#build

una-tapa:
Closing because of dup of #17586

Issue #16317, opened 2021-03-22
Bell service implementations cannot load SPI


tjwatson:
A bell in liberty is configured with a reference to a shared library configuration that provides Java SE service loader meta-data (META-INF/services) to specify implementations of services that can be used to extend the functionality of Liberty.  The class loader that is used to load shared libraries is only allowed to access API packages that are exposed by the features enabled with the server configuration.

This prevents a bell from implementing service interfaces which are included in SPI packages in Liberty.  There are many scenarios where a Liberty feature has SPI to allow product extensions to extend the functionality of Liberty through the implementation of a service interface included in one of the Liberty SPI packages.  Because SPI are only available to product extensions this forces developers to use their own product extension to register an OSGi service that implements the SPI service interface.

See thread https://groups.io/g/openliberty/topic/81482285#646

This design issue is to discuss the options for enabling shared libraries that are configured for a bell have access to SPI packages.  Shared libraries can be configured for application class loader delegation.  This enhancement for bells must not expose application class loaders to SPI packages.

Possible solutions
1. Add a boolean attribute to `<library/>` configuration to enable access to SPI called `spiVisibilityEnabled`, default false
1. Add a string attribute to `<library/>` configuration to specify SPI types visible called `spiTypeVisibility`, default to empty.  Only supported type is `ibm-spi`. Currently Liberty does not differentiate types of SPI, a package is either SPI or not, but this option would allow us to add different types later.
1. Others

Regardless of configuration, the solution must ensure that delegation to common shared libraries does not expose the application class loader to more packages than configured with the application class loader.  The application class loader must never have access to SPI packages.  A quick look at the application class loader delegation implementation seems to allow common shared library package visibility to be different from the application class loader package visibility setting already.

NottyCode:
Another possible implementation would be to allow someone to configure via the existing apiTypeVisibility:

```
  <library apiTypeVisibility="api" />
```

which was suggested in the email thread. My concern with this is if we do that and do not open this up to applications an application could never reference the library which would make it impossible for the bell to provide an application api (the reason for this is that I think the library and application class loader apiTypeVisbility needs to match). I'm also not keen on exposing SPIs to applications. However perhaps since this isn't default I'm being overly paranoid and we shouldn't worry about it.

I do think that bells is a simpler way to extend the runtime than expecting people to write OSGi features so I'm in favour of increasing the applicability of it.

tjwatson:
[...]

PR #17580, opened 2021-06-17
[DRAFT] Support library SPI visibility for BELLs

dazavala:
For issue #16317

This PR adds beta-edition SPI type visibility support to libraries that are referenced by BELL services and to liberty service components that use library classloaders.  The support includes changes to the library configuration, the classloading service, and the kernel's region digraph.  

Updates to the classloading.bells_fat verify three new use-cases.
- BELL services can see SPI packages according to the new library spiTypeVisibility configuration option.
- Applications do not see SPI packages when sharing a library configured with spiTypeVisibility.
- Library SPI type visibility enables within beta-edition servers, only.  

Updates to the classloading_test (unittest) verify behaviors of the new SpiType.


dev/com.ibm.ws.classloading.bells_fat/test-applications/testSpiTypeVisible.jar/src/com/ibm/ws/test/SpiTypeVisible.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class SpiTypeVisible implements TestInterface {
+
+    @Override
+    public String isThere(String name) {
+        final String spiClassName = "com.ibm.wsspi.rest.handler.RESTHandler";
+        Class spiClass = null;
+        try {
+            // Try current classloader
+            ClassLoader libCl = this.getClass().getClassLoader();
```
tjwatson: Why are you not just using Class.forName here instead of getting the class loader and explicitly calling `loadClass` on it?

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/GatewayBundleFactory.java
```diff
@@ -153,16 +154,16 @@ private Bundle createGatewayBundle(GatewayConfiguration gwConfig, ClassLoaderCon
                         .requireBundles(gwConfig.getRequireBundle())
                         .dynamicallyImportPackages(gwConfig.getDynamicImportPackage())
                         .addAttributeValues(GATEWAY_BUNDLE_MARKER, "true")
-                        .addManifestAttribute(MANIFEST_GATEWAY_ALLOWEDTYPES_PROPERTY_KEY, gwConfig.getApiTypeVisibility())
+                        .addManifestAttribute(MANIFEST_GATEWAY_ALLOWEDTYPES_PROPERTY_KEY, gwConfig.getApiTypeVisibility())  // WOT DIS?
```
tjwatson: I believe this is just an informational header describing what APIs the gateway bundle has access to.  It has no purpose at runtime as far as I know.

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/GatewayBundleFactory.java
```diff
@@ -227,6 +231,12 @@ private void connectToApiRegions(Region region, EnumSet<ApiType> apiTypeVisibili
             Region apiRegion = copy.getRegion(apiRegionType.getRegionName());
             region.connectRegion(apiRegion, allBuilder.build());
         }
+        if (spiTypeVisibility) {
+            SpiRegion spiRegionType = SpiRegion.valueFromSpiType("ibm-spi");
+            connectToInternal |= spiRegionType.delegateInternal();
```
tjwatson: I have my doubts that we need to be concerned about connectToInternal for the SPI region connection.

dev/com.ibm.ws.classloading/src/com/ibm/ws/library/internal/LibraryGeneration.java
```diff
@@ -88,6 +90,7 @@
         this.files = files;
         this.folders = folders;
         this.apiTypeVisibility = apiTypeVisibility;
+        this.isIbmSpiVisible = "ibm-spi".equalsIgnoreCase((String)props.get("spiTypeVisibility"));
```
tjwatson: Being OpenLiberty, I'm not sure we should use `ibm` in the value here.

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/classloading/GatewayConfiguration.java
```diff
@@ -89,6 +89,13 @@
      */
     GatewayConfiguration setApiTypeVisibility(Iterable<ApiType> types);
 
+    /**
+     * Sets the allowed SPI type to "ibm-spi".  Once set, it cannot be unset
```
tjwatson: This javadoc is confusing.  We are not setting the type to `ibm-spi`.  We are just enabling with a boolean visibility to SPI.

dev/com.ibm.ws.classloading/src/com/ibm/ws/library/internal/LibraryGeneration.java
```diff
@@ -44,6 +44,8 @@
     private volatile boolean cancelled;
     private final SharedLibraryImpl library;
 
+    private final boolean isIbmSpiVisible;  // true iff spiTypeVsibility="ibm-spi" and library instance is referenced by Bell.
```
tjwatson: This only works if the library is referenced by a Bell?  I would think a library could set `spiTypeVsibility="ibm-spi"` independently of a `<bell/>` configuration.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/SpiRegion.java
```diff
@@ -0,0 +1,173 @@
...
+ * by the liberty features. The {@link #THREAD_CONTEXT} type is used to grant the
+ * thread context class loaders access to packages exported with the matching
+ * attribute thread-context=true.
+ */
+public enum SpiRegion {
+
+    /**
+     * IBM SPI, delegates to internal
+     */
+    IBM("ibm-spi", "liberty.ibm.spi", true);
```
tjwatson: As mentioned earlier.  I think we want to avoid adding more references to IBM in open liberty code.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/SpiRegion.java
```diff
@@ -0,0 +1,173 @@
...
+        return regionName;
+    }
+
+    /**
+     * Returns true if the region delegates to the internal region
+     * to grant access to internal packages
+     *
+     * @return true if delegation to the internal region.
+     */
+    public boolean delegateInternal() {
```
tjwatson: I'm not sure if we need this delegateInternal stuff for SPI.  On the other hand it would be consistent with the other API regions.  I guess we should do that to be consistent.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/SpiRegion.java
```diff
@@ -0,0 +1,173 @@
...
+     *                     the specified digraph. The returned digraph will be used to
+     *                     call {@link RegionDigraph#replace(RegionDigraph)} on the
+     *                     specified digraph.
+     * @throws BundleException  if the update could not complete or if the
+     *                              callable throws a BundleException it is rethrown.
+     * @throws RuntimeException if the callable throws any exception other
+     *                              than a BundleException.
+     */
+    @FFDCIgnore({ BundleException.class, InterruptedException.class })
+    public static void update(RegionDigraph digraph, Callable<RegionDigraph> updateTo) throws BundleException {
```
tjwatson: I think this is just a copy paste form APIRegion and is not actually needed.

dev/com.ibm.ws.classloading.bells_fat/fat/src/com/ibm/ws/classloading/BellFatTest.java
```diff
@@ -1,13 +1,12 @@
 package com.ibm.ws.classloading;
 
-import static org.hamcrest.MatcherAssert.assertThat;
+//import static org.hamcrest.MatcherAssert.assertThat;
```
tjwatson: Do not leave commented out imports.  That will only confuse future readers of this.

dev/com.ibm.ws.classloading.bells_fat/fat/src/com/ibm/ws/classloading/BellFatTest.java
```diff
@@ -74,14 +74,13 @@ static void buildAndExportBellLibrary(LibertyServer targetServer, String archive
     public static void tearDown() throws Throwable {
         stopServer();
         try {
-            server.uninstallUserFeature(USER_FEATURE_NAME);
-            server.uninstallUserBundle(USER_BUNDLE_NAME);
+//            server.uninstallUserFeature(USER_FEATURE_NAME);
```
tjwatson: This is likely a temporary thing for local testing.

dev/com.ibm.ws.classloading.bells_fat/fat/src/com/ibm/ws/classloading/BellFatTest.java
```diff
@@ -110,157 +109,187 @@ static void stopServer() {
...
-     */
-    @Test
-    @AllowedFFDC({INSTANTIATION_EXCEPTION, CLASS_NOT_FOUND_EXCEPTION, NO_CLASSDEF_FOUND_EXCEPTION, EXCEPTION} )
-    public void testMultipleImplsOfSingleService() throws Exception {
-        final int numOfLogEntries = server.waitForMultipleStringsInLog(2, ".*CWWKL0050I: .*testMultipleImplsOfSingleService");
-        assertEquals("Should load both implementations of the service.",
-                     2,
-                     numOfLogEntries);
-    }
+//    /**
```
tjwatson: This is clearly a work in progress.

dazavala:
#build

dazavala:
#build 
Changes to modify minor version on classloading spi bundle, and attempt to fix library metatype validation failure. 

> ERROR: com.ibm.websphere.appserver.classloading-1.0
[BAD_BUNDLE_VERSION dev/spi/ibm/com.ibm.websphere.appserver.spi.classloading_1.4.59.jar]
Bundle com.ibm.websphere.appserver.spi.classloading known at location dev/spi/ibm/com.ibm.websphere.appserver.spi.classloading_1.4.59.jar used to have version 1.4.54.202106021349 and is now known at version 1.4.59.202111301344. Minor package changes have been made in the bundle, and it's version should be changed from 1.4.XXX to 1.5.XXX
For More Information, visit http://was.pok.ibm.com/xwiki/bin/view/Build/FeatureChecker

>unknown.attribute} > an unknown attribute was found: description="%library.spi.spi.desc" -- for sharedlibrary.spiTypeVisibility.MetatypeAdOption{value='spi' label='%library.spi.spi'} unknown attribute

dazavala:
#build review changes.
@tjwatson Review changes are in the stacked commit.  Will post summary of our review discussions after final review.

tjwatson (commented):
Just committing my past review comments that I never submitted.  This may all be addressed or outdated now.

dev/com.ibm.ws.classloading.bells_fat/publish/servers/bell_FAT_Server/jvm.options
```diff
@@ -0,0 +1 @@
+-Dcom.ibm.ws.beta.edition=true
```
tjwatson: Slight concern that this will run all bell tests with beta.edition.  Risk being that we break some existing bell function when not beta.

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/classloading/GatewayConfiguration.java
```diff
@@ -89,6 +89,22 @@
...
+     */
+    GatewayConfiguration setSpiTypeVisibility(SpiType... types);
+    
+    /**
+     * Sets the allowed SPI types. 
+     *
+     * @param types The type that of SPIs that are allowed to be loaded by the gateway class loader
+     * @return The GatewayConfiguration with this property set
+     */
+    GatewayConfiguration setSpiTypeVisibility(Iterable<SpiType> types);
```
tjwatson, line 106: If this is real SPI then we have to throw an exception when it is called for the first time with non-beta?

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/classloading/GatewayConfiguration.java
```diff
@@ -109,7 +125,21 @@
...
+     * </p>
+     * <ul>
+     * <li><code>null</code>: Do not restrict the packages visible to the class loader</li>
+     * <li>Empty: Restrict the packages available to just "internal" packages</li>
+     * <li>Non Empty Set: Restrict the packages available to just the types supplied</li>
+     * </ul>
+     * 
+     * @return The set of allowed SPI types
+     */
+    EnumSet<SpiType> getSpiTypeVisibility();
```
tjwatson, line 140: Same beta question

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/library/Library.java
```diff
@@ -69,4 +70,8 @@
      */
     Collection<ArtifactContainer> getContainers();
 
+    /**
+     * Get the allowed SPI types for this shared library.
+     */
+    EnumSet<SpiType> getSpiTypeVisibility();
```
tjwatson, line 76: Same beta question, if this really is SPI

tjwatson (commented):
A few minor comments that I would like answered before approving.

dev/com.ibm.ws.classloading.bells_fat/test-applications/testSpiTypeVisible.jar/src/com/ibm/ws/test/SpiTypeVisible.java
dazavala: Done.  The intent was to make the code self-documenting for the classloading novice.  Plus, Class.forName() invokes native methods which may complicate debugging.

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/classloading/GatewayConfiguration.java
dazavala: Yes.  Let's add the beta fence and no-op the beta spi operations.

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/classloading/GatewayConfiguration.java
dazavala: Answered above.

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/library/Library.java
dazavala: Answered above.  All beta SPI operations are fenced and will no-op when invoked in a non-beta edition.

dev/com.ibm.ws.classloading.bells_fat/publish/servers/bell_FAT_Server/jvm.options
dazavala: All new Bell FAT tests were changed to run as beta and non-beta editions.  Utility methods were added to set and remove the ProductInfo beta-edition jvm option on/from the test server.

dazavala:
Responses and changes for all review comments are ready for resolution and approval.  Changes were amended to the "Review..." commit.

Here is a summary of our earlier review discussion, the changes from which also appear in the "Review..." commit:

`GatewayConfiguration`
- Remove statement "Once set it cannot be unset" from javadoc of new methods.  
- Because this interface is wsspi, modify new methods for get/setSpiTypeVisibility() to return an `EnumSet` of `SpiType` and to accept arguments of type `SpiIType` in the same manner as get/setApiTypeVisibility(). 

> Done.  A new wsspi enum, SptType, was added.  The  enum supports one SPI type, SPI, which designates all SPI packages exported by the server's feature configuration.  Unlike ApiType, SpiType does not produce lists/sets of SPI types strings greater than length=1, nor does it support`+`/`-` prefixes in spi type values.  Like ApiType, SpiType emits an error message for invalid spi type strings.  A unit test for SpiType was added to project classloading_test 

`GatewayBundleFactory`
- In createGatewayBundle(), modify the call to addAttributeValues(MANIFEST_GATEWAY_ALLOWEDSPITYPES_PROPERTY_KEY, gwConfig.getSpiTypeVisibility() ? "spi" : "")  to return "[spi]" or "[]".  The manifest must contain data in the same format as that returned by APIType.toString()
> The method was changed to call gw.Config.getSpiTypeVisibility(), which will return the EnumSet "[spi]" or "[]". 
- Q: Can apiTypeVisibility ever be null or "[]"?
> Yes.  A user may configure apiTypeVisibilty="", which translates to the empty set, "[]".  Also, wsspi method setApiTypeVisibility() allows the caller to pass null, "[]", or any list of API types.  Null and "[]" have different semantics for api package visibility at the gateway; not so for spiTypeVisibility.
- Change method connectToApiRegions to connectToApiSpiRegions()
- In connectToApiSpiRegions(), do not connect all-spi region to internal (remove line 237).
> Done.  We needn't connect the new spi region to internal, because any configuration of api type visibility (e.g. null, ""=[], [api, spec, ibmapi, stable, _etc_]) will enable internal packages to be visible at the gateway.  

`Library`
- This interface is wsspi.  Modify new method getSpiTypeVisibility() to return an `EnumSet` of `SpiType` in the same manner as getApiTypeVisibility(). 
> Done.

Library.xml, library props
- Remove <Option> from the new spiTypeVisibility attribute, and remove the option's name and desc property strings.
- Set all properties in spiTypeVisibility to "internal"
> Done.   By removing the <option>, we can no longer rely on the config service to catch invalid values for spiTypeVisibility.  Validation now appears in the new SpiType enum.
> spiTypeVisibility attribute properties that will be reinstated at GA remain declared in library.props

`LibraryGeneration`
- Collect the library configuration using using the new SPIType support.
- Remove the Tr.info statement for non-beta images in betaFenceCheck()
> Done.  A new attribute, spiTypeVisibility, was added to SharedLibraryConstants.SharedLibraryAttribute. 

`Provisioner`
- In connectToGatewayHubs() do not add SPI packages to osgiAppsAPIFilters.
> Done.  This was a definite bug.  Thanks!

#build

dazavala:
#build
Recent updates that prevent the AppClassLoader from using common libraries (delegates) that have SPI type visibility.  The solution implicitly applies to global libraries, resource adapters, and ThreadContext classloaders.

dazavala:
#build  Changes to beta-fence the library delegate solution to avoid any possibility of emitting new message CWWKL0019W.

dev/com.ibm.ws.classloading.bells_fat/test-applications/testSpiTypeVisible.jar/src/com/ibm/ws/test/SpiTypeVisible.java
dazavala: The BELL SPI visibility FAT is now enabled the use both forName and loadClass operations, and the FAT variations exercises both load patterns.  In particular, the BELL service is exercised at least once using forName to verify SPI classes are not visible to the application after the BELL library loader has become the initiating loader for the same SPI classes.

dev/com.ibm.ws.classloading/src/com/ibm/wsspi/classloading/GatewayConfiguration.java
dazavala: Done.

dazavala:
#build
Changes rebased on origin/release; no merge conflicts

dazavala:
#build
Update javadoc bundle version in feature classloading-1.0; fixes build break: 
> com.ibm.websphere.appserver.spi.classloading_1.4-javadoc.zip for feature com.ibm.websphere.appserver.classloading-1.0 is missing. At the time of writting this was always caused by missing JavaDoc bundles where their generation failed due to not having a dependency on the OSGi spec JARs. If the file reported is a JavaDoc file see 97157 for a possible fix.

dazavala:
This is the prototype. Closing now because the socialized version of this enhancement recently delivered via [PR 21708 ](https://github.com/OpenLiberty/open-liberty/pull/21708).  The branch will remain for a while as it contains a full implementation of library spiTypeVisibility, which was removed from the socialized design.

PR #17660, opened 2021-06-23
policy executor cancel bugs

### LINKED ISSUE #17658: concurrencyPolicy loses queue slots when managed executor deactivates and erroneously cancels tasks of other executors
While looking at the code to prepare for adding some new function, I noticed that the identifier that is supplied by the managed executor to PolicyExecutor.cancel is ignored by the policy executor and it appears to indiscriminately cancel everything.  I predicted that this would cause a bug where 2 or more managed executors sharing the same concurrencyPolicy and used by different applications where only one managed executor deactivates, would erroneously see cancellation of tasks belonging to the other managed executor as well.  After writing a test case to try it out, I first ran into another bug, which is that tasks which are canceled from the queue upon deactivate, although removed from the queue and stopped from ever starting, still count against the maxQueueSize of the concurrency policy.  After I fixed this, I then reproduced the issue that I originally expected to see, which I'll also use this issue to fix.

njr-11:
fixes #17658


njr-11:
#build

jmstephensgit (approved):

njr-11:
build has 1 test failure, which is a duplicate in an unrelated test bucket

PR #17662, opened 2021-06-23
Issue 16237: Make EJB 4 RMIC Compatible

### LINKED ISSUE #16237: Jakarta EE 9: Enterprise Beans 4.0 always RMIC Compatible
For Enterprise Beans 4.0 (and future versions), generated remote Stub and Tie classes should always be RMIC compatible.

In EJB 3.2, the behavior is configured with the following JVM property with the default being FALSE:

com.ibm.websphere.ejbcontainer.rmicCompatible

For Enterprise Beans 4.0, this JVM property should be ignored, and the behavior should always be rmicCompatible="ALL".

The following needs to be considered:
1 - application server process
2 - application client process

Also, note that the application client process currently makes a call into the server to obtain a list of EJB interfaces that should be generated with RMIC compatible; this call will no longer be required for enterprise beans 4.0.

tkburroughs:
For Enterprise Beans 4.0 all remote interfaces will use RMIC compatible
stub and tie classes; the setting will not be configurable.

Also, the application client will no longer need to obtain a list
of RMIC compatible interfaces from the server.

fixes #16237

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/wsspi/ejbcontainer/JITDeploy.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/wsspi/ejbcontainer/JITDeploy.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2017 IBM Corporation and others.
+ * Copyright (c) 2008, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -12,6 +12,8 @@ package com.ibm.wsspi.ejbcontainer;
 
 import java.util.Arrays;
 import java.util.List;
+
+import javax.ejb.EJBException;
 
 import com.ibm.ejs.container.EJBConfigurationException;
 import com.ibm.websphere.ras.Tr;
@@ -27,8 +29,7 @@ import com.ibm.ws.ejbcontainer.jitdeploy.JIT_StubPluginImpl;
  * @since WAS 7.0
  * @ibm-spi
  **/
-public final class JITDeploy
-{
+public final class JITDeploy {
     private static final TraceComponent tc = Tr.register(JITDeploy.class,
                                                          "JITDeploy",
                                                          "com.ibm.ejs.container.container");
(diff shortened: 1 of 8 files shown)


tkburroughs:
#build

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/wsspi/ejbcontainer/JITDeploy.java
```diff
@@ -80,41 +81,27 @@ private JITDeploy()
...
 
-        if (options == null)
-        {
+        if (options == null) {
             flags = RMIC_COMPATIBLE_DEFAULT;
-        }
-        else if (options.equals("none"))
-        {
+        } else if (options.equals("none")) {
             flags = 0;
```
brideck: Should this and the next two cases use the constants for consistency in the code?

brideck (commented):
Do we already have sufficient remote + EE9 FAT in place to cover these code paths, or is that yet to come in our near future FAT enablement work?

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/wsspi/ejbcontainer/JITDeploy.java
tkburroughs: @brideck  Existing code that wasn't using constants... but that is because constants didn't exist previously.  Since I had added a new constant for "all", then for consistency, I added another constant for "none" and set the default constant to the new "none" constant and updated the code to use the constants. Constants are now used consistently and support the ability to change the "default" value without changing the rest of the code (not expecting to ever do that).

tkburroughs:
> Do we already have sufficient remote + EE9 FAT in place to cover these code paths, or is that yet to come in our near future FAT enablement work?

@brideck 
The "com.ibm.ws.ejbcontainer.remote.client_fat" project does cover the bulk of the effected paths and does run with Jakarta EE 9 even in Lite mode. Specifically, it does use an application client and injects remote EJBs into that client as well as looks up remote EJBs via JNDI.

Additional remote FAT should be included later as part of the FAT enablement work.

tkburroughs:
#run-libby-bot
#build

brideck (approved):
Looks and sounds good.

PR #17663, opened 2021-06-23
Fix Key/Trust store type in config

c00crane:
Added a default key and trust store a while back to prevent an intermittent issue with creating the default keystore.
The new keystores are p12 type files, but, the config set:
type="jks"
and should have set:
type="pkcs12"
doh...
Things work fine with Java 11 and above, but, Java 8 can't handle it.
Java 8 gives a message indicating that the password was invalid in some cases, and the format was invalid in other cases.
Updating the type fixes the issue...


c00crane:
#build

ayoho (approved):

PR #17667, opened 2021-06-24
Rename reserved word in ReduceableSemaphore

njr-11:
Greg pointed out that https://openjdk.java.net/jeps/409 is in Java 17.
It shows "permits" becoming a reserved word, so we will need to rename a variable in ReduceableSemaphore which has that same name so that compilation against Java 17 will be possible.


njr-11:
#build

gjwatts (approved):

njr-11:
Build has 1 test failure which is a duplicate of an unrelated issue.

PR #17697, opened 2021-06-28
Support creating a merged OpenAPI document when multiple apps are deployed

Azquelt:
Initial implementation for #16628 

When multiple applications are deployed, mpOpenApi-2.0 can be configured to generate an OpenAPI document for each of them and then merge them together into a single OpenAPI document which is presented on the `/openapi` endpoint.

There is no change to the default behaviour, the `mp.openapi.extensions.liberty.merged.include` property must be set to `"all"` or to a list of applications to include to enable the merged view.

Currently this functionality is only available in nightly/beta releases.


Azquelt:
#build

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/clashing-path-merged.yaml
```diff
@@ -0,0 +1,120 @@
+openapi: 3.0.3
+info:
+  title: "Merged documentation"
+  version: "1.0"
+paths:
+  "/foo/foo":
```
msmiths, line 6: This behaviour concerns me because it implies that the path for the resource has been changed as a result of the merge. Is this correct, or am I missing something?

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/clashing-path-merged.yaml
```diff
@@ -0,0 +1,120 @@
...
+        200:
+          description: success
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/foo"
+          links:
+            "list":
+              operationId: getAllFoo
+  "/foo/login":
```
msmiths, line 33: This behaviour concerns me because it implies that the path for the resource has been changed as a result of the merge. Is this correct, or am I missing something?

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/clashing-path-merged.yaml
```diff
@@ -0,0 +1,120 @@
...
+                properties:
+                  sessionKey:
+                    type: string
+        400:
+          description: "failure"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
+  "/bar/bar":
```
msmiths, line 52: This behaviour concerns me because it implies that the path for the resource has been changed as a result of the merge. Is this correct, or am I missing something?

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/clashing-path-merged.yaml
```diff
@@ -0,0 +1,120 @@
...
+        200:
+          description: success
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/bar"
+          links:
+            "list":
+              operationId: getAllBar
+  "/bar/login":
```
msmiths, line 79: This behaviour concerns me because it implies that the path for the resource has been changed as a result of the merge. Is this correct, or am I missing something?

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/operationid-clash-merged.yaml
```diff
@@ -0,0 +1,66 @@
...
+        200:
+          description: "success"
+          links:
+            idLink:
+              operationId: "getTest"
+            refLink:
+              operationRef: "#/paths/~1test1~1testA/get"
+  "/test2/testB":
+    get:
+      operationId: getTest1
```
msmiths, line 28: OperationId are mainly intended as a guide for code generation tools when generating client/server stubs from the OpenAPI doc. I am little concerned that the result of merging the OpenAPI docs will result in different code being generated... and that affected methods will depend on the order in which the applications are processed and the docs are merged. But then I guess that this is a result of the fact that the merge processing is not clearly defined in the spec.

msmiths (commented):
There is a lot of code here and, while I believe that I understand the bulk of it, I cannot guarantee that there are issues that I have not missed.

I have raised a couple of questions/comments regarding the document merge processing... specifically for path and operationId merging... based on the YAML test docs.

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/clashing-path-merged.yaml
Azquelt: The path has had the context root added to the front of it. The context root will also be removed from the `servers` block so that the final URL ends up being the same.

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/operationid-clash-merged.yaml
Azquelt: This is true, but we run into the problem that operationIds are required to be unique by the OpenAPI spec, so we have no choice but to change them.

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/clashing-path-merged.yaml
Azquelt: In this test, there's no severs in the source document so it would be generated by the servlet, which would work out the host and port and then append `openApiProvider.getApplicationPath()` which would return the context root.

However, when we merge documents, the resulting provider returns `null` from `getApplicationPath()` and so nothing is appended to the server URL in the final document.

Azquelt:
#build

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationProcessor.java
```diff
@@ -65,7 +77,12 @@
     
     @Reference
     private ClassLoadingService classLoadingService;
-
+    
+    // This needs to be an optional dynamic reference to accommodate the circular dependency
+    // It should almost always be present
+    @Reference(cardinality = OPTIONAL, policy = DYNAMIC, policyOption = GREEDY)
+    private volatile ApplicationRegistry appRegistry;
```
tevans78: What happens if ApplicationRegistry is null? If ApplicationRegistry and ApplicationProcessor depend on each other, why have them as separate components?

Azquelt: The code that calls it checks for nullness first.

They're separate components since they're responsible for different things, the application processor handles processing a single application, the application registry keeps track of all the deployed applications. The link this direction is needed because we want to log a message if multiple modules are deployed but merging modules is disabled. The ApplicationRegistry is responsible for making sure this message is only emitted once.

skasund:
L2 message review completed

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,29 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
 OPENAPI_APPLICATION_PROCESSING_ERROR.explanation=A runtime error occurred when processing the application.
-OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
\ No newline at end of file
+OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
+
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W=CWWKO1662W: The following problems occurred while merging OpenAPI documents: {0}
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W.explanation=The OpenAPI documentation from multiple web modules could not be successfully merged. If the reported problems are not fixed, the endpoints from the reported modules will not be included in the generated OpenAPI document.
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W.useraction=Review the merge problems to identify why the generated OpenAPI documents could not be merged.
+
+OPENAPI_MERGE_PROBLEM_PATH_CLASH=Path {0} from module {1} clashes with a path from module {2}. {1} will not be merged.
```
tevans78: I think this message still needs it's own code

Azquelt: Even though it's included as part of another message? It's one of the problems which will be listed under CWWKO1662W (actually I think it's currently the only problem but it can happen multiple times).

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+ */
+public class MergeProcessor {
+
+    private static final TraceComponent tc = Tr.register(MergeProcessor.class);
+
+    private static final Info MERGED_INFO;
+
+    static {
+        MERGED_INFO = OASFactory.createInfo();
+        MERGED_INFO.setTitle("Merged documentation");
```
tevans78: NLS?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+        MERGED_INFO.setVersion("1.0");
```
tevans78: Constant?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+    public static OpenAPIProvider mergeDocuments(List<OpenAPIProvider> documents) {
+
+        MergeProcessor mergeProcessor = new MergeProcessor();
+        
+        if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
+            Tr.event(mergeProcessor, tc, "Beginning merge of OpenAPI documents for " + documents);
+        }
+
+        for (OpenAPIProvider document : documents) {
+            mergeProcessor.add(document);
```
tevans78: why not pass the docs in on the constructor?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+                Tr.event(this, tc, "Starting path processing for " + provider);
+            }
+
+            // Make a deep copy, to avoid modifying the input model
+            OpenAPI model = (OpenAPI) ModelCopy.copy(provider.getModel());
+
+            RenameHolder renameHolder = new RenameHolder();
+
+            // Add context root to the front of paths if possible
+            prependPaths(model, provider.getApplicationPath(), renameHolder);
```
tevans78: instead of creating a new RenameHolder and passing it in, wouldn't it be cleaner to have prependPaths create and return the RenameHolder?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+        boolean serversIdentical = isServersIdentical(inProgressModels);
+
+        if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
+            Tr.event(this, tc, "Security identical = " + securityIdentical);
+            Tr.event(this, tc, "Info identical = " + infoIdentical);
+            Tr.event(this, tc, "External docs identical = " + externalDocsIdentical);
+            Tr.event(this, tc, "Servers identical = " + serversIdentical);
+        }
+
+        // Second loop which moves and renames parts of the models where necessary to avoid clashes
```
tevans78, line 183: maybe some of the process method should be split out into separate methods?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+            }
+            return new MergedOpenAPIProvider(includedProviders.get(0).getModel(), mergeProblems);
+        }
+
+        OpenAPI merged = OASFactory.createOpenAPI();
+        for (OpenAPI model : processedModels) {
+            merged = MergeUtil.merge(merged, model);
+        }
+
+        merged.setOpenapi("3.0.3");
```
tevans78: Constant?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+     * Actually assumes that equals is implemented properly and just checks that the first element is equal to all others
+     * <p>
+     * If {@code collection} contains less than two elements, this method will always return {@code true}.
+     * 
+     * @param <T> the element type
+     * @param collection the collection of elements to test for equality
+     * @param comparator the function to use to test equality
+     * @return {@code true} if all elements of {@code collection} are equal, {@code false} otherwise
+     */
+    private static <T> boolean allEqual(Collection<? extends T> collection, BiPredicate<? super T, ? super T> comparator) {
```
tevans78: should these 4 static methods be in a generic utility class?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
...
+    }
+
+    /**
+     * Creates and stores new unique names for model elements.
+     * <p>
+     * Certain names must be unique within an OpenAPI document. When merging documents, these names may have to be changed so that they don't clash.
+     * <p>
+     * This class handles detecting clashes, creating new names as required and looking up the new names of renamed elements.
+     */
+    public class RenameHolder {
```
tevans78: In general the MergeProcessor class seems a little long and there are things that could be moved ... RenameHolder could be moved to its own standalone class, although maybe with a better name?

tevans78: ~If I wasn't suggesting moving this out, I would be wondering if RenameHolder should be a private static class?~ Ah used by RenameReferenceVisitor ... so needs to be public, but what about static?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
```diff
@@ -0,0 +1,773 @@
+package io.openliberty.microprofile.openapi20.merge;
```
tevans78, line 11: Missing copyright headers from many new files

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelCopy.java
```diff
@@ -0,0 +1,63 @@
+package io.openliberty.microprofile.openapi20.merge;
```
tevans78, line 11: Copyright

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelEquality.java
```diff
@@ -0,0 +1,89 @@
+package io.openliberty.microprofile.openapi20.merge;
```
tevans78, line 11: Copyright

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: Which bits did you have in mind?

At the moment, the process method does very little other than call other private methods which do well defined tasks. As it is, I think it's fairly easy to see the complete set of steps taken to merge several models so I'm not sure that breaking it up any more would make it easier to read or understand.

There are some slightly verbose bits, maybe `InProgressModel` should have a constructor so that its creation doesn't take up so many lines. (It was added fairly late when I realised I needed two loops so that I could remove any models which couldn't be merged before I calculate whether the info, servers and security sections are identical across all models).

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelEquality.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
+
+public class ModelEquality {
```
tevans78, line 26: No javadoc

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelType.java
```diff
@@ -0,0 +1,230 @@
+package io.openliberty.microprofile.openapi20.merge;
```
tevans78, line 11: Copyright

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: Probably not. `prependPaths` deals with moving the context root from servers to paths, there's no reason it should create the `renameHolder`. `renameHolder` is also passed as a parameter in the same way to other methods further down.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelEquality.java
```diff
@@ -0,0 +1,89 @@
...
+        Iterator<?> ai = a.iterator();
+        Iterator<?> bi = b.iterator();
+        while (ai.hasNext()) {
+            if (!equals(ai.next(), bi.next())) {
+                return false;
+            }
+        }
+
+        return true;
+    }
```
tevans78: These two are generic utility methods and could go in a generic class elsewhere

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelType.java
```diff
@@ -0,0 +1,230 @@
...
+import org.eclipse.microprofile.openapi.models.security.OAuthFlow;
+import org.eclipse.microprofile.openapi.models.security.OAuthFlows;
+import org.eclipse.microprofile.openapi.models.security.SecurityRequirement;
+import org.eclipse.microprofile.openapi.models.security.SecurityScheme;
+import org.eclipse.microprofile.openapi.models.servers.Server;
+import org.eclipse.microprofile.openapi.models.servers.ServerVariable;
+import org.eclipse.microprofile.openapi.models.tags.Tag;
+
+/**
+ * Tools for reflectively navigating model objects
```
tevans78: is this specific to an open api version?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: I wasn't sure. They are very generic, but on the other hand I'm not planning to use them anywhere else right now. I would definitely move them to a utility class if I had somewhere else I wanted to use them.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
```diff
@@ -0,0 +1,230 @@
+/*******************************************************************************
+ * Copyright (c) 2017 IBM Corporation and others.
```
tevans78: Wrong date

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
```diff
@@ -0,0 +1,230 @@
...
+            String newName = renameHolder.lookupName(nameType, oldName);
+            if (!oldName.equals(newName)) {
+                referrer.setRef(prefix + newName);
+            }
+        }
+    }
+
+    @Override
+    public Example visitExample(Context context, Example example) {
+        updateReference(example, NameType.EXAMPLES, "#/components/examples/");
```
tevans78: Constant?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
```diff
@@ -0,0 +1,230 @@
...
+        if (link.getRef() != null) {
+            updateReference(link, NameType.LINKS, "#/components/links/");
+        } else {
+            link.setOperationId(renameHolder.lookupName(NameType.OPERATION_ID, link.getOperationId()));
+            link.setOperationRef(updateOperationRef(link.getOperationRef()));
+        }
+        return link;
+    }
+
+    private static final Pattern PATH_REF_PATTERN = Pattern.compile("#/paths/([^/]+)/.*");
```
tevans78, line 103: If the component strings above are constants then maybe this path pattern goes with them?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
tevans78: all of the strings that start `#/` e.g. `#/components/links/`

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
```diff
@@ -0,0 +1,230 @@
...
+
+    }
+
+    private String decodePathReference(String pathReference) {
+        return pathReference.replace("~1", "/").replace("~0", "~");
+    }
+
+    private String encodePathReference(String path) {
+        return path.replace("~", "~0").replace("/", "~1");
+    }
```
tevans78: static methods?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: I don't disagree with a better name :stuck_out_tongue_closed_eyes: 

It currently has to be a non-static inner class because it references `MergeProcessor.namesInUse`.

Possibly we could move it out into a separate file, but it would still have to be an outer and an inner class, with the outer class owning `namesInUse`.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/utils/MessageConstants.java
```diff
@@ -24,6 +24,12 @@
     public static final String OPENAPI_FILTER_LOAD_ERROR            = "OPENAPI_FILTER_LOAD_ERROR"; //$NON-NLS-1$
     public static final String OPENAPI_MODEL_READER_LOAD_ERROR      = "OPENAPI_MODEL_READER_LOAD_ERROR"; //$NON-NLS-1$
     public static final String UNSUPPORTED_CSS_VALUE                = "UNSUPPORTED_CSS_VALUE"; //$NON-NLS-1$
+    public static final String OPENAPI_MERGE_PROBLEMS_CWWKO1662W    = "OPENAPI_MERGE_PROBLEMS_CWWKO1662W"; //$NON-NLS-1$
```
tevans78, line 27: Copyright date

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelEquality.java
Azquelt: They aren't because they call `ModelEquality.equals()`

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/utils/OpenAPIUtils.java
```diff
@@ -13,11 +13,17 @@
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
```
tevans78, line 20: Copyright date

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/ModelType.java
Azquelt: Hmm, it probably is. Will add that to the doc.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/utils/OpenAPIUtils.java
```diff
@@ -13,11 +13,17 @@
...
 
+import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.openapi.OASFactory;
 import org.eclipse.microprofile.openapi.models.OpenAPI;
+import org.eclipse.microprofile.openapi.models.info.Info;
 import org.eclipse.microprofile.openapi.models.servers.Server;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
```
tevans78, line 29: Are you sure you want to use jackson directly and not the standard jsonp APIs?

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/fat/src/io/openliberty/microprofile/openapi20/fat/FATSuite.java
```diff
@@ -16,12 +16,16 @@
 
 import io.openliberty.microprofile.openapi20.fat.cache.CacheTest;
 import io.openliberty.microprofile.openapi20.fat.deployments.DeploymentTest;
+import io.openliberty.microprofile.openapi20.fat.deployments.MergeConfigTest;
```
tevans78, line 19: Copyright date

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/fat/src/io/openliberty/microprofile/openapi20/fat/deployments/DeploymentTest.java
```diff
@@ -51,7 +51,7 @@
 
     @Server("OpenAPITestServer")
     public static LibertyServer server;
-    
+
```
tevans78, line 64: Missing Copyright header

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/fat/src/io/openliberty/microprofile/openapi20/fat/deployments/MergeTest.java
```diff
@@ -0,0 +1,297 @@
+package io.openliberty.microprofile.openapi20.fat.deployments;
```
tevans78, line 11: Copyright header

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/utils/OpenAPIUtils.java
Azquelt: Sadly, I think so

1. We are already including Jackson as a dependency
2. We're calling `InfoReader.readInfo` from smallrye which takes a Jackson `ObjectNode`

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/fat/src/io/openliberty/microprofile/openapi20/fat/deployments/TestServlet.java
```diff
@@ -0,0 +1,23 @@
+package io.openliberty.microprofile.openapi20.fat.deployments;
```
tevans78, line 11: Copyright

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/fat/src/io/openliberty/microprofile/openapi20/fat/utils/OpenAPITestUtil.java
```diff
@@ -10,13 +10,17 @@
  *******************************************************************************/
 package io.openliberty.microprofile.openapi20.fat.utils;
 
+import static org.hamcrest.Matchers.hasItems;
```
tevans78, line 13: Copyright date

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/fat/src/io/openliberty/microprofile/openapi20/fat/utils/OpenAPITestUtil.java
```diff
@@ -237,8 +241,8 @@ public static void checkInfo(JsonNode root, String defaultTitle, String defaultV
         String title = infoNode.get("title").textValue();
```
tevans78, line 241: Aren't `title` and `version` constants somewhere?

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/publish/servers/OpenAPIMergeTestServer/jvm.options
```diff
@@ -0,0 +1 @@
+-Dcom.ibm.ws.beta.edition=true
```
tevans78, line 1: why not put this in bootstrap.properties?

Azquelt: It doesn't work in bootstrap.properties because the server reads the system property and stores the value before it reads bootstrap.properties. (That took ages to work out. I did update the wiki page but didn't think to add a comment here.)

dev/io.openliberty.microprofile.openapi.2.0.internal_fat/fat/src/io/openliberty/microprofile/openapi20/fat/utils/OpenAPITestUtil.java
Azquelt: Oddly, it doesn't look like they are. Usually we have a model object and we're calling `Info.getTitle()` instead, or we're not in java code and can't use a constant anyway.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Override
+    public Schema visitSchema(Context context, Schema schema) {
+        updateReference(schema, NameType.SCHEMAS, "#/components/schemas/");
+        return schema;
+    }
+
+    @Override
+    public Schema visitSchema(Context context, String key, Schema schema) {
+        updateReference(schema, NameType.SCHEMAS, "#/components/schemas/");
```
Emily-Jiang: This method has the same impl as the method on 177

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Override
+    public Parameter visitParameter(Context context, Parameter p) {
+        updateReference(p, NameType.PARAMETERS, "#/components/parameters/");
+        return p;
+    }
+
+    @Override
+    public Parameter visitParameter(Context context, String key, Parameter p) {
+        updateReference(p, NameType.PARAMETERS, "#/components/parameters/");
```
Emily-Jiang: why not just call the previous method?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Override
+    public RequestBody visitRequestBody(Context context, RequestBody rb) {
+        updateReference(rb, NameType.REQUEST_BODIES, "#/components/requestBodies/");
+        return rb;
+    }
+
+    @Override
+    public RequestBody visitRequestBody(Context context, String key, RequestBody rb) {
+        updateReference(rb, NameType.REQUEST_BODIES, "#/components/requestBodies/");
```
Emily-Jiang: why not just call the previous method?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationProcessor.java
Azquelt: Moved the method which emits the message into its own component which can be reliably injected into both locations without causing a loop.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: I don't think so, we don't NLS the other default values for the OpenAPI model. I will make it a constant.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: Moved these methods to `OpenAPIUtils`

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: Moved `RenameHolder` into a different file.

Now have `NameProcessor` which holds `namesInUse` with an inner class `DocumentNameProcessor`.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/MergeProcessor.java
Azquelt: I've shortened some of the verbose bits in the first loop.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
Azquelt: As you suggested, I've added these strings to the `NameTypes` enum (which has also moved to its own file).

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
Azquelt: Didn't move this as it's very tied to the method where it's used.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/merge/RenameReferenceVisitor.java
Azquelt: I've made this change. I'm not entirely sure about it, but it does at least reduce the duplication of code slightly and reinforces that the implementation of both methods is the same.

I'm not totally sure whether this is worth adding the extra step of redirection or calling a method declared on the interface in a different way to how it was designed. However, we did spend way too long thinking about it.

Azquelt:
#build

Emily-Jiang (approved):
LGTM!

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
tevans78: @OpenLiberty/message-reviewer we need some guidance here. Even though this message will be nested inside another one, I think it still needs a NLS code?

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
 
 OPENAPI_APPLICATION_PROCESSING_ERROR=CWWKO1661E: An error occurred when processing application {0} and an OpenAPI document was not produced. The error was: {1}.
 OPENAPI_APPLICATION_PROCESSING_ERROR.explanation=A runtime error occurred when processing the application.
-OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
\ No newline at end of file
+OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
+
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W=CWWKO1662W: The following problems occurred while merging OpenAPI documents: {0}
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W.explanation=The OpenAPI documentation from multiple web modules could not be successfully merged. If the reported problems are not fixed, the endpoints from the reported modules will not be included in the generated OpenAPI document.
```
helyarp: Change "could not be" to "cannot be"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W.useraction=Review the merge problems to identify why the generated OpenAPI documents could not be merged.
+
+OPENAPI_MERGE_PROBLEM_PATH_CLASH=Path {0} from module {1} clashes with a path from module {2}. {1} will not be merged.
+OPENAPI_MERGE_PROBLEM_PATH_CLASH.explanation=The given path is present in two modules. Both instances of the path can't be included in the same OpenAPI document. This message is returned as part of CWWKO1662.
```
helyarp: Maybe change "can't" to "cannot"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_PROBLEM_PATH_CLASH.useraction=Either change one of the conflicting modules or configure the mpOpenApi-2.0 feature to exclude one of them. This problem can be caused if servers are included in the OpenAPI document and the server URL does not end with the context root of the application (e.g. because the application will be accessed at a different URL through a reverse proxy).
```
helyarp: Change the "(e.g. ..)" phrase to "(for example, because the application is accessed at a different URL through a reverse proxy)"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+
+OPENAPI_MERGE_DISABLED_CWWKO1663I=CWWKO1663I: Combining OpenAPI documentation from multiple modules is disabled. Only {0} will be used to generate OpenAPI documentation. To enable merging of OpenAPI documentation from all modules, set config property mp.openapi.extensions.liberty.merged.include = all.
```
helyarp: Change "will be" to "is"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_DISABLED_CWWKO1663I.explanation=Unless OpenAPI documentation merging is enabled, OpenAPI documentation will only be generated for the first OpenAPI module deployed.
```
helyarp: Change "will only be generated" to "is generated only"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_DISABLED_CWWKO1663I.useraction=If you want all of your web modules to be used to generate OpenAPI documentation, set the MicroProfile Config property listed in the message. This can be set in any of the locations that MicroProfile Config searches for config properties, e.g. in a Java system property or an environment variable.
```
helyarp: - Change "property listed" to "property that is listed" or to "that is"
- Change "This" to "The property"
- Change "e.g." to "such as"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+
+OPENAPI_MERGE_INFO_INVALID_CWWKO1664W=CWWKO1664W: The info object configured with property {0} is not valid. The title and version properties must be set. The invalid value was: {1}
```
helyarp: - If "info" is not the object name, please change it to "information."
- Change "property {0}" to "the {0} property"
- Change "was" to "is"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_INFO_INVALID_CWWKO1664W.explanation=If the property is set, it must be set to a complete info section in JSON format, with the title and version properties set to valid values.
+OPENAPI_MERGE_INFO_INVALID_CWWKO1664W.useraction=Correct the configured info value.
+
+OPENAPI_MERGE_INFO_PARSE_ERROR_CWWKO1665W=CWWKO1665W: The config property {0} could not be parsed as JSON. The value was: {1}\nThe error was: {2}
```
helyarp: Change "The config property {0} could not be parsed as JSON. The value was: {1}\nThe error was: {2}" to "The {0} config property cannot be parsed as JSON. The value is: {1}\nThe error is: {2}"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_INFO_PARSE_ERROR_CWWKO1665W.explanation=If the property is set, it must be set to a complete info section in JSON format.
+OPENAPI_MERGE_INFO_PARSE_ERROR_CWWKO1665W.useraction=Review the error text to identify the problem and correct the configured info value so that it is valid JSON.
+
+OPENAPI_MERGE_INVALID_NAME_CWWKO1666W=CWWKO1666W: Invalid name found in {0}: {1}. Each name must be of the format "applicationName" or "applicationName/moduleName"
```
helyarp: Change `be of the format "applicationName" or "applicationName/moduleName"` 
to:
`use the "applicationName" or "applicationName/moduleName" format.`

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_INVALID_NAME_CWWKO1666W.explanation=The given config property must contain a comma-separated list of application and module names, or one of the special values "all", "first", or "none".
+OPENAPI_MERGE_INVALID_NAME_CWWKO1666W.useraction=Review the error text to identify the problem and correct the indicated configuration property.
+
+OPENAPI_MERGE_UNUSED_INCLUDE_CWWKO1667W=CWWKO1667W: The configuration property {0} included the value {1} but that did not match any deployed application or web module
```
helyarp: Change "The configuration property {0} included the value {1}"
to:
"The {0} configuration property included the {1} value"

helyarp (commented):
Some changes suggested.

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
Azquelt: "info" is the object name. I will make the other changes.

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
Azquelt: Hmm, that reads strangely to me. It sounds like `{1}` is a type of value, rather than the value itself. Would "The {0} configuration property included {1}" be ok? Possibly with the value in quotes?

Reading it now, I also wonder whether this should be in the present tense, as you've suggested for some of the other messages.

Examples with placeholders filled in:
* `The mp.openapi.extensions.liberty.merged.include configuration property included the myTestApplication value but that did not match any deployed application or web module.`
* `The mp.openapi.extensions.liberty.merged.include configuration property included myTestApplication but that did not match any deployed application or web module.`
* `The mp.openapi.extensions.liberty.merged.include configuration property includes myTestApplication but that does not match any deployed application or web module.`
* `The mp.openapi.extensions.liberty.merged.include configuration property includes "myTestApplication" but that does not match any deployed application or web module.`

Azquelt: I've gone with the last suggested example, but I'm happy to change it again.

Azquelt:
@helyarp I've made most of your suggested changes, except where noted.

Did you also have an opinion on whether this message should have its own message ID? https://github.com/OpenLiberty/open-liberty/pull/17697/files#diff-53fa1abc4002268be3f730d1c6de09a39bd57b4f450847224b06f24d3ca8f0aeR72-R74

It's included in part of CWWKO1662, but does it need its own message code to be included and findable in the documentation?

We expect the full message to look something like this:
```
CWWKO1662W: The following problems occurred while merging OpenAPI documents: 
 - Path /app/test1 from module myTestApplication/module1 clashes with a path from module myOtherApplication/myOtherApplication. myTestApplication/module1 will not be merged.
 - Path /app/test2 from module myTestApplication/module2 clashes with a path from module myOtherApplication/myOtherApplication. myTestApplication/module2 will not be merged.
```

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
 
 OPENAPI_APPLICATION_PROCESSING_ERROR=CWWKO1661E: An error occurred when processing application {0} and an OpenAPI document was not produced. The error was: {1}.
 OPENAPI_APPLICATION_PROCESSING_ERROR.explanation=A runtime error occurred when processing the application.
-OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
\ No newline at end of file
+OPENAPI_APPLICATION_PROCESSING_ERROR.useraction=Review the FFDC logs and exception text to identify the problem.
+
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W=CWWKO1662W: The following problems occurred while merging OpenAPI documents: {0}
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W.explanation=The OpenAPI documentation from multiple web modules cannot be successfully merged. If the reported problems are not fixed, the endpoints from the reported modules will not be included in the generated OpenAPI document.
```
helyarp: Change "will not be included" to "are not included"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_PROBLEMS_CWWKO1662W.useraction=Review the merge problems to identify why the generated OpenAPI documents could not be merged.
```
helyarp: Change "could not be merged" to "cannot be merged"

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
helyarp: In message strings with a variable, please try to use "The _variable_ noun" format.

Suggested change to message:

The {0} path from the {1} module clashes with a path from the {2} module. {1} cannot be merged.

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+
+OPENAPI_MERGE_PROBLEM_PATH_CLASH=Path {0} from module {1} clashes with a path from module {2}. {1} will not be merged.
+OPENAPI_MERGE_PROBLEM_PATH_CLASH.explanation=The given path is present in two modules. Both instances of the path cannot be included in the same OpenAPI document. This message is returned as part of CWWKO1662.
```
helyarp: Change "The given path is present in two modules." to "The path is in two modules."

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_INFO_INVALID_CWWKO1664W.explanation=If the property is set, it must be set to a complete info section in JSON format, with the title and version properties set to valid values.
+OPENAPI_MERGE_INFO_INVALID_CWWKO1664W.useraction=Correct the configured info value.
+
+OPENAPI_MERGE_INFO_PARSE_ERROR_CWWKO1665W=CWWKO1665W: The {0} config property cannot be parsed as JSON. The value is: {1}\nThe error is: {2}
+OPENAPI_MERGE_INFO_PARSE_ERROR_CWWKO1665W.explanation=If the property is set, it must be set to a complete info section in JSON format.
+OPENAPI_MERGE_INFO_PARSE_ERROR_CWWKO1665W.useraction=Review the error text to identify the problem and correct the configured info value so that it is valid JSON.
+
+OPENAPI_MERGE_INVALID_NAME_CWWKO1666W=CWWKO1666W: Invalid name found in {0}: {1}. Each name must use the "applicationName" or "applicationName/moduleName" format.
+OPENAPI_MERGE_INVALID_NAME_CWWKO1666W.explanation=The given config property must contain a comma-separated list of application and module names, or one of the special values "all", "first", or "none".
```
helyarp: Remove "given" ("The config property must contain...")

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_INVALID_NAME_CWWKO1666W.useraction=Review the error text to identify the problem and correct the indicated configuration property.
+
+OPENAPI_MERGE_UNUSED_INCLUDE_CWWKO1667W=CWWKO1667W: The {0} configuration property includes "{1}" but that does not match any deployed application or web module
```
helyarp: Add a period (.) to the end of the message.

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
```diff
@@ -63,4 +63,32 @@ OPENAPI_APPLICATION_PROCESSED.useraction=No action is required.
...
+OPENAPI_MERGE_UNUSED_INCLUDE_CWWKO1667W.explanation=The server has been configured to create OpenAPI documentation for the given application or module name, but no application or web module with that name was deployed.
```
helyarp: Change "has been" to "is" and remove "given"  ("The server is configured to create OpenAPI documentation for the application or module name, ...)

helyarp (commented):
Suggested changes to messages.

Azquelt:
#run-libby-bot

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
Azquelt: Isn't this the wrong tense after an if?

> If the reported problems are not fixed, the endpoints from the reported modules are not included in the generated OpenAPI document

Maybe I should also get rid of "If the reported problems are not fixed" altogether?

> The OpenAPI documentation from multiple web modules cannot be successfully merged. The endpoints from the reported modules are not included in the generated OpenAPI document.

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
Azquelt: I've changed this as recommended but I'm not sure about it. I'm happy with "The _variable_ noun" where the variable is a name, but it seems odd when the variable is a value (as it is here).

Example with placeholders filled in:
> The /product/{id} path from the foo/api module clashes with a path from the bar/products module. foo/api cannot be merged.

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
Azquelt: I've removed the if clause and changed the later text to "are not included"

Azquelt:
@helyarp Sorry for the delay, I've finally got back to this and I've made most of the requested changes except where noted.

Github attached one of your comments to an existing thread further up the issue, you can find my reply to it here: https://github.com/OpenLiberty/open-liberty/pull/17697#discussion_r688631589

Azquelt:
#run-libby-bot

Azquelt:
rebased on integration
#run-libby-bot
#build

dev/io.openliberty.microprofile.openapi.internal.common/resources/io/openliberty/microprofile/openapi/internal/resources/OpenAPI.nlsprops
helyarp: The "The _variable_ noun" example looks okay.

Azquelt: Cool, that's what we now have. Thanks :)

helyarp (commented):
Looks good!

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationReadException.java
```diff
@@ -0,0 +1,30 @@
+package io.openliberty.microprofile.openapi20;
```
tevans78, line 11: Copyright header?

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/MergedOpenAPIProvider.java
```diff
@@ -0,0 +1,45 @@
+package io.openliberty.microprofile.openapi20;
```
tevans78, line 11: Copyright header

dev/io.openliberty.microprofile.openapi.2.0.internal/bnd.bnd
```diff
@@ -34,7 +34,8 @@ Export-Package: \
     io.openliberty.microprofile.openapi20.validation
 
 Private-Package: \
-    io.openliberty.microprofile.openapi20.cache
+    io.openliberty.microprofile.openapi20.cache,\
```
tevans78, line 37: I think that all these `openapi20.*` packages should be `openapi20.internal.*` because they are not API. Also makes it consistent with the previous version (`io.openliberty.microprofile.openapi.internal`).

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/OpenAPIProvider.java
```diff
@@ -12,61 +12,44 @@
 
 import java.util.List;
 
-import org.eclipse.microprofile.openapi.models.servers.Server;
-
-import io.smallrye.openapi.runtime.io.Format;
+import org.eclipse.microprofile.openapi.models.OpenAPI;
```
tevans78, line 15: copyright date

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
```diff
@@ -1,18 +1,13 @@
 package io.openliberty.microprofile.openapi20;
```
tevans78, line 11: copyright header

dev/io.openliberty.microprofile.openapi.2.0.internal/bnd.bnd
Azquelt: I agree, I'll make a more generalized cleanup issue which includes this, since there are a few other changes which should be made which will affect lots of files.

dev/io.openliberty.microprofile.openapi.2.0.internal/test/io/openliberty/microprofile/openapi20/test/merge/MergeProcessorTest.java
```diff
@@ -0,0 +1,444 @@
+package io.openliberty.microprofile.openapi20.test.merge;
```
tevans78, line 11: copyright header

dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/ApplicationProcessorTest.java
```diff
@@ -144,6 +144,10 @@ public void testApplicationProcessor() throws Exception {
         assertEquals("FAIL: Only a single application must be processed by the application processor.", app1Doc,
             openapi);
 
+        // Check we didn't output the beta message
```
tevans78, line 147: strictly not a beta-only message

dev/io.openliberty.microprofile.openapi.2.0.internal/bnd.bnd
Azquelt: Raised #18790

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/MergeDisabledAlerter.java
```diff
@@ -0,0 +1,42 @@
...
+    private boolean multiAppWarningGiven = false;
+
+    /**
+     * Called at the point where we notice that multiple modules are deployed but the OpenAPI config is set to only generate docs from the first module
+     * @param firstModule the module which <i>is</i> being used to generate openAPI documentation
+     */
+    public void setUsingMultiModulesWithoutConfig(OpenAPIProvider firstModule) {
+        synchronized (this) {
+            if (!multiAppWarningGiven && ProductInfo.getBetaEdition()) {
+                Tr.info(tc, MessageConstants.OPENAPI_MERGE_DISABLED_CWWKO1663I, firstModule);
```
tevans78, line 36: Is there a test for this message?

dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/ApplicationProcessorTest.java
Azquelt: Not quite sure what you're getting at. Outputting the message should be gated behind a beta fence.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/MergeDisabledAlerter.java
Azquelt: Yes `MergeConfigTest.testFirstModuleOnly()` and `.testFirstModuleOnlyEar()` look for it.

`ApplicationProcessorTest` checks that the message is not emitted when not running the beta edition.

However, both sets of tests use the message string rather than the message code, which should probably be changed.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/WebModuleOpenAPIProvider.java
Azquelt: This was missing the copyright header before. It looks like Martin created it in 2020, so I've put that as the initial date in the header.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/MergeDisabledAlerter.java
Azquelt: I've updated the test to use the message code. Some of them still do validation on the message string to ensure the right values are filled in in the placeholders, but they all now include the message code so it should be easier to find the tests.

Azquelt:
#run-libby-bot
#build

tevans78 (approved):

dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/ApplicationProcessorTest.java
tevans78: I just mean that at some point the beta guard will be removed and this message will be output from GA code given the appropriate config options

Azquelt: Ah sure, but at that point we'll need to remove this part of the test anyway since it'll start failing.

Azquelt:
Code review commits squashed.
#run-libby-bot

No code changes since build requested

Azquelt:
Build appears to have hung.
#build

Azquelt:

PR #17721, opened 2021-06-29
Allow a larger window when assuming exp time is valid

c00crane:
When validating token content, we can't validate the exact value that will be in the exp attribute - we have to say that it's in a window of time around the time that the test request the token to be committed.  We were allowing 10 seconds, but, on slow machines that isn't always a large enough window.  Increasing the time to 20 seconds.


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.oidc.server_fat

ayoho (approved):

PR #17722, opened 2021-06-29
Issue17706-ReduceSubjectToStringCalls

emilytee:
Note: the fix introduced here helps to reduce performance overhead via calls to Subject.toString().
There is no related functional issue or regression.


emilytee:
#build

emilytee:
#run-libby-bot

PR #17735, opened 2021-06-30
Issue 16994: Update how dynamic reconfig of social login affects discovery data

### LINKED ISSUE #16994: Dynamic reconfig of discovery endpoint not updating endpoints in all cases
If Social client is reconfigured in the following steps, it will not set the authorization, token, ... endpoints.

1) configure server with a valid `discoveryEndpointUrl` attribute and none of the discoverable config attributes set 
2) start the server
3) Update the server config - replace the `discoveryEndpointUrl` value with ""
4) Note the message in the server's messages.log: `CWWKS5500E: A required configuration parameter authorizationEndpoint is missing or has an invalid value null`
5) Update the server config again - replace the `discoveryEndpointUrl` value with the original good discovery endpoint.
6) Note the message in the server's messages.log: `CWWKS5500E: A required configuration parameter authorizationEndpoint is missing or has an invalid value null`
7) Attempt to access a social login protected app, you'll receive a 403 in the response and find this message in the server log: `CWWKS5500E: A required configuration parameter authorizationEndpoint is missing or has an invalid value null`

It looks like the initial config sets up the _discoveryDocumentHash_ which is used to populate the config.
The second config finds that the discovery endpoint is not set, so, it skips using the _discoveryDocumentHash_ and tries to init the config from the values found in server.xml.
The hash instance is still set to the values from the first config.
The third config comes along and runs discovery.  If then compares the current _discoveryDocumentHash_ to the still existing _discoveryDocumentHash_ - they're the same, so, it skips setting the values into the config object. 

We need to do one of two things:
1) when we find that the discovery endpoint is not set, we should discard the _discoveryDocumentHash_ (I'm not sure if we can get to the hash at that point to do this)
2) stop checking if the _discoveryDocumentHash_ values are the same and if set, always set the config values from the hash. (I don't know why that check was put in place - could we be breaking something else if we make this change)

ayoho:
Resolves #16994

This now sets the `discoveryDocumentHash` in `OidcLoginConfigImpl` back to null if the `discoveryEndpointUrl` is dynamically removed from the configuration or set to an empty string after discovery has already been performed. This should ensure that the various endpoint information in the discovery data gets re-populated if the `discoveryEndpointUrl` is dynamically re-added.

### DIFF
--- a/dev/com.ibm.ws.security.social/src/com/ibm/ws/security/social/internal/OidcLoginConfigImpl.java
+++ b/dev/com.ibm.ws.security.social/src/com/ibm/ws/security/social/internal/OidcLoginConfigImpl.java
@@ -139,19 +139,19 @@ public class OidcLoginConfigImpl extends Oauth2LoginConfigImpl implements Conver
 
     @Override
     protected void setAllConfigAttributes(Map<String, Object> props) throws SocialLoginException {
-        this.clientId = configUtils.getConfigAttribute(props, KEY_clientId);
-        this.clientSecret = configUtils.processProtectedString(props, KEY_clientSecret);
-        this.useSystemPropertiesForHttpClientConnections = configUtils.getBooleanConfigAttribute(props, KEY_USE_SYSPROPS_FOR_HTTPCLIENT_CONNECTONS, false);
-        this.sslRef = configUtils.getConfigAttribute(props, KEY_sslRef);
-        this.discoveryEndpointUrl = configUtils.getConfigAttribute(props, KEY_DISCOVERY_ENDPOINT);
+        clientId = configUtils.getConfigAttribute(props, KEY_clientId);
+        clientSecret = configUtils.processProtectedString(props, KEY_clientSecret);
+        useSystemPropertiesForHttpClientConnections = configUtils.getBooleanConfigAttribute(props, KEY_USE_SYSPROPS_FOR_HTTPCLIENT_CONNECTONS, false);
+        sslRef = configUtils.getConfigAttribute(props, KEY_sslRef);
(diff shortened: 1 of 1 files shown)


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat,com.ibm.ws.security.social_fat.LibertyOP,com.ibm.ws.security.social_fat.delegated,com.ibm.ws.security.social_fat.multiProvider,com.ibm.ws.security.social_fat.okdServiceLogin,com.ibm.ws.security.social_fat.OpenShift

dev/com.ibm.ws.security.social/src/com/ibm/ws/security/social/internal/OidcLoginConfigImpl.java
```diff
@@ -163,45 +163,49 @@ protected void setAllConfigAttributes(Map<String, Object> props) throws SocialLo
...
         } else {
-            this.userInfoEndpoint = configUtils.getConfigAttribute(props, KEY_USERINFO_ENDPOINT);
-            this.authorizationEndpoint = getRequiredConfigAttribute(props, KEY_authorizationEndpoint);
-            this.tokenEndpoint = configUtils.getConfigAttribute(props, KEY_tokenEndpoint);
-            this.jwksUri = configUtils.getConfigAttribute(props, KEY_jwksUri);
-            this.issuer = configUtils.getConfigAttribute(props, KEY_ISSUER);
+            if (tc.isDebugEnabled()) {
+                Tr.debug(tc, "discoveryEndpointUrl is null");
+            }
+            discoveryDocumentHash = null;
```
ayoho, line 169: This is the only functional change.

teddyjtorres (approved):

PR #17746, opened 2021-07-01
config: fix getObscuredDefaultValue

Azquelt:
Previously, if getObscuredDefaultValue decided that the value did not
need to be obscured, it would return the value, rather than the
defaultValue.


Azquelt:
@brenthdaniel I found this because my trace was displaying the wrong default value. I'm not sure whether the problem would manifest anywhere user visible, so I'm not sure whether this needs a `release bug` issue.

Azquelt:
#build

brenthdaniel (approved):

Azquelt:

PR #17761, opened 2021-07-02
mpOpenApi-2.0: ignore non-public properties by default

### LINKED ISSUE #17428: MicroProfile OpenAPI 2.0 includes non-public fields in the generated documentation
**Describe the bug**

When a schema is created for a class which includes a private field, the private field is listed in the generated OpenAPI document

E.g.
```java
@Schema
public class Example {
    private String field1;
    public String field2;
}
```

results in
```yaml
  schemas:
    Example:
      type: object
      properties:
        field1:
          type: string
        field2:
          type: string
```

**Expected behavior**
```yaml
  schemas:
    Example:
      type: object
      properties:
        field2:
          type: string
```

**Diagnostic information:**

**Additional context**

An option was added to smallrye open-api to enable the correct behaviour for private properties under smallrye/smallrye-open-api#418 but the default behaviour will not change until the next major version to ensure backwards compatibility. A second issue smallrye/smallrye-open-api#825 is open for protected and package-scoped properties

Once we update to a version of smallrye-open-api with the fix in #17440, it will be possible to set the property `mp.openapi.extensions.smallrye.private-properties.enable = false` to get the correct behaviour for private fields, but protected and package-scoped fields are still included.

We _could_ default this property to `false` ourselves, though this would risk breaking anyone relying on the current behaviour.

There are instances where users might _want_ private fields included:
* the fields and methods used by Json-B for serialization can be changed using `@JsonbVisibility` or globally using config
* if Jackson is used instead of Json-B, private fields can be included by annotating them with `@JsonProperty`

In these situations, if `mp.openapi.extensions.smallrye.private-properties.enable = false` is set, smallrye-open-api does not automatically use these annotations to decide which fields to include.

In either case, users can work around the problem by manually annotating fields they want included with `@Schema` and annotating those they don't want included with `@Schema(hidden=true)`.

Azquelt:
Fixes #17428

* ~Updates smallrye-open-api to 2.1.7~ (already done elsewhere)
* sets the property to disable non-public properties by default
* adds tests

### DIFF
--- a/dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/FATSuite.java
+++ b/dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/FATSuite.java
@@ -15,6 +15,7 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 import org.junit.runners.Suite.SuiteClasses;
 
+import com.ibm.ws.microprofile.openapi.fat.annotations.AnnotationProcessingTest;
 import com.ibm.ws.microprofile.openapi.fat.filter.FilterConfigTest;
 import com.ibm.ws.microprofile.openapi.validation.fat.OpenAPIValidationTestFive;
 import com.ibm.ws.microprofile.openapi.validation.fat.OpenAPIValidationTestFour;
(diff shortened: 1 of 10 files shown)


Azquelt:
There's currently a problem with this. The default value added by `defaultInstances.xml` isn't picked up by MicroProfile Config.

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

Azquelt:
#build

tevans78 (approved):

Azquelt:
#run-libby-bot

PR #17780, opened 2021-07-06
Fix ApacheDS stop and add port checks

kristip17:
Some FATs on windows for `com.ibm.ws.security.wim.adapter.ldap_fat.krb5` are having startup failures. The HTTP team traced it back to possible issues with the ApacheDS servers not stopping correctly. It looks like there was a bug on calling stop. I also added port checks to make sure they are free as expected in case updating the stop call doesn't fix it.

For RTC 282634


kristip17:
#build

kristip17:
All acme_fat tests ran successfully. One `build.featureStart_fat0` child didn't complete, but all others completed.

jvanhill (approved):

PR #17782, opened 2021-07-06
Fix OpenLiberty build issue for JPA 2.0 FAT

dazey3:
Running the JPA 2.0 FAT locally with only OpenLiberty, I can see that the results indicate no tests returned any results. For gradle, that does not result in test failures or exception, but for internal builds using only OpenLiberty, the JPA 2.0 FAT fails when no jpa-2.0 feature is present and no tests are executed as a result. 

For that reason, this change moves the JPA 2.0 FAT tests into a separate Suite so that the `componenttest.custom.junit.runner.AlwaysPassesTest` can still execute and return a result, allowing the internal builds to not fail. Using this change, I can observe that now at least 1 result is returned for the `componenttest.custom.junit.runner.AlwaysPassesTest` when run locally.


olendvcook (approved):

dazey3:
#build

jgrassel (approved):

PR #17785, opened 2021-07-07
Sleep for revoke completion on Boulder for Acme

kristip17:
It appears that there is some variability on when the revoke is complete on the Boulder server. Add additional time before restarting the server.

RTC 284855


kristip17:
#build

jvanhill (approved):

PR #17806, opened 2021-07-08
fix xml-apis jar version

c00crane:
Make sure we pick u the 1.4 version of he xml-apis jar instead of the version specified by the xalan serializer which is 1.3.4


c00crane:
#build

ayoho (approved):

PR #17814, opened 2021-07-09
Ping LDAP remote servers and remove any that are down for FATs

kristip17:
We frequently have the problem that some remote LDAPs are down and they take a long time to fail (Read timed out vs connection refused), while we do failover to a valid LDAP, if we have to timeout everytime, the FAT times out. 

Test out the LDAPs returned from Consul and remove any that don't ping. If we have no running LDAPs (for an LDAP group type), throw an exception and fail back to running local, if allowed by the suite/build type.

RTC 284763

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
@@ -18,6 +18,7 @@ import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Map.Entry;
@@ -543,6 +544,32 @@ public class LDAPUtils {
         for (int requested = count; requested > 0; requested--) {
             try {
                 List<ExternalTestService> services = new ArrayList<ExternalTestService>(ExternalTestService.getServices(requested, service));
+
+                /*
+                 * Remove services that are not available.
+                 */
+                Set<ExternalTestService> toRemove = new HashSet<ExternalTestService>();
+                for (ExternalTestService serviceToPing : services) {
+                    try {
+                        if (!isLdapServerAvailable(serviceToPing.getAddress(),
+                                                   serviceToPing.getProperties().get(CONSUL_LDAP_PORT_KEY), false,
+
+                                                   serviceToPing.getProperties().get(CONSUL_BIND_DN_KEY),
+                                                   serviceToPing.getProperties().get(CONSUL_BIND_PASSWORD_KEY))) {
+                            Log.warning(c, "For service, " + service + ", Ldap at " + serviceToPing.getAddress() + " failed to ping");
+                            toRemove.add(serviceToPing);
+                        }
+                    } catch (Exception e) {
+                        Log.warning(c, "For service, " + service + ", Ldap at " + serviceToPing.getAddress() + " failed to ping with exception: " + e.getMessage());
+                        toRemove.add(serviceToPing);
+                    }
+                }
+                if (services.size() == toRemove.size()) {
+                    throw new Exception("For service, " + service + ", no LDAPs responded to ping, follow regular failover path which will run local LDAP if possible.");
+                } else if (!toRemove.isEmpty()) {
+                    Log.warning(c, "For service, " + service + ", some LDAPS failed to ping, will remove from services list and run with good LDAPs only.");
+                    services.removeAll(toRemove);
+                }
 
                 /*
                  * Copy unique instances to fill out the requested count of services.


kristip17:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.wim.adapter.ldap_fat

jvanhill (changes requested):
If there is only one available, use it for all the slots instead of re-ordering. We already do this if we ask for n servers but Consul returns less than n servers and this is also eventually what would happen if we add LDAP/LDAPS statusing to Consul.

kristip17:
Single PB failure is a known issue.

kristip17:
#libby

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
```diff
@@ -201,20 +201,39 @@ private static void initializeRemoteLdapServers() throws Exception {
         try {
             services = getLdapServices(2, CONSUL_LDAP_AD_SVT_SERVICE);
 
+            /**
```
jvanhill: It seems like all this logic should be in `getLdapServices` so we don't have to replicate it.

If you check that method you will see we request X LDAP services from Consul, if we can't get X, we try X-1, etc until we either get at least 1 LDAP service.

I think we could do something like:

```
                List<ExternalTestService> services = new ArrayList<ExternalTestService>(ExternalTestService.getServices(requested, service));

                // NEW CODE STARTS HERE!!!!
                /*
                 * Remove services that are not available.
                 */
                Set<ExternalTestService> toRemove = new HashSet<ExternalTestService>();
                for (ExternalTestService service : services) {
                    if (!isLdapServerAvailable(service...) ) {
                       toRemove.add(service);
                    }
                }
                services.removeAll(toRemove);
                
                if (services.isEmpty()) {
                   // NO AVAILABLE LDAP SERVERS, FAIL BACK TO LOCAL????
                }
                // NEW CODE STOPS HERE!!!!
                
                /*
                 * Copy unique ...
                 */
```

Note: the isLdapServerAvailable may never return false. Looking at it, it appears to throw an Exception so that code above may need to change, or the method should perhaps be written better.

jvanhill (changes requested):
See my comments. Instead of re-ordering, lets remove all the bad servers. This code can go into the getLdapServices method, where the code can be common. That method already does some handling of filling out the required LDAP servers if enough are not available, so we can take advantage of some of the logic already there in the case that 1 is down, and the other is up. This will also have the added advantage of allowing us to fail over to local LDAP if no servers are available. It also eliminates the need to make assumptions about which LDAP server mosts tests are using as primary.

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
kristip17: Made updates -- `isLdapServerAvailable` usually throws an Exception. Since the method is called in several areas, I decided the risk wasn't worth the reward to rewrite it right now.

kristip17:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.wim.adapter.ldap_fat

kristip17:
PB Failures are known issues.

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
```diff
@@ -544,6 +545,32 @@ private static void logLdapServerInfo() {
...
+                        if (!isLdapServerAvailable(serviceToPing.getAddress(),
+                                                   serviceToPing.getProperties().get(CONSUL_LDAP_PORT_KEY), false,
+
+                                                   serviceToPing.getProperties().get(CONSUL_BIND_DN_KEY),
+                                                   serviceToPing.getProperties().get(CONSUL_BIND_PASSWORD_KEY))) {
+                            Log.warning(c, "Ldap at " + serviceToPing.getAddress() + " failed to ping");
+                            toRemove.add(serviceToPing);
+                        }
+                    } catch (Exception e) {
+                        Log.warning(c, "Ldap at " + serviceToPing.getAddress() + " failed to ping with exception: " + e.getMessage());
```
jvanhill: Add service name we are looking up to help facilitate in debug.

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
```diff
@@ -544,6 +545,32 @@ private static void logLdapServerInfo() {
...
+                        toRemove.add(serviceToPing);
+                    }
+                }
+                if (services.size() == toRemove.size()) {
+                        throw new Exception("No LDAPs responded to ping, follow regular failover path which will run local LDAP if possible.");
```
jvanhill: Add service name we are looking up to help facilitate in debug.

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
```diff
@@ -544,6 +545,32 @@ private static void logLdapServerInfo() {
...
+                } else if (!toRemove.isEmpty()) {
+                        Log.warning(c, "Some LDAPS failed to ping, will remove from services list and run with good LDAPs only.");
```
jvanhill: Add service name we are looking up to help facilitate in debug.

jvanhill (approved):
It would be nice to have the service name we had trouble pinging the servers for in the exceptions and trace so it is clear what group the server is failing for.

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
kristip17: Added

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
kristip17: Added

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
kristip17: Added

kristip17:
Failures are known issues.

kristip17:
#libby

PR #17822, opened 2021-07-12
update testcontainer bundle

KyleAure:
Ensure that `META-INF/services` are copied from testcontainers-core project into the shared `io.openliberty.org.testcontainers` bundle.  This will ensure that all docker client services are registered and available. 

Switch back to using a custom gradle task to include runtime dependency on the `io.openliberty.org.testcontainers` project.  When using the `requiredLibs` configuration gradle would copy the project jar and also all of the dependencies meaning that most dependencies were being loaded twice at runtime. This should improve memory and performance.

### DIFF
--- a/.github/test-categories/CLASSLOADING
+++ b/.github/test-categories/CLASSLOADING
@@ -1,0 +1,1 @@
+com.ibm.ws.classloading.bells_fat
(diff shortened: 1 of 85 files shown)


gjwatts (approved):

KyleAure:
#build

KyleAure:
#libbybot

PR #17866, opened 2021-07-15
criu: need SnapShotHook to support handling .sCommand under multiple restores

sebratton:


sebratton:
This was opened in error

Issue #16073, opened 2021-02-28
Password is exposed as clear text in the logs when running OL wssecurity CXF FAT with EE8


meiaus:
Per discussion with @arunavemulapalli, we're tracking this issue for further action.

@neuwerk @BerksanAtes @sawadood,
Please let me know if any test logs are needed here.

jimmy1wu:
changes in https://github.com/OpenLiberty/open-liberty/pull/17871 allowed us to put the password tests back into CxfUNTBasicTests

however, there are still other places where the password is logged, which needs to be investigated further

PR #17871, opened 2021-07-15
hide passwords in cxf FAT logs

jimmy1wu:
for https://github.com/OpenLiberty/open-liberty/issues/16073

### DIFF
--- a/dev/com.ibm.ws.org.apache.cxf.cxf.rt.bindings.soap.3.2/src/org/apache/cxf/binding/soap/interceptor/SoapActionInInterceptor.java
+++ b/dev/com.ibm.ws.org.apache.cxf.cxf.rt.bindings.soap.3.2/src/org/apache/cxf/binding/soap/interceptor/SoapActionInInterceptor.java
@@ -44,6 +44,9 @@ import org.apache.cxf.service.model.BindingOperationInfo;
 import org.apache.cxf.service.model.OperationInfo;
 import org.apache.cxf.ws.addressing.JAXWSAConstants;
 
+import com.ibm.websphere.ras.annotation.Trivial; // Liberty code change
+
+@Trivial // Liberty code change
 public class SoapActionInInterceptor extends AbstractSoapInterceptor {
 
     private static final Logger LOG = LogUtils.getL7dLogger(SoapActionInInterceptor.class);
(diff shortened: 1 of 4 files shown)


jimmy1wu:
#build

jimmy1wu:
#build

jimmy1wu:
#build

jimmy1wu:
#libby

jimmy1wu:
@arunavemulapalli @neuwerk 

changes allow us to put the password tests back into `CxfUNTBasicTests`.

however, a quick search after running the build still shows a bunch of other places where the password is logged 

might be good to merge this pr to re-enable the password tests in `CxfUNTBasicTests`, but leave the issue open to investigate the other places where the password is logged further

dev/com.ibm.ws.org.apache.cxf.cxf.rt.bindings.soap.3.2/src/org/apache/cxf/binding/soap/interceptor/SoapActionInInterceptor.java
```diff
@@ -251,6 +254,7 @@ public void handleMessage(SoapMessage message) throws Fault {
             if (boi == null) {
                 return;
             }
+            LOG.info("boi = " + boi + " action = " + action + " message = " + message); // Liberty code change
```
neuwerk: I'm not sure if we actually need to print this message, but if it helps for debug purposes, you should change it to `LOG.fine()` in order to prevent unnecessary logging in the message.log.  LOG.info() can often have a huge hit on performance, so we should only use it when absolutely necessary for usability.

neuwerk (changes requested):
Everything looks good with the exception of the one LOG.info() call. See my comments below. Thanks for doing this Jimmy!

jimmy1wu:
#libby
#build

dev/com.ibm.ws.org.apache.cxf.cxf.rt.bindings.soap.3.2/src/org/apache/cxf/binding/soap/interceptor/SoapActionInInterceptor.java
jimmy1wu: thanks for the suggestion, i have updated it to use `LOG.fine()`

in terms of the message, i was mostly going off of what was being logged in the 2.6.2 version
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.org.apache.cxf.cxf.rt.bindings.soap.2.6.2/src/org/apache/cxf/binding/soap/interceptor/SoapActionInInterceptor.java#L272

neuwerk (approved):
Changes look good to me!

PR #17874, opened 2021-07-15
Add wait in EJB automatic timer FAT

tkburroughs:
Normal setup during a local full run takes 4 minutes and the test is only waiting 5,
so it seems reasonable to have the test wait another minute.... and to have the additional
minute be a separate wait after the first 5 minute wait.  By adding a second wait for
1 minute, the second wait will occur completely after any server pause, at which time
the timers are likely running rapidly to catchup from the pause... and thus very likely
they will all have caught up by the time the test checks again after a second wait time.


tkburroughs:
#build

olendvcook (approved):

PR #17877, opened 2021-07-16
Fix Race Condition Regarding JSP's classloaderCreated variable

### LINKED ISSUE #17828: Update JSP Logic to Avoid Race Condition Regarding trackDependencies
After #17532, it was discovered that JSP may not always update the dependentsList properly when JSP handles concurrent requests.  This means dependents are not tracked, and the JSP fails to updated when any changes occur to those include JSPs. 

- Occurs when trackDependencies is true.
 - Two simultaneous  requests come in to the same jsp with after the server starts up. 
   - First request comes in, the page is translated/compiled, and classloaderCreated is set as true within it’s synchronous block.
   - Second requests comes in just after the first request leaves the synchronous block. and then classloaderCreated is to to false. 
- The class object containing this variable is a the same for the two requests so classloaderCreated also changes for the first request.
classloaderCreated needs to be true for the check within loadClassInformation as that’s where the dependentsList is populated.  If it’s not, then future requests to that page won’t check if the dependents are outdated (since it’s empty).

Scenarios where this can occur:
- Large Enough JSP.  Code is synchronized, and the translation/compilation take a while, This creates a backlog of requests waiting to go into that sync block.  When the first request leaves, the next comes comes in and sets classloaderCreated back to false.  Then that first request, fails to add dependents.  Then dependencies are not tracked.  
- High Load ( Might not be ideal to update the JSP during this time)

volosied:
fixes #17828

### DIFF
--- a/dev/com.ibm.ws.jsp.2.3_fat/fat/src/com/ibm/ws/jsp23/fat/tests/JSPTests.java
+++ b/dev/com.ibm.ws.jsp.2.3_fat/fat/src/com/ibm/ws/jsp23/fat/tests/JSPTests.java
@@ -781,7 +781,7 @@ public class JSPTests {
         String url = JSPUtils.createHttpUrlString(server, TestEDR_APP_NAME, "index.jsp");
         LOG.info("url: " + url);
 
-        runEDR(url, true);
+        runEDR(url, false);
     }
 
     /**
@@ -800,7 +800,7 @@ public class JSPTests {
         runEDR(url, true);
     }
 
-    private void runEDR(String url, boolean makeConcurrentRequests){
+    private void runEDR(String url, boolean makeConcurrentRequests) throws Exception {
         String expect1 = "initial EDR header";
         String expect2 = "updated EDR header";
         String orgEdrFile = "headerEDR1.jsp";
@@ -815,7 +815,7 @@ public class JSPTests {
 
         if(makeConcurrentRequests) {
             // Make 2 requests. 
-            makeConcurrentRequests(wc1, request1, 2)
+            makeConcurrentRequests(wc1, request1, 2);
         }
 
         WebResponse response1 = wc1.getResponse(request1);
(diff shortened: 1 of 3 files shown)


volosied:
#build #spawn.fullfat.buckets=com.ibm.ws.jsp.2.3_fat, com.ibm.ws.jsf.2.2_fat, com.ibm.ws.jsf.2.3_fat, com.ibm.ws.webcontainer.servlet.4.0_fat, com.ibm.ws.jsp_fat

volosied:
#build

dev/com.ibm.ws.jsp.2.3_fat/test-applications/TestEDR.war/resources/trackDependencies.jsp
```diff
@@ -0,0 +1,31 @@
...
+<meta http-equiv="Content-Type" content="text/html; charset=ISO-8859-1">
+<%@ taglib uri = "http://java.sun.com/jsp/jstl/core" prefix = "c" %>
+<title>Test trackDependencies</title>
+</head>
+<body>
+<p>This tests trackDependencies under concurrent requests.</p>
+
+<!-- Slow Down JSP Translation By Just a Bit  -->
+<c:forEach var = "i" begin = "1" end = "5"> </c:forEach>
+<c:forEach var = "i" begin = "1" end = "5"> </c:forEach>
```
volosied, line 25: Probably doesn't do much, but I thought it help would the chances for this issue to reoccur.

epj (approved):
verified in my workspace too

dev/com.ibm.ws.jsp.2.3_fat/fat/src/com/ibm/ws/jsp23/fat/tests/JSPTests.java
```diff
@@ -802,6 +836,34 @@ public void testEDR() throws Exception {
...
+                    return true;
+                }
+            }));
+        }
+        
+        // check runs completed successfully
+        for (Future<Boolean> task : tasks) {
+            try {
+                if (!task.get())
+                    throw new Exception("0");
```
wtlucy, line 862: are the `"0"` and `"1"` just used for differentiation?

volosied: I think so. I left this alone from the other example, but I can change it if you prefer? 

https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Future.html#get--

wtlucy (approved):

volosied:
#run-libby-bot

dev/com.ibm.ws.jsp.2.3_fat/test-applications/TestEDR.war/resources/trackDependencies.jsp
volosied: Update -- more local testing shows that it does matter.

PR #17884, opened 2021-07-16
Do not use soap 1.2 methods to process soap 1.1 SoapFault

sawadood:
This PR is created for fixing issue: 17883.

If an application sends a SOAP 1.1 SOAPFaultException to Liberty server, it tries to process it as a SOAP 1.2 SoapFault which results in the following error: SAAJ0303: Operation getFaultSubcodes not supported by SOAP 1.1.


sawadood:
#build Let's do a personal build

neuwerk (approved):
Changes look good!

Issue #17545, opened 2021-06-15
Investigate jsfContainer-2.3 with MyFaces 2.3-Next


volosied:
Worth looking into this configuration to see if jsfContainer is compatible with MyFaces 2.3-Next. Initial testings shows it is not, but it may be a trivial fix to address this problem (or not).

MyFaces 2.3-Next is not necessarily the supported implementation as opposed to the preferred regular 2.3.x implementation. Nevertheless, some users may find it useful.

pnicolucci:
We already offer a facesContainer-3.0 and facesContainer-4.0 features so this work isn't necessary at this time. Closing the issue.

PR #17898, opened 2021-07-19
jsfContainer MyFaces 2.3-next-* version string support

wtlucy:
Testing with basic apps shows that MyFaces `2.3-next` works correctly with `jsfContainer-2.3` once the jsfContainer version string check is updated in 1eb8b21867f3557ae92a1c0ce9e6b584396749fe. However in order to test this properly we need to update `com.ibm.ws.jsfContainer_fat_2.3` to test `2.3-next`,  which will include determining which tests in that bucket are actually valid for `2.3-next` (ie. not dependent on JSF managed beans).


volosied:
Linking #17545

PR #17901, opened 2021-07-19
identifyException should not be top level config

### LINKED ISSUE #17591: identifyException accidentally externalized as unusable top level config element
identifyException was intended to be available only as a nested element of dataSource,

```
<dataSource jndiName="jdbc/ds1" jdbcDriverRef="mydriver">
  <identifyException sqlState="08000" as="StaleConnection"/>
  <identifyException sqlState="08004" as="StaleConnection"/>
  <properties .../>
</dataSource>
```

When I went to promote this nested element from nonship to beta, I discovered that due to the way in which the metatype was written, the developer had accidentally exposed identifyException as an unusable top level element, which made it into Open Liberty doc.  See https://openliberty.io/docs/21.0.0.1/reference/config/identifyException.html

This means that a user could currently write the following as valid top level config, which would be a no-op because it doesn't apply to anything:
```
<identifyException sqlState="08000" as="StaleConnection"/>
```

Options:

1. Remove `identifyException` as top level config and add an exclude to the MetaTypeValidator. This shouldn't cause any real errors because the top level element never did anything in the first place, although schema-based config editors would now start noticing it as bad config, which in a way is helpful because it would inform users that their top level `identifyException` doesn't do anything.
2. Leave `identifyException` as useless top level config, which the server will continue to ignore. The usability problem of having a useless configuration element will remain.
3. Allow `identifyException` as top level config and come up with a way to make it useful by giving dataSource an identifyExceptionRef attribute. This complicates the configuration somewhat, but at least has a valid use case behind it.  It would unfortunately force us to switch away from the flattened config optimization for identifyException. 

Preferred option 1 was approved by the POC.

njr-11:
fixes #17591

### DIFF
--- a/dev/com.ibm.ws.jdbc.metatype/resources/OSGI-INF/metatype/metatype.xml
+++ b/dev/com.ibm.ws.jdbc.metatype/resources/OSGI-INF/metatype/metatype.xml
@@ -108,8 +108,7 @@
   <Object ocdref="com.ibm.ws.jdbc.dataSource.identifyException"/>
  </Designate>
 
- <!-- TODO This should not have been externalized as a top level element, where it is unusable. Need to determine if we we are stuck with it. -->
- <OCD id="com.ibm.ws.jdbc.dataSource.identifyException" ibm:alias="identifyException" name="%identifyException" description="%identifyException.desc" ibmui:localization="OSGI-INF/l10n/metatype">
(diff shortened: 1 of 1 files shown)


njr-11:
#build

jmstephensgit (approved):

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build complains of a timeout starting parallel execution, which was due to an outage, however, the async test runs did eventually start after that and all passed.

PR #17906, opened 2021-07-20
Refactor to remove SnapshotFailed Exception

### LINKED ISSUE #17754: Rework Checkpoint and ExecuteCRIU interfaces
Open J9 is proposing new API that allows us to use CRIU to checkpoint a JVM process.  So far the following enum values are being proposed for result of performing a checkpoint:

```
	public static enum CRIUResultType {
		SUCCESS,
		UNSUPPORTED_OPERATION,
		INVALID_ARGUMENTS,
		SYSTEM_CHECKPOINT_FAILURE,
		JVM_CHECKPOINT_FAILURE,
		JVM_RESTORE_FAILURE;
	}
```
We need to rework the `io.openliberty.checkpoint.internal.criu.ExecuteCRIU` internal interface that is used to abstract the way we invoke CRIU.  Currently there are two implementations of this:

- `io.openliberty.checkpoint.openj9.internal.ExecuteCRIU_OpenJ9` - that uses Open J9 to do the checkpoint
- `io.openliberty.jigawatts.internal.ExecuteCRIU_JNI` - that uses jigawatts to do the checkpoint

The `io.openliberty.checkpoint.internal.criu.ExecuteCRIU` interface is going to need to be changed to return something more advanced than a simple `int` to indicate success or failure.  We could change `io.openliberty.checkpoint.spi.SnapshotFailed` to be `io.openliberty.checkpoint.spi.SnapshotResult` instead to mimic the `org.eclipse.openj9.criu.CRIUSupport.CRIUResult` being proposed in Open J9.

So we would end up with something like this:

```java
@ProviderType
public class SnapshotResult {
    public enum Type {
        UNSUPPORTED,
        PREPARE_ABORT,
        SNAPSHOT_FAILED,
        RESTORE_ABORT;
    }

    private final int errorCode;
    private final Type type;
    private final String msg;
    private final Exception cause;

    public SnapshotFailed(Type type, String msg, Exception cause, int errorCode) {
        this.msg = msg;
        this.cause = cause;
        this.type = type;
        this.errorCode = errorCode;
    }
   // getter methods

}
```
We could then enhance our enum Type with the result types form j9 `INVALID_ARGUMENTS, SYSTEM_CHECKPOINT_FAILURE, JVM_CHECKPOINT_FAILURE, JVM_RESTORE_FAILURE`

The `ExecuteCRIU` dump method would then look like this:
```java
    SnapshotResult dump(File directory);
```

We would also update the `io.openliberty.checkpoint.spi.Checkpoint.snapshot(Phase)` method to look like this:

```java
    SnapshotResult snapshot(Phase phase);
```

The `CheckpointImpl` class needs to be updated to properly handle prepare errors vs. restore errors from Open J9.

The following error types from Open J9 should result in aborting any prepare hooks called and then return a proper SnapshotResult indicating the failure to checkpoint:
- UNSUPPORTED_OPERATION
- INVALID_ARGUMENTS
- SYSTEM_CHECKPOINT_FAILURE
- JVM_CHECKPOINT_FAILURE

For `JVM_RESTORE_FAILURE` it is unclear what we should do with respect to our own restore hooks.  This type would be returned when CRIU was able to successfully restore the process but then the JVM itself had some issues with its own restore hooks before it could return execution back to us.  At this point the JVM probably is in an unstable state and we likely should just return a `SnapshotResult` indicating the JVM failure.  Ultimately this probably should result in us exiting out of the Java process.

anjumfatima90:
fixes  #17754


anjumfatima90:
#build

dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/Checkpoint.java
```diff
@@ -54,10 +54,8 @@ public static Phase getPhase(String p) {
     /**
      * Asserts that this platform supports checkpoint. Depending on implementation may
      * call out to native library support.
-     *
-     * @throws SnapshotFailed if support for snapshot functionality not available.
      */
-    default void checkpointSupported() throws SnapshotFailed {
+    default void checkpointSupported() {
```
tjwatson, line 58: This likely should be renamed to `isCheckpointSupported` and return a boolean.  We should then update `ExecuteCRIU` interface to also have a corresponding method that the `CheckpointImpl` can call to get the answer.

dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/SnapshotResult.java
```diff
@@ -0,0 +1,63 @@
...
+import org.osgi.annotation.versioning.ProviderType;
+
+@ProviderType
+public class SnapshotResult {
+
+    public enum SnapshotResultType {
+        SUCCESS(1),
+        UNSUPPORTED_OPERATION(-1),
+        INVALID_ARGUMENTS(-2),
+        SYSTEM_CHECKPOINT_FAILURE(-3),
```
tjwatson, line 23: I assume this is the type used when criu itself fails.  In that case we will probably have access to the actual error code from criu.  We may need to codes here.  One that is the liberty exit code and one that is the error code from criu dump result.

dev/io.openliberty.checkpoint.openj9/src/io/openliberty/checkpoint/openj9/internal/ExecuteCRIU_OpenJ9.java
```diff
@@ -42,17 +42,51 @@ public void stop(BundleContext bc) {
...
+
+            case UNSUPPORTED_OPERATION:
+                snapshotResult = createSnapshotResult(SnapshotResultType.UNSUPPORTED_OPERATION, "The criu dump command failed with error: Unsupported Operation");
+
+            case INVALID_ARGUMENTS:
+                snapshotResult = createSnapshotResult(SnapshotResultType.INVALID_ARGUMENTS, "The criu dump command failed with error: Invalid Arguments");
+                break;
+
+            case SYSTEM_CHECKPOINT_FAILURE:
+                snapshotResult = createSnapshotResult(SnapshotResultType.SYSTEM_CHECKPOINT_FAILURE, "The criu dump command failed with error: System Checkpoint Failure");
```
tjwatson, line 68: Here I expect j9 will get some updates to give us access to the error code from `criu dump` operation

dev/io.openliberty.checkpoint.openj9/src/io/openliberty/checkpoint/openj9/internal/ExecuteCRIU_OpenJ9.java
```diff
@@ -42,17 +42,51 @@ public void stop(BundleContext bc) {
...
-        return -50;
+        return snapshotResult;
+    }
+
+    /**
+     * @return
+     */
+    private SnapshotResult createSnapshotResult(SnapshotResultType type, String msg) {
+        SnapshotResult snapshotResult;
+        snapshotResult = new SnapshotResult(type, msg, null);
```
tjwatson, line 88: this is overly verbose.  Why not just have:
```
        return new SnapshotResult(type, msg, null);
```

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -79,53 +79,66 @@ public CheckpointImpl(ComponentContext cc) {
...
+     * @returns SnapshotResult if the snapshot fails
      */
     @Deprecated
     @Descriptor("Take a snapshot")
-    public void snapshot(@Parameter(names = "-p", absentValue = "") @Descriptor("The phase to snapshot") Phase phase,
-                         @Descriptor("Directory to store the snapshot") File directory) throws SnapshotFailed {
-        doSnapshot(phase, directory);
+    public SnapshotResult snapshot(@Parameter(names = "-p", absentValue = "") @Descriptor("The phase to snapshot") Phase phase,
+                                   @Descriptor("Directory to store the snapshot") File directory) {
+        SnapshotResult result = doSnapshot(phase, directory);
```
tjwatson, line 88: just:
```
        return doSnapshot(phase, directory);
```

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -144,57 +157,56 @@ private void doSnapshot(Phase phase, File directory) throws SnapshotFailed {
...
+        SnapshotResult result = new SnapshotResult(SnapshotResultType.SUCCESS, "Prepare Success", null);
         for (SnapshotHook snapshotHook : snapshotHooks) {
             try {
                 perform.accept(snapshotHook);
                 called.add(snapshotHook);
             } catch (Exception abortCause) {
                 for (SnapshotHook abortHook : called) {
                     try {
-                        abort.accept(abortHook, abortCause);
+                        result = snapshotResult.apply(abortCause);
```
tjwatson, line 173: Should only call snapshotResult.apply(abortCause) once before the for loop

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -79,53 +79,66 @@ public CheckpointImpl(ComponentContext cc) {
...
-            if (e instanceof SnapshotFailed) {
-                throw (SnapshotFailed) e;
+
+            if (dumpResult.getType() == SnapshotResultType.SUCCESS) {
+                SnapshotResult restoreResult = restore(phase, snapshotHooks);
+                if (restoreResult.getType() != SnapshotResultType.SUCCESS) {
+                    return restoreResult;
+                }
+            } else if (dumpResult.getType() == SnapshotResultType.JVM_RESTORE_FAILURE) {
+                return new SnapshotResult(SnapshotResultType.JVM_RESTORE_FAILURE, "Unexpected JVM failure.", null);
```
tjwatson, line 132: Why create a new SnapshotResult instead of returning the `dumpResult` directly here?

tjwatson (commented):
I think we should move the Checkpoint interface to the package `io.openliberty.checkpoint.internal` and move that to the `io.openliberty.checkpoint` project itself as in internal implementation detail for that bundle.  The Checkpoint service should only be called by `io.openliberty.checkpoint.internal.CheckpointFeatures` and `io.openliberty.checkpoint.internal.CheckpointApplications` components so we might as well make the Checkpoint service internal and only have the hook and result classes left in `io.openliberty.checkpoint.spi`

I would probably then rename

- SnapshotHook -> CheckpointHook
- SnapshotHookFactory -> CheckpointHookFactory
- SnashotResult -> CheckpointResult

anjumfatima90:
Changes are no longer needed. Will do it in another PR if required in future.

PR #17918, opened 2021-07-20
282675 social fats cleanup web client

c00crane:
htmlunit has a bug where it does not clean up the web clients that each test creates.
I'm replacing the call to getWebClient with a method that not only creates a new webClient, it tracks the instance and at the end of each test, we clean up all tracked webClient.
This should resolve the OOM issue with htmlunit


c00crane:
#run-libby-bot
#build

clarkek123 (approved):
Approve contingent on good personal build results.

PR #17966, opened 2021-07-26
No obvious cause of of cached streams if junits are executed sequenti…

Channyboy:
…ally (which they should be). Change is to restore stream to original System.out/err regardless of current stream state

#build 
addresses #17768

### DIFF
--- a/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
+++ b/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
@@ -260,7 +260,7 @@ public class BaseTraceService implements TrService {
      * of system properties we expect (for FFDC and logging).
      *
      * @param config a {@link LogProviderConfigImpl} containing TrService configuration
-     *            from bootstrap properties
+     *                   from bootstrap properties
      */
     @Override
     public void init(LogProviderConfig config) {
@@ -286,10 +286,12 @@ public class BaseTraceService implements TrService {
             }
 
             @Override
-            public void flush() {}
+            public void flush() {
+            }
 
             @Override
-            public void close() {}
+            public void close() {
+            }
         });
     }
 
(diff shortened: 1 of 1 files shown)


Channyboy:
#build 
Seem an issue with the builds prevented the build from fetching source.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -1658,18 +1661,23 @@ protected void captureSystemStreams() {
...
-        if (System.out == teeOut)
-            System.setOut(systemOut.getOriginalStream());
-        if (System.err == teeErr)
-            System.setErr(systemErr.getOriginalStream());
+        /*
+         * OL17768 - No obvious cause of OL17768.
+         * Disable this check and "restore"
+         * it regardless to evaluate effects.
+         */
+        //if (System.out == teeOut)
```
fmhwong, line 1669: The if statements are for efficiency purpose only.  It doesn't change the logic.  I'm not sure this change will have any effect to the issue.

Channyboy: Right, the cause of the issue seems to be that the `System.out` is not set back to the _original_ `System.out`. It fails since the check for `System.out` shows that the reference is pointing to a `TeePrintStream` (i.e. `teeOut`).

This check is done before each Junit execution. After each Junit execution it's _supposed_ to clean up (i.e. restore the System.out to the original stream). The Junits are suppose to execute sequentially so there isn't any obvious cause of issue here (i.e. shouldn't be a concurrency issue). This seems like it's the only possible area where it _may_ go wrong ( as in for some reason it never restored the original stream during the cleanup of the Junit (which is done as part of the TestRule)). As I am unable to reproduce  the issue at all, this PR is an attempt to evaluate if this could cause any change (if any). As this is just replacing the the reference (and only would be really hit through junits as runtime we would not expect the `BaseTraceService` to be shutting down during runtime, this would not impact runtime at all.

If no change is in effect, we can chose to revert this PR.

Channyboy:
#build 
Builds are still wonky, the build took over a day to finish resulting in some timeouts :(

Channyboy:
#build 
re-running for fresh build after long weekend

fmhwong (approved):
LGTM

PR #17967, opened 2021-07-26
Refactor checkpointImpl

anjumfatima90:


dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -27,28 +28,36 @@
...
+import com.ibm.ws.kernel.feature.ServerReadyStatus;
 
 import io.openliberty.checkpoint.internal.criu.ExecuteCRIU;
-import io.openliberty.checkpoint.internal.Checkpoint;
 import io.openliberty.checkpoint.spi.CheckpointHook;
 import io.openliberty.checkpoint.spi.CheckpointHookFactory;
 import io.openliberty.checkpoint.spi.CheckpointHookFactory.Phase;
-import io.openliberty.checkpoint.spi.SnapshotFailed;
-import io.openliberty.checkpoint.spi.SnapshotFailed.Type;
+import io.openliberty.checkpoint.internal.CheckpointFailed;
```
tjwatson: imports contained in the same package as the defining class will show up as unnecessary imports.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -59,60 +68,102 @@
...
         this.cc = cc;
+        this.checkpointFeatures = "features".equals(cc.getBundleContext().getProperty(BootstrapConstants.CHECKPOINT_PROPERTY_NAME));
+        this.checkpointApplications = "applications".equals(cc.getBundleContext().getProperty(BootstrapConstants.CHECKPOINT_PROPERTY_NAME));
     }
 
     // only for unit tests
     CheckpointImpl(ComponentContext cc, ExecuteCRIU criu) {
         this.cc = cc;
         this.criu = criu;
+        this.checkpointFeatures = "features".equals(cc.getBundleContext().getProperty(BootstrapConstants.CHECKPOINT_PROPERTY_NAME));
```
tjwatson: These calls are not necessary for unit tests to set the checkpointFeatures and checkpointApplications variable.  I would think the unit tests would fail with these calls here.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -59,60 +68,102 @@
...
+     * @param phase the phase to take the checkpoint
+     * @throws CheckpointFailed if the checkpoint fails
      */
     @Deprecated
-    @Descriptor("Take a snapshot")
-    public void snapshot(@Parameter(names = "-p", absentValue = "") @Descriptor("The phase to snapshot") Phase phase,
-                         @Descriptor("Directory to store the snapshot") File directory) throws SnapshotFailed {
-        doSnapshot(phase, directory);
+    @Descriptor("Do a checkpoint")
+    public void checkpoint(@Parameter(names = "-p", absentValue = "") @Descriptor("The phase to checkpoint") Phase phase,
```
tjwatson: We should get rid of this gogo command method altogether.  I don't think we have any need for it anymore.  The method is still used by unit tests so we can make it package private.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -27,28 +28,36 @@
...
-import io.openliberty.checkpoint.spi.SnapshotFailed.Type;
+import io.openliberty.checkpoint.internal.CheckpointFailed;
+import io.openliberty.checkpoint.internal.CheckpointFailed.Type;
 
 @Component(
            reference = @Reference(name = "hookFactories", service = CheckpointHookFactory.class, cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC,
                                   policyOption = ReferencePolicyOption.GREEDY),
-           property = { "osgi.command.function=snapshot", "osgi.command.scope=criu" })
-public class CheckpointImpl implements Checkpoint {
+           property = { "osgi.command.function=snapshot", "osgi.command.scope=criu", Constants.SERVICE_RANKING + ":Integer=-10000" })
```
tjwatson: We should get rid of the `osgi.command` service properties at this point.  There is no need to keep on maintaining this test gogo command for doing a checkpoint.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -59,60 +68,102 @@
...
+    public void checkpoint(@Parameter(names = "-p", absentValue = "") @Descriptor("The phase to checkpoint") Phase phase,
+                         @Descriptor("Directory to store the checkpoint") File directory) throws CheckpointFailed {
+        doCheckpoint(phase, directory);
     }
 
-    @Override
-    @Descriptor("Take a snapshot")
-    public void snapshot(Phase phase) throws SnapshotFailed {
-        doSnapshot(phase,
+    @Descriptor("Do a checkpoint")
```
tjwatson: Remove the gogo `@Descriptor` annotation and make this method package private.

dev/io.openliberty.checkpoint/test/io/openliberty/checkpoint/internal/CheckpointImplTest.java
```diff
@@ -31,8 +31,8 @@
 import io.openliberty.checkpoint.spi.CheckpointHookFactory.Phase;
 import io.openliberty.checkpoint.spi.CheckpointHook;
 import io.openliberty.checkpoint.spi.CheckpointHookFactory;
-import io.openliberty.checkpoint.spi.SnapshotFailed;
-import io.openliberty.checkpoint.spi.SnapshotFailed.Type;
+import io.openliberty.checkpoint.internal.CheckpointFailed;
```
tjwatson: unnecessary imports.

tjwatson:
fixes #17754

PR #17970, opened 2021-07-26
waitForStringInLog method with default timeout is used

### LINKED ISSUE #17969: Unable to find the output in the server log
Despite the fact that "Unable to find the output in the server log" exception is created, searched String appears in the log later.  As stated in LibertyServer comments, waitForStringInLog method with default timeout signature is used to prevent timeout issue.

BerksanAtes:
Fixes #17969


BerksanAtes:
#build

neuwerk (approved):
Changes look good! Thanks for doing this

PR #17975, opened 2021-07-27
New config attribute ibm-type="schedule"

KyleAure:
Introduce new metatype attribute `ibm:type="schedule"`
Parse config into objects that can be used by components to query if we are in the date/time range configured.

Multiple cardinality of this attribute can be done in one of two ways:
```
<!-- Single Attribute -->
<fakeElement1 scheduleAttribute="MON-FRI 8:00-17:30, SAT 9:00, SUN 9:00" />

<!-- Multiple Cardinality-->
<fakeElement2 scheduleAttribute="MON-FRI 8:00-17:30" scheduleAttribute="SAT 9:00" />
```


dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+        } else {
+            this.startup = false;
+        }
+
+        //Split MON-FRI 8:00-10:00 to [MON-FRI, 8:00-10:00] OR
+        //Split MON 8:00 to [MON, 8:00]
+        strVal = strVal.trim();
+        String[] splitDaysTimes = strVal.split(" ");
+        if (splitDaysTimes.length != 2) {
+            throw new IllegalArgumentException("Could not parse configuration value as a schedule: " + strVal);
```
njr-11: Translation is needed for error messages to the user, here and in similar places.
The error message that we do give should be more helpful.  Maybe add "... A schedule must include both day-of-week and hours, for example: MON-FRI 9:00-17:00."

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        //Split "MON-FRI" and "8:00" to [MON, FRI] [8:00]
+        String[] splitDays = splitDaysTimes[0].split("-");
+        String[] splitTimes = splitDaysTimes[1].split("-");
+
+        if (splitDays.length > 2 || splitDays.length < 1 || splitTimes.length > 2 || splitTimes.length < 1) {
+            throw new IllegalArgumentException("Could not parse days configured in schedule: " + strVal);
```
njr-11: More helpful message could add "... Day-of-week and time ranges must be delimited by a single hyphen (-). For example, MON-FRI 9:00-17:00."

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+
+        this.validDays = getValidDays(evaluateDayOfWeek(splitDays[0]), splitDays.length == 2 ? evaluateDayOfWeek(splitDays[1]) : null);
+        this.startTime = evaluateLocalTime(splitTimes[0]);
+        this.endTime = splitTimes.length == 2 ? evaluateLocalTime(splitTimes[1]) : null;
+    }
+
+    private static List<DayOfWeek> getValidDays(DayOfWeek startDay, DayOfWeek endDay) {
+        //Handle a single day
+        if (endDay == null) {
+            return new ArrayList<DayOfWeek>(Arrays.asList(startDay));
```
njr-11: Just return the Arrays.asList result instead of creating another list from it.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        return result;
+    }
+
+    private static DayOfWeek evaluateDayOfWeek(String strVal) {
+        try {
+            //Will return successfully for monday, MONDAY, Monday
+            return DayOfWeek.valueOf(strVal.toUpperCase());
+        } catch (IllegalArgumentException e) {
```
njr-11: FFDCIgnore is needed for exceptions that are expected code path. Although that said, we should not cause an exception on expected code path given that we are recommending the short form (MON, TUE).  Instead, do the switch first, and move the valueOf into the default:, and then the exception will no longer be expected and FFDC will be fine.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+     * Try to parse string representation of time to meet format requirements in
+     * {@link java.time.format.DateTimeFormatter#ISO_LOCAL_TIME}
+     */
+    private static LocalTime evaluateLocalTime(String strVal) {
+        try {
+            //Optimistic string was configured correctly
+            return LocalTime.parse(strVal);
+        } catch (DateTimeParseException e) {
+            //Need to call LocalTime.parse with leading 0
+            //Example: 8:00 will throw exception, need to provide "08:00"
```
njr-11: I could see plenty of users doing the following without thinking about it:
MON 8-5:30
We should consider doing something like the following (before attempting parse) to correct it for them automatically:
If there is no second character, then prepend "0" and append ":00".
Else if the second character is ':' then preprend "0".
Else if the next character isn't ':' then append ":00".

If we don't fix it for them, then make sure the error message adequately shows what is wrong and includes a valid example.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+
+        if (o == this) {
+            return true;
+        }
+
+        SerializableSchedule c = (SerializableSchedule) o;
+
+        //Compare fields taking into account possible null values
+        boolean startupBool = Boolean.compare(this.startup, c.startup) == 0;
+        boolean validDays = this.validDays == null ? c.validDays == null : this.validDays.equals(c.validDays);
```
njr-11: List compare will take into account order.  Should  FRI-THU be equal to SAT-FRI or not ?

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+     * @return true we are within the schedule, false otherwise.
+     */
+    public boolean isNowWithinTheSchedule(long marginNS) {
+
+        // If day is not valid don't check time
+        if (validDays.contains(LocalDate.now().getDayOfWeek()))
+            return false;
+
+        LocalTime startAdjusted = startTime.minusNanos(marginNS / 2);
+        LocalTime endAdjusted = endTime.plusNanos(marginNS / 2);
```
njr-11: The margin is confusing.  From reading the JavaDoc, I didn't expect it to be split in half.  But I also don't see it being useful in our particular case because it would cause us to start performing work beyond the end of the interval.  Maybe we should defer this sort of logic to the consumer of the schedule?  Or maybe rewrite this method to be 
boolean withinSchedule(LocalDateTime time) where the caller has control over the time they want to consider and can adjust it or check a future time as they see fit?

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.wsspi.kernel.service.utils;
+
+import java.io.Serializable;
+import java.time.DayOfWeek;
+import java.time.LocalDate;
+import java.time.LocalTime;
```
njr-11, line 17: I'm sorry that I didn't think of this during the design review, but I'm concerned about the use of local time because it requires the user to have knowledge of which time zone(s) their servers are running in and forces them to think in that time zone versus the time zone that their business is running in, which will be more natural for them. I wonder if we should add a (optional?) time zone/zone id to the configuration, `MON-FRI 9:00-17:00 America/Chicago` and operate in ZonedDateTime.  Time zones like CDT or CST are more familiar to most people, but would lead to confusion when crossing over the daylight saving time boundary outside of what they pertain to.

njr-11 (dismissed):
Kyle - excellent work getting these changes into the config code! I'll defer to Brent on the review of that part and concentrate on other aspects.  None of my comments should hold up delivery of this, and can be addressed in follow-on pulls.  Some are just things to think about.

brenthdaniel (dismissed):

KyleAure:
#build

KyleAure:
#build

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -71,123 +84,191 @@ public SerializableSchedule() {
...
-        //Split MON 8:00 to [MON, 8:00]
-        strVal = strVal.trim();
-        String[] splitDaysTimes = strVal.split(" ");
-        if (splitDaysTimes.length != 2) {
-            throw new IllegalArgumentException("Could not parse configuration value as a schedule: " + strVal);
+        //Split "MON-FRI 8:00-10:00 America/Chicago"  to [MON-FRI, 8:00-10:00, America/Chicago]
+        String[] splitDaysTimesZone = strVal.split(" ");
+
+        if (splitDaysTimesZone.length < 2 || splitDaysTimesZone.length > 3) {
+            throw new IllegalArgumentException(Tr.formatMessage(tc, "invalidScheduleDelimit", originalStr, "MON-FRI 8-17:00 GMT-6"));
```
njr-11: Let's use `America/Chicago` in the example because `GMT-6` suggests that the schedule won't account for Daylight Saving Time, when Chicago (and the rest of Central Time) are on `GMT-5`.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -71,123 +84,191 @@ public SerializableSchedule() {
...
+        }
+
+        //Parse DayOfWeek
+        //Split "MON-FRI" to [MON, FRI] or "MON" to [MON]
+        String[] splitDays = splitDaysTimesZone[0].split("-");
+
+        // Ensure only one or two values
+        if (splitDays.length < 1 || splitDays.length > 2)
+            throw new IllegalArgumentException(Tr.formatMessage(tc, "invalidScheduleDelimitRange", splitDaysTimesZone[0], "MON-FRI"));
```
njr-11: If someone enters the wrong order and mistakenly uses - instead of _,
`Asia/Ho-Chi-Minh 9:00-17:00 MON-FRI`
then this error message will tell them that they need to use a single hyphen for day ranges, such as MON-FRI and they will suspect that Liberty is at fault instead of their input because they have MON-FRI already.  I think we should at least include the full example with all 3 parts to help them get it right.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -71,123 +84,191 @@ public SerializableSchedule() {
...
-        //Split "MON-FRI" and "8:00" to [MON, FRI] [8:00]
-        String[] splitDays = splitDaysTimes[0].split("-");
-        String[] splitTimes = splitDaysTimes[1].split("-");
+        //Parse LocalTime
+        //Split "8:00-17:00" to [8:00, 17:00] or "8:00" to [8:00]
+        String[] splitTimes = splitDaysTimesZone[1].split("-");
+
+        //Ensure only one or two values
+        if (splitTimes.length < 1 || splitTimes.length > 2)
+            throw new IllegalArgumentException(Tr.formatMessage(tc, "invalidScheduleDelimitRange", splitDaysTimesZone[1], "8:00-17:00"));
```
njr-11: Similar to previous comment.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -71,123 +84,191 @@ public SerializableSchedule() {
...
         }
 
-        //
         ArrayList<DayOfWeek> result = new ArrayList<>();
-        for (int d = startDay.getValue(); d <= endDay.getValue(); d = d == 7 ? 1 : d + 1) {
-            result.add(DayOfWeek.of(d));
+
+        //Increment through circular array [0-6]
+        for (int day = startDayPosition; day != endDayPosition; day = (day + 1) % 7) {
+            result.add(DayOfWeek.of(day + 1));
```
njr-11: I think your algorithm produces the correct result, but I found the variable names confusing when trying to follow it.  `startDayPosition` is actually the day before the start day.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -215,44 +296,86 @@ public boolean equals(Object o) {
 
         //Compare fields taking into account possible null values
         boolean startupBool = Boolean.compare(this.startup, c.startup) == 0;
-        boolean validDays = this.validDays == null ? c.validDays == null : this.validDays.equals(c.validDays);
+        boolean validDaysBool = this.validDays == null ? c.validDays == null : (this.validDays.containsAll(c.validDays) && c.validDays.containsAll(this.validDays));
```
njr-11: Should the data type be a Set rather than a List? Then Set.equals could be used for comparison, and validDays.contains would also be more efficient.
I suppose the disadvantage to this would be if the value gets printed, it will look fairly random rather than orderly.  TreeSet (which is a SortedSet) could help with that.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -215,44 +296,86 @@ public boolean equals(Object o) {
...
+     * Method to determine if a LocalDateTime is within the schedule represented by this object.
+     * Note: This method assumes that the LocalDateTime provided represents a ZonedDateTime within the timezone of this schedule.
      *
-     * Example1: if the schedule is "MON 8:00" and you are checking the schedule every 6 minutes then calling
-     * this method between 7:57 and 8:03 will return true on Monday.
+     * @param time - LocalDateTime
+     * @return boolean - true the time provided is within the scheduled represented by this object, false otherwise.
+     */
+    public boolean withinSchedule(LocalDateTime time) {
+        return withinSchedule(time.atZone(timezone));
```
njr-11: I wouldn't expect the consumer of the schedule to want to supply a time that they have converted to be in the configured zone.  It seems more likely they will have something like Instant.now() or ZonedDateTime.now(), which you already have a method for.

This is also error prone.  I could see someone supplying LocalDateTime.now() which is computed in the server's time zone then being interpreted by this method as though the value were the configured time zone.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -215,44 +296,86 @@ public boolean equals(Object o) {
...
+        final boolean result;
+
+        // If day is not valid don't check time
+        if (!validDays.contains(time.getDayOfWeek())) {
+            result = false;
+        } else {
+            LocalTime check = time.toLocalTime();
+            result = startTime.isBefore(endTime) ? // Dont need to worry about wrapped time
+                            (startTime.isBefore(check) && check.isBefore(endTime)) : // [ xxx start --- check --- end xxx ]
+                            (check.isBefore(endTime) || startTime.isBefore(check)); //  [ --- check --- end xxx start --- check --- ]
```
njr-11: It seems like performing the comparisons in LocalTime would not properly consider DaylightSavingTime of the zone.  I would have expected the LocalTime + configured zone to be made into a ZonedDateTime, which could then be compared to the time argument (ZonedDateTime) that is supplied this method.

KyleAure:
#libby-build

KyleAure:
#libby-build

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
 invalidDuration.explanation=The value for the specified configuration parameter is not, or cannot be converted to, a duration value.
 invalidDuration.useraction=Ensure that the value specified in the server configuration is a valid representation of a duration value.
 
+invalidScheduleDelimit=CWWKE0108E: Schedule components must be delimited by a single space ( ). Provided: {0}. Valid example: {1}
+invalidScheduleDelimit.explanation=A schedule must include both a day-of-week and a time component. Optionally, a schedule can also have a timezone component.   
+invalidScheduleDelimit.useraction=Ensure that the schedule specified in the server configuration uses a single space ( ) to delimit day-of-week, time, and timezone components.
```
dmuelle:
```suggestion
invalidScheduleDelimit.useraction=Ensure that the schedule that is specified in the server configuration uses a single space ( ) to delimit day-of-week, time, and timezone components.
```

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
...
+
+invalidScheduleDelimitRange=CWWKE0108E: Schedule ranges must be delimited by a single hyphen (-). Provided: {0}. Valid example: {1}
+invalidScheduleDelimitRange.explanation=A range of days or times must have a start and end component delimited by a single hyphen (-).
```
dmuelle:
```suggestion
invalidScheduleDelimitRange.explanation=A range of days or times must have a start component and an end component that are delimited by a single hyphen (-).
```

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
...
+invalidScheduleDelimitRange.useraction=Ensure that the schedule specified in the server configuration uses a single hyphen (-) to delimit day and time ranges.
```
dmuelle:
```suggestion
invalidScheduleDelimitRange.useraction=Ensure that the schedule that is specified in the server configuration uses a single hyphen (-) to delimit day and time ranges.
```

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
...
+
+invalidScheduleParse=CWWKE0109E: Unable to parse the provided parameter to a valid {0}. Provided: {1}. Valid example(s): {2}.
```
dmuelle: what are the possible values (or kinds of values) for {0}?

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
...
+invalidScheduleParse.explanation=A schedule must be able to evaluate days as java.time.DayOfWeek, times as java.time.LocalTime, and timezones as java.time.ZoneID.
+invalidScheduleParse.useraction=Ensure that all of the schedule components specified in the server configuration use the correct syntax to be evaluated as a corresponding java.time object.
```
dmuelle:
```suggestion
invalidScheduleParse.useraction=Ensure that all of the schedule components that are specified in the server configuration use the correct syntax to be evaluated as a corresponding java.time object.
```

dmuelle (changes requested):
ID review- a few suggestions and a question

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
KyleAure: They are Java SE class names such as `java.time.DayOfWeek` `java.time.LocalTime` and `java.time.ZoneId`

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
...
 
+invalidScheduleDelimit=CWWKE0108E: Schedule components must be delimited by a single space ( ). Provided: {0}. Valid example: {1}
+invalidScheduleDelimit.explanation=A schedule must include both a day-of-week and a time component. Optionally, a schedule can also have a timezone component.   
+invalidScheduleDelimit.useraction=Ensure that the schedule that is specified in the server configuration uses a single space ( ) to delimit day-of-week, time, and timezone components.
+
+invalidScheduleDelimitRange=CWWKE0108E: Schedule ranges must be delimited by a single hyphen (-). Provided: {0}. Valid example: {1}
+invalidScheduleDelimitRange.explanation=A range of days or times must have a start component and an end component that are delimited by a single hyphen (-).
+invalidScheduleDelimitRange.useraction=Ensure that the schedule that is specified in the server configuration uses a single hyphen (-) to delimit day and time ranges.
+
+invalidScheduleParse=CWWKE0109E: Unable to parse the provided parameter to a valid {0}. Provided: {1}. Valid example(s): {2}.
```
dmuelle:
```suggestion
invalidScheduleParse=CWWKE0109E: Unable to parse the provided parameter to a valid {0} class. Provided: {1}. Valid example(s): {2}.
```

dmuelle: thanks, updated mysuggestion

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
 invalidDuration.explanation=The value for the specified configuration parameter is not, or cannot be converted to, a duration value.
 invalidDuration.useraction=Ensure that the value specified in the server configuration is a valid representation of a duration value.
 
+invalidScheduleDelimit=CWWKE0108E: Schedule components must be delimited by a single space ( ). Provided: {0}. Valid example: {1}
```
njr-11, line 59: Missing period at end of sentence like CWWKE0109E has.

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
...
+invalidScheduleDelimit.explanation=A schedule must include both a day-of-week and a time component. Optionally, a schedule can also have a timezone component.   
+invalidScheduleDelimit.useraction=Ensure that the schedule that is specified in the server configuration uses a single space ( ) to delimit day-of-week, time, and timezone components.
+
+invalidScheduleDelimitRange=CWWKE0108E: Schedule ranges must be delimited by a single hyphen (-). Provided: {0}. Valid example: {1}
```
njr-11, line 63: Missing period at end of sentence like CWWKE0109E has.

dmuelle (approved):
LGTM, thanks

dev/com.ibm.ws.kernel.service/resources/com/ibm/ws/kernel/service/utils/resources/ServiceMessages.nlsprops
```diff
@@ -56,6 +56,18 @@ invalidDuration=CWWKE0107W: The specified parameter does not represent a duratio
...
+invalidScheduleDelimitRange.explanation=A range of days or times must have a start component and an end component that are delimited by a single hyphen (-).
+invalidScheduleDelimitRange.useraction=Ensure that the schedule that is specified in the server configuration uses a single hyphen (-) to delimit day and time ranges.
+
+invalidScheduleParse=CWWKE0109E: Unable to parse the provided parameter to a valid {0} class. Provided: {1}. Valid example(s): {2}.
```
njr-11, line 67: "parse the provided parameter to a valid {0} class" is ambiguous.
Does it mean that we cannot supply {1} as a parameter to a method or constructor that we want to invoke on class {0} because we cannot parse {1}?
Does it mean that we cannot convert {1} into an instance of class {0}?
Will the user understand what "parameter" is?

After locating where this is used in code, here is one possible way to update it:
"Term {1} of the provided schedule cannot be parsed into a {0} class. Valid examples(s): {2}."

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,434 @@
...
+        } catch (DateTimeParseException e) {
+            throw new IllegalArgumentException(Tr.formatMessage(tc, "invalidScheduleParse", LocalTime.class, splitDaysTimesZone[1], "[8:00, 17:00]"), e);
+        }
+
+        //Parse ZoneID
+        try {
+            this.timezone = splitDaysTimesZone.length == 3 ? ZoneId.of(splitDaysTimesZone[2]) : ZoneId.systemDefault();
+        } catch (DateTimeException e) {
+            throw new IllegalArgumentException(Tr.formatMessage(tc, "invalidScheduleParse", ZoneId.class, splitDaysTimesZone[2],
+                                                                ZoneId.systemDefault().getDisplayName(TextStyle.SHORT, Locale.getDefault())), e);
```
njr-11, line 160: The JavaDoc for `getDisplayName` states,
```
Gets the textual representation of the zone, such as 'British Time' or '+02:00'.
This returns the textual name used to identify the time-zone ID, suitable for presentation to the user. The parameters control the style of the returned text and the locale.
If no textual mapping is found then the [full ID](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#getId--) is returned.
```
Is there documentation anywhere that guarantees that the display name generated here will be parseable back into a ZoneId?  It would be safest to just hard code one that is known to be valid.

dev/com.ibm.ws.kernel.service/src/com/ibm/wsspi/kernel/service/utils/SerializableSchedule.java
```diff
@@ -0,0 +1,434 @@
...
+        switch (strVal.toUpperCase()) {
+            case "MON":
+                return DayOfWeek.MONDAY;
+            case "TUE":
+            case "TUES":
+                return DayOfWeek.TUESDAY;
+            case "WED":
+                return DayOfWeek.WEDNESDAY;
+            case "THU":
+            case "THURS":
```
njr-11, line 232: "THUR" is also used to abbreviate Thursday

dev/com.ibm.ws.kernel.service/test/com/ibm/wsspi/kernel/service/utils/MetaTypeUtilsTest.java
```diff
@@ -384,4 +401,289 @@ public void testLocale() throws Exception {
...
+
+        //Test Days of week
+        testSchedule = MetatypeUtils.evaluateSchedule("MON-SUN 8")[0];
+        assertEquals(alldays, testSchedule.getValidDays());
+
+        testSchedule = MetatypeUtils.evaluateSchedule("WED-TUES 8")[0];
+        assertEquals(alldays, testSchedule.getValidDays());
+
+        testSchedule = MetatypeUtils.evaluateSchedule("WED-WED 8")[0];
+        assertEquals(wednesday, testSchedule.getValidDays());
```
njr-11, line 445: This one is confusing and I wonder if we should reject it.  It isn't clear to me if this should mean all days in the week or only Wednesday.

dev/com.ibm.ws.kernel.service/test/com/ibm/wsspi/kernel/service/utils/MetaTypeUtilsTest.java
```diff
@@ -384,4 +401,289 @@ public void testLocale() throws Exception {
...
+    }
+
+    @Test
+    public void testScheduleExceptions() throws Exception {
+
+        //Constructor exceptions
+        try {
+            //At minimum need startDay and startTime
+            new SerializableSchedule(null, null, null, null, null);
+        } catch (Exception e) {
```
njr-11, line 510: A general comment for the tests in this method is that they are missing enforcement that an exception is raised.

njr-11 (approved):

posmith:
L2 message review completed

PR #18027, opened 2021-08-02
Kerberos test issues on linux using kinit

KyleAure:
Some of our SOE test systems have installed a version of `kinit` that does not support the lifetime flag resulting in an error message during the com.ibm.ws.jdbc_fat_krb5 suite `Bad principal name: -l`.

This only seems to be happening on Linux Platforms with IBM JDK installed.  It could be the case that when these systems are installed with the IBM JDK an alternative version of kinit is installed.  The IBM JDK does come with a `kinit` class that also does not support lifetime configurations.  https://www.ibm.com/docs/en/sdk-java-technology/7.1?topic=jgss-kinit 

Since this is only affecting two tests that only run in full mode I have it so that these tests are skipped if we determine that the OS does not support the lifetime flag.  If in the future these systems do support the flag then these tests will run.


KyleAure:
#build

KyleAure:
#build

mswatosh (approved):

KyleAure:
#libbybot

PR #18028, opened 2021-08-02
create TCK bucket for mp context propagation 1.3

njr-11:
This pull gets as far as we can without having the nonship copies of the mpConfig-3.0 and mpContextPropagation-1.3 features yet.  That means to start out with, the new bucket just runs the 1.2 TCK, but is ready to switch over as soon as we have these features.


njr-11:
#build

dev/fattest.simplicity/build.gradle
```diff
@@ -63,10 +65,30 @@ task extractArquillianSupportFeature(type:Sync) {
...
+}
+
 task publishArquillianSupportFeature(type:Copy) {
     dependsOn extractArquillianSupportFeature
     from('build/arquillian-feature-extract/extension')
     into(buildImage.file('wlp'))
+
+    dependsOn extractArquillianSupportJakartaFeature
+    from('build/arquillian-jakarta-feature-extract/extension')
+    into(buildImage.file('wlp'))
```
Azquelt, line 90: I don't think you need two `into`s here. I think the one above will apply to the whole task.

Azquelt (approved):
Approval for the gradle change to install the jakarta support feature.

dev/fattest.simplicity/build.gradle
njr-11: Thanks for spotting that. I removed the extra.

njr-11:
#run-libby-bot
#build

PR #18050, opened 2021-08-04
run TCK against jakarta packages with local snapshot

njr-11:
Create non-ship MP Context Propagation 1.3 feature and update the TCK bucket to use a local snapshot of the MP Context Propagation 1.3 TCK.  After doing this, tests are passing at 86%, where some failures are expected due to the lack of a MicroProfile 5.0 MP Config feature that some of the TCK tests require.


njr-11:
#build

mswatosh (approved):

njr-11:
Feature message is copied from 1.2 feature, there is no need for an ID review

skasund:

PR #18058, opened 2021-08-04
Add Overlay to Eclipselink 3.0.2 to address Eclipselink#1225

jgrassel:
Eclipselink 3.0.2 shipped without updating all ASM8 references to ASM9 (see https://github.com/eclipse-ee4j/eclipselink/issues/1225).  To support Java 17, we need to update two classes with an overlay which properly references ASM9.

This change is equal to the corrective PR on the Eclipselink repo, https://github.com/eclipse-ee4j/eclipselink/pull/1226


jgrassel:
#build

PR #18060, opened 2021-08-05
Run completion listeners on disconnect instead of closing event source

andymc12:
Many SSE TCK test cases are failing because the `SseEventSource#isOpen` method in the client is returning false after the server has disconnected. The javadoc for `isOpen` says that it should return true until the user has called the `close` method. This change ensures that when the server ends the connection the client will invoke event and completion listeners, but will not close the event source until the user has called close.


andymc12:
#build

jim-krueger (approved):

andymc12:
#build

andymc12:
made an additional change - new #build

andymc12:
5 total failures, but all are unrelated to these changes. Further, the build analyzer tool shows that they occur in several other builds that do not have these changes.

#libby

PR #18064, opened 2021-08-05
autofeature to enable CDI context propagation for v1.3

njr-11:
Create autofeature to enable integration between MP Context Propagation 1.3 and CDI 3.0.
With this, some additional TCK tests are now passing.  The only TCK tests remaining that need to be excluded are those that rely on MicroProfile Config, which needs to be disabled for now because it doesn't support jakarta packages yet.


jmstephensgit (approved):

njr-11:
#build

PR #18067, opened 2021-08-05
Update JBoss ClassFileWriter bundles

tevans78:
We only need one version of ClassFileWriter 1.x
#build


Emily-Jiang (approved):
LGTM as long as the FAT tests can run successfully.

PR #18074, opened 2021-08-06
Revert change to IBM Java detection in PluginGenerator

Azquelt:
There's something wrong with this change which breaks when running on IBM Java 8.

I would guess that the xalan transformer is available and should be used but the implementation class isn't loadable by application code?


jhanders34 (approved):

Azquelt:
Merged without build as it's a revert and the current code is broken.

PR #18092, opened 2021-08-06
Remove docker constraint for libhpike on ACME FATs

kristip17:
Try removing the host constraint for libhpike docker servers for acme_fat. Previously the boulder image would time out while starting on this bank of servers. Locally, it worked for me. Remove the constraint and see what occurs -- `AcmeRevocationTest` which runs on FULL mode will fail if we can't start the boulder image.


kristip17:
#build

mhldr (approved):

PR #18103, opened 2021-08-09
Remove service filters for testcontainer buckets

KyleAure:
Remove service filtering for JDBC buckets to ensure that they do not accidentally filter out the only docker-engine machines they can run against.


KyleAure:
#build

mhldr (approved):

PR #18106, opened 2021-08-09
test bucket for MP Context Propagation 1.3

njr-11:
Create a test bucket where we can place our own tests of MicroProfile Context Propagation 1.3 with Open Liberty.  To start with, I added 4 tests, including the provision of a custom context type via a bell/shared library, and usage of the default ManagedExecutorService for Jakarta EE as a MicroProfile ManagedExecutor, along with a couple of scenarios demonstrating that function from 1.2 is still working.  One of these actually uncovered a bug with ManagedExecutor.copy, which I'm also fixing in this pull.


mswatosh (approved):

njr-11:
#build

njr-11:
Build has 3 test failures, which are preexisting issues in unrelated test buckets.

PR #18116, opened 2021-08-09
Disable access token cache for re-auth jaxrs tests

c00crane:
We've recently added support for an inbound access token cache.  For the tests that are trying to test when we're required to re-authenticate, the new cache can cause problems.
We have tests that will validate the caching behavior - these tests need to focus on token lifetimes, so, we will disable the new cache.


c00crane:
#build

teddyjtorres (approved):

PR #18126, opened 2021-08-10
18068 jax release cc after service invoke

### LINKED ISSUE #18068: JaxRsFactoryImplicitBeanCDICustomizer does not release CreationalContexts
In JaxRsFactoryImplicitBeanCDICustomizer.getClassFromCDI() creates a CreationalContext but JAX-RS does not call the release() method.

I have noticed that this means that in the following scenario: 

- You have a bean annotated `@Dependent` with a `preDestroy()` method

- That is injected into a class annotated `@Path` 

The preDestroy method will not be called. It is possible that other problems may result too.

benjamin-confino:
fixes #18068 for EE7 and EE8 but not EE9

### DIFF
--- a/dev/com.ibm.ws.jaxrs.2.0.cdi.1.2_fat/bnd.bnd
+++ b/dev/com.ibm.ws.jaxrs.2.0.cdi.1.2_fat/bnd.bnd
@@ -19,6 +19,7 @@ src: \
     test-applications/complex/src,\
     test-applications/contextandclient/src,\
     test-applications/contextandCDI/src,\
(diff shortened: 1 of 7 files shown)


andymc12 (changes requested):
@benjamin-confino Thanks so much for this fix.  The new test files are missing copyright headers and some of the other files need the copyright date changed to 2021, but otherwise this looks good.  Thanks again!

benjamin-confino:
#build

andymc12 (approved):

benjamin-confino:
#run-libby-bot

PR #18141, opened 2021-08-11
Fix various missing tested.features

tkburroughs:
Fix a few soft merge conflicts with the simplicity feature dependency
checker fixes, and also account for running on open-liberty only images
where servlet-3.0 does not exist, so servlet-3.1 is used instead.


tkburroughs:
#build

olendvcook (approved):

PR #18150, opened 2021-08-11
Remove testcase due to spec clarification.

jim-krueger:
This PR is removing a testcase variation due to a specification clarification for Server Sent Events indicating that reconnect delay is only applicable when a connection to the SSE endpoint is lost. This test variation tested that a reconnect would take whenever reconnect delay was set, even when normal event processing took place with no errors or loss of connection. This testcase variation has therefore been removed.


jim-krueger:
This PR is removing a testcase variation due to a specification clarification for Server Sent Events indicating that reconnect delay is only applicable when a connection to the SSE endpoint is lost.  This test variation tested that a reconnect would take whenever reconnect delay was set, even when normal event processing took place with no errors or loss of connection.   This testcase variation has therefore been removed.

jim-krueger:
#build

WhiteCat22 (approved):

PR #18171, opened 2021-08-12
[18147] Initial impl of `restfulWSLogging-3.0` feature to log HTTP data

### LINKED ISSUE #18147: Serviceability / JAX-RS 3.0: Logging filter for HTTP request/response data
Part of epic issue #11803:

As part of JAX-RS 3.0, we'd like to add a serviceability feature (in the form of a liberty feature, like `restfulWSLogging-1.0`) that would enable a container filter that would log HTTP data like:

```
>POST /app1/resouce1?id=123

Authorization: Basic blahblahblah
Content-Type: application/json
...
Accept: application/json

{
    someObject: { ... }
}

<200 OK

Cookie: someCookieData
Content-Type: application/json

{
    accepted: true
}
```

We'll probably want some switches to determine whether to log all requests/response or just specific requests. We'll probably also want some switches to determine whether to print the entire contents of the HTTP body or not (it's expensive).

andymc12:
Fixes #18147 - adds new `restfulWSLogging-3.0` feature that includes `restfulWS-3.0` and a new filter that logs requests/responses.


andymc12:
#build

dev/com.ibm.websphere.appserver.features/visibility/public/restfulWSLogging-3.0/resources/l10n/io.openliberty.restfulWSLogging-3.0.properties
```diff
@@ -0,0 +1,18 @@
...
+#     IBM Corporation - initial API and implementation
+###############################################################################
+#
+#ISMESSAGEFILE FALSE
+#NLS_ENCODING=UNICODE
+#NLS_MESSAGEFORMAT_NONE
+#
+
+description=This feature enables HTTP request and response logging support for Jakarta RESTful Web Services 3.0. \
+ When enabled, HTTP requests that are matched to Jakarta RESTful resource methods will be printed to log output.
```
helyarp: Change "will be" to "are"

helyarp (commented):
One minor suggested change to text

dev/com.ibm.websphere.appserver.features/visibility/public/restfulWSLogging-3.0/resources/l10n/io.openliberty.restfulWSLogging-3.0.properties
andymc12: will do - thanks!

andymc12:
Addressed message review comment and removed extraneous system-out. New #build

helyarp (commented):
Looks good!

dev/io.openliberty.restfulWS.internal.logging.filter/bnd.bnd
```diff
@@ -0,0 +1,46 @@
...
+#*******************************************************************************
+-include= ~../cnf/resources/bnd/bundle.props
+bVersion=1.0
+
+Bundle-SymbolicName: io.openliberty.restfulWS.internal.logging.filter
+
+app-resources= \
+  META-INF/services/javax.ws.rs.ext.Providers
+
+Service-Component: \
```
WhiteCat22, line 19: This is more of an education question, but can you explain what this service component does?

dev/io.openliberty.restfulWS.internal.logging.filter/src/io/openliberty/restfulWS/internal/logging/filter/LoggingFilter.java
```diff
@@ -0,0 +1,157 @@
...
+
+        // it's expensive to read the HTTP entity (requires copying the stream - and the full stream onto the heap)
+        // so best to avoid unless user really wants to see it:
+        final String entity;
+        if (!LOG_ENTITY) {
+            entity = UNLOGGED;
+        } else if (!reqCtx.hasEntity()) {
+            entity = NO_ENTITY;
+        } else {
+            byte[] entityBytes = toByteArray(reqCtx.getEntityStream());
```
WhiteCat22, line 99: Uh, not 100% sure, but I'm worried about this causing the entity stream to be read when the backend tries to read it. Have you tested this?

WhiteCat22 (approved):

dev/io.openliberty.restfulWS.internal.logging.filter/bnd.bnd
andymc12: This is basically the same thing as what `@Component` does in the code - it sets up an OSGi Declarative Service.  But this is a unique one in that it is actually setting up a DS component for another bundle (the classloading service bundle) - so it cannot be created using annotations.  This is what makes it possible for user applications to load resource files (like META-INF/services/... files) from Liberty bundles.

dev/io.openliberty.restfulWS.internal.logging.filter/src/io/openliberty/restfulWS/internal/logging/filter/LoggingFilter.java
andymc12: yes, this will read the entity stream - which is why I have it set to not read/print it by default.  I tested it to verify that the resource will still be able to read the stream (from the newly created ByteArrayInputStream in the code below). It's not great for performance, but I think users would only enable this if they were willing to give up some performance for additional logging.

hcs-ibm:
L2 Message Review Complete

andymc12:
#libby

PR #18172, opened 2021-08-12
Modifying RESTEasy behavior to always return 200 when no events are sent

andymc12:
This change was suggested by the JAX-RS project lead in response to TCK challenge issue https://github.com/eclipse-ee4j/jaxrs-api/issues/1007. 

The test being challenged is related to reconnect behavior, but it assumes that the server sends a 200 HTTP response code when not sending any events. RESTEasy has (perhaps correctly?) decided to return a 204 code instead since there is no content being returned to the client - so this disrupts the TCK test.

This change may end up being temporary depending on the result of the challenge, and I expect that the spec and/or TCK will be clarified in a future release.


andymc12:
#build

jim-krueger (approved):

andymc12:
New changes affecting client behavior - ensuring the client reconnects after receiving a 200 with no events. New #build

jim-krueger (approved):

andymc12:
Failures are unrelated - #libby

PR #18184, opened 2021-08-13
Plugin generator java8

tevans78:
On some versions of IBM Java SDKs, the default Transformer Factory is com.ibm.xtq.xslt.jaxp.compiler.TransformerFactoryImpl. That is a compiler processor but we don't want to use that. We want to use the interpretive processor.

https://www.ibm.com/docs/en/sdk-java-technology/8?topic=guide-using-xml#xml_using__select_processor

> com.ibm.xtq.xslt.jaxp.compiler.TransformerFactoryImpl
>     Use the XL TXE-J compiler. This value is the default.
> org.apache.xalan.processor.TransformerFactoryImpl
>     Use the XSLT4J interpreter. 

Previously, the code would have looked at the JVM vendor and version to estabilsh if running on IBM Java 8. However, that approach can not be guaranteed to work in the future. So instead we look directly at which classes are available from the system classloader. If the IBM XSLT classes are not all available then we use the oracle/sun fallback processor, the same one which is defined as a fallback in the API.

This PR also contains a few minor cleanup items such as removing unused imports and code.

### DIFF
--- a/dev/com.ibm.ws.webcontainer/src/com/ibm/ws/webcontainer/osgi/mbeans/PluginGenerator.java
+++ b/dev/com.ibm.ws.webcontainer/src/com/ibm/ws/webcontainer/osgi/mbeans/PluginGenerator.java
@@ -23,7 +23,6 @@ import java.net.UnknownHostException;
 import java.nio.channels.FileChannel;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
-import java.nio.file.StandardCopyOption;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.text.SimpleDateFormat;
@@ -44,6 +43,9 @@ import java.util.Set;
 
 import javax.servlet.SessionCookieConfig;
 import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.parsers.SAXParser;
+import javax.xml.parsers.SAXParserFactory;
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerFactory;
@@ -51,13 +53,17 @@ import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
 
-import java.io.IOException;
-import java.io.StringReader;
-
-import javax.xml.parsers.ParserConfigurationException;
-import javax.xml.parsers.SAXParser;
-import javax.xml.parsers.SAXParserFactory;
-
+import org.apache.commons.io.FileUtils;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceReference;
+import org.w3c.dom.Attr;
+import org.w3c.dom.Comment;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.NodeList;
 import org.xml.sax.Attributes;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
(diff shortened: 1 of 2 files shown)


tevans78:
#build

tevans78:
#build

tevans78:
#build

tevans78:
#build

dev/com.ibm.ws.webcontainer/src/com/ibm/ws/webcontainer/osgi/mbeans/PluginGenerator.java
```diff
@@ -905,65 +887,29 @@ public static void copyFile(File in, File out)
...
+                                    JavaInfo.isAvailable(XALAN_TRANSFORMER_FACTORY_CLASS_NAME) &&
+                                    JavaInfo.isAvailable(SAX_LEXICAL_HANDLER_CLASS_NAME);
+        
+        String transformerFactory = changeTransformer ? XALAN_TRANSFORMER_FACTORY_CLASS_NAME : FALLBACK_XALAN_TRANSFORMER_FACTORY_CLASS_NAME;
+        
+        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+            Tr.debug(tc, "JDK : Use transformer factory: " + transformerFactory);
+        }
+        
+        tf = TransformerFactory.newInstance(transformerFactory, ClassLoader.getSystemClassLoader());
```
Azquelt: I think this has the potential to break in future if the FALLBACK_XALAN processor is not present in any JVM which doesn't have all of XALAN, IBM_XLTXEJ and the SAX lexical handler.

I think it would be better to retain the old behaviour and call `TransformerFactory.newInstance()` to obtain the default implementation if `changeTransformer` is false.

dev/com.ibm.ws.webcontainer/src/com/ibm/ws/webcontainer/osgi/mbeans/PluginGenerator.java
```diff
@@ -905,65 +887,29 @@ public static void copyFile(File in, File out)
...
-            }
-        } else {
-            tf = TransformerFactory.newInstance();
+        //We want to use the the Interpretive processor (Apache Xalan) but need to work out which one
+        //This is either going to be the one package with an IBM Java SDK or the one packaged as a fallback by Oracle
+        boolean changeTransformer = JavaInfo.isAvailable(IBM_XLTXEJ_COMPILED_TRANSFORMER_FACTORY_CLASS_NAME) &&
+                                    JavaInfo.isAvailable(XALAN_TRANSFORMER_FACTORY_CLASS_NAME) &&
+                                    JavaInfo.isAvailable(SAX_LEXICAL_HANDLER_CLASS_NAME);
+        
+        String transformerFactory = changeTransformer ? XALAN_TRANSFORMER_FACTORY_CLASS_NAME : FALLBACK_XALAN_TRANSFORMER_FACTORY_CLASS_NAME;
```
Azquelt: I think we could have a better name than `changeTransformer`, possibly `useApacheXalanTransformer` or `useIBMJDK8Workaround`?

Azquelt (changes requested):

tevans78:
#build

dev/com.ibm.ws.webcontainer/src/com/ibm/ws/webcontainer/osgi/mbeans/PluginGenerator.java
```diff
@@ -885,85 +867,48 @@ protected synchronized void generateXML(String rootLoc, String serverName,
...
-            // 2. Thread 1 sets jvm property
-            // 3. Thread 2 gets jvm property set by Thread 1
-            // 4. Thread 1 resets jvm property to value obtained at 1.
-            // 5. Thread 2 resets jvm property to value set by Thread 1.
-            synchronized (transformerLock) {
+        //We want to use the the Interpretive processor (Apache Xalan) but need to work out which one
+        //This is either going to be the one package with an IBM Java SDK or the one packaged as a fallback by Oracle
+        boolean useApacheXalanTransformer = JavaInfo.isAvailable(IBM_XLTXEJ_COMPILED_TRANSFORMER_FACTORY_CLASS_NAME) &&
+                                    JavaInfo.isAvailable(XALAN_TRANSFORMER_FACTORY_CLASS_NAME) &&
+                                    JavaInfo.isAvailable(SAX_LEXICAL_HANDLER_CLASS_NAME);
```
Azquelt: Comment needs updating I think?

IIRC, if we see the XLTXEJ transformer and the XALAN transformer are both available, we need to manually select the XALAN transformer, as we will hit problems with XLTXEJ?

Also unsure if the check for `SAX_LEXICAL_HANDLER_CLASS_NAME` is actually necessary.

Azquelt (approved):

wtlucy (approved):

tevans78:
#run-libby-bot

PR #18231, opened 2021-08-17
check the server status when a FAT test server fails to stop within the timeout.

WhiteCat22:
Only throw an exception if the status call returns anything other that 1

I've been seeing more and more tests fail because the server failed to stop within the 15 seconds (local, 30 seconds on remote builds) timeout.

This PR is the result of an old slack conversation I found and some experimenting.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -2454,9 +2454,16 @@ public class LibertyServer implements LogMonitorClient {
 
             int serverStopRC = output.getReturnCode();
             if (serverStopRC != 0) {
-                throw new RuntimeException("Server stop failed with RC " + serverStopRC +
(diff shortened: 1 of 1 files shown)


dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -2454,9 +2454,16 @@ public ProgramOutput stopServer(boolean postStopServerArchive, boolean forceStop
 
             int serverStopRC = output.getReturnCode();
             if (serverStopRC != 0) {
-                throw new RuntimeException("Server stop failed with RC " + serverStopRC +
-                                           ".\nStdout:\n" + output.getStdout() +
-                                           "\nStderr:\n" + output.getStderr());
+                if (serverStopRC >= 20) {
```
WhiteCat22, line 2457: only do additional processing if the RC was >= 20, anything less and there's probably a legitimate problem, so lets save the processing cycles.

PR #18241, opened 2021-08-18
Only allow a single checkpoint to be called

dettmoney:
When doing a "features" checkpoint, then changing the location of the keyStore in the server.xml, and finally doing a restore, the restore portion fails trying to do another checkpoint due to the config change. Add a test to ensure we can only call checkpoint once.

### DIFF
--- a/dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
+++ b/dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -119,6 +122,10 @@ public void check() {
     }
 
     void checkpoint(Phase phase) throws CheckpointFailed {
+        // Checkpoint can only be called once
+        if (checkpointCalled.get()) {
```
tjwatson: This should be:
```java
        if (!checkpointCalled.compareAndSet(false, true) {
            return;
        }
```

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -153,6 +160,7 @@ void checkpoint(Phase phase) throws CheckpointFailed {
             throw new CheckpointFailed(Type.SNAPSHOT_FAILED, "Failed to do checkpoint.", e, 0);
         }
         restore(phase, checkpointHooks);
+        checkpointCalled.set(true);
```
tjwatson: No need for this to be here if you use compareAndSet

tjwatson (changes requested):

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
dettmoney: Fixed up and added a test

dettmoney:
#build

tjwatson (approved):

PR #18242, opened 2021-08-18
Deleting duplicated suite class

andymc12:
This project contains both a `FATSuite` and `FATSuiteLite` class that includes the same test classes.  This project has been failing in Java 17 builds - possibly due to this duplication, though I suspect it may be something else. In any case, there seems to be nothing wrong with deleting the `FATSuiteLite` test class.


andymc12:
#build

jim-krueger (approved):

PR #18253, opened 2021-08-19
Add transaction isolation config to Persistence Service

dazey3:
#### Current persistence service failure:
```
org.postgresql.util.PSQLException: Cannot change transaction isolation level in the middle of a transaction.
at org.postgresql.jdbc.PgConnection.setTransactionIsolation(PgConnection.java:844)
at java.base/java.lang.reflect.Method.invoke(Method.java:566)
at org.postgresql.ds.PGPooledConnection$ConnectionHandler.invoke(PGPooledConnection.java:337)
at com.sun.proxy.$Proxy10.setTransactionIsolation(Unknown Source)
at com.ibm.ws.rsadapter.impl.WSRdbManagedConnectionImpl.setTransactionIsolation(WSRdbManagedConnectionImpl.java:3989)
at com.ibm.ws.rsadapter.impl.WSRdbManagedConnectionImpl.synchronizePropertiesWithCRI(WSRdbManagedConnectionImpl.java:2065)
at com.ibm.ws.rsadapter.impl.WSRdbManagedConnectionImpl.getConnection(WSRdbManagedConnectionImpl.java:2352)
at com.ibm.ejs.j2c.MCWrapper.getConnection(MCWrapper.java:1933)
at com.ibm.ejs.j2c.ConnectionManager.allocateConnection(ConnectionManager.java:318)
at com.ibm.ws.rsadapter.jdbc.WSJdbcDataSource.getConnection(WSJdbcDataSource.java:140)
at com.ibm.ws.rsadapter.jdbc.WSJdbcDataSource.getConnection(WSJdbcDataSource.java:114)
at org.eclipse.persistence.sessions.JNDIConnector.connect(JNDIConnector.java:135)
at com.ibm.wsspi.persistence.internal.eclipselink.WrappingConnector.connect(WrappingConnector.java:40)
at org.eclipse.persistence.sessions.DatasourceLogin.connectToDatasource(DatasourceLogin.java:162)
...
at org.eclipse.persistence.queries.ObjectLevelReadQuery.executeInUnitOfWork(ObjectLevelReadQuery.java:1230)
at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.internalExecuteQuery(UnitOfWorkImpl.java:2900)
at org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery(AbstractSession.java:1866)
at org.eclipse.persistence.internal.sessions.AbstractSession.retryQuery(AbstractSession.java:1936)
at org.eclipse.persistence.sessions.server.ClientSession.retryQuery(ClientSession.java:698)
at org.eclipse.persistence.internal.sessions.UnitOfWorkImpl.retryQuery(UnitOfWorkImpl.java:5540)
at org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery(AbstractSession.java:1902)
at org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery(AbstractSession.java:1848)
at org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery(AbstractSession.java:1813)
at org.eclipse.persistence.internal.jpa.QueryImpl.executeReadQuery(QueryImpl.java:258)
at org.eclipse.persistence.internal.jpa.QueryImpl.getResultList(QueryImpl.java:473)
at com.ibm.ws.concurrent.persistent.db.DatabaseTaskStore.findUnclaimedTasks(DatabaseTaskStore.java:796)
```

#### Explanation:

The Persistence Service first sets the Transaction Isolation to TRANSACTION_READ_UNCOMMITTED on EclipseLink's `org.eclipse.persistence.sessions.DatabaseLogin` API.

EclipseLink uses that isolation level to set the isolation level on the first connection: 
`org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.checkTransactionIsolation()`

Then, that first query fails with an exception (table does not exist).

EclipseLink attempts to determine if the failure was due to a communication issue with the database. 
EclipseLink executes a "ping" query to determine if it was a communication failure: 
org.eclipse.persistence.internal.databaseaccess.DatabasePlatform.wasFailureCommunicationBased

EclipseLink determines that the failure was a communication issue because of the  `java.sql.SQLException` from the ping query: 
org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.processExceptionForCommError

Then, EclipseLink sees that the query was a readQuery and a communication issue, so EclipseLink tries to retry executing the query: 
org.eclipse.persistence.internal.sessions.AbstractSession.executeQuery

In order to re-execute the query, EclipseLink tries to get a new connection from the datasource pool.

However, connection sharing is enabled, and connectionSharing=MatchOriginalRequest is used (by default), so the connection pool matches that same connection by its original request, and attempts to set the TRANSACTION_READ_COMMITTED isolation level that we are asking for before handing it back to. That is where it fails because PostgreSQL is very strict.

#### Solution:

One solution is to remove the `org.eclipse.persistence.sessions.DatabaseLogin` API call to set transaction isolation within EclipseLink and instead set the transaction isolation on the `javax.sql.DataSource` so that all Connections will be the expected isolation level.

### DIFF
--- a/dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/DatabaseStore.java
+++ b/dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/DatabaseStore.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015 IBM Corporation and others.
+ * Copyright (c) 2015, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -10,23 +10,43 @@
  *******************************************************************************/
 package com.ibm.wsspi.persistence;
 
+import java.util.Map;
+
 /**
  * Enables configuration of a database as a persistent store that can be shared by multiple components.
  */
 public interface DatabaseStore {
+
     /**
      * Create a persistence service unit for the specified entity classes.
      * The invoker of this method is responsible for closing the persistence service unit
      * and for participating in DDL generation.
-     * 
-     * The persistence service feature is currently not supported for informix.
-     * 
+     *
      * @param loader class loader for the entity classes.
      * @param entityClasses list of entity classes.
      * @return the persistence service unit.
      * @throws Exception if a failure occurs.
+     * @throws UnsupportedOperationException for Informix
      */
     PersistenceServiceUnit createPersistenceServiceUnit(ClassLoader loader, String... entityClassNames) throws Exception;
+
+    /**
+     * Create a persistence service unit for the specified entity classes.
+     * Allows the invoker to pass configuration properties.
+     * The invoker of this method is responsible for closing the persistence service unit
+     * and for participating in DDL generation.
+     *
+     * The persistence service feature is currently not supported for informix.
+     *
+     * @param loader class loader for the entity classes.
+     * @param properties map of configuration properties
+     * @param entityClasses list of entity classes.
+     * @return the persistence service unit.
+     * @throws Exception if a failure occurs.
+     * @throws UnsupportedOperationException for Informix
+     * @throws UnsupportedOperationException if a configuration property is not supported
+     */
+    PersistenceServiceUnit createPersistenceServiceUnit(ClassLoader loader, Map properties, String... entityClassNames) throws Exception;
 
     /**
      * Get the configured schema name.
--- a/dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
+++ b/dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2019 IBM Corporation and others.
+ * Copyright (c) 2014, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -149,10 +149,10 @@ public class DatabaseStoreImpl implements DatabaseStore {
     @Activate
     @Trivial
     protected void activate(ComponentContext context) throws Exception {
-        properties = context.getProperties();
+        this.properties = context.getProperties();
         final boolean trace = TraceComponent.isAnyTracingEnabled();
         if (trace && tc.isEntryEnabled())
-            Tr.entry(this, tc, "activate", properties);
+            Tr.entry(this, tc, "activate", this.properties);
 
         // The database store should be lazily initialized by the components using it, so there is no need
         // for the database store implementation to have its own lazy initialization.
@@ -166,12 +166,20 @@ public class DatabaseStoreImpl implements DatabaseStore {
      */
     @Override
     public PersistenceServiceUnit createPersistenceServiceUnit(ClassLoader loader, String... entityClassNames) throws Exception {
+        return createPersistenceServiceUnit(loader, new HashMap<String, Object>(), entityClassNames);
+    }
+
+    /**
+     * @see com.ibm.wsspi.persistence.DatabaseStore#createPersistenceServiceUnit(java.lang.ClassLoader, java.util.Map properties, java.lang.String[])
+     */
+    @Override
+    public PersistenceServiceUnit createPersistenceServiceUnit(ClassLoader loader, Map properties, String... entityClassNames) throws Exception {
         final boolean trace = TraceComponent.isAnyTracingEnabled();
         if (trace && tc.isEntryEnabled())
             Tr.entry(this, tc, "createPersistenceServiceUnit", loader, Arrays.asList(entityClassNames));
 
         Map<String, Object> puProps = new HashMap<String, Object>();
-        tablePrefix = (String) properties.get("tablePrefix");
+        tablePrefix = (String) this.properties.get("tablePrefix");
         tablePrefixLength = tablePrefix.length();
         for (int i = 0; i < tablePrefixLength; i++) {
             int codepoint = tablePrefix.codePointAt(i);
@@ -181,7 +189,7 @@ public class DatabaseStoreImpl implements DatabaseStore {
             }
         }
 
-        schema = (String) properties.get("schema");
+        schema = (String) this.properties.get("schema");
         schemaLength = schema == null ? -1 : schema.length();
         for (int i = 0; i < schemaLength; i++) {
             int codepoint = schema.codePointAt(i);
@@ -192,19 +200,34 @@ public class DatabaseStoreImpl implements DatabaseStore {
         }
 
         // Look for EclipseLink persistence properties
-        if (properties.get("persistenceProperties.0.config.referenceType") != null) {
-            for (String key : Collections.list(properties.keys())) {
+        if (this.properties.get("persistenceProperties.0.config.referenceType") != null) {
+            for (String key : Collections.list(this.properties.keys())) {
                 if (key.startsWith("persistenceProperties.0") && !key.equals("persistenceProperties.0.config.referenceType")) {
-                    Object value = properties.get(key);
+                    Object value = this.properties.get(key);
                     key = key.substring(24);
                     puProps.put(key, value);
                 }
             }
         }
 
+        // Look for additional configuration properties
+        int isolationLevel = Connection.TRANSACTION_READ_COMMITTED;
+        if (properties.get("transactionIsolationLevel") != null) {
+            Object configIsolationLevel = properties.get("transactionIsolationLevel");
+
+            if(configIsolationLevel instanceof Number) {
+                int newisolationLevel = ((Number) configIsolationLevel).intValue();
+                if(supportsIsolationLevel(newisolationLevel)) {
+                    isolationLevel = newisolationLevel;
+                } else {
+                    throw new UnsupportedOperationException(Tr.formatMessage(tc, "UNSUPPORTED_ISOLATION_LEVEL_CWWKD0293E", isolationLevel));
+                }
+            }
+        }
+
         ResourceConfig resourceInfo = resourceConfigFactory.createResourceConfig(DataSource.class.getName());
         resourceInfo.setSharingScope(ResourceConfig.SHARING_SCOPE_SHAREABLE);
-        resourceInfo.setIsolationLevel(Connection.TRANSACTION_READ_COMMITTED);
+        resourceInfo.setIsolationLevel(isolationLevel);
         resourceInfo.setResAuthType(ResourceConfig.AUTH_CONTAINER);
 
         if (authDataRef != null) {
@@ -246,7 +269,7 @@ public class DatabaseStoreImpl implements DatabaseStore {
             throw new IllegalStateException(errMsg);
         }
 
-        strategy = (String) properties.get("keyGenerationStrategy");
+        strategy = (String) this.properties.get("keyGenerationStrategy");
         if ("AUTO".equals(strategy)) {
             strategy = dbProductName.contains("oracle") ? "SEQUENCE"
                             : dbProductName.contains("adaptive server") || dbProductName.contains("sybase") ? "TABLE"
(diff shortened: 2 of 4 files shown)


dazey3:
#build #spawn.fullfat.buckets=com.ibm.ws.persistence_fat,com.ibm.ws.concurrent.persistent_fat

njr-11 (approved):

jgrassel (approved):

dazey3:
@jgrassel @njr-11 

Running against Oracle fails with the exception:
```
Stack Dump = java.sql.SQLException: READ_COMMITTED and SERIALIZABLE are the only valid transaction levels
	at oracle.jdbc.driver.PhysicalConnection.setTransactionIsolation(PhysicalConnection.java:3302)
	at oracle.jdbc.OracleConnectionWrapper.setTransactionIsolation(OracleConnectionWrapper.java:220)
	at com.ibm.ws.rsadapter.impl.WSRdbManagedConnectionImpl.setTransactionIsolation(WSRdbManagedConnectionImpl.java:3989)
	at com.ibm.ws.rsadapter.impl.WSRdbManagedConnectionImpl.synchronizePropertiesWithCRI(WSRdbManagedConnectionImpl.java:2065)
	at com.ibm.ws.rsadapter.impl.WSRdbManagedConnectionImpl.<init>(WSRdbManagedConnectionImpl.java:530)
	at com.ibm.ws.rsadapter.impl.WSManagedConnectionFactoryImpl.createManagedConnection(WSManagedConnectionFactoryImpl.java:696)
	at com.ibm.ejs.j2c.FreePool.createManagedConnectionWithMCWrapper(FreePool.java:1373)
	at com.ibm.ejs.j2c.FreePool.createOrWaitForConnection(FreePool.java:1247)
	at com.ibm.ejs.j2c.PoolManager.reserve(PoolManager.java:1442)
	at com.ibm.ejs.j2c.ConnectionManager.allocateMCWrapper(ConnectionManager.java:574)
	at com.ibm.ejs.j2c.ConnectionManager.allocateConnection(ConnectionManager.java:306)
	at com.ibm.ws.rsadapter.jdbc.WSJdbcDataSource.getConnection(WSJdbcDataSource.java:140)
	at com.ibm.ws.rsadapter.jdbc.WSJdbcDataSource.getConnection(WSJdbcDataSource.java:114)
	at com.ibm.wsspi.persistence.internal.DatabaseStoreImpl.getDatabaseProductName(DatabaseStoreImpl.java:748)
	at com.ibm.wsspi.persistence.internal.DatabaseStoreImpl.createPersistenceServiceUnit(DatabaseStoreImpl.java:252)
	at com.ibm.ws.concurrent.persistent.db.DatabaseTaskStore.getPersistenceServiceUnitReadUncommitted(DatabaseTaskStore.java:968)
	at com.ibm.ws.concurrent.persistent.db.DatabaseTaskStore.findUnclaimedTasks(DatabaseTaskStore.java:788)
```

It appears that creating a PersistenceServiceUnit with READ_UNCOMMITTED, then trying to determine if the database platform is Oracle, so that a new PersistenceServiceUnit can be created with READ_COMMITTED, causes the concurrent persistence to fail for Oracle.

Also, the method concurrent persistence uses to determine the database platform is to use reflection and use EclipseLink specific API. That isn't the CLEANEST implementation and I think we should remove that from concurrent persistence. Instead, we can open the Persistence API to allow users to ask the `com.ibm.wsspi.persistence.DatabaseStore`, which requires a DataSource anyway, what the database platform is. We can create a temporary DataSource from the datasourceFactory to look at the metadata. 
This should be less intensive than creating a full EntityManagerFactory and relying on EclipseLink specific API.

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/DatabaseStore.java
```diff
@@ -10,24 +10,53 @@
...
+
+    /**
+     * Obtain the database product name from the database to which the data source connects.
+     * <p>
+     * {@link com.ibm.ws.jdbc.WSDataSource#getDatabaseProductName()}
+     * {@link java.sql.DatabaseMetaData#getDatabaseProductName()}
+     *
+     * @return the database product name or 'UNKNOWN' if unable
+     */
+    public String getDatabaseProductName();
```
njr-11: Rather than going to all of the trouble of getting the db product name and interpreting the meaning of the value, you could just ask the DatabaseMetaData directly if it supports the isolation level,
https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#supportsTransactionIsolationLevel(int)

dazey3: The persistence service would first create a DataSource with `Connection.TRANSACTION_READ_COMMITTED`, so that we can get a connection and therefore the `DatabaseMetaData`, so that we can ask if it supports `Connection.TRANSACTION_READ_UNCOMMITTED`? Then, if it does support it, we create a new DataSource with `Connection.TRANSACTION_READ_UNCOMMITTED` and throw away the old one anyway. And all this logic under 

Getting access to the javax.sql.DatabaseMetaData doesn't seem quicker and makes the persistence service care more about this decision that they aren't responsible for in the first place. Would it make more sense if we maybe could ask `com.ibm.ws.jdbc.WSDataSource` if they support the isolation level?

dazey3:
@jgrassel @njr-11 

I took @njr-11 suggestion to use https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#supportsTransactionIsolationLevel(int) and updated this PR with a possible implementation using that API. Thoughts? 

Any suggestions on more debug would be appreciated.

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/DatabaseStore.java
njr-11: WSDataSource doesn't actually know anything. I'm pretty sure it would need to get the DatabaseMetaData and ask it

dazey3:
#build #spawn.fullfat.buckets=com.ibm.ws.persistence_fat,com.ibm.ws.concurrent.persistent_fat

njr-11 (approved):

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
```diff
@@ -738,6 +781,52 @@ else if (suspendedLTC != null)
...
+                Connection con = dataSource.getConnection();
+                try {
+                    return con.getMetaData().supportsTransactionIsolationLevel(isolationLevel);
+                } finally {
+                    con.close();
+                }
+            } finally {
+                try {
+                    if (tranStarted)
+                        tranMgr.commit();
```
jgrassel: Are we changing anything here?  If not, might be better to rollback than commit.

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
```diff
@@ -738,6 +781,52 @@ else if (suspendedLTC != null)
...
+                } finally {
+                    // resume
+                    if (suspendedTran != null)
+                        tranMgr.resume(suspendedTran);
+                    else if (suspendedLTC != null)
+                        localTranCurrent.resume(suspendedLTC);
+                }
+            }
+           } catch (Exception e) {
+               if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
```
jgrassel: Is this something we might want to FFDC?  If something's going wrong here in the field, we won't know until trace is turned on.

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
```diff
@@ -192,19 +200,54 @@ public PersistenceServiceUnit createPersistenceServiceUnit(ClassLoader loader, S
...
                 if (key.startsWith("persistenceProperties.0") && !key.equals("persistenceProperties.0.config.referenceType")) {
-                    Object value = properties.get(key);
+                    Object value = this.properties.get(key);
                     key = key.substring(24);
                     puProps.put(key, value);
                 }
             }
         }
 
+        // Look for additional configuration properties
```
jgrassel, line 213: This is a big chunk of specialized code to add to this method, I wonder if the code involved in determining the isolation level should be moved to its own method, and we just use the value returned by that method in resourceInfo.setIsolationLevel().  This way it's less of a change to the original method body.

jgrassel (changes requested):

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
dazey3: If an error occurs getting the resource from the factory, or while trying to establish a connection, should we fail this whole process or just default back to READ_COMMITTED and proceed?

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
dazey3: Sound fine to me

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
dazey3: Sure, I'll move to a sub-method

dev/com.ibm.ws.persistence/src/com/ibm/wsspi/persistence/internal/DatabaseStoreImpl.java
jgrassel: They requested READ_UNCOMMITTED for a reason, which could lead to concurrent behaving unexpectedly or erroneously.  Without a FFDC or warning, the reason for this would be silent without trace enabled, and users are not likely going to have trace enabled until asked to enable it by support.  If there's a warning or FFDC, then they can at least bring that up at the very beginning when they engage support.

njr-11: While defaulting to READ_COMMITTED might be convenient for the particular scenario where persistent executor is using the persistence service with Oracle, behavior of the persistence service needs to be generally applicable for all components, and therefore an exception should go back to the caller.  Whether or not to FFDC as well should depend on whether we expect this to be something the caller will trap and handle another way.  If you have externalized the supportsIsolationLevel information to the caller and pattern is for the caller to create the READ_COMMITTED persistence unit, then ask it whether the database also supports READ_UNCOMMITTED then the caller has a way to know whether creation of the persistence unit with READ_UNCOMMITTED is valid or not before attempting it, in which case I would say go ahead and log an FFDC because they shouldn't be attempting this.  If, however, no provision is made for the caller to be able to detect whether the isolation level is supported, such that attempting to create the persistence unit and trapping for an exception _is_ the mechanism, then do not log FFDC because the exception is part of the expected programming model.  Sorry  that my answer is conditional, but github has marked this comment as outdated and will not place it in the code so I can't really see which way you have chosen or are planning to choose here.

njr-11 (approved):

dazey3:
The latest design is that `com.ibm.wsspi.persistence.internal.DatabaseStoreImpl.createPersistenceServiceUnit()` will throw an UnsupportedOperationException if the configuration property is unsupported. It is the caller's responsibility to catch that exception and make appropriate adjustments.

For the isolation level, if the driver does not support the transaction isolation level supplied, then an UnsupportedOperationException is thrown. `com.ibm.ws.concurrent.persistent.db.DatabaseTaskStore.getPersistenceServiceUnitReadUncommitted() ` catches that exception and falls back to creating a default PersistenceServiceUnit with no additional configuration.

jgrassel (approved):

dazey3:
#build #spawn.fullfat.buckets=com.ibm.ws.persistence_fat,com.ibm.ws.concurrent.persistent_fat

dazey3:
#run-libby-bot

PR #18270, opened 2021-08-19
Added missing transformations

isaacrivriv:
Added missing transformations required for CL lWAS bucket to properly function


isaacrivriv:
Closing PR went with a better way to override rule transformations in the test that needed it instead of putting it in the transformer itself

PR #18292, opened 2021-08-23
Fix missing tested.features in Security WIM FAT

tkburroughs:
Fix an open-liberty image conflict with the simplicity feature dependency
checker fix; account for running on open-liberty only images
where servlet-3.0 does not exist, so servlet-3.1 is used instead.

Note: servlet-3.1 either directly or indirectly via auto-features
enables distributedMap-1.0 and jndi-1.0, so those do not need to
be included in the test.features list.


tkburroughs:
#build

tkburroughs:
#build

kristip17 (approved):

PR #18293, opened 2021-08-23
Add a checkpoint action

tjwatson:
Checkpoint is now done with the following:

server checkpoint <server-name> --at=features|applications

Restore happens when you issue the run action

server run <server-name>

For now we do not allow for a restore from the start action.  If start is done on a checkpoint server then a normal server start is done, not from a checkpoint.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -825,12 +825,12 @@ javaCmd()
     if [ -n "${option}" ]; then
       if [ -z "${option##-D*}" ] || [ -z "${option##-X*}" ]; then
         JVM_OPTIONS_QUOTED="$JVM_OPTIONS_QUOTED '$option'"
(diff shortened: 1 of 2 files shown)


tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1118,14 +1155,19 @@ export INVOKED
 case "$ACTION" in
 
   # Start the server in the foreground
-  run | debug)
+  run | debug | checkpoint)
+
+    if restoreServer false
```
sebratton: I don't get this syntax. Waht is the trailing `false` in
`if restoreServer false`

sebratton (approved):
It's not clear how a someone 'recovers' from a checkpointed state if after checkpoint someone wants to do a non-checkpointed run/debug `--clean`? Maybe a normal start should remove the workarea/checkpoint dir. 

...actually, a `--clean` will not work since it's executed by the launcher which never executes in the case of a criu restore. So it seems that manually cleaning the workarea/checkpoint dir is the only way tot reset.

sebratton (approved):

PR #18308, opened 2021-08-24
The JVM option -XX:MaxPermSize is not valid for Java 17

tjwatson:
As a result the test server would fail to even start
Java 17.  It is unclear why the test was using this
particular JVM option.  It seems just as good to use
any one of the expected -D JVM arguments to verify
the dump is correct


tjwatson:
#build

gjwatts (approved):

PR #18317, opened 2021-08-25
Check for null BundleContext

tjwatson:
The bundle could be stopped or uninstalled which makes the
BundleContext be null.  We should handle this and carry on.


tjwatson:
#build

sebratton (approved):

PR #18334, opened 2021-08-26
Always write to a console log

tjwatson:
This allows for successful restore in the background (server start) even
when the checkpoint is done in the foreground.  When restoring in the
foreground (server run) a tail process is used to display the
console.log to the standard out console.  When restoring the existing
console.log is saved off to checkpoint-console.log and a new console.log
is started.


tjwatson:
#build

sebratton (approved):

PR #18345, opened 2021-08-27
Fix EJB remote FAT test order bug

tkburroughs:
Add JNDI lookup with iiop port prior to JNDI
lookup with iiops port to ensure consistent behavior.


tkburroughs:
#build

olendvcook (approved):

PR #18354, opened 2021-08-27
Remove dependency on MP Config from JAX-RS 3.0

andymc12:
RESTEasy made it's dependency on MicroProfile Config to be optional.  We don't want to use it in Liberty (we want to keep the config function isolated - allowing the server to be leaner), so we (JAX-RS / RESTful WS 3.0) need to explicitly exclude that dependency.

### DIFF
--- a/dev/io.openliberty.org.jboss.resteasy.common/bnd.bnd
+++ b/dev/io.openliberty.org.jboss.resteasy.common/bnd.bnd
@@ -1,5 +1,5 @@
 #*******************************************************************************
-# Copyright (c) 2020 IBM Corporation and others.
+# Copyright (c) 2020, 2021 IBM Corporation and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License v1.0
 # which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


andymc12:
#build

dev/io.openliberty.org.jboss.resteasy.common/bnd.bnd
```diff
@@ -128,6 +128,7 @@ Import-Package: \
   !com.sun.xml.bind.marshaller, \
   !org.apache.cxf.*, \
   !org.apache.james.mime4j.*, \
+  !org.eclipse.microprofile.config.*, \
```
tevans78, line 131: Rather than being removed completely, doesn't this become an optional dependency on MP Config 3.0?

andymc12:
Current build seems hung - trying new #build

dev/io.openliberty.org.jboss.resteasy.common/bnd.bnd
andymc12: This line will prevent the resteasy.common bundle from importing MP Config packages.

tevans78: if someone does enable the mpConfig-3.0 feature, shouldn't resteasy be able to use it as well? We can address that as a follow-up though

WhiteCat22 (approved):

PR #18385, opened 2021-08-31
Add delay for Boulder revoke on Acme FAT

kristip17:
Hit the same thing as #17785 in another test. It appears that there is some variability on when the revoke is complete on the Boulder server. `Boulder: welcome-to-the-purge, purge expected in: 153s`

RTC 286124


kristip17:
#build

jvanhill (approved):

PR #18415, opened 2021-09-01
Ensure get(Class<?>) works even when response has no content-type header

andymc12:
This fixes a `@SkipForRepeat` issue where the cause of the problem was that the HTTP response did not contain a `Content-type` header.  This prevents RESTEasy's client implementation from automatically converting the `Response` to a `String`.


andymc12:
#build

andymc12:
Build seems hung building Liberty IM images... trying again with new #build

andymc12:
#build

andymc12:
Need to add `@Sensitive` to overlaid class to avoid tracing sentitive data. New #build

WhiteCat22 (approved):

andymc12:
#libby

Issue #18439, opened 2021-09-03
netty - implement UDP component


wtlucy:
similar to #18082. This is required for #15779.

Issue #18081, opened 2021-08-06
Modularize com.ibm.ws.channelfw bundle


wtlucy:
The bundle `com.ibm.ws.channelfw` provides all of the channel framework interfaces and implementations. This bundle needs to be split up into a handful of new bundles so that we can share existing implementations with the new netty implementation.

hutchig:
I will close this for now, we can always open a phase 2 one if needed for wsspi renaming or whatever.

Issue #15669, opened 2021-01-26
netty startup and shutdown lifecycle


pnicolucci:
- define and create a set of utilities that can be used to ensure that new netty endpoints follow the same initialization steps and behaviors as existing channelfw chains
- similarly, create common utils that can be used to ensure that endpoint teardown and update behaviors don't change.
- consider server quiesce and port open timing
- special attention must be given to info and warning messages, which need to be preserved

PR #17940, opened 2021-07-22
Split the WsByteBuffer classes off into a seperate bundle

hutchig:
This was originally a draft pull request.

PR #18441, opened 2021-09-03
Netty framework and TCP, UDP support

wtlucy:
For #15669 and #18082 and #18439

This PR sets up a foundation for the use of Netty in Liberty. A few key parts:
- A new bundle `io.openliberty.netty.internal` which provides via `NettyFramework` the internal APIs Liberty bundles will need to manage endpoints
- An new bundle `io.openliberty.netty.internal.impl` which provides implementations for the APIs defined in `NettyFramework`, via a new `NettyFrameworkImpl` service
- A TCP component which provides the behaviors, messages, and (public) config options previously provided by the channelfw
- A UDP component which mirrors the behaviors previously provided by the channelfw

Note this code currently has a runtime dependency on the channelfw bundle. This is needed so the [channelfw config](https://www.openliberty.io/docs/latest/reference/config/channelfw.html) can continue to be read. This dependency will be less problematic once #18081 progresses.


wtlucy:
#build

mrsaldana (approved):
Newest commit looks good

dev/io.openliberty.netty.internal.impl/bnd.bnd
```diff
@@ -0,0 +1,64 @@
...
+bVersion=1.0
+
+Bundle-Name: Liberty Netty Transport Implementation
+Bundle-SymbolicName: io.openliberty.netty.internal.impl
+Bundle-Description: Liberty Netty Transport Implementation ${bVersion}
+WS-TraceGroup: Netty
+
+Import-Package: \
+  io.openliberty.netty.internal,\
+  *
```
hutchig: Your choice, but might be better as
	${defaultPackageImport}
?

hutchig: If you do go back in, a lot of them have the Export-Package before the imports but again, your choice.

dev/io.openliberty.netty.internal.impl/bnd.bnd
```diff
@@ -0,0 +1,64 @@
...
+# reuse the existing translated messages from the TCP Channel
+Private-Package: \
+  com.ibm.ws.tcpchannel.internal.resources, \
+  com.ibm.ws.udpchannel.internal.resources
+
+-dsannotations: \
+  io.openliberty.netty.internal.impl.NettyFrameworkImpl
+
+-buildpath: \
+    com.ibm.ws.channelfw;version=latest,\
```
hutchig: I don't think there will ever be more than one com.ibm.ws.channelfw project in a build.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+/**
+ * Liberty NettyFramework implementation bundle
+ */
+@Component(configurationPid = "io.openliberty.netty.internal",
+    immediate = true, 
+    service = { NettyFramework.class, ServerQuiesceListener.class }, 
+    property = { "service.vendor=IBM" })
+public class NettyFrameworkImpl implements ServerQuiesceListener, NettyFramework {
+
+    private static final TraceComponent tc = Tr.register(NettyFrameworkImpl.class, NettyMessageConstants.NETTY_TRACE_NAME, NettyMessageConstants.NETTY_BUNDLE);
```
hutchig: Would NettyConstants be a more genericly useful classname as in other examples non message related contants are also held in the equivalent class.
Also is this the 'NETTY_BUNDLE'? Not really I think, more the 'BASE_BUNDLE' conventionally perhaps.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+     *
+     * @param bundle CHFWBundle instance to unset
+     */
+    protected void unsetChfwBundle(CHFWBundle bundle) {}
+
+    @Activate
+    protected void activate(ComponentContext context, Map<String, Object> config) {
+        // use the executor service provided by Liberty
+        parentGroup = new NioEventLoopGroup(1, executorService);
+        childGroup = new NioEventLoopGroup(0, executorService);
```
hutchig, line 108: Is it worth a comment that 0 for the threads uses the 'default' number of threads - and is is OK for the external library to have control over this when we have 'io.openliberty.io.netty;version=latest' in the bnd file (so I guess it could change if netty version changes?)

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+     * When notified that the server is going to stop, pre-quiesce all chains in the runtime.
+     * This will be called before services start getting torn down..
+     *
+     * @see com.ibm.wsspi.kernel.service.utils.ServerQuiesceListener#serverStopping()
+     */
+    @Override
+    public void serverStopping() {
+        if (isActive) {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
+                Tr.event(this, tc, "Destroying all endpoints");
```
hutchig: Maybe it is preferable to preserve this trace string as service will be familiar otherwise "Closing all channels" reads more accurately.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+            // set up future to log channel stopped messages
+            for (Channel channel : activeChannels) {
+                channel.closeFuture().addListener(future -> {
+                    logChannelStopped(channel);
+                });
+            }
+            Future<?> parent = null;
+            Future<?> child = null;
+            if (parentGroup != null) {
+                parent = parentGroup.shutdownGracefully(0, timeout, TimeUnit.MILLISECONDS);
```
hutchig, line 170: Not sure, without looking at the source code if a quiet period (first param) of 0 is defaulting to 2 seconds or results in no required quiet period.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+
+        synchronized (syncStarted) {
+            serverCompletelyStarted.set(true);
+            isActive = true;
+            syncStarted.notifyAll();
+        }
+    }
+
+    /**
+     * Method is called to run a task if the server has already started, if the server has not started that task is queue to be run when the server start signal
```
hutchig: queue -> queued

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+                serverStartedTasks.add(callable);
+                return null;
+            }
+        }
+        return callable.call();
+    }
+
+    /*
+     * If the server has not completely started, then wait until it has been.
+     * The server will be "completely" stated when the server start signal has been received and any task waiting on that signal before running have now been run.
```
hutchig: stated -> started
task -> tasks (or have now->has)

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+     */
+    @FFDCIgnore({ InterruptedException.class })
+    public static void waitServerCompletelyStarted() {
+        synchronized (syncStarted) {
+            if (serverCompletelyStarted.get() == false) {
+                try {
+                    syncStarted.wait();
+                } catch (InterruptedException x) {
+                    // assume we can go one then
```
hutchig: one -> on   ??

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+                }
+            }
+        }
+        return;
+    }
+
+    /**
+     * non-blocking method to return the state of server startup with respect to the server being completely started.
+     * The server will be "completely" stated when the server start signal has been received and any task waiting on that signal before running have now been run.
```
hutchig: same comment as above task-has or tasks-have

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+        return serverCompletelyStarted.get();
+    }
+
+    /**
+     * Declarative Services method for unsetting the ServerStarted service
+     *
+     * @param ref reference to the service
+     */
+    protected synchronized void unsetServerStarted(ServiceReference<ServerStarted> ref) {
+        // server is shutting down
```
hutchig, line 281: serverCompletelyStarted.set(false); ?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
```diff
@@ -0,0 +1,359 @@
...
+    @Override
+    public ChannelFuture start(BootstrapExtended bootstrap, String inetHost, int inetPort, ChannelFutureListener bindListener) throws NettyException {
+        return UDPUtils.start(this, bootstrap, inetHost, inetPort, bindListener);
+    }
+
+    @Override
+    public ChannelFuture stop(Channel channel) {
+        if (isActive && channel.isOpen()) {
+            ChannelFuture closeFuture = channel.close();
+            // TODO: should we wait until close actually completes to remove?
```
hutchig: If we did the server stop would send a second close as it loops round the activeChannels in serverStopping without an obvious equivalent of 
```if (isActive && channel.isOpen())``` 
in the serverStopping loop for any channel that was 'slow' to close and thus closing when the server was stopped. 
The inverse is not true (serverStopping sets isActive to false).

File: dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java [hunk 13]
```diff
@@ -0,0 +1,359 @@
...
+     *
+     * @see com.ibm.wsspi.kernel.service.utils.ServerQuiesceListener#serverStopping()
+     */
+    @Override
+    public void serverStopping() {
+        if (isActive) {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
+                Tr.event(this, tc, "Destroying all endpoints");
+            }
+            isActive = false;
```
hutchig, line 157: need to be careful as this flag is used in an if statement in the stop(channel) method and setting it false will noop that method (if my reading is correct) so it would be ineffectual if used in the serverStopping impl/loop below (directly or indirectly)

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyMessageConstants.java
```diff
@@ -8,16 +8,13 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
-package io.openliberty.netty.internal.codec;
+package io.openliberty.netty.internal.impl;
 
-import io.netty.channel.CombinedChannelDuplexHandler;
+public interface NettyMessageConstants {
```
hutchig: NettyConstants is better IMHO

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyMessageConstants.java
```diff
@@ -8,16 +8,13 @@
...
 
-/**
- * Codec for converting ByteBuf to WsByteBuffer and vice versa
- */
-public class WsByteBufferCodec extends CombinedChannelDuplexHandler<ByteBufToWsByteBufferDecoder, WsByteBufferToByteBufEncoder> {
+    /** RAS trace bundle for NLS */
+    String NETTY_BUNDLE = "io.openliberty.netty.internal.impl.resources.NettyFrameworkMessages";
```
hutchig: BASE_BUNDLE is more conventional

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/TCPUtils.java
```diff
@@ -0,0 +1,124 @@
...
+import io.openliberty.netty.internal.tcp.TCPConfigurationImpl;
+
+public class TCPUtils {
+
+    private static final TraceComponent tc = Tr.register(NettyFrameworkImpl.class, TCPMessageConstants.NETTY_TRACE_NAME, TCPMessageConstants.TCP_BUNDLE);
+    private static final int timeBetweenRetriesMsec = 1000; // make this non-configurable
+
+    public static ServerBootstrapExtended createTCPBootstrap(NettyFrameworkImpl framework, Map<String, Object> tcpOptions) throws NettyException {
+        // TODO: toggle inbound
+        BootstrapConfiguration config = new TCPConfigurationImpl(tcpOptions, true);
```
hutchig, line 38: I was never allowed to merge TODOs previously. I had to raise an issue or do. If nothing else, it pollutes the TODO list of everyone else when they are used for unmerged feature work.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/TCPUtils.java
```diff
@@ -0,0 +1,124 @@
...
+                
+                // set common channel attrs
+                bindFuture.channel().attr(ConfigConstants.NameKey).set(channelName);
+                bindFuture.channel().attr(ConfigConstants.HostKey).set(inetHost);
+                bindFuture.channel().attr(ConfigConstants.PortKey).set(inetPort);
+
+                Tr.info(tc, TCPMessageConstants.TCP_CHANNEL_STARTED, new Object[] { channelName, inetHost, String.valueOf(inetPort) });
+            } else {
+                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                    Tr.debug(tc, "bindHelper failed due to: " + future.cause().getMessage());
```
hutchig: I think it would be a good idea to include the channelName, similarly to line 66 above?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/TCPUtils.java
```diff
@@ -0,0 +1,124 @@
...
+                }
+
+                if (retryCount > 0) {
+                    if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                        Tr.debug(tc, "attempt to bind again after a wait of " +timeBetweenRetriesMsec +"ms; " + retryCount + " attempts remaining");
```
hutchig: channelName?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/TCPUtils.java
```diff
@@ -0,0 +1,124 @@
...
+                @Override
+                public ChannelFuture call() {
+                    TCPConfigurationImpl config = (TCPConfigurationImpl) bootstrap.getConfiguration();
+                    int bindRetryCount = config.getPortOpenRetries();
+                    return bindHelper(framework, bootstrap, inetHost, inetPort, bindListener, bindRetryCount);
+                }
+            });
+        } catch (Exception e) {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "caught exception performing late cycle server startup task: " + e);
```
hutchig: It might get good to pull out the cause and message?  Or do you want to use an overloaded toString?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/TCPUtils.java
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+        return null;
+    }
+
+    public static void logChannelStopped(Channel channel) {
+        String channelName = channel.attr(ConfigConstants.NameKey).get();
```
hutchig, line 117: I was wondering about port unification as in https://netty.io/4.0/xref/io/netty/example/portunification/PortUnificationServerHandler.html, if the channel name would always be enough and if we would want/need channel.id

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/UDPUtils.java
```diff
@@ -0,0 +1,158 @@
...
+                } else {
+                    Tr.error(tc, UDPMessageConstants.BIND_FAILURE, new Object[] { channelName, inetHost, String.valueOf(inetPort) });
+                }
+            }
+        });
+        return bindFuture;
+    }
+
+    /**
+     * UDP start helper implementation
```
hutchig, line 101: ...but this start 'helper's method name is not postfixed by Helper but the bindHelper is?
Personally, in a *Utils class, for static methods that will have that classname on the call line
I think the postfix 'Helper' does not add much except extra chars. Your choice, but best to be consistent
unless these is some reason. Is it because bindHelper helps TCP and UDP binding?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/UDPUtils.java
```diff
@@ -0,0 +1,158 @@
...
+                    String newHost = inetHost;
+                    if (newHost.equals("*")) {
+                        newHost = "0.0.0.0";
+                    }
+                    address = new InetSocketAddress(newHost, inetPort);
+                    if (address.isUnresolved()) {
+                        final String channelName = ((UDPConfigurationImpl) bootstrap.getConfiguration()).getExternalName();
+                        Tr.error(tc, UDPMessageConstants.DNS_LOOKUP_FAILURE, 
+                                new Object[] { channelName, newHost, String.valueOf(inetPort) });
+                        throw new NettyException("local address unresolved");
```
hutchig: It would be helpful to tag on the host and port?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/UDPUtils.java
```diff
@@ -0,0 +1,158 @@
...
+                    }
+
+                    return bindHelper(framework, bootstrap, newHost, inetPort, bindRetryCount, bindRetryInterval, bindListener);
+                }
+            });
+        } catch (Exception e) {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "caught exception performing late cycle server startup task: " + e);
```
hutchig: "Error scheduling bind of UDP channels on server startup?"

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/AccessLists.java
```diff
@@ -0,0 +1,195 @@
+/*******************************************************************************
+ * Copyright (c) 2005, 2006 IBM Corporation and others.
```
hutchig: © dates?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/AccessLists.java
```diff
@@ -0,0 +1,195 @@
...
+     * @param remoteAddr
+     * @return boolean
+     */
+    public boolean accessDenied(InetAddress remoteAddr) {
+
+        String hostname = null; //F184719
+
+        // check the inclusion lists first to see if the client matches
+        if (includeAccess.getActive() || includeAccessNames.getActive()) {
+            boolean closeSocket = true;
```
hutchig, line 120: I think this would be better to be called 'denied' to be obvious it is the result, and the caller chooses what to do with the result - but if not new code then perhaps best left as is. Also, I presume often we would not open a socket then close it again, we would just not open it in the first place.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/FastSynchHashTable.java
```diff
@@ -0,0 +1,336 @@
+/*******************************************************************************
+ * Copyright (c) 2005, 2006 IBM Corporation and others.
```
hutchig: dates?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/FastSynchHashTable.java
```diff
@@ -0,0 +1,336 @@
...
+ * FastSyncHashTable.
+ * 
+ * This is a table implementation that does very little besides basic table operations.
+ * In making it fast, it does not maintain counts of items in the table, cannot grow or change in size once instantiated,
+ * and therefore has no global tablewide locks. This locks at the bucket level, providing for smaller granularity of
+ * of locking and therefore faster access of data. The main table is a 2-d array allowing for some variation. Both the
+ * hashCode() and equals() methods are used to locate and compare keys to find entries in the table.
+ * 
+ * Copied from {@link com.ibm.ws.bytebuffer.internal.FastSynchHashTable}
+ * TODO: get rid of this when we have a common WsByteBuffer bundle
```
hutchig: Convert TODO to issue?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/FastSynchHashTable.java
```diff
@@ -0,0 +1,336 @@
...
+     * Does not hold synchronization throughout...which means a window is open for
+     * a get to miss.
+     * 
+     * @param key
+     *            Object to identify the value by
+     * @param value
+     *            Object to store in the table
+     * @return value of the object in the cache
+     */
+    public Object forcePut(int key, Object value) {
```
hutchig: Can't find a caller of this.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/FilterListStr.java
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
+ * Copyright (c) 2005, 2006 IBM Corporation and others.
```
hutchig: Dates? If you want to leave them as file unchanged I have no strong opinion.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/FilterListSlowStr.java
```diff
@@ -0,0 +1,214 @@
+/*******************************************************************************
+ * Copyright (c) 2005, 2006 IBM Corporation and others.
```
hutchig: Dates?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/FilterListFastStr.java
```diff
@@ -0,0 +1,396 @@
+/*******************************************************************************
+ * Copyright (c) 2005, 2006 IBM Corporation and others.
```
hutchig: Dates?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/TCPConfigConstants.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * The size of the tcp socket receiving buffer
+     */
+    String RCV_BUFF_SIZE = "receiveBufferSize";
+
+    /**
+     * Minimum Port Value
+     */
+    int PORT_MIN = 0;
```
hutchig, line 83: Can we ever make use of port 0 to get the OS to give us a free port in OpenLiberty and track it etc?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/TCPConfigurationImpl.java
```diff
@@ -0,0 +1,1285 @@
...
+                    keyType = ValidateUtils.KEY_TYPE_INT;
+                    oldValue = this.commOption;
+                    if (convertIntegerValue(value) != oldValue) {
+                        result = ValidateUtils.VALIDATE_NOT_EQUAL;
+                    }
+                    continue;
+                }
+
+                if (key.equalsIgnoreCase("chainData")) {
+                    // this parameter is made available by the channel framework
```
hutchig, line 872: 'channel framework' seems odd and so does 'this channel' - is this really any channel, TCP channels?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
```diff
@@ -0,0 +1,467 @@
+/*******************************************************************************
+ * Copyright (c) 2003, 2006 IBM Corporation and others.
```
hutchig: Dates?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
```diff
@@ -0,0 +1,467 @@
...
+            outputConfigToTrace();
+        }
+    }
+
+    /**
+     * @see com.ibm.ws.udpchannel.internal.UDPNetworkLayer.initDatagramSocket(VirtualConnection)
+     */
+    @Override
+    public void applyConfiguration(Bootstrap bootstrap) {
+        // TODO: UDPNetworkLayer sets reuseaddr to false, but if we don't do that here we can't bind TCP and UDP to the same port...?
```
hutchig: Convert to issue? (retain comments of course just the "TODO" in merged code

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
```diff
@@ -0,0 +1,467 @@
...
+                if (key.equalsIgnoreCase(UDPConfigConstants.ADDR_INC_LIST)) {
+                    if (value instanceof String) {
+                        this.addressIncludeList = convertToArray((String) value);
+                    } else {
+                        this.addressIncludeList = (String[]) value;
+                    }
+                    continue;
+                }
+
+//                if (key.equalsIgnoreCase(ChannelFrameworkConstants.CHAIN_DATA_KEY)
```
hutchig: 'this channel' seems odd - does it mean this UDP channel impl does not make us if it?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    /**
+     * Query the external name of this channel.
+     *
+     * @return String
+     */
+    public String getExternalName() {
+        return this.externalName;
+//        return this.channelData.getExternalName();
```
hutchig: should a reader worry about this?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
```diff
@@ -0,0 +1,467 @@
...
+    /**
+     * Access the channel framework configuration object for this channel.
+     *
+     * @return ChannelData
+     */
+    public ChannelData getChannelData() {
+        return this.channelData;
+    }
+
+//    /**
```
hutchig: Why commented out/left?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
```diff
@@ -0,0 +1,467 @@
...
+    private static int DEFAULT_READ_BUFFER_SIZE = 1024000;
+    private Map<String, Object> channelProperties = null;
+
+
+    private ChannelData channelData;
+    private String hostname = null;
+    private int port = 0;
+    private int sendBufferSize = 0;
+    private int receiveBufferSize = 0;
+    private int channelReceiveBufferSize = UDPConfigConstants.MAX_UDP_PACKET_SIZE;
```
hutchig, line 43: I might understand why this is the max PACKET size and not RECEIVE_BUFFER_SIZE_MAX but just want to check it is what is meant?

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    /**
+     * Set the size of the bytebuffer to allocate when receiving data.
+     *
+     * @param size
+     */
+    public void setChannelReceiveBufferSize(int size) {
+        this.channelReceiveBufferSize = size;
+        if (size < 0 || size > UDPConfigConstants.MAX_UDP_PACKET_SIZE) {
```
hutchig, line 370: In this method, I think  we should be using "RECEIVE_BUFFER_SIZE_MAX"

dev/io.openliberty.netty.internal/bnd.bnd
```diff
@@ -12,27 +12,26 @@
...
+Bundle-Name: Liberty Netty Transport Framework
 Bundle-SymbolicName: io.openliberty.netty.internal
-Bundle-Description: Liberty Netty Transport Implementation ${bVersion}
-WS-TraceGroup: ChannelFramework
+Bundle-Description: Liberty Netty Transport Framework ${bVersion}
+WS-TraceGroup: Netty
 
--dsannotations: io.openliberty.netty.internal.NettyFramework,\
+Export-Package: \
+  io.openliberty.netty.internal, \
```
hutchig, line 21: Having 'internal' in the exported packages is understood - we will just have to remember that using
the buildpath and auto-generated Import list in clients' bnds will skip these.

hutchig (changes requested):
Only concrete thing I saw that the UDP buffer/packet max receive size [here](https://github.com/OpenLiberty/open-liberty/pull/18441#discussion_r707276251), the rest
was just comments really that you can feel free to 'resolve' once considered.

hutchig (approved):
I have discussed all my comments with Bill and am happy to approve based on Bill's plan.
There is nothing in the current code that is of material import.

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
wtlucy: I've opened an issue to revisit this https://github.com/OpenLiberty/open-liberty/issues/18515

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
wtlucy: `0` does result in no quiet period

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
wtlucy: As we discussed - this logic isn't currently _wrong_, but we'll probably need to improve this area in the future. I've opened https://github.com/OpenLiberty/open-liberty/issues/18518

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/UDPUtils.java
wtlucy: I agree with getting rid of the _Helper_ postifx

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/UDPUtils.java
wtlucy: this is actually handling the exception from `Callable.call()`, I think it's fine to leave it as-is

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/AccessLists.java
wtlucy: This is taken directly from the channelfw tcpchannel - I agree it would be better if we didn't have to accept the connection at all, but when this code is run the socket has been opened and we do need to close it.

wtlucy: either way we'll be changing this up later with https://github.com/OpenLiberty/open-liberty/issues/18438

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/TCPConfigurationImpl.java
wtlucy: I wanted to preserve config parsing as closely as possible for this PR

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
wtlucy: https://github.com/OpenLiberty/open-liberty/issues/18520

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/udp/UDPConfigurationImpl.java
wtlucy: As we discussed - this is confusing because the UDP channel keeps track of both a `receiveBufferSize` and a `channelReceiveBufferSize`. `receiveBufferSize` maps to `RECEIVE_BUFFER_SIZE_MAX`, and `channelReceiveBufferSize` maps to `MAX_UDP_PACKET_SIZE`.

wtlucy: Earlier I proposed changing this `MAX_UDP_PACKET_SIZE` to something like `CHANNEL_RECEIVE_BUFFER_SIZE_MAX`, but looking at this again I think there's some value in leaving it as `MAX_UDP_PACKET_SIZE`

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/TCPConfigConstants.java
wtlucy: both the channelfw and netty will get this behavior from the JDK https://docs.oracle.com/javase/7/docs/api/java/net/ServerSocket.html#ServerSocket%28int%29

dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/tcp/FastSynchHashTable.java
wtlucy: you've already fixed this with https://github.com/OpenLiberty/open-liberty/pull/17940 - I'll update this PR to use your new project

wtlucy:
code review commits and rebase #build

wtlucy:
#build

wtlucy:
#build

wtlucy:
#build

wtlucy:
I'm happy with the progress of the latest PB - this new code is only currently exercised via unittests. #run-libby-bot

PR #18457, opened 2021-09-06
Try to clean up global config instances

### LINKED ISSUE #18426: MicroProfile 4.x: OutOfMemoryError: GC overhead limit exceeded when starting/stopping lots of apps
When running the CDI TCK with `microProfile-4.1` added to the Liberty server feature list, the server eventually runs out of memory, throwing `java.lang.OutOfMemoryError: GC overhead limit exceeded` from all of its threads.  This behavior does not normally occur with this TCK, and is only present when the microProfile feature is included in the server config.

The CDI TCK consists of hundreds of applications that are added to the server, started, and then stopped as the tests execute.  Analysis of a heap dump collected when the server was actively throwing the OOM errors has indicated that OLSmallRyeConfigProviderResolver is failing to release resources when the test applications are stopped.  There is a known memory leak in SmallRye (https://github.com/smallrye/smallrye-config/issues/277) and an attempt has been made to work around it in Open Liberty (https://github.com/OpenLiberty/open-liberty/pull/14550).  The current suspicion is that the GLOBAL_CONFIG_APPLICATION_NAME case is not being handled correctly, so those applications are never released.

tevans78:
Strictly this behaviour isn't quite right but it may be better than the potential memory leak.
fixes https://github.com/OpenLiberty/open-liberty/issues/18426

#build

### DIFF
--- a/dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigWrapper.java
+++ b/dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigWrapper.java
@@ -16,12 +16,19 @@ import java.util.Set;
 
 import org.eclipse.microprofile.config.Config;
 
+import com.ibm.websphere.ras.annotation.Trivial;
+
 /**
  * Wrapper to record which applications are using a Config.
  *
  * This class is not thread-safe and is package scoped only
  */
+@Trivial
 class ConfigWrapper {
+
+    //We try to keep track of which application is using which Config. There are cases where the Config is used by a global component
+    //or we just can't work out which app it is. Then we fall back to this global name.
+    static final String GLOBAL_CONFIG_APPLICATION_NAME = "!GLOBAL_CONFIG_APPLICATION_NAME!";
 
     private final Config config;
     private final Set<String> applications = new HashSet<>();
@@ -34,22 +41,29 @@ class ConfigWrapper {
         this.applications.add(appName);
     }
 
+    /**
+     * Remove an application from this ConfigWrapper. If only the global app is left, remove that too.
+     *
+     * @param appName The application to remove
+     * @return true if there are no applications left in this ConfigWrapper
+     */
     boolean removeApplication(String appName) {
-        boolean close = false;
-        boolean removed = this.applications.remove(appName);
-        if (removed && (this.applications.size() == 0)) {
-            close = true;
+        boolean empty = false;
+        this.applications.remove(appName);
+        //if all applications except for the global one have been removed, remove that too
+        if (this.applications.size() == 1) {
+            this.applications.remove(GLOBAL_CONFIG_APPLICATION_NAME);
         }
-        return close;
+        if (this.applications.size() == 0) {
+            empty = true;
+        }
+        return empty;
     }
 
     Set<String> listApplications() {
         return Collections.unmodifiableSet(applications);
     }
 
-    /**
-     * @return
-     */
     Config getConfig() {
         return this.config;
     }
--- a/dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
+++ b/dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
@@ -10,7 +10,10 @@
  *******************************************************************************/
 package io.openliberty.microprofile.config.internal.extension;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintWriter;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -23,6 +26,9 @@ import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.ConfigurationPolicy;
 
 import com.ibm.websphere.csi.J2EEName;
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.websphere.ras.annotation.Trivial;
 import com.ibm.ws.container.service.app.deploy.ApplicationInfo;
 import com.ibm.ws.container.service.app.deploy.extended.ExtendedApplicationInfo;
 import com.ibm.ws.container.service.state.ApplicationStateListener;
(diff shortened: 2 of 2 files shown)


dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigWrapper.java
```diff
@@ -37,8 +41,14 @@ void addApplication(String appName) {
...
-        if (removed && (this.applications.size() == 0)) {
-            close = true;
+        if (removed) {
+            //if all applications except for the global one have been removed, remove that too
+            if (this.applications.size() == 1) {
+                this.applications.remove(GLOBAL_CONFIG_APPLICATION_NAME);
+            }
+            if (this.applications.size() == 0) {
+                close = true;
+            }
```
Azquelt: I'm unsure of this logic. It would be quite possible for a liberty feature to request a config and to still need it even if all applications are removed.

Then again, is there anything that would stop them using the config after this point if we close it? Usually using a resource after it's been closed causes an exception, but I'm not sure if that's the case here.

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
```diff
@@ -157,7 +153,7 @@ private static String getApplicationName() {
         if (applicationName == null) {
             //There are cases where the Config is used by a global component or we just can't work out which app it is. Then we fall back to this global name.
             //Configs used "globally" will be shutdown when the server is shutdown
-            applicationName = GLOBAL_CONFIG_APPLICATION_NAME;
+            applicationName = ConfigWrapper.GLOBAL_CONFIG_APPLICATION_NAME;
         }
```
Azquelt: Can we add some trace here to identify when and why this happens?

I guess it's a bit tricky, on the one had we want to dump the stack to see who called us, on the other hand there are valid cases where there's no application and in those cases we don't want the cost of dumping the stack.

Maybe we could just include the thread name (which would show if it was an unmanaged thread) and the TCCL?

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigWrapper.java
Emily-Jiang: what is the useful of it when all apps are closed?

tevans78: Unlike our previous implementations, all this does is to remove the config from caches. It does not actually close them. Therefore, if an internal feature still has a reference then it should still work. If it does not then a new one should be created upon request.

Azquelt: > what is the useful of it when all apps are closed?

There are lots of features which do things even if no apps are deployed. E.g. MP Metrics will still report the JVM and app server metrics, MP OpenAPI will report an empty OpenAPI document. I'm more concerned about whether they'll cause errors rather than if their functionality with no applications is useful.

Azquelt: > Unlike our previous implementations, all this does is to remove the config from caches. It does not actually close them.

It's not clear that that's the case from this bit of code. Possibly renaming the `close` variable and adding some javadoc to the method would help.

tevans78: yeah, fair enough, `close` was a hangover from the previous version

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
tevans78: i'll try to add some debug before we merge this ... will wait for Brian to try out the TCK first though

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigWrapper.java
Emily-Jiang: > > what is the useful of it when all apps are closed?
> 
> There are lots of features which do things even if no apps are deployed. E.g. MP Metrics will still report the JVM and app server metrics, MP OpenAPI will report an empty OpenAPI document. I'm more concerned about whether they'll cause errors rather than if their functionality with no applications is useful.

It closes `config`. I am not sure MP Metrics and MP OpenAPI needs a runtime config laying around in the background, besides MP Metrics has no dependency on config.

tevans78: Technically it does not "close" the config ... the standard Config API does not have a close method. All it does is call `releaseConfig(Config)` which SmallRye implements as simply removing the Config from the cache.
https://github.com/smallrye/smallrye-config/blob/08918436d1a48b411da530be99e3da97df1e5bfe/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java#L124

tevans78:
#build

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
```diff
@@ -136,6 +133,8 @@ public void applicationStopping(ApplicationInfo appInfo) {
             for (Config config : configsToRelease) {
                 releaseConfig(config);
             }
+
+            dumpCache();
```
Emily-Jiang: I'm wondering whether to always have this when trace is enabled. Thoughts?

Emily-Jiang: oh, ignore my comments since this can be quite chatty.

tevans78:
@build

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigProviderResolver.java
tevans78: I have tried to tune this a little ... added some more cache debug trace but also made ConfigWrapper trivial which cuts the trace down quite a bit

tevans78:
Well clearly that should not have been @build ... #build

Emily-Jiang (approved):
LGTM

tevans78:
#build

tevans78:
#build

tevans78:
#build

tevans78:
#run-libby-bot

PR #18463, opened 2021-09-07
prepare Concurrency CDI bucket for EE 10

njr-11:
Unlike some other test buckets that will simply repeat tests with the new EE 10 features, this area will need a lot of new tests specifically written to cover the new ways that the Concurrency spec will interact with CDI in EE 10.  This will allow us to test and experiment with new interceptors.  The existing tests within the bucket are minimal, but are a great starting point for testing and experimenting with the new function, so in this particular case, it makes sense to copy them and start from there.


njr-11:
#build

gjwatts (approved):

PR #18483, opened 2021-09-08
Update to latest Open J9 criu API

tjwatson:


tjwatson:
#build

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -79,25 +79,18 @@ public CheckpointImpl(ComponentContext cc, @Reference WsLocationAdmin locAdmin)
         this.cc = cc;
         this.criu = criu;
         this.locAdmin = locAdmin;
-        this.checkpointFeatures = "features".equals(cc.getBundleContext().getProperty(BootstrapConstants.CHECKPOINT_PROPERTY_NAME));
-        this.checkpointApplications = "applications".equals(cc.getBundleContext().getProperty(BootstrapConstants.CHECKPOINT_PROPERTY_NAME));
+        String phase = cc.getBundleContext().getProperty(BootstrapConstants.CHECKPOINT_PROPERTY_NAME);
+        this.checkpointAt = phase == null ? null : Phase.getPhase(phase);
```
sebratton, line 83: Isn't `phase == null` an error condition? Seems like it'll be silently ignored and server will launch normally. I think the launcher does some verification of the --at-location-internal passed in. Is there some other supported way to arrive at Phase == null  which would ignore an error in phase name (system property?) ?

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -110,19 +103,23 @@ public void failedCompletion(Future<Boolean> future, Throwable t) {
...
 
-    void checkpoint(Phase phase) throws CheckpointFailed {
+    void checkpointOrExitOnFailure(Phase phase) {
+        try {
+            checkpoint(phase);
+        } catch (CheckpointFailedException e) {
+            // TODO log error informing we are exiting
+
+            // TODO is there any type of failure where we would not want to exit?
+            new Thread(() -> System.exit(e.getErrorCode()), "Checkpoint failed, exiting...").start();
```
sebratton, line 125: Can you add a comment as to why you are launching a new Thread? I think you said it avoids a quiesce wait?

sebratton (approved):

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
tjwatson: This is the constructor for the service component, which is an immediate component.  So this gets called even when not doing a `checkpoint` action.  I considered adding a NONE enum value for Phase, but thought that would be a strange external for the SPI.  Instead the phase is null when we should have the various events from "features", "applications" etc. be a no-op.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
tjwatson: I'll add a comment.  The extra thread is needed to avoid blocking the current thread while the shutdown hooks are run (which starts a server shutdown and quiesce).  The issue is that if we block the event thread then there is a deadlock that prevents the server shutdown from happening until a 30 second timeout is reached.

PR #18509, opened 2021-09-10
Add Java 17 FAT test

gjwatts:
Adds a basic Java 17 test to verify we are running with the proper version of Java.  It uses the Java 17 features:
```
JEP 356: Enhanced Pseudo-Random Number Generators - https://openjdk.java.net/jeps/356
JEP 406: Pattern Matching for switch (Preview) - https://openjdk.java.net/jeps/406
```


gjwatts:
#build

dev/com.ibm.ws.java.level_fat/src-reference/java17/org/gjwatts/liberty/TestApp.java
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.gjwatts.liberty;
```
KyleAure: I would maybe re-iterate here as a note that changes to the application code will not be honored, and that the application is built externally and uploaded as a binary.

KyleAure (changes requested):
Can the `java17-app.war` be uploaded to GHE and pulled at build/runtime?  I think that would be a better design than uploading a binary to GitHub.

KyleAure (approved):
Plan is to move this binary to an external repo in a future PR. Because of the small size it shouldn't impact git history much.

dev/com.ibm.ws.java.level_fat/src-reference/java17/org/gjwatts/liberty/TestApp.java
gjwatts: Thanks for the suggestion.  I have added this to the FAT README.txt file.

gjwatts:
#build

gjwatts:
#libbybot

KyleAure (approved):

Issue #16439, opened 2021-03-31
Move WsByteBuffer Implementation and SPI into independent bundle


wtlucy:
Currently the `WsByteBuffer` code (`com.ibm.wsspi.bytebuffer` and `com.ibm.ws.bytebuffer.internal`) lives in the `com.ibm.ws.channelfw` bundle. We should explore moving that code out of the `channelfw` bundle and into a standalone bundle along the lines of `com.ibm.ws.wsbytebuffer` or `io.openliberty.wsbytebuffer`. This will allow other projects which depend on `WsByteBuffer` to use it without pulling in the entire channel framework.

hutchig:
- [x] Unit tests pass on branch hutchig/wsbytebuffer_16439
- [x] FAT tests pass on branch
- [x] CF stripped of internal WsByteBuffer on branch 
- [x] CF UT and FATs OK
- [x] WS-CD-Open tested
- [x] Resolve WS-TraceGroup: ChannelFramework

I have detached the goal of creating a new separate FAT bucket as this buffer bundle will
not be used by any other bundles that the existing set which all have a good coverage
of tests already and there has been no function/interface changes as the Java level, just OSGi 
repackaging.

It would be more cost effecting to put effort into easy-to-run performance tests that could then
be used to compare with the Netty equivalent.

hutchig:
OSGI-INF/metatype/metatype.xml has 'com.ibm.ws.bytebuffer.metatype' rather than 'com.ibm.ws.wsbytebuffer.metatype' which it might be if more consistent?

hutchig:
#build

hutchig:
Done

PR #18522, opened 2021-09-13
Remove duplicate bundle from channelfw feature

wtlucy:
`com.ibm.websphere.appserver.channelfw-1.0.feature` has the bundle `com.ibm.ws.timer` defined twice. This isn't a functional problem, but the duplicate should be cleaned up. The duplicate was added in #17940, for #16439


wtlucy:
#build

hutchig (approved):
LGTM

Issue #17599, opened 2021-06-18
HTTP upgrade to WebSocket can cause quiesce errors


loriadi:
**Describe the bug**  
When a websocket connection is started, it starts as an HTTP connection. If an error occurs during the transition between an HTTP and a websocket connection, which is known to be a websocket upgrade, the error processing neglects to decrement a connection counter, which then causes the server to believe there is an open connection during server stop.

There are two scenarios we are aware of when these quiesce errors occur. As an HTTP upgrade connection request is sent to the server.

1) When a read error occurs during the transition between an HTTP and a websocket connection, the error processing neglects to decrement a connection counter, which then causes the server to believe there is an open connection during server stop.

2) If a client immediately closes the websocket connection after it is opened, the original upgrade request handling may not have had enough time to close properly on the server. Once again, the connection counter fails to decrement leading the server to believe there is an open connection during the server stop.

**Steps to Reproduce**  
This problem is not manually reproduceable and is intermittent.

**Expected behavior**  
The connection counter should be decremented properly when an error occurs during the websocket upgrade process.

**Diagnostic information:**  
 - OpenLiberty Version: all
 - Affected feature(s): websocket-1.0, websocket-1.1

**Additional context**  
None

====================
Summary:

The problem is intermittent.

The triggering factor is when an unexpected read error occurs, like this one:
```
[6/14/21 8:37:03:735 PDT] 0000001e id=207ec47a com.ibm.ws.tcpchannel.internal.SocketIOChannel 1 Read -1(-1) bytes, 1 requested on local: localhost/127.0.0.1:8010 remote: localhost/127.0.0.1:57364
[6/14/21 8:37:03:735 PDT] 0000001e id=00000000 com.ibm.ws.tcpchannel.internal.WorkQueueManager > requestComplete Entry
[6/14/21 8:37:03:735 PDT] 0000001e id=00000000 com.ibm.ws.tcpchannel.internal.WorkQueueManager 3 calling cc.error callback
```

In this case, the wsoc connection has received an initial request with an upgrade header. The 101 response has been sent back. Then a read on a new thread is issued, which immediately returned with a read error.

When this error happens, a CLOSE frame is sent out which is successful, then some clean up work is done.

Then we get to this cleanup where the httpdispatcher link isn't closed, it's just destroyed. This is all being done on the read thread.

```
[...]

PR #18524, opened 2021-09-13
Fix WebSocket Quiesce Error During Concurrent Requests 

volosied:
My attempt at addressing defect 276206. It still needs review from the team, but I think it's promising. 

#17599

Basically, my idea for the fix is to use a countdown latch which is created during the upgrade initialization.   HttpDispatcherLink.Close needs to countdown before HttpDispatcherLink.destroy can be called.  99% of the time HttpDispatcherLink.close is completes first, so there shouldn’t  be any impact for customers.  But in case there are these two concurrent requests (upgrade & close wsoc call), then it should wait at most 30 milliseconds for the close to complete first. 

I'm confident there needs to be a wait in any solution we provide in order to avoid this problem, so it would be best when the connection is closing. 

Part of the problem is that the 101 is send to the client first (via the finishKeepConnection call) before the web-socket initialization is fully complete (via handler.init). 

https://github.com/OpenLiberty/open-liberty/blob/5d81f9cbe0e1b3d1ac09f18d298e61d420d4a359/dev/com.ibm.ws.webcontainer.servlet.3.1/src/com/ibm/ws/webcontainer31/osgi/srt/SRTConnectionContext31.java#L155-L161

But once the client receives the 101, it is free to send out any wsoc request, including close which the server isn’t ready for since it is a bit slow on closing up the upgrade request.   It's better to have a wait on the close frame request, since timing is more important for opening connections. 

Based on the logs I looked at, the timing differences between the close & destroy ranged from 8 to 25 milliseconds.  30 should be fairly safe?   30 should greatly reduce the number of defect occurrences.

### DIFF
--- a/dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/dispatcher/internal/channel/HttpDispatcherLink.java
+++ b/dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/dispatcher/internal/channel/HttpDispatcherLink.java
@@ -18,8 +18,10 @@ import java.nio.charset.Charset;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import com.ibm.websphere.ras.Tr;
 import com.ibm.websphere.ras.TraceComponent;
@@ -249,6 +251,11 @@ public class HttpDispatcherLink extends InboundApplicationLink implements HttpIn
                 this.myChannel.decrementActiveConns();
             }
         }
+
+        if (this.vc != null && this.vc.getStateMap().get(TransportConstants.UPGRADED_CONNECTION) != null) {
+            CountDownLatch countDownLatch = (CountDownLatch) (this.vc.getStateMap().get(TransportConstants.ON_CLOSE_COUNTDOWN_LATCH));
+            countDownLatch.countDown();
+        }
     }
 
     /*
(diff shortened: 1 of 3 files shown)


volosied:
Build: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_PrYOsBSpEey6_dRKrYJmeg&tab=com.ibm.team.build.web.ui.internal.editors.result.test.TestProvider

Note: different branch, but the same code

volosied:
#build  -- ABANDONED

dev/com.ibm.ws.webcontainer.servlet.3.1/src/com/ibm/ws/webcontainer31/osgi/srt/SRTConnectionContext31.java
```diff
@@ -135,6 +137,7 @@ public void finishConnection()
                             // remove the TransportConstants which if added for Upgrade previously
                             vc.getStateMap().put(TransportConstants.CLOSE_UPGRADED_WEBCONNECTION, null);
                             vc.getStateMap().put(TransportConstants.UPGRADED_LISTENER, null);
+                            vc.getStateMap().put(TransportConstants.ON_CLOSE_COUNTDOWN_LATCH, null);
```
volosied, line 140: Not sure if I should add this, but it seemed needed to keep it consistent with the puts on lines 110-112.

volosied:
#build  #spawn.fullfat.buckets=io.openliberty.wsoc.internal_fat,com.ibm.ws.webcontainer.servlet.4.0_fat,com.ibm.ws.webcontainer.servlet.3.1_fat,com.ibm.ws.wsoc_fat

dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/dispatcher/internal/channel/HttpDispatcherLink.java
```diff
@@ -249,6 +251,11 @@ public void close(VirtualConnection conn, Exception e) {
                 this.myChannel.decrementActiveConns();
             }
         }
+
+        if (this.vc != null && this.vc.getStateMap().get(TransportConstants.UPGRADED_CONNECTION) != null) {
+            CountDownLatch countDownLatch = (CountDownLatch) (this.vc.getStateMap().get(TransportConstants.ON_CLOSE_COUNTDOWN_LATCH));
+            countDownLatch.countDown();
```
volosied, line 257: Should i have a countDownLatch != null check just in case for here and just before the await?

wtlucy: They're not really necessary, but null checks would mirror how the statemap is typically used

loriadi (approved):
Crossing my fingers that this will work!

volosied:
#build

volosied:
Closing -- #18670 is more favored

PR #18542, opened 2021-09-14
Add Deployment phase for checkpoint

tjwatson:
This PR also adds another beta-guard for the `checkpoint` action.

### DIFF
--- a/dev/com.ibm.websphere.appserver.features/visibility/public/checkpoint-1.0/io.openliberty.checkpoint-1.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/public/checkpoint-1.0/io.openliberty.checkpoint-1.0.feature
@@ -4,6 +4,7 @@ visibility=public
 singleton=true
 IBM-ShortName: checkpoint-1.0
 IBM-Process-Types: server
+IBM-API-Package: io.openliberty.checkpoint.internal.criu; type="internal"
 Subsystem-Name: Checkpoint and Restore Support 1.0
 -bundles=io.openliberty.checkpoint,\
  io.openliberty.jigawatts,\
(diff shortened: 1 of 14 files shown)


tjwatson:
#build

sebratton (commented):
I'm unclear on how classes get into the shared cache. Is it possible for an already transformed class to get inserted and then transformed again?

sebratton (commented):
Do we need an additional beta guard in checkpointImpl to guard against an alternate way of setting the checkpoint property?

sebratton (approved):

tjwatson:
> I'm unclear on how classes get into the shared cache. Is it possible for an already transformed class to get inserted and then transformed again?

Transformed classes do not get added to the shared classes cache.  That is by design.  See method `com.ibm.ws.classloading.internal.AppClassLoader.definePackageAndClass(String, String, ByteResourceInformation, byte[])`
It has this block:
```java
            if (sharedClassCacheURL != null && Arrays.equals(bytes, byteResourceInformation.getBytes())) {
                hook.storeClass(sharedClassCacheURL, clazz);
```
This is where it checks that the defined bytes equals the original resource bytes.  Only if the bytes were not transformed at all does it store the class in the shared class cache.

The difference with the new type of transformer I added here is that it is allowed to transform classes that already exist in the shared classes cache.  But even when I do this transformation the transformed class still does not replace the original class in the cache.

This is also by design because I don't want the transformed class to affect other server configurations in the same installation.  But I must transform already stored classes because 

1. I want the server to be able to be primed with a normal run which may have already stored the application classes
2. I need to transform all application classes that get loaded during a checkpoint so I can detect when they `<clinit>`
3. I don't want to store the class because the server may be started with clean in which case I want the untransformed class to be loaded again out of the cache.

> Do we need an additional beta guard in checkpointImpl to guard against an alternate way of setting the checkpoint property?

I figured since the checkpoint feature will be beta itself there is no need to beta-guard code contained in it.  The beta-guard should only be needed for beta code that exists in a bundle that itself is a GA bundle.

But on that point,we probably do need the beta-guard placed around the calls to the new type of "system" transformers.  I'll add that.

PR #18554, opened 2021-09-15
[18552] Return 400 when client sends invalid content type or accept hdr

### LINKED ISSUE #18552: JAX-RS 2.0 and 2.1 implementation is executing resource method when Content-Type or Accept header contains invalid values
**Describe the bug**
Resource method might look like this:
```
@Consumes("application/json")
@Produces("application/json")
@Path("somePath")
public class MyResource {
    @Context
    HttpHeaders headers;

    @POST
    public String post(String jsonString) {
        assert headers.getHeaderString("Accept").startsWith("application/json") && headers.getHeaderString("Content-Type").startsWith("application/json");
    }
}
```

The problem is that the `post` method is executed, and the assertion fails when a client sends a request that includes headers such as `Accept: blah`, `Content-Type: hello`, etc. where there is no slash in the header value (note that the value of `*` is allowed as the `Accept` header).

If there is a stack trace, please include the FULL stack trace (without any `[internal classes]` lines in it). To find the full stack trace, you may need to check in `$WLP_OUTPUT_DIR/messages.log`

N/A 

**Steps to Reproduce**

Build a JAX-RS App using the resource code above then invoke it using invalid media types - e.g.:
`curl http://localhost:9080/contextRoot/app/somePath -d "foo" -H "Accept: ANYTHING"`
or:
`curl http://localhost:9080/contextRoot/app/somePath -d "foo" -H "Content-type: blah"`

This will invoke the `post` method.

**Expected behavior**
The `post` method should not be invoked, and the JAX-RS implementation should return a 400 (Bad Request).

**Diagnostic information:**
 - OpenLiberty Version: 21.0.0.9
 - Java Version: 11 (though should happen on other levels too)
 - server.xml configuration: `jaxrs-2.1` feature (`jaxrs-2.0` has the same behavior)
 - If it would be useful, upload the messages.log file found in `$WLP_OUTPUT_DIR/messages.log` - N/A

**Additional context**
Originally reported over slack by @MarkBluemel

andymc12:
Fixes #18552


andymc12:
#build

jim-krueger (approved):

PR #18566, opened 2021-09-15
[18565] Handle unmapped applications

### LINKED ISSUE #18565: JAX-RS 3.0 App endpoints never register when unmapped application is found in WAR classpath
@rumanaHaque discovered an issue with her SVT app.  The app was an EAR file that contained multiple WARs. None of them had an `Application` subclass but specified a `<servlet-name>jakarta.ws.rs.core.Application</servlet-name>`.  However, the EAR had a library that contained an `Application` subclass, so the `RESTfulServletContainerInitializer` was incorrectly attempting to map that application subclass to the servlet mapping class.  The result is that the `<servlet-class>` was never added to the servlet registration and all JAX-RS endpoints in the WARs were ignored.

andymc12:
Also refactor FAT bucket to be flatter - and clean up compiler warnings

Fixes #18565


andymc12:
#build

WhiteCat22 (approved):

PR #18567, opened 2021-09-15
Set the starter xml file for DynamicUpdateTest

kristip17:
In the FAT failure, we're not flagging that the server.xml should be updated when it should be. Set it in the `setup` method -- it seems to be failing on the EE9 repeat at the start of the test so maybe we're running with an old config reference.

RTC 286331


kristip17:
#build

jvanhill (approved):

PR #18580, opened 2021-09-16
Allow SSL failures possibly resolved by retries

c00crane:
The server records error message CWWKO0801E, but does sometimes retry and the connection will work.  The FAT framework searches for extra error or warning messages on server shutdown.  It finds this message and fails the test even though everything worked just fine.
I'm adding this error message to the allowed errors for all servers.  If there is truly a connection failure, the test case will NOT get the response that it expects and the test will fail.
This should not mask any real issues.


c00crane:
#build

c00crane:
#run-libby-bot

ayoho (approved):

PR #18587, opened 2021-09-16
allow even more time in calc of exp time (for slow machines)

c00crane:
When validating token content, we can't validate the exact value that will be in the exp attribute - we have to say that it's in a window of time around the time that the test request the token to be committed. We were allowing 10 seconds, and then 20, but, on slow machines that still isn't always a large enough window. Increasing the time to 60 seconds.  Our typical exp times are 2 hours, so, 1 minute off is not the end of the world.


c00crane:
#build

ayoho (approved):

PR #18324, opened 2021-08-25
Java 16/17 add --add-exports for JMX FAT

gjwatts:
One of the tests in the JMX FAT test bucket currently has a dependency on `com.sun.tools.attach.*`.  This causes an exception stacktrace like the example below since it violates the Java strong encapsulation of internal jdk classes in [JEP 403](https://openjdk.java.net/jeps/403):
```
java.lang.reflect.InaccessibleObjectException: 2021-08-10-03:46:30:804 Unable to make public java.util.Properties sun.tools.attach.HotSpotVirtualMachine.getSystemProperties() throws java.io.IOException accessible: module jdk.attach does not "exports sun.tools.attach" to unnamed module @626b2d4a
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
	at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
	at java.base/java.lang.reflect.Method.setAccessible(Method.java:193)
	at com.ibm.ws.jmx.fat.attach.ForwardingInvocationHandler.invoke(ForwardingInvocationHandler.java:35)
	at jdk.proxy2/jdk.proxy2.$Proxy15.getSystemProperties(Unknown Source)
	at com.ibm.ws.jmx.fat.AttachSupport.<init>(AttachSupport.java:49)
	at com.ibm.ws.jmx.fat.JMXTest.testMBeanAttachAPI(JMXTest.java:96)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at componenttest.custom.junit.runner.FATRunner$1.evaluate(FATRunner.java:197)
	at componenttest.custom.junit.runner.FATRunner$2.evaluate(FATRunner.java:318)
	at componenttest.custom.junit.runner.FATRunner.run(FATRunner.java:171)
```

[...]

PR #18601, opened 2021-09-16
Add strong encapsulation exception for JMX FAT with Semeru

gjwatts:
This is the same issue as described in fix #18324, except this for the Semeru JVM.  It needs a slightly different strong encapsulation exception `--add-exports jdk.attach/com.ibm.tools.attach.attacher=ALL-UNNAMED`.


gjwatts:
#build

jhanders34 (approved):

gjwatts:
The Java 17 Semeru PB tests all passed, but the build got hung up on a network problem trying to talk to a backend server.

PR #18613, opened 2021-09-17
add ssl trace to debug keystore issue with Java 8

c00crane:
With Java 1.8.0_301, we're getting the failure: 
The keystore [rsa_trust] is not present in the configuration.

The server.xml contains an entry for rsa_trust and the file that it references rsa_trust.jks does exist.  We can use keytool to open it with the password specified in the config.
We don't see any errors occurring during keystore init to explain why the keystore is not initialized and therefore can't be found.  We're hoping the extra trace will give us a clue.


c00crane:
#build

ayoho (approved):

c00crane:
#run-libby-bot

PR #18617, opened 2021-09-18
Fix spnego_fat.1 on jdk8 build break 

Zech-Hein:
Fixes build break 285663

Problem: Jar used for the apacheKDCforSpnego class for the spnego_fat causes the spnego_fat.1 to fail on java8 because of an unsigned entries exception.

Solution: Move the jar and the classes that use it from the spnego.fat.common project to the spnego_fat project.


Zech-Hein:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.spnego_fat,com.ibm.ws.security.spnego_fat.1

utle (approved):

Zech-Hein:
LibbyBot went down briefly yesterday so the build status didn't get posted to the PR. The build completed successfully.

PR #18652, opened 2021-09-21
Avoid EJB timer cancel while timer running

tkburroughs:
For EJB timer missed timer action tests, have the timer
self cancel to avoid rollbacks if timer concurrently
canceled while the timer is running.


tkburroughs:
#build

brideck (approved):

PR #18682, opened 2021-09-23
combine MP ThreadContext impl with ContextService

### LINKED ISSUE #18667: Concurrency 3.0: ThreadContext methods from MicroProfile
User story:
"As a user of Liberty, I would like to use the MicroProfile ThreadContext methods that have been brought over to ContextService so that I can get individual completion stage actions to run with the desired thread context independently whether other completion stages in the pipeline have context propagation at all."

njr-11:
In preparation for Jakarta EE 10, which will be adding the MicroProfile ThreadContext methods to ContextService, combine our implementations of these classes, which already have a lot of overlap. The MicroProfile thread context providers still don't have the ability to serialize thread context, so the particular Jakarta/Java EE methods that require serializable context will continue to be prohibited when the MicroProfile builders are used, which we will do by raising UnsupportedOperationException in that case. The only way a user can even reach that code path is by casting their MicroProfile ThreadContext to an EE ContextService, which is not part of the MicroProfile programming model, so it is acceptable to continue to have this limitation in place.


njr-11:
#build

mswatosh (approved):

PR #18718, opened 2021-09-24
replace getContextService with captureThreadContext (part 1)

njr-11:
Part of the removal of WSManagedExecutorService.getContextService, so that we can replace this method signature with the new method signature with the same name that is being added in Concurrency 3.0.  This is a 3 step process, which involves providing an alternative way to capture thread context and mostly removing getContextService under part 1, then updating the ManagedExecutorExtensions to switch over to the new interface under part 2, then complete removal under part 3.
Also, update the mock Async signature to match the spec proposal.


njr-11:
#build

njr-11:
#build
#run-libby-bot

KyleAure (approved):

njr-11:
#build

njr-11:
#build

PR #18750, opened 2021-09-27
Use osgiresourcelocator to load xmlBinding impl

Azquelt:
Get the XML Binding implementation by using the glassfish OSGi resource
locator to get it if we couldn't service load it using the TCCL.

This ensures we can get the XML Binding implementation outside of the
context of an application (e.g. when batch uses it to deserialize
messages received over JMS).

### DIFF
--- a/dev/cnf/oss_dependencies.maven
+++ b/dev/cnf/oss_dependencies.maven
@@ -436,6 +436,7 @@ org.glassfish.hk2:hk2-locator:2.5.0-b42
 org.glassfish.hk2:hk2-utils:2.3.0-b10
 org.glassfish.hk2:hk2-utils:2.5.0-b42
 org.glassfish.hk2:osgi-resource-locator:1.0.1
+org.glassfish.hk2:osgi-resource-locator:1.0.3
 org.glassfish.jaxb:codemodel:3.0.1
 org.glassfish.jaxb:jaxb-core:3.0.1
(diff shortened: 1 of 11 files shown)


Azquelt:
#build

neuwerk (approved):
Changes look good to me! Thanks so much for doing this.

Azquelt:
#run-libby-bot

Azquelt:
The one failure in the build matches a known intermittent problem.

dev/cnf/oss_dependencies.maven
```diff
@@ -436,6 +436,7 @@ org.glassfish.hk2:hk2-locator:2.5.0-b42
 org.glassfish.hk2:hk2-utils:2.3.0-b10
 org.glassfish.hk2:hk2-utils:2.5.0-b42
 org.glassfish.hk2:osgi-resource-locator:1.0.1
+org.glassfish.hk2:osgi-resource-locator:1.0.3
```
tevans78, line 439: Please remove the old version

Azquelt: The old version is still used in a separate test project where they use a lot of the hk2 jars. I'm _assuming_ they're using a consistent set of jars corresponding to a particular release so I didn't touch them.

tevans78: we want to keep to a single major version wherever possible

PR #18773, opened 2021-09-28
286891 Fix io.openliberty.java.internal_fat to pass at least 1 test

gjwatts:
This fixes RTC defect 286891.  FAT tests need to pass at least one test.  The main test can only run at Java 17, so when this test is run at Java 11, it fails to run any tests and shows up as a failing test.  Added the `AlwaysPassesTest.class` to fix.

### DIFF
--- a/dev/io.openliberty.java.internal_fat/bnd.bnd
+++ b/dev/io.openliberty.java.internal_fat/bnd.bnd
@@ -18,7 +18,7 @@ fat.project: true
 
 # To define a global minimum java level for the FAT, use the following property.
 # If unspecified, the default value is ${javac.source}
-fat.minimum.java.level: 17
+#fat.minimum.java.level: 17
 
 # Dependencies may be local bundles (e.g. com.ibm.websphere.javaee.servlet.3.1)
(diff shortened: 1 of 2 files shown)


dev/io.openliberty.java.internal_fat/bnd.bnd
```diff
@@ -18,7 +18,7 @@ fat.project: true
 
 # To define a global minimum java level for the FAT, use the following property.
 # If unspecified, the default value is ${javac.source}
-fat.minimum.java.level: 17
+#fat.minimum.java.level: 17
```
gjwatts, line 21: Specifying this value breaks this test when you run it locally since the logic in launch.xml detects if the `fat.minimum.java.level` specified here is greater than the running JVM.  If so, launch.xml tries to perform these 2 commands:
```
<var name="test.bucket.class" unset="true" />
<var name="test.bucket.name" unset="true" />
```
but the `var` type which is a tool not available to local runs (but is on the test servers).

gjwatts:
#build

KyleAure (approved):

PR #18792, opened 2021-09-29
Fix handling of MP JWT default instance config

ayoho:
The `setGenericAndFilteredConfigTaiRequestInfoFromConfigServices()` method in the `TAIRequestHelper` class has some quirky logic for adding "generic" MP JWT configs to the `MicroProfileJwtTaiRequest` object. In short, if a server admin doesn't configure an `<mpJwt>` element, the default `<mpJwt>` element that's provided by the feature doesn't actually get added as a generic config for the request. That results in the MP JWT feature believing there are no configs present to handle a protected resource request, so Liberty moves on to the next authentication method. The default `<mpJwt>` element should be considered a generic config for this scenario.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.jwt_fat.builder,com.ibm.ws.security.jwt_fat.consumer,com.ibm.ws.security.jwtsso_fat,com.ibm.ws.security.mp.jwt.1.1_fat,com.ibm.ws.security.mp.jwt.1.2_fat

PR #18802, opened 2021-09-29
fix installUtility bug 

jjiwooLim:
When checking for auto features, if the same feature exists in the same directory as one specified in the command, the esa cache that stores the esa file path was getting overwritten (possibly to older version)

Added a debug log that shows the esa file being installed


jjiwooLim:
#build

jjiwooLim:
#libby-bot

jjiwooLim:
#build

bensonlatibm (approved):
lgtm!

jjiwooLim:
#libby-bot

PR #18807, opened 2021-09-30
checkpoint fat

sebratton:


sebratton:
#build

sebratton:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -450,10 +452,76 @@ public String getPathToAutoFVTNamedServer() {
...
+     * <code> "bin/server start"</code>
+     * </pre>
+     *
+     * @param phase
+     * @param autoRestor    if true initiate restore as part of serverStart
+     * @param beforeRestore beforeRestore lambda is called just before the server start
      */
-    private Phase checkpointPhase = null;
+    public void setCheckpoint(Phase phase, boolean autoRestore, Consumer<LibertyServer> beforeRestoreLambda) {
+        if (phase == null) {
```
tjwatson: Why check for null here but not in the other setCheckpoint method above?  Why doesn't the simple `setCheckpoint(Phase phase)` method just call this one with `true` and a `null` consumer?

tjwatson (approved):
Just a small comment on the constructor.  Overall it seems like the right direction.

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
sebratton: Thanks. I moved the null check to constructor and redirected one parameter setCheckpoint method to call three parameter version.

sebratton:
#build

sebratton:
#build

sebratton:
#build

PR #18814, opened 2021-09-30
NullPointerException from transactions while computing message for different error

### LINKED ISSUE #18813: Test Failure: testJTATransactionUsedSeriallyWithOverlapAndCommitWithinLastStage NullPointerException
**Describe the bug**  
An application uses a ManagedExecutor to attempt to propagate Transaction context of an active UserTransaction to an asynchronous completion stage that overlaps the UserTransaction.  Normally, the completion stage would raise IllegalStateException to reject the condition where it exists on 2 threads at the same time.  However, if the application code suspends or resolves the UserTransaction while the asynchronous completion stage is attempting to start, the transaction code that ought to raise the IllegalStateException can hit a timing window where NullPointerException occurs.

Here is the code from the automated test case where we saw the failure occur intermittently,

```
        CompletableFuture<Integer> stage;
        tx.begin();
        try {
            stage = txExecutor.supplyAsync(() -> {
                try (Connection con = defaultDataSource.getConnection(); Statement st = con.createStatement()) {
                    return st.executeUpdate("INSERT INTO MNCOUNTIES VALUES ('Freeborn', 30619)");
                } catch (SQLException x) {
                    throw new CompletionException(x);
                }
            }).thenApplyAsync(numUpdates -> {
                try (Connection con = defaultDataSource.getConnection(); Statement st = con.createStatement()) {
                    numUpdates += st.executeUpdate("INSERT INTO MNCOUNTIES VALUES ('Mower', 39386)");
                    tx.commit();
                    return numUpdates;
                } catch (Exception x) {
                    throw new CompletionException(x);
                }
            }).exceptionally(x -> {
                try {
                    tx.rollback();
                } catch (Exception x2) {
                    x2.printStackTrace();
                }
                if (x instanceof RuntimeException)
                    throw (RuntimeException) x;
                if (x instanceof Error)
                    throw (Error) x;
                throw new CompletionException(x);
            });
        } finally {
            tm.suspend();
        }

        try {
            stage.get(TIMEOUT_NS, TimeUnit.NANOSECONDS);
        } catch (ExecutionException x) {
            Throwable cause = x.getCause();
            if (cause instanceof IllegalStateException) // transaction used on 2 threads at once
                return; // successful, got the expected exception
            else
                throw x; // unexpected failure, such as the NullPointerException
        }
``` 

If there is a stack trace, please include the FULL stack trace (without any `[internal classes]` lines in it). To find the full stack trace, you may need to check in `$WLP_OUTPUT_DIR/messages.log`

```
java.util.concurrent.ExecutionException: java.lang.NullPointerException: Cannot invoke "java.lang.Thread.getId()" because the return value of "com.ibm.tx.jta.embeddable.impl.EmbeddableTransactionImpl.getThread()" is null
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2092)
	[internal class]
	at concurrent.mp.fat.tx.web.MPConcurrentTxTestServlet.testJTATransactionUsedSeriallyWithOverlapAndCommitWithinLastStage(MPConcurrentTxTestServlet.java:879)
	at concurrent.mp.fat.tx.web.MPConcurrentTxTestServlet.doGet(MPConcurrentTxTestServlet.java:136)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:686)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:791)
        [internal classes]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:853)
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Thread.getId()" because the return value of "com.ibm.tx.jta.embeddable.impl.EmbeddableTransactionImpl.getThread()" is null
	[internal classes]
```

**Steps to Reproduce**  
Run the scenario within the following automated test case until it eventually hits the timing window and fails,
```
com.ibm.ws.concurrent.mp.fat.MPConcurrentTxTest.testJTATransactionUsedSeriallyWithOverlapAndCommitWithinLastStage_MP_CONTEXT_PROP_10
```

**Expected behavior**  
`CompletableFuture.get` must raise `ExecutionException` with chained `IllegalStateException` indicating that the transaction cannot be propagated to the asynchronous completion stage thread while the same transaction remains active on the main servlet thread.

**Diagnostic information:**  
 - OpenLiberty Version: all versions including the mpContextPropagation-1.0 or 1.2 feature up through 21.0.0.10.
 - Affected feature(s) mpContextPropagation-1.0 and 1.2
 - Java Version: all
 - server.xml configuration
 ```
<feature>mpContextPropagation-1.0</feature>
```
  or
```
<feature>mpContextPropagation-1.0</feature>
```

**Additional context**  
Because the error occurs while reporting an application bug, applications are recommended to fix the application bug by ensuring that the UserTransaction has been suspended from the servlet thread before permitting the start of any dependent completion stages to which the same Transaction context is propagated.

njr-11:
fixes #18813

The NullPointerException is due to a bug in transactions code where it falsely assumes that if getThread() returned a value once then it will return the same non-null value again shortly after.  There is a small timing window where that won't be the case which leads to this NullPointerException.

### DIFF
--- a/dev/com.ibm.ws.tx.embeddable/src/com/ibm/tx/jta/embeddable/impl/EmbeddableTranManagerImpl.java
+++ b/dev/com.ibm.ws.tx.embeddable/src/com/ibm/tx/jta/embeddable/impl/EmbeddableTranManagerImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2020 IBM Corporation and others.
+ * Copyright (c) 2009, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -119,8 +119,9 @@ public class EmbeddableTranManagerImpl extends TranManagerImpl {
 
             if (!t.isResumable()) {
                 final IllegalStateException ise;
-                if (t.getThread() != null) {
-                    ise = new IllegalStateException("Transaction already active on thread " + String.format("%08X", t.getThread().getId()));
+                Thread thread = t.getThread(); // avoid race condition where value becomes null after first check
+                if (thread != null) {
+                    ise = new IllegalStateException("Transaction already active on thread " + String.format("%08X", thread.getId()));
                 } else {
                     ise = new IllegalStateException("Transaction cannot be resumed on this thread");
                 }


njr-11:
#build

jonhawkes (approved):

PR #18879, opened 2021-10-05
Swap WIM Reg FATs to use Inbound ID LDAP where possible

kristip17:
Only two tests in the wim.registry FAT bucket were still using remote LDAP. `EnumerateUsersTest` requires remote LDAP (need a long enough time for the LDAP calls to show a difference in timing), but `DefaultWIMRealmMultipleReposTest` does not.

I updated DefaultWIMRealmMultipleReposTest to use all InboundID LDAP.

I set this test to only allow remote LDAPs and EnumerateUsersTest should only run if remote is available. We don't want to start local ApacheDS if no tests are going to use it.

RTC 286750


kristip17:
#build

jvanhill (approved):

kristip17:
#libby

PR #18883, opened 2021-10-05
objectClass checking in userFilter should be case insensitive

ryanesch:


ryanesch:
#build

ryanesch:
Handled in another PR

PR #18884, opened 2021-10-05
Pass the class name instead of the resource name to transformers

tjwatson:
The refactoring of the transformer code broke by passing the resource name to the transformers.


tjwatson:
#build

sebratton (approved):

PR #18906, opened 2021-10-06
Improve zOS runs when remote LDAP is down

kristip17:
If we're running zOS, the local ApacheDS will not work. Try to setup as many remote LDAPs are available instead of bailing at the first failure. This builds on my previous fix to force zOS runs to stay remote, but resulted in NPEs setting the remote server names because we stop pulling consul servers when we hit an error.

RTC 287143

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
@@ -31,9 +31,11 @@ import javax.naming.NamingException;
 import javax.naming.directory.DirContext;
 import javax.naming.directory.InitialDirContext;
 
+import com.ibm.websphere.simplicity.OperatingSystem;
 import com.ibm.websphere.simplicity.RemoteFile;
 import com.ibm.websphere.simplicity.log.Log;
 
+import componenttest.common.apiservices.LocalMachine;
 import componenttest.topology.impl.LibertyServer;
 
 /**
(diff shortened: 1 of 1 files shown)


kristip17:
#build

dev/fattest.simplicity/src/componenttest/topology/utils/LDAPUtils.java
```diff
@@ -137,25 +137,29 @@
...
     static {
 
         /*
          * Determine whether we are running remote or locally.
          */
         boolean uselocalLDAP = isLocalLdapExpectedToBeUsed();
 
+        String os = System.getProperty("os.name").toLowerCase();
+        isZOS = os.startsWith("z/os");
+
```
Azquelt: In other FAT tests, I can see we're doing `LocalMachine.getInstance().getOperatingSystem()` to do this sort of check which might be more reliable.

kristip17: I'll switch it.

kristip17:
Build failure was 1 ldap_fat read time out failure which is a known intermittent issue. I also ran a z/OS build with ldap_fat and it passed successfully -- https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_O9YnoCguEeyWeMDU9jagxQ

jvanhill (approved):

kristip17:
#libby

PR #18122, opened 2021-08-10
Initial ConnectionBuilder path for Oracle Kerberos

mswatosh:
Adds new property to allow for using the OracleConnectionBuilder to create Connections with Kerberos.
fixes #18121

Issue #18121, opened 2021-08-10
Add support for Connection Builder path for Oracle Kerberos connections


mswatosh:
In some cases, there is an issue between the Oracle driver and IBM Java SDK when running under the DataSource.getConnection() path. This can be mitigated by running under the DataSource.getConnectionBuilder() path.

PR #18920, opened 2021-10-07
Reverting Oracle Connection Builder path

mswatosh:
Reverts #18122
Removes changes for issue #18121

It was determined that the connection builder path does not provide the intended benefit, so it is being removed.


mswatosh:
#build

frowe (approved):

PR #18923, opened 2021-10-07
join and get method exceptions should include stack

njr-11:
When CompletableFuture join/get/getNow operations raise CancellationException or CompletionException which was created earlier, the exception stack doesn't show the join/get/getNow operation from the application code that triggered raising the error in the first place. This is extremely misleading and confusing to anyone who is trying to debug application code and determine what/where is causing the exception to be raised. This behavior is inherited from java.util.CompletableFuture, but we should be able to correct it by replacing the top level exception class within the join/get/getNow operation of our subclass.


KyleAure (approved):

njr-11:
#build

PR #18954, opened 2021-10-11
Enable FT 4.0 - Metric 4.0 integration

tevans78:


tevans78:
#build

tevans78:
#build

tevans78:
#build

dev/com.ibm.ws.microprofile.faulttolerance.cdi_fat_metrics/fat/src/com/ibm/websphere/microprofile/faulttolerance/metrics/fat/tests/MetricRemovalTest.java
```diff
@@ -125,10 +118,8 @@ private String getMetricsPage() throws IOException {
      * undeploying while the server is running.
      */
     private void deployApp(Archive<?> archive) throws Exception {
-        ShrinkHelper.exportArtifact(archive, ".");
-        System.out.println("I'm putting the archive here: " + new File(".").getAbsolutePath());
         server.setMarkToEndOfLog();
-        server.copyFileToLibertyServerRoot(".", "dropins", archive.getName());
+        ShrinkHelper.exportDropinAppToServer(server, archive, DeployOptions.SERVER_ONLY, DeployOptions.DISABLE_VALIDATION);
         assertNotNull(archive.getName() + " started message not found", server.waitForStringInLog("CWWKZ000[13]I.*" + getAppName(archive)));
```
Azquelt, line 123: Not sure why we need to disable validation and then manually check for the started message.

dev/fattest.simplicity/src/componenttest/rules/repeater/JakartaEE9Action.java
```diff
@@ -363,16 +363,29 @@ public static void transformApp(Path appPath, Path newAppPath) {
...
                 }
-            } catch (IOException e) {
+            } else {
                 Log.info(c, m, "Unable to create backup directory.");
-                Log.error(c, m, e);
-                throw new RuntimeException(e);
+                FileNotFoundException fnfe = new FileNotFoundException("Parent path not found: " + appPath.toAbsolutePath().toString());
+                Log.error(c, m, fnfe);
             }
         } else {
```
Azquelt, line 390: I'm not sure of the logic of this change.

When either `getParent()` returns `null` (which is unlikely, unless you're right at the top of the filesystem), an error is logged but you continue.

If creating the backup directory fails (which is possible if you try to put it somewhere you don't have permission), an exception is thrown.

dev/com.ibm.ws.microprofile.faulttolerance.cdi_fat_metrics/fat/src/com/ibm/websphere/microprofile/faulttolerance/metrics/fat/tests/MetricRemovalTest.java
tevans78: because the validation flag is persisted and would happen again on the next server start (when the app isn't there)

dev/fattest.simplicity/src/componenttest/rules/repeater/JakartaEE9Action.java
tevans78: When the Path used was a relative (e.g. "."), getParent() was returning null. converting to an absolute path first resolves that. A permissions problem should show up differently.

Azquelt (approved):

tevans78:
#run-libby-bot

PR #18974, opened 2021-10-13
use PrintStream to avoid risk of out of memory errors

benjamin-confino:
An OOM was seen from from calling Log on detector.listProcesses() so lets use a print stream set to flush after every new line.


benjamin-confino:
#build

tevans78 (approved):

Issue #18973, opened 2021-10-13
Investigate weld-osgi-bundle versions in feature files


benjamin-confino:
Our feature files have outdated versions of the weld osgi bundle. We need to investigate if this is causing problems, if it can be updated without issue, and if there's a way to link things so this automatically updates to the latest version.

benjamin-confino:
I'm closing this so it will appear in the release fix lists. The problem is resolved but currently we'll need to manually update these file next time we update weld.

mbroz2:
@benjamin-confino please update the description of this release bug so that it follows/conforms to the [Bug report template](https://github.com/OpenLiberty/open-liberty/issues/new?assignees=&labels=release+bug&template=bug_report.md&title=)

PR #18977, opened 2021-10-13
Update maven coordinates

benjamin-confino:
This will help with #18973 but a more permanent solution would be nice.


benjamin-confino:
#build

tevans78 (approved):

PR #18978, opened 2021-10-13
Reactive Messaging: allow slow container startup

Azquelt:
Increase the startup timeout and number of attempts to make when
starting a Kafka broker container.

This is trying to work around an issue we're seeing regularly but
infrequently in the build where the test fails because the container
does not start and open ports within the timeout.

For RTC 280258


tevans78 (approved):

Azquelt:
#build

PR #18980, opened 2021-10-13
Fix EJB Container API Type in features

tkburroughs:
The EJB Container API package com.ibm.websphere.ejbcontainer is incorrectly
set to a type of "internal" rather than "ibm-api" and needs to be corrected.
The API jar is already present in wlp/dev/api/ibm and FAT tested.


tkburroughs:
#build

jhanders34 (approved):

PR #19004, opened 2021-10-14
concurrency 3.0 initial development snapshot build

njr-11:
Update to a snapshot build 3.0.0.20211013 of Jakarta Concurrency, which was created locally from my dev environment based on the master branch.  This will get the resource definition annotations (such as ManagedExecutorDefinition) in place so that we can try to figure out how to process those.  The changes that I brought in don't have the asynchronous methods annotation yet, so code simulating that will not yet be replaced.  The concurrent_fat_cdi bucket is updated under this pull, but the concurrent_fat_jakarta bucket will mostly need to wait in order to avoid merge conflicts with another pull.


njr-11:
#build

KyleAure (approved):

njr-11:
Build has 3 test failures in an unrelated test bucket which are duplicates of a pre-existing issue.

PR #19014, opened 2021-10-15
http2_fat updates

wtlucy:
* slim down http2_fat default trace spec - currently the trace file is unbounded and can get very large
* correct the exception processing in `MultiSessionTests`


wtlucy:
#build

loriadi (approved):
Looks good, Bill.

PR #19050, opened 2021-10-20
remove instramentation from packages with no source because it can ca…

benjamin-confino:
…use problems when code is back ported for an IFIX


benjamin-confino:
#build

Azquelt (approved):

Azquelt:
Looks sensible to me. There's no source in either project so there's nothing to instrument unless we're overlaying a fix to a class in service in which case we don't want to instrument anyway.

benjamin-confino:
#build

PR #19058, opened 2021-10-20
GrpcMetricsTest is not correct for sequential runs 

wtlucy:
* reset `GrpcMetricsTest` counter between class invocations
* increase the app update timeout length for `grpc_fat` - I get occasional server update timeouts on my 2017 MBP. The default _local run_ timeout is 12s; I'll set it to the default value for _non-local runs_


wtlucy:
#build

volosied (approved):
Both look good

PR #19063, opened 2021-10-20
Better handling of connection error events on free connections

jmstephensgit:
Resource adapter provider may attempt to use connection error event for a connection that is not being used by an application.   The code changes will better tolerate a connection error event for a free connection and avoid potential problems if the connection error event is called multiple times for the same connection.


jmstephensgit:
#build

njr-11 (approved):

PR #19077, opened 2021-10-21
getContextService with multiple managed executors

njr-11:
Multiple managed executors can be configured to share a single contextService.
When this happens, the result of the ManagedExecutorService.getContextService method needs to be aware of which managed executor it came from so that managed executor can back completion stages that are created by ContextService.withContextCapture


njr-11:
#build

jmstephensgit (approved):

PR #19092, opened 2021-10-21
LastExecution methods with ZoneId

njr-11:
Add the LastExecution methods that specify a ZoneId.
Write a test case that covers these as well as confirming that the default impl for ZonedTrigger methods produce correct behavior when the trigger only implements the new getNextRunTime/getZoneId methods.


njr-11:
#build

jmstephensgit (approved):

PR #19111, opened 2021-10-22
Fixing testCacheInvalidationAfterTimeout

prova-anika-216474306:
Catching exceptions and logging timeout exceptions


prova-anika-216474306:
#build

zhang-ca:
If this to fix the **java.lang.RuntimeException**: Internal Server Error due to server die,
try catch (java.lang.RuntimeException e) instead of just catch (Exception e)

Issue #18581, opened 2021-09-16
Iterate on LibertyResteasyEndpointIntrospector to work with more than dropin applications


WhiteCat22:
Currently, when deploying an application via `server.xml`, there are 2 unresolved issues with how we assemble the JAX-RS endpoints to log.

1. How do we get the application name when it's different from the Context-Root? Is organizing the endpoints via Context-Root good enough?

2. When deploying an application via `<webApplication/>` in `server.xml` (with a custom context-root), the endpoint's `fullpath` as built by Resteasy is missing the `@ApplicationPath`'s string.

For example:
`/com.ibm.ws.jaxrs.2.1.client_fat/fat/src/com/ibm/ws/jaxrs21/client/fat/test/JAXRS21ReactiveSampleTest.java`

server.xml:
```
<server>
    ...
    <webApplication location="${server.config.dir}/apps/jaxrs21reactivesample.war" contextRoot="/"/>
    ...
</server>
```

```
@ApplicationPath("reactive")
public class ReactiveTestApp extends Application {
```

```
@Path ("/test")
public class ReactiveTestResource {
```

expected endpoint: `/reactive/test`
actual endpoint: `/test`

As you can see, "reactive" is missing for some reason.

PR #19117, opened 2021-10-24
18581 - iterate on the RESTfulWS-3.0 endpoint introspector

WhiteCat22:
#18581 

I improved the dump introspector to print out the servlet mapping (which was previously missing).


WhiteCat22:
#build

I do have 1 more thing I'd like to improve. Ideally, I'd like to find a way to get the `ServletConfig` from a JAX-RS Dynamic feature so I can port this introspector back to `jaxrs-2.0` and `jaxrs-2.1`.

Barring that, there may be a better spot to call `logEndpoints()` from than `ServletContainerDispatcher`.

I mainly picked `ServletContainerDispatcher` because we already override it in OpenLiberty, rather than pull in another overlay class.

WhiteCat22:
#build

WhiteCat22:
#build #run-libby-bot

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,239 @@
...
+            }
+            StringBuffer sb = new StringBuffer();
+            sb.append("Endpoints for Application: ").append(appname).append(" - ").append(servletName).append("\n");
+            
+            if (resources != null) {
+                for (String clazz : resources) {
+                    Class resource;
+                    try {
+                        // load the class
+                        resource = Thread.currentThread().getContextClassLoader().loadClass(clazz);
```
andymc12: This (getting a classloader) needs to be in a doPriv.  The loadClass might also... I'd suggest creating a new static method that loads the class from the TCCL and it can do both in a doPriv if a security manager is present.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,239 @@
...
+                            
+                            // ignore methods without verbs
+                            if (httpMethod != null) {
+                                // check if the method also has a @Path annotation
+                                String methodPath = null;
+                                if (m.isAnnotationPresent(Path.class)) {
+                                    methodPath = m.getAnnotation(Path.class).value();
+                                }
+    
+                                // get the @Consumes MediaTypes
```
andymc12, line 109: This code will only get the `@Consumes`/`@Produces` annotations from the method, but we should also get it from the class, since that would apply to all method that don't have those annotations.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,239 @@
...
+                                        i++;
+                                    }
+                                    sb.append("}");
+                                }
+                                sb.append("\n");
+                            }
+                        }
+                        sb.append("\n");
+                    } catch (ClassNotFoundException e) {
+                        e.printStackTrace();
```
andymc12: Let's not print the stack to the logs.  If auto-FFDC is enabled, that is good.  We could also print it to the introspector output - something like:
```
ERROR WHILE INTROSPECTING - ClassNotFoundException: <SomeClass>
at ...
```

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,239 @@
...
+                    }
+                }
+            }
+
+            introspector.addEndpoints(servletConfig, sb);
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.info(tc, sb.toString());
+            }
+        } else if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+            Tr.info(tc, "unable to find LibertyResteasyEndpointLoggingIntrospector");
```
andymc12: This either needs to be a debug message or needs to be translated.

dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java
```diff
@@ -51,6 +52,7 @@ public ServletContainerDispatcher(final ServletConfig servletConfig)
    {
       this.servletConfig = servletConfig;
       ResteasyContext.pushContext(ServletConfig.class, servletConfig);
+      RestfulWSUtils.logEndpoints(servletConfig); // Liberty Change
```
andymc12, line 55: Just to avoid adding code to overlays, I wonder if this could be put into the `RESTfulServletContainerInitializer` or somewhere else?

andymc12 (changes requested):

dev/io.openliberty.org.jboss.resteasy.common/src/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java
WhiteCat22: I'd like to iterate on that in the future, right now I'm not sure how to get the `ServletConfig` in the `RESTfulServletContainerInitializer` nor in a JAXRS Dynamic Feature.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+                    Class resource = loadClass(clazz, sb);
+                    if (resource != null) {
+
+                        sb.append(resource.getName()).append("\n");
+
+                        // check if the class is annotated with @Path, if so, get the value
+                        String classPath = getAnnotationValue(resource, Path.class, String.class);
+
+                        // get the @Consumes MediaTypes
+                        String classConsumes[] = getAnnotationValue(resource, Consumes.class, String[].class);
```
andymc12:
```suggestion
                        String[] classConsumes = getAnnotationValue(resource, Consumes.class, String[].class);
```

This looks like C style arrays - Java styling prefers the bracket to be near to the type to make clear that the type is an array of Strings.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+
+                        // get the @Produces mediatypes
+                        String classProduces[] = getAnnotationValue(resource, Produces.class, String[].class);
```
andymc12:
```suggestion
                        String[] classProduces = getAnnotationValue(resource, Produces.class, String[].class);
```

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+                            String httpMethod = getHttpMethod(m);
+
+                            // ignore methods without verbs
+                            if (httpMethod != null) {
+
+                                // check if the method also has a @Path annotation
+                                String methodPath = getAnnotationValue(m, Path.class, String.class);
+
+                                // get the @Consumes MediaTypes
+                                String consumes[] = getAnnotationValue(m, Consumes.class, String[].class);
```
andymc12:
```suggestion
                                String[] consumes = getAnnotationValue(m, Consumes.class, String[].class);
```

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+
+                                // get the @Produces mediatypes
+                                String produces[] = getAnnotationValue(m, Produces.class, String[].class);
```
andymc12:
```suggestion
                                String[] produces = getAnnotationValue(m, Produces.class, String[].class);
```

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+                    return (RESTfulEndpointLoggingIntrospector)bc.getService(sr);
+                }
+                } catch (NoClassDefFoundError ncdfe) {
+                    // ignore - return null
+                }
+                return null;
+            }});
+    }
+
+    public static Class<?> loadClass(String clazz, StringBuffer sb) {
```
andymc12: Can we make this private?  Or at least non-public?  Otherwise somebody else could invoke the method to make an unauthorized call to loadClass.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+                    return Thread.currentThread().getContextClassLoader().loadClass(clazz);
+                } catch (ClassNotFoundException e) {
+                    sb.append("ERROR WHILE INTROSPECTING - ").append(e.getMessage());
+                }
+                return null;
+            }
+        });
+    }
+
+    public static Method[] getMethods(Class<?> clazz) {
```
andymc12: Same thing here - non-public.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+            }
+            if (annotation.annotationType().equals(OPTIONS.class)) {
+                httpMethod = "OPTIONS";
+            }
+            if (annotation.annotationType().equals(POST.class)) {
+                httpMethod = "POST";
+            }
+            if (annotation.annotationType().equals(PUT.class)) {
+                httpMethod = "PUT";
+            }
```
andymc12, line 248: (1) Can we add support for PATCH?
(2) In the future, we may also want to log custom HTTP verbs.

dev/io.openliberty.org.jboss.resteasy.common/src/io/openliberty/restfulWS/utils/RestfulWSUtils.java
```diff
@@ -0,0 +1,277 @@
...
+            Method valueMethod;
+            try {
+                valueMethod = annoType.getMethod("value");
+                Annotation anno = element.getAnnotation(annoType);
+                return valueType.cast(valueMethod.invoke(anno));
+            } catch (NoSuchMethodException e) {
+            } catch (SecurityException e) {
+            } catch (IllegalAccessException e) {
+            } catch (IllegalArgumentException e) {
+            } catch (InvocationTargetException e) {}
```
andymc12: Can we consolidate this?  Maybe use a multi-catch (like `catch (NoSuchMethodException | SecurityException | ...)`) or just catch `Exception`?

andymc12 (changes requested):
Getting much closer.  This is looking really good.  Thanks!

andymc12 (approved):
Changes look good - thanks again!

WhiteCat22:
#run-libby-bot #build

Issue #19125, opened 2021-10-25
Avoid use of standard in from TTY when doing checkpoint


tjwatson:
For checkpoint we already avoid using the standard out/err from the TTY.  This was to allow us to more easily switch between foreground and background restores with `server run|start <server name>`.  It also solved some issues with doing a restore in a container without requiring it to be interactive (e.g. by using -it docker option).

Today the server script in the `criu` branch does the following to checkpoint:

```
"${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1
```
Sending standard out/err to the console log.  But this still sends standard in to the java process during checkpoint.  We should add `< /dev/null` to the end of this command to avoid checkpointing standard in from the TTY to the checkpoint process.  There is no need for standard in for a liberty server.

PR #19126, opened 2021-10-25
Do not capture standard in when doing a checkpoint

tjwatson:
There is no need to capture standard in when doing a
checkpoint.  This also helps for restore if we can
avoiding having to connect standard in to a TTY
for some restore scenarios


tjwatson:
#build

tjwatson:
Fixes #19125

sebratton (approved):

PR #19128, opened 2021-10-25
testTLD assertion fix

### LINKED ISSUE #19123: testTLD() assertion error in com.ibm.ws.jsp.2.3_fat
**Describe the bug**  
In FAT com.ibm.ws.jsp.2.3_fat, test testTLD(), there can be an assert error:
testTLD:junit.framework.AssertionFailedError: ... Log should not contain SRVE0253I: Destroy successful.

**Steps to Reproduce**  
The test fails intermittently when using FULL mode.

**Expected behavior**  
The test will be updated to avoid the assertion error.

epj:
Fixes #19123 
The issue is that an assertion trips as false because an SRVE0253I is within the log mark window.  This was caused by the first log mark being set too early.


epj:
#build
#spawn.fullfat.buckets=com.ibm.ws.jsp_fat,com.ibm.ws.jsp.2.3_fat,com.ibm.ws.jsp_fat_lWAS

volosied (approved):

PR #19145, opened 2021-10-26
Add allowed warning to stopserver list for OutboundSSLLDAPTest

kristip17:
Add `CWPKI0815W` as an allowed message in the stopServer checks. `CWPKI0815W: More than one OutboundConnection element specifies the [was-ldap21.fyre.ibm.com] host and [636] port as a filter. The OutboundConnection element configured on the [AlternateSSLSettings] SSL element will be used.` We are intentionally testing multiple SSL configs -- `* Test to make sure the correct SSL outbound filter configuration is used.`

RTC 287495


kristip17:
#build

jvanhill (approved):

PR #19152, opened 2021-10-27
EJB NP Timer test add another stage of CountDownLatches right before testing getNextTimeout.

olendvcook:
To make sure we fired right before calling getNextTimeout


olendvcook:
#build

tkburroughs (approved):

PR #19160, opened 2021-10-27
concurrent-2.0 should not tolerate mpContextPropagation 1.0

njr-11:
concurrent-2.0 is from Jakarta EE 9, which shouldn't interoperate with mpContextPropagation-1.0


njr-11:
#build

jhanders34 (approved):

jhanders34 (changes requested):
Yeah this won't work because of eeCompatible conflicts.  Going to have to change to io.openliberty.org.eclipse.microprofile.contextPropagation-1.3 and update that feature to have the jar file in the lib dir instead of of dev/api until we GA MP 5.0.

njr-11:
#build
#run-libby-bot

njr-11:
not needed because it was included in a different pull

PR #19173, opened 2021-10-28
Add 'zip.reaper.assume.valid' with its default set to true.

tbitonti:
Experimental change:

The zip caching subsystem checks the zip file size and last modified times when a new open is performed on a zip file.

(See: com.ibm.ws.artifact.zip.cache.internal.ZipFileData.reacquireZipFile().)

This change adds a property 'zip.reaper.assume.valid', which when true disables calls to obtain the size and last modified times.

For this PR only, the default value of the new is set to 'true', with the intent to simplify testing the setting.

The motivation for this change is a customer issue where the reacquire size and last modified checks were using about 0.3s each.  This is highly unusual and indicates a problem in the customer's file system.  However, a question lingered: How much time do these checks cost us?

The new property has been added to answer this question.  The goal is to run standard performance tests with this PR and compare those results against a usual performance test (against 'integration', or any other suitable build) to tell if there is any appreciable difference.


tbitonti:
#build

PR #19174, opened 2021-10-28
Fix checking of URL string to allow for a user name

acdemyers:
Fix to make sure the user qualified keyring can be used properly.


acdemyers:
#build

awisniew90 (approved):
LGTM

PR #19188, opened 2021-10-29
Handle CRIU not supported scenarios

sebratton:
Consolidate checks for criu support to a single class and add a check for criu not installed on platform


sebratton:
#build

tjwatson (approved):

PR #19190, opened 2021-10-29
Update security repeat rules

c00crane:
Update the security repeat rules.  Add a method that will allow us to not run some repeats with EE9 on windows.  Add a method that will only run classes on windows.
The two rules together allow us to run a repeat with EE9, but when on windows that repeat really just runs normally without EE9.
The tests are timing out just trying to run the feature replacement - it's soo slow.


c00crane:
#build
#run-libby-bot

c00crane:
replaced with 19283 and 19284

PR #19194, opened 2021-10-29
mpOpenApi-2.0 merge improvements

### LINKED ISSUE #19153: MP OpenAPI 2.0: Strip trailing slash from context root when merging
Not sure if this will happen in practise, but if `MergeProcessor` is given a context root which ends in `/`, the character should be removed.

### LINKED ISSUE #19154: MP OpenAPI 2.0: Handle server defined with relative URL
In an OpenAPI document, a server can be defined with a relative URL. In some cases I think we can prepend this to the path and remove the server element entirely which makes for a nicer merged document.

We particularly want to avoid ending up with paths which have a server defined with a URL of `/`

Azquelt:
* detect and remove a trailing slash in the context root
* if the server information is completely absorbed into the path names, clean up the empty server definitions
* port merge unit tests from openapi-3.1

Fixes #19154 
Fixes #19153


Azquelt:
#build
#libby

tevans78 (approved):

PR #19203, opened 2021-10-31
Remove unneeded restfulWSCleint feature from microProfile-5.0 feature file

jim-krueger:
The microProfile-5.0 feature brings in the restfulWS-3.0 feature, which in turn brings in the restfulWSClient-3.0 feature. So it is not necessary for microProfile-5.0 to also bring in the restfulWSClient-3.0 feature.


jim-krueger:
#build

andymc12 (approved):

PR #19209, opened 2021-11-01
configurable transaction context on ContextServiceDefinition

njr-11:
Honor the propagation setting for Transaction context on ContextServiceDefinition,
including getting the precedence correct with the similar execution property, which should override.


njr-11:
#build

KyleAure (approved):

PR #19241, opened 2021-11-04
beta of Concurrency 3.0

njr-11:
Pull to use for beta of Concurrency 3.0.  This pull was prepared in advance of being ready to save us some time on figuring out what changes are needed to get us through builds when this is eventually ready for beta.

### DIFF
--- a/dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.concurrency-3.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.concurrency-3.0.feature
@@ -5,6 +5,6 @@ singleton=true
 #TODO remove toleration of eeCompatible-9.0 once other EE 10 features are usable in beta form
 -features=com.ibm.websphere.appserver.eeCompatible-10.0; ibm.tolerates:="9.0"
 -bundles=io.openliberty.jakarta.concurrency.3.0; location:="dev/api/spec/,lib/"; mavenCoordinates="io.openliberty.jakarta.enterprise.concurrent:jakarta.enterprise.concurrent-api:3.0.0.20220120"
-kind=noship
-edition=full
+kind=beta
+edition=core
 WLP-Activation-Type: parallel
--- a/dev/com.ibm.websphere.appserver.features/visibility/auto/io.openliberty.cdi-concurrent3.0.feature
+++ b/dev/com.ibm.websphere.appserver.features/visibility/auto/io.openliberty.cdi-concurrent3.0.feature
@@ -8,5 +8,5 @@ IBM-Provision-Capability: \
 -bundles=\
   io.openliberty.concurrent.cdi.jakarta
 IBM-Install-Policy: when-satisfied
-kind=noship
-edition=full
+kind=beta
+edition=core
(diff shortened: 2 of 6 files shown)


njr-11:
#build

tevans78 (changes requested):
I think the feature must at least be prioritised before doing this

njr-11:
> I think the feature must at least be prioritised before doing this

I'll be sure not to merge it until that happens.  Prioritization has been waiting on the Upcoming Feature Overview doc, which has been waiting on the specification being far enough along to present one, but it is scheduled for tomorrow (Nov 4) and after that I'll bring this up on the Nov 8 prioritization call.  The goal is to have this ready with the option to beta if it gets approved.  It should be noted that there is quite a lot of matching function that we already get from having our MicroProfile implementation that goes a long way toward being Concurrency 3.0 compliant.

njr-11:
#build

njr-11:
The build had just one unit test failure in an unrelated area, and is good enough for merging, but this pull will be waiting for other approvals before that happens.

njr-11:
Rebased on to the latest. We still need official copies of the spec api jar and tck before merging.

njr-11:
#run-libby-bot
#build

njr-11:
VisibilityTest found an error, that eeCompatible-10.0 wasn't also marked beta. Fixed this, and rebuilding,
#build
#run-libby-bot

njr-11:
Rebased onto latest to pick up candidate final release concurrency API & TCK
#run-libby-bot
#build

njr-11:
@tevans78 Could you review this now? The feature is prioritized, had its Upcoming Feature Overview presented, has a green build and we're now using the 3.0 candidate for final release.

dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.concurrency-3.0.feature
```diff
@@ -5,6 +5,6 @@ singleton=true
 #TODO remove toleration of eeCompatible-9.0 once other EE 10 features are usable in beta form
 -features=com.ibm.websphere.appserver.eeCompatible-10.0; ibm.tolerates:="9.0"
 -bundles=io.openliberty.jakarta.concurrency.3.0; location:="dev/api/spec/,lib/"; mavenCoordinates="io.openliberty.jakarta.enterprise.concurrent:jakarta.enterprise.concurrent-api:3.0.0.20220120"
```
tevans78, line 7: perhaps doesn't matter so much for a beta but these maven co-ords are wrong.
```suggestion
-bundles=io.openliberty.jakarta.concurrency.3.0; location:="dev/api/spec/,lib/"; mavenCoordinates="io.openliberty.jakarta.enterprise.concurrent:jakarta.enterprise.concurrent-api:3.0.0.20211206"
```

tevans78 (approved):

dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.jakarta.concurrency-3.0.feature
njr-11: I don't see how 20220120 is wrong and the older 20211206 would be correct.  3.0.0.20220120 is how I versioned Liberty's [copy](http://public.dhe.ibm.com/ibmdl/export/pub/software/olrepo/io/openliberty/jakarta/enterprise/concurrent/jakarta.enterprise.concurrent-api/3.0.0.20220120/) of the Concurrency spec API jar that is the latest candidate for release of that specification which was built on January 20 (matching the version number).  20211206 is an older development built snapshot driver.  As you point out, by the time this is GA (and probably well before that point), we will be able to switch this to the real location on Maven of the official final release of the spec API jar.

tevans78: Yeah, sorry, I looked at the wrong file, my fault

PR #19260, opened 2021-11-05
Change variable resolution to variable expansion

jimblye:
Just changing the name we use to enable variable expansion in server.env to better align with the documentation.
The support for variable expansion in server.env is not released yet.  So this won't affect anyone.
Currently  # enable_variable_**resolution** is used, but it is more consistent with the doc if we use # enable_variable_**expansion**


hlhoots (approved):
LGTM

PR #19261, opened 2021-11-05
Increase the timeout duration for jdbc FAT testConfigChangeDataSource

gjwatts:
In the `com.ibm.ws.jdbc_fat` FAT bucket, the test `testConfigChangeDataSource` occasionally runs past its 5 second timeout value.  In this test, we want to make sure that the prepared statement is picking up the timeout value set in the transaction and not using the default 30 second timeout from the dataSource (defined in server.xml).  So as long as the `queryTimeout` value is greater than 30 (and less than or equal to 90), we can still be confident that the timeout is properly being propagated.  This increases our wait time from 5 seconds to 59.

The timing has been increased from:
```
if (queryTimeout < 85 || queryTimeout > 90)
```

to:
```
if (queryTimeout < 31 || queryTimeout > 90)
```


njr-11 (approved):

gjwatts:
#build

PR #19273, opened 2021-11-08
use io.openliberty.io.netty bundle from grpc client

### LINKED ISSUE #19246: Update gRPC client to use io.openliberty.io.netty bundle
A handful of gRPC bundles, for example `io.openliberty.grpc.1.0.internal.client` and `io.openliberty.grpc.client.1.0.thirdparty`, point to individual netty binaries rather than the `io.openliberty.io.netty` bundle Liberty provides (currently not shipped). These bundles should be cleaned up to only point to the `io.openliberty.io.netty` bundle. Additionally, `io.openliberty.grpc.1.0.internal.client` should not export the netty packages, but rather the grpc client feautre should be updated to use `io.openliberty.io.netty` at runtime.

wtlucy:
Fixes #19246


wtlucy:
#build

wtlucy:

PR #19279, opened 2021-11-08
Fix jwtSso-1.0 EE9 enablement in the OIDC server fat

c00crane:
The tests that use jwtSso-1.0 in the oidc server fat project are not properly triggering the including of EE9 features.  The apps are being "transformed", but the jars they need after the transform and not available because the proper features are not available.
Updating the one config file with a feature that will be converted instead of adding .alwaysAddFeature(“servlet-5.0”) to the repeat in the FATSuite.  That would add some overhead to every config file.


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.oidc.server_fat.oidc

teddyjtorres (approved):

PR #19280, opened 2021-11-08
Check default config settings to drive reprocessing config

tjwatson:
The variablesChanged check did not correctly check the defined variable
default values when calculating if any variable values had changed.

Also make the defaultValue @Sensitive for the ConfigVariable
constructor.


tjwatson:
#build

brenthdaniel (approved):

PR #19281, opened 2021-11-08
Increase timeout for testConfigChangeWithActiveConnections

gjwatts:
In the `com.ibm.ws.jdbc_fat` FAT, the client side test `testConfigChangeWithActiveConnections` calls a servlet that runs over the course of 40 time intervals.  At every interval, it creates an SQL Statement from a connection (defined as dataSource "dsfat5derby" in the `server.xml`) and checks the `queryTimeout` value of it.  In parallel, the client side attempts to manipulate that same dataSource definition and slowly increase the default `queryTimeout` value by 1 second every 100ms.  At the end of the test, the client checks to see if the default `queryTimeout` ever goes above 30s.

In this particular build break, we saw a `J2CA8040E` in messages.log likely because the server side was trying to use the dataSource information from the server.xml file at the same time the client side was trying to manipulate it and got caught.  For what little information about J2CA8040E I could find, [look here](https://www.ibm.com/docs/pt-br/was-zos/9.0.5?topic=SS7K4U_9.0.5/com.ibm.websphere.wlp.zseries.doc/ae/rwlp_restrict.html).

In trying to reproduce this problem locally (Windows 10), I found that the servlet piece ran so fast, it completely finished before the client side could even update the default `queryTimeout` once.  The number of servlet checking intervals was reduced from 40 to 20 and the interval duration was changed from 100ms to 867ms to be an uneven interval of the sleep delay on the client side (100ms).  This will hopefully reduce the likelihood that the servlet will be trying to access the dataSource at the same time the client is updating it.

### DIFF
--- a/dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
+++ b/dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
@@ -112,6 +112,11 @@ public class ConfigTest extends FATServletClient {
     private static ServerConfiguration originalServerConfig;
     private static ServerConfiguration originalServerConfigUpdatedForJDBC;
 
+    static final long FIVE_MINUTE_MS_TIMEOUT = 300000; // Five minutes in milliseconds
+
+    // Interval (in milliseconds) that tests should use for polling
+    static final long POLLING_INTERVAL_MS = 500; // 500 milliseconds
+
     @BeforeClass
     public static void setUp() throws Exception {
         // Delete the Derby database that might be left over from last run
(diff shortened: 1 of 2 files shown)


njr-11 (commented):
This may help some, but we ought to align this with the timing tolerance that we are giving to other tests.  Throughout other FAT buckets, a 2 minute timeout is used for simple operations to tolerate various build infrastructure.  A configuration update is more complex and involves additional filesystem access.  Rather than fixed iterations of 40 or 20, maybe up to 5 minutes would be good here?  An increase in the 100ms sleep as you are doing is fine, and then stop once you see the first change so that the test will normally run quickly.

The corresponding code in ConfigUpdate should follow this same pattern, looping until it sees a result from the servlet method indicating that it saw the change.

gjwatts:
@njr-11 Good suggestions.  Thank you for the feedback Nathan!
I have implemented the changes you mentioned.  Would you please re-review them?

dev/com.ibm.ws.jdbc_fat/fat/src/com/ibm/ws/jdbc/fat/tests/ConfigTest.java
```diff
@@ -1183,18 +1192,21 @@ public void run() {
...
             throw x;
         } finally {
-            Object result = results.poll(10, TimeUnit.SECONDS);
-            if (result == null)
-                throw new Exception("Test did not complete within allotted time");
-            else if (result instanceof Throwable)
-                throw (Throwable) result;
+            if (!success) {
+                throw new Exception("Test testConfigChangeWithActiveConnections did not complete within the allotted time");
+            }
```
njr-11: When `success` is false because the servlet method failed, this finally block obscures the actual failure and replaces it with a more obscure error that will make debugging more difficult.  You should move this code block out of the finally and put at the end of the try block where it is known that no other error has occurred.

gjwatts: Good catch!  I have updated that and added the actual timeout value to the timeout exception thrown.

njr-11 (approved):

gjwatts:
#build

gjwatts:
#build

gjwatts:
#build

gjwatts:
#build

gjwatts:
#libbybot

PR #19283, opened 2021-11-09
Update security repeat rules

c00crane:
Update the security repeat rules. Add a method that will allow us to not run some repeats with EE9 on windows. Add a method that will only run classes on windows.
The two rules together allow us to run a repeat with EE9, but when on windows that repeat really just runs normally without EE9.
On windows, the tests are timing out just trying to run the feature replacement (that's part of the EE9 run) - it's too slow.
I've removed some project specific rules and replaced use of them with more generic "security" rules.
I've removed the transform code copied into many of the FATSuite.java files with calls to common transform code that looks in our standard app locations and builds the list of apps that need to be transformed.  (replacing calls that needed to pass a list of all the apps that have to be transformed)


c00crane:
#build

ayoho (approved):

PR #19305, opened 2021-11-09
add code to wait for keystore readiness

c00crane:
The consumer tests are failing because the rsa_trust keystore was not loaded before the test was run.  I'm adding some code to the setup method that will hopefully make the test class wait for keystore config to be loaded before running any tests.  If this works, other test classes can use the "waitForKeyStores" method in ServerInstanceUtils.java
There is no real message that we can check for in the messages.log.
The new method will list the keystores that are in the config and search for a specific trace entry.
We'll have to see if this check is sufficient to prevent the issue we're seeing in automated testing.


c00crane:
#build

c00crane:
I can't repro the issue locally, so, I'm not sure if the config will have the keystores before they keymanager is initialized.  If the config is only updated after the keymanager is updated, my change will have to be modified to have the calling method pass in the list of keystore id's to wait for instead of querying the config.

ayoho (approved):

c00crane:
#build

c00crane:
#build

PR #19312, opened 2021-11-10
Allow utility scripts to expand variables in server.env

### LINKED ISSUE #19311: Utiltity scripts do not expand variables in server.env
issue #14901 added support for expanding variables in the server.env on Linux.   For example, you can now specify the following in server.env and the value of B will be expanded and assigned to the variable A.
A=$B

However, there are about a dozen scripts in the wlp/bin directory that also read server.env, and they do not support variable expansion.

All of these scripts share common code which reads server.env.   So only one file needs to be updated to add this support:

/dev/cnf/resources/bin/tool

jimblye:


jimblye:
For this, I just took the readServerEnv "method" from the server script and replaced the same method in
/dev/cnf/resources/bin/tool

No change is necessary on Windows.
During the build, this tool script is magically merged into several other utility scripts in wlp/bin

jimblye:
#build #libby

cbridgha (approved):
Thanks - looks good

jimblye:
#build
#libby

PR #19339, opened 2021-11-11
Fix the failure caused during restore when messages.log is deleted

### LINKED ISSUE #18395: Restore from checkpoint fails if messages.log is removed
If a server is restored from a checkpoint after the logs/ directory has been cleared then the criu restore operation will fail with the following in the restore.log:

```
(00.007538) 203711: Error (criu/files-reg.c:1767): Can't open file .../wlp/usr/servers/petclinic/logs/messages.log: No such file or directory
(00.007542) 203711: Error (criu/files.c:1210): Unable to open fd=13 id=0x39
(00.007622) Error (criu/cr-restore.c:2397): Restoring FAILED.
(00.007816) Error (criu/cr-restore.c:1479): 203711 killed by signal 9: Killed

```

A checkpoint hook is needed to close/reopen the messages.log on checkpoint/restore to avoid checkpointing the file handle for messages.log in the checkpoint image.

anjumfatima90:
fixes #18395

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
+++ b/dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
@@ -32,6 +32,7 @@ import java.util.Hashtable;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
@@ -50,7 +51,9 @@ import org.osgi.framework.launch.FrameworkFactory;
 import com.ibm.ejs.ras.TraceNLS;
 import com.ibm.websphere.ras.DataFormatHelper;
 import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TrConfigurator;
 import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.ws.ffdc.FFDCConfigurator;
 import com.ibm.ws.ffdc.annotation.FFDCIgnore;
 import com.ibm.ws.kernel.LibertyProcess;
 import com.ibm.ws.kernel.boot.BootstrapConfig;
@@ -68,6 +71,7 @@ import com.ibm.ws.kernel.boot.internal.commands.ServerDumpUtil;
 import com.ibm.ws.kernel.boot.jmx.internal.PlatformMBeanServerBuilder;
 import com.ibm.ws.kernel.boot.jmx.internal.PlatformMBeanServerBuilderListener;
 import com.ibm.ws.kernel.boot.jmx.service.MBeanServerPipeline;
+import com.ibm.ws.kernel.launch.internal.LauncherDelegateImpl.ReadOnlyFrameworkProperties;
 import com.ibm.ws.kernel.launch.internal.Provisioner.InvalidBundleContextException;
 import com.ibm.ws.kernel.launch.service.ClientRunner;
 import com.ibm.ws.kernel.launch.service.ForcedServerStop;
@@ -84,6 +88,8 @@ import com.ibm.ws.kernel.provisioning.BundleRepositoryRegistry;
 import com.ibm.wsspi.logging.Introspector;
 import com.ibm.wsspi.logprovider.LogProvider;
 
+import io.openliberty.checkpoint.spi.CheckpointHook;
+import io.openliberty.checkpoint.spi.CheckpointHookFactory;
 import io.openliberty.checkpoint.spi.CheckpointHookFactory.Phase;
 
 /**
@@ -198,7 +204,7 @@ public class FrameworkManager {
      *                        framework management activities (start/stop/.. ), or null
      * @param callback
      */
-    public void launchFramework(BootstrapConfig config, LogProvider logProvider) {
+    public void launchFramework(final BootstrapConfig config, final LogProvider logProvider) {
         if (config == null)
             throw new IllegalArgumentException("bootstrap config must not be null");
         boolean isClient = config.getProcessType().equals(BootstrapConstants.LOC_PROCESS_TYPE_CLIENT);
@@ -272,6 +278,31 @@ public class FrameworkManager {
             }
             // Set the framework variables only if everything succeeded.
             systemBundleCtx = fwk.getBundleContext();
+
+            systemBundleCtx.registerService(CheckpointHookFactory.class, new CheckpointHookFactory() {
+
+                @Override
+                public CheckpointHook create(Phase phase) {
+                    CheckpointHook checkpointHook = new CheckpointHook() {
+                        @Override
+                        public void prepare() {
+                            logProvider.stop();
+                        }
+
+                        @Override
+                        public void restore() {
+                            config.put(BootstrapConstants.RESTORE_ENABLED, "true");
+                            Map<String, Object> configMap = Collections.<String, Object> unmodifiableMap(new ReadOnlyFrameworkProperties(config));
+                            TrConfigurator.update(configMap);
+                            FFDCConfigurator.update(configMap);
+                        }
+
+                    };
+                    return checkpointHook;
+                }
+
+            }, null);
+
             framework = fwk;
         } catch (BundleException e) {
             throw new RuntimeException(e);
--- a/dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
+++ b/dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
@@ -41,6 +41,7 @@ import java.util.concurrent.TimeUnit;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleException;
+import org.osgi.framework.Constants;
 import org.osgi.framework.FrameworkEvent;
 import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.InvalidSyntaxException;
@@ -53,7 +54,6 @@ import com.ibm.websphere.ras.DataFormatHelper;
 import com.ibm.websphere.ras.Tr;
 import com.ibm.websphere.ras.TrConfigurator;
 import com.ibm.websphere.ras.TraceComponent;
-import com.ibm.ws.ffdc.FFDCConfigurator;
 import com.ibm.ws.ffdc.annotation.FFDCIgnore;
 import com.ibm.ws.kernel.LibertyProcess;
 import com.ibm.ws.kernel.boot.BootstrapConfig;
@@ -71,7 +71,6 @@ import com.ibm.ws.kernel.boot.internal.commands.ServerDumpUtil;
 import com.ibm.ws.kernel.boot.jmx.internal.PlatformMBeanServerBuilder;
 import com.ibm.ws.kernel.boot.jmx.internal.PlatformMBeanServerBuilderListener;
 import com.ibm.ws.kernel.boot.jmx.service.MBeanServerPipeline;
-import com.ibm.ws.kernel.launch.internal.LauncherDelegateImpl.ReadOnlyFrameworkProperties;
 import com.ibm.ws.kernel.launch.internal.Provisioner.InvalidBundleContextException;
 import com.ibm.ws.kernel.launch.service.ClientRunner;
 import com.ibm.ws.kernel.launch.service.ForcedServerStop;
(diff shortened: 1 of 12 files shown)


dev/com.ibm.ws.kernel.boot_test/test/com/ibm/ws/kernel/boot/internal/framework/BasicLogProvider.java
```diff
@@ -26,4 +26,8 @@ public void configure(Map<String, String> config,
 
     @Override
     public void stop() {}
+
```
tjwatson: Why is this override here?

dev/com.ibm.ws.logging.core/src/com/ibm/wsspi/logprovider/LogProvider.java
```diff
@@ -54,4 +54,12 @@
      * open resources when this method is called.
      */
     void stop();
+    
+    /**
+     * Update the configuration of the Log provider: config is a map containing string key/object
+     * value pairs. 
+     * @param config
+     */
+    void update(Map<String, Object> config);
```
tjwatson: This is related to my question above about why is this method overriden.  I believe this package to be SPI and that LogProvider may be implemented by other product extensions (not part of Liberty itself).  As such I think this will need to be a default method.

dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
```diff
@@ -262,6 +267,49 @@ public Void run() throws Exception {
...
+                public CheckpointHook create(Phase phase) {
+                    CheckpointHook checkpointHook = new CheckpointHook() {
+                        @Override
+                        public void prepare() {
+                            logProvider.stop();
+                        }
+
+                        @Override
+                        public void restore() {
+                            ClassLoader loader = config.getFrameworkClassloader();
```
tjwatson: Can you get the logProvider class from the logProvider instance that is there at checkpoint time?  I am pretty sure it has to be the exact same class name that you would get below from `config.getKernelResolver().getLogProvider()`

With that said, I am unsure why you need to create a new instance of the logProvider vs. constructing a new instance and calling update on it.  If you are creating a new instance why would it not get initialized the same way as it would the first time it is created before in checkpoint?

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/LogProviderImpl.java
```diff
@@ -34,6 +34,12 @@ public void configure(Map<String, String> config, File logLocation, TextFileOutp
         FFDCConfigurator.init(loggingConfig);
     }
 
+    @Override
+    public void update(Map<String, Object> config) {
+        TrConfigurator.update(config);
```
tjwatson: It seems all implementations of this do the same thing.  Can we not just call these two static methods ourselves?  Would that simplify things so we don't have to update the LogProvider interface?

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/LoggingConstants.java
```diff
@@ -65,4 +65,6 @@
     enum FFDCSummaryPolicy {
         DEFAULT, IMMEDIATE
     };
+
+    String CHECKPOINT_ENABLED = "CHECKPOINT_ENABLED";
```
tjwatson: is this used?

dev/com.ibm.ws.logging.hpel/src/com/ibm/ws/logging/internal/hpel/HpelLogProviderImpl.java
```diff
@@ -46,6 +46,12 @@ public void configure(Map<String, String> config, File logLocation, TextFileOutp
         HpelConfigurator.init(loggingConfig);
     }
 
+    @Override
+    public void update(Map<String, Object> config) {
+        TrConfigurator.update(config);
```
tjwatson: I see no updates to TrConfigurator or FFDCConfigurator that would use the updated key.  Not sure how this flows down.

dev/com.ibm.ws.kernel.boot_test/test/com/ibm/ws/kernel/boot/internal/framework/BasicLogProvider.java
anjumfatima90: This is because it implements LogProvider. I will make the update method default then i won't have to override it.

dev/com.ibm.ws.logging.hpel/src/com/ibm/ws/logging/internal/hpel/HpelLogProviderImpl.java
anjumfatima90: TrConfigurator.update(Map<String, Object>) has a line `config.update(newConfig);` which calls `LogProviderConfigImpl.update(Map<String, Object> -> LogProviderConfigImpl.doCommonInit(Map, boolean))` where I made changes.

dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
```diff
@@ -272,6 +278,31 @@ public Void run() throws Exception {
...
+                            Map<String, Object> configMap = Collections.<String, Object> unmodifiableMap(new ReadOnlyFrameworkProperties(config));
+                            TrConfigurator.update(configMap);
+                            FFDCConfigurator.update(configMap);
+                        }
+
+                    };
+                    return checkpointHook;
+                }
+
+            }, null);
```
tjwatson: Is a service.ranking property needed here to ensure the hooks prepare is called last and the restore is called first?  I would think the service.ranking should be Integer min value so that it gets sorted properly when CheckpointImpl calls it.

dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
```diff
@@ -272,6 +278,31 @@ public Void run() throws Exception {
...
+                    CheckpointHook checkpointHook = new CheckpointHook() {
+                        @Override
+                        public void prepare() {
+                            logProvider.stop();
+                        }
+
+                        @Override
+                        public void restore() {
+                            config.put(BootstrapConstants.RESTORE_ENABLED, "true");
+                            Map<String, Object> configMap = Collections.<String, Object> unmodifiableMap(new ReadOnlyFrameworkProperties(config));
```
tjwatson: Why wrapper the ReadOnlyFrameworkProperties as an unmodifiableMap?  Seems redundant?

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -700,9 +715,20 @@ public void stop() {
```
tjwatson, line 719: On prepare we call stop().  This seems to restoreSystemStreams.  On restore I assume the system streams go back to the state they were?

anjumfatima90:
#build

dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
tjwatson: I worry that this update with the original configMap may override settings that could happen dynamically from:

com.ibm.ws.logging.internal.osgi.LoggingConfigurationService.updated(Dictionary)

Is there any way to make it such that the TrConfigurator and FFDCConfigurator updates only pay attention to the RESTORE_ENABLED key when present and ignore all other keys?  This way we could have a singleton Map that only contains RESTORE_ENABLED key and all other settings that are currently set stay the same?

anjumfatima90:
#build

anjumfatima90:
#build

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/LogProviderConfigImpl.java
```diff
@@ -196,6 +200,8 @@ public LogProviderConfigImpl(Map<String, String> config, File logLocation, TextF
         serverName = config.get("wlp.server.name");
 
         wlpUsrDir = config.get("wlp.user.dir");
+
+        checkpoint = config.get(LoggingConstants.CHECKPOINT_PROPERTY_NAME) != null ? true : false;
```
pgunapal, line 204: It would be better to use our existing method to get the Boolean value here, to be consistent:
`checkpoint = LoggingConfigUtils.getBooleanValue(config.get(LoggingConstants.CHECKPOINT_PROPERTY_NAME), checkpoint );`

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/LogProviderConfigImpl.java
```diff
@@ -126,6 +126,10 @@
     /** Allow JSON from applications write directly to System.out/System.err */
     protected volatile boolean appsWriteJson = false;
 
+    private final boolean checkpoint;
```
pgunapal, line 129: - Initialize this to equal false.
- Variables should be in camelCase to match other variables in the class.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/utils/FileLogHolder.java
```diff
@@ -142,7 +142,7 @@ private static TraceComponent getTc() {
     public static FileLogHolder createFileLogHolder(TraceWriter oldLog, FileLogHeader logHeader,
```
pgunapal, line 142: Update copyright header to 2021.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/LogProviderConfigImpl.java
```diff
@@ -126,6 +126,10 @@
     /** Allow JSON from applications write directly to System.out/System.err */
```
pgunapal, line 126: Update copyright header to 2021.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/Jsr47TraceService.java
```diff
@@ -56,13 +56,17 @@ protected void unregisterLoggerHandlerSingleton() {}
     protected void initializeWriters(LogProviderConfigImpl config) {
         if (!WsLogManager.isConfiguredByLoggingProperties()) {
             // createFileLog may or may not return the original log holder..
```
pgunapal, line 60: Update copyright header to 2021.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -315,11 +319,18 @@ protected void unregisterLoggerHandlerSingleton() {
     @Override
     public synchronized void update(LogProviderConfig config) {
         LogProviderConfigImpl trConfig = (LogProviderConfigImpl) config;
+        checkpoint = trConfig.isCheckpoint();
+        restore = trConfig.isRestore();
+        if (isRestore()) {
```
pgunapal: Isn't it better to use the restore value you got from line 324, instead of calling `isRestore()` directly?

pgunapal (changes requested):

anjumfatima90:
#build

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/LogProviderConfigImpl.java
tjwatson: > * Initialize this to equal false.

It is a final field, it cannot be initialized and changed later

> * Variables should be in camelCase to match other variables in the class.

`checkpoint` is a single word, not "check point".

dev/com.ibm.ws.kernel.boot.nested/src/com/ibm/ws/kernel/launch/internal/FrameworkManager.java
```diff
@@ -272,6 +277,30 @@ public Void run() throws Exception {
             }
             // Set the framework variables only if everything succeeded.
             systemBundleCtx = fwk.getBundleContext();
+
+            if (LaunchArguments.isBetaEdition()) {
```
tjwatson, line 281: We don't need this beta block here.  CheckpointHookFactory services only get used when the beta checkpoint feature is enabled.

tjwatson (approved):
I'm going to merge this into criu to make progress in other areas.  @anjumfatima90 please open a separate PR to address the remaining two comments here.

PR #19354, opened 2021-11-15
Don't exposed processed file paths

### LINKED ISSUE #19366: JMX file transfer errors shouldn't expose resolved file paths
If a request is made to the file transfer API and the relevent file doesn't exist, a JSON error object is returned. E.g. if the request is:
 DELETE /IBMJMXConnectorREST/file/${server.config.dir}/server.xml
then the response will be something like:
  {
    "throwable": "..."
    "error": "CWWKX0121E: Access denied to path /opt/IBM/WebSphereLiberty/usr/servers/server2/server.xml."
  }
where the server.config.dir property is resolved and exposed in the error message. Exposing server internals is not good practice, and this error message should be changed.

idlewis:
If a file is not accessible from the FileTransfer rest handler,
then the error message returned should not contain the processed
path, it should only contain the originally requested path.
Otherwise, internal file path information could be leaked back
to the requester

fixes #19366

### DIFF
--- a/dev/com.ibm.ws.jmx.connector.server.rest/src/com/ibm/ws/jmx/connector/server/rest/helpers/FileTransferHelper.java
+++ b/dev/com.ibm.ws.jmx.connector.server.rest/src/com/ibm/ws/jmx/connector/server/rest/helpers/FileTransferHelper.java
@@ -238,12 +238,12 @@ public class FileTransferHelper {
 
         if (!checkAccess(processedPath, readOnly)) {
             //Return an exception
-            Object[] params = new String[] { processedPath };
+            Object[] params = new String[] { filePath };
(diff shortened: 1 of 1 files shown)


idlewis:
#build

idlewis:
#build

idlewis:
#build

idlewis:
#build

idlewis:
#build

idlewis:
#run-libby-bot

leochr (approved):
@idlewis Looks good. Thanks

leochr:
@idlewis Noticed that deleteInternal and uploadInternal also have similar issues. Processed path is used in error message if delete or exapansion fails. It'll be good to address those as well. Thanks.

PR #19363, opened 2021-11-15
Update to GA binaries of MP context propagation 1.3

njr-11:
This pull updates us to the GA spec binaries and TCK of MicroProfile Context Propagation 1.3.  To clarify, this is not the same as GA of the mpContextPropagation-1.3 Liberty feature, which will be done as a follow-on pull after this.

### DIFF
--- a/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
+++ b/dev/cnf/dependabot/check_this_in_if_it_changes/pom.xml
@@ -1879,7 +1879,7 @@
     <dependency>
       <groupId>org.eclipse.microprofile.context-propagation</groupId>
       <artifactId>microprofile-context-propagation-api</artifactId>
-      <version>1.3-RC1</version>
+      <version>1.3</version>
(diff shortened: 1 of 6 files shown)


njr-11:
#build

njr-11:
#build

njr-11:
The above build has only the error
`Could not find artifact org.eclipse.microprofile.context-propagation:microprofile-context-propagation-tck:jar:1.3 `
which is exactly what we want to see at this point given that the final 1.3 tck jar has not yet been pushed to maven.
One the 1.3 specification is finalized and the binaries are pushed over, I'll resubmit the build and everything should be working (the TCK passes locally with the 1.3 TCK, and is already passing when running with the identical 1.3-RC1 TCK from Maven, so we do not expect any failures)

njr-11:
#build

njr-11:
#run-libby-bot

njr-11:
#build

nmittles (approved):

PR #19387, opened 2021-11-18
mpOpenApi-2.0: avoid processing apps on shutdown

### LINKED ISSUE #19375: mpOpenApi 2.0 cache fails intermittently, which increases startup time.
I have a test case where I have 10 apps and I am running in the default "openapi mode" where only 1 of them will be processed by mpOpenAPI.

At server start, 1 of the 10 apps is selected as the default and generates a model and gets put into the cache (if it's not there already and has not been "changed"). This also happens during a server stop, and sometimes it does it multiple times, 1 to 3 apps may be added to the cache during a stop. I am guessing as the "default app" gets stopped, a new app is made the "default app" so it becomes the one openapi cares about and gets processed. Should this be happening during a stop?

However, sometimes, the isCacheUpToDate is false  here:
https://github.com/OpenLiberty/open-liberty/blob/3cc24d590ff5305ee82cded977ab9cce814873be/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/cache/CacheEntry.java#L198

It looks like the properties differ during a server start compared to later at a server stop (see privatePropertiesEnable below). So, if an app gets processed during a different step than it was prior, the props will differ, and it will be marked as a cache miss and get regenerated (which adds to startup time if it is during a server start)

Start
```
{privatePropertiesEnable=false, getDefaultProduces=Optional.empty, allowNakedPathParameter=Optional.empty, getSchemas=, getDefaultConsumes=Optional.empty, scanExcludeClasses=, sortedPropertiesEnable=false, arrayReferencesEnable=true, scanPackages=, scanDependenciesDisable=false, scanExcludePackages=(\Qjava.lang\E), scanClasses=, scanDisable=false, servers=, propertyNamingStrategy=IDENTITY, applicationPathDisable=false, scanDependenciesJars=}
```
 
Stop
```
{privatePropertiesEnable=true, getDefaultProduces=Optional.empty, allowNakedPathParameter=Optional.empty, getSchemas=, getDefaultConsumes=Optional.empty, scanExcludeClasses=, sortedPropertiesEnable=false, arrayReferencesEnable=true, scanPackages=, scanDependenciesDisable=false, scanExcludePackages=(\Qjava.lang\E), scanClasses=, scanDisable=false, servers=, propertyNamingStrategy=IDENTITY, applicationPathDisable=false, scanDependenciesJars=}
```

Guessing this was introduced with this (Pull 17761 - by default privatePropertiesEnable=false).
https://github.com/Azquelt/open-liberty/blob/65071eafa9c641e3b0e786a15ac2894bc4034b3a/dev/io.openliberty.microprofile.openapi.2.0.internal/resources/OSGI-INF/wlp/defaultInstances.xml
 
If I set mp.openapi.extensions.smallrye.private-properties.enable=true, the mismatch goes away.
 
So, when/why does it get changed to true after startup?

Azquelt:
In the default configuration, where only one app is processed to create
OpenAPI documentation, when that app is removed we then processed the
next app so that while any app is deployed we will have OpenAPI
documentation for one of them.

During shutdown all apps are removed in turn so we can end up creating
OpenAPI documentation for some apps while the server is shutting down,
which is a waste of time.

Fixes #19375

### DIFF
--- a/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationRegistry.java
+++ b/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/ApplicationRegistry.java
@@ -35,6 +35,7 @@ import com.ibm.ws.container.service.app.deploy.ContainerInfo.Type;
 import com.ibm.ws.container.service.app.deploy.WebModuleInfo;
 import com.ibm.ws.ffdc.annotation.FFDCIgnore;
 import com.ibm.ws.kernel.feature.ServerStartedPhase2;
+import com.ibm.wsspi.kernel.service.utils.FrameworkState;
 
 import io.openliberty.microprofile.openapi20.merge.MergeProcessor;
 import io.openliberty.microprofile.openapi20.utils.Constants;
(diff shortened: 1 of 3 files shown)


Azquelt:
#build

tevans78 (approved):

PR #19390, opened 2021-11-18
Register the Phase as a service

tjwatson:
This PR has a few fixes to clean up checkpoint

1. The Phase enum is registered by the kernel.boot as a service.  This allows service components reference the phase as a requirement to activate.  The phase service also has its type stored as a service property so particular phases can be targeted with a filter.  For example: `(io.openliberty.checkpoint=APPLICATIONS)`.
2. A couple more env keys are filtered out for restore to avoid changing the values of the X_LOG_DIR and X_LOG_FILE.
3. The restore method of the hooks is called in the reverse order of the prepare hooks.  This follows a similar pattern to things like interceptors that use a call order like A->B->C action C->B->A.
4. Adds a hook to handle prepare/restore for the OSGi console.


tjwatson:
#build

tjwatson:
#build

anjumfatima90 (approved):

sebratton (approved):

PR #19407, opened 2021-11-19
Make sure the release lock request is based on the same authdata that was used for the obtain lock

### LINKED ISSUE #19567: Memory Leak with mpJWT.
Describe the bug:
When using  JWTs to authenticate into a server, there may be a memory leak after running for a long time leading to an OutOfMemory Error.

Steps to Reproduce:
Use JWTs to authenticate to a server many, many, times.

Expected behavior
No memory leak and no OutOfMemoryError

Diagnostic information:
OpenLiberty Version: 21.0.0.11 (but may have been around for a few releases prior)
Affected feature(s): mpJWT-x.x

A workaround is to disable the authCache by adding the following to the server.xml.
```
 <authentication id="Basic" cacheEnabled="false" />
 ```

jdmcclur:
I hit an Out of Memory (OOM) Error in one of test cases. In my case, the HashMap in AuthenticationGuard never has entries removed from it because the key used for the put (hashtableAuthData) is different from the key used later to see if it should be removed (authenticationData). This eventually leads to a large HashMap and OOM Error.The change here is to make sure the same key is used in relinquishAccess that was used in requestAccess, which ensures that entries are removed from the HashMap.

fixes #19567


jdmcclur:
#build

utle (approved):

PR #19409, opened 2021-11-19
NullPointerException when max or priority are unspecified

njr-11:
This fixes the NullPointerException that can arise from a couple of different places if one of the int type attributes (max or priority) is unspecified.


mswatosh (approved):

njr-11:
#build

PR #19419, opened 2021-11-19
Create FAT for new config attr jwtAccessTokenRemoteValidation

### LINKED ISSUE #17448: Allow OIDC clients to introspect propagated JWT access tokens
For inbound propagation scenarios, OIDC clients will only perform local validation of access tokens in JWT format. If validation of the JWT fails and `inboundPropagation="required"`, we will fail the request. If validation of the JWT fails and `inboundPropagation="supported"`, we'll fall back to the normal OIDC login flow.

This issue is for adding support for doing remote validation of propagated JWT access tokens. Possible usage scenarios are:
- Server admins want propagated JWTs to be validated by the OP
- JWTs not understood by our RP (for whatever reason) could still be permitted by falling back to validating with the OP

c00crane:
Create new FAT tests in support of jwtAccessTokenRemoteValidation.
Add tests to verify that when set to none, we'll only do local validation for propagated tokens.
Add tests to verify that when set to allow, we'll do local validation for propagated tokens and if that fails, we'll use the validation endpoint to try to validate the token.
Add tests to verify that when set to required, we'll only use the validation endpoint to try to validate the propagated token.

### DIFF
--- a/dev/com.ibm.ws.security.oauth.oidc_fat.common/src/com/ibm/ws/security/oauth_oidc/fat/commonTest/TestServer.java
+++ b/dev/com.ibm.ws.security.oauth.oidc_fat.common/src/com/ibm/ws/security/oauth_oidc/fat/commonTest/TestServer.java
@@ -6,7 +6,7 @@
  * http://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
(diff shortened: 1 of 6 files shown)


c00crane:
#build

ayoho (approved):

PR #19284, opened 2021-11-09
Split jwtsso fat

c00crane:
Split the jwtsso fat.
This fat runs the same tests multiple times - each run specifies a different setup of mpJwt.
It runs without mpJwt (and that runs with and without EE9)
It runs with mpJwt 1.1, mpJwt 1.2 and mpJwt 2.0.  This just got too large and time consuming.
com.ibm.ws.security.jwtsso_fat			
has now become:
com.ibm.ws.security.jwtsso_fat.commonTest - contains the tests
com.ibm.ws.security.jwtsso_fat.mpJwt-1.1 - runs with mpJwt 1.1
com.ibm.ws.security.jwtsso_fat.mpJwt-1.2 - runs with mpJwt 1.2
com.ibm.ws.security.jwtsso_fat.mpJwt-2.0 - runs with mpJwt 2.0 (which include EE9)
com.ibm.ws.security.jwtsso_fat.noMpJwt - has no mpJwt in the config and runs with/without EE9

PR #19429, opened 2021-11-22
Update to xercesImpl 2.12.1

tevans78:
This is just step 1 of the intended changes; to update the xercesImpl version to the latest available.

#build


benjamin-confino (approved):

wraschke:
@tevans78 your PB has a hard break in it, caused by https://github.com/OpenLiberty/open-liberty/pull/19284. This PR builds on top of that, which we need to revert, and so we also need to revert yours. See my comment in 19284 for why.

PR #19435, opened 2021-11-22
#19433 Treat AdapterAlreadyExists as unexceptional

### LINKED ISSUE #19433: JNDI lookup to CORBA URL can hang
**Describe the bug**  
A naming lookup that results in the first call to the CORBA COSNaming NameService service should cause the service to be activated.

If two concurrent naming lookups find the service not yet activated, a race condition can occur, and one of the calls may hang.

This arises because the activator incorrectly propagates an exception:
`org.omg.PortableServer.POAPackage.AdapterAlreadyExists`

**Steps to Reproduce**  
This bug is hard to reproduce as it relies on hitting a very narrow timing window.

**Expected behavior**  
The `NamingServiceAdapterActivator` should simply return false to indicate that activation failed.

One call should activate the service and use it, and the other call should use the already-activated service.
Both calls should succeed.

**Diagnostic information:**  
 - OpenLiberty Version: 21.x
 - Affected feature(s) ejbRemote
 - Java Version: 8+

joe-chacko:
Fixes #19433


joe-chacko:
#build

ngmr (approved):
Looks good to me.

PR #19462, opened 2021-11-23
JAX-WS: Undo LibertyApplicationBusFactory classloader change

neuwerk:
This PR undoes a change made earlier to the LibertyApplictionBusFactory, as an alternative route was fixed in xmlBinding-3.0.


neuwerk:
#build

PR #19466, opened 2021-11-24
Add support to start selectors immediately for outbound channels

### LINKED ISSUE #19702: Support for outbound channel selectors to start immediately
**Describe the bug**  
There is a tcpOptions attribute, `waitToAccept`, that is meant to query whether the TCP Channel will delay accepting connections until the server start. As the attribute suggests, this possible delay should apply to selectors used for accepting inbound connections. However, any selector, be it created for inbound or outbound scenarios is currently subject to this behavior.  

Outbound selectors should not have a delayed start. Outbound selectors, by default, will currently wait until the kernel has signaled that the server has fully started to be initialized. The TCP transport should be updated to allow selectors in outbound cases to start immediately. 

**Expected behavior**  
Selectors started in outbound scenarios will not delay starting.

**Diagnostic information:**  
 - OpenLiberty Version: [e.g. 22.0.0.1]
 - Affected feature(s) - Users of outbound TCP channels created using the channelfw project
 - Java Version: 11.0.3

mrsaldana:
Fixes #19702

### DIFF
--- a/dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/TCPChannelConfiguration.java
+++ b/dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/TCPChannelConfiguration.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2021 IBM Corporation and others.
+ * Copyright (c) 2005, 2020 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -40,7 +40,6 @@ public class TCPChannelConfiguration implements TCPConfigConstants, FFDCSelfIntr
     protected static final String DIRECT_BUFFS = "allocateBuffersDirect";
     protected static final String ACCEPT_THREAD = "acceptThread";
     protected static final String WAIT_TO_ACCEPT = "waitToAccept";
-    protected static final String START_OUTBOUND_SELECTORS_IMMEDIATELY = "startOutboundSelectorsImmediately";
     protected static final String PORT_OPEN_RETRIES = "portOpenRetries";
     protected static final String COMM_OPTION = "commOption";
     protected static final String DUMP_STATS_INTERVAL = "dumpStatsInterval";
@@ -84,7 +83,6 @@ public class TCPChannelConfiguration implements TCPConfigConstants, FFDCSelfIntr
     private int sendBufferSize = -1;
     private boolean acceptThread = false;
     private boolean waitToAccept = false;
-    private boolean startOutboundSelectorsImmediately = false;
     private int dumpStatsInterval = 0;
     private String endPointName = null;
 
@@ -383,15 +381,6 @@ public class TCPChannelConfiguration implements TCPConfigConstants, FFDCSelfIntr
                     if (key.equalsIgnoreCase(PORT_OPEN_RETRIES)) {
                         //This is a valid configuration option but outbound channels do not use it
                         //Adding this prevents a message from being output saying it's invalid
-                        continue;
-                    }
-
-                    if (key.equalsIgnoreCase(START_OUTBOUND_SELECTORS_IMMEDIATELY)) {
-                        // convert and check
-                        keyType = ValidateUtils.KEY_TYPE_BOOLEAN;
-                        this.startOutboundSelectorsImmediately = convertBooleanValue(value);
-                        result = ValidateUtils.VALIDATE_OK;
-
                         continue;
                     }
                 }
@@ -736,15 +725,6 @@ public class TCPChannelConfiguration implements TCPConfigConstants, FFDCSelfIntr
                     }
                 } else {
                     // outbound only parameters
-                    if (key.equalsIgnoreCase(START_OUTBOUND_SELECTORS_IMMEDIATELY)) {
-                        // convert and check
-                        keyType = ValidateUtils.KEY_TYPE_BOOLEAN;
-                        oldBool = this.startOutboundSelectorsImmediately;
-                        if (convertBooleanValue(value) != oldBool) {
-                            result = ValidateUtils.VALIDATE_NOT_EQUAL;
-                        }
-                        continue;
-                    }
                 }
 
                 if (key.equalsIgnoreCase(NO_DELAY)) {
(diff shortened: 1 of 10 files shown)


mrsaldana:
#build

mrsaldana:
#build - Disregard this build

mrsaldana:
#build

dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/TCPChannelConfiguration.java
```diff
@@ -293,6 +295,15 @@ private void setValues() throws ChannelException {
                         continue;
                     }
 
+                    if (key.equalsIgnoreCase(START_OUTBOUND_SELECTORS_IMMEDIATELY)) {
```
wtlucy: This statement is in a `isInbound()` block - I don't think that was your intention?

dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/WorkQueueManager.java
```diff
@@ -200,7 +205,7 @@ protected void startSelectors(boolean inBound) throws ChannelException {
                 // add
                 connectCount = new int[maxChannelSelectorsPerFlow]; // 269309 add
 
-                connect[0] = new ConnectChannelSelector(this, 0, CS_CONNECTOR);
+                connect[0] = new ConnectChannelSelector(this, 0, CS_CONNECTOR, startImmediately);
                 createNewThread(connect[0], CS_CONNECTOR, 1);
 
                 for (int i = 0; i < maxChannelSelectorsPerFlow; i++) {
```
wtlucy, line 211: should line 439 - `CS[nextOpen] = new ConnectChannelSelector(this, nextOpen, CS_CONNECTOR)` - pass `startImmediately` as well?

dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/WorkQueueManager.java
```diff
@@ -100,6 +101,11 @@ protected WorkQueueManager() {
...
     }
 
+    protected void startSelectors(boolean inBound, boolean startImmediately) throws ChannelException {
+        this.startImmediately = startImmediately;
+        this.startSelectors(inBound);
+    }
+
     protected void startSelectors(boolean inBound) throws ChannelException {
         if (TraceComponent.isAnyTracingEnabled() && tc.isEntryEnabled()) {
             Tr.entry(tc, "startSelectors " + inBound);
```
wtlucy, line 111: should this class _always_ pass `_checkCancel` to the `SocketRWChannelSelector` ctor? There are a few instances in `startSelectors()` and `moveIntoPosition()` where we don't

mrsaldana: The intention is that the new `startSelector(boolean, boolean)` signature is used by outbound scenarios. When used it will set a class level `startImmediately` flag, such that when the original `startSelector(boolean)` is invoked, all constructors for outbound scenarios will inherit this `startImmediately` flag. All other constructors (inbound cases) should remain unchanged to preserve current behavior.

dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/WorkQueueManager.java
mrsaldana: The constructor in line 439 is triggered when all active selectors are full. The change being proposed by this PR is during the scope of the server's startup. It is unlikely that there will be enough traffic during startup to need to need to create additional selectors.

dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/WorkQueueManager.java
wtlucy: Thanks for the explanation!

dev/com.ibm.ws.channelfw/src/com/ibm/ws/tcpchannel/internal/WorkQueueManager.java
mrsaldana: As a connect selector, line 439 should be outbound, so passing the config value directly. For the SocketRWChannelSelector, checking if its an oubtound read/write before passing the configuration value.

mrsaldana:
#build

wtlucy (approved):
LGTM!

mrsaldana:
#build

mrsaldana:
#build

mrsaldana:
Latest build is good, two errors not related to change.

AndrewDJBanks (dismissed):
Manuel, looking at this we don't appear to have any way to detect whether "startOutboundSelectorsImmediately" has taken the default value or has been configured to false. We don't want to unilaterally set this to true if the user has configured it to false. This would lead to confusing behaviour and would correctly be described as a bug.

Could you change the default to true. Or better still reverse the meaning to "deferOutboundConnectionsUntilServerStarted". The current problem is that the channel framework is blocking outbound connections when it did not need to. The users of the channel framework can decide when they wish to make a connection and the only need for a parameter is to allow the possibility of restoring the old behaviour.

mrsaldana:
> Manuel, looking at this we don't appear to have any way to detect whether "startOutboundSelectorsImmediately" has taken the default value or has been configured to false. We don't want to unilaterally set this to true if the user has configured it to false. This would lead to confusing behaviour and would correctly be described as a bug.
> 
> Could you change the default to true. Or better still reverse the meaning to "deferOutboundConnectionsUntilServerStarted". The current problem is that the channel framework is blocking outbound connections when it did not need to. The users of the channel framework can decide when they wish to make a connection and the only need for a parameter is to allow the possibility of restoring the old behaviour.

@AndrewDJBanks  Setting this to true would be a direct change to behavior, which we shouldn't do. From the channel's point of view this has to be guarded with a property, defaulted to the same behavior as it is today. Users creating channels can specify to use the new behavior through the custom property, or keep the old behavior by setting the property to false or by not defining it. 

Perhaps SIB can have its own property that users can set when they desire outbound channels to start with this new property?

mrsaldana:
#build

mrsaldana:
Based on discussion from POC meeting, this has been adjusted to be default behavior for outbound selectors.

mrsaldana:
#build

joe-chacko (approved):
lgtm

joe-chacko:
#run-libby-bot

AndrewDJBanks (approved):

joe-chacko:
Also fixes #19756

PR #19492, opened 2021-11-29
Move additional test to krb5.2 fat bucket

kristip17:
First bucket still occasionally timing out on Windows. Bump a test to krb5.2, after splitting the krb5 fat bucket due to slow Windows performance.

RTC 287932


kristip17:
#build

ryanesch (approved):

PR #19512, opened 2021-11-30
Clean up and add trace for acmeCA FATs

kristip17:
Clean up and add some trace for acmeCA FATs. We hit a possible HTTP port not open issue in a build, but it seems like the test continued fine anyway. Added some more trace for better debug and timestamps if it continues to happen.

RTC 287937


kristip17:
#build

ryanesch (approved):

PR #19515, opened 2021-11-30
Make tests wait for Discovery to complete and tokenEndpoint app to load

c00crane:
We have tests that are failing because they try to access a protected app and the authoriation endpoint has not been discovered by the time we try to use it.  We have a single attempt in the class startup method and I'm adding a retry loop to that.  I've also changed that startup code to not result in a failure if those multiple attempts fail - if there is an issue with discovery, the tests will fail (there is time between the end of setup and the first test - that may give it more time to complete).
We also have a test that is failing because we're not waiting for the TokenEndpointServlet in the test class setup.  When the first test runs, it tries to create and save a token and we can not because the app isn't ready.  When the runtime tries to retrieve the token, it gets a null and the test case fails.  The app is finally up by the end of that test, so all remaining tests succeed.


c00crane:
#run-libby-bot
#build

teddyjtorres (approved):

PR #19518, opened 2021-11-30
Code is incorrectly expecting Void.class to identify void return type

njr-11:
Code has a bug where it is expecting Void.class to identify void return type on a method.  The way to refer to void.class in Java is with the Void.TYPE constant, which ought to be used instead.


njr-11:
#build

KyleAure (approved):

njr-11:
Build has 1 test failure, which is a duplicate of a pre-existing issue in an unrelated test bucket

PR #19535, opened 2021-12-01
Refactor our consumption of Swagger UI

### LINKED ISSUE #19325: Streamline consumption of new versions of Swagger UI
Many of our OpenAPI related features include a customized version of [Swagger UI][swagger-ui] which provides a web interface for viewing and browsing the OpenAPI documentation we generate.

Our customizations directly modify the Swagger UI code, so updating to a new version requires checking out the new version of swagger UI, manually re-applying all of our changes, building and then checking in the built css and js files.

We need to find a way to make updating Swagger UI easier, as this is something we should need to do regularly to pick up bug fixes and security patches.

Ideally we would restrict our changes to just use the [documented extension mechanisms][extensions] so that we can depend on Swagger UI without needing to make any changes to its code.
<br/><br/><br/>  

[swagger-ui]: https://github.com/swagger-api/swagger-ui
[extensions]: https://github.com/swagger-api/swagger-ui/blob/master/docs/SUMMARY.md#customization

Azquelt:
Switch from including the entire swagger UI source (with modifications) to pulling in Swagger UI as a dependency and making our modifications as extensions on top of it.

This should make it much easier to consume new versions of Swagger UI.

Fixes #19325


Azquelt:
I note that libby has flagged the following:
> WARNING: This pull request contains binary files, which should be managed in a Maven repository.

The only binary file I can see is a .png used as the favicon, which I think is ok.

However, this PR also contains very large minified css, js and map files which, though not binary, are likely to change in ways that don't create nice diffs when we consume new versions of swagger UI. This is no worse than the currently checked in code, but we should check we're still ok with this approach.

The only alternative I can see would be to put the code in a separate repository and upload the built js, css and map files to an artifact repository and pull those files down during the open liberty build.

Azquelt:
Pushing a rebase onto the current integration branch and a couple of minor changes.

Azquelt:
#libby
#build

Azquelt:
Well noted @LibbyBot, however those are files which are being _removed_ in this PR...

Azquelt:
#build

Emily-Jiang (commented):
LGTM! Thanks @Azquelt !

Emily-Jiang (approved):

Azquelt:
#libby

Failures in the build are known issues.

PR #19539, opened 2021-12-01
Transform classes found in shared class cache

### LINKED ISSUE #19026: Improve application class sharing for dynamic classes
The server does not transform application classes found in the shared class cache (SCC) nor does it store transformed classes to the cache.  That is, the server does not cache dynamic application classes.  This behavior is problematic whenever a server must transform an application class, but does not because the class had been unexpectedly cached by another server process.  To resolve this problem the user must stop all server processes attached to the SCC and then destroy the SCC -- OR -- disable application class sharing.  Problem determination typically requires two support interactions.  And though the problem does not occur frequently, it seems something we might avoid altogether.

We could change the application classloader's caching behavior to prevent this problem.  For instance, the loader could transform a class found in the cache, and if the cached and transformed bytes differ, the loader would read the class bytes from the classspath, transform it, and define it.  Another change could add intelligence to the transformer(s) that indicates a class requires transformation. And again, if a class requires transformation and is found in the cache, the classloader would read the class bytes from the classpath, transform it, and define it.

dazavala:
For issue #19026

Modify the AppClassLoader to transform classes found in the shared class cache and update the classloading unittest to verify the new behavior.

### DIFF
--- a/dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/AppClassLoader.java
+++ b/dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/AppClassLoader.java
@@ -313,37 +313,37 @@ public class AppClassLoader extends ContainerClassLoader implements SpringLoader
     }
 
     byte[] transformClassBytes(String name, ByteResourceInformation toTransform) throws ClassNotFoundException {
-        // If a class is loaded from the shared classes cache it cannot be transformed.  The bytes are not in 
-        // the normal class bytes format, but rather a cookie to where the bytes are stored in the classes cache.
-        // Since the class was put into the classes cache, it was not transformed previously.  We are
-        // inferring it will not be transformed this time either.  If there is a change requiring the class
-        // to be transformed the user will need to delete the shared classes cache.  If we do not make this
-        // inference, the transform code will fail to process the class bytes resulting in a number of different
-        // types of exceptions or quitting out of the transformation code which will result in the class not being
-        // transformed anyway.
-        byte[] bytes;
-        if (toTransform.foundInClassCache()) {
-            if (systemTransformers.isEmpty()) {
-                bytes = toTransform.getBytes(); 
-                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                    Tr.debug(tc, "Not transforming class " + name + " because it was found in the shared classes cache.");
-                }
-            } else {
-                if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
-                    Tr.debug(tc, "Attempt to transform " + name + " from shared class cache because we have system tranformers");
-                }
-                try {
-                    bytes = toTransform.getActualBytes();
-                    bytes = transformClassBytes(bytes, name, systemTransformers);
-                } catch (IOException e) {
-                    // auto FFDC
-                    // fallback to the cached class
-                    bytes = toTransform.getBytes();
-                }
-            }
+        byte[] originalBytes;
+        boolean fromSCC = toTransform.foundInClassCache();
+        if (!fromSCC) {
+            originalBytes = toTransform.getBytes();
         } else {
-            bytes = transformClassBytes(toTransform.getBytes(), name, transformers);
-            bytes = transformClassBytes(bytes, name, systemTransformers);
+            // Since the class was stored to the shared class cache (SCC), it was not transformed previously.
+            // Transforming the actual class bytes ensures the expected bytes and alleviates the need for the
+            // user to delete the SCC when there is a change requiring the class be transformed.
+            // By not transforming bytes loaded from the SCC we avoid numerous types of exceptions or quitting
+            // out of the transformation code, which will result in the class not being transformed anyway.
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "Attempt to transform \" + name + \" found in shared class cache because we have "
+                         + (!transformers.isEmpty() ? "[transformers]" : "")
+                         + (!systemTransformers.isEmpty() ? "[system transformers]" : ""));
+            }
+            try {
+                originalBytes = toTransform.getActualBytes();
+            } catch (IOException e) {
+                // auto FFDC
+                // fallback to the cached class
+                return toTransform.getBytes();
+            }
+        }
+        byte[] bytes = transformClassBytes(originalBytes, name, transformers);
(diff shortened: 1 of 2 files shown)


dazavala:
The new solution simplifies the transform path because now, since all transformer types must execute upon finding a class in the cache.  Just run the original bytes through all transformers.  Checking for empty transformer lists is offset by the near calls to transformClassBytes(cfts), which return immediately when the transformer list _cfts_ is empty.  So reducing the logic seemed appropriate.  These checks appear in the trace output for serviceability.

dazavala:
#build

dazavala:
#build

tjwatson (approved):

dazavala:
Here are @jhanders34's review concerns and our conclusions for each.
  
Based off of the new code we lose the benefits of the SCC completely when a transformer is involved.
1 - There seems like little reason to use the SCC with transformers. 
2 - There isn't any logic to say if the transformer didn't change anything to just use the SCC version. 

>Issue 2. is the overall issue with 1.  **The solution must be changed to return the original SCC bytes if the transform didn't change anything.  This will require a some clever refactoring:**  

```
byte[] originalBytes;
boolean fromSCC = toTransform.foundInClassCache(); 
if (!fromSCC) {
    originalBytes = toTransform.getBytes();
} else {
        Tr.debug(tc, "Attempt to transform \" + name + \" found in shared class cache because we have "
                 + (!transformers.isEmpty() ? "[transformers]" : "")
                 + (!systemTransformers.isEmpty() ? "[system transformers]" : ""));
    }
    try {
        originalBytes = toTransform.getActualBytes();
    } catch (IOException e) {
        // auto FFDC
        // fallback to the cached class
        return toTransform.getBytes();
    }
}
byte[] bytes = transformClassBytes(originalBytes, name, transformers);
bytes = transformClassBytes(bytes, name, systemTransformers);
if (fromSCC) {
    // if the transform didn't change the bytes, then return the original bytes
    // returned from the shared classes cache.
    if (bytes == originalBytes || Arrays.equals(bytes, originalBytes)) {
        bytes = toTransform.getBytes();
    }
}
return bytes;
```

3 - It really makes me question if we should store the transformed version of the class now with this newer logic.

>Determined moot.  Initial thought was, if we always call the transformer there is no reason not to store the transformed bytes to SCC, unless there is concern that the transformer could change.  And as such, we would not want the results from the previous transformation when there is a new transformer.  But ultimately it doesn't matter, because even though a class is found in the SCC, the classloader always calls the transformer after this change.   We will never store a transformed version. 

4 - Since we are getting the actual bytes now, is there a reason to have the try / catch any longer for the transform when gotten from the SCC.

>We need the try-catch for the `IOException` that `ByteResourceInformation.getActualBytes()` may throw.  There's no reason to modify getActualBytes() to eat the exception and instead return the cached bytes (i.e. return `ByteResourceInformation.getBytes())`, so leave the solution as-is in this regard.  **But, the solution can be factored/simplified to first determine which bytes to transform, and then make a single call to the transformers over those bytes.  (done)**

5 - Send the classloading bundle to @gjdeval in order to benchmark these changes.

Q: Do you know why Tom decided to call the system transformers after the other transformers?
I do not.  If there's a rationale here, let's ask @tjwatson. 

Q: Will there be a test added for the JPA scenario itself?
Yes, I can do that.  I consulted @jgrassel for advice regarding which JPA FAT project is appropriate for this test, and I'll consult with him further as I begin the task.

#build

jhanders34 (approved):

tjwatson:
> Q: Do you know why Tom decided to call the system transformers after the other transformers?
> I do not. If there's a rationale here, let's ask @tjwatson.

Because checkpoint needs to be run after any other transformers that also may add `<clinit>` code.  Checkpoint needs to be inserted as the very first instruction run curing `<clinit>` to ensure no user code is run before we checkpoint on `deployment`.  Running global transformers last allows us to rank our checkpoint transformer such that we can have the most likely chance of running last to avoid other transformers from inserting instructions before the checkpoint.

tjwatson:
> Determined moot. Initial thought was, if we always call the transformer there is no reason not to store the transformed bytes to SCC, unless there is concern that the transformer could change. And as such, we would not want the results from the previous transformation when there is a new transformer. But ultimately it doesn't matter because, even though a class is found in the SCC, the classloader always calls the transformer after this change. We will never store a transformed version.

I think there would still be benefits if we could figure out how to store transformed classes.  The act of defining a new class after the IO to read the class still takes significant time vs getting the existing class from SCC.  But there are many concerns with storing transformed classes because we have no easy way, that I can tell, of knowing the output of the transformers each run for a given class will be the same as the previous time.  We would need to:

1. Be able to determine the exact list (including order) of transformers is present that were present when we stored the transformed class in the SCC.  If the list is different then we have to transform the class again and define the transformed bytes to get the class.
2. If the list of transformers is the same then can we make the assumption that such transformers perform consistent results for a given class?  If this assumption is true (have my doubts) then we could skip calling the transformers and simply load the SCC version of the transformed class.
3. If assumption 2 is not safe then we have no choice but to transform again.  But here we have no way to know if the resulting bytes is the same as the transformed bytes that were used when we stored the transformed class in the SCC from a previous run.

To do 1 and 2 above will require a bit of work, likely need to introduce some marker interface that we can use to identify "consistent" transformers.  This interface could have a default method that would return a unique ID for the transformer.  Something like this (assuming the consistent transformers are all from our own bundles)

```java
default String getTransformerID() {
    Class<?> clazz = this.getClass();
    Bundle b = FrameworkUtil.getBundle(clazz);
    if (b == null) {
        return null;
    }
    return clazz.getName() + ";" + b.getSymbolicName() + ";" b.getVersion();
}
```

If the same transformer implementation is added multiple times to the same class loader then they would have to override the getTransformerID to differentiate each instance of the transformer.  If all this is possible we would still have to figure out a way to store the transformer ID list for each transformed class so that we can lookup the ID list on subsequent loads of that class.

If we have to do 3) then it is even more complicated and I have my doubts that it would end with any benefits for performance.  Only way it is feasible is if the SCC has some way to get the byte[] of the class stored in the SCC so that we can compare the array with the byte[] we just transformed again.

jgrassel:
I'm not sure if we want to retain transformed classes.  There is always the chance that an application might be reconfigured to use a different JPA persistence provider (switching from Hibernate to Eclipselink for example), and there is always the chance that an update to the same persistence provider impl might change the bytecode enhancer in a way which the updated JPA provider impl is dependent on.  Apache OpenJPA actually embeds the version of OpenJPA which enhanced the class, and emits a warning when that version no longer matches the current active implementation of OpenJPA.

dazavala:
Perf benchmark results are in and look good.  Time to merge...
#libby

PR #19548, opened 2021-12-03
OpenAPI FAT: Set read timeout

Azquelt:
Set the read timeout on the connection when downloading the OpenAPI
document. Hopefully this will prevent test timeouts and allow us to
debug further why we occasionally see this fail.

Also stop swallowing the real exception :angry: 

For RTC283961

### DIFF
--- a/dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/utils/OpenAPIConnection.java
+++ b/dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/utils/OpenAPIConnection.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2018 IBM Corporation and others.
+ * Copyright (c) 2017, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


Azquelt:
#build

Emily-Jiang (approved):

dev/com.ibm.ws.microprofile.openapi_fat/fat/src/com/ibm/ws/microprofile/openapi/fat/utils/OpenAPIConnection.java
```diff
@@ -136,6 +136,7 @@ public OpenAPI downloadModel() throws Exception {
     public HttpURLConnection getConnection() throws IOException, ProtocolException {
         HttpURLConnection conn = HttpUtils.getHttpConnection(constructUrl(), expectedResponseCode, null, 30, method,
             headers, streamToWrite);
+        conn.setReadTimeout(30 * 1000);
```
tevans78, line 139: Are these two "30" seconds supposed to be the same thing? Shoud it be a constant? Should the ReadTimeout be set inside HttpUtils at the same time as the ConnectTimeout?

Azquelt: They're similar but not the same thing. The connect timeout is how long you wait for the connection to be estabilshed whereas the read timeout is how long you wait for the server to say something after the application tries to read from the connection.

For the OpenAPI tests, we always expect a prompt response so it's safe to set a read timeout (and hopefully this will result in a more useful failure than the whole test bucket timing out). The 30 seconds for each timeout is somewhat arbitrary, we don't expect the timeout to ever be hit, it's just a failsafe so that we won't wait forever.

I didn't want to change `HttpUtils` since that's a global test class and we have tests which expect the server to take a long time to respond (e.g. `FATServletClient`) so setting a read timeout would not be appropriate.

PR #19549, opened 2021-12-03
OpenAPI: Cancel listener synchronously

Azquelt:
OpenAPI runs a scheduled task which processes already deployed modules.
If required services are not ready it repeats. When it succeeds, it
cancels itself.

The cancellation is done asynchronously which can cause problems if the
server is shutting down as the executor serivce will no longer be
available.

I think the easiest way to avoid the problem is to do the cancellation
synchronously instead, which doesn't require the executor service.

For RTC 281148 and RTC 280848


Azquelt:
#build

Emily-Jiang (approved):

PR #19598, opened 2021-12-08
Update string to search for in WS-Security SAML CXF test

ayoho:
For 287982. The code is looking for "verifySubjectConfirmationMethod A Bearer Assertion was not signed" in the trace, but a recent run shows that "fillCallerData A Bearer Assertion was not signed" is being emitted now. I believe this might be due to an open source project having been updated, but I'm not entirely sure yet. I can update the test to just look for "A Bearer Assertion was not signed" since that should suffice.


ayoho:
#build

c00crane (approved):

PR #19603, opened 2021-12-09
mpOpenAPI: Don't scan if scanning is disabled

### LINKED ISSUE #19585: Classes are still indexed by mpOpenAPI when mp.openapi.scan.disable=true
**Describe the bug**  

When the config option `mp.openapi.scan.disable=true`, application classes are still indexed using Jandex though the result is discarded and isn't used to generate the OpenAPI document.

This is unhelpful since scanning is a relatively slow operation.

**Steps to Reproduce**  

* Set `mp.openapi.scan.disable=true`
* Enable trace specification `MPOPENAPI=all`
* Note that lots of scanning takes place within `IndexUtils.getIndexView()`

**Expected behavior**  
Classes should not be scanned

**Diagnostic information:**  
 - OpenLiberty Version: 21.0.0.11 (and probably older)
 - Affected feature(s): mpOpenApi-2.0 and 3.0
 - Java Version: N/A

Azquelt:
Previously, if scanning was disabled we would scan but then discard the result.

Fixes #19585


Azquelt:
#build
#libby-bot

Azquelt:
#build

hutchig (approved):
Checked out OK.

Azquelt:
#libby

PR #19604, opened 2021-12-09
Use class loader of jdbcDriver libraryRef to load custom exceptions and data store helpers

njr-11:
Use the class loader from the libraryRef of the jdbcDriver to load custom exception and data store helpers.
Previously, it was obtaining the class loader from the data source implementation class, which ends up with different behavior on IBM i


njr-11:
#build

gjwatts (approved):

gjwatts:
FYI, the single failure was in an existing RTC defect 287799.

njr-11:
The personal build has just 1 test failure and it is an exact match for a pre-existing issue in an unrelated test bucket.

PR #19610, opened 2021-12-09
Enable user-provided interceptors to work with MP Rest Client 3.0

andymc12:
This PR partially resolves an issue in MP Rest Client 3.0 wrt to CDI interceptors.  Prior to this change user-provided CDI interceptors (i.e. interceptors in the user application)  enabled via the `@Priority` annotation and bound to MP Rest Client interface methods would be executed twice.  This change resolves this issue so that they are only executed once.

Note that interceptors bound to rest client interface methods that are enabled via the `beans.xml` file are not invoked at all like they were with previous versions of the MP Rest Client. 

While Liberty's CDI implementation does invoke user-provided interceptors when they are enabled via `@Priority`, CDI does not invoke MP Fault Tolerance interceptors (perhaps because those aren't annotated with `@Priority`??).  So this fix (somewhat hackishly) explicitly invokes MP FT interceptors via the Rest Client implementation. This means that potentially other Liberty-provided interceptors would not be invoked. This should be investigated to see if a better solution is possible.

### DIFF
--- a/dev/com.ibm.ws.microprofile.rest.client.FT_fat/test-applications/retryApp/src/mpRestClientFT/retry/Driver.java
+++ b/dev/com.ibm.ws.microprofile.rest.client.FT_fat/test-applications/retryApp/src/mpRestClientFT/retry/Driver.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2019, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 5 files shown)


andymc12:
#build

Azquelt (approved):
I'd have liked a nice comment pointing out that this is a horrible hack because RestEasy's interceptor handling doesn't see the interceptor binding added by fault tolerance, but otherwise this looks good.

WhiteCat22 (commented):
lgtm

andymc12:
#libby

andymc12:
Only failure was an unrelated ui test.

PR #19611, opened 2021-12-10
Reduce Oracle database helper `DSRA8207I` messages

### LINKED ISSUE #19608: Oracle database helper logging `DSRA8207I` too frequently
**Describe the bug**  

As a user of Liberty, JDBC, and the Oracle JDBC driver when I use third-party libraries or applications that are written in a general manner for non-specific database calls, I find that I often have an over-abundance of `DSRA8207I` messages in my messages.log file.  This is caused by the library or application inappropriately calling `con.setReadOnly(false)` which the oracle driver does not support. Since this is logged each time this method is called on a connection it could lead to the messages.log file increasing to an undesirable size. 

**Expected behavior**  
I would expect that this information message would be emitted once to notify the user that all method calls to `con.setReadOnly(false)` are ignored.  This would align with the existing behavior of `DSRA8770I` which is also logged at the same time. 

**Diagnostic information:**  
 - OpenLiberty Version: 21.0.0.12
 - Affected feature(s) [JDBC-4.x]
 - Java Version: All

**Current workaround:**
Currently, I have to work around this issue by ignoring these messages using:
`com.ibm.ws.logging.hideMessage=DSRA8207I`

KyleAure:
Fixes #19608

### DIFF
--- a/dev/com.ibm.ws.jdbc_fat_oracle/fat/src/com/ibm/ws/jdbc/fat/oracle/OracleTraceTest.java
+++ b/dev/com.ibm.ws.jdbc_fat_oracle/fat/src/com/ibm/ws/jdbc/fat/oracle/OracleTraceTest.java
@@ -10,8 +10,10 @@
  *******************************************************************************/
 package com.ibm.ws.jdbc.fat.oracle;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
+import java.util.List;
 import java.util.regex.Pattern;
 
 import org.junit.AfterClass;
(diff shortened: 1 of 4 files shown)


dev/com.ibm.ws.jdbc_fat_oracle/fat/src/com/ibm/ws/jdbc/fat/oracle/OracleTraceTest.java
```diff
@@ -88,4 +90,12 @@ public void testConnectionPropsEmbeddedPasswordNotLogged() throws Exception {
         if (sslPasswords > 0)
             fail("SSL passwords logged in trace " + sslPasswords + " times. SSL Password: " + SSL_PASSWORD);
     }
+
+    @Test
+    public void testReadOnlyInfo() throws Exception {
+        server.setTraceMarkToEndOfDefaultTrace();
+        runTest(server, JEE_APP + "/" + SERVLET_NAME, getTestMethodSimpleName());
+        List<String> occurrences = server.findStringsInLogsAndTraceUsingMark("DSRA8207I");
+        assertEquals("The DSRA8207I message should have only been logged exactly once in log, and once in trace.", 2, occurrences.size());
```
njr-11: The test should check messages.log only (I think there is a similar findStringsInLogsUsingMark method for this) so that we are only testing the externals.  I'd rather not have tests that are vulnerable to changes in trace and it also costs more time to scan through traces.

njr-11 (approved):

KyleAure:
#build

KyleAure:
#libbybot

KyleAure:
#build

KyleAure:
The build was successful and the reported errors were due to existing errors in the find-bugs part of the build.

PR #19634, opened 2021-12-14
Fail to load forbidden classes

### LINKED ISSUE #19630: Block Loading of Classes with Known Vulnerabilities (Apache log4j “JndiLookup”).
Applications deployed to Liberty may run versions of Log4j2 that are affected by the Log4Shell (CVE-2021-44228) and related vulnerabilities.

Issue 19630 modifies the application and library class loaders to block the loading of class org.apache.logging.log4j.core.lookup.JndiLookup, which is the cause of the vulnerability.

Users should analyze their applications for use of Log4j2 with urgency; in the meantime this fix may help mitigate Log4Shell and other vulnerabilities related to class org.apache.logging.log4j.core.lookup.JndiLookup. This issue will not protect in cases where the Log4j2 classes have been renamed (a process known as "shading") or if Log4j2 is loaded by other class loaders, such the Java system class loader or user-created class loaders.

tbitonti:
Add an internal properties file that lists forbidden classes that our application class loader must fail to load.

Fixes #19630.


tbitonti:
#build

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/AppClassLoader.java
```diff
@@ -71,11 +74,63 @@
...
+        }
+        
         ClassLoader.registerAsParallelCapable();
     }
-    static final TraceComponent tc = Tr.register(AppClassLoader.class);
 
+    private static final String FORBIDDEN_PROPERTIES = "forbidden.properties";
+
+    private static Properties loadForbidden() {
+        InputStream inputStream = AppClassLoader.class.getResourceAsStream(FORBIDDEN_PROPERTIES);
```
tjwatson: Use try-with-resource/finally here to manage the inputstream closing

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/AppClassLoader.java
```diff
@@ -544,6 +599,19 @@ final ByteResourceInformation findClassBytes(String className, String resourceNa
...
+        if ( forbiddenClasses.contains(name) ) {
+            if ( TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled() ) {
+                Tr.debug(tc, "loadClass " + name + " forbidden");
+            }
+            // Does this need to use the 'suggestion' mechanism?
+            // Since this is not a usual class-not-found case,
+            // I'm having it directly throw a simple CNFE.  We
+            // may want to change that to indicate that the class
+            // is designated as to be ignored.
+            throw new ClassNotFoundException(name);
```
tjwatson, line 603: As the comment suggests, we could give a more detailed error message.  In most cases the JVM is going to catch the CNFE, forget the message and then throw a NoClassDefFoundError with only the class name, so having a more detailed message gives little value besides Class.forName and direct ClassLoader.loadClass calls by the applications.

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/AppClassLoader.java
```diff
@@ -71,11 +74,63 @@
...
 public class AppClassLoader extends ContainerClassLoader implements SpringLoader {
+    static final TraceComponent tc = Tr.register(AppClassLoader.class);
+
+    private static final Set<String> forbiddenClasses = new HashSet<>();
+
     static {
+        // Per CVE XXX, Apache log4j class "JndiLookup" is vulnerable
+        // before package version 2.15.0.
+        
+        String jndiLookupClassName = "org.apache.logging.log4j.core.lookup.JndiLookup";
```
tjwatson: Why is this not contained in the properties file we include?

dev/com.ibm.ws.classloading/src/com/ibm/ws/classloading/internal/AppClassLoader.java
```diff
@@ -71,11 +74,63 @@
  * to discover the special methods:
  */
 public class AppClassLoader extends ContainerClassLoader implements SpringLoader {
+    static final TraceComponent tc = Tr.register(AppClassLoader.class);
+
+    private static final Set<String> forbiddenClasses = new HashSet<>();
```
tjwatson: Make this an unmodifiable set in the static block below instead of assigning it to a modifiable set here.

tjwatson (changes requested):
I see little value in having some hard-coded classes directly in the `.java` code and in the `.properties` file contained in the bundle.  In both cases the class name is hard coded and the classloading bundle must be rebuilt to update the forbidden class set.

Is the design to not have a `.properties` file in the classloading bundle?  Instead have a hard-coded set in the `.java` code and be able to load the `.properties` file from a fragment of classloading?  I have my doubts that is the intention.  If so then I suggest we need more comments in the code indicating the intent.  I also suggest that `java.lang.ClassLoader.getResources(String)` is used to allow multiple fragments to supply the set, otherwise only the "first" fragment will win which is unpredictable.

tbitonti:
#build

tjwatson:
#build
#libby

tjwatson (approved):
I approve, but should be reviewed by another for the changes I made.  We have a good build, but we should add a test.  I'll work on that now, but we could consider merging this now I think.

jaridk (approved):

NottyCode (approved):

PR #19645, opened 2021-12-14
openapi: cancel module listener job on shutdown

Azquelt:
When the OpenAPIModuleListener component is deactivated, it should
cancel its scheduled task if it's still running.

For RTC 274019


Azquelt:
#build

tevans78 (approved):

PR #19649, opened 2021-12-14
featureUtility fat test clean up 

jjiwooLim:
- testInvalidMultiVersionFeatures() had intermittent connection time out issue. Added a property to use local repo so that it doesn't connect to Maven central repo. 
- switched the order of assertion so that exit codes are checked later. This will help us see the error we are getting. 
- reintroduced testAlreadyInstalledFeatures() for installServerFeatureTest. It should exit with 0 even if the features are already installed. 

Things that can be improved/cleaned up:
- output.contains("[feature]") is probably not a good way to check whether the features are installed properly. We could check /lib/features/[feature].mf file 
- instead of copying json and esa file for each test case, copy once before class (if possible)


jjiwooLim:
#build

jjiwooLim:
#build

jacobwdv (approved):
looks good!

jjiwooLim:
#libby-bot

Issue #15979, opened 2021-02-18
Enable JPA bind literal as parameter at a server scope


NottyCode:
## Issue
Enabling EclipseLink to bind a literal as a parameter requires a `persistence.xml` to be updated:

```xml
    <persistence-unit name="TestPU" transaction-type="JTA">
        <jta-data-source>jdbc/jpaDataSource</jta-data-source>  
        <properties>
            <property name="eclipselink.target-database-properties" value="shouldBindLiterals=true"/>
        </properties>
    </persistence-unit>
```

There may be many `persistence.xml` files, spread across several applications, in a Liberty server. This makes it very difficult to add a persistence property to all `persistence.xml` files; requiring repackaging/rebuilding of possibly hundreds of applications. It would be good to be able to configure this for all `persistence.xmls` in the server.

## Description of the high level feature, including any external spec links: 

This feature will enable the configuration of default persistence properties in server.xml, using the existing JPA container configuration `<jpa>`. The new `defaultProperties` format will follow the format used in `persistence.xml` as follows:

```xml
    <jpa>
        <defaultProperties>
            <property name="javax.persistence.lock.timeout" value="10000"/>
            <property name="eclipselink.target-database-properties" value="shouldBindLiterals=true"/>
        </defaultProperties>
    </jpa>
```

The new properties will be described as:

> Default, integration-level persistence properties for use by the persistence provider when a container managed entity manager factory is created. Specified persistence properties will be included when the PersistenceProvider.createContainerEntityManagerFactory method is called.

According to the JPA specification (3.4.4.3):
```
If the same property or hint is specified more than once, the following order of overriding applies, in
order of decreasing precedence:
    • argument to method of EntityManager, Query, or TypedQuery interface
    • specification to NamedQuery (annotation or XML)
    • argument to createEntityManagerFactory method
    • specification in persistence.xml
```

This means these `defaultProperties` values will override the same properties specified in `persistence.xml`. However, property values specified through PersistenceContext annotation, or the _persistence-context-ref_ deployment descriptor element, will override these `defaultProperties`.

Basically, the new `defaultProperties` provide a mechanism to apply a persistence property to all `EntityManager` instances that do not currently have the property specifically configured.

In the event a default property is configured incorrectly (empty name or value), the following warning will be logged and the property ignored:

> CWWJP0056W: A JPA default property configured in the server.xml file contains an empty name or empty value: name = [No Value], value = []. The property will be ignored.

[...]

PR #19658, opened 2021-12-16
Improve JPA config trace and obfuscate passwords

dazey3:
for #15979

### DIFF
--- a/dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
+++ b/dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
@@ -308,9 +308,14 @@ ERROR_TRANSFORMING_CLASS_CWWJP0055E.useraction=The class cache might be corrupte
 
 #-------------------------------------------------------------------------------
 # 0=property name, 1=property value
-EMPTY_PROP_NAME_VALUE_CWWJP0056W=CWWJP0056W: A JPA default property configured in the server.xml file contains an empty name or empty value: name = [{0}], value = [{1}]. The property will be ignored.
+EMPTY_PROP_NAME_VALUE_CWWJP0056W=CWWJP0056W: A JPA default property that contains an empty name or empty value is configured in the server.xml file: name = [{0}], value = [{1}]. This property is ignored.
 EMPTY_PROP_NAME_VALUE_CWWJP0056W.explanation=The name and value attributes for a JPA default property must not be empty.
 EMPTY_PROP_NAME_VALUE_CWWJP0056W.useraction=Change the server configuration to provide a non-empty name and value for the property or remove the property.
+
+# 0=property name, 1=property value
+JDBC_PROP_NAME_CWWJP0057W=CWWJP0057W: A JPA default property that is intended for use only in Java SE environments is configured in the server.xml file: name = [{0}], value = [{1}]. This property is ignored.
+JDBC_PROP_NAME_CWWJP0057W.explanation=This JPA default property is being used in a Java EE environment, but is intended for use only in Java SE environments.
+JDBC_PROP_NAME_CWWJP0057W.useraction=Change the server and application configuration to use a JNDI data source and remove the property.
 
 #-------------------------------------------------------------------------------
 # Error messages from the JPA Provider below this point
--- a/dev/com.ibm.ws.jpa.container/src/com/ibm/ws/jpa/container/osgi/internal/JPAComponentImpl.java
+++ b/dev/com.ibm.ws.jpa.container/src/com/ibm/ws/jpa/container/osgi/internal/JPAComponentImpl.java
@@ -46,9 +46,11 @@ import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
 
 import com.ibm.websphere.csi.J2EEName;
+import com.ibm.websphere.ras.ProtectedString;
 import com.ibm.websphere.ras.Tr;
 import com.ibm.websphere.ras.TrConfigurator;
 import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.websphere.ras.annotation.Sensitive;
 import com.ibm.websphere.ras.annotation.Trivial;
 import com.ibm.ws.Transaction.UOWCurrent;
 import com.ibm.ws.container.service.app.deploy.ApplicationClassesContainerInfo;
@@ -89,6 +91,7 @@ import com.ibm.wsspi.classloading.ClassLoadingService;
 import com.ibm.wsspi.kernel.service.utils.AtomicServiceReference;
 import com.ibm.wsspi.kernel.service.utils.ConcurrentServiceReferenceSet;
 import com.ibm.wsspi.kernel.service.utils.FrameworkState;
+import com.ibm.wsspi.kernel.service.utils.SerializableProtectedString;
 import com.ibm.wsspi.kernel.service.utils.ServiceAndServiceReferencePair;
 import com.ibm.wsspi.logging.Introspector;
 import com.ibm.wsspi.resource.ResourceBindingListener;
@@ -205,6 +208,7 @@ public class JPAComponentImpl extends AbstractJPAComponent implements Applicatio
         }
     }
 
+    @Sensitive
     private void setDefaultProperties(Dictionary<String, Object> properties) {
         // Look for default integration-level properties
         Map<String, String> dProperties = new HashMap<String, String>();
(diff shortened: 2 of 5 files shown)


dazey3:
#build

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
```diff
@@ -312,6 +312,11 @@ EMPTY_PROP_NAME_VALUE_CWWJP0056W=CWWJP0056W: A JPA default property configured i
 EMPTY_PROP_NAME_VALUE_CWWJP0056W.explanation=The name and value attributes for a JPA default property must not be empty.
 EMPTY_PROP_NAME_VALUE_CWWJP0056W.useraction=Change the server configuration to provide a non-empty name and value for the property or remove the property.
 
+# 0=property name, 1=property value
+JDBC_PROP_NAME_CWWJP0057W=CWWJP0057W: A JPA default property configured in the server.xml file is not recommended to be used here: name = [{0}], value = [{1}].
```
tkburroughs: "here" seems like it could be confusing; perhaps "is not recommended to be used in a EE server:" but perhaps wait for feedback from ID before changing.

dmuelle: - in general, we're advised to avoid making recommendations- can we instead just tell them not to use it?
- what does it mean that a "default" property is configured? Isn't the default what you get when you don't specify configuration? Also, the message calls it a "JPA default property" and the explanation says "JDBC property". If they're the same thing it's better to use one consistent term.

suggested edit:
 The following JDBC property, which is configured in the server.xml file, is not intended for Java or Jakarta EE applications: name = [{0}], value = [{1}].

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
```diff
@@ -312,6 +312,11 @@ EMPTY_PROP_NAME_VALUE_CWWJP0056W=CWWJP0056W: A JPA default property configured i
...
+JDBC_PROP_NAME_CWWJP0057W.explanation=This JDBC property is not recommended for EE applications and should only be used in JSE environments.
```
dmuelle: IBM style guidance is to avoid modal "should"

suggested edit:
This JDBC property is not intended for Java or Jakarta EE applications and must be used only in Java SE environments.

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
```diff
@@ -312,6 +312,11 @@ EMPTY_PROP_NAME_VALUE_CWWJP0056W=CWWJP0056W: A JPA default property configured i
...
+JDBC_PROP_NAME_CWWJP0057W.useraction=Change the server/application configuration to use a JNDI data source and remove the property.
```
dmuelle: Change the server or application configuration to use a JNDI data source and remove the property.

dmuelle (changes requested):
ID review- let me know if you have any questions. Once the comments are addressed, I'll add the ID Reviewed label

dev/com.ibm.ws.jpa.container/src/com/ibm/ws/jpa/container/osgi/internal/JPAComponentImpl.java
```diff
@@ -205,19 +208,37 @@ protected void modified(Map<?, ?> newProperties) {
...
             if (name != null) {
-                String value = (String) properties.get(DEFAULT_PROPS_PREFIX + defaultPropertiesIndex + DEFAULT_PROPS_VALUE);
-                if (name.length() > 0 && value != null && value.length() > 0) {
-                    dProperties.put(name, value);
+                // See metatype: 'ibm:obscure="true"'
+                SerializableProtectedString value = (SerializableProtectedString) properties.get(DEFAULT_PROPS_PREFIX + defaultPropertiesIndex + DEFAULT_PROPS_VALUE);
+
+                if (value != null) {
+                    if (!name.isEmpty() && !value.isEmpty()) {
+                        // Convert to java.lang.String to pass the value to the persistence providers
```
tkburroughs, line 224: Did you consider leaving the passwords, or possibly all values as SerializableProtectedString, and only convert them when they are passed to the JPA provider, and/or only converting the non-passwords for trace?

Converting things later would be less efficient, but potentially safer.  I was also wondering, should we be protecting this list from modification by the JPA provider?  For example, what if the JPA provider added more properties to the list, or changed the list values, then all subsequent EMF creates would see the modified list, and not the original.  This could be a good reason to create a new list to pass on every create, and then convert from protected string at that time.

tkburroughs (commented):
I think what you have is fine; pending ID approval; so my comments are just some thoughts to consider.

tkburroughs (approved):

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
dazey3: > what does it mean that a "default" property is configured?

Users configure "default" JPA properties in their server.xml configuration. These default properties become the defaults for all  container managed persistence units in all applications running on that server

Here is an example of a JPA JDBC property:
```xml
<server>
    <featureManager>
        ...
    </featureManager>

    <jpa>
        <defaultProperties>
            <property name="javax.persistence.jdbc.url" value="jdbc:mysql://localhost/testDB" />
        </defaultProperties>
    </jpa>
</server>
```
If a customer uses one of the `javax.persistence.jdbc.*` properties (referred to as JDBC properties), then we will log warning `CWWJP0057W` to inform them that this property should not be used in this way.

> we're advised to avoid making recommendations- can we instead just tell them not to use it?

Ill have to discuss with my team if we can ignore these properties instead of just warn against their use. If so, I can update the message accordingly

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
dazey3: How about:
```
    This JDBC property is being used in a Java EE environment, but is intended for use in Java SE environments only
```

dev/com.ibm.ws.jpa.container/src/com/ibm/ws/jpa/container/osgi/internal/JPAComponentImpl.java
dazey3: > Converting things later would be less efficient, but potentially safer.

That was the major reason. I didnt want to have to iterate all the properties again to convert before sending to the provider.

>I was also wondering, should we be protecting this list from modification by the JPA provider?

I think it is protected. If you read through `com.ibm.ws.jpa.management.JPAPUnitInfo.createEMFactory`, we create a temporary Map of properties, then add all the integration/default properties to that Map, finally sending it to the provider via `javax.persistence.spi.PersistenceProvider.createContainerEntityManagerFactory(PersistenceUnitInfo arg0, Map arg1)`. If the provider modifies that Map in their implementation, thats fine because we are done with it after passing it in

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
dmuelle: looks good- slight edit to move the "only"

    This JDBC property is being used in a Java EE environment, but is intended for use only in Java SE environments.

dazey3:
@dmuelle I update the message incorporating your review comments. Let me know if there are any further refinements you believe would be beneficial. Thanks!

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
```diff
@@ -308,10 +308,15 @@ ERROR_TRANSFORMING_CLASS_CWWJP0055E.useraction=The class cache might be corrupte
 
 #-------------------------------------------------------------------------------
 # 0=property name, 1=property value
-EMPTY_PROP_NAME_VALUE_CWWJP0056W=CWWJP0056W: A JPA default property configured in the server.xml file contains an empty name or empty value: name = [{0}], value = [{1}]. The property will be ignored.
+EMPTY_PROP_NAME_VALUE_CWWJP0056W=CWWJP0056W: A JPA default property, configured in the server.xml file, contains an empty name or empty value: name = [{0}], value = [{1}]. The property will be ignored.
```
dmuelle: per IBM style, avoid future tense

The property will be ignored
--->
This property is ignored

dmuelle: slight edit suggestion

 A JPA default property that is configured in the server.xml file contains an empty name or empty value:

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
```diff
@@ -308,10 +308,15 @@ ERROR_TRANSFORMING_CLASS_CWWJP0055E.useraction=The class cache might be corrupte
...
 EMPTY_PROP_NAME_VALUE_CWWJP0056W.explanation=The name and value attributes for a JPA default property must not be empty.
 EMPTY_PROP_NAME_VALUE_CWWJP0056W.useraction=Change the server configuration to provide a non-empty name and value for the property or remove the property.
 
+# 0=property name, 1=property value
+JDBC_PROP_NAME_CWWJP0057W=CWWJP0057W: A JPA default property, configured in the server.xml file, is intended for use in Java SE environments only: name = [{0}], value = [{1}]. The property will be ignored.
```
dmuelle: misplaced "only"

A JPA default property that is intended for use only in Java SE environments is configured in the server.xml file : name = [{0}], value = [{1}]. This property is ignored.

dev/com.ibm.ws.jpa.container.core/resources/com/ibm/ws/jpa/jpa.nlsprops
```diff
@@ -308,10 +308,15 @@ ERROR_TRANSFORMING_CLASS_CWWJP0055E.useraction=The class cache might be corrupte
...
+JDBC_PROP_NAME_CWWJP0057W.explanation=This JPA default property is being used in a Java EE environment, but is intended for use in Java SE environments only.
```
dmuelle: ...for use in Java SE environments only
--->
...for use only in Java SE environments

dmuelle (changes requested):
looks good- a few more minor suggestions

dazey3:
#build

dmuelle (approved):
LGTM thanks!

dazey3:
#run-libby-bot

PR #19661, opened 2021-12-16
RM: fix kafka container version

Azquelt:
Requesting the kafka image from testcontainers results in a container
name with :latest.

We only want to move up to latest versions after testing them as they
may need changes in the test setup.

For RTC 288260


Azquelt:
#build

benjamin-confino (approved):

PR #19679, opened 2021-12-21
suppress printed 405 stack trace in messages.log for jaxrs-2.0 and jaxrs-2.1

WhiteCat22:
If the server throws an HTTP 405 from either the `jaxrs-2.0` or `jaxrs-2.1` (CXF) features it prints the error stack to the `messages.log` regardless of trace level.

We extend CXF's `WebApplicationExceptionMapper` with our own `CustomWebApplicationExceptionMapper`, so all I had to do was call `WebApplicationExceptionMapper.setPrintStackTrace(false);` to disable printing the stack trace without pulling in more overlay files.


WhiteCat22:
#build #run-libby-bot

jim-krueger (approved):

ayoho (approved):
Verified the change locally 👍

WhiteCat22:
#build

PR #19689, opened 2021-12-22
Remove any empty strings in hideMessageids

### LINKED ISSUE #19688: Empty com.ibm.ws.logging.hideMessage hides all messages and does not create messages.log
**Describe the bug**  
When using the bootstrap property value `com.ibm.ws.logging.hideMessage` and no value is provided a `messages.log` is not created upon start up of the server.

**Steps to Reproduce**  
Set `com.ibm.ws.logging.hideMessage=` in the `bootstrap.properties` file.

**Expected behavior**  
Since no value is provided, no messages should be hidden and a normal `messages.log` should be expected (and created).

**Diagnostic information:**  
 - OpenLiberty Version: [e.g. 21.0.0.8 - 21.0.0.10]
 - Affected feature(s) [logging]
 - Java Version: [n/a]

**Additional context**

Channyboy:
#build

fixes #19688 

The alternative place to fix this would have been in `LogProviderConfigImpl` where we parse the `config` passed in. The change would be to the method which parses in a `Collection` of values from `config`. However, not knowing if any other attribute (present or future) that may rely on an empty string (""), the fix is made in `BaseTraceService`.

### DIFF
--- a/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
+++ b/dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
@@ -33,6 +33,7 @@ import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import com.ibm.websphere.logging.WsLevel;
 import com.ibm.websphere.ras.Tr;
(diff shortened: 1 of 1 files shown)


Channyboy:
#build

Channyboy:
#build

Channyboy:
#build 
Failures scene in first two builds are related to a dependency upgrade on build machines for finding bugs.  The bugs found were not a result of this PR.
Reverting to original commit.

dev/com.ibm.ws.logging/src/com/ibm/ws/logging/internal/impl/BaseTraceService.java
```diff
@@ -319,7 +320,8 @@ public synchronized void update(LogProviderConfig config) {
         javaLangInstrument = trConfig.hasJavaLangInstrument();
         consoleLogLevel = trConfig.getConsoleLogLevel();
         copySystemStreams = trConfig.copySystemStreams();
-        hideMessageids = trConfig.getMessagesToHide();
+        //Remove any items in hideMessageids that are empty strings. Create a "new" list as original is backed by an array and cannot be removed.
+        hideMessageids = trConfig.getMessagesToHide().stream().filter(s -> !s.equals("")).map(s -> s).collect(Collectors.toList());
```
fmhwong: 1. I'd use `s.isEmpty()` since it is a bit faster than `s.equals("")`

1. `map(s -> s)` is not needed.

fmhwong (changes requested):
See comments

Channyboy:
#build

fmhwong (approved):
LGTM

Channyboy:
Errors/Warnings are related to original explanation made earlier where find bugs are triggered (for many components) due to a dependency change. Not Test failures.
#libby

PR #19690, opened 2021-12-23
MP Config: improve performance of ServerXML vars

### LINKED ISSUE #19624: Look into caching InternalConfigSource lookup results
Looking at profiles for some microprofile-primitives, I see some overhead on calls to getOptionalValue (coming from mpOpenTracing and mpJwt).
 
Some of the path is coming from here:
 ```
  Parent      0   0.02   2.10           1         101    J:io/openliberty/microprofile/config/internal/common/InternalConfigSource.getValue(Ljava/lang/String;)Ljava/lang/String;
 
    Self      0   0.02   2.10           1         101    J:io/openliberty/microprofile/config/internal/serverxml/ServerXMLVariableConfigSource.getProperties()Ljava/util/Map;
 
   Child      0   0.08   1.27           4          61    J:io/openliberty/microprofile/config/internal/serverxml/ServerXMLDefaultVariableConfigSource.getServerXMLVariables()Ljava/util/Map;
   Child      0   0.00   0.81           0          39    J:io/openliberty/microprofile/config/internal/serverxml/ServerXMLVariableConfigSource.getServerXMLVariables()Ljava/util/Map;
 ```
It would be great if we could cache the properties map in the two child methods above, and only regenerate it if something has changed, or if it's been 500 ms or something similar.

Azquelt:
Improve performance when finding properties defined as variables in
server.xml by eliminating copying of maps when retrieving variables.

Note this changes the semantics of
`ConfigVariables.getUserDefinedVariableDefaults`. Previously the default
value of a variable would not be returned if it was overridden with an
actual value, whereas now it is.

Fixes #19624

### DIFF
--- a/dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/variables/ConfigVariableRegistry.java
+++ b/dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/variables/ConfigVariableRegistry.java
@@ -64,6 +64,11 @@ public class ConfigVariableRegistry implements VariableRegistry, ConfigVariables
     // variables explicitly defined in bundle defaultInstances files
     private Map<String, LibertyVariable> defaultConfigVariables;
 
+    // Immutable map of name to value for user defined variables, replaced when updated, never modified
+    private volatile Map<String, String> userDefinedVariableMap;
+    // Immutable map of name to defaultValue for user defined variables, replaced when updated, never modified
+    private volatile Map<String, String> userDefinedVariableDefaultsMap;
+
     // cache of external variables
     private Map<String, Object> variableCache;
     // cache of variables defined in default configurations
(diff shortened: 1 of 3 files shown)


Azquelt:
#build
#spawn.fullfat.buckets=com.ibm.ws.microprofile.config.1.3_fat,com.ibm.ws.microprofile.config.1.4_fat,io.openliberty.microprofile.config.2.0.internal_fat,io.openliberty.microprofile.config.3.0.internal_fat,com.ibm.ws.config_fat,com.ibm.ws.config.variables_fat

tevans78 (approved):
From a MP Config point of view this is ok

brenthdaniel (approved):
Looks fine to me.. Normally I would be concerned about the change in behavior, but given that this has been around for years and MP config is still the only consumer it's OK.

jdmcclur:
The change looks good performance-wise.

Azquelt:
Thanks everyone for the reviews.

The failures in the build are known issues which are not related to this PR.

PR #19701, opened 2021-12-31
Add FFDCs for failover on UserEnumerationTest

kristip17:
Even when we successfully failover between remote LDAP servers, we log an FFDC (`javax.naming.CommunicationException`) which is fine, but gets flagged by the FAT infrastructure. The `UserEnumerationTest` has to run remotely, so will ignore Comm exceptions to avoid bugs when we've failed over during the test.

RTC 288470


kristip17:
#build

jvanhill (approved):

PR #19715, opened 2022-01-04
Wait for token and userinfo test endpoints to be ready

c00crane:
In order to test encryption and additional signature algorithm for OIDC and Social configs, we need token and userinfo endpoints that can handle them.  Some of the test classes were not waiting for the test apps that are providing that endpoint support to start.  In most cases, they start pretty quickly, but that's not always the case.  So, I'm adding checks (with embedded waits) for those test apps.


c00crane:
#build

c00crane:
#run-libby-bot
#build

ayoho (approved):

PR #19730, opened 2022-01-06
Renew Lease on TX leaseRenewer creation

### LINKED ISSUE #19729: Update Tx Peer recovery lease on creation of lease renewer
When the leaseRenewer is created we should update the lease as well as scheduling the next renewal.

neilgsyoung:
fixes #19729


neilgsyoung:
#build

jonhawkes (approved):

PR #19734, opened 2022-01-06
Remove check for shib_idp_session cookie on logout.

c00crane:
Remove the check for the shib_idp_session cookie after an IDP initiated logout since the SP does not handle cleaning it up.
Shibboleth controls the creation and removal of the shib_idp_session cookie - in most cases, it cleans it up, but not all.
This is an external service that we don't have control over, so, we'll stop checking for the cookie in this case.


c00crane:
#build

ayoho (approved):

PR #19737, opened 2022-01-06
test currentContextExecutor and getExecutionProperties

njr-11:
Add tests for the new currentContextExecutor method.
Also test what happens when a contextualized function created with one of the new API methods is supplied to the existing getExecutionProperties API method.  The spec JavaDoc indicates it must raise IllegalArgumentException.
Also, include an unrecognized context type under the configured `unchanged` list for `ContextServiceDefinition`. This should have no impact and not cause a failure.


njr-11:
#build

nmittles (approved):

PR #19744, opened 2022-01-07
Disable EE9 repeat on Windows in a couple long running OIDC Fats

c00crane:
We've run into some bucket timeouts with OIDC when run on Windows with EE9 enabled.  The FeatureReplacementAction process is taking in once case 55 out of the 180 allowed minutes.  The FAT ends up timing out.  I'm going to update the repeat rules for these long running projects to run without EE9 on Windows.  We have other FATs that are smaller that will test OIDC with EE9 on Windows and not time out.  The projects that are being updated will still run on Windows and will run with EE9 on all other platforms.

    public static RepeatTests repeat = RepeatTests.with(new EmptyAction().liteFATOnly())
            .andWith(new SecurityTestRepeatAction().onlyOnWindows().fullFATOnly())
            .andWith(new SecurityTestFeatureEE9RepeatAction().notOnWindows().alwaysAddFeature("servlet-5.0").fullFATOnly());


c00crane:
#build

ayoho (approved):

PR #19760, opened 2022-01-10
add missing code from DataSourceDefinition code path

njr-11:
Implementation for the Concurrency 3.0 resource definitions (like ContextServiceDefinition) was based on ConnectionFactoryDefinition, but I recently noticed that DataSourceDefinition has some additional code path. This pull adds that code path, illustrating the differences.  The decision to merge this or not merge it should be made by the team with expertise in the injection engine, which I'll ask to review.

### DIFF
--- a/dev/com.ibm.ws.injection.core/src/com/ibm/wsspi/injectionengine/ComponentNameSpaceConfiguration.java
+++ b/dev/com.ibm.ws.injection.core/src/com/ibm/wsspi/injectionengine/ComponentNameSpaceConfiguration.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007, 2015 IBM Corporation and others.
+ * Copyright (c) 2007, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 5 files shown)


njr-11:
#build

tkburroughs:
I recommend this pull request be discarded.  While all of the changes would be consistent with the way DataSourceDefinition is processed; the code would serve no purpose, as the binding file format does not currently support these data types.

Very specifically, this code is to enable an override of the annotation, for example, to completely ignore an `@DataSourceDefinition` annotation, and instead map the JNDI name of the annotation to a global annotation then one might create an ibm-ejb-jar-bnd.xml file or ibm-web-bnd.xml file with an entry something like this:

```
   <session name="JavaAppBean">
      <data-source name="java:app/env/dataSourceBinding" binding-name="jdbc/TestDataSource"/>
   </session>

```
An entry in the binding file like this would indicate that the `java:app/env/dataSourceBineding`  `@DataSourceDefinition` should be ignored, and instead when someone looks it up or injects it, then instead this should result in a lookup of `jdbc/TestDataSource` instead.

Looking in `com.ibm.ws.javaee.dd.commonbnd.RefBindingsGroup`, I see no evidence that the `ibm-*-bnd.xml` files would support the datatypes being addressed here.... and thus there could never be a binding override.  All of the code being added by this pull request would never be executed.

I would prefer not to add this code unless it is decided that the format of the XML binding file needs to be updated to support such an override.

If the XML file format is updated, then in addition to the changes in this PR, likely other code in DataSourceDefinition should also be included that does not log an error when there are 2 annotations with the same JNDI name but different attribute values.  For example, if two servlets in the same module have an @DataSource annotation with the same JNDI name `java:app/env/dataSourceBineding`, then the DataSourceDefinition binding object will throw an error if all of the attributes on both annotations don't match exactly...... unless there is a binding, in which case all of the attributes are ignored and a lookup of the DataSourceDefinition instead is re-directed to whatever datasource is identified by the binding.  None of the InjectionBinding objects for the Definitions modified in this PR have such code to avoid the error like DataSourceDefinition does.

Basically, we should either fully support overriding the "definitions" with a binding in ibm-*-bnd.xml; or not add any of the code, as it just adds complexity, unnecessary dependencies, and possibly confusion.

njr-11:
The build has lots of findbugs errors that are pre-existing errors in unrelated components, but no other errors

njr-11:
> I recommend this pull request be discarded. While all of the changes would be consistent with the way DataSourceDefinition is processed; the code would serve no purpose, as the binding file format does not currently support these data types.

Tracy - thanks for identifying purpose of the code in DataSourceDefinition.  Given that it does not apply in the case of the annotations for Concurrency 3.0, I agree we should not add it.

PR #19761, opened 2022-01-10
Add additional port check on stop for AcmeClientTest

kristip17:
Having problems with port 5002 not being open after `AcmeClientTest`, had a couple issues where `AcmeSimpleTest` fails to get the port but later tests are fine, add a longer grace period on the `HTTPServer` shutdown and add a port check after the HTTP server stop in `AcmeClientTest` with an extra long wait time.

RTC 288792


kristip17:
#build

jvanhill (approved):

kristip17:
#libby

PR #19766, opened 2022-01-11
EE repeat actions

tevans78:
Allow EE tests to be repeated in a similar way to MicroProfile
#build

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/rules/repeater/RepeatActions.java
+++ b/dev/fattest.simplicity/src/componenttest/rules/repeater/RepeatActions.java
@@ -15,9 +15,12 @@ import java.util.HashSet;
 import java.util.Set;
 
 import componenttest.custom.junit.runner.Mode.TestMode;
-import componenttest.rules.repeater.EERepeatTests.EEVersion;
 
 public class RepeatActions {
+
+    public static enum EEVersion {
+        EE6, EE7, EE8, EE9, EE10
+    }
 
     /**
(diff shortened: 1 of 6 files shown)


dev/fattest.simplicity/src/componenttest/rules/repeater/RepeatActions.java
```diff
@@ -0,0 +1,97 @@
...
+        if (eeVersion == EEVersion.EE7)
+            action = new EE7FeatureReplacementAction();
+        else if (eeVersion == EEVersion.EE8)
+            action = new EE8FeatureReplacementAction();
+        else if (eeVersion == EEVersion.EE9)
+            action = new JakartaEE9Action();
+        else if (eeVersion == EEVersion.EE10)
+            action = new JakartaEE10Action();
+        else
+            action = new FeatureReplacementAction();
```
Azquelt: If I'm reading this correctly, we define a FeatureSet for each EE version, but then actually we throw the FeatureSet away and use a specific `FeatureReplacementAction` instead?

Why don't we just use the `FeatureReplacementAction` (with all of its downsides) in the first place?

Azquelt: Ok, we discussed this. We do need to call the various EE `FeatureReplacementAction` classes because they do extra stuff like transform applications for jakarta.

The featureSet does know about the features that need added and removed and those are still added to the FRA in the lines below, but it doesn't know about any transformations needed.

Some comments and more curly brackets here would be nice.

tevans78:
#build

Azquelt (approved):

tevans78:
#libby

PR #19773, opened 2022-01-11
Merge integration into criu

tjwatson:


tjwatson:
#build

tjwatson:
@jhanders34 only need you to review 2f9fd13 and 75fdabc

For now to make progress I just added back the class.  I think the class overall is not that useful and would be for removing again, but probably after we get the `criu` branch merged into `integration`

jhanders34 (approved):
Approving the changes, but would like to see it change to not have ConcurrentHashSet in a future PR.  May make sense to have a helper method to create a concurrent set now that there are more  places that are doing it.

tjwatson:
#build

PR #19776, opened 2022-01-12
Fix threading unittest ScheduledExecutorImplTest

dazavala:
Modify ScheduledExecutorImpTest to initialize logging and test artifacts within junit lifecycle events rather than `<clinit>`.  This will ensure test artifacts are initialized after junit has created the test instance, reducing the chance the test fails due to artifacts initializing in an indeterminate order or state.


dazavala:
#build

tjwatson (approved):

dazavala:
Fixed copyright and removed a redundant comment.  Merge pending a passing build.
#build

dazavala:
#libby

PR #19782, opened 2022-01-12
Fix MDB RA FAT synchronization

tkburroughs:
Make a local copy of static synchronization object in MDB FAT to
ensure notifyAll is called on same object that is synchronized.


tkburroughs:
#build

olendvcook (approved):

PR #19783, opened 2022-01-12
fix memory leak in provider tracking

njr-11:
Resolve the TODO comment in ThirdPartyContextCoordinator where we hadn't gotten around to clearing old third-party thread context providers and their class loader when the application stops.


njr-11:
#build

jmstephensgit (approved):

njr-11:
Build has lots of pre-existing findbugs errors, all in unrelated components. Those are the only errors.

PR #19788, opened 2022-01-12
Asynchronously attempt to load ThreadPoolStats

### LINKED ISSUE #19780: Adding Monitor Filter increases Startup Time.
**Describe the bug**  
A user will experience slower startup (around 4 or 5 seconds) if mpMetrics-2.3 or newer is enabled and a monitor filter that does not include **ThreadPoolStats** in the server.xml.

**Steps to Reproduce**  
Enable mpMetrics-2.3, mpMetrics-3.0., or mpMetrics-4.0 and have the following in the server.xml
```
<monitor  filter="REST"/>
```

**Expected behavior**  
Startup does not hang for 5 seconds.

**Diagnostic information:**  
 - OpenLiberty Version: 22.0.0.1 and prior
 - mpMetrics-2.3 and greater.
 - Java Version: N/A
 - server.xml configuration
 A monitor filter element is included and does not contain  ThreadPoolStats.
 ```
 <monitor  filter="REST"/>
 ```

**Additional context**  
Will be fixed in 22.0.0.2 and startup will not hang for 5 seconds.

Channyboy:
As we are unable to read the server.xml faithfully during runtime. This PR will attempt to asynchronously load the ThreadPoolStats related MBeans in a separate thread. As a result it will not affect the sequential start-up of the bundle and hence the server.

Fixes #19780

The issue in #19780 is that there is a 5 second wait time in monitor metrics for the retrieval and registration of the `ThreadpoolStat` Mbeans as MP metrics. The 5 second was provided as it may take some time for the Mbean to be created and present in the Mbean server and the thread pool metrics are important for the MP metric feature. The 5 seconds is a worst-case scenario as normally it shouldn't take that long.  

  However, this is an issue when the `monitor-1.0` filter has filtered out the `ThreadpoolStats`. This forces the server to always wait 5 seconds at start-up.
  
  As we are unable to read the configuration values of other features from the monitor metrics bundle,  this PR opts to change the dynamic of how the `ThreadpoolStats` (and ultimately threadpool metrics) are loaded. Typically a for loop is created and goes through a table of values we expect to load up on start-up. This is a simple and quick process for each item in the table unless it is the Threadpool stats, which has its own unique logic (which blocks the for loop until it is done). To avoid the hard 5 second impact to the start-up of the bundle and hence the server, we will execute the loading of the thread pool metric in a separate thread.

### DIFF
--- a/dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
+++ b/dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
@@ -38,8 +38,6 @@ import com.ibm.websphere.ras.TraceComponent;
 import com.ibm.ws.ffdc.FFDCFilter;
 import com.ibm.ws.microprofile.metrics.impl.SharedMetricRegistries;
 
-import io.openliberty.microprofile.metrics.internal.monitor.MonitorMetricsHandler;
-
 @Component(name = "com.ibm.ws.microprofile.metrics.monitor.MonitorMetricsHandler", property = { "service.vendor=IBM"})
 public class MonitorMetricsHandler {
 	
--- a/dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
+++ b/dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
@@ -14,7 +14,6 @@ import java.lang.management.ManagementFactory;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanServer;
@@ -56,14 +55,22 @@ public class MonitorMetricsHandler {
         Tr.info(tc, "FEATURE_REGISTERED");
     }
 
-	@Reference
-    public void getSharedMetricRegistries(SharedMetricRegistries sharedMetricRegistry) {
+    @Reference
+    public void setSharedMetricRegistries(SharedMetricRegistries sharedMetricRegistry) {
         this.sharedMetricRegistry = sharedMetricRegistry;
     }
-	
+    
+    public void unsetSharedMetricRegistries(SharedMetricRegistries sharedMetricRegistry) {
+        this.sharedMetricRegistry = null;
+    }
+       
     @Reference
-    public void getExecutorService(ExecutorService execServ) {
+    public void setExecutorService(ExecutorService execServ) {
         this.execServ = execServ;
+    }
+    
+    public void unsetExecutorService(ExecutorService execServ) {
+        this.execServ = null;
     }
     
     @Deactivate
@@ -136,7 +143,7 @@ public class MonitorMetricsHandler {
             Set<ObjectInstance> mBeanObjectInstanceSet;
             try {
                 mBeanObjectInstanceSet = mbs.queryMBeans(new ObjectName(sName), null);
-                if (sName.contains("ThreadPoolStats") && mBeanObjectInstanceSet.isEmpty()) {
+                if (sName.contains("ThreadPoolStats") && mBeanObjectInstanceSet.isEmpty() && execServ != null) {
                     execServ.execute(() -> {
                         final int MAX_TIME_OUT = 5000;
                         int currentTimeOut = 0;
--- a/dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
+++ b/dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2019, 2021 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -13,6 +13,9 @@ package com.ibm.ws.microprofile.metrics.monitor;
 import java.lang.management.ManagementFactory;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.locks.ReentrantLock;
 
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanServer;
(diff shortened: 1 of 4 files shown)


Channyboy:
#build

Channyboy:
#build
#libby

Channyboy:
Errors are related to the findbugs issue (none are related to this PR and related bundles).
Related FATS have passed.

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
```diff
@@ -123,35 +126,70 @@ protected void unregister(String objectName) {
...
+                    lock.unlock();
+                }
+
+                if (sName.contains("ThreadPoolStats") && mBeanObjectInstanceSet.isEmpty()) {
+                    ExecutorService execServ = Executors.newSingleThreadExecutor();
+                    execServ.execute(() -> {
+                        final int MAX_TIME_OUT = 5000;
+                        int currentTimeOut = 0;
+                        Set<ObjectInstance> mBeanObjectInstanceSetTemp = mBeanObjectInstanceSet;
+                        while (mBeanObjectInstanceSet.isEmpty() && currentTimeOut <= MAX_TIME_OUT) {
```
fmhwong: The while loop is checking `mBeanObjectInstanceSet` is empty but the body is setting `mBeanObjectInstanceSetTemp`.

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
```diff
@@ -123,35 +126,70 @@ protected void unregister(String objectName) {
...
+                            try {
+                                Thread.sleep(50);
+                                lock.lock();
+                                try {
+                                    mBeanObjectInstanceSetTemp = mbs.queryMBeans(new ObjectName(sName), null);
+                                } finally {
+                                    lock.unlock();
+                                }
+                                currentTimeOut += 50;
```
fmhwong, line 156: Is there a possibility of endless loop here?  When `mbs.queryMBeans()` throws an exception, it will jump to the catch block and skips the increment of `currentTimeOut` and go back to the beginning of the while loop.

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
```diff
@@ -123,35 +126,70 @@ protected void unregister(String objectName) {
...
-			}
-		}
-	}
+        MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
+        ReentrantLock lock = new ReentrantLock();
+        for (String sName : mappingTable.getKeys()) {
+            Set<ObjectInstance> mBeanObjectInstanceSet;
+
+            try {
+                lock.lock();
```
fmhwong: Why do we need a lock?  Is `mbs.queryMBeans()` not thread safe?

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2019, 2021 IBM Corporation and others.
```
fmhwong: 2019, 2022

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
Channyboy: fixed

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
Channyboy: Yes. It's unlikely, but fixed to exit the thread.

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
Channyboy: 😂

Channyboy:
#build 
#libby

dev/com.ibm.ws.microprofile.metrics.2.0.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
Channyboy: Wasn't sure, the java doc doesn't seem to say whether it is or not. But given the role of the MBeanserver and how it's used it _should_ be. But since there wasn't an explicit statement I've locked in just in case.

Channyboy: But then again, there shouldn't be any concurrency concerns.

Channyboy:
#build 
#libby

fmhwong (commented):
One thing that I am not sure is whether it is okay to use `Executors.newSingleThreadExecutor()`.  Or do we need to use `ManagedExecutorService`

Channyboy:
#build 
#libby

dev/com.ibm.ws.microprofile.metrics.2.3.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
```diff
@@ -57,6 +60,11 @@ protected void activate(ComponentContext context) {
     public void getSharedMetricRegistries(SharedMetricRegistries sharedMetricRegistry) {
         this.sharedMetricRegistry = sharedMetricRegistry;
     }
+	
+    @Reference
+    public void getExecutorService(ExecutorService execServ) {
```
fmhwong: This is usually called `setExecutorService()` and there should be a corresponding `unsetExecutorService()`.

dev/com.ibm.ws.microprofile.metrics.2.3.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
```diff
@@ -123,35 +131,57 @@ protected void unregister(String objectName) {
...
-			}
-		}
-	}
+        MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
+        for (String sName : mappingTable.getKeys()) {
+            Set<ObjectInstance> mBeanObjectInstanceSet;
+            try {
+                mBeanObjectInstanceSet = mbs.queryMBeans(new ObjectName(sName), null);
+                if (sName.contains("ThreadPoolStats") && mBeanObjectInstanceSet.isEmpty()) {
+                    execServ.execute(() -> {
```
fmhwong, line 147: It may be safer to check for null in `execServ`

dev/com.ibm.ws.microprofile.metrics.2.3.monitor/src/com/ibm/ws/microprofile/metrics/monitor/MonitorMetricsHandler.java
Channyboy: I don't think that service will be unloaded unless the server is shut down. But added it anyways.

Channyboy:
#build
#libby

fmhwong (approved):
LGTM

Issue #19787, opened 2022-01-12
NullPointerException when installing WEBENAB on iSeries with Turkish locale


jacobwdv:
With English locale, the failure does not occur. 

On Turkish locale, installations fail with 
```
<entry num='1503' time='1641925340712' elapsed='01:54.46' level='DEBUG' thread='Thread-19'>
 <logger>WASInstallLogger</logger>
 <class>com.ibm.cic.agent.internal.core.IMLoggerImpl</class>
 <method>debug</method>
 <message>Beklenmeyen özel durum</message>
 <exception>
  java.lang.NullPointerException
  <stack>com.ibm.ws.install.internal.ExceptionUtils.create(ExceptionUtils.java:457)</stack>
  <stack>com.ibm.ws.install.internal.ResolveDirector.resolve(ResolveDirector.java:652)</stack>
  <stack>com.ibm.ws.install.internal.ResolveDirector.resolveMap(ResolveDirector.java:714)</stack>
  <stack>com.ibm.ws.install.internal.Director.installAssets(Director.java:1437)</stack>
  <stack>com.ibm.ws.install.internal.InstallKernelImpl.installAsset(InstallKernelImpl.java:538)</stack>
  <stack>com.ibm.ws.install.internal.InstallKernelImpl.installAsset(InstallKernelImpl.java:524)</stack>
  <stack>com.ibm.was.liberty.dynamic.feature.install.InstallAsset.installAsset(InstallAsset.java:367)</stack>
  <stack>com.ibm.was.liberty.dynamic.feature.install.InstallAsset.perform(InstallAsset.java:185)</stack>
  <stack>com.ibm.was.liberty.dynamic.feature.install.InstallAsset.run(InstallAsset.java:63)</stack>
  <stack>sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)</stack>
  <stack>sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:90)</stack>
  <stack>sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:55)</stack>
  <stack>java.lang.reflect.Method.invoke(Method.java:508)</stack>
  <stack>com.ibm.cic.agent.core.commonNativeInstallAdapter.Invoke$InvokeRunnable.run(Invoke.java:424)</stack>
  <stack>java.lang.Thread.run(Thread.java:825)</stack>
 </exception>
```
[...]

PR #19791, opened 2022-01-12
preventing NPE by checking if the String returned by getMessageKey() …

jacobwdv:
…is null before doing .equals() comparisons.

maintenance of code to prevent NPE as seen in https://github.com/OpenLiberty/open-liberty/issues/19787

The NPE only happened because of an odd behaviour in the repository code. In any case, preventing an NPE would have resulted in better error output and user experience. This code change adds safeguards to check for null from the `getMessageKey()` API before using the value.


jacobwdv:
#build

bensonlatibm (approved):
lgtm!

jacobwdv:
#build

jacobwdv:
all failures in build analysis show they are existing defects. Therefore my changes should be fit to merge.

jacobwdv:
#libby-bot

PR #19794, opened 2022-01-13
Update date used in testUserAgentNeedsUpdateFalseCondition

### LINKED ISSUE #19793: Fix test bug in testUserAgentNeedsUpdateFalseCondition
Test uses Jan 1 00:00 GMT as the header date, but this can cause problems if the test runs just before midnight in different time zones.  

Test uses GMT in the header, but if the server is running on PST then the test can fail. (GMT is 8 hours ahead of PST).   

I think changing the header date to Jan 2nd or 3rd should fix this small bug.

volosied:
Fixes #19793
- Update date to avoid timezone problems around new years


volosied:
#build

volosied:
force pushed to update copyright

pnicolucci (approved):

pnicolucci:
#libby

PR #19805, opened 2022-01-13
EJBTimerTest fix interval countdown latch waiting in timeout

olendvcook:
if statement caused it to only wait on the 2nd interval and not the 3rd one I had added to fix a build break.


olendvcook:
#build

brideck (approved):

PR #19806, opened 2022-01-13
Update kafka containers to latest

KyleAure:
In an effort to keep better control over test containers that are potentially vulnerable to log4shell we should ensure that we are only using one version of the image `confluentrinc/cp-kafka` which does not currently use log4j 2.x, but is still using the out of date log4j 1.x versions.


KyleAure:
#build

KyleAure:
confluent has released a statement indicating they DO NOT plan to patch their image and intend to keep log4j v1.2.17
https://support.confluent.io/hc/en-us/articles/4412615410580-December-2021-Log4j-Vulnerabilities-Advisory
Closing this issue, and deferring to the teams who own these test suites to come up with an alternative solution.

PR #19811, opened 2022-01-13
Add JNDI trace to widen the login window for UserEnum test

kristip17:
The `UserEnumerationTest`  `testUserEnumerationDisabled()` had a valid/invalid count failure, but was working correctly (login delay disabled), but the times are very short to do the valid/invalid user calculations and popped some false invalid/valid determinations. Adding the JDNI BER trace to see if that gives us better numbers with longer times to work with.

Also updated the method name in the test.

RTC 288790


kristip17:
#build

jvanhill (approved):

PR #19820, opened 2022-01-14
Update JPA FAT Bucket for Oracle database support.

jgrassel:
Looks like the bucket needs to be updated to include JDBC library in the application's classloader, in order for OpenJPA to find Oracle's Blob classes.

### DIFF
--- a/dev/com.ibm.ws.jpa.tests.spec10.entity_fat.common/fat/src/com/ibm/ws/jpa/tests/spec10/entity/tests/Entity_EJB.java
+++ b/dev/com.ibm.ws.jpa.tests.spec10.entity_fat.common/fat/src/com/ibm/ws/jpa/tests/spec10/entity/tests/Entity_EJB.java
@@ -244,11 +244,18 @@ public class Entity_EJB extends JPAFATServletClient {
             ConfigElementList<ClassloaderElement> cel = appRecord.getClassloaders();
             ClassloaderElement loader = new ClassloaderElement();
             loader.getCommonLibraryRefs().add("HibernateLib");
+            loader.getCommonLibraryRefs().add("AnonymousJDBCLib");
             cel.add(loader);
         } else if (AbstractFATSuite.repeatPhase != null && AbstractFATSuite.repeatPhase.contains("openjpa")) {
             ConfigElementList<ClassloaderElement> cel = appRecord.getClassloaders();
             ClassloaderElement loader = new ClassloaderElement();
             loader.getCommonLibraryRefs().add("OpenJPALib");
+            loader.getCommonLibraryRefs().add("AnonymousJDBCLib");
+            cel.add(loader);
+        } else {
+            ConfigElementList<ClassloaderElement> cel = appRecord.getClassloaders();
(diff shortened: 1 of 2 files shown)


jgrassel:
#build

dev/com.ibm.ws.jpa.tests.spec10.entity_fat.common/fat/src/com/ibm/ws/jpa/tests/spec10/entity/tests/Entity_EJB.java
```diff
@@ -244,11 +244,18 @@ public boolean include(ArchivePath arg0) {
             ConfigElementList<ClassloaderElement> cel = appRecord.getClassloaders();
             ClassloaderElement loader = new ClassloaderElement();
             loader.getCommonLibraryRefs().add("HibernateLib");
+            loader.getCommonLibraryRefs().add("AnonymousJDBCLib");
```
dazey3: This appears to be a requirement for old versions of OpenJPA (2.2.x specifically, because it was fixed in 3.0.0) and for Oracle only. I would rather not add the JDBC Library to all providers classpaths for all databases. It might mask real issues for other providers, on other platforms, that will need to be investigated (like if we start testing against Sybase and EclipseLink has some hard dependency we need to fix)

dazey3 (changes requested):

dev/com.ibm.ws.jpa.tests.spec10.entity_fat.common/fat/src/com/ibm/ws/jpa/tests/spec10/entity/tests/Entity_EJB.java
dazey3: How about the following instead?

```java
        /*
         * TODO: OpenJPA 2.2.x (JPA 2.0) has hard dependencies on the Oracle JDBC driver classes and
         * therefore requires the driver be added to the application classloader
         *
         * https://issues.apache.org/jira/projects/OPENJPA/issues/OPENJPA-2602
         * https://issues.apache.org/jira/projects/OPENJPA/issues/OPENJPA-2690
         */
        if (AbstractFATSuite.repeatPhase != null && AbstractFATSuite.repeatPhase.contains("20")
            && DatabaseVendor.ORACLE.equals(getDbVendor())) {
            ConfigElementList<ClassloaderElement> cel = appRecord.getClassloaders();
            ClassloaderElement loader = new ClassloaderElement();
            loader.getCommonLibraryRefs().add("AnonymousJDBCLib");
            cel.add(loader);
        }
```

jgrassel: Done.

jgrassel:
#run-libby-bot
#build

jgrassel:
#run-libby-bot

dazey3 (approved):
lgtm

PR #19823, opened 2022-01-14
resource definitions in EJB deployment descriptor

njr-11:
Tests (disabled for now) of the Concurrency resource definitions in an EJB deployment descriptor that we can switch on when ready.


njr-11:
#build

mswatosh (approved):

Issue #15789, opened 2021-02-03
EjbTimerTest.testCDIScopeViaEJBTimer may be operating incorrectly


tevans78:
https://github.com/OpenLiberty/open-liberty/pull/15583#discussion_r569518954

> the call to /ejbTimer/Timer appears to increment the session scoped counter after setting up a timer, rather than when that timer executes, contrary to the javadoc on this test method. (See EjbSessionBean#incCountersViaTimer)

tevans78:
@benjamin-confino  Can this be closed? What is remaining?

benjamin-confino:
It can be closed.

PR #19827, opened 2022-01-16
Ejb timer test

benjamin-confino:
The old test was a maze which we did not understand, so we're creating a new simpler one and will debate if we should keep the old one around just in case

This is part of #15789

### DIFF
--- a/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/EjbSessionBean.java
+++ b/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/EjbSessionBean.java
@@ -1,0 +1,93 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.cdi.ejb.apps.newtimer;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.annotation.Resource;
+import javax.ejb.EJB;
+import javax.ejb.Local;
+import javax.ejb.LocalBean;
+import javax.ejb.Stateless;
+import javax.ejb.Timeout;
+import javax.ejb.Timer;
+import javax.ejb.TimerConfig;
+import javax.ejb.TimerService;
+import javax.inject.Inject;
+
+import javax.enterprise.context.ContextNotActiveException;
+
+/**
+ * Session Bean implementation class EjbSessionBean
+ */
+@Stateless
+@Local(EjbSessionBeanLocal.class)
+@LocalBean
+@MyCDIInterceptorBinding
+public class EjbSessionBean implements EjbSessionBeanLocal {
+    private static final long TIMEOUT = 1;
+
+    @Resource
+    TimerService timerService;
+
+    @Inject RequestScopedBean rsb;
+
+    @Inject SessionScopedBean ssb;
+
+    private int requestScopedIndex = -1;
+    public static Boolean canAccessRequestScope = null;
+    public static Boolean seperateRequestScopes = null;
+    public static Boolean sessionScopeInactive = null;
+
+
+    @Timeout
+    void ping(Timer timer) {
+        //Test One. Is a request scope active in the context of an EJB timer?
+        try {
+            if (rsb.toString() != null) {
+                canAccessRequestScope = true;
+            } else {
+                canAccessRequestScope = false;
+            }
+        } catch (ContextNotActiveException e) {
+            canAccessRequestScope = false;
+        }
+
+        //Test two. Is it a different request scope to outside the timer?
+        try {
+            if (rsb.getIndex() != requestScopedIndex) {
+                seperateRequestScopes = true;
+            } else {
+                seperateRequestScopes = false;
+            }
+        } catch (ContextNotActiveException e) {
+            canAccessRequestScope = false;
+        }
+
+        //Test three. Is the session scope inactive in the context of an EJB timer?
+        try {
+            ssb.toString();
+            sessionScopeInactive = false;
+        } catch (ContextNotActiveException e) {
+            sessionScopeInactive = true;
+        }
+    }
+
+    @Override
+    public void initTimer() {
+        requestScopedIndex = rsb.getIndex();
+
+        System.out.println("CL: " + timerService.getClass().getClassLoader());
+        TimerConfig config = new TimerConfig();
+        config.setPersistent(false);
+        timerService.createSingleActionTimer(TIMEOUT, config);
+    }
+}
--- a/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/MyCDIInterceptor.java
+++ b/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/MyCDIInterceptor.java
@@ -1,0 +1,35 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.cdi.ejb.apps.newtimer;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.annotation.Priority;
+import javax.annotation.Resource;
+import javax.inject.Inject;
+import javax.interceptor.AroundTimeout;
+import javax.interceptor.Interceptor;
+import javax.interceptor.InvocationContext;
+
+@MyCDIInterceptorBinding
+@Interceptor
+@Priority(1)
+public class MyCDIInterceptor {
+
+    public static boolean interceptorCalled = false;
+
+    @AroundTimeout
+    public Object invoke(InvocationContext context) throws Exception {
+        interceptorCalled = true;
+        return context.proceed();
+    }
+
+}
--- a/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/MyCDIInterceptorBinding.java
+++ b/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/MyCDIInterceptorBinding.java
@@ -1,0 +1,29 @@
+/*******************************************************************************
+ * Copyright (c) 2015 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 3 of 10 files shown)


benjamin-confino:
#build

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/EjbSessionBean.java
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            canAccessRequestScope = false;
+        }
+
+        //Test two. Is it a different request scope to outside the timer?
+        if (rsb.getIndex() != requestScopedIndex) {
+            seperateRequestScopes = true;
+        } else {
+            seperateRequestScopes = false;
+        }
```
Azquelt: Given that your first test is checking the request scope is active, it should really be in a try-catch for `ContextNotActiveException` like the session scope test (even though we do expect it to be active).

The second test also relies on the request scope being active, so it should also be within a try-catch.

If you really wanted to get fancy, you could make sure that any exception thrown is passed back out and thrown from the test method as well so that you get the actual `ContextNotActiveException` with the right line number in the test result.

Azquelt (changes requested):

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/MyCDIInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+@Interceptor
+@Priority(1)
+public class MyCDIInterceptor {
+
+    public static boolean interceptorCalled = false;
+
+    @AroundTimeout
+    public Object invoke(InvocationContext context) throws Exception {
+        interceptorCalled = true;
+        return context.proceed();
```
Azquelt: Trailing whitespace on this line.

Surprised this wasn't removed by the IDE automatically.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/TestEjbTimerServlet.java
```diff
@@ -0,0 +1,48 @@
...
+    @EJB
+    EjbSessionBeanLocal bean;
+
+    @Test
+    public void testEjbTimers() throws ServletException, IOException {
+        bean.initTimer();
+        try {
+            Thread.sleep(3000);
+        } catch (InterruptedException e) {
+            assertTrue("The timer exploded", false);
```
Azquelt: 1. There's a `fail` method, which is preferable to using `assertTrue(false)`.
2. It's usually makes it easier to debug the problem if you allow the original exception to be thrown, rather than catching it and failing the test.
   a. If you do want to add a message to say what happened, you can do `throw new AssertionError("my message", e);` to add your message while preserving the original exception.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/TestEjbTimerServlet.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final long serialVersionUID = 1L;
+    @EJB
+    EjbSessionBeanLocal bean;
+
+    @Test
+    public void testEjbTimers() throws ServletException, IOException {
+        bean.initTimer();
+        try {
+            Thread.sleep(3000);
+        } catch (InterruptedException e) {
```
Azquelt: Can we avoid using `sleep`? It means we always have to wait the sleep time and there's a risk that something is slow and we don't wait long enough.

We could have a `CompletableFuture` stored in a static field which is completed by the timeout method. In this method we would wait on the `CompletableFuture` so that the rest of the test runs as soon as it completes.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/tests/NewEjbTimerTest.java
```diff
@@ -0,0 +1,79 @@
...
+    })
+    public static LibertyServer server;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        WebArchive ejbTimer = ShrinkWrap.create(WebArchive.class, EJB_TIMER_APP_NAME + ".war")
+                                        .addPackage(com.ibm.ws.cdi.ejb.apps.newtimer.SessionScopedBean.class.getPackage())
+                                        .add(new FileAsset(new File("test-applications/" + EJB_TIMER_APP_NAME + ".war/resources/META-INF/permissions.xml")),
+                                             "/META-INF/permissions.xml")
+                                        .add(new FileAsset(new File("test-applications/" + EJB_TIMER_APP_NAME + ".war/resources/WEB-INF/beans.xml")), "/WEB-INF/beans.xml");
```
Azquelt: I would prefer these files to be in the `com.ibm.ws.cdi.ejb.apps.newtimer` package, next to the application classes, rather than in `test-applications`.

dev/com.ibm.ws.cdi.ejb_fat/test-applications/newEjbTimer.war/resources/WEB-INF/beans.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://java.sun.com/xml/ns/javaee"
+   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+   xsi:schemaLocation="
+      http://java.sun.com/xml/ns/javaee
+      http://java.sun.com/xml/ns/javaee/beans_1_0.xsd">
+
+</beans>
```
Azquelt: What's the intention of this beans.xml?

I think it's using the CDI-1.0 schema and ends effectively setting `bean-discovery-mode=all`?

Azquelt:
Please add a link to the issue in the PR description

Azquelt:
IIRC, the issue was a test that wasn't working properly? However, this PR appears to just add stuff. Was that intentional? Please document in the PR description.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/MyCDIInterceptorBinding.java
```diff
@@ -0,0 +1,29 @@
+/*******************************************************************************
+ * Copyright (c) 2015 IBM Corporation and others.
```
Azquelt, line 2: I assume this "new" file has been copied without really changing anything?

If not the copyright date should be 2022.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/EjbSessionBean.java
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            if (rsb.toString() != null) {
+                canAccessRequestScope = true;
+            } else {
+                canAccessRequestScope = false;
+            }
+        } catch (ContextNotActiveException e) {
+            canAccessRequestScope = false;
+            throw e;
+        }
```
Azquelt: Throwing the exception here prevents the rest of the test from running, though I guess that doesn't matter too much because each of the boolean fields are asserted in turn within one test.

As this is a `@Timeout` method, I guess the `ContextNotActiveException` will get logged/FFDC'd, so I guess that could be useful?

Was that what you were intending?

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/MyCDIInterceptorBinding.java
benjamin-confino: You assume correctly.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/tests/NewEjbTimerTest.java
benjamin-confino: I agree, but didn't want to introduce an inconsistency in the bucket. At some point we'll be moving the lot across.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/EjbSessionBean.java
```diff
@@ -0,0 +1,95 @@
...
+
+        //Test three. Is the session scope inactive in the context of an EJB timer?
+        try {
+            ssb.toString();
+            sessionScopeInactive = false;
+        } catch (ContextNotActiveException e) {
+            sessionScopeInactive = true;
+        }
```
Azquelt: Now that you have three separate tests, you definitely don't want the `throw e;` lines within your catch blocks.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/TestEjbTimerServlet.java
Azquelt: This still needs fixed. This should do it:
```
throw new AssertionError("The timer exploded", e);
```

dev/com.ibm.ws.cdi.ejb_fat/test-applications/newEjbTimer.war/resources/WEB-INF/beans.xml
Azquelt: Beans.xml was unnecessary and has been removed.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/TestEjbTimerServlet.java
```diff
@@ -0,0 +1,72 @@
...
+import componenttest.app.FATServlet;
+import org.junit.Test;
+
+@WebServlet("/Timer")
+public class TestEjbTimerServlet extends FATServlet {
+    private static final long serialVersionUID = 1L;
+    @EJB
+    EjbSessionBeanLocal bean;
+
+    private boolean needToSetUp = true;
```
Azquelt: I'm not sure that the servlet lifecycle guarantees that every invocation of this servlet will use the same instance. Maybe make this field static to be sure we won't repeat the setup?

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/newTimer/TestEjbTimerServlet.java
```diff
@@ -0,0 +1,72 @@
...
+            } catch (InterruptedException e) {
+                assertTrue("The timer exploded", false);
+            }
+        }
+    }
+
+    @Test
+    public void testRequestScopeActiveWithinTimeout() throws ServletException, IOException {
+        setUp();
+        assertTrue("The request scope was not active inside a timeout method", EjbSessionBean.canAccessRequestScope);
```
Azquelt: Because the field is a `Boolean` rather than a `boolean`, I think using `assertTrue` will result in a `NullPointerException` during autoboxing if the field is `null`.

Using `assertEquals(Boolean.TRUE, MyClass.myField);` here would be better as that will handle the `null` case.

Note: If you do `assertEquals(true, MyClass.myField);`, I'm not sure whether the constant gets boxed or the field gets unboxed (or the compiler complains) so it's probably better to stick with `Boolean.TRUE` instead.

benjamin-confino:
#build

Azquelt (approved):

benjamin-confino:
#build

benjamin-confino:
#run-libby-bot

PR #19859, opened 2022-01-20
ContextServiceDefinition used from different EJB

njr-11:
Verify that a resource definition in one EJB can use a java:module resource definition from another EJB when both are in the same EJB module.


njr-11:
#build

KyleAure (approved):

njr-11:
Added a commit with a file that I forgot to check in (the EJB to which I moved the annotation)
#run-libby-bot
#build

PR #19871, opened 2022-01-20
Fix RepositoryResolver install lists for tolerated dependencies

### LINKED ISSUE #19860: Updating MicroProfile versions on server.xml causes issues with install manager
1: use start.micropropfile.io to create.an app
2. update pom.xml with microrpofile version to 5.0 and server.xml to microProfile-5.0 on server.xml for both service-a and service-b
3. used transformer to transform all of the sources to jakarta
4. start my new jakarta app using mvn clean liberty:dev
5. got the following error 

[INFO] [ERROR   ] CWWKF0044E: The cdi-3.0 and io.openliberty.distributedMapInternal-1.0 features cannot be loaded at the same time. The cdi-3.0 feature of Jakarta EE 9 is incompatible with the io.openliberty.distributedMapInternal-1.0 feature of Java EE 6. The microProfile-5.0 and microProfile-5.0 configured features include an incompatible combination of features. Your configuration is not supported. Update the configuration to use features that support either the Jakarta EE or Java EE programming models, but not both.
[INFO] [WARNING ] CWWKF0046W: The configuration includes an incompatible combination of features. As a result, the feature manager did not install any features.

Azquelt:
Fix the creation of install lists when a feature is installed but its
dependencies aren't. This can happen when a feature tolerates two
versions of a dependency, only one of which is installed, and the
solution from the kernel resolver uses the other one.

This situation is common in Liberty where we have a single feature which
supports both EE8 and EE9 by having a private feature for each and
tolerating both of them. The eeCompatible singleton features will ensure
that the correct one is used, depending on which other features are
enabled on the server.

This fix includes the following changes:
* When building an install list, recurse through the dependencies of any
  installed features rather than stopping. The installed features are
  removed later when features are mapped to EsaResources.
* Refactor the code for finding the features used to satisfy the
  ProvisionCapability header of an auto-feature so that it can be used
  for features which are installed as well as features in the
  repository.
* When building an install list for samples, don't try to build an
  install list for a feature which wasn't found.
* Accept `ProvisioningFeatureDefinition.getTolerates()` returning `null`
* Ensure we don't try to build install lists for features which didn't resolve
  * Previously this wasn't a problem since we'd detect that we didn't have an `EsaResource` and assume it was already installed
* Ensure we don't report conflicting features as missing

The following tests were added:
* unit test for `createInstallList` where a feature is installed but its dependency is not
* unit tests for `RepositoryResolver` where:
  * a feature is already installed but resolution requires its tolerated dependency to be satisfied by a feature which is not installed
  * an auto-feature is already installed but resolution requires its tolerated dependency to be satisfied by a feature which is not installed
the cases where the feature with the tolerated dependency is a regular feature and an auto-feature
  * two requested features conflict
  * two requested features have dependencies which conflict
  * two requested features conflict and also have dependencies which are missing from the repository
  * an unrelated feature which is already installed does not have its dependencies satisfied (i.e. the existing runtime is broken, but not in a way that affects the features requested for install)
* Repeat all basic unit tests for `RepositoryResolver` for both the `resolve()` method and the `resolveAsSet()` method
  * In most situations, both methods should give the same result
* unit tests to verify the behaviour of the kernel resolver when a feature is missing and when features conflict

Fixes #19860

### DIFF
--- a/dev/com.ibm.ws.repository.resolver/test/com/ibm/ws/repository/resolver/MissingRequirementMatcher.java
+++ b/dev/com.ibm.ws.repository.resolver/test/com/ibm/ws/repository/resolver/MissingRequirementMatcher.java
@@ -1,0 +1,66 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.repository.resolver;
+
+import java.util.Objects;
+
+import org.hamcrest.Description;
+import org.hamcrest.TypeSafeMatcher;
+
+import com.ibm.ws.repository.resolver.RepositoryResolutionException.MissingRequirement;
+import com.ibm.ws.repository.resources.RepositoryResource;
+
+/**
+ * Verify a {@link MissingRequirement}
+ */
+public class MissingRequirementMatcher extends TypeSafeMatcher<MissingRequirement> {
+
+    private final String requirementName;
+
+    private final RepositoryResource owningResource;
+
+    public static MissingRequirementMatcher missingRequirement(String requirementName, RepositoryResource owningResource) {
+        return new MissingRequirementMatcher(requirementName, owningResource);
+    }
+
+    private MissingRequirementMatcher(String requirementName, RepositoryResource owningResource) {
+        this.requirementName = requirementName;
+        this.owningResource = owningResource;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void describeTo(Description desc) {
+        desc.appendText("MissingRequirement [");
+        desc.appendText("name = ").appendValue(requirementName);
+        desc.appendText(", ");
+        desc.appendText("owning resource = ").appendValue(owningResource);
+        desc.appendText("]");
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    protected void describeMismatchSafely(MissingRequirement item, Description desc) {
+        desc.appendText("MissingRequirement [");
+        desc.appendText("name = ").appendValue(item.getRequirementName());
+        desc.appendText(", ");
+        desc.appendText("owning resource = ").appendValue(item.getOwningResource());
+        desc.appendText("]");
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    protected boolean matchesSafely(MissingRequirement other) {
+        return Objects.equals(requirementName, other.getRequirementName())
+               && Objects.equals(owningResource, other.getOwningResource());
+    }
+
+}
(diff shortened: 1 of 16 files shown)


Azquelt:
#build

Azquelt:
#build
#libby

Azquelt:
The last build showed failures in the installUtility fat, which I think I've now got to the bottom of.

The original fix made the way that `createInstallLists()` iterates through feature dependencies more consistent, in the original code, missing dependencies were handled differently depending on whether they occurred at the top level or further down the dependency tree. However, this exposed that featureUtility wasn't expecting conflicting features to be reported as "missing" at all (which makes sense because they are present in the repository, they just weren't selected because they conflict).

My second fix, made in the last commit, changes the way that we find missing features so that we start with the list that the kernel resolver said were missing and then try to match that up with resources from the repository which did resolve that have a dependency on the feature reported as missing.

I've also updated the unit tests so that the basic scenarios are run against both the `resolve()` and `resolveAsSet()` methods and added some new tests for conflict scenarios.

Azquelt:
#build
#libby

Azquelt:
The last change I made has a problem, if there's something invalid in the server install (e.g. a feature with a missing dependency), then any further install will fail because when we do individual feature resolution, we ask the kernel resolver to resolve all installed features so that we pick up any auto-features which should be installed but aren't. The kernel resolver therefore reports the missing dependency and the resolver reports that as a failure.

To revise the design slightly, we do want to walk the dependency tree starting with the requested features to find missing dependencies, but we only want to report them as missing if the kernel resolver also reported them as missing (otherwise they will have been reported as a conflict instead).

Azquelt:
#libby

Azquelt:
#build

dev/com.ibm.ws.repository.resolver/test/com/ibm/ws/repository/resolver/MissingRequirementMatcher.java
```diff
@@ -0,0 +1,66 @@
+/*******************************************************************************
```
pnickoll, line 1: Is this class used?

dev/com.ibm.ws.repository.resolver/test/com/ibm/ws/repository/resolver/RepositoryResolverTest.java
```diff
@@ -258,6 +258,36 @@ public void testCreateInstallListAutoFeature() {
...
+
+        EsaResourceWritable internalA20 = WritableResourceFactory.createEsa(null);
+        internalA20.setProvideFeature("com.example.internalFeatureA-2.0");
+        internalA20.addRequireFeatureWithTolerates("com.example.base-2.0", Collections.emptyList());
+
+        EsaResourceWritable featureB = WritableResourceFactory.createEsa(null);
+        featureB.setProvideFeature("com.example.featureB-1.0");
+        featureB.addRequireFeatureWithTolerates("com.example.base-2.0", Collections.emptyList());
+
+        RepositoryResolver resolver = testResolver().withResolvedInstalledFeature(base20, featureA)
```
pnickoll, line 284: Should InternalA10 already be listed as installed?

dev/com.ibm.ws.repository.resolver/test/com/ibm/ws/repository/resolver/MissingRequirementMatcher.java
pnickoll: Ah nvm was in a part of the diff that was large and needed to hit "display diff" to see it, can see it now. Ignore ^^

dev/com.ibm.ws.repository.resolver/test/com/ibm/ws/repository/resolver/RepositoryResolverTest.java
Azquelt: Possibly it should be included as installed but not resolved? I don't _think_ that matters in this test but yes, should change it.

pnickoll (approved):

PR #19874, opened 2022-01-20
Update issue template creating manually setting issue configuration

dazey3:
As template information (`assignees`, `labels`, ect) changes, explicit links in other locations (wikis, other templates, ect) that set their own values are not updated to match. This leads to incorrectly created issues (incorrect assignees, invalid labels, ect) unless every location is updated to match. Remembering to update all these locations is error prone.

Alternatively, we could just reference `https://github.com/OpenLiberty/open-liberty/issues/new/choose` and inform which template to choose. This would help create a single point of control as the templates change labels, assignees, titles


dazey3:
#run-libby-bot

PR #19884, opened 2022-01-21
Avoid NPE in ExternalTestService

KyleAure:
Avoid a NullPointerException, and instead, throw a more meaningful exception if we cannot access our external test services.


gjwatts (approved):

KyleAure:
#build

KyleAure:
#libbybot

PR #19909, opened 2022-01-24
tests for app-defined Concurrency resources in web.xml

njr-11:
Write tests to cover app-defined resources in the web.xml deployment descriptor for
`<context-service>`, `<managed-executor>`, `<managed-scheduled-executor>`, and `<managed-thread-factory>`
which we can enable once ready


njr-11:
#build

KyleAure (approved):

PR #19922, opened 2022-01-25
Fix test failure in RTC283807.

jvanhill:
We have a few more tests that are failing due to windows not releasing file locks. We will be skipping them when running with the transformer in Jakarta EE9.


jvanhill:
#build

kristip17 (approved):

PR #19945, opened 2022-01-27
log stdout of find command

bensonlatibm:
add extra debugging output (capture stdout) for the find command used in the InstallPackaging FAT.
Also, scrape output from find command to explicitly check for matching folder entries.

### DIFF
--- a/dev/com.ibm.ws.install.packaging_fat/fat/src/com/ibm/ws/install/packaging/fat/InstallPackagesToolTest.java
+++ b/dev/com.ibm.ws.install.packaging_fat/fat/src/com/ibm/ws/install/packaging/fat/InstallPackagesToolTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 IBM Corporation and others.
+ * Copyright (c) 2018, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


bensonlatibm:
#build

bensonlatibm:
#link 288859

jjiwooLim (commented):
.

dev/com.ibm.ws.install.packaging_fat/fat/src/com/ibm/ws/install/packaging/fat/InstallPackagesToolTest.java
```diff
@@ -443,7 +443,14 @@ protected static boolean checkUserGroupOwnership(String METHOD_NAME, String User
         if ((po1.getStdout().length() == 0) && (po2.getStdout().length() == 0)) {
             returnCode = true;
         } else {
-            returnCode = false;
+            // ensure stdout of either find command contains the folder we are supposed to be checking
+            if (po1.getStdout().contains(Folder) || po1.getStdout().contains(Folder)) {
```
jjiwooLim: Is it possible to get a folder name that doesn't exist? If the folder doesn't exist, it would print something like
find: 'folder_name': No such file or directory on linux system. I am not sure what the output would be on different systems, but maybe we could change .contains to .equals ?

bensonlatibm: Good point.  If a folder which doesn't exist is specified, this won't do what I'd expect, as it'd output a message stating folder was not found.
I think I can add in a check for the RC.
find will return an RC of 1 if an invalid folder, user, or group is specified.

If the folder exists, and a valid user/group is specified, find will return 0 (regardless if any entries are found)

Note, I can't use .equals, since find will return a list of all files in the path specified.
eg. in this case, all files/folders in /var/lib/openliberty are not owned by root.
this is what find returns:

```
[root@donut81 ~]# find /var/lib/openliberty \! -user root ; echo $?
/var/lib/openliberty
/var/lib/openliberty/usr
/var/lib/openliberty/usr/servers
/var/lib/openliberty/usr/servers/defaultServer
/var/lib/openliberty/usr/servers/defaultServer/apps
/var/lib/openliberty/usr/servers/defaultServer/dropins
/var/lib/openliberty/usr/servers/defaultServer/server.xml
/var/lib/openliberty/usr/servers/defaultServer/workarea
/var/lib/openliberty/usr/servers/defaultServer/workarea/.sLock
/var/lib/openliberty/usr/servers/defaultServer/workarea/org.eclipse.osgi
/var/lib/openliberty/usr/servers/defaultServer/workarea/org.eclipse.osgi/10
/var/lib/openliberty/usr/servers/defaultServer/workarea/org.eclipse.osgi/10/data
/var/lib/openliberty/usr/servers/defaultServer/workarea/org.eclipse.osgi/10/data/configStamp
/var/lib/openliberty/usr/servers/defaultServer/server.env
```

bensonlatibm:
#build

jjiwooLim (approved):
lgtm!

bensonlatibm:
#libby-bot

PR #19958, opened 2022-01-27
resource definitions in EJB deployment descriptor are not part of EE 10

njr-11:
Replace tests of the Concurrency 3.0 resource definitions being used in EJB deployment descriptors because it was just determined that the ejb-jar.xml schema is NOT being updated for EE 10.  The jakartaee-schemas project erroneously indicated EE 10 support within the ejb deployment descriptor,

```
 <xsd:include schemaLocation="jakartaee_10.xsd"/>
```

causing us to write these tests, but this was found to be an error and is being removed.

All of the tests for merging deployment descriptors with annotations should be done in web module for now.


njr-11:
#build

jmstephensgit (approved):

PR #19962, opened 2022-01-28
Ejb timer test

benjamin-confino:
Remove the old ejb timer test which as far as we can tell did not work.


benjamin-confino:
#build

Azquelt (approved):

PR #19965, opened 2022-01-28
Stop packaging OpenAPI UI source maps

### LINKED ISSUE #19931: Remove OpenAPI UI source maps
Although there's nothing secret in there as all the code is open source anyway, @mbroz2 pointed out that they're quite big and inflate the runtime size.

Azquelt:
The source maps for the OpenAPI UI are quite big and packaging them in the server increases the download size.

This PR removes them from the runtime while leaving them packaged in the source tree so they can be retrieved and loaded manually if required.

An alternative option would be to host them somewhere publicly available so that they can be linked to by the minified code without shipping them in the runtime.

Fixes #19931


Azquelt:
#build

Emily-Jiang (approved):

Azquelt:
The build has one failure:
```
 Errors found during packageVerify run:
  ERROR: wlp-bluemix-extended-22.0.0.2.jar(62851660 bytes) is less than 0.9 times the size of the compare file wlp-bluemix-extended-22.0.0.1.jar (70204708 bytes).
  ```

The drop in size is slightly larger than I'd anticipated, so I dug into it a bit:
Though size reduction wasn't a goal, it looks like #19535 reduced the size of each bundle from 2.9MB to 2.0MB
Removing the source maps in this PR then reduced the bundle size again from 2.0MB to 460kB

Between the three bundles which contain the UI, that's a 2.5MB * 3 = 7.5MB reduction in size for the files within this archive, which roughly corresponds with numbers in the error.

I assume this is the only affected package with a small enough total size to be pushed over the 10% reduction threshold by this PR.

tevans78:
#build

PR #19972, opened 2022-01-28
serialization of third-party context is not supported

njr-11:
The spec API does not provide any way to serialize or deserialize third-party context types that are captured for propagation, and requires that we reject the methods of ContextService that create serializable contextual proxies when such unserializable context is present.  This pull implements and tests that requirement.

### DIFF
--- a/dev/io.openliberty.concurrent/src/io/openliberty/concurrent/context/ThirdPartyContext.java
+++ b/dev/io.openliberty.concurrent/src/io/openliberty/concurrent/context/ThirdPartyContext.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2013 IBM Corporation and others.
+ * Copyright (c) 2012, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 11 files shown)


njr-11:
#build

njr-11:
#build

njr-11:
#run-libby-bot

dev/io.openliberty.concurrent/src/io/openliberty/concurrent/context/ThirdPartyContext.java
```diff
@@ -149,6 +150,26 @@ public ThirdPartyContext clone() {
...
+    @Override
+    @Trivial
+    public boolean isSerializable() {
+        UnsupportedOperationException x = null;
+        if (!propagated.isEmpty() || "propagated".equals(remaining)) {
+            List<String> unser = new ArrayList<String>(propagated); // TODO what if configured propagated types aren't available?
+            if ("propagated".equals(remaining)) // TODO avoid failing if no remaining types were found???
+                unser.add(ContextServiceDefinition.ALL_REMAINING);
+
+            x = new UnsupportedOperationException("Thread context configured for propagation is not serializable: " + unser); // TODO NLS
```
KyleAure, line 162: It looks like there should have been an `else` here that is missing. Otherwise, this will always throw an `UnsupportedOperationException`

njr-11: I want it to always throw the exception. The condition for entering this code block is `!propagated.isEmpty() || "propagated".equals(remaining)` which means that there is either a specific list of context types to propagate, or the ALL_REMAINING constant is specified for propagated context, or both.  Whenever there is configuration to propagate any third-party context type and make it Serializable, we need to reject it, which is accomplished with the exception.  The TODO covers some future optimizations where we might be able to detect that ALL_REMAINING wouldn't add any types, in which case we would need to fail, but that work is deferred to a future pull.

KyleAure (approved):

njr-11:
#build
#run-libby-bot

njr-11:
Build has 1 test failure, which is a duplicate of a preexisting issue in an unrelated test bucket

PR #19983, opened 2022-01-31
Metrics Default Exception Mapper to throw warning instead of IllegalArgumentException

### LINKED ISSUE #19843: Modify MP Metrics default exception mapper to throw IllegalArgumentException only if resource does not exist
Currently, when a JAX-RS resource that doesn't exist or an endpoint that doesn't exist is "hit" then an IllegalArgumentException is thrown by the `MetricsJaxRsEMCallbackImpl` class. This forces a `500` server response.

>"If an exception mapping provider (see Section 4.4) is available for the exception or one of its super- classes, an implementation MUST use the provider whose generic type is the nearest superclass of the exception to create a Response instance that is then processed according to Section 3.3.3. If the exception mapping provider throws an exception while creating a Response then return a server error (status code 500) response to the client. "

Modify the `MetricsJaxRsEMCallbackImpl` to only throw a warning.

Channyboy:
#build
fixes #19843

### DIFF
--- a/dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
+++ b/dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2021 IBM Corporation and others.
+ * Copyright (c) 2020, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


dev/io.openliberty.microprofile.metrics.internal.3.0.monitor/src/io/openliberty/microprofile/metrics/internal/monitor/MetricsJaxRsEMCallbackImpl.java
```diff
@@ -117,8 +125,8 @@ public void getSharedMetricRegistries(SharedMetricRegistries sharedMetricRegistr
 
 		if (fullMethodSignature == null || fullyQualifiedClassName == null || fullMethodSignature.isEmpty()
 				|| fullyQualifiedClassName.isEmpty()) {
-			throw new IllegalStateException("The following values are either null or empty - fullyQualifiedClassName: "
-					+ fullyQualifiedClassName + " fullMethodSignature: " + fullMethodSignature);
+		    Tr.warning(tc, "Could not resolve the class name or method signature class:[" + fullMethodSignature + "] method:[" +  fullMethodSignature+"]");
```
fmhwong: The warning message needs to be translated.

Channyboy: Since an empty `ResourceInfo` with no resource class or resource method originates from a non-existing URI, change to be a debug statement.

Channyboy: The new commit also formats the class, only change is to the `Tr.warning` to `Tr.debug`

fmhwong: Did you push your changes?

Channyboy:
#build #libby

fmhwong (approved):
LGTM

PR #19993, opened 2022-02-01
serialization permitted if no third-party context is propagated

njr-11:
Serializable contextual proxies should still work when third-party context is present but not configured for propagation.

### DIFF
--- a/dev/io.openliberty.concurrent/src/io/openliberty/concurrent/context/ThirdPartyContext.java
+++ b/dev/io.openliberty.concurrent/src/io/openliberty/concurrent/context/ThirdPartyContext.java
@@ -13,6 +13,7 @@ package io.openliberty.concurrent.context;
 import java.io.IOException;
 import java.io.ObjectInputStream.GetField;
 import java.io.ObjectOutputStream;
+import java.io.ObjectOutputStream.PutField;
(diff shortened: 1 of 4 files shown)


njr-11:
#build

dev/io.openliberty.concurrent/src/io/openliberty/concurrent/context/ThirdPartyContext.java
```diff
@@ -150,13 +149,42 @@ public ThirdPartyContext clone() {
         return clone;
     }
 
+    /**
+     * This method is invoked after deserialization to initializes the context types to clear.
```
nmittles, line 153: Should `initializes` be initialize?

nmittles (approved):

dev/io.openliberty.concurrent/src/io/openliberty/concurrent/context/ThirdPartyContext.java
njr-11: Yes.  It look like you've already approved and we have a green build for this, so I'll correct that typo in the following pull where I address the TODOs for enabling tests of deserializing unchanged transaction context.

PR #20032, opened 2022-02-03
fix assertion error in testTLD

### LINKED ISSUE #20023: assertion error in com.ibm.ws.jsp.2.3_fat
Describe the bug
In FAT com.ibm.ws.jsp.2.3_fat, test testTLD(), there can be an assert error:
testTLD:junit.framework.AssertionFailedError: ... Log should not contain SRVE0253I: Destroy successful.
The error is only happening on Windows platforms.

Steps to Reproduce
The test fails intermittently when using FULL mode.

Expected behavior
The test will be updated to avoid the assertion error which is caused by a unexpected compile of a jsp when it is called a 2nd time with no intervening dependent file updates.  This is similar to issue 19123 except here the failure is Windows specific.  An additional sleep will be added before the setMarkToEndOfLog() to ensure the dependent file timestamps being looked at are clearly outside of the mark window.

epj:
fixes #20023


epj:
#build

epj:
build is all green

volosied (approved):
I have some doubts (what's causing the recompile..?), but I think this change worth trying to see what happens.

PR #20037, opened 2022-02-03
Use the updated defaultValue of the variables configured during restore

anjumfatima90:


anjumfatima90:
#build

tjwatson (approved):
I approve to get this merged quickly.  Anjum can you open a separate issue to add a test that uses a defaultValue in the server.xml and then have a testcase that overrides the defaultValue with an ENV value an restore?

anjumfatima90:
Opened this issue #20071

PR #20051, opened 2022-02-05
Disable context pool for better avgs on UserEnum FAT

kristip17:
Averages still a little inconsistent for the disable test and making false negative failures. I disabled the searchCache, but should disable the contextPool as well so every call has to do a bind to the ldap.

RTC: 288790


kristip17:
#build

jvanhill (approved):

Issue #20076, opened 2022-02-08
document new key to retrieve third party id token from tokenMap in IDTokenMediator SPI


jimmy1wu:
pending feature approval and ga, document the new key to retrieve the third party id token from the tokenMap in the IDTokenMediator SPI to enable customizing the new ID token using the 3rd party ID token claims.

jimmy1wu:
new spi is no longer needed

the third party id token will be passed into `IDTokenMediator` through the existing `tokenMap` used in `IDTokenMediator.mediateToken(tokenMap);`

all that is needed for this issue is to just document the new key to retrieve the third party id token from `tokenMap` when the feature ga's

PR #20090, opened 2022-02-09
add new thirdpartycontext spi for oidc

jimmy1wu:
for #20076

adding a new `io.openliberty.security.openidconnect.spi.ThirdPartyContext​` SPI to be used by the `com.ibm.wsspi.security.openidconnect.IDTokenMediator` SPI implementations. this will allow the `IDTokenMediator` to be able to obtain the 3rd party ID token and fully customize the new ID token using the 3rd party token claims.​


jimmy1wu:
new spi is no longer needed

the third party id token will be passed into `IDTokenMediator` through the existing `tokenMap` used in  `IDTokenMediator.mediateToken(tokenMap);`

PR #20109, opened 2022-02-10
EJB persistent timers with concurrency 3.0

njr-11:
Ensure that EJB persistent timers work when used with the Concurrency 3.0 feature.
Persistent timers shouldn't invoke any of the new 3.0 methods, which aren't compatible with an implementation that is spread across multiple servers and has no way of knowing that completion happened elsewhere.


njr-11:
#build

jmstephensgit (approved):

PR #20116, opened 2022-02-11
discontinue clearing of MicroProfile context for EE10 and above

njr-11:
Now that Concurrency 3.0 in Jakarta EE 10 has its own mechanism for plugging in third-party context propagation, we should stop performing the automatic clearing of context via the similar MicroProfile mechanism because that might interfere with any type of context that provides itself as both.  This can be done on feature boundaries because MP Context Propagation 1.3 and lower apply to Jakarta EE 9 and lower.  We can move the MicroProfile context clearing service to an autofeature that only enables when MP Context Propagation 1.3 or lower is enabled, and then it will stay out of the way of Jakarta EE 10.


njr-11:
#build

jmstephensgit (approved):

njr-11:
#run-libby-bot
#build

PR #20223, opened 2022-02-22
Remove WL Resolver error reporting differences

cbridgha:


jimblye (commented):
System.out.println?
For commenting inside a method, better to use //

PR #20225, opened 2022-02-22
Add CDI core bucket

### LINKED ISSUE #20173: Add a new FAT bucket for golden path scenarios with current best practices
While we have lots of existing FAT tests for CDI, many of them run with `bean-discovery-mode=all`, which was commonly used when they were written.

Now, it's much more common to use the `annotated` bean discovery mode, and the CDI 4.0 spec encourages this by changing the behaviour of an empty beans.xml from discovery mode `all` to `annotated`.

We would like to have a set of tests covering the golden paths for CDI in the `annotated` discovery mode.

Areas to cover:
* Scopes (application, request and dependent)
* Producers
* Bean injection (field, method and constructor)
* Interceptors
  * bound on class and method, for `AROUND_INVOKE`
  * bound on class for `POST_CREATE` and `PRE_DESTROY`
* Events
* Programmatic lookup (using `CDI.current().select()`)

Azquelt:
Tests all golden path scenarios across all CDI versions using current
best practices.

In particular, it uses bean-discovery-mode=annotated, rather than all.

Fixes #20173


dev/io.openliberty.cdi.core_fat/bnd.bnd
```diff
@@ -0,0 +1,32 @@
...
+#*******************************************************************************
+
+-include= ~../cnf/resources/bnd/bundle.props
+bVersion=1.0
+
+src: \
+    fat/src
+
+tested.features:\
+    cdi-2.0,\
```
Emily-Jiang, line 19: what about cdi-1.2?

Emily-Jiang: I saw you mentioned EE7 later.

Azquelt: Features which are listed directly in the `server.xml` do not also get listed in `tested.features`.

dev/io.openliberty.cdi.core_fat/fat/src/io/openliberty/cdi/internal/core/injection/CDIInjectionTest.java
```diff
@@ -0,0 +1,48 @@
...
+    public static LibertyServer server;
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        server = FATSuite.server;
+
+        Package appPackage = CDIInjectionTestServlet.class.getPackage();
+        WebArchive war = ShrinkWrap.create(WebArchive.class, "cdiInjection.war")
+                                   .addPackage(appPackage)
+                                   .addAsWebInfResource(appPackage, "beans.xml", "beans.xml");
```
Emily-Jiang: I thought we said we don't really want any beans.xml. Can you delete this file?

dev/io.openliberty.cdi.core_fat/fat/src/io/openliberty/cdi/internal/core/interceptors/CDIInterceptorTest.java
```diff
@@ -0,0 +1,48 @@
...
+    public static LibertyServer server;
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        server = FATSuite.server;
+
+        Package appPackage = CDIInterceptorTestServlet.class.getPackage();
+        WebArchive war = ShrinkWrap.create(WebArchive.class, "cdiInterceptor.war")
+                                   .addPackage(appPackage)
+                                   .addAsWebInfResource(appPackage, "beans.xml", "beans.xml");
```
Emily-Jiang, line 38: delete this one

dev/io.openliberty.cdi.core_fat/fat/src/io/openliberty/cdi/internal/core/injection/CDIInjectionTest.java
Azquelt: Hmm, I added one because it's required for CDI Lite. I know our implementation doesn't require it but it would be best practise for writing a portable application, right?

dev/io.openliberty.cdi.core_fat/fat/src/io/openliberty/cdi/internal/core/interceptors/CDIInterceptorTest.java
Azquelt: Yes, if we decide not to include a beans.xml, I will remove it from all apps.

dev/io.openliberty.cdi.core_fat/bnd.bnd
Emily-Jiang: I figured out when I saw the last file.

dev/io.openliberty.cdi.core_fat/fat/src/io/openliberty/cdi/internal/core/injection/CDIInjectionTest.java
Emily-Jiang: ignore me as it will cause different behaviours for different versions.

Emily-Jiang: As for empty beans.xml behaviour for cdi 4.0, we need to add more tests on that instead.

Emily-Jiang (commented):
I did not find the programmatic api tests such as CDI.current() etc. I thought we agreed to have some tests on that, right?

dev/io.openliberty.cdi.core_fat/fat/src/io/openliberty/cdi/internal/core/injection/CDIInjectionTest.java
Azquelt: I think from CDI 1.2 onwards, having no beans.xml should always be interpreted as discovery-mode=annotated? It's only in CDI 1.0 and CDI Lite where having no beans.xml implies discovery-mode=none.

As you said, an empty beans.xml changed behaviour in 4.0 (from all to annotated) and we'll have specific tests for that elsewhere.

Azquelt:
> I did not find the programmatic api tests such as CDI.current() etc. I thought we agreed to have some tests on that, right?

It's in [CDILookupTestServlet](https://github.com/OpenLiberty/open-liberty/pull/20225/files#diff-5068f30274b1bc8c1fb6c5daa618e7e0592e059da0503bcb8278befd1bef6a98R28-R31)

It's a very short test, but I didn't think there was anything else to add there that would be considered golden path.

dev/io.openliberty.cdi.core_fat/fat/src/io/openliberty/cdi/internal/core/injection/CDIInjectionTest.java
Emily-Jiang: You were right. I confused about empty beans.xml and no beans.xml.

> Hmm, I added one because it's required for CDI Lite. I know our implementation doesn't require it but it would be best practise for writing a portable application, right?

We are not running in CDI Lite. We are on CDI full. I think we need to execute the tests without a beans.xml. Not sure whether you can do it with the current beans.xml and without beans.xml and they should produce the same result.

Azquelt: I've updated the test with a repeat rule to run the tests with and without a beans.xml.

Azquelt:
#build
#libby

Azquelt:
#build

Emily-Jiang (approved):
LGTM

PR #20226, opened 2022-02-22
built-in context type with properties

njr-11:
Add test coverage for a built-in thread context type which has its own properties. For this, we will copy from the z/OS WLM context, which is the only built-in type that I'm aware of that does this.


njr-11:
#build

njr-11:
#run-libby-bot
#build

KyleAure (approved):

PR #20230, opened 2022-02-22
Fix Ldap registry config

c00crane:
Fix the LDAP registry config - have it point to the correct real servers.
z/OS can't use the in memory servers, and ldap's 10 and 12 do not have the user that we need


c00crane:
#build

ayoho (approved):

PR #20242, opened 2022-02-23
Server config tool accessible name does not match or contain the visible label violation

lavenal:
This PR addresses the accessibility violation by replacing the aria-label with the alt attribute. Screen readers such as JAWS will continue to provide the additional non-screen text to users with disabilities.


lavenal:
#build

k8vance88 (approved):

lavenal:
Connection time out in the earlier PB.

lavenal:
#build

lavenal:
Two UI tests failed in the build, one in Explore tool and one in Server Config tool. We can safely ignore the failure in Explore tool as the code changes affect only Server Config tool. The test in Server Config failed when checking for the visibility of the Close button. The screen capture showed the Close button being there. Since the same test bucket was run twice, once in lite mode and once in full mode, the failed test passed in the lite mode. The code changes have nothing to do with the Close button. I believe it should be safe to ignore this failure.

lavenal:
#run-libby-bot

PR #20251, opened 2022-02-23
Swap to CustomRepository to stabilize UserEnum Test timing

kristip17:
Using LDAP has been unreliable to do our User Enum testing. Swap to a local Custom Repository and add sleeps between getting the user and doing a password check to simulate LDAP. There is a 1 second sleep at the beginning of the login and another 1 second sleep before we do the password check (if we found a user).

I copied `SampleCustomRepository` and pared it down for `SampleCustomRepository`

RTC 289475


kristip17:
#build

jvanhill (approved):

kristip17:
#libby

PR #20267, opened 2022-02-25
Issue #19618: Tighten control over lookup keys in CacheObject so it i…

### LINKED ISSUE #19618: SEC 380: Test verifying lookupKeys in CacheObject
We need to verify that the CacheObjects are being handled similarly to what they are in the in-memory authentication cache. It is possible that the in-memory authentication cache was using the same CacheObject for all keys that represented the same subject. If that is the case, then we might have issues when using JCache.

jvanhill:
…s clear that they are not updated in the (distributed) cache.

fixes #19618


jvanhill:
#build

utle (approved):

jvanhill:
#build

jvanhill:
#build

PR #20271, opened 2022-02-25
Explore tool accessibility fixes

lavenal:
This PR fixes the following accessibility issues in Explore admin center tool:

- img role with no label
  - use aria-labelledby attribute to point to an existing label
  - otherwise use aria-label
- accessible name not matching visible label text
  - replace aria-label with aria-labelledby attribute to point to an existing label
  - replace aria-label with alt attribute
  - remove unnecessary aria-label when there is visible text or title attribute

The Explore dashboard (for both collective and standalone server), collective view, and object view panels are all scanned with no accessibility violation. In addition, those panels are read by the JAWS screen reader to ensure the correct info is provided to users with disabilities.


lavenal:
#build

k8vance88 (approved):

lavenal:
#build

lavenal:
#build

lavenal:
#build

lavenal:
#run-libby-bot

PR #20281, opened 2022-02-28
Server config tool: Text contrast with its background is less than the WCAG AA minimum requirements

lavenal:
This PR fixes the accessibility violation of minimum contrast ratio requirement on the active tab in the server config tool.


lavenal:
#build

k8vance88 (approved):

PR #20288, opened 2022-03-01
Check if failure list is empty in FailoverTest

### LINKED ISSUE #20287: Guard against empty failure list in FailoverTest
In FailoverTest.check() in the com.ibm.ws.transaction.hadb_fat.derby.1 project, we need an extra guard in this code fragment....

            List<String> probableFailure = server.findStringsInLogs("WTRN0107W: Caught SQLException when opening SQL RecoveryLog");
            if (probableFailure != null) {
                fail(probableFailure.get(0));
            } else {
                fail(method + " did not return " + SUCCESS + ". Returned: " + sb.toString());
            }

The probableFailure list can be empty, so should check for that as well as the null test.

neilgsyoung:
fixes #20287


neilgsyoung:
#build

jonhawkes (approved):

PR #20326, opened 2022-03-04
improve trace of context propagation

njr-11:
It is difficult to debug trace of context being propagated to managed executor tasks and actions because of numerous misleading trace points for run, taskStarting, and taskStopping that don't actually indicate the running of the task.  Some of these come from PrivilegedAction classes, others from what is actually context being propagated or cleaned up some time before or after running the task.  There are also a lot of useless trace points that fill up the log and obscure the valuable information here.  This pull aims to suppress the useless information and improve on the beneficial trace points to provide a clearer and more concise picture of what is actually happening.


njr-11:
#build

jmstephensgit (approved):

Issue #20303, opened 2022-03-02
NPE during handshake when CLIENT_AUTH or SERVER_AUTH is missing in the certificate extension 


una-tapa:
**Describe the bug**  

NullPointerException is thrown in the following code path during SSLHandshake. 

```
product = WAS FOR Z/OS 21.0.0.12, z/OS Connect EE Unlimited 03.00.53 (wlp-1.0.59.cl211220211117-1256)
wlp.install.dir = /usr/lpp/IBM/zosconnect/v3r0/wlp/
server.config.dir = /var/zosconnect/servers/zceesrv/

[2/25/22 14:09:42:551 GMT] 00000036 id=00000000 SystemErr                                                    R Exception in thread "Default Executor-thread-4"
[2/25/22 14:09:42:551 GMT] 00000036 id=00000000 SystemErr                                                    R java.lang.NullPointerException
[2/25/22 14:09:42:551 GMT] 00000036 id=00000000 SystemErr                                                    R  at com.ibm.ws.ssl.core.WSX509TrustManager.processClientException(WSX509TrustManager.java:860)
[2/25/22 14:09:42:554 GMT] 00000036 id=00000000 SystemErr                                                    R  at com.ibm.ws.ssl.core.WSX509TrustManager.processClientTrustError(WSX509TrustManager.java:923)
[2/25/22 14:09:42:554 GMT] 00000036 id=00000000 SystemErr                                                    R  at com.ibm.ws.ssl.core.WSX509TrustManager.checkClientTrusted(WSX509TrustManager.java:699)
[2/25/22 14:09:42:555 GMT] 00000036 id=00000000 SystemErr                                                    R  at com.ibm.jsse2.z$c.a(z$c.java:74)
```

**Steps to Reproduce**  
Use certificate without  clientAuth and severAuth set in extKeyUsage
Perform SSL communication using the certificate above

**Expected behavior**  
SSLHandshake is performed without error. 

**Diagnostic information:**  
 - OpenLiberty Version: 21.0.0.12 
 - Affected feature(s) transportSecurity-1.0, ssl-1.0 

**Additional context**  
There must be an existing error thrown from Java.  
It is suspected that Java was looking CLIENT_AUTH or SERVER_AUTH in the certificate extension and threw an error as the usage is not found.  Liberty was trying to get the cause from the exception from Java. In this scenario, the cause was null.  The fix is to add null check.

una-tapa:
[...]

PR #20330, opened 2022-03-04
check for null cause before assignng

acdemyers:
Need to check for a null cause before assigning the cause.


acdemyers:
#build

una-tapa (approved):
Looks good to me. Thanks!

acdemyers:
fixes #20303

PR #20336, opened 2022-03-07
update AnnotationScanTest to provide more diagnostic info

WhiteCat22:
Modifying the assert method to indicate which check was done, javax or jakarta.

This is to help debug an intermittently failing test where the ```server.findStringsInLogs()``` call isn't finding the matching message in the logs, but I am able to manually match the method.

I suspect that maybe we may be trying to match the javax message when running EE9.


jim-krueger (approved):

WhiteCat22:
#run-libby-bot #build

jim-krueger (approved):

PR #20361, opened 2022-03-08
Update swagger-ui build and test instructions

### LINKED ISSUE #20321: Use personal build for OpenAPI UI manual test process
Currently, when we make changes to the OpenAPI UI, we test before the code is checked in, so that we can notice and fix any mistakes as early as possible.

However, this won't catch any mistakes which occur after the testing has been completed.

Update the process so that the manual testing uses the result of a personal build.

Azquelt:
* A personal build should be used to run the manual test, rather than a local build
* Add a list of expected file changes when we update dependencies

Fixes #20321


tevans78 (approved):

Azquelt:
#build

PR #20369, opened 2022-03-08
Skip prop test on z/OS

c00crane:
Having issues running propagation tests on z/OS as the in memory server used for the distributed platforms has a different set of users with extended attributes than the real servers that z/OS has to use.
I'd rather use in memory on dist and skip 3 tests on z/OS than make the distributed platforms use real ldap servers.  We just have too many issues keeping the ldap servers up and accessible.


c00crane:
#build

ayoho (approved):

PR #20378, opened 2022-03-09
Use BeanManager.getEvent() to fire events on CDI 2.0+

### LINKED ISSUE #20241: Create a CDI 4.0 version of the Transactions CDI extension
https://github.com/OpenLiberty/open-liberty/blob/312bbd1a865085ad90d88b0691b039727b7d91b5/dev/com.ibm.ws.transaction.cdi/src/com/ibm/tx/jta/cdi/TransactionContext.java#L99

TransactionContext makes use of BeanManager.fireEvent() which is a method that has now been removed in CDI 4.0. We need to create a new version of the bundle which uses BeanManager.getEvent().fire() instead.

tevans78:
The CDI 1.2 version of the transactions CDI extension used BeanManager.fireEvent(). That method was deprecated in CDI 2.0 and removed in CDI 4.0. This PR allows the transactions CDI extension to use the newer method on CDI 2.0 and above.

#build


dev/io.openliberty.transaction.internal.cdi12/bnd.bnd
```diff
@@ -0,0 +1,46 @@
...
+-buildpath: \
+        com.ibm.wsspi.org.osgi.service.component.annotations,\
+        com.ibm.websphere.appserver.spi.logging;version=latest,\
+        com.ibm.websphere.javaee.cdi.1.2;version=latest,\
+        com.ibm.tx.jta;version=latest,\
+        com.ibm.tx.util;version=latest,\
+        com.ibm.ws.cdi.interfaces;version=latest, \
+        com.ibm.ws.transaction.cdi;version=latest
+
+jakartaeeMe: true
```
tevans78: This bundle does not need to be transformed

tevans78:
fixes #20241

jonhawkes (approved):
Can't fault this by inspection. Please ensure your PB runs com.ibm.ws.transaction.core_fat.8 successfully.

dev/com.ibm.ws.transaction.cdi/bnd.bnd
```diff
@@ -55,4 +55,4 @@ Export-Package: \
 	../build.sharedResources/lib/junit/old/junit.jar;version=file, \
 	com.ibm.ws.logging;version=latest
 
-jakartaeeMe: true
+test.project: true
```
tevans78: Not 100% sure that this is the right thing to put here ... but it does mean that the `com.ibm.ws.transaction.cdi` bundle will never be copied into wlp/lib, which would be bad!

tevans78:
#build

dev/com.ibm.ws.transaction.cdi/bnd.bnd
Azquelt: It looks like other projects which contain product code but don't themselves get included in the image use `publish.wlp.jar.disabled: true`.

I would worry that `test.project: true` may have other side effects.

Azquelt: Might also be worth a comment somewhere if you're sucking out the classes into both of the auto-feature bundles.

tevans78:
```suggestion
# This bundle is not shipped itself. Instead, the classes are included in one of two other bundles, depending on the
# cdi version being used: io.openliberty.transaction.internal.cdi12 or io.openliberty.transaction.internal.cdi20
publish.wlp.jar.disabled: true
```

Azquelt (approved):

tevans78:
#libby

PR #20399, opened 2022-03-11
Add test for configuration of product extension on restore

tjwatson:


tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1213,11 +1213,6 @@ serverCmd()
 #        checkpoint/workarea contains the backup of workarea contents, less the checkpoint dir
 backupWorkarea()
 {
-  if $X_WLP_IN_CONTAINER
```
sebratton, line 1216: I think this was the only place where behavior varied based on WLP_IN_CONTAINER. Do you intend to leave the reference to it in the file header and where it's initialized in this script? I suppose there are also external uses of WLP_IN_CONTAINER which are now obosolete unless you think the behavior may vary in the future

tjwatson: Right, I left the setting of the environment variable in case we need it later.  But as of now I don't think there are any other places this variable would get used.  If you prefer I can remove it.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/TestSPIConfig.java
```diff
@@ -0,0 +1,142 @@
...
+    @BeforeClass
+    public static void addTestFeature() throws Exception {
+        server.copyFileToLibertyInstallRoot(USER_FEATURE_PATH, USER_FEATURE_USERTEST_MF);
+        server.copyFileToLibertyInstallRoot(USER_BUNDLE_PATH, USER_FEATURE_USERTEST_JAR);
+
+    }
+
+    @AfterClass
+    public static void cleanup() throws Exception {
+        //server.deleteFileFromLibertyInstallRoot(USER_FEATURE_PATH + USER_FEATURE_USERTEST_MF);
```
sebratton: Did you forget to comment these back in?

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
tjwatson: I removed the WLP_IN_CONTAINER stuff altogether.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/TestSPIConfig.java
tjwatson: yes, should be fixed now.

tjwatson:
#build

sebratton (approved):

tjwatson:
#build

tjwatson:
#libby

PR #20442, opened 2022-03-14
timing window where task cancel is ignored

### LINKED ISSUE #20441: Timing window where cancellation of scheduled task is ignored
**Describe the bug**  
There is a brief timing window between the completion (or skip) of a ManagedScheduledExecutorService task execution and the scheduling of the next execution where a cancel that is issued can be overwritten and ignored.

Due to this issue, you might intermittently see subsequent executions of a scheduled task be attempted even though the application has requested to cancel the task.

**Steps to Reproduce**  
Due to the brief timing window, the problem is intermittent and extremely unlikely to reproduce.

**Expected behavior**  
The scheduled task should stop attempting subsequent executions after it is canceled.

**Diagnostic information:**  
 - OpenLiberty Version: up to 22.0.0.3.
 - Affected feature(s) concurrent-1.0, concurrent-2.0
 - Java Version: all
 - server.xml configuration ` <feature>concurrent-1.0</feature>` or ` <feature>concurrent-2.0</feature>`

**Additional context**  
The issue appears to be due to code that is using `AtomicReference.set` rather than `AtomicReference.compareAndSet` in order to take into account the possibility of an intermediate state change.

njr-11:
fixes #20441

### DIFF
--- a/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ScheduledTask.java
+++ b/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ScheduledTask.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2013, 2021 IBM Corporation and others.
+ * Copyright (c) 2013, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


njr-11:
#build

KyleAure (approved):

PR #20465, opened 2022-03-15
cleared context defaults to Transaction

njr-11:
When cleared context is absent from a context-service definition in a deployment descriptor, it must default to clearing Transaction context, per the xsd documentation, which also aligns with the default from the ContextServiceDefinition annotation.


njr-11:
#build

KyleAure (approved):

njr-11:
There were no test failures, but child build 12 initially failed due to an infrastructure error.  After resubmitting, it ran cleanly as well.

PR #20470, opened 2022-03-15
Reduce to 1 sleep for Custom Repo on UserEnumTest

kristip17:
Adjust the sleep time in the Custom test repo for the `UserEnumerationTest` -- we really only need a sleep between the search and login to create an exaggerated difference in the login (originally tested out with 2 sleeps). Also adjusted the `testUserEnumerationCustomConfig` so the window hopefully provides more consistent results on assorted test environments.

RTC 289674


kristip17:
#build

jvanhill (approved):

PR #20471, opened 2022-03-15
Pre-initialize statics for ORB ValueHandler and other classes during applications checkpoint

dettmoney:
I'm doing criu testing with the daytrader7 application (specifically the ejb3/PingServlet2Session/Remote servlet), trying to reduce restore times by moving work into the checkpoint phase. Profiles show javax/rmi/CORBA/Util.createValueHandler() and org/apache/yoko/rmi/impl/ValueHandlerImpl.getRunTimeCodeBase() calling static initializers during the first request to the application. We can move those static initializers into the checkpoint time by adding the following call to ORBWrapperInternal.activate:  _Util.createValueHandler().getRunTimeCodeBase()_. This simple change give us a 5-10% reduction in restore time. We will only do this when doing an APPLICATIONS checkpoint, as that is the only phase that shows the gain.


dettmoney:
#build

sebratton (approved):
Checkpoint logic looks good to me. Can't speak for semantics of javax.rmi.CORBA.Util

dettmoney:
#build
#libby

dettmoney:
#build
#libby

PR #20507, opened 2022-03-18
merging for context service definition

njr-11:
Implement merging of the context type lists of context service definition between the context-service in web.xml and the ContextServiceDefinition annotation in the web module.  For each category (propagated, cleared, unchanged), the deployment descriptor should override the entire context list if configured there.   Ensure that we can distinguish unspecified from the default value.

### DIFF
--- a/dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ContextServiceDefinitionBinding.java
+++ b/dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ContextServiceDefinitionBinding.java
@@ -11,12 +11,16 @@
 package io.openliberty.concurrent.internal.processor;
 
 import java.lang.reflect.Member;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


njr-11:
#build

dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ContextServiceDefinitionBinding.java
```diff
@@ -11,12 +11,16 @@
 package io.openliberty.concurrent.internal.processor;
 
 import java.lang.reflect.Member;
+import java.util.Arrays;
```
nmittles, line 14: Update copyright date.

nmittles (approved):
Looks good.  Marked one file that should have the copyright date updated.  This can be done in a future PR.

njr-11:
#build

njr-11:
#build

njr-11:
Original build had 4 failures, all of which are pre-existing issues.  3 of these had not previously been reported, so builds were initially resubmitted, but those errors can now be seen in builds from other developers as well who do not have any of these changes, so they are clearly pre-existing issues.

PR #20515, opened 2022-03-21
Add abort for in use connections

### LINKED ISSUE #20726: Add maxInUseTime duration to abort connections.
Proposal:  Add maxInUseTime to configuration for supporting idle time out of in use connections. 

As a user of data source connections, I want the connection pool to automatically remove connections that normally would be destroyed, but are not being removed from the connection pool after a specified amount of time.

The most common way a bad connection is removed from the connection pool is when a connection error event has occurred.  The error event normally occurs while the connection is used by the application.  A in use bad connection may wait until the connection is closed and transaction completes before being removed from the connection pool.

Customers due to various reasons encounter in use bad connections that will not be removed from the connection pool.   The most common cause of the problem is a missing connection close on an error path in the customers application.  Other causes can be due to jdbc driver code bugs that can leave connections in unexpected states.  Less common problems are timing issue when the root cause of the problem occurs (database, network, security, etc.), resulting in connection not being removed from the connection pool.

For datasources that can encounter connections that are in a remove connection pool state, but are not being removed from the connection pool, customers will be able to provide a duration time to abort connections when they exceed the maxInUseTime.

New property,

maxInUseTime=Remove long running in use connections
maxInUseTime.desc=This property is a temporary workaround for application code that does not close connections. It configures an amount of time after which pool maintenance can invoke Connection.abort on a connection that is being used by the application. Stopping connections that are in use can result in unexpected and unpredictable threading errors. The proper solution is to update the application to always close connections and not use this property. If you must configure it as a workaround, choose a value of 20 minutes or more depending on the normal application transaction times. If transaction times are high, 10 minutes or more, then double your longest running transaction time for this value.

Note, additional properties can be added to provide more ways of aborting bad connection.

jmstephensgit:


dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -28,6 +28,9 @@ connectionManager.desc=Connection Manager configuration
 agedTimeout=Aged timeout
 agedTimeout.desc=Amount of time before a connection can be discarded by pool maintenance. A value of -1 disables this timeout. A value of 0 discards every connection, which disables connection pooling.
 
+abortLongRunningInuseConnections=Abort long running in use connections
+abortLongRunningInuseConnections.desc=Amount of time an in use connection can be abort by pool maintenance. A value of -1 disables this timeout. Aborting connections that are in use can result in unexpected behaviors,  The recommended value is 20 minutes or more depending on the applications normal transaction times.  If transaction times are high, 10 minutes or more, then double your longest running transaction time for this value.
```
KyleAure: The first sentence is a bit confusing. Perhaps reword as: 
The amount of time before an in-use connection can be aborted by pool maintenance.

Also throughout in-use should be hyphenated.

dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/ConnectionManagerServiceImpl.java
```diff
@@ -538,6 +542,9 @@ private J2CGlobalConfigProperties processServerPoolManagerProperties(AbstractCon
             if (pm.gConfigProps.getAgedTimeout() != agedTimeout)
                 pm.gConfigProps.setAgedTimeout(agedTimeout);
 
+            if (pm.gConfigProps.getAbortLongRunningInuseConnections() != abortLongRunningInuseConnections)
+                pm.gConfigProps.setAbortLongRunningInuseConnections(abortLongRunningInuseConnections);
```
KyleAure: Be careful with this extra line between if statement and set.  Without {} you may call `setAbortLongRunningInuseConnections` regardless of the condition.

dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/J2CGlobalConfigProperties.java
```diff
@@ -996,4 +1027,20 @@ public void throwExceptionOnMCThreadCheck(
         changeSupport.firePropertyChange("throwExceptionOnMCThreadCheck", this.throwExceptionOnMCThreadCheck, throwExceptionOnMCThreadCheck);
         this.throwExceptionOnMCThreadCheck = throwExceptionOnMCThreadCheck;
     }
+
+    /**
+     * @return
+     */
+    public int getAbortLongRunningInuseConnections() {
+        // TODO Auto-generated method stub
```
KyleAure: You can get rid of this TODO

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
KyleAure: What should a value of 0 mean? Or should we ensure that this value is >= reap time?

KyleAure (changes requested):

jmstephensgit:
#build

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -28,6 +28,9 @@ connectionManager.desc=Connection Manager configuration
 agedTimeout=Aged timeout
 agedTimeout.desc=Amount of time before a connection can be discarded by pool maintenance. A value of -1 disables this timeout. A value of 0 discards every connection, which disables connection pooling.
 
+abortLongRunningInuseConnections=Abort long running in use connections
```
helyarp: Replace "abort" with "cancel," "stop," "end," or "fail."

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -28,6 +28,9 @@ connectionManager.desc=Connection Manager configuration
...
+abortLongRunningInuseConnections.desc=The amount of time before an in-use connection can be aborted by pool maintenance.  A value of -1 or 0 disables this timeout. Aborting connections that are in use can result in unexpected behaviors,  The recommended value is 20 minutes or more depending on the applications normal transaction times.  If transaction times are high, 10 minutes or more, then double your longest running transaction time for this value.
```
helyarp: Suggested change to remove "abort," fix spacing, etc.:

The amount of time before pool maintenance can stop an in-use connection. A value of -1 or 0 disables this timeout. Stopping connections that are in use can result in unexpected behaviors. The recommended value is 20 minutes or more depending on the normal application transaction times. If transaction times are high, 10 minutes or more, then double your longest running transaction time for this value.

helyarp (commented):
Suggested changes

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
njr-11: If we externalize/document this capability, I would like to call it maxInUseTime "Maximum In Use Time" (or pick some other synonym for "in use", in order to align with our other connection manager properties like maxIdleTime and maxPoolSize.

KyleAure (dismissed):

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
njr-11: To align with other timeout properties in Liberty, use -1 to disable.  Disallow 0.  The recommendation is to not set it all (or -1 disabled) because it is unsafe.  Here is a suggestion:

"This property is a temporary workaround for application code that does not close connections. It configures an amount of time after which pool maintenance can invoke Connection.abort on a connection that is being used by the application. Stopping connections that are in use can result in unexpected and unpredictable threading errors. The proper solution is to update the application to always close connections and not use this property. If you must configure it as a workaround, choose a value of 20 minutes or more depending on the normal application transaction times. If transaction times are high, 10 minutes or more, then double your longest running transaction time for this value."

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -22,6 +22,7 @@
 
  <OCD id="com.ibm.ws.jca.connectionManager" name="%connectionManager" description="%connectionManager.desc" ibm:alias="connectionManager" ibmui:localization="OSGI-INF/l10n/metatype">
   <AD id="agedTimeout"                        name="%agedTimeout"  description="%agedTimeout.desc"  default="-1"     type="String"  ibm:type="duration(s)" min="-1"/>
+  <AD id="abortLongRunningInuseConnections"   name="abortLongRunningInuseConnections"  description="%abortLongRunningInuseConnections.desc"  default="-1"     type="String"  ibm:type="duration(s)" min="-1"/>
```
njr-11: Consider changing to minutes `ibm:type="duration(m)"` in order to prevent the configuration of values like 1 second, 5 seconds, or 10 seconds.

dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/ConnectionManagerServiceImpl.java
```diff
@@ -493,6 +495,8 @@ private J2CGlobalConfigProperties processServerPoolManagerProperties(AbstractCon
             numConnectionsPerThreadLocal = 0;
         } else {
             agedTimeout = validateProperty(map, J2CConstants.POOL_AgedTimeout, -1, TimeUnit.SECONDS, -1, Integer.MAX_VALUE, true, connectorSvc);
+            abortLongRunningInuseConnections = validateProperty(map, J2CConstants.POOL_AbortLongRunningInuseConnections, -1, TimeUnit.SECONDS, -1, Integer.MAX_VALUE, true,
```
njr-11: reject 0

dev/com.ibm.ws.jdbc_fat_v41/test-applications/basicfat/src/jdbc/fat/v41/web/BasicTestServlet.java
```diff
@@ -915,6 +918,47 @@ public void getSingleConnectionAfterAbort() throws Exception {
...
+    public void testAbortedConnectionDestroyedAuto() throws Exception {
+        Connection c = ds2abort.getConnection();
+        Executor testExecutor = new Executor() {
+            @Override
+            public void execute(Runnable arg0) {
+                arg0.run();
+            }
+        };
+        try {
+            Thread.sleep(15000);
```
njr-11: Vulnerable to test infrastructure slowness.  Can we switch to periodic polling via the mbean to identify when the connection pool size is dropped, allowing at least 2 minutes?  that would be less vulnerable to slowness, and also should typically run much faster because the polling can detect the successful result much sooner.

helyarp (commented):
Looks good!

jmstephensgit:
#build

dev/com.ibm.ws.jdbc_fat_v41/test-applications/basicfat/src/jdbc/fat/v41/web/BasicTestServlet.java
```diff
@@ -915,6 +926,123 @@ public void getSingleConnectionAfterAbort() throws Exception {
...
+     * @throws Exception
+     */
+    @Test
+    public void getSingleConnectionAfterAbortAuto60Seconds() throws Exception {
+        Connection c = ds2abort60s.getConnection();
+        c.close();
+    }
+
+    /**
+     * Test using maxInUseTime="1m"
```
njr-11, line 1036: There is no point in testing both
maxInUseTime="60s"
and
maxInUseTime="1m"
especially when the test takes so long to run.
The duplication would only be testing the ibm:type=duration({UNIT}) capability of Liberty config.

Even with just 1 test, I'd recommend designating it for FULL mode.

skasund:

jmstephensgit:
#build

PR #20519, opened 2022-03-21
Work around issue with JSONB TCK and JDK 13+

KyleAure:
Per discussion here: https://github.com/eclipse-ee4j/jsonb-api/issues/272


KyleAure:
#build

nmittles (approved):

KyleAure:
The first build was successful on JDK 11
The second build was successful on JDK 17
The second build had a known failure with one of the child builds, but the JDK 17 one was successful.

PR #20529, opened 2022-03-22
Removing the problematic pom file from built jar

cbridgha:
Making sure the Meta-Inf/maven directory isn't shipped in built bundle.   Not used in build.


cbridgha:
#build

Issue #18191, opened 2021-08-13
 Create schemaGen wrapper script for ws-schemagen.jar


cbridgha:
## Description of the high level feature, including any external spec links:  

Liberty has many executable tool JARs that also have wrapped scripts handling command line help, error codes, env vars etc...

ws-schemagen.jar also needs a script that helps pass supported / documented parameters.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## When complete & mandatory, add links to the UFO (Upcoming Feature Overview) document, FTS (Feature Test Summary), blogs post issues(s), and [Aha](https://cloud-platform.ideas.aha.io/) (externally raised RFEs):
- UFO:  N/A
- FTS:  https://github.com/OpenLiberty/open-liberty/issues/19967
- Beta Blog Post (if applicable): N/A
- Blog Post:  https://github.com/OpenLiberty/open-liberty/issues/20929
- Aha: 
   - Also, add a comment into the Aha idea linking to this issue.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## List of Steps to complete or get approvals / sign-offs for Onboarding to the Liberty release (GM date)

Instructions:
- Do the actions below and mark them complete in the checklist when they are done.
- Make sure all feature readiness approvers put the appropriate tag on the epic to indicate their approval.

## **Design**
Before Development Starts or 8 weeks before Onboarding
- [ N/A] POC Design / UFO Review Scheduled (David Chang) or N/A.
- [ N/A] POC Design / UFO Reviewed (Feature Owner) or N/A.
- [ N/A] Complete any follow-ons from the POC Review.
- [ N/A] Design / UFO Approval ([Alasdair Nottingham](https://github.com/NottyCode)) or N/A.
- [ X ] No Design / No UFO Approval ([Alasdair Nottingham](https://github.com/NottyCode)) or N/A.
  * Note: Design is required if any of the following apply: SVT or Performance testing, new UI, Servicibility considerations, significant documentation changes
- [ x] ID Requirements identified ([Documenting Open Liberty](https://github.com/OpenLiberty/open-liberty/wiki/Documenting-Open-Liberty)). (Epic owner / Feature owner with ID focal point)
[...]

PR #20545, opened 2022-03-23
Create a unit test for schemaGen script

jimblye:
I added 3 tests for the wlp/bin/schemaGen script which was added by issue #18191 .

This is the output from the tests when they run successfully on Linux.  I still need to try it on Windows.

==================== testSchemaGenNoParms ...

The schemaGen command generates the XML schema for the                       
Liberty core configuration and other installed products                
extensions in a single output file.

Usage: ./schemaGen [options] outputFile

PASSED
==================== testSchemaGenHelp ...

The schemaGen command generates the XML schema for the                       
Liberty core configuration and other installed products                
extensions in a single output file.

Usage: ./schemaGen [options] outputFile

Options:

--compactOutput
The output schema will not contain indenting 
spaces, new line feeds, or XML comments.

--encoding
The character encoding to use for the output.

--ignorePidsFile
A file name containing a list of pids to ignore.

--locale
The language to use when you are creating the output
file. This string consists of the ISO-639 two-letter lowercase language
code, optionally followed by and underscores and the ISO-3166 uppercase
two-letter country code.

--outputVersion
If outputVersion=2.0 is specified, only the xsd:any element is used    
in the output file, so that unknown elements pass XSD validation at    
the expense of losing validation for known elements.

--schemaVersion
If schemaVersion=1.1 is specified, then both explicitly named child    
elements and the xsd:any element are written to the output file.

PASSED
==================== testSchemaGenOutput ...
PASSED

### DIFF
--- a/dev/com.ibm.ws.config.schemagen/test/com/ibm/ws/config/schemagen/internal/SchemaGenTest.java
+++ b/dev/com.ibm.ws.config.schemagen/test/com/ibm/ws/config/schemagen/internal/SchemaGenTest.java
@@ -27,12 +27,13 @@ import org.junit.Test;
  */
 public class SchemaGenTest {
 
-    public static String WLP_BIN_DIR = "../build.image/wlp/bin";
-    public static String OUTPUT_FILE = "schemaGenOutput.xsd";
-    public static String HELP_OPTION = "-help";
-    public static String SCHEMAGEN_BAT = "./schemaGen.bat";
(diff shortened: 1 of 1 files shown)


jimblye:
#build #libby

hlhoots (approved):
Seems good to me.  Would run a Win build to make sure that path works.

jimblye:
#run-libby-bot

jimblye:
#libby

PR #20551, opened 2022-03-23
ensure PrometheusBuilder does not make separate retrieveal of SimpleTimer max/min values

### LINKED ISSUE #20476: NPE when outputting SimpleTimer close to the end of a full minute.
**Describe the bug**  
When the `/metrics` endpoint is hit and a `SimpleTimer` metric is being formatted there are two explicit calls made to retrieve the value of the `SimpleTimer` (of the min time duration or max time duration). It could be that these two calls are made before and after the completion of a full minute. This might lead to the a `NullPointerException` if the SimpleTimer removes the value as part of the completion of the full minute (i.e. in that the previous full minute did not record any values to update the min/max values)

```
SRVE0777E: Exception thrown by application class 'io.openliberty.microprofile.metrics30.internal.helper.PrometheusBuilder30.buildSimpleTimer30:111'
java.lang.NullPointerException
	at io.openliberty.microprofile.metrics30.internal.helper.PrometheusBuilder30.buildSimpleTimer30(PrometheusBuilder30.java:111)
	at io.openliberty.microprofile.metrics30.internal.writer.PrometheusMetricWriter30.writeMetricMapAsPrometheus(PrometheusMetricWriter30.java:101)
	at io.openliberty.microprofile.metrics30.internal.writer.PrometheusMetricWriter30.writeMetricsAsPrometheus(PrometheusMetricWriter30.java:47)
	at com.ibm.ws.microprofile.metrics.writer.PrometheusMetricWriter.write(PrometheusMetricWriter.java:83)
	at com.ibm.ws.microprofile.metrics.BaseMetricsHandler.handleRequest(BaseMetricsHandler.java:83)
```

**Steps to Reproduce**  
There is no concrete steps to reproduce. But if a query to `/metrics` at the exactly right time around the completion of a full minute. It could also be that the thread is interrupted and stops close to the full minute completion and resumes after in-between the two method calls.

**Expected behavior**  
There should not be a NullPointerException

**Diagnostic information:**  
 - Affected feature(s): `mpMetrics-3.0` and `mpMetrics-4.0`

Channyboy:
, but rather use a saved value when formatting

#build 
fixes #20476

### DIFF
--- a/dev/io.openliberty.microprofile.metrics.internal.3.0/src/io/openliberty/microprofile/metrics30/internal/helper/PrometheusBuilder30.java
+++ b/dev/io.openliberty.microprofile.metrics.internal.3.0/src/io/openliberty/microprofile/metrics30/internal/helper/PrometheusBuilder30.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2020 IBM Corporation and others.
+ * Copyright (c) 2017, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


Channyboy:
No errors encountered in the FAT tests

pgunapal (approved):
LGTM

PR #20552, opened 2022-03-23
CDI Startup and Shutdown events

tevans78:
Pull in Weld 5.0.0.CR2 and then fire the Startup & Shutdown events ourselves


tevans78:
#build

tevans78:
#build

tevans78:
#build

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/startupEvents/StartupEventsTest.java
```diff
@@ -0,0 +1,81 @@
...
+    @Server(SERVER_NAME)
+    @TestServlets({
+                    @TestServlet(servlet = StartupEventsServlet.class, contextRoot = STARTUP_EVENTS_APP_NAME)
+    })
+    public static LibertyServer server;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+
+        JavaArchive startupEventsSharedLib = ShrinkWrap.create(JavaArchive.class, STARTUP_EVENTS_APP_NAME + "SharedLib.jar")
```
Emily-Jiang, line 52: I am wondering whether we should also add a test with 2 wars as well. thoughts?

Emily-Jiang: Can you also test the shutdown event?

tevans78: > I am wondering whether we should also add a test with 2 wars as well. thoughts?

It would be easy to add a second war module. I don't think it would prove much but since it is so easy there is no reason not to.

tevans78: > Can you also test the shutdown event?

Although the test is called "startupEvents" it does check for shutdown too. I might be able to make that clearer.

Emily-Jiang (approved):
LGTM

tevans78:
#build

tevans78:
#libby

PR #20557, opened 2022-03-23
18105 20284 logout token validation

c00crane:
Create tests to:
1) Verify the logout_token validation code in the RP

- test code creates logout_tokens with valid values and expects the RP to handle the token as valid
- test code creates logout_tokens with invalid content (missing claims, dis-allowed claims, claim values that are not correct for the openidconnect (RP) client config.

2) Validate the content in the logout_token created by the OP
- The OP has multiple providers with 1-n clients using various configurations. - tests validate that the logout_token created is correct for the config in the provider


c00crane:
commits containing id will be removed before delivering - those will be delivered by Adam under different pull(s).

skasund:
L2 message review completed

c00crane:
Delivered with another pull

PR #20558, opened 2022-03-23
enable Concurrency TCK to run with other EE 10 features

njr-11:
The Concurrency TCK temporarily runs with concurrent-3.0 (EE 10 level) but other features (like servlet) at the EE 9 level, because those features are available as beta or ga.  This pull makes it possible to run with all EE 10 features, although we will not switch to that yet because those features are non-ship and we should be using at least beta to collect official TCK results.


njr-11:
#run-libby-bot

KyleAure (approved):

njr-11:
#build

njr-11:
#run-libby-bot
#build

njr-11:
#build

Issue #20368, opened 2022-03-08
CRIU : EJBContainer settings for deferred EJB initialization and bean pool preloading


tkburroughs:
There are 3 EJBContainer settings that should improve restore performance by enabling additional application initialization prior to checkpoint that occurs after application start:

**1 - startEJBsAtAppStart in server.xml (or com.ibm.websphere.ejbcontainer.initializeEJBsAtStartup ) as a JVM property.**
```
    <ejbContainer startEJBsAtAppStart="true"/>
```
or, in `ibm-ejb-jar-ext.xml`
```
<session name="AccountServiceBean">
 <start-at-app-startvalue="TRUE"/>
</session>
```
The default is "false", but switching the default to "true" should be considered for the checkpoint scenario.  A customer could specifically set it to "false" in `ibm-ejb-jar-ext.xml` to disable for a bean.  Should decide what effect setting this in server.xml would have for checkpoint (would checkpoint override?)

For testing, I'd recommend the setting in server.xml (or as a JVM property) for now, but be aware the setting in ibm-ejb-jar-ext.xml would override it.

Recommendation:
- EJBContainer can check for "checkpoint" and change the default setting for this config option if not explicitly disabled

**2 - com.ibm.websphere.ejbcontainer.poolSize as a JVM property.**
```
com.ibm.websphere.ejbcontainer.poolSize=*=H50,500
```
This will preload the bean pools with 50 instances per stateless bean type.  The pool may grow to a max of 500.

The default is 50,500, adding the H makes it a "hard limit", so is a request to preload to the minimum value.  Default is to start the pool at 0 and let it grow as used. Adding an H to the second value (500) would set a hard max limit, so the EJB Container would never create more than that; generally not recommended.

The * is the bean type (all), so you may specify this for individual beans with a comma separate list.

Note: the bean pools are "preloaded" using a background thread, so the application may report that it has started before the pools have been preloaded. 

- For the CRIU scenario, perhaps the default should change to `H50,500`.  This would be overridden by the JVM setting.
- EJBContainer should check if the server is for "checkpoint", and then preload the pools synchronously, rather than asynchronously.

**3 - `@Startup` or ` <init-on-startup>` for Singleton beans to force preCreate of bean.**

An instance of a Singleton bean is normally created on first use, but an app developer may add `@Startup` to force the bean to be created during application start.  Similar behavior could be added automatically when creating a checkpoint.

[...]

PR #20595, opened 2022-03-25
Issue #20368: Default EJBInitAtStartup for CRIU

tkburroughs:
For CheckpointPhase APPLICATIONS, EJBInitAtStart will default to true
unless explicitly configued otherwise

for #20368

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
+++ b/dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2021 IBM Corporation and others.
+ * Copyright (c) 2012, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -172,6 +172,8 @@ import com.ibm.wsspi.kernel.service.utils.AtomicServiceReference;
 import com.ibm.wsspi.kernel.service.utils.OnErrorUtil.OnError;
 import com.ibm.wsspi.kernel.service.utils.ServerQuiesceListener;
 
+import io.openliberty.checkpoint.spi.CheckpointPhase;
+
 @Component(service = { ApplicationStateListener.class, DeferredMetaDataFactory.class, EJBRuntimeImpl.class, ServerQuiesceListener.class },
            configurationPid = "com.ibm.ws.ejbcontainer.runtime",
            configurationPolicy = ConfigurationPolicy.REQUIRE,
@@ -254,6 +256,13 @@ public class EJBRuntimeImpl extends AbstractEJBRuntime implements ApplicationSta
     private static final String BIND_TO_JAVA_GLOBAL = "bindToJavaGlobal";
     private static final String DISABLE_SHORT_DEFAULT_BINDINGS = "disableShortDefaultBindings";
     private static final String CUSTOM_BINDINGS_ON_ERROR = "customBindingsOnError";
+
+    private final CheckpointPhase checkpointPhase;
+
+    @Activate
+    public EJBRuntimeImpl(@Reference(cardinality = ReferenceCardinality.OPTIONAL) CheckpointPhase checkpointPhase) {
+        this.checkpointPhase = checkpointPhase;
+    }
 
     @Override
     public void serverStopping() {
@@ -1681,6 +1690,28 @@ public class EJBRuntimeImpl extends AbstractEJBRuntime implements ApplicationSta
                 remoteLatch.countDown();
             }
         }
+    }
+
+    /**
+     * Determines if application start should optimize for checkpoint after all configured applications
+     * have been identified, but before any application code has been called.
+     *
+     * @return true if application start should optimize for checkpoint deployment; false otherwise.
+     */
+    @Override
+    public boolean isCheckpointDeployment() {
+        return CheckpointPhase.DEPLOYMENT == checkpointPhase;
+    }
+
+    /**
+     * Determines if application start should optimize for checkpoint after all configured applications
+     * have started.
+     *
+     * @return true if application start should optimize for checkpoint applications; false otherwise.
+     */
+    @Override
+    public boolean isCheckpointApplications() {
+        return CheckpointPhase.APPLICATIONS == checkpointPhase;
     }
 
     @Override
--- a/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/EJSHome.java
+++ b/dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/EJSHome.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1998, 2021 IBM Corporation and others.
+ * Copyright (c) 1998, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -562,7 +562,7 @@ public abstract class EJSHome implements PoolDiscardStrategy, HomeInternal, Sess
     private void preLoadBeanPool() {
         final boolean isTraceOn = TraceComponent.isAnyTracingEnabled();
         if (isTraceOn && tc.isEntryEnabled())
-            Tr.debug(tc, "preLoadBeanPool: " + j2eeName);
+            Tr.entry(tc, "preLoadBeanPool: " + j2eeName);
 
         synchronized (beanPool) {
             Object oldClassLoader = ThreadContextAccessor.UNCHANGED;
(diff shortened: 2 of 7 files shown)


tkburroughs:
#build

tkburroughs:
#libby #build

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
```diff
@@ -242,6 +244,7 @@ private static RuntimeException rethrow(Throwable t) {
 
     private WSEJBHandlerResolver webServicesHandlerResolver;
     private EJBPMICollaboratorFactory ejbPMICollaboratorFactory;
+    private CheckpointPhase checkpointPhase;
```
tjwatson: Make this final and then add this constructor:
```java
     private final CheckpointPhase checkpointPhase;

     @Activate
     public EJBRuntimeImpl(@Reference(cardinality = ReferenceCardinality.OPTIONAL) CheckpointPhase checkpointPhase) {
         this.checkpointPhase = checkpointPhase;
     }
```

Note that is is acceptable to have an `@Activate` constructor and an`@Activate` method.  Both will be called (obviously the constructor will be called first when activating the component).

There is no reason to make this reference dynamic.  The phase service will either be there right from the start, or never at all.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
```diff
@@ -1683,6 +1686,26 @@ protected void unsetLibertyFeature(ServiceReference<LibertyFeature> feature) {
...
+
+    /**
+     * Determines if application start should optimize for checkpoint after all configured applications
+     * have started.
+     *
+     * @return true if application start should optimize for checkpoint; false otherwise.
+     */
+    @Override
+    public boolean isCheckpointApplications() {
+        return CheckpointPhase.APPLICATIONS == checkpointPhase;
```
tjwatson, line 1714: I question why we would not want to do this for both `DEPLOYMENT` and `APPLICATIONS`.  When this returns true are we pretty much guaranteed to run user code if the application has EJBs?  If so then I understand why we would limit this to the `APPLICATIONS` phase.  But if not, then it seems we would want to do this extra work as long as we don't have to execute `<clinit>` on any application code.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
```diff
@@ -1683,6 +1686,26 @@ protected void unsetLibertyFeature(ServiceReference<LibertyFeature> feature) {
...
     }
 
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
+    protected void setCheckpointPhase(CheckpointPhase phase) {
+        checkpointPhase = phase;
+    }
+
+    protected void unsetCheckpointPhase(CheckpointPhase phase) {
+        checkpointPhase = null;
+    }
```
tjwatson: Remove these methods in favor of a static optional requirement on `CheckpointPhase` with constructor injection.

olendvcook (approved):

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tkburroughs: @tjwatson  I made the reference dynamic in case Checkpoint/Restore ever wanted to support application restart, where the app restart would behave normally as without checkpoint.  That is... if the CheckpointPhase could be "unset" after restore, then any subsequent application restarts would restart faster.

This could also be done with a restore hook; but that would require every component that does something similar to have a restore hook.

The UFO doesn't seem to address application restart; while not an intended scenario, would be nice if it clarified the expected behavior.  Is the intended behavior that any new or restarted application after a restore should behave exactly as if a checkpoint is being taken?

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tkburroughs: @tjwatson I'm not aware of any potential EJB optimizations for CRIU that could occur prior to application start.

For some bean types, EJB Container attempts to defer some processing, such as loading the bean classes until first use; the property optimized here, EJBInitAtStartup, specifically moves loading of customer classes to application start.

With or without this change, if any of this EJB runtime code is running during DEPLOYMENT, then there is a significant problem for EJB applications.  EJBInitAtStartup only a hint to defer; it is not applicable to all bean types and configurations, so there will definitely be application code run in this code path.

Exactly how can I tell which EJB runtime code might be executed during DEPLOYMENT?

tjwatson: > Exactly how can I tell which EJB runtime code might be executed during DEPLOYMENT?

I'm not sure I understand this question or at least how to answer it because I do not know the EJB runtime code.  For DEPLOYMENT the startup path is going to be the same as APPLICATIONS except as soon as the runtime does any class static initialization of application code (run the app class `<clinit>`) a checkpoint will be done before proceeding to run the code in the application.  This ensures that if the DEPLOYMENT phase is used that we will not execute user application code prior to the checkpoint being taken.  Does that help?

If I understand your comment above then I think it is appropriate to only have this change for the APPLICATIONS phase regardless.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tjwatson: I don't think I'll want to have the CheckpointPhase service unregistered on the restore side to be able to detect if we have restored.  I think I would prefer a method on the CheckpointPhase (e.g. `boolean restored()`) that returns true once the process has restored.  Two reasons for this.

1) I want the CheckpointPhase to either always there or never there to simplify injection into components (can use static optoinal constructor injection).
2) Allows easy introspection on a dump to see if the checkpoint service is there.

Additionally we can have the CheckpointPhase service have a service property that indicates if it is the restore side.  Then if components would rather do what you are with dynamic optional references then they can use a target filter like:

```java
    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC, target="(io.openliberty.checkpoint.restored=false)")
```
This would then cause the reference to be unset on the restore side when we set the service property to true for restored.

tkburroughs: @tjwatson Thanks for that explanation. Agree there is no reason to make the reference dynamic now; so have changed to use the `@Activate` constructor; final variable.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tkburroughs: Removed.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tkburroughs: Ah, I was thinking it was based on appmanager or runtime framework lifecycle; but since it will be based on a class initializer, I did some further testing of how this configuration option works.  At this point in time, enabling this option for CRIU will cause application code to run (a `<clinit>`) much earlier than would otherwise occur, thus reducing the benefit one would see with DEPLOYMENT.  To avoid that, at least for this first change, I'd prefer to stay with APPLICATIONS for now.

However, as it turns out, at least the main reason classes would be loaded earlier is a bug; which was identified by testing from @dettmoney, related to the second config option in issue #20368.  Once I deliver the fix to that bug, and then can enable pool pre-loading for CRIU, I should also be able to switch this change for `EJBInitAtStartup` to also apply to DEPLOYMENT in addition to APPLICATIONS.

Net is; this is a good catch, but enabling for DEPLOYMENT will need to be done later.  To that end, I've added isCheckpointDeployment to the EJBRuntime interface for use later.

tkburroughs:
#libby #build

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tjwatson: I'm fine with only doing this for APPLICATIONS for now.

tjwatson (approved):

tkburroughs:
#build

PR #20614, opened 2022-03-28
Update FAT bnd.bnd files to reference correct fat.minimum.java.level property

ayoho:
Many FAT projects reference an outdated `runtime.minimum.java.level` property in their bnd.bnd files. That property should be `fat.minimum.java.level`. FATs also run, by default, on a minimum of Java 8, so any uses of that property to specify a minimum of Java 8 have either been removed (if the property wasn't commented out) or updated to reference Java 11 (if the property was commented out). Now if people need a reference to see how to require a minimum Java level they'll see a valid sample using Java 11 as the example.


ayoho:
#build

tevans78 (approved):

ayoho:
#build

PR #20632, opened 2022-03-29
Concurrency 3.0 resources defined in web fragment

njr-11:
Make it possible to define in a web fragment the four new resource types from Jakarta Concurrency 3.0 (Jakarta EE 10).  Add tests for each type, partly by enhancing some existing tests for JSP that were using an empty servlet to define resources.  We can get rid of that servlet entirely and use a web fragment instead, which will conveniently test this new spec function.

The test cases uncovered a bug where an empty list of context is causing a NullPointerException, so I'm fixing that as well.


njr-11:
#build

nmittles (approved):

PR #20657, opened 2022-03-31
Publish new schemaGen script

jimblye:
The last step for the wrapper script is to publish it.   #18191


cbridgha (approved):
LGTM

jimblye:
#build

hlhoots:
#20697 should push back in the reverted code for publishing via .mf file as well as the rc=0 scenario.

PR #20674, opened 2022-04-01
EJB TimedBean Test Fix

olendvcook:
Took out one of the CountDownLatches completely, it was used to have interval timers wait in their timeouts before the tests told them to continue, but this was only really because the tests checked they timed out a very specific number of times, and that after a certain number of times they could cancel themselves

Instead, made the test not care how many times they timed out, just that it happened multiple times (because it's testing single action and interval timers), and have a flag to let them auto cancel. This causes the beans to not sit there in the timeouts and get behind which was messing up getNextTimeout(), and un-complicates the tests a little bit.

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer.mdb.ra_fat/test-applications/MsgEndpointEJB.jar/src/com/ibm/ws/ejbcontainer/fat/msgendpoint/ejb/MDBTimedCMTBean.java
+++ b/dev/com.ibm.ws.ejbcontainer.mdb.ra_fat/test-applications/MsgEndpointEJB.jar/src/com/ibm/ws/ejbcontainer/fat/msgendpoint/ejb/MDBTimedCMTBean.java
@@ -60,7 +60,6 @@ public class MDBTimedCMTBean implements MessageListener {
     private static CountDownLatch svTimerResultsLatch = new CountDownLatch(1);
     private static CountDownLatch svTimerLatch = new CountDownLatch(1);
     private static CountDownLatch svTimerIntervalLatch = new CountDownLatch(1);
-    private static CountDownLatch svTimerIntervalWaitLatch = new CountDownLatch(1);
 
     // The time for the container to run post invoke processing for @Timeout.
     // Should be used after a Timer has triggered a CountDownLatch to insure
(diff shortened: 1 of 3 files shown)


olendvcook:
#build

dev/com.ibm.ws.ejbcontainer.mdb.ra_fat/test-applications/MsgEndpointEJB.jar/src/com/ibm/ws/ejbcontainer/fat/msgendpoint/ejb/MDBTimedCMTBean.java
```diff
@@ -75,6 +74,8 @@
 
     public MDBTimedCMTBean() {}
 
+    private static boolean svCancelTimer = false;
```
tkburroughs, line 77: Perhaps could add volatile to this to ensure the timer thread sees the change in state immediately after being set to true

dev/com.ibm.ws.ejbcontainer.mdb.ra_fat/test-applications/MsgEndpointEJB.jar/src/com/ibm/ws/ejbcontainer/fat/msgendpoint/ejb/MDBTimedCMTBean.java
```diff
@@ -404,6 +396,11 @@ public void onMessage(Message msg) {
                         // 23 - Verify NoSuchObjectLocalException occurs accessing self cancelled
                         //      timer
                         // -----------------------------------------------------------------------
+                        svTimerIntervalLatch = new CountDownLatch(2);
```
tkburroughs, line 399: I assume the latch is set to 2 because there are 2 timers remaining.  However, there appears to be a very narrow window where a timer thread could be performing a countdown at exactly the same time the new latch is created, but before the next line where cancel is true... This could lead to the countdown latch tripping, and yet a timer still exists.  Making both svTimerIntervalLatch and svCancelTimer volatile might help reduce the window further, but won't completely eliminate that.  Reversing the order doesn't really help either.

Adding a sync block around these 2 lines (i.e. creating the latch and setting cancel to true), and then sync'ing on the same object when performing the countdown in the timeout method could eliminte the window; but it seems pretty unlikely it would ever be hit.

tkburroughs (approved):
I think there is an extremely small timing window around when cancel is set to true; but I'm approving anyway, as I think the window is unlikely and this is better than what was there before. You can try and fix that if you want; or I'm fine delivering as-is.

PR #20683, opened 2022-04-01
Add support for CRIU unprivileged mode

ymanton:
This PR adds support for CRIU unprivileged mode to the checkpoint feature and the server script.

Regarding the checkpoint feature implementation:

> If the io.openliberty.checkpoint.criu.unprivileged property is set to
> to true, set unprivileged mode to true before invoking CRIU dump.
> 
> If NoSuchMethodError is thrown it means that the JVM does not support
> unprivileged mode, in which case throw an exception to that effect.

Regarding the server script implementation:

> Introduce the CRIU_EXTRA_ARGS env var to allow the user to
> pass extra arguments to `criu restore`.
> 
> Introduce the CRIU_UNPRIVILEGED env var to allow the user to
> control/override whether CRIU is invoked in privileged or unprivileged
> mode.
> 
> Check whether CRIU supports unprivileged mode by looking for
> "--unprivileged" in the output of `criu --help`.
> 
> If the script is invoked as root or CRIU doesn't support unprivileged
> mode or CRIU_UNPRIVILEGED is false, invoke CRIU in privileged mode.
> 
> If the script is invoked as non-root and CRIU supports unprivileged
> mode, or CRIU_UNPRIVILEGED is true, invoke CRIU in unprivileged mode.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -325,7 +325,7 @@ checkCriuUnprivileged()
   # Determine if CRIU supports unprivileged mode
   criu --help | grep -q -e '--unprivileged' && CRIU_SUPPORTS_UNPRIVILEGED=true || CRIU_SUPPORTS_UNPRIVILEGED=false
 
-  # Determine which CRIU mode to use, based on current euid and values of CRIU_SUPPORTS_UNPRIVILEGED and CRIU_UNPRIVILEGED
+  # Determine which CRIU mode to use, based on current euid and value of CRIU_SUPPORTS_UNPRIVILEGED
   DO_CRIU_UNPRIVILEGED=false
   if [ "$(id -u)" != 0 ]; then
     # Not root, unprivileged by default if CRIU supports it
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -30,6 +30,15 @@
 #TODO need to reflect this to criu checkpoint
 #                  Set to 1-4 with 4 being max verbosity. Two "2" is the default.
 #
+# CRIU_UNPRIVILEGED - Use criu in unprivileged mode.
+#                     Set to one of true, True, TRUE, 1
+#                     or false, False, FALSE, 0 to continue
+#                     to use criu in privileged mode.
+#
+# CRIU_EXTRA_ARGS   - Pass extra arguments to `criu restore`. Extra arguments are
+#                     appended to the end of the list of arguments and can therefore
+#                     override existing arguments, if desired.
+#
 # PID_DIR    - The directory that should be used for server pid file(s).
 #              The default value is ${WLP_OUTPUT_DIR}/.pid
 #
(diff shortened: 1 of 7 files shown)


ymanton:
FYI @tjwatson @tajila

The corresponding OpenJ9 is https://github.com/eclipse-openj9/openj9/pull/14836

tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -296,6 +305,31 @@ if [ -n "${WLP_SKIP_BOOTSTRAP_AGENT}" ]; then
...
+checkCriuUnprivileged()
+{
+  # Determine if CRIU supports unprivileged mode
+  criu --help | grep -q -e '--unprivileged' && CRIU_SUPPORTS_UNPRIVILEGED=true || CRIU_SUPPORTS_UNPRIVILEGED=false
+
+  # Determine which CRIU mode to use, based on current euid and values of CRIU_SUPPORTS_UNPRIVILEGED and CRIU_UNPRIVILEGED
+  DO_CRIU_UNPRIVILEGED=false
+  if [ "$(id -u)" != 0 ]; then
+    # Not root, unprivileged by default if CRIU supports it
+    if [ $CRIU_SUPPORTS_UNPRIVILEGED = true -a "${CRIU_UNPRIVILEGED}" != "false" -a "${CRIU_UNPRIVILEGED}" != "1" -a "${CRIU_UNPRIVILEGED}" != "FALSE" -a "${CRIU_UNPRIVILEGED}" != "False" ]; then
```
ymanton: Just noticed the `"${CRIU_UNPRIVILEGED}" != "1"` part should be `"${CRIU_UNPRIVILEGED}" != "0"`.

ymanton: Fixed.

tjwatson: My latest commit removes the false check altogether.  If not set to one of the "true" values then it just sets `DO_CRIU_UNPRIVILEGED` to false.

I also moved the check for the `CRIU_UNPRIVILEGED` variable to be first and if set then it never does a check for support in a `criu` help check.  @ymanton let me know if you see an issue with that.  I figured if someone sets the variable then just take what they specified and there is no need to run a `criu` to see if it is supported.

tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -296,6 +305,36 @@ if [ -n "${WLP_SKIP_BOOTSTRAP_AGENT}" ]; then
...
+    else
+      DO_CRIU_UNPRIVILEGED=false
+    fi
+    return
+  fi
+
+  # Determine if CRIU supports unprivileged mode
+  criu --help | grep -q -e '--unprivileged' && CRIU_SUPPORTS_UNPRIVILEGED=true || CRIU_SUPPORTS_UNPRIVILEGED=false
+
+  # Determine which CRIU mode to use, based on current euid and values of CRIU_SUPPORTS_UNPRIVILEGED and CRIU_UNPRIVILEGED
```
ymanton:
```suggestion
  # Determine which CRIU mode to use, based on current euid and value of CRIU_SUPPORTS_UNPRIVILEGED
```

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
ymanton: Looks good, it reads clearly to me. I updated the comment as a suggestion.

ymanton:
For completeness, here are the results of various combinations of support amongst the relevant components:

| uid | open-liberty | openj9 | criu | resulting criu mode | result |
| --- | --- | --- | --- | --- | ---  |
| root | - | - | - | privileged | works, baseline case |
| non-root | - | - | - | privileged | fails (expected, no privileges), baseline case |
| non-root + caps | - | - | - | privileged | works, baseline case |
| root | supports unpriv | supports unpriv | supports unpriv | privileged | works |
| non-root | supports unpriv | supports unpriv | supports unpriv | unprivileged | works |
| non-root + caps | any | any | supports unpriv | undefined | fails, unprivileged mode PR breaks this case |
| non-root | supports unpriv | supports unpriv | - | privileged | fails (expected, no privileges) |
| non-root | supports unpriv | - | supports unpriv | - | fails, liberty catches `NoSuchMethodError` from openj9, throws `CheckpointFailedException` |
| non-root | - | supports unpriv | supports unpriv | privileged | fails (no privileges, unprivileged mode needs to be explicitly requested by liberty) |

tjwatson (approved):

tjwatson:
#libby

PR #20694, opened 2022-04-04
Revert "Publish new schemaGen script"

tevans78:
Reverts OpenLiberty/open-liberty#20657

Tests fail due to a no-args call to the schemaGen script returning non-zero return code


cbridgha (approved):
Yep needs reverting

PR #20712, opened 2022-04-05
Create Faces 4.0 Placeholder Bundles for EL 5.0

### LINKED ISSUE #20774: Create Faces 4.0 Placeholder Bundles for EL 5.0
EL 5.0 created feature resolution problems for faces-4.0 since EL 5.0 provided the JavaEL version=5.0.0 Capability, but faces-4.0 is just the faces-3.0 underneath (which provided the JavaEL osgi contract version 4.0.0). I'm creating these 3 placeholder bundles, so that they can be updated to use expression language 5.0 on the build path in order to require JavaEL version=5.0.0 instead.

Both the API and implementation use the expression language API.

io.openliberty.faces.4.0.thirdparty uses the Fragment-Host: attribute, so it also needs to be updated.

In the EL 5.0 PR, these placeholder bundles will be updated to use the EL 5.0 api in the build path.

volosied:
fixes #20774

I don't think any tests should be enabled here as no development has really started here.  Faces-4.0 is no-ship, as well. 

Using the 4.0 snapshot api and implementation would have is too much hassle as this point as we are focusing on EL 5.0 & Servlet 6.0.  I think keeping it as faces 3.0 is fine for now. 

Buildpaths will be updated to use EL 5.0 the EL 5.0 PR.

### DIFF
--- a/dev/io.openliberty.jakarta.faces.4.0/build.gradle
+++ b/dev/io.openliberty.jakarta.faces.4.0/build.gradle
@@ -1,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2020, 2021 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+// Global vars
+File downloadSrcDir = project.file('build/source')
+File downloadJavadocDir = project.file('build/javadoc')
+
+configurations {
+    sourceJar
+    javadocJar
+}
+
+dependencies {
+   sourceJar 'org.apache.myfaces.core:myfaces-api:3.0.1:sources'
+   javadocJar 'org.apache.myfaces.core:myfaces-api:3.0.1:javadoc'
+}
+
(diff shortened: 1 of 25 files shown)


volosied:
#build

pnicolucci:
@volosied  can we get this associated with an issue for the EL 5.0 epic?

pnicolucci (commented):
Vlad, can you call out specifically via a comment where these bundles differ currently from their Jakarta EE9 counterparts? Or is the intention here that they are identical and then you'll update them to point to Expression Language 5.0 in another PR where you're integrating Expression Language 5.0?

pnicolucci:
> Vlad, can you call out specifically via a comment where these bundles differ currently from their Jakarta EE9 counterparts? Or is the intention here that they are identical and then you'll update them to point to Expression Language 5.0 in another PR where you're integrating Expression Language 5.0?

I talked with Vlad and the issue this references now gives good details.

dev/io.openliberty.jakarta.faces.4.0/build.gradle
```diff
@@ -0,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2020, 2021 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
```
pnicolucci, line 3: New file shouldn't this just be 2022?

pnicolucci (approved):

volosied:
#libby

PR #20753, opened 2022-04-08
Avoid compiling regular expressions every request.

jdmcclur:
We spend a bit of time compiling regular expressions in the acmeAirMS benchmark. This is an attempt to only do the compile once.

```
  Parent      0   0.00   0.05           0           4    J:java/lang/String.split(Ljava/lang/String;I)[Ljava/lang/String;
  Parent      0   0.00   0.21           0          18    J:com/ibm/ws/security/AccessIdUtil.getUniqueId(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;
  Parent      0   0.00   0.33           0          29    J:java/util/regex/Pattern.matches(Ljava/lang/String;Ljava/lang/CharSequence;)Z
 
    Self      0   0.00   0.58           0          51    J:java/util/regex/Pattern.compile(Ljava/lang/String;)Ljava/util/regex/Pattern;
 
   Child      0   0.01   0.58           1          51    J:java/util/regex/Pattern.<init>(Ljava/lang/String;I)V
``` 

Not sure if the change in AccessIdUtil is the best idea. Will there be a lot of realms or just a few? If just a few, will probably be fine, but wondering if it needs to be a a ConcurrentHashMap?

### DIFF
--- a/dev/com.ibm.ws.security.common/src/com/ibm/ws/security/common/crypto/KeyAlgorithmChecker.java
+++ b/dev/com.ibm.ws.security.common/src/com/ibm/ws/security/common/crypto/KeyAlgorithmChecker.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020 IBM Corporation and others.
+ * Copyright (c) 2020, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


jdmcclur:
#build

ayoho (approved):

jdmcclur:
#libby

jhanders34 (approved):

PR #20755, opened 2022-04-08
Use service property to determine local network address, if available

### LINKED ISSUE #20754: Use the local network address for accessessing external test services where available
To reduce network usage, if the test machine and external test service are on the same network, and there is a network local IP address defined on the service, then use this in preference to the external IP address

rickettmwork:
fixes #20754


websterc87 (approved):
Looks good to me pending a greenish build

PR #20765, opened 2022-04-11
PR for issue 20714. Catch SIConnectionDroppedException in SibRaListener.close()

### LINKED ISSUE #20714: Exceptions thrown from MDBs when stopping a server that is running a JMS messaging engine
When stopping a Liberty server that runs a JMS server, MDBs can generate FFDC incidents involving the JMS server not being accessible.
This appears to occur because, when the MDBs are begin closed, they internally attempt to close their JMS Connections, which are no longer value because the Messaging Engine has already quiesced.

sakerl:


joe-chacko (approved):
lotm

sakerl:
#build

PR #20770, opened 2022-04-11
Add AllowedFFDC for CommExceptions when we failover remote LDAPs

kristip17:
If we failover LDAPs at the WIM level due to a `javax.naming.CommunicationException`, we will log an FFDC. When the failover is successful, the test is fine, but then fails with an unexpected FFDC. Adding `AllowedFFDC` on a case by case basis to reduce false BBs. If we can't failover successfully, the test will likely fail due to an Assert or other test check.

RTC 290230


kristip17:
#build

jvanhill (approved):

kristip17:
#libby

PR #20803, opened 2022-04-13
Reject Concurrent Asynch on EJBs

KyleAure:
Ensure Jakarta EE 10 Concurrency compliance by rejecting the use of `@Asynchronous` on EJBs. 

Spec Documentation: https://github.com/jakartaee/concurrency/blob/master/specification/src/main/asciidoc/jakarta-concurrency.adoc#5-asynchronous-methods

> The jakarta.enterprise.concurrent.Asynchronous annotation annotates a CDI managed bean method to run asynchronously. The CDI managed bean must not be a Jakarta Enterprise Bean, and neither the method nor its class can be annotated with the MicroProfile Asynchronous annotation.

Behavior:
If the `jakarta.enterprise.concurrent.Asynchronous` annotation is used on a Jakarta Enterprise Bean method then we will both log and throw an error to notify the user that the annotation they used is not supported within the EJB container.

If the `jakarta.enterprise.concurrent.Asynchronous` annotation is used on a Jakarta Enterprise Interface method then we will log a warning to notify the user that the annotation they used is not supported within the EJB container.  However, since users should not expect interface method annotations to be inherited and the specification offers no guidance on how to handle this situation we will allow the application to be installed and used.

### DIFF
--- a/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceBean.java
+++ b/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceBean.java
@@ -1,0 +1,51 @@
+/**
+ *
+ */
+package test.jakarta.concurrency.ejb.error;
+
+import static org.junit.Assert.assertNotNull;
+
+import java.util.concurrent.CompletableFuture;
+
+import jakarta.ejb.Local;
+import jakarta.ejb.Stateless;
+import jakarta.enterprise.concurrent.Asynchronous;
+
+@Stateless
+@Local(AsynchInterfaceLocal.class)
+public class AsynchInterfaceBean {
+
+    public void getThreadName() {
+        System.out.println("Thread name async: " + Thread.currentThread().getName());
+    }
+
+    public void getThreadNameNonAsyc() {
+        System.out.println("Thread name non-async:" + Thread.currentThread().getName());
+    }
+
+    public CompletableFuture<String> getState(String city) {
+        if (city == "Rochester")
+            return Asynchronous.Result.complete("Minnesota");
+        else
+            return Asynchronous.Result.complete(null);
+    }
+
+    public CompletableFuture<String> getStateFromService(String city) {
+
+        CompletableFuture<String> future = Asynchronous.Result.getFuture();
+
+        assertNotNull(future);
+
+        try {
+            if (city == "Rochester")
+                future.complete("Minnesota");
+            else
+                future.complete(null);
+        } catch (Exception x) {
+            future.completeExceptionally(x);
+        }
+
+        return future;
+    }
+
+}
--- a/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceBean.java
+++ b/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceBean.java
@@ -1,6 +1,13 @@
-/**
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
  *
- */
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
 package test.jakarta.concurrency.ejb.error;
 
 import static org.junit.Assert.assertNotNull;
--- a/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceLocal.java
+++ b/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceLocal.java
@@ -1,0 +1,29 @@
+/**
+ *
+ */
+package test.jakarta.concurrency.ejb.error;
+
+import java.util.concurrent.CompletableFuture;
+
+import jakarta.enterprise.concurrent.Asynchronous;
+
+/**
+ * This generic interface is using the Jakarta Concurrency @Asynchronous annotations.
+ * Since users should not expect annotations on methods to be inherited from interfaces
+ * (only super-classes) these annotations should be ignored.
+ * The application should be installed, and a warning given to the user.
+ */
+public interface AsynchInterfaceLocal {
+
+    @Asynchronous
+    public void getThreadName();
+
+    public void getThreadNameNonAsyc();
+
+    @Asynchronous
+    public CompletableFuture<String> getState(String city);
+
+    @Asynchronous(executor = "java:comp/DefaultManagedExecutorService")
+    public CompletableFuture<String> getStateFromService(String city);
+
+}
--- a/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceLocal.java
+++ b/dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceLocal.java
@@ -1,6 +1,13 @@
-/**
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
  *
- */
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
 package test.jakarta.concurrency.ejb.error;
 
 import java.util.concurrent.CompletableFuture;
--- a/dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
+++ b/dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
@@ -2223,10 +2223,10 @@ CNTR9425E_NO_URL_SPEC_SPECIFIED=CNTR9425E: The embeddable Enterprise JavaBeans (
 CNTR9425E_NO_URL_SPEC_SPECIFIED.explanation=The properties that the embeddable Enterprise JavaBeans (EJB) container uses contain a URL resource with no specification specified.
 CNTR9425E_NO_URL_SPEC_SPECIFIED.useraction=Add a specification property for this URL resource to the properties file.
 
-CNTR9426E_INCORRECT_ASYNC_ANNO=CNTR9426E: The {0} bean in the {1} module of the {2} application has one or more methods with the Jakarta Concurrent Asynchronous annotation.
-CNTR9426E_INCORRECT_ASYNC_ANNO.explanation=Jakarta Concurrent Asynchronous annotations are only valid for CDI beans, not EJBs.
-CNTR9426E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrent Asynchronous annotation from method.
+CNTR9426E_INCORRECT_ASYNC_ANNO=CNTR9426E: The {0} bean in the {1} module of the {2} application has one or more methods with the Jakarta Concurrency Asynchronous annotation, which must only be used on CDI beans and is not valid on Jakarta Enterprise Beans.
+CNTR9426E_INCORRECT_ASYNC_ANNO.explanation=Jakarta Concurrency Asynchronous annotations are only valid for CDI beans, not Jakarta Enterprise Beans.
+CNTR9426E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrency Asynchronous annotation from the method.
 
-CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR9427W: The {0} interface has one or more methods with the Jakarta Concurrent Asynchronous annotation.
-CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=Jakarta Concurrent Asynchronous annotations are only valid for CDI beans, not EJBs or Interfaces.
-CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction=Remove the Jakarta Concurrent Asynchronous annotation from interface method.
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR9427W: The Jakarta Concurrency Asynchronous annotation must only be used on CDI beans. It is not valid on the {0} interface.
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=Jakarta Concurrency Asynchronous annotations are only valid for CDI beans, not Jakarta Enterprise Beans or Interfaces.
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction=Remove the Jakarta Concurrency Asynchronous annotation from the interface method.
--- a/dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
+++ b/dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
@@ -2223,3 +2223,10 @@ CNTR9425E_NO_URL_SPEC_SPECIFIED=CNTR9425E: The embeddable Enterprise JavaBeans (
 CNTR9425E_NO_URL_SPEC_SPECIFIED.explanation=The properties that the embeddable Enterprise JavaBeans (EJB) container uses contain a URL resource with no specification specified.
 CNTR9425E_NO_URL_SPEC_SPECIFIED.useraction=Add a specification property for this URL resource to the properties file.
 
+CNTR9426E_INCORRECT_ASYNC_ANNO=CNTR9426E: The {0} bean in the {1} module of the {2} application has one or more methods with the Jakarta Concurrent Asynchronous annotation.
+CNTR9426E_INCORRECT_ASYNC_ANNO.explanation=Jakarta Concurrent Asynchronous annotations are only valid for CDI beans, not EJBs.
+CNTR9426E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrent Asynchronous annotation from method.
+
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR9427W: The {0} interface has one or more methods with the Jakarta Concurrent Asynchronous annotation.
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=Jakarta Concurrent Asynchronous annotations are only valid for CDI beans, not EJBs or Interfaces.
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction=Remove the Jakarta Concurrent Asynchronous annotation from interface method.
--- a/dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
+++ b/dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
 UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W.explanation=The remote enterprise bean binding cannot be created because the CORBA name server is not available.
 UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W.useraction=Check the log file for exceptions that are related to the ORB or CORBA name server initialization.
 
+CNTR0342E_INCORRECT_ASYNC_ANNO=CNTR0342E: The {0} bean in the {1} module of the {2} application has one or more methods with the jakarta.enterprise.concurrent.Asynchronous annotation, which must not be used on enterprise beans. The jakarta.ejb.Asynchronous annotation provides asynchronous methods for enterprise beans.
+CNTR0342E_INCORRECT_ASYNC_ANNO.explanation=The Jakarta Concurrency specification does not allow the jakarta.enterprise.concurrent.Asynchronous annotation to be used on Jakarta Enterprise Beans.
+CNTR0342E_INCORRECT_ASYNC_ANNO.useraction=Remove the jakarta.enterprise.concurrent.Asynchronous annotation from the method or class. Use the jakarta.ejb.Asynchronous annotation to identify asynchronous enterprise bean methods.
+
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR0343W: The {0} business interface contains a jakarta.enterprise.concurrent.Asynchronous annotation. Asynchronous annotations are not valid on any interface, and the Jakarta Enterprise Beans container ignores them.
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=The jakarta.enterprise.concurrent.Asynchronous annotation must not be used with Jakarta Enterprise Beans. The Jakarta Enterprise Beans container ignores these annotations on interfaces. Unless the methods are declared asynchronous in the ejb-jar.xml deployment descriptor or the jakarta.ejb.Asynchronous annotation is specified on the bean class or a bean class method, the bean methods run synchronously.
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction=Remove any jakarta.enterprise.concurrent.Asynchronous annotation from the specified business interface and verify that the jakarta.ejb.Asynchronous annotation is specified correctly in the bean class.
+
 #-------------------------------------------------------------------------------
 #
 # CNTR4000 to CNTR4999 reserved for Liberty.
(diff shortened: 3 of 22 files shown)


KyleAure:
#build

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -2223,3 +2223,10 @@ CNTR9425E_NO_URL_SPEC_SPECIFIED=CNTR9425E: The embeddable Enterprise JavaBeans (
 CNTR9425E_NO_URL_SPEC_SPECIFIED.explanation=The properties that the embeddable Enterprise JavaBeans (EJB) container uses contain a URL resource with no specification specified.
 CNTR9425E_NO_URL_SPEC_SPECIFIED.useraction=Add a specification property for this URL resource to the properties file.
 
+CNTR9426E_INCORRECT_ASYNC_ANNO=CNTR9426E: The {0} bean in the {1} module of the {2} application has one or more methods with the Jakarta Concurrent Asynchronous annotation.
```
njr-11: We should change "Jakarta Concurrent Asynchronous annotation" to "Jakarta Concurrency Asynchronous annotation" here and elsewhere within the 2 new messages.

It would also be nice to add the reason to the end of the message because users need to go out of the way to ever find the explanation/useraction.
"... with the Jakarta Concurrency Asynchronous annotation, which must only be used on CDI beans and is not valid on Jakarta Enterprise Beans."

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -2223,3 +2223,10 @@ CNTR9425E_NO_URL_SPEC_SPECIFIED=CNTR9425E: The embeddable Enterprise JavaBeans (
...
+CNTR9426E_INCORRECT_ASYNC_ANNO.explanation=Jakarta Concurrent Asynchronous annotations are only valid for CDI beans, not EJBs.
+CNTR9426E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrent Asynchronous annotation from method.
+
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR9427W: The {0} interface has one or more methods with the Jakarta Concurrent Asynchronous annotation.
```
njr-11: Can we say, "The Jakarta Concurrent Asynchronous annotation must only be used on CDI beans. It is not valid on the {0} interface." ?

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
```diff
@@ -340,6 +341,53 @@ public static boolean getAsynchronousMethods(Method[] beanMethods,
...
+            if (beanMethod == null)
+                continue; //onto next method
+
+            asynchMethodAnnotations = beanMethod.getAnnotations();
+
+            if (asynchMethodAnnotations == null)
+                continue; //onto next method
+
+            for (Annotation beanAnno : asynchMethodAnnotations) {
+                if (beanAnno.annotationType().getName().contains("jakarta.enterprise.concurrent.Asynchronous")) {
```
njr-11: Should use `startsWith` instead of `contains` because it's more efficient

dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceBean.java
```diff
@@ -0,0 +1,51 @@
+/**
```
njr-11: Copyright header is missing from this class.

dev/com.ibm.ws.concurrent_fat_jakarta/test-applications/ConcurrencyTestError/src/test/jakarta/concurrency/ejb/error/AsynchInterfaceLocal.java
```diff
@@ -0,0 +1,29 @@
+/**
```
njr-11: Copyright header is missing from this class.

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -2223,10 +2223,10 @@ CNTR9425E_NO_URL_SPEC_SPECIFIED=CNTR9425E: The embeddable Enterprise JavaBeans (
...
 CNTR9425E_NO_URL_SPEC_SPECIFIED.useraction=Add a specification property for this URL resource to the properties file.
 
-CNTR9426E_INCORRECT_ASYNC_ANNO=CNTR9426E: The {0} bean in the {1} module of the {2} application has one or more methods with the Jakarta Concurrent Asynchronous annotation.
-CNTR9426E_INCORRECT_ASYNC_ANNO.explanation=Jakarta Concurrent Asynchronous annotations are only valid for CDI beans, not EJBs.
-CNTR9426E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrent Asynchronous annotation from method.
+CNTR9426E_INCORRECT_ASYNC_ANNO=CNTR9426E: The {0} bean in the {1} module of the {2} application has one or more methods with the Jakarta Concurrency Asynchronous annotation, which must only be used on CDI beans and is not valid on Jakarta Enterprise Beans.
+CNTR9426E_INCORRECT_ASYNC_ANNO.explanation=Jakarta Concurrency Asynchronous annotations are only valid for CDI beans, not Jakarta Enterprise Beans.
+CNTR9426E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrency Asynchronous annotation from method.
```
njr-11: Sorry I didn't catch this trivial grammar improvement before - "from method" ought to be "from the method".  This also applies to CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
```diff
@@ -369,7 +369,7 @@ public static boolean getConcurrentAsynchronousMethods(Method[] beanMethods) {
                 continue; //onto next method
 
             for (Annotation beanAnno : asynchMethodAnnotations) {
-                if (beanAnno.annotationType().getName().contains("jakarta.enterprise.concurrent.Asynchronous")) {
+                if (beanAnno.annotationType().getName() == "jakarta.enterprise.concurrent.Asynchronous") {
```
njr-11: Good idea using equals to be even more efficient, although `==` will do an instance comparison which will usually work, but there are cases where it might not because there are different instances of the string value, possibly due to different class loaders.  The safer way to do this is:
```
"jakarta.enterprise.concurrent.Asynchronous".equals(beanAnno.annotationType().getName())
```

njr-11 (approved):
Only very minor updates needed. Sorry I didn't catch the grammar update to make during the previous review.

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -2223,3 +2223,10 @@ CNTR9425E_NO_URL_SPEC_SPECIFIED=CNTR9425E: The embeddable Enterprise JavaBeans (
 CNTR9425E_NO_URL_SPEC_SPECIFIED.explanation=The properties that the embeddable Enterprise JavaBeans (EJB) container uses contain a URL resource with no specification specified.
 CNTR9425E_NO_URL_SPEC_SPECIFIED.useraction=Add a specification property for this URL resource to the properties file.
 
+CNTR9426E_INCORRECT_ASYNC_ANNO=CNTR9426E: The {0} bean in the {1} module of the {2} application has one or more methods with the Jakarta Concurrency Asynchronous annotation, which must only be used on CDI beans and is not valid on Jakarta Enterprise Beans.
```
tkburroughs: CNTR94xx is a message ranged reserved for use with the Embeddable EJB Container, which is not this scenario.  This code is part of the "core" container.

Both messages should be moved earlier in the file; since this is the "core", this would be in the normal range, and the next would be CNTR0342E and CNTR0343W.  Please move up earlier in the file and change the number.

tkburroughs: While the message is worded nicely, it still may not be obvious to the app developer what the problem is.  I'd prefer the message be very explicit about the problem and tell them what they really need to know, which is how to fix it. Explaining which specification provides specific annotations and restrictions should be secondary to resolving the issues; something for the explanation/user action.

I'd recommend something like this:

CNTR0342E_INCORRECT_ASYNC_ANNO=CNTR0342E: The {0} bean in the {1} module of the {2} application has one or more methods with the jakarta.enterprise.concurrent.Asynchronous annotation, which must not be used on enterprise beans. The jakarta.ejb.Asynchronous annotation provides asynchronous methods for enterprise beans.
CNTR0342E_INCORRECT_ASYNC_ANNO.explanation=The Jakarta Concurrency specification requires that the jakarta.enterprise.concurrent.Asynchronous annotation not be used on Jakarta Enterprise Beans.
CNTR0342E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrency Asynchronous annotation, jakarta.enterprise.concurrent.Asynchronous, from the method or Class. Use the Jakarta Enterprise Beans Asynchronous annotation, jakarta.ejb.Asynchronous, to identify asynchronous enterprise bean methods.

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -2223,3 +2223,10 @@ CNTR9425E_NO_URL_SPEC_SPECIFIED=CNTR9425E: The embeddable Enterprise JavaBeans (
...
+CNTR9426E_INCORRECT_ASYNC_ANNO.explanation=Jakarta Concurrency Asynchronous annotations are only valid for CDI beans, not Jakarta Enterprise Beans.
+CNTR9426E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrency Asynchronous annotation from the method.
+
+CNTR9427W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR9427W: The Jakarta Concurrency Asynchronous annotation must only be used on CDI beans. It is not valid on the {0} interface.
```
tkburroughs: As mentioned above, the message is in the wrong range; should be CNTR0343W.  This would tend to be an unlikely scenario, so I'd be fine if you just used the same message as is currently used for javax.ejb.Asynchronous when found on the interface... CNTR0305W.  However, a new message could be more specific, but as above, I think it should be more specific, including the full packages of the annotations to make it very clear what the problem is.... and then also direct them to the solution; mention the spec in the explanation/user action.  I'd prefer the message focus on EJB, and not bother mentioning CDI Beans, as I don't think that is necessarily the proper way to refer to them, and that is only a current limitation; @Asynchronous is the future for Jakarta EE and so may be valid with other types beyond CDI as well.  So, something like this:

CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR0343W: The {0} business interface contains a jakarta.enterprise.concurrent.Asynchronous annotation.  These annotations are not valid on interface types, and the Jakarta Enterprise Beans container ignores them.
CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=The jakarta.enterprise.concurrent.Asynchronous annotation must not be used with Jakarta Enterprise Beans.  The EJB container ignores these annotations on the interfaces.  Unless the jakarta.ejb.Asynchronous annotation is specified on the bean class or a bean class method, or if the methods are declared asynchronous in the ejb-jar.xml deployment descriptor, the bean methods will run synchronously.
CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction=Remove any jakarta.enterprise.concurrent.Asynchronous annotation from the specified business interface, and verify that the jakarta.ejb.Asynchronous annotation is specified correctly in the bean class.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/BeanMetaData.java
```diff
@@ -1129,6 +1129,12 @@ public int getActivationPolicy() {
      */
     public boolean ivHasAsynchMethod = false;
 
+    /**
```
tkburroughs: While this works, BeanMetaData is not really the proper place to keep track of error only data.  We work to not bloat BeanMetaData since it remains in memory for the life of the application. Since this is an error only scenario, we should just throw the error when it is found.

I see that this follows a similar pattern as javax.ejb.Asynchronous, but the field is used for working configurations, not just for detecting errors.

Please just remove all changes from BeanMetaData, and instead throw the error when the annotation is identified.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/BeanMetaData.java
```diff
@@ -2449,6 +2456,17 @@ public void validate() throws EJBConfigurationException {
 
         } // Validation of asynchronous methods
 
+        if (ivHasConcurrentAsynchMethod) {
```
tkburroughs: As above, remove this, and instead log the error where the annotation is found.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
```diff
@@ -340,6 +341,53 @@ public static boolean getAsynchronousMethods(Method[] beanMethods,
...
+    /**
+     * <code>getConcurrentAsynchronousMethods<code> checks all beanMethods to determine
+     * if any beans are annotated with the Jakarta Concurrent Asynchronous annotation.
+     * Returns true when found, or false otherwise.
+     *
+     * @param beanMethods Input array of Method objects representing the methods of this EJB.
+     *
+     * @return True if at least one method was found on this bean to have the Concurrent Asynchronous annotation.
+     */
+    public static boolean getConcurrentAsynchronousMethods(Method[] beanMethods) {
```
tkburroughs: Perhaps change the name of this method to `verifyNoConcurrentAsynchronousMethods` and have it log the error and throw the exception.  There is no need to save that to be done later.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ws/metadata/ejb/EJBMDOrchestrator.java
```diff
@@ -1296,6 +1296,9 @@ private MethodDataWrapper initializeBeanMethodMD(boolean isEntity,
                                                                               methodInterface); //d599046
         }
 
+        //Set a flag if any method is annotated with the concurrency Asychronous annotation.
```
tkburroughs: This call must be guarded by 

`if (!bmd.metadataComplete) {`

Specifically, the "metadataComplete" setting is a performance setting that an application developer may use to tell the container to NOT look at any annotations.  Therefore, we should not be scanning for this annotation if metadataComplete is set to true.  `metadata-complete` is a setting that my be specified in either ejb-jar.xml (for EJB JARs) or in web.xml (when the EJB is in a WAR module).

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ws/metadata/ejb/EJBMDOrchestrator.java
```diff
@@ -1296,6 +1296,9 @@ private MethodDataWrapper initializeBeanMethodMD(boolean isEntity,
...
+        bmd.ivHasConcurrentAsynchMethod |= MethodAttribUtils.getConcurrentAsynchronousMethods(ejbMethods);
```
tkburroughs: Per the above comment, there is no need to save this in bmd; just call the method (perhaps renamed) and have the method log the error and throw the exception.

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
tkburroughs: unlike getAsynchronousMethods, I notice this method does not check for the annotation on the Class.  If the annotation can be on the class, is it worth checking?  Could probably add the class as a parameter to this method, and then should be sure to look on super classes as well, up to java.lang.Object.

tkburroughs (changes requested):

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
```diff
@@ -340,6 +343,96 @@ public static boolean getAsynchronousMethods(Method[] beanMethods,
...
+            //Check method level annotations
+            Annotation[] asynchMethodAnnotations = null;
+            for (Method beanMethod : beanMethods) {
+
+                if (beanMethod == null)
+                    continue; //onto next method
+
+                //First compile a list of classes to check
+                Class<?> beanClass = beanMethod.getDeclaringClass();
+                if (beanClass != null && !classesToCheck.contains(beanClass)) {
```
tkburroughs: HashSet does not allow duplicates; the add() method already determines if the element is present, so it would be more efficient to just always call add(); no need for contains check.

tkburroughs (approved):
Looks good; just the one comment about a minor efficiency around use of HashSet; not worth holding up approval.

I see it has been flagged for message review; so just that remaining

tkburroughs:
#libby

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
 UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W.explanation=The remote enterprise bean binding cannot be created because the CORBA name server is not available.
 UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W.useraction=Check the log file for exceptions that are related to the ORB or CORBA name server initialization.
 
+CNTR0342E_INCORRECT_ASYNC_ANNO=CNTR0342E: The {0} bean in the {1} module of the {2} application has one or more methods with the jakarta.enterprise.concurrent.Asynchronous annotation, which must not be used on enterprise beans. The jakarta.ejb.Asynchronous annotation provides asynchronous methods for enterprise beans.
+CNTR0342E_INCORRECT_ASYNC_ANNO.explanation=The Jakarta Concurrency specification requires that the jakarta.enterprise.concurrent.Asynchronous annotation not be used on Jakarta Enterprise Beans.
```
dmuelle:
```suggestion
CNTR0342E_INCORRECT_ASYNC_ANNO.explanation=The Jakarta Concurrency specification does not allow the jakarta.enterprise.concurrent.Asynchronous annotation to be used on Jakarta Enterprise Beans.
```

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
...
+CNTR0342E_INCORRECT_ASYNC_ANNO.useraction=Remove the Jakarta Concurrency Asynchronous annotation, jakarta.enterprise.concurrent.Asynchronous, from the method or Class. Use the Jakarta Enterprise Beans Asynchronous annotation, jakarta.ejb.Asynchronous, to identify asynchronous enterprise bean methods.
```
dmuelle:
```suggestion
CNTR0342E_INCORRECT_ASYNC_ANNO.useraction=Remove the jakarta.enterprise.concurrent.Asynchronous annotation from the method or Class. Use the jakarta.ejb.Asynchronous annotation to identify asynchronous enterprise bean methods.
```

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
...
+
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR0343W: The {0} business interface contains a jakarta.enterprise.concurrent.Asynchronous annotation. These annotations are not valid on interface types, and the Jakarta Enterprise Beans container ignores them.
```
dmuelle: `These annotations are not valid on interface types`

This reads strangely- is the annotation not valid for _any_ interface type?

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
...
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=The jakarta.enterprise.concurrent.Asynchronous annotation must not be used with Jakarta Enterprise Beans. The EJB container ignores these annotations on the interfaces. Unless the jakarta.ejb.Asynchronous annotation is specified on the bean class or a bean class method, or if the methods are declared asynchronous in the ejb-jar.xml deployment descriptor, the bean methods will run synchronously.
```
dmuelle:
```suggestion
CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=The jakarta.enterprise.concurrent.Asynchronous annotation must not be used with Jakarta Enterprise Beans. The EJB container ignores these annotations on the interfaces. Unless the jakarta.ejb.Asynchronous annotation is specified on the bean class or a bean class method or the methods are declared asynchronous in the ejb-jar.xml deployment descriptor, the bean methods run synchronously.
```

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
...
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction=Remove any jakarta.enterprise.concurrent.Asynchronous annotation from the specified business interface, and verify that the jakarta.ejb.Asynchronous annotation is specified correctly in the bean class.
```
dmuelle:
```suggestion
CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.useraction=Remove any jakarta.enterprise.concurrent.Asynchronous annotation from the specified business interface and verify that the jakarta.ejb.Asynchronous annotation is specified correctly in the bean class.
```

dmuelle (changes requested):
ID review- made a few suggestions and asked a question. Let me know if you have any questions, thanks

dev/com.ibm.ws.ejbcontainer.core/src/com/ibm/ejs/container/util/MethodAttribUtils.java
KyleAure: Good catch Tracy!  I switched to HashSet for this reason and then forgot to remove the check!

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
KyleAure: @dmuelle I agree this statement was confusing.  I have changed it to: 
```
Asynchronous annotations are not valid on any interface, and the Jakarta Enterprise Beans container will ignore them.
```
To indicate that in addition to it being the wrong annotation, a user should know that even the EJB Asynchronous annotation is not valid on an interface. 
FYI - @tkburroughs

tkburroughs: 👍

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
 UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W.explanation=The remote enterprise bean binding cannot be created because the CORBA name server is not available.
 UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W.useraction=Check the log file for exceptions that are related to the ORB or CORBA name server initialization.
 
+CNTR0342E_INCORRECT_ASYNC_ANNO=CNTR0342E: The {0} bean in the {1} module of the {2} application has one or more methods with the jakarta.enterprise.concurrent.Asynchronous annotation, which must not be used on enterprise beans. The jakarta.ejb.Asynchronous annotation provides asynchronous methods for enterprise beans.
+CNTR0342E_INCORRECT_ASYNC_ANNO.explanation=The Jakarta Concurrency specification does not allow the jakarta.enterprise.concurrent.Asynchronous annotation to be used on Jakarta Enterprise Beans.
+CNTR0342E_INCORRECT_ASYNC_ANNO.useraction=Remove the jakarta.enterprise.concurrent.Asynchronous annotation from the method or Class. Use the jakarta.ejb.Asynchronous annotation to identify asynchronous enterprise bean methods.
```
dmuelle:
```suggestion
CNTR0342E_INCORRECT_ASYNC_ANNO.useraction=Remove the jakarta.enterprise.concurrent.Asynchronous annotation from the method or class. Use the jakarta.ejb.Asynchronous annotation to identify asynchronous enterprise bean methods.
```

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
...
+
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR0343W: The {0} business interface contains a jakarta.enterprise.concurrent.Asynchronous annotation. Asynchronous annotations are not valid on any interface, and the Jakarta Enterprise Beans container will ignore them.
```
dmuelle:
```suggestion
CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE=CNTR0343W: The {0} business interface contains a jakarta.enterprise.concurrent.Asynchronous annotation. Asynchronous annotations are not valid on any interface, and the Jakarta Enterprise Beans container ignores them.
```

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
```diff
@@ -1566,6 +1566,14 @@ UNABLE_TO_BIND_REMOTE_BINDING_CNTR0341W=CNTR0341W: Unable to bind the {0} interf
...
+CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=The jakarta.enterprise.concurrent.Asynchronous annotation must not be used with Jakarta Enterprise Beans. The EJB container ignores these annotations on interfaces. Unless the jakarta.ejb.Asynchronous annotation is specified on the bean class or a bean class method or the methods are declared asynchronous in the ejb-jar.xml deployment descriptor, the bean methods will run synchronously.
```
dmuelle:
```suggestion
CNTR0343W_INCORRECT_ASYNC_ANNO_INTERFACE.explanation=The jakarta.enterprise.concurrent.Asynchronous annotation must not be used with Jakarta Enterprise Beans. The EJB container ignores these annotations on interfaces. Unless the methods are declared asynchronous in the ejb-jar.xml deployment descriptor or the jakarta.ejb.Asynchronous annotation is specified on the bean class or a bean class method, the bean methods run synchronously.
```

dmuelle (changes requested):
thanks for clarifying- a few more suggestions

dev/com.ibm.ws.ejbcontainer.core/resources/com/ibm/ejs/container/container.nlsprops
tkburroughs: Please do not use "EJB" in any new messages.  "EJB" is a registered trademark and is no longer used to refer to the Jakarta Enterprise Beans technology.  Use either "Jakarta Enterprise Beans conainer" or "enterprise beans container".  It is of course acceptable to use "ejb" when it is part of the file name (ejb-jar.xml) or API package (jakarta.ejb.Asynchronous).

KyleAure:
@dmuelle and @tkburroughs I have updated the messages per both of your comments. 😄

tkburroughs (approved):

dmuelle (approved):
LGTM thanks

KyleAure:
#build

skasund:
L2 message review completed

KyleAure:
#libbybot

PR #20798, opened 2022-04-13
CDI JEE FAT Updates

tevans78:
The main aim of this PR is to repeat the `com.ibm.ws.cdi.jee_fat` bucket with EE10. However, we also take the opportunity to tidy up and update the bucket. Most notably we want to avoid use of the deprecated SharedServer code.

I have also included an update to FATRunner which warns if the `componenttest-x.x` feature is not enabled on the test server.

PR #20817, opened 2022-04-14
Check for componenttest feature

Azquelt:
This change was removed from #20798 because it caused failures in the build, so it needs some fixing up before it can be delivered.


Azquelt:
@tevans78 You can see the failures caused by this change in this build result: https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_njr3cLs4EeyhmZ5zGPuLaA

PR #20820, opened 2022-04-14
Reduce sampling from 3 to 1 for UserEnumTest

kristip17:
Reduce the sampling for `UserEnumerationTest`. It looks like now we're averaging out the invalid/valid tests so they are more matchy-matchy which is generally fine, but not consistent for the FAT test. Since we have a more consistent response with the CustomRepo, I think we can compare 1 invalid to 1 valid call instead of the average of 3 to 3. Just changing the loop count for now, if it doesn't work, we can change the loop count again.

RTC 290280


kristip17:
#build

jvanhill (approved):

PR #20824, opened 2022-04-14
Add a running condition for Liberty

### LINKED ISSUE #20660: Add mechanism to conditionally enable service components on restore or for normal launches
Some components may not want to be enabled when a checkpoint is going to be taken.  Instead these components would like to be enabled once the server process has been restored.  Additionally, these components would like to be enabled as soon as possible (depending on other required references availability) when not in a checkpoint run.

We can consider adding a [condition service](https://docs.osgi.org/specification/osgi.core/8.0.0/service.condition.html) that is registered by the kernel to indicate these two cases.  When the kernel is run without checkpoint this condition service would be registered right from the point the framework is initialized (ASAP).  When the kernel is run with a checkpoint action then this condition service would be registered at the point the checkpoint process has been restored.

I don't have a perfect name for this condition ID.  For the context of this discussion I will call it `io.openliberty.process.restored`.  I know it doesn't fit well for the case where this condition is registered ASAP on non-checkpoint launches.

A component that must only be enabled during non-checkpoint launches or after a checkpoint launch has been restored would have an `@Component` that looks like this:

```java
@Component(
           reference = {
                   @Reference(
                           service = Condition.class, 
                           target = "(" + Condition.CONDITION_ID + "=io.openliberty.process.restored)"
                   )
           }
)
```

tjwatson:
Fixes #20660

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
@@ -46,6 +46,15 @@ public enum CheckpointPhase {
      */
     public static final String CHECKPOINT_RESTORED_PROPERTY = "io.openliberty.checkpoint.restored";
 
+    /**
+     * The ID of the condition service that indicates the Liberty process is running. A Liberty process
+     * is considered running if Liberty was launched with no checkpoint or if the Liberty process
+     * has been restored from a checkpoint image. When Liberty is launched to create a checkpoint image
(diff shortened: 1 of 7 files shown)


tjwatson:
#build

tjwatson:
#build

tjwatson:
> For help analyzing your personal build, go to https://cognitive.hursley.ibm.com/buildAnalysis.html?uuid=_GJstYL_tEeyhmZ5zGPuLaA

The analysis didn't show green for the failures in com.ibm.ws.jmx.connector.client.rest_fat FAT tests but they match RTC defect 289481.

sebratton (approved):

tjwatson:
#libby

PR #20832, opened 2022-04-15
Reduce trace in DynamicJACCFeatureTest

kristip17:
Server startup can be slow on Windows, causing a `Start Server error: The server did not show kernel had started. The server did not show kernel had started. The standard error was: ....`. Try reducing the trace for this test.

RTC 290123


kristip17:
#build

jvanhill (approved):

Issue #20781, opened 2022-04-12
Add introspector for feature manager.


tjwatson:
It would have been useful many times to be able to get a dump that includes information about all features enabled at runtime, including all public, protected, private and auto-features.

PR #20855, opened 2022-04-18
Implemented feature introspector to output server feature definitions

KeenanMcCarthy:
In the event of a dump, the feature introspector will iterate over all server features and output
the feature name, as well as its visibility, version, simple name, and symbolic name. 
for #20781

### DIFF
--- a/dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/FeatureIntrospector.java
+++ b/dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/FeatureIntrospector.java
@@ -13,6 +13,7 @@ package com.ibm.ws.kernel.feature.internal.subsystem;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.Collection;
+import java.util.List;
 import java.util.Set;
 
 import org.osgi.service.component.annotations.Activate;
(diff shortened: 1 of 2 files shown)


KeenanMcCarthy:
#build

tjwatson:
#build

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureIntrospector.java
```diff
@@ -0,0 +1,67 @@
...
+     * Requests the set of features that are installed on the server and outputs
+     * relevant data to the log file
+     */
+    @Override
+    public void introspect(PrintWriter out) throws Exception {
+        Set<String> features = this.provisioner.getInstalledFeatures();
+        for (String feature : features) {
+            FeatureDefinition featureDef = this.provisioner.getFeatureDefinition(feature);
+            out.println(featureDef.getFeatureName() + "[");
+            out.println("\tvisibility: " + featureDef.getVisibility());
```
tjwatson: I think we should use something like 4 spaces here instead tabs.  I only say this because that seems to be what other introspectors do.

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureIntrospector.java
```diff
@@ -0,0 +1,67 @@
...
+            out.println("\tversion: " + featureDef.getVersion());
+            out.println("\tsimple name: " + featureDef.getFeatureName());
+            out.println("\tsymbolic name: " + featureDef.getSymbolicName());
+            out.println("]");
```
tjwatson: We should be able to safely cast the `featureDef` to `com.ibm.ws.kernel.feature.provisioning.ProvisioningFeatureDefinition` here.  I suggest we do an `instanceof` check and then cast and also provide the following information:

ProvisioningFeatureDefinition.getBundleRepositoryType()
ProvisioningFeatureDefinition.getConstituents(SubsystemContentType)
ProvisioningFeatureDefinition.getSupersededBy()
ProvisioningFeatureDefinition.isAutoFeature()
ProvisioningFeatureDefinition.isSingleton()

KeenanMcCarthy:
#build

KeenanMcCarthy:
#build

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/FeatureIntrospector.java
```diff
@@ -0,0 +1,94 @@
...
+            out.println("    Symbolic Name: " + featureDef.getSymbolicName());
+            if (featureDef instanceof ProvisioningFeatureDefinition) {
+                ProvisioningFeatureDefinition proFeatDef = (ProvisioningFeatureDefinition) featureDef;
+                String bundleRepType = proFeatDef.getBundleRepositoryType();
+                out.println("    Bundle Repository Type: " + (bundleRepType == null || bundleRepType.isEmpty() ? "core" : bundleRepType));
+                out.println("    Auto Feature: " + proFeatDef.isAutoFeature());
+                out.println("    Singleton: " + proFeatDef.isSingleton());
+                out.print("    Superseded By: ");
+                try {
+                    out.println(proFeatDef.getSupersededBy()); //Throws Exception if called outside of provisioning operation
```
tjwatson: I think we can work around this by using internals of this project/bundle.  First you will need to move this class into the package `com.ibm.ws.kernel.feature.internal.subsystem` so that you can use package private methods of that package.  Then you can do something like how `com.ibm.ws.kernel.feature.internal.subsystem.FeatureRepository.LibertyFeatureServiceFactory.getService(Bundle, ServiceRegistration<LibertyFeature>)` does to set the details so we can get this information.

KeenanMcCarthy:
#build

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/FeatureIntrospector.java
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                    boolean isSuperseded = details.isSuperseded();
+                    out.println("    Superseded: " + isSuperseded);
+                    if (isSuperseded) {
+                        out.println("        Superseded by: " + details.getSupersededBy());
+                    }
+                    Collection<FeatureResource> constituents = details.getConstituents(null);
+                    out.println("    Constituents: ");
+                    for (FeatureResource constituent : constituents) {
+                        out.println("        " + constituent.getSymbolicName());
```
tjwatson: I would like to see the tolerates information for each included FeatureResource.  Something like this.
```suggestion
                        List<String> tolerates = constituent.getTolerates();
                        String toleratesString = tolerates == null ? "" : ": tolerates:=" + tolerates;
                        out.println("        " + constituent.getSymbolicName() + toleratesString);
```

dev/com.ibm.ws.kernel.feature.core/src/com/ibm/ws/kernel/feature/internal/subsystem/FeatureIntrospector.java
```diff
@@ -0,0 +1,113 @@
...
+                        out.println("            " + constituent.getAttributes().toString());
+                    }
+                    if (detailsAreSet) {
```
tjwatson: Just after the initialization of `detailsAreSet` at line 82 I would add a try-finally block.  Then put this `if (detailsAreSet) {}` code inside the `finally` block.

tjwatson (commented):
I would be fine merging this as is.  But the build failed in weird ways.  I doubt it failed because of your change.  But since you have to run another build anyways we might as well get these suggestions in first.

KeenanMcCarthy:
Fixes #20781

KeenanMcCarthy:
#build

tjwatson:
#build

tjwatson:
There is a good build in #20998 with the same changes: https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_jtbk0MshEeyhmZ5zGPuLaA

The build issues have to do with the size of Keenan's github ID.  We can merge this change.

tjwatson (approved):

tjwatson:
#libby

PR #20618, opened 2022-03-28
Ensure JSONB TCK compatibility with Windows and Semaru

KyleAure:
The JSONB TCK is currently failing when running on Windows (any version) with Semaru (any version).
The failure is due to the Signature Test plugin failing to run in static mode because it does not recognize or thinks that a certain argument is missing.  This leads it to just output the "Help" message instead of running.

Example output: 
```txt
Calling:  com.sun.tdk.signaturetest.SignatureTest() with following args:
   testArguments[0] = -Static
   testArguments[1] = -CheckValue
   testArguments[2] = -mode
   testArguments[3] = src
   testArguments[4] = -Verbose
   testArguments[5] = -FileName
   testArguments[6] = C:\Users\JAZZ_B~1.EBC\AppData\Local\Temp\jakarta.json.bind.sig_3.0.0
   testArguments[7] = -Classpath
   testArguments[8] = C:\Users\jazz_build\Build\jazz-build-engines\wasrtc\EBCPROD\build\dev\io.openliberty.jakarta.jsonb.3.0_fat_tck\autoFVT\results\tck_defaultServer/signaturedirectory/jakarta.json.bind-api.jar:C:\Users\jazz_build\Build\jazz-build-engines\wasrtc\EBCPROD\build\dev\io.openliberty.jakarta.jsonb.3.0_fat_tck\autoFVT\results\tck_defaultServer/jdk11-bundle/java.base:C:\Users\jazz_build\Build\jazz-build-engines\wasrtc\EBCPROD\build\dev\io.openliberty.jakarta.jsonb.3.0_fat_tck\autoFVT\results\tck_defaultServer/jdk11-bundle/java.rmi:C:\Users\jazz_build\Build\jazz-build-engines\wasrtc\EBCPROD\build\dev\io.openliberty.jakarta.jsonb.3.0_fat_tck\autoFVT\results\tck_defaultServer/jdk11-bundle/java.sql:C:\Users\jazz_build\Build\jazz-build-engines\wasrtc\EBCPROD\build\dev\io.openliberty.jakarta.jsonb.3.0_fat_tck\autoFVT\results\tck_defaultServer/jdk11-bundle/java.naming
   testArguments[9] = -Package
   testArguments[10] = jakarta.json.bind
   testArguments[11] = -Exclude
   testArguments[12] = jakarta.json.bind.adapter
   testArguments[13] = -Exclude
   testArguments[14] = jakarta.json.bind.annotation
   testArguments[15] = -Exclude
[...]

PR #20870, opened 2022-04-19
Skip JSON-B TCK Signature Tests on Windows

KyleAure:
This change is to expand on #20618 
Our SOE testing has shown that the signature plugin also fails on other JDKs not just semaru on windows.


tabishop (approved):

KyleAure:
#build

PR #20889, opened 2022-04-21
test getContextService from executor definitions

njr-11:
Test that when getContextService is used to obtain a ContextService from managed executors that are defined by `@ManagedExecutorDefinition` and `@ManagedScheduledExecutorDefinition`, the same executor remains associated with the context service and enforces its maxAsync on async dependent stages that are created by the context service.
In this test, we have two executor definitions with differing maxAsync constraints each pointing at the same `@ContextServiceDefinition`.


njr-11:
#build

mswatosh (approved):

PR #20900, opened 2022-04-21
Skip custom RW FATs on iSeries for acmeCA

kristip17:
We have some acmeCA tests that programmatically set some config related files to unreadable/unwritable in the JUNIT FAT. This works on most OS, but not Windows and not iSeries currently. Add a check and skip on iSeries OS.

RTC 290401


kristip17:
#build

jvanhill (approved):

kristip17:
#libby

Issue #20908, opened 2022-04-23
Default session meta cache name failed with RH DataGrid


zhang-ca:
Describe the bug
RH DataGrid or Datagrid Administration GUI Console unable to find percent encoded infinispan caches name

Steps to Reproduce
When Liberty create infinispan caches name with percent encoding, Datagrid Administration GUI Console failed to retrieve the cache name. 

Diagnostic information:

    OpenLiberty Version: [all]

Additional context
Example infinispan caches name:
com.ibm.ws.session.meta.default_host%2FGestionPedidos
- “GestionPedidos” is the application web context
- “%2F” is the encoded character for “/”
RH DataGrid failed to retrieve the above cache name.

PR #20909, opened 2022-04-23
Cache name separator property

zhang-ca:
Session custom property CacheSeparator to customize the cache naming

cacheSeparator=Cache Name Separator
cacheSeparator.desc=The single character used to separate the session meta cache name. The default value % should usually be used.


zhang-ca:
fixes #20908

zhang-ca:
for #20908

zhang-ca:
#build

zhang-ca:
#build

PR #20926, opened 2022-04-26
Optimize service lookup by filter

tjwatson:
This is to pull in https://github.com/eclipse-equinox/equinox.framework/pull/46 before it is released.


tjwatson:
#build

tjwatson:
#build

PR #20947, opened 2022-04-28
AppSecurity-5.0 still requires request context

Azquelt:
Add appSecurity-5.0 (EE10) to the list of features which mean the
WebAppSecurityCollaborator reports that CDI is needed.

Fixes RTC 289901


Azquelt:
I suspect this is the cause of the error we're seeing in the security TCK:
```
org.jboss.weld.contexts.ContextNotActiveException: WELD-001303: No active contexts for scope type jakarta.enterprise.context.RequestScoped
```

Azquelt:
#build

hutchig (approved):
One does wonder if a form of semantic 'contains' that is version independent might be of use in the future.

jvanhill (approved):
Looks good.

When I added 4.0, I was not a big fan of the fact we are looking for features in the code, especially features that have a tendency to change revisions quite often such as appSecurity. Not sure if there is much else we can do here.

EDIT: I should mention, I had made it version check to consider future revisions, but it was a performance degradation, so I had to remove it:

https://github.com/OpenLiberty/open-liberty/commit/652e439435abd7be84b568bcdef49de29306d483

PR #20957, opened 2022-04-28
optimize and better debug for fixed-rate scheduled task

njr-11:
Additional trace points are needed for debugging the computations involved in fixed-rate scheduled tasks.
Also, in most cases we should be able to skip those computations entirely by just adding the fixed-rate amount to the expected start time of the current execution if the resulting time is past the current time.  The more complex computation is only needed when what would have been the execution has been overlapped.


jmstephensgit (approved):

njr-11:
#build

njr-11:
#build

njr-11:
#build

PR #20977, opened 2022-04-29
remove compilation workaround for Trigger and LastExecution

njr-11:
The workaround to declare the Trigger and LastExecution method parameters as Object is no longer needed now that projects can depend on Jakarta-transformed bundles.


njr-11:
#build

tabishop (approved):

Issue #20981, opened 2022-05-02
ArrayOutOfBounds exception on z/OS with either full or JMX audit events enabled on shutdown 


emilytee:
**Describe the bug**  
ArrayOutOfBounds exception with either full or JMX audit events enabled on shutdown 

Stack Dump = java.lang.ArrayIndexOutOfBoundsException: Array index out of range: 10
	at com.ibm.ws.security.audit.source.AuditServiceImpl.sendEvent(AuditServiceImpl.java:483)
	at com.ibm.ws.request.probe.audit.servlet.AuditPE.auditEventJMXMBeanRegister01(AuditPE.java:648)
	at com.ibm.ws.request.probe.audit.servlet.AuditPE.processCounter(AuditPE.java:184)
	at com.ibm.ws.request.probe.RequestProbeService.processAllCounterProbeExtensions(RequestProbeService.java:231)
	at com.ibm.wsspi.request.probe.bci.TransformDescriptorHelper.exitHelper(TransformDescriptorHelper.java:195)
	at com.ibm.ws.security.audit.Audit.audit(Audit.java:68)
	at com.ibm.ws.security.audit.pipeline.AuditJMXPipeline.emitJMXMBeanRegisterEvent(AuditJMXPipeline.java:544)
	at com.ibm.ws.security.audit.pipeline.AuditJMXPipeline.unregisterMBean(AuditJMXPipeline.java:509)
	at com.ibm.ws.kernel.boot.jmx.internal.PlatformMBeanServer.unregisterMBean(PlatformMBeanServer.java:470)
	at com.ibm.websphere.monitor.meters.MeterCollection.lambda$MXBeanHelper$2(MeterCollection.java:149)
	at java.security.AccessController.doPrivileged(AccessController.java:738)
	at com.ibm.websphere.monitor.meters.MeterCollection.MXBeanHelper(MeterCollection.java:148)
	at com.ibm.websphere.monitor.meters.MeterCollection.remove(MeterCollection.java:194)
	at com.ibm.ws.jaxrs.monitor.JaxRsMonitorFilter.cleanApplication(JaxRsMonitorFilter.java:320)
	at com.ibm.ws.jaxrs.monitor.MonitorAppStateListener.applicationStopped(MonitorAppStateListener.java:53)
	at com.ibm.ws.container.service.state.internal.ApplicationStateManager.fireStopped(ApplicationStateManager.java:144)
[...]

PR #20984, opened 2022-05-02
Issue20981-AOOBException

emilytee:


una-tapa (approved):
- The list no longer has max in order to avoid ArrayIndexOutOfBoundsException.
- I checked if any object can get NULL - looks good.

una-tapa:
Fixes #20981

PR #21005, opened 2022-05-04
Ensure Oracle UCP 2PC works

KyleAure:
Use two different data sources to ensure that we participate in a 2PC when testing Oracle UCP XA connections.


frowe (approved):

KyleAure:
#build

Issue #20402, opened 2022-03-12
CRIU: Support Non-persistent EJB Timers created during checkpoint


tkburroughs:
An application may be configured to automatically create EJB timers during deployment (both persistent and non-persistent). For CRIU, the EJB Container should be updated to determine when a checkpoint is occurring and defer the creation of automatic non-persistent timers until after restore, or reschedule the start time upon restore.

**Background**
EJB timers in general (including "non-persistent" timers) have interesting semantics around scheduling. Specifically, EJB timers are expected to schedule next timeout operations based on when an application creates the timer, rather than when the last timeout occurred.

For example, if an application creates a timer at exactly 3:00, which is scheduled to run every 5 minutes; then the next timeout is calculated from 3:00, and not just the last time it successfully ran. If at 3:30, the server is stopped, and then re-started an hour later at 4:30, then the next scheduled timeout is still 3:35 in the past.... so when the server re-starts, the timer will run 12 times immediately to catch up on all the missed expiration.

Therefore, when using CRIU, if an image of the server is captured today, which includes existing EJB timers (either persistent or non-persistent), then every time that image is used, the timers will run "catch-up" timeouts from the point the CRIU image was captured. The number of "catch-up" timeouts will increase over time. If the image is restarted an hour later, there will be 12 catch-up timeouts, after 2 hours, it would be 24 etc.

One option considered for CRIU for automatically created EJB timers would be to defer the creation of the timers until after restore. likely some type of hook or listener would need to be implemented to create the timers after restore. Unfortunately this solution would require considerable change effecting both the automatic and programmatic creation of timers and would not be compliant with the EJB specification regarding the transactional nature of timer creation.

**Solution**

The implemented solution is as follows:
1. Non-persistent timers (whether automatically created by the container or programmatically created during application start from a `@Startup @Singleton`) are created normally.
2. When any checkpoint phase is enabled, the execution of all above created non-persistent timers is paused until after restore.
3. On restore; the start time of each non-persistent timer is recalculated to ensure the first timeout is after the time of restore (to avoid timers running repeatedly for catch-up processing).
4. If the EJB Container is restarted for any reason after restore, non-persistent timer processing is as if checkpoint were not enabled.

**Note:** Persistent timers have additional considerations that will be addressed by #18778

tjwatson:
@tkburroughs Is there more work needed on this one?

tkburroughs:
[...]

PR #21007, opened 2022-05-04
Issue #20402: Pause non-persistent timers until Checkpoint restore

tkburroughs:
Non-persistent timer service will be paused on checkpoint and resume
on checkpoint restore; next timeout will be advanced past the
time of restore, skipping missed expirations

for #20402

### DIFF
--- a/dev/io.openliberty.checkpoint_fat/test-applications/ejbapp1/src/ejbapp1/TimerTest.java
+++ b/dev/io.openliberty.checkpoint_fat/test-applications/ejbapp1/src/ejbapp1/TimerTest.java
@@ -18,6 +18,7 @@ import java.util.concurrent.atomic.AtomicLong;
 
 import javax.ejb.Schedule;
 import javax.ejb.Singleton;
+import javax.ejb.Timer;
 
 @Singleton
 public class TimerTest {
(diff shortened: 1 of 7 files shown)


tkburroughs:
#build

dev/io.openliberty.checkpoint_fat/test-applications/ejbapp1/src/ejbapp1/TimerTest.java
```diff
@@ -0,0 +1,50 @@
...
+            return;
+        }
+        long deltaTime = currentTime - lastTime;
+        long deltaMillis = TimeUnit.NANOSECONDS.toMillis(deltaTime);
+        // here we assume if delta is less than 700 millis then something is wrong
+        if (deltaMillis < 800) {
+            failed.set(true);
+        }
+        if (numRun.addAndGet(1) == 5) {
+            System.out.println("TIMER RUN 5 TIMES");
```
tkburroughs, line 43: I would recommend that this method be changed to accept a Timer parameter, and then add a timer.cancel() here, once the timer has expired 5 times.  While the test should work as coded, there will likely be intermittent build break issues opened due to the timer continuing to run when the test finishes and attempts to shutdown the server.

Under normal conditions, the test will probably complete and shutdown the server before the timer runs again; however, due to the sometimes erratic performance behavior of the test systems, there can be pauses that will result in the timer running before shutdwon, or more specifically during the server shutdown.  While the EJB Timer Service blocks timers running once server shutdown is initiated, if a timer is already running, it can result in various errors and or FFDC, which would need to be allowed by the test.  Adding a timer.cancel() eliminates that concern.

tkburroughs:
#libby

olendvcook (approved):

tkburroughs:
#libby

tkburroughs:
#build

PR #21008, opened 2022-05-04
port jca_bvt.example bucket from WebSphere Liberty and switch to Jakarta

njr-11:
Port the com.ibm.ws.jca_bvt.example bucket to a FAT bucket in Open Liberty. Then switch it to use Jakarta EE feature and packages.  This will be a good bucket to later switch to the Jakarta Connectors 2.1 API because it uses/implements some of the interfaces, both in the (mock resource adapter as well as in the test application) that generics have been added to in 2.1.


njr-11:
#build

mswatosh (approved):

PR #21010, opened 2022-05-05
Jsonp 2.1 container testing

KyleAure:
Update current tests that use the jsonpContainer-x.x features to re-run using the latest 2.1 feature.


dev/com.ibm.ws.jsonb_fat/build.gradle
```diff
@@ -8,34 +8,82 @@
...
+task addYasson3(type: Copy) {
+  from configurations.yasson3
+  into "${buildDir}/autoFVT/publish/shared/resources/yasson/3.0.0-RC2/"
+  rename "yasson-.*.jar", "yasson.jar"
 }
 
 addRequiredLibraries {
   dependsOn addJohnzon
-  dependsOn addRefImpls
+  dependsOn addJonzonJakarta
```
nmittles: Missing 'h' in 'Jonzon'?  See line 13 as well.

dev/com.ibm.ws.jsonb_fat/fat/src/com/ibm/ws/jsonb/fat/FATSuite.java
```diff
@@ -38,32 +49,58 @@
...
+                                    .forServers(servers)
+                                    .removeFeatures(new HashSet<>(Arrays.asList("usr:testFeatureUsingJsonb-1.0", "usr:testFeatureUsingJsonb-2.0")))
+                                    .addFeature("usr:testFeatureUsingJsonb-3.0")
+                                    .removeFeatures(new HashSet<>(Arrays.asList("usr:testFeatureUsingJsonp-1.1", "usr:testFeatureUsingJsonp-2.0")))
+                                    .addFeature("usr:testFeatureUsingJsonp-2.1"));
 
+    //JSONB providers
     public static final String PROVIDER_YASSON = "org.eclipse.yasson.JsonBindingProvider";
-    public static final String PROVIDER_JOHNZON = "org.apache.johnzon.jsonb.JohnzonProvider";
+    public static final String PROVIDER_JOHNZON_JONSB = "org.apache.johnzon.jsonb.JohnzonProvider";
```
nmittles: '_JSONB' instead of '_JONSB'?

nmittles (approved):
Approved.  Added two typo related comments to consider.

dev/com.ibm.ws.jsonb_fat/fat/src/com/ibm/ws/jsonb/fat/FATSuite.java
```diff
@@ -113,8 +159,125 @@ public static void beforeSuite() throws Exception {
...
+
+    /**
+     * Returns the jsonp provider class name that is available to the applications running
+     * based on the jakarta level being used.
+     *
+     * @param isAvailable - provide a false value if you want a provider class name that ISN'T available for negative testing.
+     * @return String - classname
+     */
+    public static String getJsonpProviderClassName(boolean... isAvailable) {
+        if (isAvailable.length > 1) {
```
tabishop, line 236: Is there a reason we want to allow someone to call this with multiple input parms and get an exception back instead of just having 2 methods with one of them taking a single boolean and the other like this:
public static String getJsonpProviderClassName(){
return Fatsuite.getJsonpProviderClassName(true);
}

Same comment apples to getJsonbProviderClassName

tabishop (approved):
Only one minor nit/question, but not going to hold up approval over it.

dev/com.ibm.ws.jsonb_fat/fat/src/com/ibm/ws/jsonb/fat/FATSuite.java
KyleAure: So realistically I could have just made this a single boolean method where a "user" would provide true if they wanted the provider class name that should be loaded, and false if they wanted a provider class name that should NOT be loaded. 

But really the true case should be the default, and they should optionally be able to provide a false value if they want to perform negative tests. However, java doesn't support optional method parameters except for variable-length parameters (using the ... syntax).  So this is my shining example of how support for optional parameters could be useful in java.

KyleAure:
Waiting until artifacts are uploaded to Artifactory before performing a build.

KyleAure:
#build

KyleAure:
#libbybot

Issue #14594, opened 2020-10-21
httpSessionCache Config issue ?


c-koell:
**Describe the bug**
I have tried to run the example application from https://openliberty.io/guides/sessions.html on my local pc without docker.
I have opened a issue (https://github.com/OpenLiberty/guide-sessions/issues/89) because the example doesn't worked.
As @gkwan-ibm mentioned that i have to use the same uri (as described in the issue). With the same uri the session cache worked fine.
Now my question is how should that work if i deploy the application to 2 servers with different os (windows/linux) ?

I tried it on my local pc to point the uri to a http based endpoint where i can read the hazelcast-config.xml

Example server.xml:
````
<httpSessionCache libraryRef="jCacheVendorLib"
        uri="https://files.int.srv/rest/Configuration/Files/Liberty/example/hazelcast-config.xml" />
````
With that uri the sessionCache does not work anymore on my local pc if i start 2 instances of the example application.

**Expected behavior**
A clear and easy way to configure a session cache.

**Diagnostic information:**
 - OpenLiberty Version: 20.0.0.10
 - Java Version: Eclipse OpenJ9 VM, Version 11.0.8+10

Maybe i don't understand something here. 
Can somebody explain me how i should configure a sessionCache with 2 servers (windows/linux) or why the sessionCache don't work locally if i use a http uri.

Thanks much !

**Comment by nmittles** [comment]
Hello @c-koell 

Instead of specifying the `uri` attribute on `httpSessionCache` can you instead provide the configuration file via the `hazelcast.config.location` on `properties` like the example below?

```
 <httpSessionCache libraryRef="jCacheVendorLib" >
        <properties hazelcast.config.location="file:${server.config.dir}/hazelcast-config.xml"/>
 </httpSessionCache>
```

c-koell:
Hi @nmittles !

That's it ! If i use the property the sessionCache works fine. I haven't found anything about that behaviour in the documentaion.
Thanks for help. Do you think it should better documented ?

greets
claus

c-koell:
Ah i haven't read https://github.com/OpenLiberty/guide-sessions/issues/89 really good, because @gkwan-ibm mentioned the property already ... sorry
But as written ... if you don't know it .. it doesn't work ;-)

c-koell:
I will close this issue since it works now. Maybe someone extends the documentation about the different config properties.

zhang-ca:
[...]

PR #21015, opened 2022-05-05
Set uri value as part of the vendorProperties

zhang-ca:
The uri attribute directly on the httpSessionCache following JCache API:
javax.cache.spi.CachingProvider.getCacheManager(uri, classLoader, vendorProperties)
The getCacheManager API states, "Multiple calls to this method with the same URI and ClassLoader must return the same CacheManager instance". The URI is different in the some scenarios, since ${server.config.dir} is different for the two Liberty servers. This results in two separate cache managers --> different JCaches --> unshared sessions.

Therefore pass the uri value as part of the vendorProperties to ensure the uri attribute method works with latest Hazelcast versions.


zhang-ca:
fixes #14594

zhang-ca:
for #14594

zhang-ca:
#build

bensonlatibm (approved):
lgtm!

zhang-ca:
#build

PR #21028, opened 2022-05-06
Fix isEnabled impl for JPA31 Runners

jgrassel:
`RepeatWithJPA31.java` overrode the `isEnabled()` method and just returns true (which clobbers `componenttest.rules.repeater.FeatureReplacementAction.isEnabled()` 's jdk level check):

```
public class RepeatWithJPA31 extends JakartaEE10Action implements RepeatTestAction {
    public static final String ID = "JPA31";

    [@Override](https://wasrtc.hursley.ibm.com:9443/jazz/users/Override)
    public boolean isEnabled() {
        return true;
    }
```

This copy/paste mistake needs to be removed in order for the JPA 3.1 FAT portion to not be executed with JDKs < 11.


jgrassel:
#build

tkburroughs (approved):

jgrassel:
pushed an update to include a fix for com.ibm.ws.persistence_fat, effectively the same issue.

jgrassel:
#run-libby-bot

PR #21062, opened 2022-05-10
OLGH21060: Correct Service Release and Fixpack processing in JavaInfo

### LINKED ISSUE #21060: Correct Service Release and Fixpack processing in JavaInfo
There exists Java implementations (such as Java Semeru for z/OS) which can return unexpected values for the `java.runtime` property, such as `IBM Semeru Runtime Certified Edition for z/OS (11.0.15-internal+0-adhoc.JENKINS.BuildJDK11s390xzosRelease)`.  When trying to process a Service Release number out of this, JavaInfo submits "" as an argument to `JavaInfo.parseIntSafe()`.

Before calling `parseIntSafe()`, there should be a `len > 0` check, as len will not increment past 0 if there is a non-digit character after the "sr" it finds in the string.

This is a low-risk bug, since the NumberFormatException is caught and logged with FFDC, and `parseIntSafe()` returns 0.  However, the extra FFDC can be avoided with some extra value checking.

jgrassel:
Fixes #21060


jgrassel:
#build

jhanders34 (approved):

PR #21085, opened 2022-05-12
Delete transformed jar after use

benjamin-confino:
We are suspicious that the transformer might not be overwriting a jar if called twice, so we delete the first output.


benjamin-confino:
#build

tevans78 (approved):

PR #21092, opened 2022-05-12
Add a CDI 1.0 beans.xml test to CDI 4.0 bucket

tevans78:
We initially assumed that CDI 4.0 would treat a v1.0 beans.xml (not empty but no version specified) the same way as an empty beans.xml file. This turns out not to be the case and upon closer inspection of the spec, it does not really cover this case. Therefore it is perfectly correct for Weld to just use the default bean discovery mode (ANNOTATED). This PR adds a test for that.

#build


benjamin-confino (approved):

tevans78:
#libby

PR #21116, opened 2022-05-13
Add default LTPA keys for ConfigManagerFeatureTest

kristip17:
Add default LTPA keys instead of creating from scratch for this short run test, `ConfigManagerFeatureTest` -- otherwise the test is over before the LTPA keys are finished on slower build systems, causing FFDCs to be flagged.

RTC 290524


kristip17:
#build

jvanhill (approved):

PR #21119, opened 2022-05-13
use final 3.0.0 TCK for Concurrency

njr-11:
Point the test bucket for running the Concurrency 3.0 TCK at the final release copy from Maven instead of staging.


KyleAure (approved):

njr-11:
#build

PR #21209, opened 2022-05-20
Nonship property to enable orm xml to be provided for unannotated entities

njr-11:
This will be useful for experimentation with ideas being considered for Jakarta Data.

Also included is a bug fix to a NullPointerException that prevented components other than Persistent Executor and Batch from using the persistence service because they aren't recognized as providing an entity within a special set.  I added an enumeration value NONE to indicate this rather than using a null which causes errors elsewhere when the value is assumed to be non-null.


jgrassel (approved):

njr-11:
#build

njr-11:
All tests that started were successful.  However, it looks like some sort of infrastructure issue might have prevented additional buckets from being attempted, so I'm resubmitting the build:
#build

njr-11:
In the latest build, all tests that started were again successful.  BVT tests were not attempted due to inability to reach RTC, so I have resubmitted just the async BVT build.

njr-11:
The resubmit of the async BVT was successful,
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_hEnNodwmEeypPP_gDLm9Dg

PR #21210, opened 2022-05-20
Pre-initalize SSLSocketFactory during checkpoint

dettmoney:
Profiling criu restore time to first response using acmeair microservices applications shows time being spent in javax/net/ssl/SSLSocketFactory.getDefault() during the first request to the application. If we can move this call into the criu checkpoint phase, we can reduce the restore time up to 20%.

### DIFF
--- a/dev/com.ibm.ws.ssl/src/com/ibm/ws/ssl/internal/SSLComponent.java
+++ b/dev/com.ibm.ws.ssl/src/com/ibm/ws/ssl/internal/SSLComponent.java
@@ -52,6 +52,8 @@ import com.ibm.ws.ssl.provider.AbstractJSSEProvider;
 import com.ibm.wsspi.kernel.service.location.WsLocationAdmin;
 import com.ibm.wsspi.kernel.service.location.WsLocationConstants;
 
+import io.openliberty.checkpoint.spi.CheckpointPhase;
+
 /**
  * Component for the SSL configuration bundle.
  */
@@ -85,6 +87,13 @@ public class SSLComponent extends GenericSSLConfigService implements SSLSupportO
     private boolean transportSecurityEnabled;
 
     private ExtComponentContext componentContext;
+
+    private final CheckpointPhase checkpointPhase;
+
+    @Activate
+    public SSLComponent(@Reference(cardinality = ReferenceCardinality.OPTIONAL) CheckpointPhase checkpointPhase) {
+        this.checkpointPhase = checkpointPhase;
+    }
 
     /**
      * DS method to activate this component.
@@ -332,6 +341,9 @@ public class SSLComponent extends GenericSSLConfigService implements SSLSupportO
                                                              isServer,
                                                              transportSecurityEnabled,
                                                              repertoirePIDMap);
+                if (checkpointPhase != null) {
+                    SSLSocketFactory.getDefault();
+                }
             } catch (SSLException e) {
                 if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
                     Tr.event(tc, "Exception processing SSL configuration; " + e);
--- a/dev/com.ibm.ws.ssl/bnd.bnd
+++ b/dev/com.ibm.ws.ssl/bnd.bnd
@@ -38,6 +38,7 @@ Import-Package: \
 	sun.security.pkcs11;resolution:=optional, \
     com.ibm.security.cmskeystore;resolution:=optional, \
     !*.internal.*, \
+    ${defaultPackageImport}, \
     *
 
 Private-Package: \
--- a/dev/com.ibm.ws.ssl/test/com/ibm/ws/ssl/internal/SSLComponentTest.java
+++ b/dev/com.ibm.ws.ssl/test/com/ibm/ws/ssl/internal/SSLComponentTest.java
@@ -176,7 +176,7 @@ public class SSLComponentTest {
 
     @Before
     public void setUp() {
-        sslComponent = new SSLComponent();
+        sslComponent = new SSLComponent(null);
 
         mock.checking(new Expectations() {
             {


dettmoney:
#build

dettmoney:
#build

acdemyers:
HI @dettmoney I don't know when this taking place.  My concert is an oddity we have to deal with in the JDK.   We have to make sure the Liberty SSL initialization takes place before the SSLSocketFactory.getDefault() takes place.   We setup a Security properties to point to a Liberty custom socket factory.   This has to take place before a call to SSLSocketFactory.getDefault() takes place or out dynamic outbound SSL feature will not work.

The description makes it sound like the change will take place during application access??   We just need to make sure it happens after Liberty SSL initialization takes place.

dev/com.ibm.ws.ssl/src/com/ibm/ws/ssl/internal/SSLComponent.java
```diff
@@ -332,6 +341,9 @@ private synchronized void processConfig(boolean updateSSLConfigManager) {
                                                              isServer,
                                                              transportSecurityEnabled,
                                                              repertoirePIDMap);
+                if (checkpointPhase != null) {
+                    SSLSocketFactory.getDefault();
```
tjwatson, line 345: We need to check with the OpenJ9 team to make sure this is safe to do on the checkpoint side.  We need to ask @alon-sh.

dettmoney:
@acdemyers I looked at a liberty ssl trace, and when my code is run at checkpoint time, the SSLSocketFactory.getDefault() seems to invoke the LibertySSLSocketFactory constructor. Based on your comment, does that mean it's doing the right thing?

A small snippet of the trace output:
```
[5/25/22, 2:01:16:008 EDT] 0000001f id=00000000 com.ibm.ws.ssl.config.SSLConfigManager                       < initializeSSL Exit
[5/25/22, 2:01:16:008 EDT] 0000001f id=00000000 SystemOut                                                    O ***JTD: calling SSLSocketFactory.getDefault() from processConfig
[5/25/22, 2:01:16:014 EDT] 0000001f id=00000000 com.ibm.ws.ssl.protocol.LibertySSLSocketFactory              > LibertySSLSocketFactory Entry
[5/25/22, 2:01:16:014 EDT] 0000001f id=00000000 com.ibm.ws.ssl.config.SSLConfigManager                       > getDefaultSystemProperties Entry
[5/25/22, 2:01:16:014 EDT] 0000001f id=00000000 com.ibm.ws.ssl.config.SSLConfigManager                       > parseDefaultSecureSocketLayer Entry
[5/25/22, 2:01:16:015 EDT] 0000001f id=00000000 com.ibm.ws.ssl.JSSEProviderFactory                           > getInstance: null Entry
[5/25/22, 2:01:16:015 EDT] 0000001f id=00000000 com.ibm.ws.ssl.JSSEProviderFactory                           < getInstance: com.ibm.ws.ssl.provider.IBMJSSEProvider@42f70f5a Exit
```
....

alon-sh:
I was looking at this code - it should be disabled during checkpoint, JSSE should not be initialized at all during checkpoint. Also this code where a new security provider is inserted into the security providers list: 
```
AccessController.doPrivileged(new PrivilegedAction<Object>() {
            @Override
            public Object run() {
                Provider provider = Security.getProvider(contextProviderPriv);

                if (provider == null) {
                    try {
......
                            Security.addProvider(provider);
                        }
                    } catch (Exception e) {
                        if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
                            Tr.debug(tc, "Exception loading provider: " + contextProviderPriv + "; " + e);
                    }
                }
                return null;
            }
        });
```

is not a good idea as it could circumvent various restrictions (CRIU and FIPS). It is actually on the list of to-dos to disable the Security.addProvider() for CRIU checkpoint mode (and FIPS).  So the code will stop working and throw an error (as it should during CRIU checkpoint).

I recommend enumerating the existing providers and if neither IBMJSSE2 nor SunJSSE are available - throw an error saying there is an error in JDK security configuration. and disabling it for CRIU checkpoint mode.

acdemyers:
@dettmoney After closer examination of the code I realized that the SSL socket factories are being set prior to your code.

acdemyers:
@alon-sh That is some code that I was looking at recently and actually look ups the history of why it's there and as it turns out it is some code carried from tradition and really should be redone.  I agree an exception should be thrown if no JSSE is found.

dettmoney:
@alon-sh mentioned the call to `Security.addProvider`. Is that called as part of the `SSLSocketFactory.getDefault()` path? Must we delay the getDefault call until restore time?

acdemyers:
The call to addProvider would happen during server startup.  But mostly will not get called.  It some really old code that is a carry over from tWAS that needs to be redone.   It would only get called if a JSSE is not present.

From: Joshua Dettinger ***@***.***>
Date: Friday, June 24, 2022 at 2:23 PM
To: OpenLiberty/open-liberty ***@***.***>
Cc: Alaine DeMyers ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [OpenLiberty/open-liberty] Pre-initalize SSLSocketFactory during checkpoint (PR #21210)
@alon-sh mentioned the call to Security.addProvider. Is that called as part of the SSLSocketFactory.getDefault() path? Must we delay the getDefault call until restore time? — Reply to this email directly, view it on GitHub, or ‍ ‍ ‍ ‍ ‍ ‍
ZjQcmQRYFpfptBannerStart
This Message Is From an External Sender
This message came from outside your organization.
ZjQcmQRYFpfptBannerEnd

@alon-sh<https://github.com/alon-sh> mentioned the call to Security.addProvider. Is that called as part of the SSLSocketFactory.getDefault() path? Must we delay the getDefault call until restore time?

alon-sh:
we should delay all SSL code to after checkpoint due to potential security issues. we should ensure that addProvider() is not called but rather move this code path to restore time.

In terms of performance, there may be a performance impact - the security related code path not only do their specific thing but also will cause a lazy load of security providers in the JDK that are needed for those operations.

PR #21225, opened 2022-05-23
Fix EE10 app transforms during repeats

KyleAure:
If an application was previously transformed using EE9 the archive will persist when we try to repeat for EE10.
When this happens the transformer will see the `.jakarta` archive already exists and fail to transform the app for EE10. 
In situations where the archive produced for EE10 and EE9 need to be different, we need to make sure to delete the archive between repeats.


KyleAure:
Worked this fix into https://github.com/OpenLiberty/open-liberty/pull/21010 instead.

PR #21273, opened 2022-05-30
Ensure we run some CDI core tests in lite mode

Azquelt:
Don't try to run only the EE10 repeat if on Java version < 11. EE10
requires Java 11 so all tests will be skipped.

Fixes RTC 290805


Azquelt:
#libby #build

tevans78 (approved):

PR #21281, opened 2022-05-31
Move and Simplify loginMethod app check

kristip17:
Windows machines are taking a long time to start the test app for `StaticAnnotationsTest`, my earlier fix did some Windows checks and didn't realize we were already doing a `CWWKZ0001I`. Simplified the test for all OS doing one `CWWKZ0001I` check with a longer wait.

RTC 290799


kristip17:
#build

kristip17:
It looks like some of the extra at.build.response.for.child-build.featureStart.mToZ_fat0 children timed out (Wendy says these are a bug anyway). The full fat for this test ran successfully.

jvanhill (approved):

PR #21359, opened 2022-06-06
Update JPA Entity bucket's Postgres DDL files

jgrassel:
The Postgresql DDL files for the JPA Entity bucket needed to take advantage of "CREATE TABLE IF NOT EXISTS" and "DROP TABLE IF EXISTS" as Postgresql is a bit less forgiving than other databases, and will mark the transaction for rollback on any failure.  This causes test cases to fail due to missing tables.


jgrassel:
#build

dazey3 (dismissed):
It looks like the HIBERNATE DDL files were missed. Otherwise, LGTM

jgrassel:
#run-libby-bot

tkburroughs (approved):

PR #21370, opened 2022-06-07
Give a warning when using v1.0 beans.xml with cdi-4.0

### LINKED ISSUE #21096: Attempt to output a useful message if a v1.0 beans.xml is used in CDI 4.0
In CDI 4.0, a v1.0 beans.xml file will end up being treated as bean-discovery-mode = ANNOTATED. This is different to previous versions where it would have been the equivalent of ALL.

Since we are not going to provide a means for backwards compatibility in this case, we should attempt to output a useful message.

tevans78:
In CDI 4.0/Weld 5.0, a non-empty unversioned beans.xml file will now be considered to be bean-discovery-mode=annotated. In previous versions this would have been considered to be "all". We are not providing a backwards compatibility option for this but we are going to output a warning when a DeploymentException occurs.


tevans78:
#build

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
```diff
@@ -172,20 +176,26 @@ public Void run() {
...
+                webSphereCDIDeployment = null;
             }
 
+        } catch (DeploymentException e) {
+            if (webSphereCDIDeployment != null) {
+                CDIContainerEventManager eventManager = this.cdiRuntime.getCDIContainerEventManager();
+                if (eventManager != null) {
+                    eventManager.processDeploymentException(webSphereCDIDeployment, e);
+                }
+            }
```
Azquelt: Does this block need to re-throw the `DeploymentException` ?

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
```diff
@@ -718,4 +720,27 @@ public CDI<Object> getCDI() {
...
+                if (parsedURL != null) {
+                    //if the beans.xml was not an empty file then check if the version was set or not
+                    boolean unversionedBeansXml = beansXml.getVersion() == null;
+                    if (unversionedBeansXml) {
+                        URL unversionedBeansXmlURL = bda.getBeansXmlResourceURL();
+                        unversionedBeansXmlURLs.add(unversionedBeansXmlURL);
+                    }
+                }
+            }
+        }
```
Azquelt: I would suggest non-nested ifs that `continue` would be easier to read than all these nested if blocks.

This could also be done with streams though it might be harder to add comments to explain what each step means.

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
```diff
@@ -0,0 +1,103 @@
...
+        //ANNOTATED. Unlike an empty beans.xml, the legacy configuration option has no effect. Therefore we will output a warning when the
+        //DeploymentException occurs
+        WebArchive legacyBeans10XMLWar = ShrinkWrap.create(WebArchive.class, LEGACY_BEANS10_APP_NAME + ".war")
+                                                   .addClass(UnannotatedErrorServletBeans10.class)
+                                                   .addClass(UnannotatedBean.class)
+                                                   .addClass(RequestScopedBean.class);
+        CDIArchiveHelper.addBeansXML(legacyBeans10XMLWar, CDIVersion.CDI10);
+        ShrinkHelper.exportDropinAppToServer(server, legacyBeans10XMLWar, DeployOptions.SERVER_ONLY);
+
+        server.startServer(true, false);
```
Azquelt: Need to explain what `true, false` means here.

Alternatively, it might be clearer to use the `DISABLE_VALIDATION` option in the call to `exportDropinAppToServer`

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Test
+    public void testUnversionedBeansXmlWarning() throws Exception {
+        List<String> warningMessages = server.findStringsInLogs("CWOWB1018W:.*");
+        assertTrue("Message CWOWB1018W not found", warningMessages.size() > 0);
+        assertEquals("Message CWOWB1018W was found more than once", 1, warningMessages.size());
+    }
```
Azquelt: I think you should assert that both apps have failed to start before checking for the message.

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
```diff
@@ -0,0 +1,103 @@
...
+public class LegacyConfigTestError extends FATServletClient {
+    public static final String SERVER_NAME = "CDI40Server";
+
+    public static final String LEGACY_EMPTY_BEANS_APP_NAME = "UnannotatedErrorServletEmptyBeansXml";
+    public static final String LEGACY_BEANS10_APP_NAME = "UnannotatedErrorServletBeans10";
+
+    @Server(SERVER_NAME)
+    @TestServlets({
+                    @TestServlet(servlet = UnannotatedErrorServletEmptyBeansXml.class, contextRoot = LEGACY_EMPTY_BEANS_APP_NAME)
+    })
```
Azquelt: I don't think you should have a test servlet if the app is not expected to deploy. The test framework may even try to call the servlet despite the app containing it not starting.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
tevans78: ah yeah, it was there before but i missed the else block when i moved the "process" part out to a different service

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
tevans78: i disagree ... having multiple if blocks is easier to read, particularly given that I can comment next to each one. Streams are definitely harder to read and perform worse so why use them unless really needed.

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
tevans78: only one of them fails to start ... but I guess we could check for that first

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
tevans78: this servlet does run ... this is the one for empty beans.xml and I wanted to make sure that the warning only applied to a non-empty unversioned beans.xml

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
Azquelt: Right, but I'd suggest something like this to avoid so many nested blocks:

```suggestion
        for (WebSphereBeanDeploymentArchive bda : getWebSphereBeanDeploymentArchives()) {
            BeansXml beansXml = bda.getBeansXml();
            if (beansXml == null) { //check that there is a beans.xml file
                continue;
            }
            
            URL parsedURL = beansXml.getUrl();
            //if the URL is null then that means it was an empty beans.xml file
            //note that this may be an undocumented "feature" of the Weld SPI
            if (parsedURL == null) {
                continue;
            }
            
            //if the beans.xml was not an empty file then check if the version was set or not
            boolean unversionedBeansXml = beansXml.getVersion() == null;
            if (unversionedBeansXml) {
                URL unversionedBeansXmlURL = bda.getBeansXmlResourceURL();
                unversionedBeansXmlURLs.add(unversionedBeansXmlURL);
            }
        }
```

I'd also suggest than when it's laid out like this, the `parsedURL` and `unversionedBeansXml` variables are superfluous.

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
Azquelt: Huh, your comments suggest that both apps should fail to start.

Line 71:
```
        //an empty beans.xml file should not produce a warning even though this app will result in a DeploymentException
        WebArchive legacyEmptyBeansXMLWar = ShrinkWrap.create(WebArchive.class, LEGACY_EMPTY_BEANS_APP_NAME + ".war")
```

Line 79:
```
        //This legacyBeans10XMLWar war contains a non-empty CDI 1.0 beans.xml. When using CDI 4.0, the discovery mode will default to
        //ANNOTATED. Unlike an empty beans.xml, the legacy configuration option has no effect. Therefore we will output a warning when the
        //DeploymentException occurs
        WebArchive legacyBeans10XMLWar = ShrinkWrap.create(WebArchive.class, LEGACY_BEANS10_APP_NAME + ".war")
```

I understood that to mean that both apps should fail to start, but only one should output a warning when it does so.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
tevans78: ok

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
Azquelt: Hmm, maybe just the comments that need fixed then, the referenced servlet says:
```
        //this should never be called because app won't deploy
```

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
tevans78: fixed

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
tevans78: done

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
tevans78: The comments should be accurate now

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
tevans78: I removed that comment

tevans78:
#build

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
```diff
@@ -0,0 +1,107 @@
...
+    @Test
+    public void testUnversionedBeansXmlWarning() throws Exception {
+        List<String> warningMessages = server.findStringsInLogs("CWOWB1018W:.*");
+        assertTrue("Message CWOWB1018W not found", warningMessages.size() > 0);
+        assertEquals("Message CWOWB1018W was found more than once", 1, warningMessages.size());
+
+        List<String> errorMessages = server.findStringsInLogs("CWWKZ0002E:.*");
+        assertTrue("Message CWWKZ0002E not found", errorMessages.size() > 0);
+        assertEquals("Message CWWKZ0002E was found more than once", 1, errorMessages.size());
+    }
```
Azquelt: Please check for the application name in the error message, and add a comment that we're checking that the app failed to start.

dev/io.openliberty.cdi.4.0.internal.weld/src/io/openliberty/cdi40/internal/weld/CDI40ContainerEventManagerImpl.java
```diff
@@ -88,4 +99,17 @@ public boolean isEEModulesAware() {
         }
 
     }
+
+    /** {@inheritDoc} */
+    @Override
+    public DeploymentException processDeploymentException(WebSphereCDIDeployment webSphereCDIDeployment, DeploymentException e) {
+        //this extension point allows us to output an additional warning message when a DeploymentException is caught
+        if (TraceComponent.isAnyTracingEnabled() && tc.isWarningEnabled()) {
+            Collection<URL> unversionedBeansXmlURLs = webSphereCDIDeployment.getUnversionedBeansXmlURLs();
```
Azquelt: Do not check `isAnyTracingEnabled()` here. Warnings should still be emitted if tracing is disabled.

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTest.java
```diff
@@ -42,8 +42,8 @@
 public class LegacyConfigTest extends FATServletClient {
     public static final String SERVER_NAME = "CDI40Server";
 
-    public static final String LEGACY_EMPTY_BEANS_APP_NAME = "LagacyEmptyBeans";
-    public static final String LEGACY_BEANS10_APP_NAME = "LagacyBeans10";
+    public static final String LEGACY_EMPTY_BEANS_APP_NAME = "LegacyEmptyBeans";
+    public static final String LEGACY_BEANS10_APP_NAME = "LegacyBeans10";
```
Azquelt, line 52: Sort of unrelated to this change but further down in this file on line 84, you start the server with validation disabled, even though you expect all the deployed applications to start correctly.

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -99,3 +99,6 @@ enableImplicitBeanArchives.conflict.CWOWB1017W=CWOWB1017W: The enableImplicitBea
 enableImplicitBeanArchives.conflict.CWOWB1017W.explanation=Although you can set the enableImplicitBeanArchives attribute on both of the cdi12 and cdi configuration element configuration elements, you should not set this attribute with different values in the same server configuration.
 enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml to fix the conflict. The recommended action is to remove the cdi12 configuration element because it is superseded by the cdi configuration element.
 
+unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file may lead to unexpected behaviour when using CDI 4.0 or higher: {0}
+unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive should be treated as an explicit bean archive (all classes may be beans). In CDI 4.0 or higher, the archive is taken to be an implicit bean archive (only annotated classes may be beans).
+unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application.
```
Emily-Jiang:
```suggestion
unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file with the correct version and the bean-discovery-mode attribute propertly set within your application.
```

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/CDIContainerEventManager.java
```diff
@@ -18,16 +20,37 @@
...
     public Environment getEnvironment();
+
+    /**
+     * Not sure if this is the correct place for this method but don't really want to create yet another separate service.
+     *
+     * This extension point allows us to perform additional operations when a DeploymentException is caught. In most cases
+     * it is expected that the implementation will do something and then just return that same exception... but it could be
+     * a modified instance of the exception
+     */
+    public DeploymentException processDeploymentException(WebSphereCDIDeployment deployment, DeploymentException e);
```
Emily-Jiang, line 55:
```suggestion
    public DeploymentException addDignosticHelp(WebSphereCDIDeployment deployment, DeploymentException e);
```

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
tevans78: The problem here is that you will probably also need to update the namespaces and schema... which might also then lead to other changes being required?

tevans78: Updated to mention changing the schema as well (since that implies the namespace change too).

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTestError.java
tevans78: done

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTest.java
tevans78: I consolidated the tests, hopefully they make more sense now

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -99,3 +99,6 @@ enableImplicitBeanArchives.conflict.CWOWB1017W=CWOWB1017W: The enableImplicitBea
 enableImplicitBeanArchives.conflict.CWOWB1017W.explanation=Although you can set the enableImplicitBeanArchives attribute on both of the cdi12 and cdi configuration element configuration elements, you should not set this attribute with different values in the same server configuration.
 enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml to fix the conflict. The recommended action is to remove the cdi12 configuration element because it is superseded by the cdi configuration element.
 
+unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file may lead to unexpected behaviour when using CDI 4.0 or higher: {0}
```
tevans78: @OpenLiberty/message-reviewer please review this new message

tevans78: plus the explanation and user action as well, thank you

tevans78:
#libby

tevans78:
#build

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
helyarp: Use "might" instead of "may". It is preferable to use "later" instead of "higher" ("CDI 4.0 or later").

skasund:
L2 message review completed

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTest.java
```diff
@@ -33,60 +39,70 @@
...
-        cdi.setEmptyBeansXMLExplicitBeanArchive(true);
-        cdis.add(cdi);
+        server.startServer();
+    }
 
-        server.updateServerConfiguration(config);
+    @Test
+    public void testUnversionedBeansXmlWarning() throws Exception {
+        //check for a warning message about a non-empty unversioned beans.xml file
+        List<String> warningMessages = server.findStringsInLogs("CWOWB1018W: .*[wsjar:file:.*" + LEGACY_BEANS10_APP_NAME + ".war!/WEB-INF/beans.xml]");
```
Azquelt: `findStringsInLogs` takes a regular expression so you need to escape the `[` and `]` characters (or just remove them)

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTest.java
```diff
@@ -33,60 +39,70 @@
...
+        assertTrue("Message CWOWB1018W not found", warningMessages.size() > 0);
+        assertEquals("Message CWOWB1018W was found more than once", 1, warningMessages.size());
 
-        server.startServer(true, false);
+        //check for the DeploymentException
+        List<String> errorMessages = server.findStringsInLogs("CWWKZ0002E:.*" + LEGACY_BEANS10_APP_NAME + ".*DeploymentException");
+        assertTrue("Message CWWKZ0002E not found", errorMessages.size() > 0);
+        assertEquals("Message CWWKZ0002E was found more than once", 1, errorMessages.size());
```
Azquelt, line 100: I think this will work, but I'm not sure if `LibertyServer` strictly guarantees that the application will have failed to start before `startServer()` returns.

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/beansxml/AllBeansServlet.java
```diff
@@ -29,14 +27,11 @@ public class AllBeansServlet extends FATServlet {
...
+    private UnannotatedBean unannotatedBean;
 
     @Test
     public void testAllBeansInjected() {
         assertNotNull(requestScopedBean);
         assertEquals("RequestScopedBean", requestScopedBean.test());
-
-        assertTrue(unannotatedBean.isResolvable());
-        assertEquals("UnannotatedBean", unannotatedBean.get().test());
     }
```
Azquelt: Should you be calling the `unannotatedBean` here? I don't expect it to ever fail if the app successfully starts, but it would be consistent with the test for `requestScopedBean`.

dev/io.openliberty.cdi.4.0.internal.weld/src/io/openliberty/cdi40/internal/weld/CDI40ContainerEventManagerImpl.java
tevans78: done

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTest.java
tevans78: It waits for the "ready to run a smarter planet" message which will always be output after CDI has processed the applications... which is what gives the DeploymentException

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
tevans78: done

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/LegacyConfigTest.java
tevans78: done

dev/io.openliberty.cdi.4.0.internal_fat/fat/src/io/openliberty/cdi40/internal/fat/config/beansxml/AllBeansServlet.java
tevans78: done

tevans78:
#build

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -99,3 +99,6 @@ enableImplicitBeanArchives.conflict.CWOWB1017W=CWOWB1017W: The enableImplicitBea
 enableImplicitBeanArchives.conflict.CWOWB1017W.explanation=Although you can set the enableImplicitBeanArchives attribute on both of the cdi12 and cdi configuration element configuration elements, you should not set this attribute with different values in the same server configuration.
 enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml to fix the conflict. The recommended action is to remove the cdi12 configuration element because it is superseded by the cdi configuration element.
 
+unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
+unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive should be treated as an explicit bean archive (all classes may be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes may be beans).
```
helyarp: Changes to remove "should" and "may":

- "should be treated" to "was treated"
- "all classes may be beans" to "all classes might be beans" ("may" to "might")
- "only annotated classes may be beans" to "only annotated classes can be beans"

helyarp (commented):
Suggested changes

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
tevans78: @helyarp "all classes might be beans" really doesn't sound right to me. I would rather phrase it as "all classes are eligible to be beans".

helyarp: "all classes are eligible to be beans" is okay. A bit of a mouthful but accurate.

Azquelt (approved):

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
tevans78: done

tevans78:
#build

tevans78:
#libby

PR #21374, opened 2022-06-07
EJBNPTimerTest: change to interval timer so timer doesn't expire and cancel before test is done

olendvcook:
The test doesn't really care what type of timer it is, or if it expires or not.

Also the test structure already handles cancelling the timers when the test is done for all tests in the bucket by calling clearAllTimers() in the AbstractServlet.doGet() method.


olendvcook:
#build

tkburroughs (approved):

PR #21375, opened 2022-06-07
Fix JCA test suite derby thread interruption

KyleAure:
Working around the situation where Derby could put an interrupt status on the resource adapter thread during deactivation.


njr-11 (approved):

KyleAure:
#build

KyleAure:
Child build failed due to a file system issue. Re-built child and it passed.

PR #21405, opened 2022-06-09
Retry checking for FFDC if ExpectedFFDC was not found

kristip17:
Sometimes the FATRunner checks for FFDC and we don't find the ExpectedFFDC right away -- seems to be a timing or flush problem. Added 3 retries with a very short sleep to start, if the ExpectedFFDCs aren't found.

In the test I debugged, a test had an ExpectedFFDC and then test cleanup didn't find it, but milliseconds later, the next test picked it up in the pre-test FFDC check.

I manually tested this by adding an ExpectedFFDC that wouldn't occur and check that we retry and fail. Then I checked on a test with valid ExpectedFFDCs that we only did one check. Then I added some temp code to force us to do 1 retry, then find the ffdc and exit the retry loop.

RTC 272561

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
+++ b/dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
@@ -198,19 +198,48 @@ public class FATRunner extends BlockJUnit4ClassRunner {
 
                     superStatement.evaluate();
 
-                    // If we got to here without error, do a final check that
-                    // any FFDCs were expected
-                    Map<String, FFDCInfo> ffdcAfterTest = retrieveFFDCCounts();
-                    Map<String, FFDCInfo> unexpectedFFDCs = filterOutPreexistingFFDCs(ffdcBeforeTest, ffdcAfterTest);
+                    int retryCount = 3;
 
+                    Map<String, FFDCInfo> unexpectedFFDCs = null;
                     ArrayList<String> errors = new ArrayList<String>();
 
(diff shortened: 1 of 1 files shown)


kristip17:
#build

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -198,20 +198,34 @@ public void evaluate() throws Throwable {
...
+                     * the test expecting the FFDC or it bleeds over into another test. Attempting to mitigate with a short
+                     * retry.
+                     */
+                    for (int i = 1; i <= retryCount; i++) {
+                        // If we got to here without error, do a final check that
+                        // any FFDCs were expected
+                        Map<String, FFDCInfo> ffdcAfterTest = retrieveFFDCCounts();
+                        unexpectedFFDCs = filterOutPreexistingFFDCs(ffdcBeforeTest, ffdcAfterTest);
+
+                        List<String> expectedFFDCs = getExpectedFFDCAnnotationFromTest(method);
```
tevans78: no point in retrieving the annotations inside the for loop, those won't change

kristip17: Moved out of the retry loop.

kristip17:
#build

kristip17:
There was one failed test which looks like known issue RTC 290765.

tevans78 (approved):

kristip17:
#libby

PR #21410, opened 2022-06-09
Issue 21376: Remove access and refresh tokens from the OP cache

### LINKED ISSUE #21376: Remove access tokens from OP cache after back-channel logout
ID tokens, access tokens, and refresh tokens should probably be removed from the OP's cache after a user performs back-channel logout. This will ensure those tokens cannot be reused.

ayoho:
Removes any ID, access, and refresh tokens that are associated with the user and RP that is being logged out as part of OIDC back-channel logout. All of the user's cached tokens are examined to find ID and access tokens associated with the user and client, and then any refresh token that is associated with each of those tokens is also removed from the cache.

Resolves #21376


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.oidc.social_fat.backchannelLogout

teddyjtorres (approved):

Issue #21235, opened 2022-05-24
Checkpoint - MicroProfile fault tolerance @Timeout annotation


tjwatson:
The `@Timeout` annotation can be used on a method to specify a specific timeout.  If there is user application code running during a `checkpoint` operation (with `--at=applications`) then it may be possible that a thread is in the middle of calling a method annotated with `@Timeout`.  If this occurs then it seems very likely the timeout will pass before the process resumes on restore.  What should happen for this scenario?  Some possibilities:

1. Fail at checkpoint if we detect any threads in the middle of a method call annotationed with `@Timeout`
2. Just let the method pause during the checkpoint.  On restore the timeout may have passed causing an immediate `TimeoutException` to occur.
3. Depending on how the timeout is calculated it may omit the delta of time "missed" between the checkpoint and restore operation.  For example, if `System.nanoTime` is used then the JVM is going to omit the "missed" time when `System.nanoTime` is called again.  On the other hand, if this is placed on some ScheduledExecutor that runs at the specified timeout then we need to see that the executor behaves correctly.

For simplicity I think it makes some sense to fail checkpoint if a method is in the middle of an `@Timeout` method.

Azquelt:
The job of triggering the timeout is delegated to a scheduled executor: https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/state/impl/TimeoutStateImpl.java#L51

The scheduled task is a call to a method on the same class, so we would be able detect whether a checkpoint has happened and act accordingly, if the scheduled executor does run the task immediately after resuming the checkpoint.

I note that we also log a metric of how long the task took using `System.nanoTime`, I guess we might want to discard that.

tjwatson:
> The job of triggering the timeout is delegated to a scheduled executor: https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.microprofile.faulttolerance.2.0/src/com/ibm/ws/microprofile/faulttolerance20/state/impl/TimeoutStateImpl.java#L51

Open J9 will be fixing the implementations of `java.util.concurrent.ScheduledExecutorService` that the JVM provides to correctly handle the "gap" time between checkpoint and restore such that the delays and periods behave appropriately.  For example, if you are into 1 minute of a 2 minute period and a checkpoint is done the on restore it will correctly determine that 1 minute still remains in the period.

[...]

PR #21422, opened 2022-06-13
Added a checkpoint with method timeout annotation test to checkpoint_…

KeenanMcCarthy:
…fat FATSuite

Created a test-case to check the method timeout behavior when a timeout is invoked during a checkpoint. When the checkpoint command is run, the server will spawn a thread which starts a timeout. When the server is restored, the timeout method should resume ignoring time passed in which the server was inactive.

### DIFF
--- a/dev/io.openliberty.checkpoint_fat/test-applications/timeoutTest/src/timeoutTest/TimeoutService.java
+++ b/dev/io.openliberty.checkpoint_fat/test-applications/timeoutTest/src/timeoutTest/TimeoutService.java
@@ -1,0 +1,61 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package timeoutTest;
+
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.faulttolerance.Asynchronous;
+import org.eclipse.microprofile.faulttolerance.Timeout;
+import org.eclipse.microprofile.faulttolerance.exceptions.TimeoutException;
+
+@ApplicationScoped
+public class TimeoutService {
+
+    @Inject
+    @ConfigProperty(name = "timeout")
+    private Boolean createTimeout;
+
+    @Asynchronous
+    public Future<Void> spawnTimeoutThread() {
+        Future<Void> completable = CompletableFuture.runAsync(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    System.out.println(produceTimeout());
+                } catch (TimeoutException te) {
+                    System.out.println("TIMEOUT OCCURED");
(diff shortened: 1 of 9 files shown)


dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/TimeoutTest.java
```diff
@@ -0,0 +1,64 @@
...
+    public void testCheckpointStartTimeout() throws Exception {
+        server.setCheckpoint(CheckpointPhase.APPLICATIONS, false, server -> {
+            String result = server.waitForStringInLogUsingMark("TIMEOUT CALLED");
+            assertNotNull("Timeout call not found at checkpoint", result);
+        });
+        server.startServer();
+    }
+
+    @Test
+    public void testCheckpointTimeout() throws Exception {
```
tjwatson: Order of tests will not be predictable, so it is possible `testCheckpointTimeout` will be called before `testCheckpointStartTimeout`.  I think you may need to place this in a single `@Test` method.

Also, I think to test this properly you need to have a delay between the checkpoint and the restore that is greater than the `@Timeout` annotations timeout value.

dev/io.openliberty.checkpoint_fat/test-applications/timeoutTest/src/timeoutTest/TimeoutService.java
```diff
@@ -0,0 +1,51 @@
...
+                    System.out.println(e.getClass().getName());
+                }
+            }
+        });
+        return completable;
+    }
+
+    @Timeout(value = 30, unit = ChronoUnit.SECONDS)
+    public String produceTimeout() throws TimeoutException, InterruptedException {
+        Thread.sleep(Long.MAX_VALUE);
```
tjwatson: So this tests that a timeout will happen on restore.  We probably need another test that ensures a timeout does not happen on restore.

For example, have the `@Timeout` annotation specify 30 seconds to timeout, but only sleep for 10 seconds.  Then make sure the method continues on the restore side without timing out (but make sure you wait to restore from checkpoint for a period that is greater than 30 seconds.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/TimeoutTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+import componenttest.annotation.Server;
+import componenttest.annotation.SkipIfCheckpointNotSupported;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import io.openliberty.checkpoint.spi.CheckpointPhase;
+
+@RunWith(FATRunner.class)
+@SkipIfCheckpointNotSupported
+public class TimeoutTest {
```
tjwatson: Please name this MPFaultToleranceTimoutTest

dev/io.openliberty.checkpoint_fat/test-applications/timeoutTest/src/timeoutTest/TimeoutStartup.java
```diff
@@ -0,0 +1,31 @@
...
+@Singleton
+public class TimeoutStartup {
+
+    @Inject
+    TimeoutService timeoutTester;
+
+    @PostConstruct
+    public void getProperties() {
+        System.out.println("TIMEOUT CALLED");
+        timeoutTester.sqawnTimeoutThread();
```
tjwatson: I suggested a second testcase to add.  It would be nice not to have to add a new application to that.  So to do that you could check a system property to tell you which timeoutTester method to call.  One that must timeout and one that must not timeout according to the testcase running.

sebratton (approved):
Looks good to me.

KeenanMcCarthy:
#build
#libby

tjwatson:
For #21235

PR #21492, opened 2022-06-20
PR for Issue 21447: Extract user name from session security name in BCL

### LINKED ISSUE #21447: Backchannel logout not getting called when id_token_hint is not passed
I have a test that accesses a protected app using an openidconnectclient and openidconnect server.
The openidconnect provider uses a client defined in a local_store.  That client specifies a backchannelLogoutUri.
After accessing the app, I invoke end_session on the openidconnect provider and do **NOT** provide the **id_token_hint** on the request.
end_session is invoked, but the "userName" that is passed to sendBackchannelLogoutRequests is "user:BasicRealm/testuser".  When we try to match that user to an entry in the tokenCache, we don't find any matches.  Using "testuser" as the userName would match and would result in the back channel logout uri's for that user being invoked.
We need to investigate how we're obtaining the userName.

The call to request.logout(); (in processEndSesstion)  will end up in logoutUnprotectedResourceServiceRef in AuthenticateApi.  This will use getSession(req, res) to get the username which will return "user:BasicRealm/testuser".
That will then be passed to logout in BackChannelLogoutServices and then to sendBackchannelLogoutRequests who needs just the actual user "testuser" (in this case).
We probably need to update sendBackchannelLogoutRequests to transform that string as to not break the caller that may use the userName value to invoke other services.

[noIdTokenHint2.zip](https://github.com/OpenLiberty/open-liberty/files/8914069/noIdTokenHint2.zip)

ayoho:
Before, the OIDC `BackchannelLogoutService` class expected the user name it received to be the correct string as-is. However the user name that is returned from the `IHttpSessionContext` can take the format of an access ID. We therefore need to extrac the user name from that access ID if the input string is in that format.

Resolves #21447

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.server/src/io/openliberty/security/openidconnect/backchannellogout/BackchannelLogoutService.java
+++ b/dev/com.ibm.ws.security.openidconnect.server/src/io/openliberty/security/openidconnect/backchannellogout/BackchannelLogoutService.java
@@ -11,6 +11,8 @@
 package io.openliberty.security.openidconnect.backchannellogout;
 
 import java.util.Iterator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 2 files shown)


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.oidc.social_fat.backchannelLogout

teddyjtorres (approved):

PR #21494, opened 2022-06-20
Fix retry conditions for UserEnumTest

kristip17:
Previously added a retry for the UserEnumerationTest, but didn't use the right condition to retry on the "all invalid" option and our Windows boxes can be extra slow/variable.

RTC 291045


kristip17:
#build

ryanesch (approved):

PR #21508, opened 2022-06-22
Enable trace if task threads fail to stop

KyleAure:
During the server quiesce period for this test, we have observed a warning being produced that states that 6 threads did not complete.

```
CWWKE1107W: 6 threads did not complete during the quiesce period.
CWWKE1102W: The quiesce operation did not complete. The server will now stop. 
```

We were able to verify that these are referring to user threads, and not threads related to the persistent executor quiesce listeners.
Adding this trace will help us further debug this issue if it happens again.


KyleAure:
#build

tabishop (approved):

PR #21509, opened 2022-06-22
AuthenticateApi logout fall back to get user name from caller subject if session user is anonymous

ayoho:
The user name "anonymous" is returned from the HTTP session when session security is disabled. In that case, the `logoutUnprotectedResourceServiceRef()` method should fall back to getting the user name from the caller subject just as if the user name was `null`.


ayoho:
#build

teddyjtorres (approved):

PR #21514, opened 2022-06-22
samlWeb feature update

arunavemulapalli:
Since we are almost done updating jaxws-2.2 , wsSecurity and wsSecuritySaml features (using newer cxf/wss4j levels, we will not need samlWeb feature with older levels of opensaml (v2))

Update samlWeb feature to transition using opensaml v3.


wraschke:
#libby

arunavemulapalli:
I have to do this update in stages.
I will go ahead open separate PRs

Issue #21342, opened 2022-06-03
WebProfile ssl with checkpoint


tjwatson:
For checkpoint the ssl configuration needs to be able to come in on the restore side.  In particular we need to make sure that in-container tls configuration works as described at https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/

See checklist to help evaluate work needed for checkpoint: https://github.com/OpenLiberty/open-liberty/issues/18664#issuecomment-1146218456

tjwatson:
@anjumfatima90 I assigned to you for now because you already added a testcase for SSL and have some investigations on how we can get the SSL initialization to happen on the restore side.  This has an added complication that we need to re-check ssl configuration on the restore side before bringing up the SSL stack in liberty.

dettmoney:
https://github.com/OpenLiberty/open-liberty/pull/21210 is a performance prototype that reduces restore response times. It involves  pre-initializing the default SSLSocketFactory. There's a chance that will not be possible in light of this work to move ssl configuration to the restore side.

PR #21519, opened 2022-06-22
Move ltpa processing from checkpoint to restore time

### LINKED ISSUE #21518: Checkpoint fails for acmeair applications due to LTPA configuration error
During checkpoint of acmeair microservice applications, the following error is seen:

**[ERROR   ] CWWKS4106E: LTPA configuration error. Unable to create or read LTPA key file: /criu/20220620/wlp/usr/servers/AcmeAuth/resources/security/ltpa.keys**

LTPA processing should likely be moved to restore time to account for differences in environments between checkpoint and restore times.

dettmoney:
When trying to checkpoint acmeair microservice applications, I see the following error:
**[ERROR   ] CWWKS4106E: LTPA configuration error. Unable to create or read LTPA key file: /criu/20220620/wlp/usr/servers/AcmeAuth/resources/security/ltpa.keys**

Moving the call to  `com/ibm/ws/security/token/ltpa/internal/LTPAConfigurationImpl.setupRuntimeLTPAInfrastructure()` to restore time avoids the error with no effect on performance. We need to make sure we don't run into any issues with this fix where we are improperly caching anything that happens at checkpoint time because the environment may be different during restore.

This fix for this particular problem may be folded into the ssl checkpoint issue (https://github.com/OpenLiberty/open-liberty/issues/21342).

### DIFF
--- a/dev/com.ibm.ws.security.token.ltpa/bnd.bnd
+++ b/dev/com.ibm.ws.security.token.ltpa/bnd.bnd
@@ -28,6 +28,10 @@ Export-Package: com.ibm.ws.security.token.ltpa
 Private-Package: \
   com.ibm.ws.security.token.ltpa.internal.*
 
+-ds-felix-extensions=true
+-dsannotations: \
+  com.ibm.ws.security.token.ltpa.internal.LTPAConfigurationImpl
+
 # Update the metatype.xml for your component in the resources/OSGI-INF/metatype folder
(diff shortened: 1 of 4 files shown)


tjwatson:
@dettmoney can this PR be closed?  I think commit 732485675b2dff24914033f7ca68675b1e5a4908 addresses this?

PR #21545, opened 2022-06-24
Refactor cloudant tests to use couchDB

KyleAure:
This is a test to see if we can switch from using a Cloudant image to CouchDB. 
This is for serviceability since the Cloudant image has been deprecated.


KyleAure:
#build

KyleAure:
#build

KyleAure:
#build

PR #21549, opened 2022-06-24
Update servlet60 fat trace and build path

### LINKED ISSUE #21548: Update io.openliberty.webcontainer.servlet.6.0.internal_fat to build with Servlet 6.0
In addition update the server.xml that is used currently to specify the full trace spec for servlet-6.0 so that we have that in a server easily accessible if we need it.

pnicolucci:
fixes #21548


pnicolucci:
#libby #build

pnicolucci:
#libby #build

volosied (approved):
Looks fine to me

PR #21555, opened 2022-06-24
asynchronous repository methods

njr-11:
The pull has some in-depth experimentation with asynchronous repository methods.
First, I figured out a way to get Concurrency's asynchronous annotation to be transferred to the bean that backs the mock repository method, after which all the repository method needs to do is declare that it returns CompletionStage or CompletableFuture as the Concurrency spec requires for asynchronous.  If a database has its own async programming model, then Concurrency Asynchronous won't even be needed an an incomplete stage can be returned instead. This helps fill in the gap when that option isn't available.
With the ability to return completion stages, we needed a way to identify at run time whether the completion stage represents a list or single result.  Reusing the value from Limit annotation worked well for that.
Finally, one of the things I've been wanting to experiment with is to combine with processing/reduction of pages inline so that there is no need to have all of the results read into memory at once.  Streams aren't an option because the entire stream needs to exist first.  However, the Collector mechanism from streams doesn't have this requirement.  I added the ability to specify a Collector parameter that would process pages as they are read in, with a completion stage returned to the application to await or chain dependent stages to the final result.

### DIFF
--- a/dev/io.openliberty.data.internal_fat/bnd.bnd
+++ b/dev/io.openliberty.data.internal_fat/bnd.bnd
@@ -40,6 +40,7 @@ fat.project: true
   com.ibm.ws.tx.embeddable.jakarta;version=latest,\
   io.openliberty.jakarta.annotation.2.0,\
   io.openliberty.jakarta.cdi.3.0,\
+  io.openliberty.jakarta.concurrency.3.0,\
   io.openliberty.jakarta.interceptor.2.0,\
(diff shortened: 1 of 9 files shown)


njr-11:
#build

njr-11:
rebased onto latest
#build
#libby

njr-11:
#build

KyleAure (approved):

PR #21565, opened 2022-06-27
Concurrency skip signature testing on Java 19

KyleAure:
The concurrency TCK does not support running the signature test on any JDK besides 11 and 17.  
Therefore, we should skip this test on all other JDKs levels


gjwatts (approved):

KyleAure:
#build

PR #21571, opened 2022-06-28
Update CDI Config Attribute

### LINKED ISSUE #21019: Review the name of the CDI emptyBeansXMLExplicitBeanArchive config attribute
As part of CDI 4.0 and new attribute is needed to allow backwards compatibility when an empty beans.xml is used.

In previous versions of CDI, an empty beans.xml file meant that the archive was an Explicit Bean Archive ... the equivalent of `bean-discovery-mode=ALL` ... all classes in the archive are eligible to be CDI Beans

In CDI 4.0, an empty beans.xml file means that the archive is an Implicit Bean Archive ... the equivalent of `bean-discovery-mode=ANNOTATED` ... only classes that are annotated with bean defining annotations may be CDI Beans.

The initial config attribute chosen was `emptyBeansXMLExplicitBeanArchive=true|false`. This attribute is a verbose description of what the spec requires but may not be clear enough outside of that context. There was also some concern about the length of the attribute name.

tevans78:
fixes https://github.com/OpenLiberty/open-liberty/issues/21019
#build


dev/com.ibm.ws.cdi.weld/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -22,8 +22,8 @@ cdi12.enableImplicitBeanArchives.desc=The implicit bean archives are scanned for
...
-cdi.enableImplicitBeanArchives.desc=The implicit bean archives are scanned for any bean discoveries. The default value is true.
+cdi.enableImplicitBeanArchives.name=Enable scanning implicit bean archives.
+cdi.enableImplicitBeanArchives.desc=The implicit bean archives are scanned for any bean discoveries. The default value is true. Only applies to CDI 1.2 or newer.
 
-cdi.emptyBeansXMLExplicitBeanArchive.name=An archive with an empty beans.xml file is an explicit bean archive. This property applies to CDI 4.0 or newer.
-cdi.emptyBeansXMLExplicitBeanArchive.desc=CDI 4.0 and later considers an archive with an empty beans.xml file to be an implicit bean archive. This property maintains the existing behavior before to CDI 4.0. The default value is false.
+cdi.emptyBeansXmlCDI3Compatibility.name=CDI 3.0 compatibility for empty beans.xml files.
+cdi.emptyBeansXmlCDI3Compatibility.desc=CDI 4.0 and later consider an archive with an empty beans.xml file to be an implicit bean archive (bean-discovery-mode=ANNOTATED). This property allows an empty beans.xml file to be treated as it was in CDI 3.0 (bean-discovery-mode=ALL). The default value is false. Only applies to CDI 4.0 or newer.
```
tevans78, line 29: https://github.com/orgs/OpenLiberty/teams/message-reviewer As part of changing the attribute name we also changed these NLS properties because we realised that the main use of "name" was actually in the WDT tooling and should be rather shorter than it was before.

Azquelt, line 29: I don't think you want a full stop at the end of the names.

Azquelt (approved):

Azquelt:
We noted that the behaviour of the WDT tooling when you have an optional boolean property with no default in the metatype isn't brilliant, but I'm not sure how else we can transition from the old `<cdi12>` element to the new one.

helyarp (commented):
Looks good.

posmith:

PR #21574, opened 2022-06-28
Ensure JSONB property annotation collision is resolved

### LINKED ISSUE #13640: JSON-B transient and property annotation collide
When mapping a class via JSON-B, and annotating fields as follows, the result throws an error because of mismatching types:

```java
public class Coffee {

    @JsonbTransient
    private Order order;

    @JsonbProperty("order")
    private URI orderLink;

    Coffee() {
        order = new Order();
        orderLink = URI.create("http://localhost/");
    }

    // getters & setters
}
```

this results in: 

```
[ERROR   ] Unable to serialize property 'order' from com.sebastian_daschner.openliberty.Coffee
[ERROR   ] Generating incomplete JSON
[ERROR   ] Problem with writing the data, class com.sebastian_daschner.openliberty.Coffee, ContentType: application/json
```

If both properties `order` and `orderLink` have the same type, it works as expected (i.e. `orderLink` is serialized as `"order"`).

However, if we annotate `@JsonbProperty` on the setter instead, it works:

```java
public class Coffee {

    @JsonbTransient
    private Order order;

    private URI orderLink;

    // [...]

    @JsonbProperty("order")
    public URI getOrderLink() {
        return orderLink;
    }

}
```

Response:

```json
{"order":"http://localhost/"}
```

There seems to be an unexpected type link of the (ignored) property with the default serialized JSON key `"order"`, and the custom name binding.

Tested with `Open Liberty 20.0.0.8/wlp-1.0.43.cl200820200721-1900 on Eclipse OpenJ9 VM, version 14+36`

KyleAure:
Fixes #13640

### DIFF
--- a/dev/io.openliberty.jakarta.jsonb.internal_fat/test-applications/jsonbtestapp/src/test/jsonb/web/JsonBTestServlet.java
+++ b/dev/io.openliberty.jakarta.jsonb.internal_fat/test-applications/jsonbtestapp/src/test/jsonb/web/JsonBTestServlet.java
@@ -13,10 +13,12 @@ package test.jsonb.web;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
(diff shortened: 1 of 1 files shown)


tabishop (approved):

KyleAure:
#build

PR #21580, opened 2022-06-28
Remove hidden character in Bells FAT

tkburroughs:
Method name in a FAT class has a hidden character at the end which Eclipse
interprets as part of the method name and thus does not recognize as an
override of a method from parent.  Removed the hidden character.


dazavala (approved):

jhanders34 (approved):

tkburroughs:
#build

dazavala:
Thank you, @tkburroughs .

PR #21582, opened 2022-06-28
moved date in oldStaticFile.xml way into the future

### LINKED ISSUE #21581: oldStaticFile.xml is no longer valid
oldStaticFile.xml has the purpose of `iFix with a static file who's date is a long way into the future so the file will be older and therefore not valid`

However, the date is no longer way into the future.. the date was 2022-06-27 which has passed. Since then this file causes tests to fail.

jacobwdv:
Fixes #21581 

changing the date of the file  from year 2022 to 2092.


jacobwdv:
#build

bensonlatibm (approved):
lgtm!

wraschke:
#libby

wraschke:
I'm overriding the required copyright header status check and merging.

PR #21585, opened 2022-06-28
During a checkpoint enable SSL only after restore - Add a condition which gets registered only when process is running with no checkpoint or if the process has been restored from a checkpoint image.

anjumfatima90:
for issue #21342


tjwatson (approved):
Approve, but please update the commit message to something like this:


During a checkpoint enable SSL only after restore

This change adds a mandatory requirement to KeystoreConfigurationFactory and SSLComponent on the running process condition.  This condition is registered ASAP in the boot sequence when not doing a checkpoint.  That allows the components to be enabled just as before for normal launches.  For checkpoint launches this condition is not registered until after the server process has been restored.  This delays the components from be enabled until after restore.  This avoids calling JVM security related code on the checkpoint side.

tjwatson (approved):

anjumfatima90:
#build

dettmoney (approved):

anjumfatima90:
#build

anjumfatima90:
#libby

PR #21586, opened 2022-06-28
asynchronous with consumer

njr-11:
Covers the JNoSQL pattern of being able to supply a Consumer parameter which accepts each result individually.


njr-11:
#build

KyleAure (approved):

PR #21594, opened 2022-06-29
Fix SpnegoFat - JwtSsoFeature test

Zech-Hein:
On rare occasion `testSpnegoSuccessful_withJwtSsoFeature` would get ran as the first test and would fail because the SPNEGO config was not set. The test assumed SPNEGO config was already set from other test cases being run first normally.

Fix is simple to call `setDefaultSpnegoServerConfig();` at the beginning of the test to ensure the SPNEGO config has been set in the Liberty server config


Zech-Hein:
#build

utle (approved):

PR #21639, opened 2022-07-05
Refresh of Keystores should also include Hybrid Keystores

### LINKED ISSUE #20888: Refresh of certificate not working for keystores with Prefix safkeyringhybrid in WAS zOS
**Describe the bug**  
Inside the server.xml the following configuration is set:

``` sh
<keyStore id="my_racfTrustStore_1"        
  filebased="false"            
  location="safkeyringhybrid:///mykeystore"
  password="XXXX"           
  readOnly="true"             
  type="JCEHYBRIDRACFKS"/>  
```

Whenever a refresh command for keyrings with prefix `safkeyringhybrid://` or `safkeyringhw://`, runs

The following error is seen in the logs:

`CWWKB0004I: Unable to find the Keystore racfkeystore` 

However, this issue does **NOT** happen with keystores that have the following prefix
`safkeyring://`

``` sh
<keyStore id="my_racfTrustStore_2"        
  filebased="false"            
  location="safkeyring:///mykeystore2"
  password="XXXXX"           
  readOnly="true"             
  type="JCERACFKS"/>  
```

**Steps to Reproduce**  

Run the following command:
`MODIFY <server>,REFRESH,KEYSTORE`

If using zOS Connect the issue can be recreated if you run the following command:

`F  baqstrt,REFRESH,KEYSTORE,ID=racfTrustStore`

**Expected behavior**  
The refresh command should be able to handle keystores that have prefixes `safkeyringhybrid://` and `safkeyringhw://`

**Diagnostic information:**  
 - OpenLiberty Version: 21.0.0.12-22.0.0.3

wrodrig:
When updating the keystores we should also refresh those that are hybrid on zOS.


wrodrig:
The issue will be addressed in a different Pull request

PR #21642, opened 2022-07-05
Refresh of certificate not working for keystores formats

wrodrig:


wrodrig:
The issue will be addressed in a different Pull request

PR #21646, opened 2022-07-05
initial experimentation with Template

njr-11:
Copy Template from JNoSQL and get it to inject. To start with, I've only mocked a couple of the methods for insert.  Experimentation with time-to-live will be done separately.


njr-11:
#build

dev/io.openliberty.data.internal_fat/fat/src/test/jakarta/data/TemplateTest.java
```diff
@@ -0,0 +1,49 @@
...
+import componenttest.annotation.Server;
+import componenttest.annotation.TestServlet;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+import test.jakarta.data.template.web.TemplateTestServlet;
+import test.jakarta.data.web.DataTestServlet;
+
+@RunWith(FATRunner.class)
+@MinimumJavaLevel(javaLevel = 11) // TODO 17
```
KyleAure, line 30: Are we thinking that Jakarta Data will be written for a minimum of Java 17?

KyleAure (approved):

dev/io.openliberty.data.internal_fat/fat/src/test/jakarta/data/TemplateTest.java
njr-11: Regarding the spec (which hasn't been written yet), it's uncertain if it would require Java 17.
The comment here is about the minimum level for the test case, which I'd like to have cover Java Records once we can run tests against Java 17.

PR #21652, opened 2022-07-06
Fix umask command for IBM i in server script

### LINKED ISSUE #21651: 290399-Fix umask command for IBM i in server script
**Describe the bug**  
Currently, the umask shell script command used in the `server` command on IBM i is biased for QSHELL (`/bin/qsh`).  It uses the extended symbols (namely `-S`) support that not all the other shells on IBM i support.  When running in some of the other shells, an error message message like:
```
umask: bad argument count
```
can show up.  This could mean the file permissions used with an Open Liberty server were incorrectly set.

We also see this error message show up in our FAT testing:
```
.../dev/image/output/wlp/bin/server[3]: umask: 0403-008 The number of parameters specified is not correct.
```

**Steps to Reproduce**  
Run any server command that attempts to set the file permissions via the umask shell command.

**Expected behavior**  
The umask permission should be properly set.  The user and group values should be maintained from their previous setting and the other value should be set to 7 (no access).

**Diagnostic information:**  
 - OpenLiberty Version: [all]

**Additional context**

gjwatts:
Fixes #21651

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -41,7 +41,7 @@
 #
 # CRIU_RESTORE_DISABLE_RECOVERY - 
 #              When restoring a server from a checkpoint it is possible
-#              that CRIU will fail to restore the proces.  By default
+#              that CRIU will fail to restore the process.  By default
 #              any failure to restore will result in the server
 #              recovering without using the checkpoint image.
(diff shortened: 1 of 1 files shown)


KyleAure (approved):

brenthdaniel (approved):

gjwatts:
#build

gjwatts:
Build completed successfully, but this change will not really be tested until it has run through an SOE to actually run on an IBM i server.  Merging.

PR #21674, opened 2022-07-06
Port jca fat regr tests to open liberty

tabishop:
Porting the com.ibm.ws.jca_fat_regr test bucket to Open Liberty. This is a massive test bucket and was broken up into 3 separate test buckets due to the long run time. The other 2 buckets are com.ibm.ws.jca_fat_regr.cfg.prop and com.ibm.ws.jca_fat_regr.rar.beanVal.

Rar, war, and app creation was converted from ant to java code using Shrinkwrap, feature versions updated for what is supported in open liberty and removed tests for feature versions not supported in open liberty.

Will squash commits after reviews are completed.


tabishop:
#build

tabishop:
#build

tabishop:
#build

tabishop:
#libby

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/JCAFATTest.java
```diff
@@ -0,0 +1,235 @@
...
+
+import com.ibm.websphere.simplicity.RemoteFile;
+import com.ibm.websphere.simplicity.log.Log;
+
+import componenttest.topology.impl.LibertyServer;
+
+/**
+ * Common infrastructure code for JCA REGR FAT tests
+ */
+public abstract class JCAFATTest {
```
KyleAure: Can this class extend `FATServletClient`? 
This will reduce some repeated code that could be stripped out.

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/JCAFATTest.java
```diff
@@ -0,0 +1,235 @@
...
+        public void succeeded(Description description) {
+            System.out.println(timestamp() + " <--- "
+                               + description.getMethodName() + " successful");
+        }
+
+        private final String timestamp() {
+            Date date = new Date(System.currentTimeMillis());
+            return '[' + date.toString() + ']';
+        }
+    };
```
KyleAure: Can be deleted if we extend `FATServletClient`

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/JCAFATTest.java
```diff
@@ -0,0 +1,235 @@
...
+                else
+                    throw x;
+            } finally {
+                Log.info(c, "runInJDBCFATServlet", "disconnecting from servlet");
+                if (runPasswordCheck)
+                    checkForTextPasswordsInLogs();
+                con.disconnect();
+            }
+        }
+    }
```
KyleAure, line 199: Can be deleted if we extend `FATServletClient`

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/app/InboundSecurityTest.java
```diff
@@ -0,0 +1,938 @@
...
+                              "J2CA0678W: Group testGrpNonExistent" // EXPECTED
+            );
+        } finally {
+            server.updateServerConfiguration(originalServerConfig);
+        }
+    }
+
+    @Test
+    public void testCallerIdentityPropagationFromCallerPrincipalCallbackStartWork() throws Exception {
+        String testName = "testCallerIdentityPropagationFromCallerPrincipalCallbackStartWork";
```
KyleAure: If `JCAFATTest` extends `FATServletClient` then a protected field called `testName` will automatically be created.  This is important to have because if we try to repeat these tests with Jakartaee10 then the method name will be changed to `testCallerIdentityPropagationFromCallerPrincipalCallbackStartWork_EE10` (or something like that).

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/app/InboundSecurityTest.java
```diff
@@ -0,0 +1,938 @@
...
+        Log.info(c, testName, "Executing " + testName);
```
KyleAure: Also, this is automatically done when using `FATServletClient`

dev/com.ibm.ws.jca_fat_regr.cfg.prop/fat/src/suite/r80/base/jca16/TestSetupUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Class <CODE>TestSetupUtils</CODE> provides utility methods used
+ * to help set up resources needed for tests
+ */
+public class TestSetupUtils {
+
+    public static final String raDir = "test-resourceadapters/adapter.regr/resources/META-INF/";
+
+    private static JavaArchive resourceAdapterEjs_jar;
```
KyleAure, line 26: When we repeat tests using Jakarta features.  I wonder if caching the resource adapter like this will make it so the transformer doesn't transform the app. 

At first glance I don't think it will.  I believe the transformer comes in when we use `ShrinkHelper.exportToServer(server, path, app, ... options)`.

It might be worthwhile to test this locally by repeating a single test class for EE9 and make sure that the app gets transformed.

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/HttpHelper.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Vector;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import componenttest.topology.impl.LibertyServer;
+
+public class HttpHelper {
```
KyleAure: I don't think this class is used anywhere in your test classes.  
And if it is it can be replaced with functionality from `FATServletClient`

dev/com.ibm.ws.jca_fat_regr/fat/src/suite/r80/base/jca16/TestSetupUtils.java
```diff
@@ -0,0 +1,220 @@
...
+import componenttest.topology.impl.LibertyServer;
+
+/**
+ * Class <CODE>TestSetupUtils</CODE> provides utility methods used
+ * to help set up resources needed for tests
+ */
+public class TestSetupUtils {
+
+    private final static String CLASSNAME = TestSetupUtils.class.getName();
+    private final static Logger LOGGER = Logger.getLogger(CLASSNAME);
```
KyleAure: Instead of registering a logger, we should be using the simplicity logger. 
`com.ibm.websphere.simplicity.log.Log`

dev/com.ibm.ws.jca_fat_regr/fat/src/suite/r80/base/jca16/ann/AdministeredObjectMergeActionTest.java
```diff
@@ -0,0 +1,1104 @@
...
+    }
+
+    @AfterClass
+    public static void tearDown() throws Exception {
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.entering(CLASSNAME, "tearDown");
+        }
+
+        server.stopServer("CWWKE0701E", // EXPECTED
+                          "CWWKE0700W");// EXPECTED
```
KyleAure: It would be good practice here to check to make sure the server is running before calling stopServer(). 

```java
      if (server.isStarted()) {
          server.stopServer();
      }
```

dev/com.ibm.ws.jca_fat_regr/fat/src/suite/r80/base/jca16/ann/AnnUtils.java
```diff
@@ -0,0 +1,749 @@
...
+                }
+            }
+        } finally {
+            if (s != null) {
+                s.close();
+                s = null;
+            }
+        }
+
+        return null;
```
KyleAure, line 75: It looks like this method is reading every line of trace!  
This may be a really big performance hit!

Depending on how this method is used we could try one of the following options: 

1. If our tests are sequential. Meaning the first test does (x) and then gets metadata, and then the second test does (y) and then gets metadata then we don't need to search the entire trace, just from the end of the last test, to now.  For this, you could use `server.findStringsInLogsAndTraceUsingMark(regex)`. 
2. If our tests are non-sequential. Meaning the server starts and both (x) and (y) are just searching logs from the server startup (or something). Then we could use `findStringsInTrace(regex)` This will return a list containing all metatypes and we'd then have to parse out which metatype is for which test (which will be quicker then searching the logs every time.

KyleAure (changes requested):

dev/com.ibm.ws.jca_fat_regr/fat/src/suite/r80/base/jca16/ann/AnnUtils.java
tabishop: I took a look at those methods and did some experimenting and found and they won't work in this instance because they are looking for instances of the regex within single lines. The metatype definitions span multiple lines so what it does is find the line starting the metatype def ("<metatype") and then captures all the subsesquent lines until it finds a line with ":MetaData>" and all the lines between the start and end are appended into 1 multi-line metatype string. I looked to see if there was an existing method that did what we want using marks, and did not find one.

dev/com.ibm.ws.jca_fat_regr/fat/src/suite/r80/base/jca16/ann/AdministeredObjectMergeActionTest.java
tabishop: I did a quick search and added this where I found server stops

dev/com.ibm.ws.jca_fat_regr.cfg.prop/fat/src/suite/r80/base/jca16/TestSetupUtils.java
tabishop: I was able to repeat the test class that uses this class for EE9 and it transformed and ran successfully so think we should be good.

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/JCAFATTest.java
tabishop: I had talked with Nathan about this in scrum, since the parameters don't match what is provided by the runTest methods in FATServletClient, we would have to change hundreds of locations in the code where this is used. Nathan didn't think the cost justified the benefit, but could be something we capture in another work item of potential optimizations for these test buckets.

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/app/InboundSecurityTest.java
tabishop: fixed

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/app/InboundSecurityTest.java
tabishop: removed

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/HttpHelper.java
tabishop: removed

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/JCAFATTest.java
tabishop: done

dev/com.ibm.ws.jca_fat_regr/fat/src/com/ibm/ws/jca/fat/regr/util/JCAFATTest.java
tabishop: made this change

KyleAure (approved):

dev/com.ibm.ws.jca_fat_regr/fat/src/suite/r80/base/jca16/ann/AnnUtils.java
KyleAure: Thanks for looking into this.  Let's consider improving the performance of this method as a stretch goal that we can do during the further development of the connectors-2.1 feature. Overall, since this test is just moving, we aren't negatively affecting performance as it is written currently.

tabishop:
#build

tabishop:
#libbybot

KyleAure (approved):

tabishop:
#libbybot

Issue #21314, opened 2022-06-02
MicroProfile Rest Client with checkpoint


tjwatson:
See checklist to help evaluate work needed for checkpoint: https://github.com/OpenLiberty/open-liberty/issues/18664#issuecomment-1146218456

tjwatson:
In particular we need to check how the config is read for the rest client.  See https://download.eclipse.org/microprofile/microprofile-rest-client-2.0/microprofile-rest-client-spec-2.0.html#mpconfig

tjwatson:
@KeenanMcCarthy can you update this issue with a description of what is tested for MicroProfile Rest Client?

KeenanMcCarthy:
This application invokes a rest client proxy that queries an endpoint defined by the "test client" system variable. Initially this variable is set to a default endpoint and returns a default value. After the default value is confirmed, the server is checkpointed and the system variable is changed to an alternate value. Following restore, the rest client is invoked and the alternate endpoint is queried, then reset to the default through a POST request. However, in instances where the rest client is initialized at startup, the configuration change does not persist through checkpoint. This test is repeated for both Java 8 and Java 9.

PR #21687, opened 2022-07-07
Created an application for testing rest client proxy with checkpoint

KeenanMcCarthy:
After checkpoint and restore, the application accesses an endpoint using a rest client with a default URI, and then requests an alternate endpoint using a URI defined at runtime.
for #21314


KeenanMcCarthy:
#build
#libby

tjwatson (approved):
LGTM and I suggest we merge as is (after a build).  As a follow on.  It would be good to test how `@RestClient` behaves when the `@ApplicationScoped` bean is constructed before a checkpoint.

Simply adding something like this to the client `@ApplicationScoped` class should cause early construction and injection on the checkpoint side:

```java
    public void observeInit(@Observes @Initialized(ApplicationScoped.class) Object event) {
        System.out.println(getClass() + ": " + "Initializing application context");
    }
```

The idea is to see how the config change on restore impacts the injected `@RestClient` after we know it has been injected with the original value during checkpoint.

tjwatson:
This PR is part of Epic #21358

PR #21708, opened 2022-07-11
Enable BELL properties for BETA

### LINKED ISSUE #17414: Enhance the bells-1.0 feature to support the configuration of BELL service properties
For epic #19368  

Consumers currently use Liberty properties to configure services provided by their BELL libraries. We could augment the `<bell>` metatype to allow the declaration of properties for services named within a `<bell>` configuration, and enhance the feature runtime to set these properties on service implementation classes during BELL initialization.  Here's a sample BELL configuration.
```
    <server>
          <featureManager>
              <feature>servlet-3.1</feature>
              <feature>bells-1.0</feature>
       </featureManager>
       <library id="exampleBellLib" name="exampleBellLib" description="exampleBellLib">
           <fileset dir="${server.output.dir}/sharedLib" includes="exampleBellLib.jar" />
       </library>
       <bell libraryRef="exampleBellLib"/>
```
A few concerns come to mind:

1) We could exploit this: The current bells-1.0 feature support collects properties in `META-INF/services` files; it also passes them into `BundleContext.registerService()` and caches them in the corresponding service factories.  This configuration must be proprietary, because neither the [java provider-configuration](https://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html#Service_Provider) nor the [service loader](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html) docs mention property support.

2. This enhancement will require a proprietary method of setting properties on a service class.  Do we care?  The runtime would collect service properties from the `<bell>` configuration into the service factory.  Before returning the service instance the factory would invoke instance methods `setX(_xValue_)` -- a la JavaBeans -- for all properties, or invoke instance method init(map) with a Map of properties. 
 
3. [Bell documentation](https://openliberty.io/docs/21.0.0.5/reference/config/bell.html) says the user may declare the name of the Meta-inf service provided by the BELL library -- i.e., a single service name.  If this is a valid logical restriction, then the configuration of service properties could simply entail adding a property attribute to the bell OCD, as shown below.  Note that the runtime does not enforce this restriction.  So, if we expect the user to name more than one service in a `<bell>`, then the Bells metatypes will require another OCD for the server, with a required name attribute and an optional attribute for service properties.  This sounds like a breaking change to me, which is circumvented by 1).
```
    <OCD name="%bell" description="%bell.desc"
             id="com.ibm.ws.classloading.bell"
             ibm:alias="bell">
        <AD name="%bell.library.ref" description="%bell.library.ref.desc"
                id="libraryRef", required="true", type="String", ibm:type="pid"
                ibm:reference="com.ibm.ws.classloading.sharedlibrary"
         />
        <AD id="library.target" type="String" required="true" default="(service.pid=${libraryRef})" name="internal" description="internal use only" ibm:final="true" />
        <AD id="service" name="%service.name" description="%service.name.desc" required="false" type="String" cardinality="2147483647"/>
        **<AD id="service.property" name="%service.property.name" description="%service.property.desc" required="false" type="String" cardinality="2147483647"/>**
    </OCD>
```

dazavala:
The implementation and FAT for BELL properties support pursuant to socialized design changes.

### DIFF
--- a/dev/com.ibm.ws.classloading.bells/resources/OSGI-INF/l10n/bells.properties
+++ b/dev/com.ibm.ws.classloading.bells/resources/OSGI-INF/l10n/bells.properties
@@ -26,7 +26,7 @@ service.name.desc=The name of the service that the system will look up in the /M
 
 # BETA
 enableSpiVisibility.name=Enable SPI visibility
-enableSpiVisibility.desc=Indicates whether or not the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
+enableSpiVisibility.desc=Indicates whether the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
 
 # BETA
 properties.name=Properties
--- a/dev/com.ibm.ws.classloading.bells/resources/OSGI-INF/l10n/bells.properties
+++ b/dev/com.ibm.ws.classloading.bells/resources/OSGI-INF/l10n/bells.properties
@@ -26,4 +26,8 @@ service.name.desc=The name of the service that the system will look up in the /M
 
 # BETA
 enableSpiVisibility.name=Enable SPI visibility
-enableSpiVisibility.name.desc=Indicates whether or not the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
+enableSpiVisibility.desc=Indicates whether or not the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
+
+# BETA
+properties.name=Properties
+properties.desc=Name="Value" pairs the system will pass to a service implementation at creation and update. The service implementation class must declare a public constructor, or a public method named updateBell, with a single parameter of type java.util.Map<String, String>.
--- a/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
+++ b/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
@@ -102,6 +102,6 @@ bell.spi.visibility.disabled.libref.global.useraction=Modify the BELL configurat
 # {0} - the name of the implementation class
 # {1} - the location of the META-INF/services file, probably a URL that also indicates which jar it is in
 # {2} - the name of the library
-bell.missing.property.injection.methods=CWWKL0062W: The system cannot inject BELL properties into an instance of the {0} implementation class because the class is missing a public constructor or updateBell method having a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
-bell.missing.property.injection.methods.explanation=The system cannot access the public constructor or updateBell method it requires to inject BELL properties into the service implementation instance.
-bell.missing.property.injection.methods.useraction=Ensure the class defines a public constructor or updateBell method having a single parameter of type java.util.Map<String, String>.
+bell.missing.property.injection.methods=CWWKL0062W: The system did not inject BELL properties into an instance of the {0} implementation class. The class does not define a public constructor, or a public updateBell method, with a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
+bell.missing.property.injection.methods.explanation=The system cannot access the public constructor or the updateBell method required to inject BELL properties into the service implementation instance.
+bell.missing.property.injection.methods.useraction=Ensure that the class defines a public constructor, or a public updateBell method, with a single parameter of type java.util.Map<String, String>.
--- a/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
+++ b/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
@@ -97,3 +97,11 @@ bell.spi.visibility.disabled.libref.global.useraction=Modify the BELL configurat
 #bell.spi.visible.share.common.libref=CWWKL0061W: The {0} application and BELL share a common library reference, but do not reference the {1} library through a common class loader because the BELL is enabled for SPI visibility.
 #bell.spi.visible.share.common.libref.explanation=To prevent applications and resources from seeing SPI packages, the BELL references the library through a unique class loader that is not shared with the application.
 #bell.spi.visible.share.common.libref.useraction=Disable SPI visibility in the BELL configuration whenever the application and BELL must share a common library class loader. Otherwise, no action is required.
+
+# BETA
+# {0} - the name of the implementation class
+# {1} - the location of the META-INF/services file, probably a URL that also indicates which jar it is in
+# {2} - the name of the library
+bell.missing.property.injection.methods=CWWKL0062W: The system cannot inject BELL properties into an instance of the {0} implementation class because the class is missing a public constructor or updateBell method having a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
+bell.missing.property.injection.methods.explanation=The system cannot access the public constructor or updateBell method it requires to inject BELL properties into the service implementation instance.
+bell.missing.property.injection.methods.useraction=Ensure the class defines a public constructor or updateBell method having a single parameter of type java.util.Map<String, String>.
--- a/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
+++ b/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
@@ -102,6 +102,6 @@ bell.spi.visibility.disabled.libref.global.useraction=Modify the BELL configurat
 # {0} - the name of the implementation class
 # {1} - the location of the META-INF/services file, probably a URL that also indicates which jar it is in
 # {2} - the name of the library
-bell.missing.property.injection.methods=CWWKL0062W: The system did not inject BELL properties into an instance of the {0} implementation class. The class does not define a public constructor, or a public updateBell method, with a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
-bell.missing.property.injection.methods.explanation=The system cannot access the public constructor or the updateBell method required to inject BELL properties into the service implementation instance.
-bell.missing.property.injection.methods.useraction=Ensure that the class defines a public constructor, or a public updateBell method, with a single parameter of type java.util.Map<String, String>.
+bell.missing.property.injection.methods=CWWKL0062W: The system did not inject BELL properties into an instance of the {0} implementation. The class does not define a method that supports BELL property injection. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
+bell.missing.property.injection.methods.explanation=The system cannot access a method required for BELL property injection. The service implementation class must define a public constructor, or a public method named updateBell, with one parameter of type java.util.Map<string,String> in order to support BELL property injection.
+bell.missing.property.injection.methods.useraction=Ensure that the class defines a public constructor, or a public method named updateBell, with a single parameter of type java.util.Map<String, String>.
--- a/dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
+++ b/dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
@@ -16,6 +16,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.URL;
@@ -25,6 +26,7 @@ import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Dictionary;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.LinkedList;
@@ -93,16 +95,18 @@ public class Bell implements LibraryChangeListener {
         }
     }
 
-    private static final String SERVICE_ATT = "service";
     private final ReentrantLock trackerLock = new ReentrantLock();
     private ServiceTracker<Library, List<ServiceRegistration<?>>> tracker;
 
     private Library library;
 
     private ComponentContext componentContext;
+
     private Map<String, Object> config;
 
+    private static final String SERVICE_ATT = "service";
     private static final String ENABLE_SPI_VISIBILITY_ATT = "enableSpiVisibility";
+    private static final String PROPERTIES_ATT = "properties";
 
     @Activate
     protected void activate(ComponentContext cc, Map<String, Object> props) {
@@ -145,8 +149,8 @@ public class Bell implements LibraryChangeListener {
         final String libraryRef = library.id();
 
         // Determine the service filter to use for discovering the Library service this bell is for
-        // it is unclear if only looking at the id would work here.
-        // other examples in classloading use both id and service.pid to look up so doing the same here.
+        // it is unclear if only looking at the id would work here. Other examples in classloading use
+        // both id and service.pid to look up so doing the same here.
         String libraryStatusFilter = String.format("(&(objectClass=%s)(|(id=%s)(service.pid=%s)))", Library.class.getName(), libraryRef, libraryRef);
         Filter filter;
         try {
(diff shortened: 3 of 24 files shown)


dazavala:
#build
#libbybot

dazavala:
@tjwatson FYI, these changes do not perform a non-destructive update on BELL service implementation instances when the user changes a BELL configuration.  We can address this at review; introducing the behavior requires significant change to the feature design.  Note also, a non-destructive update is feasible only when properties are updated.  Any change to SPI visibility or the referenced library from the previous generation will require a destructive BELL update.

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
```diff
@@ -205,41 +209,62 @@ public void removedService(ServiceReference<Library> libraryRef, List<ServiceReg
...
+        if (isBetaEdition && !!!issuedBetaMessage) {
+            Tr.info(tc, "BETA: BELL SPI Visibility and BELL Properties has been invoked by class " + this.getClass().getName() + " for the first time.");
+            issuedBetaMessage = true;
+        }
+        return isBetaEdition;
+    }
+
+    @SuppressWarnings("restriction")
+    boolean getSpiVisibility(boolean enableSpiVisibility, String libraryId) {
+        if (enableSpiVisibility) {
```
tjwatson: Before you seemed to handle null `Boolean` object from config.  Are we now guaranteed to never have a null value for the key `ENABLE_SPI_VISIBILITY_ATT`?

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
```diff
@@ -205,41 +209,62 @@ public void removedService(ServiceReference<Library> libraryRef, List<ServiceReg
...
-            issuedBetaMessage = true;
         }
-        return isBeta;
+        return enableSpiVisibility;
     }
 
-    private static Set<String> getServiceNames(String[] configuredServices) {
-        if (configuredServices == null || configuredServices.length == 0) {
-            return Collections.emptySet();
+    private static final String propKeyPrefix = PROPERTIES_ATT + ".0.";
```
tjwatson, line 246: Where does the `.0.` come from?  Is that just inherent in the way liberty config is passed here?

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
```diff
@@ -205,41 +209,62 @@ public void removedService(ServiceReference<Library> libraryRef, List<ServiceReg
...
+     * @return a map containing zero or more BELL properties, otherwise return null whenever
+     *         the configuration lacks a properties element.
+     */
+    private Map<String, String> getProperties(Map<String, Object> config) {
+        Map<String, String> pMap = null;
+        if (config.get(propKeyPrefix + "config.referenceType") != null) {
+            pMap = new HashMap<String, String>();
+            for (String key : config.keySet()) {
+                if (key.startsWith(propKeyPrefix) && !!!key.endsWith("config.referenceType")) {
+                    String pValue = (String) config.get(key);
```
tjwatson: Do we need a safe guard check here to make sure the value is an `instanceof String` here before trying to add it to the map?

I would suggest
```java
    Object pValue = config.get(key);
    if (pValue instanceof String) {
        String pName = key.substring(propKeyPrefixLen);
        pMap.put(pName, (String) pValue);
    }
```

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
```diff
@@ -516,10 +543,27 @@ private static Class<?> findClass(final String implClass, final Library library,
...
         try {
-            service = serviceType.newInstance();
+            if (properties == null) {
+                service = serviceType.newInstance();
+            } else {
+                // Inject a shallow clone of the properties map
+                final Constructor<?> singleArgCtor;
+                final Method updateMethod;
+                if ((singleArgCtor = getConstructor(serviceType, java.util.Map.class)) != null) {
+                    service = singleArgCtor.newInstance(((HashMap<String, String>) properties).clone());
```
tjwatson: Can we just document that the map<string,string> passed to the service is unmodifiable?

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
```diff
@@ -516,10 +543,27 @@ private static Class<?> findClass(final String implClass, final Library library,
         return service;
     }
 
-    private static Object createService(final Class<?> serviceType, final String libID, final URL fileUrl) {
+    private static Object createService(final Class<?> serviceType, final String libID, final URL fileUrl, final Map<String, String> properties) {
         Object service = null;
         try {
-            service = serviceType.newInstance();
+            if (properties == null) {
+                service = serviceType.newInstance();
```
tjwatson, line 555: What if the service implementation only has a constructor that takes a `Map<String,String>`?  Here we always force them to have a no arg constructor?  Should we fall back to calling the `ctor(Map)` with an empty map if there is no, no-arg constructor available?

dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
```diff
@@ -97,3 +97,11 @@ bell.spi.visibility.disabled.libref.global.useraction=Modify the BELL configurat
...
 #bell.spi.visible.share.common.libref.useraction=Disable SPI visibility in the BELL configuration whenever the application and BELL must share a common library class loader. Otherwise, no action is required.
+
+# BETA
+# {0} - the name of the implementation class
+# {1} - the location of the META-INF/services file, probably a URL that also indicates which jar it is in
+# {2} - the name of the library
+bell.missing.property.injection.methods=CWWKL0062W: The system cannot inject BELL properties into an instance of the {0} implementation class because the class is missing a public constructor or updateBell method having a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
+bell.missing.property.injection.methods.explanation=The system cannot access the public constructor or updateBell method it requires to inject BELL properties into the service implementation instance.
+bell.missing.property.injection.methods.useraction=Ensure the class defines a public constructor or updateBell method having a single parameter of type java.util.Map<String, String>.
```
ramkumar-k-9286:
```suggestion
bell.missing.property.injection.methods.useraction=Ensure that the class defines a public constructor or the updateBell method has a single parameter of type java.util.Map<String, String>.
```

dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
```diff
@@ -97,3 +97,11 @@ bell.spi.visibility.disabled.libref.global.useraction=Modify the BELL configurat
...
 #bell.spi.visible.share.common.libref.useraction=Disable SPI visibility in the BELL configuration whenever the application and BELL must share a common library class loader. Otherwise, no action is required.
+
+# BETA
+# {0} - the name of the implementation class
+# {1} - the location of the META-INF/services file, probably a URL that also indicates which jar it is in
+# {2} - the name of the library
+bell.missing.property.injection.methods=CWWKL0062W: The system cannot inject BELL properties into an instance of the {0} implementation class because the class is missing a public constructor or updateBell method having a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
+bell.missing.property.injection.methods.explanation=The system cannot access the public constructor or updateBell method it requires to inject BELL properties into the service implementation instance.
```
ramkumar-k-9286:
```suggestion
bell.missing.property.injection.methods.explanation=The system cannot access the public constructor or the updateBell method required to inject BELL properties into the service implementation instance.
```

dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
```diff
@@ -97,3 +97,11 @@ bell.spi.visibility.disabled.libref.global.useraction=Modify the BELL configurat
...
 #bell.spi.visible.share.common.libref.explanation=To prevent applications and resources from seeing SPI packages, the BELL references the library through a unique class loader that is not shared with the application.
 #bell.spi.visible.share.common.libref.useraction=Disable SPI visibility in the BELL configuration whenever the application and BELL must share a common library class loader. Otherwise, no action is required.
+
+# BETA
+# {0} - the name of the implementation class
+# {1} - the location of the META-INF/services file, probably a URL that also indicates which jar it is in
+# {2} - the name of the library
+bell.missing.property.injection.methods=CWWKL0062W: The system cannot inject BELL properties into an instance of the {0} implementation class because the class is missing a public constructor or updateBell method having a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
```
ramkumar-k-9286:
```suggestion
bell.missing.property.injection.methods=CWWKL0062W: The system cannot inject BELL properties into an instance of the {0} implementation class. The class is missing a public constructor or the updateBell method with a single parameter of type java.util.Map<String,String>. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
```

dev/com.ibm.ws.classloading.bells/resources/OSGI-INF/l10n/bells.properties
```diff
@@ -26,4 +26,8 @@ service.name.desc=The name of the service that the system will look up in the /M
 
 # BETA
 enableSpiVisibility.name=Enable SPI visibility
-enableSpiVisibility.name.desc=Indicates whether or not the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
\ No newline at end of file
+enableSpiVisibility.desc=Indicates whether or not the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
```
ramkumar-k-9286, line 29:
```suggestion
enableSpiVisibility.desc=Indicates whether or not the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
```
```suggestion
enableSpiVisibility.desc=Indicates whether the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
```

ramkumar-k-9286:
```suggestion
enableSpiVisibility.desc=Indicates whether the system makes feature SPI packages visible to the library referenced by the BELL. Enable SPI visibility whenever the BELL provides services that use feature SPI.
```

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
dazavala: Yes.  I verified the config service sets the default value for attribute `enableSpiVisibility` when the attribute is not configured within the BELL XML and when the attribute is configured incorrectly (i.e. not configured as `true` or `false`). The feature runtime further safeguards against this concern by relying on Java to unbox the Boolean value to boolean when passing the parameter by value to method `getSpiVisibility()`.
 
 I'll add a test for invalid `enableSpiVisibility` configuration to the BELL FAT which checks for this condition:
`CWWKG0083W: A validation failure occurred while processing the [enableSpiVisibility] property, value = [XXX]. Default value in use: [false].`

tjwatson: > The feature runtime further safeguards against this concern by relying on Java to unbox the Boolean value to boolean when passing the parameter by value to method `getSpiVisibility()`.

That was what triggered my concern.  Because unboxing a null `Boolean` will result in a NPE (which is hard to debug when it does happen because the stacktrace doesn't make it obvious on Java < 17).  But if we are guaranteed to not have a null here then it should be fine.

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
dazavala: That's the inherent way liberty "flat" config is passed to the Bell DS instance on activation.  I provided a sample property config below that shows the attribute format. Attribute `properties.0.config.referenceType` appears only when the BELL XML declares a `<properties/>` element.

Some notes about property collection:

1. An empty `<properties/>` element is a valid configuration. The feature runtime will inject an empty map into the service impl instance.
2. Multiple `<property/>` elements is a valid configuration. The property attribute format remains consistent when multiple `<property/>` elements are configured. The feature runtime collects all property attributes with prefix "properties.0."
3. Properties may be declared more than once. The feature runtime overrides values for duplicate properties as they are collected from the config and _does not issue a warning_.  

{component.name=com.ibm.ws.classloading.bell, ..., properties.0.P2=P2_VAL, properties.0.P1=P1_VAL, properties.0.P0=FEED ME, SEYMORE!, properties.0.config.referenceType=com.ibm.ws.classloading.bell.properties }

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
dazavala: Oh no -- exactly the opposite of what I intended!  I hardened and clarified the solution against NPE.  
```
boolean getSpiVisibility(Boolean enableSpiVisibility, String libraryId) {
    boolean spiVisibility = Boolean.TRUE.equals(enableSpiVisibility);
    if (spiVisibility) {
        ...
```

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
dazavala: Done.

dev/com.ibm.ws.classloading.bells/resources/OSGI-INF/l10n/bells.properties
dazavala: Done.

dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
dazavala: Done.  Note the minor edit involving commas.

dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
dazavala: Hello @ramkumar-k-9286.  In the draft message the phrase "having a single parameter..." is intended to distribute -- that is, the intended meaning is this: the user must define in their service impl class either a public constructor with a single parameter of type Map<S,S>, or a public method updateBell with a single parameter of type Map<S,S>, or both.  

So, I made your changes as indicated and inserted commas to reinforce that the parameter type applies to both the constructor and the updateBell method, _rather than only the updateBell method._

Lastly, after working your changes I conjured this version.  Let me know if you prefer it.

```
bell.missing.property.injection.methods=CWWKL0062W: The system did not inject BELL properties into an instance of the {0} implementation. The class does not define a method that supports BELL property injection. The class is an implementation of the service that is declared in the {1} file and registered from the {2} library.
bell.missing.property.injection.methods.explanation=The system cannot access a method required for BELL property injection. The service implementation class must define a public constructor, or a public method named updateBell, with one parameter of type java.util.Map<string,String> ni order to support BELL property injection.
bell.missing.property.injection.methods.useraction=Ensure that the class defines a public constructor, or a public method named updateBell, with a single parameter of type java.util.Map<String, String>.
```

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
dazavala: For now I'd prefer to leave the solution as-is.  According to documentation, and some implicit convention mentioned in our design call, a service implementation class is required to provide a public zero-arg constructor. It's possible a user will ask for the behavior, and we can reconsider at that time.

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
dazavala: Yes, I'll document the map is unmodifiable, because I'm having trouble enforcing this contract in the code using a unmodifiableMap -- specifically, I'm encountered trouble passing a unmodifiableMap(hashMap<s,s>) over a parameter of type Map<s,s>:

> The following error occurred: java.lang.ClassCastException: java.util.Collections$UnmodifiableMap incompatible with java.util.HashMap
> 	at com.ibm.ws.classloading.bells.internal.Bell.createService(Bell.java:559)
> 	at [internal classes]

dazavala:
#build
#libby

dev/com.ibm.ws.classloading.bells/src/com/ibm/ws/classloading/bells/internal/Bell.java
dazavala: @tjwatson Thanks again for helping investigate this insane failure. Clearing the java shared class cache did the trick.  Changes are uploaded to the review commit. I'll still document the injected property map as unmodifiable, since it is.

dazavala:
#build

tjwatson (approved):

dazavala:
#build

dazavala:
#build

ramkumar-k-9286 (approved):

dazavala:
#build

dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
dazavala: We resolved this conversation, and the others concerning changes to message CWWKL0062W, by using the new message posted above, which required a quick typo fix to change "ni" to "in".  The new message appears in commit "ID changes".

dazavala:
#libbybot

Issue #21307, opened 2022-06-02
MicroProfile Config with checkpoint


tjwatson:
See checklist to help evaluate work needed for checkpoint: https://github.com/OpenLiberty/open-liberty/issues/18664#issuecomment-1146218456

tjwatson:
We need to add more tests around MicroProfile Config.  In particular anything that is using application scope with injected configuration.  We need to see what happens when config changes on restore for injection into application scope CDI beans.

anjumfatima90:
Injected Config is tested in the following tests. Config is injected statically using `@ConfigProperty` and dynamically using a `Config` object. The tests cover testing the injected default value, annotation value, environment variable and server config.

1. **MPConfigBean.java** and **MPConfigBeanWithApplicationScope.java**

- Injected config property (`@Configproperty`) is tested using `@RequestScoped` and `@ApplicationScoped` beans. The FAT tests if the change is config which is done after the checkpoint is positively reflected during restore. 

2. **ApplicationScopedOnCheckpointBean.java**

- Here, for `@ApplicationScoped` beans, we notice that, when the statically injected config property (`@ConfigProperty`) is accessed during the checkpoint, then the change is config done after checkpoint is not reflected on restore. Therefore, the tests verifies if a warning message `CWWKC0651W`  that lets the user know that the updated value of the property might not be used during restore is being logged or not.

3. **ApplicationScopedOnCheckpointBeanWithConfigObject.java** and **ApplicationScopedOnCheckpointBeanWithConfigObjectProperties.java** 

- Here, config is injected dynamically. It also tests that the updated config gets used on restore even after the config was accessed during the checkpoint for `@ApplicationsScoped` beans.

PR #21728, opened 2022-07-12
Test when config changes on restore for injection into application scope CDI beans.

anjumfatima90:
for #21307


dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
```diff
@@ -76,6 +76,7 @@ private void configureBeforeCheckpoint(TestMethod testMethod) {
     }
 
     private void configureBeforeRestore(TestMethod testMethod) {
+        ServerConfiguration config;
```
tjwatson, line 79: If this isn't going to be used outside the scope of the case statement I would not declare it here.  I would leave it local to each case statement that it is created in.

tjwatson (approved):

anjumfatima90:
#build

anjumfatima90:
#libby

Issue #21707, opened 2022-07-11
server checkpoint command needs better error handling


sebratton:
On a platform where checkpoint is not supported (non-Linux) or where the criu executable is not on the PATH, running the checkpoint command emits a shell scripting error

> bin/server checkpoint cp1
bin/server: line 333: criu: command not found

and possibly an intelligble NOT-SUPPORTED Liberty errorr message depending on weather or not the com.ibm.ws.beta.edition=true java system property is set.

This is caused by the callout to the function 'checkCriuUnprivileged()' which unconditionally invokes criu.

Not sure how to fix this cleanly. Maybe with a check for criu on the path prior to calling it and respond in a better way if not there. Whatever is done need to be tested on all platforms that invoke bin/server (zLinux, cygwin, OS400, linux). 

A similar problem exists on restore if criu was removed after a checkpoint image was created. But that does not seem as problematic since it does not affect platforms where we never intended criu to be supported

sebratton:
Closing as fixed by above mentioned PR.

PR #21742, opened 2022-07-13
Remove extra call to criu from the bin/server script.

sebratton:
After this change 'criu --help' is no longer used to determine whether or not to invoke criu in unprivileged mode (by passing in the --unprivileged option). Instead it's determined by a combination of an environment variable and whether or not the user is root.


tjwatson (approved):

sebratton:
#build

sebratton:
#libbybot

sebratton:
#build

sebratton:
Fixes #21707

PR #21744, opened 2022-07-13
Remove mpOpenAPI-3.1 from compatible features to avoid clash

tevans78:
The Jakarta EE 9 feature compatibility test should only use mpOpenAPI-3.0 and NOT mpOpenAPI-3.1


jhanders34 (approved):

tevans78:
Merging without PB because this is a trivial fix for something we know will be broken ... and may delay other beta work.

PR #21800, opened 2022-07-19
Update JPA Query FAT for changes needed to run on DB2 on zOS

jgrassel:
Needed to restore erroneously removed `CCSID UNICODE` in the test bucket's DDL files.  Updated the hibernate runner to require system property `jpa.enable.repeat.hibernate` to true in order to run the FAT with third party hibernate provider.


jgrassel:
#build

tkburroughs (approved):

PR #21810, opened 2022-07-19
Delay creation of LibertyProxyInvocationHandler until first invoke

tjwatson:
Creating the LibertyProxyInvocationHandler first requires building
the ResteasyClient.  That requires access to the JVM security
stack that is not available before the checkpoint. This change
delays creating the LibertyProxyInvocationHandler (and therefore
the ResteasyClient) until the rest client instance is first invoked
by application code. If nothing invokes the rest client object
before the checkpoint then that allows for the JVM security stack
to be avoided until the restore side.

This approach still has a issue that the rest client cannot be
configured with respect to the URI used to connect to for invoking
the rest client on the restore side.


tjwatson:
#build

PR #21812, opened 2022-07-20
Fix test bug

dazavala:
Fix bug in SpringBoot JavaEE tests where ORB SSL initialization exceeds 30 seconds, causing applications not to start. 

RTC 291351.


dazavala:
#build

sebratton (approved):

PR #21823, opened 2022-07-20
experiment with version

njr-11:
Experiment with identifying a version attribute and supplying it to orm.xml and with the behavior we get for the simulated Jakarta Data template/repository methods.


njr-11:
#build

mswatosh (approved):

PR #21825, opened 2022-07-21
Run test against https://github.com/eclipse-equinox/equinox/pull/68 DO NOT MERGE

tjwatson:
Testout https://github.com/eclipse-equinox/equinox/pull/68


tjwatson:
#build

PR #21838, opened 2022-07-22
Additional log mark reset for ACME fat

kristip17:
Add an additional log mark reset so we don't pick up an old Cert Checker wake up and run in `AcmeConfigVariationsTest`.

For RTC 291261


kristip17:
#build

jvanhill (approved):

PR #21877, opened 2022-07-26
Handle extra Timed out msg during server start retry

c00crane:
The SSO fats have retry logic for starting test servers.  Occasionally, we see problems that prevent test servers from starting.  Our framework will stop that server and try to start it again.
The instance that caused an overall failure was where test apps did not start and we had to retry starting the server.
The overall fat framework logged a Timed out message when it was searching for the app started message.  Our portion of the framework adds "1" to the count of allowed timed out msgs.  The problem is we're logging the failure that we got back from the overall framework.  This failure included "Time out".
So, I'm increasing the number of allowed timed out msgs by 2 instead of one.


c00crane:
#build

teddyjtorres (approved):

PR #21891, opened 2022-07-27
Keep track of testcontainer projects and images

KyleAure:
A proposed solution to keep track of testcontainer projects, and the image(s) they use.


websterc87:
So the file is generated by the developer running the gradle build on there own laptop, feels like users will forget to run it prior to delivery and it will get out of date. Feels like this is wrong approach. I think we need to find a different route. 

Would it work for the FATs read from the images from a shared file e.g. Constants in a common java class.  That way when you need to update an image you update the shared class?

KyleAure:
@websterc87 Thanks for the feedback.
I can think of a couple of different approaches and their pros/cons:
1. The approach in this PR currently: We risk becoming out of date if a developer never does a run of their PR on their local system.  However, we guarantee that if a test container is run that it is recorded.  This means that unaware developers don't need to think about it and can't forget to update this list (common issue). 
2. The approach you suggested: Having a constants class that holds all the images we use.  A developer could still declare their own image in their test class and wouldn't know about the shared class unless we threw an error.  If the developer isn't running their tests locally, that error won't show up until they've already started a personal build.  Additionally, we'd have to keep two lists, one of images, and one of test classes.
3. Or do I re-visit the idea of keeping the images used within a bucket in the bnd.bnd file? Then verify that those images are the only ones used when running both during local and remote runs.  Then the automation tool would just need to inspect the bnd file and look for a certain property to determine if it uses test containers or not. 

I'm open to other suggestions.

dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
```diff
@@ -30,21 +39,26 @@ public class ArtifactoryImageNameSubstitutor extends ImageNameSubstitutor {
 
     @Override
     public DockerImageName apply(DockerImageName original) {
-        // Priority 1: If we are using a programmatically built image, or a registry was explicit set, do not substitute
-        if (isSyntheticImage(original) || (original.getRegistry() != null && !original.getRegistry().isEmpty())) {
+        // Priority 1: If we are using a synthetic image do not substitute nor cache
+        if (isSyntheticImage(original)) {
```
rickettmwork, line 34: Could you explain why we don't want to cache in this case?

rickettmwork:
For (2) why do we need a list of test classes?

dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
KyleAure: A synthetic image is one that is built at runtime.  We discourage the use of synthetic images because of the performance impact. But we still check for them in our test infrastructure and handle them correctly. 
We wouldn't want to cache their names because by their nature they are never stored in GitHub/Artifactory. 
Their names are typically a random hash like: `localhost/testcontainers/jxbqdhfgsomphxr5:latest`

KyleAure:
> For (2) why do we need a list of test classes?

This was a request by the build team to be able to mark test suites as using testcontainers in the Cognitive Tool

KyleAure:
@rickettmwork @websterc87 Updated this PR with the original plan implemented. 
I will verify images have been added to the `bnd.bnd` file using the property `fat.test.container.images`: 
```
fat.test.container.images: kyleaure/postgres-test-table:3.0
```

Then, I added a gradle script to the `io.openliberty.org.testcontainers` project that will compile a list of projects that use testcontainers and a list of images. 
Note that the image list only contains original names, and not substituted names.  So when we preload our docker hosts we'll want to make sure to pull images from our artifactory mirror and not docker-hub.

rickettmwork (approved):

rickettmwork:
Approved, but it isn't ideal to have to run a gradle task to get the list of containers that we need to pre-cache because that adds a dependency on gradle to the infrastructure setup system.

KyleAure:
@rickettmwork 
>  it isn't ideal to have to run a gradle task to get the list of containers that we need to pre-cache

You won't have to run it, this data should stay up to date on GitHub via this line: 
https://github.com/OpenLiberty/open-liberty/blob/2856c15ccfbeb9189ec68c4e5a2cce6162a0e764/dev/io.openliberty.org.testcontainers/build.gradle#L77

Anytime someone makes an update to a project that adds/removes a container image, they will need to update their bundle.  When they run their test again this will force the test container bundle to also be re-built and the list will be updated. If the bnd file isn't updated their test will fail so this will force the developer and this list to always be up to date.

It is possible that a developer will update their bnd file, but never run their test suite locally (bad practice).  But anyone else in the org that runs one of these test suites will also re-generate the file and include the updated file in whatever PR they create.  Similar to how the dependabot list gets updated.  

**NOTE: I haven't updated all projects yet because I want to have approval on a design before I put in the effort :)**

KyleAure:
#build

Expect this build to fail as not all projects are updated. Verify the following: 
1. Projects that only use containers from `fattest.simplicity` don't require any updates to their bnd.
2. Projects that use other images will need updates and should fail.

KyleAure:
#build should work now.

KyleAure:
#build

KyleAure:
#build

KyleAure:
#build

KyleAure:
#build

KyleAure:
Re-ran child build that failed: 
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_Cw_U0B1mEe2b2O1gXXKb4w
Failures looked to be related to existing defects, and issues communicating to LibertyFS

KyleAure:
#libbybot

wraschke:
When coexisting with https://github.com/OpenLiberty/open-liberty/pull/22104, this change causes a test failure in io.openliberty.jcache.internal_fat.infinispan. See defect 292102.

PR #21906, opened 2022-07-28
Fix test bug 291351

dazavala:
Fix BELL properties test `testPropertiesInvalid` for beta and non-beta editions. The regex argument `.*CWWKG0014E:.*TIP_P0.*properties` passed into stopServer() expects the term "properties" to appear in msg CWWKG0014E. The test verifies CWWKG0014E mentions the property name w/ the syntax error, which is in the `<properties/>` element. Strange that the error message no longer contains the element name ("properties").  I'll reduce the regex to `CWWKG0014E:.*TIP_P0`.


dazavala:
#build

sebratton (approved):

dazavala:
Thanks, @sebratton. Merge pending personal build results.

PR #21908, opened 2022-07-28
Updating jca tests to not run EE10 features on Java 8

mswatosh:
Fixing issue with no tests run on Java 8 when the EE10 repeat action is the only repeat action running lite mode tests.


mswatosh:
#build

KyleAure (approved):

PR #21912, opened 2022-07-28
Adjust attributeCache timeout for LdapRegressionTest

kristip17:
Sometimes on windows runs, the attributes cache was timing out before the JUnit test could do the next step. Add a slightly longer AttributeCache timeout.

For RTC 290934


kristip17:
#build

acdemyers (approved):

PR #21938, opened 2022-08-02
tests for resource definitions with invalid ContextService

njr-11:
As requested by the functional acceptance test approver, I'm adding tests for when the resource definitions are configured with an invalid ContextService, either one that doesn't exist or with a name that does exist but isn't a ContextService.


mswatosh (approved):

njr-11:
#build

PR #21939, opened 2022-08-02
Fix build break in RTC 291008. Wait for HTTPS endpoint to startup aft…

jvanhill:
…er restarting server before calling HTTPS endpoint.


jvanhill:
#build

kristip17 (approved):

PR #21942, opened 2022-08-02
Fix NPE when executorSrvc gets nulled out on shutdown

tjwatson:
If the server is shutdown while we are submitting the
task to generate the plugin XML file an NPE is possible.

The fix is to assign the local variable to do the null check
and call to the executor


tjwatson:
#build

epj (approved):
Thanks for the quick response.   Change looks good to me.

PR #21961, opened 2022-08-03
increase allowed ffdc retries and timeout, and remove some dead code

tabishop:
Defect 291664 was caused by the test case thinking that an expected ffdc was not getting created. It turns out, it was getting created, there was just a delay ion the ffdc getting flushed so the test case gave up before the ffdc was produced. This fix increases the retries and timeout and also deletes some dead code.


tabishop:
#build

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
```diff
@@ -234,10 +235,12 @@ public void evaluate() throws Throwable {
                          * This sleep is arbitrary, if we continue to have problems, a longer sleep could be tried. Or a new solution and remove
                          * this retry.
                          */
-                        if (!expectedFFDCs.isEmpty() && !errors.isEmpty()) {
+                        if (!expectedFFDCs.isEmpty() && !errors.isEmpty() && (i + 1 <= retryCount)) {
```
KyleAure, line 238: instead of doing an extra `i` check here why not add the Thread.sleep(retryInterval) to the latter half of the for loop. 

```
for (int i = 1, retryInterval = 120; i <= retryCount; i++, Thread.sleep(retryInterval), retryInterval *= 10)
```

You'd have to check to make sure this works.  Haven't worked much with multivariable for loops in java.

KyleAure (approved):
This will work as is, but I added a suggestion on how we could avoid some extra checks to make it more understandable.

dev/fattest.simplicity/src/componenttest/custom/junit/runner/FATRunner.java
tabishop: Your way would add an extra extraneous sleep at the end without then entering the loop to check again, so I'm inclined to leave it as is.

KyleAure: Ohhh you are totally right.  👍🏼 I missed that part.

tabishop:
#libby

tabishop:
#build

kristip17 (approved):

tabishop:
The failure in the personal build is in a ui test and is unrelated to the change in this PR.

PR #21964, opened 2022-08-03
result and aggregate functions

njr-11:
Experiment with aggregate functions.
One way to do this could be on the Select annotation, which will make more sense if we separate out the type, which will also make sense for disambiguating delete/update results.


njr-11:
#build

njr-11:
#build

tabishop (approved):

PR #21987, opened 2022-08-04
Fix test bug 291759

dazavala:
Modify the SpringBoot TemplateTest to be less sensitive to Windows file locking.

### DIFF
--- a/dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/TemplateTests.java
+++ b/dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/TemplateTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 IBM Corporation and others.
+ * Copyright (c) 2018, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -18,8 +18,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.junit.After;
 import org.junit.AfterClass;
-import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


dazavala:
#build

dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/TemplateTests.java
```diff
@@ -61,34 +59,44 @@ public static void tearDown() throws Exception {
...
     public void testCreateServerFromSpringBoot2() throws Exception {
         expectedFeatures = new String[] { "servlet-4.0", "ssl-1.0", "transportSecurity-1.0", "websocket-1.1", "springBoot-2.0" };
+        serverName = "com.ibm.ws.springboot.support.fat.TemplateServer2";
+        serverPath = installPath + "/usr/servers/" + serverName;
 
         createServerFromTemplate("springBoot2", expectedFeatures);
     }
 
     void createServerFromTemplate(String templateName, String[] expectedFeatures) throws Exception {
+        cleanupServer();
```
sebratton, line 98: I don't get why you need to do this begore and after for this test sice you're using different servernames. But it seems harmless

sebratton (approved):

dazavala:
Thanks for the review, @sebratton. The PB indicates everything works as expected. Before merging I investigated a couple concerns regarding these changes: 

1. The test framework does not collect server content for servers created using simplicity's liberty server utils — i.e. deleting these test servers after each test does not change the collected test content; it only removes the server from `wlp/usr/servers` as intended. Good.

2. The test framework does not fail a test when its server cannot be deleted due to file locking (ref: `LibertyFileManager.deleteLibertyDirectoryAndContents(machine, serverPath);`). Worst case the test inadvertently litters the locked server in `wlp/usr/servers`. Before these changes the test always littered 'TemplateServer` after the second test. I thought we should fix that.

dev/com.ibm.ws.springboot.support_fat/fat/src/com/ibm/ws/springboot/support/fat/TemplateTests.java
dazavala: Sam, before these changes the test left `TemplateServer` in `wlp/usr/servers` after `testCreateServerFromSpringBoot2`.  I thought we should fix that.

PR #21996, opened 2022-08-05
Update MP Actions LATEST version to 5.0

Azquelt:
MP 5.0 is out, so this should be updated.

Originally part of #21903


Azquelt:
We got rid of latest in #22226

PR #22035, opened 2022-08-09
Log a warning when updated value of config property might not be used during restore

anjumfatima90:
for issue https://github.com/OpenLiberty/open-liberty/issues/21307


anjumfatima90:
#build

posmith:
L2 message review completed

dev/io.openliberty.io.smallrye.config/resources/io/openliberty/io/smallrye/config/resources/SmallryeConfigMessages.nlsprops
```diff
@@ -0,0 +1,26 @@
...
+#COMPONENTNAMEFOR CWWKC Smallrye Config
+
+# Message numbers CWWKC0651-700 are reserved for this file.
+# NLS_MESSAGEFORMAT_VAR                
+# MSG_DESCRIPTIVE_NAME_CWWKC0000=CWWKC0000I: This is a message with inserts {0}
+# MSG_DESCRIPTIVE_NAME_CWWKC0000.explanation=Explanation text for the message
+# MSG_DESCRIPTIVE_NAME_CWWKC0000.useraction=User action text for the message
+
+WARNING_UPDATED_CONFIG_PROPERTY_NOT_USED_CWWKC0651=CWWKC0651W: The MicroProfile configuration value for the key {0} has changed since the checkpoint action completed on the server. The updated value might not be used by the application.
+WARNING_UPDATED_CONFIG_PROPERTY_NOT_USED_CWWKC0651.explanation=The configuration value was accessed during a checkpoint action and updated before the server process got restored. Depending on how the application accesses the configuration the updated value might not be used.
```
hiltoncv, line 25: "Depending on how the application accesses the configuration the updated value might not be used."
Can we have an example for this use case? If yes, please provide at least one example to include in the message!

tjwatson: The example message will look like this:

```
[WARNING ] CWWKC0651W: The MicroProfile configuration value for the key test_key has changed since the checkpoint action completed on the server. The updated value might not be used by the application.
```

The example application class that would trigger such a message would look like this:

```java
package mpapp1;

import static org.junit.Assert.assertEquals;

import javax.enterprise.context.ApplicationScoped;
import javax.enterprise.context.Initialized;
import javax.enterprise.event.Observes;
import javax.inject.Inject;

import org.eclipse.microprofile.config.inject.ConfigProperty;

@ApplicationScoped
public class ApplicationScopedOnCheckpointBean {

    @Inject
    @ConfigProperty(name = "test_key", defaultValue = "annoValue")
    String testKey;

    public void applicationScopedValueTest() {
        // The value of test_key was not updated here because the property was accessed during the checkpoint.
        // A warning CWWKC0651W is thrown to let the user know that the updated value of the property might not be used during restore.
        check("defaultValue");
    }

    public void observeInit(@Observes @Initialized(ApplicationScoped.class) Object event) {
        System.out.println(getClass() + ": " + "Initializing application context");
        check("defaultValue");
    }

    private void check(String expected) {
        assertEquals("Wrong value for test key.", expected, testKey);
    }
}
```

Here the warning is making reference to the `test_key` that gets used by the field `testKey` that gets injected as a configuration property.  This particular application class has an `@ApplicationScope` which allows it to be created and injected before a checkpoint happens.  Here we are forcing that by having the bean observe the initialization of the ApplicationScope which forces CDI to aggressively create this bean as the application is starting.  When this happens the `testKey` will get injected with the current value set in config at the time the checkpoint occurs.  Now we can then change the value of the configuration for the `test_key` and restore the server.  At this point the server process will be restored, but this application scope bean instance will not get the updated value.

Thinking through that makes me think the message can be improved by indicating that we detected that the `test_key` was used and the value was read by the application before the checkpoint.  Maybe something like this:

```
CWWKC0651W: The MicroProfile configuration value for the key {0} has been used by an application before a checkpoint action.  The value of the key has changed since the checkpoint action completed on the server and the updated value might not be used by the application when the server process is restored.
```

Note that we say "might" because some applications may use something called [dynamic config](https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#_dynamic_configsource).  In that case the application will always get the most up to date value for the config key.

Hope that helps.

anjumfatima90:
#build

tjwatson:
@hiltoncv do you need some more information from us on this one in order to continue a review?

hiltoncv:
> @hiltoncv do you need some more information from us on this one in order to continue a review?
As discussed please update the message with this line
“If the value of the key changes after the checkpoint action, the application might not use the updated value.”

tjwatson (approved):
Looks good to me.  We should get @tevans78 or @Emily-Jiang to have a look.

anjumfatima90:
#build

anjumfatima90:
#build

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigSourceWrapper.java
```diff
@@ -0,0 +1,156 @@
...
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.websphere.ras.annotation.Trivial;
+
+import io.openliberty.checkpoint.spi.CheckpointHook;
+import io.openliberty.checkpoint.spi.CheckpointPhase;
+
+public class ConfigSourceWrapper implements ConfigSource {
```
tevans78, line 38: lots more javadoc and comments please

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigBuilder.java
```diff
@@ -55,20 +56,30 @@ protected List<ConfigSource> getDefaultSources() {
         defaultSources.add(new AppPropertyConfigSource());
         defaultSources.add(new ServerXMLVariableConfigSource());
         defaultSources.add(new ServerXMLDefaultVariableConfigSource());
-
+        CheckpointPhase phase = CheckpointPhase.getPhase();
```
tevans78, line 63: more comments

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigExtension.java
```diff
@@ -10,15 +10,59 @@
...
 import org.osgi.service.component.annotations.ConfigurationPolicy;
 
 import com.ibm.ws.cdi.extension.WebSphereCDIExtension;
 
+import io.openliberty.checkpoint.spi.CheckpointPhase;
 import io.smallrye.config.inject.ConfigExtension;
 
 @Component(service = WebSphereCDIExtension.class, configurationPolicy = ConfigurationPolicy.IGNORE, property = { "service.vendor=IBM" }, immediate = true)
 public class OLSmallRyeConfigExtension extends ConfigExtension implements Extension, WebSphereCDIExtension {
+    static final CheckpointPhase phase = CheckpointPhase.getPhase();
```
tevans78: I think this should be private?

tevans78: again, class needs lots of javadoc & comments

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigExtension.java
```diff
@@ -10,15 +10,59 @@
...
+    protected void validate(@Observes AfterDeploymentValidation adv) {
+        boolean stopped = stopRecordingReads();
+        try {
+            super.validate(adv);
+        } finally {
+            startRecordingReads(stopped);
+        }
+    }
+
+    static void startRecordingReads(boolean stopped) {
```
tevans78: these method names are confusing ... a method called "start" that takes a parameter called "stopped" which actually sets a thread local called "recordingStopped" to false?

tjwatson: An alternative would be to have `stopRecordingReads` to return an `AutoCloseable` and the `close()` method only resumes the reads if `stopRecordingReads` actually stopped them.  Then it would be code like this:

```java
    try (AutoCloseable stoppedRecording = stopRecordingReads()){
        // do stuff while not recording reads
    }
```

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigExtension.java
```diff
@@ -10,15 +10,59 @@
...
+        if (!stopped || phase == null || phase.restored()) {
+            return;
+        }
```
Azquelt: Having to check whether `phase` is `null` everywhere is unfortunate. Not sure whether the design of `CheckpointPhase.phase()` could be changed there.

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigExtension.java
Azquelt: I think this would be better. Maybe rename as `pauseRecordingReads()` or `withReadRecordingPaused()` if the design is that recording will always be resumed.

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigSourceWrapper.java
```diff
@@ -0,0 +1,156 @@
...
+    private static final TraceComponent tc = Tr.register(ConfigSourceWrapper.class);
+    private final ConfigSource source;
+    private final Map<String, String> propertiesAccessedOnCheckpoint;
+    private final CheckpointPhase phase;
+    private final AtomicBoolean hookAdded = new AtomicBoolean();
+
+    public ConfigSourceWrapper(ConfigSource source, CheckpointPhase phase) {
+        this.source = source;
+        propertiesAccessedOnCheckpoint = new HashMap<>();
```
Azquelt, line 50: Should this be a thread-safe map?

I think it's unlikely that a config source will be accessed from multiple threads during startup but I'm not sure we can guarantee it.

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigExtension.java
tjwatson: Originally I wanted `CheckpointPhase.phase()` to return an `Optional<CheckpointPhase>`, but as of now that project is still compiled to Java 7 because it contains the code that reports that Java 7 is not supported when trying to run on Java 7.

We could consider making `CheckpointPhase.phase()` never return null, but then we are in the awkward spot of figuring out the enum constant name for the `CheckpointPhase` that means no checkpoint is actually going on.

Azquelt: I guess it depends what else you're doing with `CheckpointPhase` elsewhere. Here it seems like it would be easy to have a phase named `INACTIVE` where `.restored()` always returns `true`, but I don't know what other considerations you have elsewhere in the code.

I did have a brief look in `CheckpointPhase` and it sounded like the phase may initially be `null` and then change to an actual value so it sounds like `null` can either be "we're not checkpointing" or "we haven't started thinking about checkpointing yet"?

tjwatson: The intention is that `CheckpointPhase.phase()` will return a constant value.  Once it rerturns null it must always return null.  Down in boot strap I do see a window where that may not be true.  The code probably should throw an IllegalStateException from `phase()` if `CheckpointPhase.phaseSet==false`.

I can look at changing that to return an INACTIVE enum.  But I don't want to hold up this PR waiting for that to happen.

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigExtension.java
anjumfatima90: Made phase as private and added javadoc and comments.

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/OLSmallRyeConfigExtension.java
anjumfatima90: Renamed method. Thanks!

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigSourceWrapper.java
anjumfatima90: Added synchronization while doing read/write into the hashmap.

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigSourceWrapper.java
anjumfatima90: Added comments, thanks!

dev/io.openliberty.io.smallrye.config/src/io/openliberty/microprofile/config/internal/extension/ConfigSourceWrapper.java
Azquelt: Hmm, it looks right but if we're going to do it that way, can we add a javadoc comment to the field stating that reads and writes to the map must synchronize on the map?

anjumfatima90: Added the javadoc comment. Thanks!

anjumfatima90:
#build

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
```diff
@@ -52,68 +62,117 @@ public static void copyAppToDropins() throws Exception {
...
                              server -> {
                                  assertNotNull("'SRVE0169I: Loading Web Module: " + APP_NAME + "' message not found in log before rerstore",
                                                server.waitForStringInLogUsingMark("SRVE0169I: .*" + APP_NAME, 0));
                                  assertNotNull("'CWWKZ0001I: Application " + APP_NAME + " started' message not found in log.",
                                                server.waitForStringInLogUsingMark("CWWKZ0001I: .*" + APP_NAME, 0));
-                                 configureBeforeRestore(testMethod);
+                                 configureBeforeRestore();
                              });
         server.startServer(getTestMethod(TestMethod.class, testName) + ".log");
     }
```
Azquelt, line 76: It's fairly difficult to work out what's going on in this test, you need to check several different places for each test method to find out how it's set up and what it's trying to do. It would be good to find some way to completely refactor this test so that you can see for an individual test method what environment variables are set before and after checkpoint, what log messages are expected and what the application code is doing in one place rather than split out all over the place.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
```diff
@@ -52,68 +62,117 @@ public static void copyAppToDropins() throws Exception {
...
+            case configObjectAppScopeEnvValueChangeTest:
+                configureEnvVariable("config_object_app_scope_key", "envValue");
+                break;
+            case configObjectPropertiesAppScopeEnvValueChangeTest:
+                configureEnvVariable("config_object_properties_app_scope_key", "envValue");
+                break;
+            default:
+                break;
+        }
+    }
```
Azquelt, line 95: This confuses me. You have several tests with "change" in the name where the environment variable is set to one value before checkpoint and a different value afterwards.

However, for all the other tests, you're not setting the value before checkpoint, you're only setting it afterwards. I.e. the value still changes from unset to set.

You don't appear to be doing any tests where the value is the same before checkpoint and on restore?

tjwatson:
The latest build has failures in com.ibm.ws.microprofile.openapi_fat FAT tests which appear to be an existing RTC defect 292234

There is a failure in that needs investigation because the error is happening within the code that has changed in the builder during the execution of the io.openliberty.microprofile.config.2.0.internal_fat tests.  This should NOT be using any of the new wrappers or anything during this testcase, but we need to confirm that the changes didn't cause this:

```
Stack Dump = java.util.ServiceConfigurationError: org.eclipse.microprofile.config.spi.Converter: Error locating configuration files
>	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:582)
>	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.nextProviderClass(ServiceLoader.java:1199)
>	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1221)
>	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1265)
>	at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1300)
>	at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1385)
>	at io.smallrye.config.SmallRyeConfigBuilder.discoverConverters(SmallRyeConfigBuilder.java:119)
>	at io.smallrye.config.SmallRyeConfig.buildConverters(SmallRyeConfig.java:77)
>	at io.smallrye.config.SmallRyeConfig.<init>(SmallRyeConfig.java:69)
>	at io.smallrye.config.SmallRyeConfigBuilder.build(SmallRyeConfigBuilder.java:402)
>	at io.openliberty.microprofile.config.internal.extension.OLSmallRyeConfigBuilder.build(OLSmallRyeConfigBuilder.java:78)
>	at io.openliberty.microprofile.config.internal.extension.OLSmallRyeConfigBuilder.build(OLSmallRyeConfigBuilder.java:35)
>	at io.openliberty.microprofile.config.internal_fat.apps.classLoader.ClassLoadersTestServlet.testUserLoaderErrorsConfig(ClassLoadersTestServlet.java:36)
>	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
>	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>	at componenttest.app.FATServlet.doGet(FATServlet.java:71)
>	at javax.servlet.http.HttpServlet.service(HttpServlet.java:686)
>	at javax.servlet.http.HttpServlet.service(HttpServlet.java:791)
>	at com.ibm.ws.webcontainer.servlet.ServletWrapper.service(ServletWrapper.java:1258)
>	at com.ibm.ws.webcontainer.servlet.ServletWrapper.handleRequest(ServletWrapper.java:746)
>	at com.ibm.ws.webcontainer.servlet.ServletWrapper.handleRequest(ServletWrapper.java:443)
>	at com.ibm.ws.webcontainer.filter.WebAppFilterManager.invokeFilters(WebAppFilterManager.java:1227)
>	at com.ibm.ws.webcontainer.webapp.WebApp.handleRequest(WebApp.java:5059)
>	at com.ibm.ws.webcontainer.osgi.DynamicVirtualHost$2.handleRequest(DynamicVirtualHost.java:316)
>	at com.ibm.ws.webcontainer.WebContainer.handleRequest(WebContainer.java:1007)
>	at com.ibm.ws.webcontainer.osgi.DynamicVirtualHost$2.run(DynamicVirtualHost.java:281)
>	at com.ibm.ws.http.dispatcher.internal.channel.HttpDispatcherLink$TaskWrapper.run(HttpDispatcherLink.java:1199)
>	at com.ibm.ws.http.dispatcher.internal.channel.HttpDispatcherLink.wrapHandlerAndExecute(HttpDispatcherLink.java:468)
>	at com.ibm.ws.http.dispatcher.internal.channel.HttpDispatcherLink.ready(HttpDispatcherLink.java:427)
>	at com.ibm.ws.http.channel.internal.inbound.HttpInboundLink.handleDiscrimination(HttpInboundLink.java:566)
>	at com.ibm.ws.http.channel.internal.inbound.HttpInboundLink.handleNewRequest(HttpInboundLink.java:500)
>	at com.ibm.ws.http.channel.internal.inbound.HttpInboundLink.processRequest(HttpInboundLink.java:360)
>	at com.ibm.ws.http.channel.internal.inbound.HttpInboundLink.ready(HttpInboundLink.java:327)
>	at com.ibm.ws.tcpchannel.internal.NewConnectionInitialReadCallback.sendToDiscriminators(NewConnectionInitialReadCallback.java:167)
>	at com.ibm.ws.tcpchannel.internal.NewConnectionInitialReadCallback.complete(NewConnectionInitialReadCallback.java:75)
>	at com.ibm.ws.tcpchannel.internal.WorkQueueManager.requestComplete(WorkQueueManager.java:514)
>	at com.ibm.ws.tcpchannel.internal.WorkQueueManager.attemptIO(WorkQueueManager.java:584)
>	at com.ibm.ws.tcpchannel.internal.WorkQueueManager.workerRun(WorkQueueManager.java:968)
>	at com.ibm.ws.tcpchannel.internal.WorkQueueManager$Worker.run(WorkQueueManager.java:1057)
>	at com.ibm.ws.threading.internal.ExecutorServiceImpl$RunnableWrapper.run(ExecutorServiceImpl.java:245)
>	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
>	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
>	at java.base/java.lang.Thread.run(Thread.java:866)
>Caused by: java.io.IOException
>	at io.openliberty.microprofile.config.internal_fat.apps.classLoader.CustomClassLoaderError.getResources(CustomClassLoaderError.java:25)
>	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.nextProviderClass(ServiceLoader.java:1196)
]
	at componenttest.custom.junit.runner.FATRunner.blowup(FATRunner.java:334)
	at componenttest.custom.junit.runner.FATRunner.access$600(FATRunner.java:67)
	at componenttest.custom.junit.runner.FATRunner$1.evaluate(FATRunner.java:284)
	at componenttest.rules.repeater.RepeatTests$CompositeRepeatTestActionStatement.evaluate(RepeatTests.java:143)
	at componenttest.custom.junit.runner.FATRunner$2.evaluate(FATRunner.java:352)
	at componenttest.custom.junit.runner.FATRunner.run(FATRunner.java:173)
```

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
Azquelt: Since I'm not expecting to maintain this code, you can leave it as is. If I was maintaining this, I would insist it was rewritten.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
tjwatson: > This confuses me. You have several tests with "change" in the name where the environment variable is set to one value before checkpoint and a different value afterwards.
> 
> However, for all the other tests, you're not setting the value before checkpoint, you're only setting it afterwards. I.e. the value still changes from unset to set.

From what I understand of the test all keys used should have some default value in the server.xml that is changed before restore by using a config source that will override the `defaultValue` from the server.xml.

> 
> You don't appear to be doing any tests where the value is the same before checkpoint and on restore?

We can certainly add something like that to make sure things work when you don't change the value.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
tjwatson: I'm not sure about the need to "completely refactor".  But we could place the logic within the enum TestMethod so that each named test method has the configuration built into it for the pre-checkpoint, pre-restore, post-log checks etc.  Perhaps that will help keep the details of each test more together.  Anjum can open a follow on issue to improve the tests.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
tjwatson: > We can certainly add something like that to make sure things work when you don't change the value.

Actually we already have that with the tests:

 defaultValueTest:
 appScopeDefaultValueTest:
 configObjectAppScopeDefaultValueTest:
 configObjectPropertiesAppScopeDefaultValueTest:

Azquelt:
It would have been good to see a breakdown of what your tests are supposed to cover, given how difficult it is to extract that information from the code. Here's what I think you have:

On one side of your test matrix, you're doing the config lookup in several different ways. Each option uses a different bean and test prefix
* `<no prefix>`, `MPConfigBean`
  * Injection with `@ConfigProperty` into a `RequestScoped` bean
  * It's expected that the value will not be read before checkpoint
* `appScope`, `MPConfigBeanWithApplicationScope`
  * Injection with `@ConfigProperty` into an `ApplicationScoped` bean
  * It's expected that the value will not be read before checkpoint
* `configObjectAppScope`, `ApplicationScopedOnCheckpointBeanWithConfigObject`
  * Injection of `Config` object into an `ApplicationScoped` bean, properties looked up using `getOptionalValue()`
  * Value is read before checkpoint with `@Initialized(ApplicationScoped)`
* `configObjectPropertiesAppScope`, `ApplicationScopedOnCheckpointBeanWithConfigObjectProperties`
  * Injection of `Config` object into an `ApplicationScoped` bean, properties looked up using `getConfigSources()` and calling `.getProperties().get()` on each config source
  * Value is read before checkpoint with `@Initialized(ApplicationScoped)`

On another side of your test matrix, you're changing what type of variable is set and when it's set:
* `defaultValue`
  * The test property is never set
  * The value is expected to be found in the server.xml from the `<variable>` element with a default value
* `annoValue`
  * The test property is not set before checkpoint and the `<variable>` element is removed from server.xml after checkpoint
    * Readability note: It wasn't at all obvious that the test calling `removeVariableConfig()` is not the same as not setting it
    * Before checkpoint the value is expected to be found in the server.xml from the `<variable>` element with a default value
    * After checkpoint the value is expected to be unset
  * In tests where values are injected with `@ConfigProperty`, there's a default value set in the annotation
  * In tests where values are looked up from the `Config` object, the test looks for the value to be unset
  * Not checked for `configObjectPropertiesAppScope`
* `envValue`
  * The test property is set as an environment variable. It's not set before checkpoint and is set on restore.
    * Before checkpoint the value is expected to be found in the server.xml from the `<variable>` element with a default value
    * After checkpoint the value is expected to be found from the environment variable
* `serverValue`
  * The test property is set as a `<variable>` element in server.xml with a value
    * Before checkpoint the value is expected to be found in the server.xml from the `<variable>` element with a default value
    * After checkpoint the value is expected to be found in the server.xml from the new `<variable>` element with a value

In addition you've got one random test: `applicationScopedValueTest`
* Uses `ApplicationScopedOnCheckpointBean`
  * Injection with `@ConfigProperty` into an `ApplicationScoped` bean
  * Value is read before checkpoint with `@Initialized(ApplicationScoped)`
* The test property is set as a `<variable>` element in server.xml with a value
  * Before checkpoint the value is expected to be found in the server.xml from the `<variable>` element with a default value
  * After checkpoint the value is expected to be found in the server.xml from the new `<variable>` element with a value

Azquelt:
Looking at the test coverage above, it's fairly good:

* You've tested values being added and removed as well as being changed and not updated.
* You've tested property injection, lookup from a `Config` object and lookup directly from `ConfigSource` objects.

However, there are a few things I'm less sure about.

The tests using `MPConfigBean` and `MPConfigBeanWithApplicationScope` are probably overkill since all of these tests are just "test lookup from a bean which isn't instantiated before checkpoint".

It's not clear why `ApplicationScopedOnCheckpointBean` only has one test while `ApplicationScopedOnCheckpointBeanWithConfigObject` has several. I suspect that injecting properties with `@ConfigProperty` is the more common case.

There are two tests commented out - this looks like a bug, is there an issue raised?
* `configObjectPropertiesAppScopeEnvValueTest`
* `configObjectPropertiesAppScopeEnvValueChangeTest`

It surprises me if these tests have problems but `configObjectAppScopeEnvValueTest` and `configObjectAppScopeEnvValueChangeTest` do not so I'd like to find out why that appears to be the case.

I know this test is about logging a warning, but the tests added also appear to test the correctness of config when used with checkpoint. If so, here are some more things we should test:
* Can we test a value being not present before checkpoint and then added afterwards?
  * In your current tests there is always a value set before checkpoint in the server.xml variable defaultValue
  * I would expect this to be fairly common for people who are setting optional values defined in MP specs
* Similarly, testing a value set in a config source before checkpoint, and then changing that value in that config source after checkpoint
  * Note that variable values and variable default values are separate config sources
  * I'd expect this to be fairly common too if a docker container has ENV variables declared which can be overridden when the container is started.
* If I inject `@ConfigProperty Producer<String>`, and read it pre-checkpoint, do I get an updated value post-checkpoint?
  * If users want to inject a property but their bean gets injected before checkpoint, this is how I'd expect them to work around that.
  * Test using an environment variable because in the non-checkpoint scenario we never expect it to change
* Similarly, could we test injecting `Config` and reading a value before and after checkpoint and checking that it changes
  * Again, test with environment variable

Azquelt:
I imagine these tests are quite slow to run since the server is checkpointed and started for each test. Should many of these tests be annotated with `@Mode(FULL)`?

I would suggest keeping all of the tests that use `ApplicationScopedOnCheckpointBeanWithConfigObject` in LITE and putting all but one method of each of the other beans into FULL mode.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
Azquelt: Thanks. After being confused by this, I decided I needed to write out the whole test matrix to see exactly what was an wasn't tested. See my new comments.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/MPConfigTest.java
Azquelt: Given that you're starting and stopping the server for each test, you could just have a method for each test which does the configuration, starts the server, checkpoints, calls the test servlet, stops the server and checks the logs.

dev/io.openliberty.checkpoint_fat/test-applications/mpapp1/src/mpapp1/ApplicationScopedOnCheckpointBeanWithConfigObject.java
```diff
@@ -22,13 +22,14 @@
 @ApplicationScoped
 public class ApplicationScopedOnCheckpointBeanWithConfigObject {
 
+    @Inject
+    Config config;
+
     public void observeInit(@Observes @Initialized(ApplicationScoped.class) Object event) {
         System.out.println(getClass() + ": " + "Initializing application context");
+        check("defaultValue");
     }
```
Azquelt, line 31: I don't think anything checks that this `check()` call doesn't throw an error.

If it did, I think it would fail in `configObjectAppScopeEnvValueChangeTest`

dev/io.openliberty.checkpoint_fat/test-applications/mpapp1/src/mpapp1/ApplicationScopedOnCheckpointBeanWithConfigObject.java
```diff
@@ -50,7 +51,14 @@ public void appScopeAnnoValueTest() {
...
+        //Get optional value
+        String optionalValue = config.getOptionalValue("config_object_app_scope_key", String.class).orElse("annoValue");
+        assertEquals("Wrong value for test key.", expected, optionalValue);
+
+        if (!expected.equals("annoValue")) {
+            // Get value
+            String value = config.getValue("config_object_app_scope_key", String.class);
+            assertEquals("Wrong value for test key.", expected, value);
+        }
     }
```
Azquelt, line 63: As a style point, if you have a utility method which does a non-trivial assertion like this, it's easier to read if:
* the method name starts with `assert`
* the method name includes a description of what it tests

Something like `assertConfigValueEquals(String expected)` might be better here.

Sometimes it's clearer just to return a value and do the assert in the test, but you can't do that here as you're doing several asserts in your `check()` method.

dev/io.openliberty.checkpoint_fat/test-applications/mpapp1/src/mpapp1/ApplicationScopedOnCheckpointBeanWithConfigObject.java
Azquelt: It might be that you're not fussed about actually checking what the value is at this point. In that case you could read the value from config manually rather than calling `check()`.

The problem is that it looks like there's a check being done here, but I don't think there really is.

Azquelt (commented):
I've finished my review.

Although I don't much like the way the test is written because I find it hard to read, it's your test not mine so you're welcome to ignore those comments.

You've already addressed my comments on the implementation code and it's now fine.

I would like a response to my comments on the test coverage, suggested additional tests, the commented out tests, and the split between FULL and LITE mode, but that work doesn't have to be done in this PR as long as there are other issues raised to cover it and ensure it does get done.

anjumfatima90:
I have opened up another issue to work on the tests https://github.com/OpenLiberty/open-liberty/issues/22224

Azquelt:
Thanks @anjumfatima90 

Do you know what's up with these two tests which are commented out?

* `configObjectPropertiesAppScopeEnvValueTest`
* `configObjectPropertiesAppScopeEnvValueChangeTest`

anjumfatima90:
> Thanks @anjumfatima90
> 
> Do you know what's up with these two tests which are commented out?
> 
> * `configObjectPropertiesAppScopeEnvValueTest`
> * `configObjectPropertiesAppScopeEnvValueChangeTest`

I fixed the tests to get the highest ordinal non null value. Earlier they were failing because I wasn't checking for ordinal and there were two non null values in case of these two tests. Since Environment variables have higher ordinal than the optional default value. We need to check for higher ordinal value.

anjumfatima90:
#build

Azquelt:
> I fixed the tests to get the highest ordinal non null value. Earlier they were failing because I wasn't checking for ordinal and there were two non null values in case of these two tests. Since Environment variables have higher ordinal than the optional default value. We need to check for higher ordinal value.

Thanks, that makes sense.

Azquelt (approved):

tjwatson:
#libby

PR #22051, opened 2022-08-10
xmlBinding-3.0: Change javax.xml.catalog to optional import 

neuwerk:
Since the javax.xml.catalog APIs are only available on Oracle JDKs greater than 9, the `javax.xml.catalog` APIs won't always be available for the feature to import.


neuwerk:
#build

PR #22062, opened 2022-08-11
Update abandon test to remove debug for now

dmbelina:
The one notable change is to an abandon test in BatchJobOperatorApiWithAppSecurityTest. There's a test that issues an abandon, and rarely, our test servlet call issues the abandon call twice. This causes the test to find a failure, because as-designed you can't abandon a job that has already been abandoned. I had previously put some additional tracing in and left the failure in for debugging, but at this point won't get to the debugging any time soon. So for now I'll take out the debugging extras and allow the test to pass if it sees the abandon when already abandoned error. And re-throw the error if its some other real problem.


cgianfra (approved):

dmbelina:
Personal build: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_dhuk0BlzEe2h_4Zec1JglA

PR #22068, opened 2022-08-11
Created a repeated test for LocalEJBTest class

KeenanMcCarthy:
Added a class rule to run the EJB tests as both Java 8 and Java 9


KeenanMcCarthy:
#libby
#build

tjwatson (approved):
LGTM

PR #22087, opened 2022-08-12
Issue #20416: Add test for getRealPath clarification

### LINKED ISSUE #20416: Servlet 6.0: Clarify the behaviour of getRealPath(String).
Spec issue: `https://github.com/eclipse-ee4j/servlet-api/issues/105`

See "Appendix A: Change Log" here: https://github.com/eclipse-ee4j/servlet-api/blob/master/spec/src/main/asciidoc/servlet-spec-body.adoc

ServletContext Javadoc to review: https://jakarta.ee/specifications/servlet/6.0/apidocs/jakarta.servlet/jakarta/servlet/servletcontext

> getRealPath
> 
>     [String](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html?is-external=true) getRealPath​([String](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html?is-external=true) path)
> 
>     Gets the real path corresponding to the given virtual path.
> 
>     **The path should begin with a / and is interpreted as relative to the current context root. If the path does not begin with a /, the container will behave as if the method was called with / appended to the beginning of the provided path.**
> 
>     For example, if path is equal to /index.html, this method will return the absolute file path on the server's filesystem to which a request of the form http://<host>:<port>/<contextPath>/index.html would be mapped, where <contextPath> corresponds to the context path of this ServletContext.
> 
>     The real path returned will be in a form appropriate to the computer and operating system on which the servlet container is running, including the proper path separators.
> 
>     Resources inside the /META-INF/resources directories of JAR files bundled in the application's /WEB-INF/lib directory must be considered only if the container has unpacked them from their containing JAR file, in which case the path to the unpacked location must be returned.
> 
>     This method returns null if the servlet container is unable to translate the given virtual path to a real path.
> 
>     Parameters:
>         path - the virtual path to be translated to a real path
>     Returns:
>         the real path, or null if the translation cannot be performed

At a glance I think we already behave the way: https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.webcontainer/src/com/ibm/ws/webcontainer/osgi/webapp/WebApp.java#L695

**The path should begin with a / and is interpreted as relative to the current context root. If the path does not begin with a /, the container will behave as if the method was called with / appended to the beginning of the provided path.**

Todo: Do we have any tests that verify this? If not then one should be added to make sure everything is correct.

pnicolucci:
fixes #20416

### DIFF
--- a/dev/io.openliberty.webcontainer.servlet.6.0.internal_fat/.classpath
+++ b/dev/io.openliberty.webcontainer.servlet.6.0.internal_fat/.classpath
@@ -2,6 +2,7 @@
 <classpath>
     <classpathentry kind="src" path="fat/src"/>
     <classpathentry kind="src" path="test-applications/SimpleTest.war/src"/>
+    <classpathentry kind="src" path="test-applications/GetRealPathTest.war/src"/>
     <classpathentry kind="con" path="aQute.bnd.classpath.container"/>
     <classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
     <classpathentry kind="output" path="bin"/>
--- a/dev/io.openliberty.webcontainer.servlet.6.0.internal_fat/bnd.bnd
+++ b/dev/io.openliberty.webcontainer.servlet.6.0.internal_fat/bnd.bnd
@@ -13,7 +13,8 @@ bVersion=1.0
 
 src: \
     fat/src,\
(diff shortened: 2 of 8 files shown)


pnicolucci:
#libby #build

pnicolucci:
#build

pnicolucci:
#build

volosied (approved):
Looks good to me

pnicolucci:
#libby

PR #22098, opened 2022-08-12
Disable ConcurrentErrorTest app validation

KyleAure:
When error testing we attempt to install an application that we expect to fail with an error, and therefore need to disable the validation that expects the application to be successfully installed.


njr-11 (approved):

KyleAure:
#build

Issue #21968, opened 2022-08-04
Create microProfile-6.0 feature based on EE9 and add toleration for EE10 where possible


tevans78:
Create `microProfile-6.0` convenience feature containing...
- cdi-3.0
- jsonb-2.0
- jsonp-2.0
- restfulWS-3.0
- servlet-5.0
- mpCompatible-6.0
- mpConfig-3.0
- mpFaultTolerance-4.0
- mpHealth-4.0
- mpJwt-2.1
- mpOpenAPI-3.1
- mpMetrics-5.0
- mpRestClient-3.0
- mpTelemetry-1.0
(note that the EE9 features above will later be swapped to their EE10 versions)

mpCompatible-6.0 should depend on eeCompatible-10.0 but tolerate 9.0

mpConfig-3.0, mpFaultTolerance-4.0, mpHealth-4.0 and mpRestClient-3.0 should have toleration for EE10 added  (may not be possible yet).

mpJwt-2.1, mpOpenAPI-3.1, mpMetrics-5.0 and mpTelemetry-1.0 should be based on EE10 but tolerate EE9

mpMetrics-5.0 and mpJwt-2.1 do not actually exist yet but can be stubbed out using the old impls

Tasks
- [x] Create mpCompatible-6.0 feature #22024
- [x] Update test framework to allow repeats to be added for MP60 #22024
- [x] Make mpConfig-3.0 tolerate EE10 #22024
- [x] Make mpHealth-4.0 tolerate EE10 #22089
- [x] Make mpOpenAPI-3.1 be based on EE10 and tolerate EE9 #22088 
- [x] Make mpRestClient-3.0 tolerate EE10 #22141 
  - [x] Make mpTelemetry-1.0 be based on EE10 and tolerate EE9 #22190
- [x] Make mpFaultTolerance-4.0 tolerate EE10 #22063
  - [x] Enable mpRestClient-FT integration FATs
  - [x] Create mpMetrics-5.0 stub feature to avoid waiting for the real feature to be done
- [x] Make mpContextPropagation-1.3 tolerate EE10 #22109 
  - [x] Make GraphQL-2.0 tolerate EE10 https://github.com/OpenLiberty/open-liberty/pull/23053
- [x] Create microProfile-6.0 feature https://github.com/OpenLiberty/open-liberty/pull/22144
- [x] Run high level MP tests against microProfile-6.0 https://github.com/OpenLiberty/open-liberty/issues/22201
[...]

PR #22063, opened 2022-08-11
Make mpFaultTolerance 4.0 tolerate MP 6 and EE 10

Azquelt:
mpFaultTolerance-4.0 will be included in both MP 5 and MP 6

For #21968

PR #22109, opened 2022-08-15
Update mpContextPropagation-1.3 to tolerate MP 6

Azquelt:
mpContextPropagation-1.3 is a standalone feature which will be compatible with MicroProfile 6.0

For #21968


Azquelt:
#build

Azquelt:
Previous build failed due to a problem contacting the repository.
#build
#libby

Azquelt:
Noticed that I'd excluded the MP6 features from the EE10CompatibilityTest, but I hadn't excluded the standalone features which are being made compatible with EE10.

#build
#libby

Azquelt:
Rebased on integration
#libby
#build

dev/com.ibm.ws.concurrent.mp.1.3_fat_tck/build.gradle
```diff
@@ -0,0 +1,12 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+addRequiredLibraries.dependsOn addJakartaTransformer
```
njr-11, line 12: Why does this need a Jakarta transformer?  Nothing in the 1.3 TCK should be using Java EE and in need of transformation to Jakarta EE?

Azquelt: The `JakartaEE9Action` tries to load it, even if nothing needs transformed.

njr-11 (changes requested):
The changes for com.ibm.ws.concurrent.mp_fat_jakarta need to be done differently.
This bucket is not meant to do repeats.  Instead, it has a completely different Test class and Test application covering MicroProfile Context Propagation with EE 10.  Tom Evans once disabled here, which had to be done until MicroProfile support for EE 10 was available,

https://github.com/OpenLiberty/open-liberty/commit/c448a955cd91abac663abb8f18380a1f30f120c2

Instead of adding repeats to this bucket, you should re-enable these tests.  We should also rename them to reflect that there isn't any "2.0" and MP Context Propagation remains at 1.3.  In this case, I'll recommend names like "MPContextProp1_3_EE9_Test" and "MPContextProp1_3_EE10_Test" and so forth.

dev/com.ibm.ws.concurrent.mp.1.3_fat_tck/build.gradle
Azquelt: Actually, I might be pointing the finger of blame in the wrong direction there, but when running with the `JakartaEE9Action` or `JakartaEE10Action` (which is used for the MP50 and MP60 repeats), it does try to use the transformer. I think from `ShrinkHelper` when an app is deployed.

njr-11: Would it be better in this case to just duplicate the server.xml for EE 10 features and have a second test class to launch the MP 6/EE 10 version of the TCK?  It seems like that would be pretty minimal and would avoid the possibility of having the transformer around to possibly contaminate things.

Azquelt: Given that we need to test with EE10 using a separate test class and app anyway, I will be removing the repeat and I'll remove this as well.

Azquelt: I misread this comment yesterday as applying to the regular FAT rather than the TCK.

In the TCK, I guess we shouldn't have the same issue with the transformer. I'll re-test without it.

We could still use a separate class and server.xml to run on EE10 if necessary, but then it becomes difficult to tell which run is which in your test results.

Azquelt: I tried taking the transformer library out but at the moment that causes test failures when checking for allowed FFDCs (the checker tries to transform any exception names that need it).

Stack trace:
```
java.lang.NullPointerException: 2022-08-19-12:29:02:982 null
	at componenttest.rules.repeater.EE9PackageReplacementHelper.replacePackages(EE9PackageReplacementHelper.java:72)
	at componenttest.rules.repeater.EE9PackageReplacementHelper.replacePackages(EE9PackageReplacementHelper.java:54)
	at componenttest.custom.junit.runner.FATRunner.getAllowedFFDCAnnotationFromTest(FATRunner.java:733)
	at componenttest.custom.junit.runner.FATRunner.access$400(FATRunner.java:67)
	at componenttest.custom.junit.runner.FATRunner$1.evaluate(FATRunner.java:249)
	at componenttest.rules.repeater.RepeatTests$CompositeRepeatTestActionStatement.evaluate(RepeatTests.java:143)
	at componenttest.custom.junit.runner.FATRunner$2.evaluate(FATRunner.java:352)
	at componenttest.custom.junit.runner.FATRunner.run(FATRunner.java:173)
```

I guess it's a fairly recent problem that we have tests which we want to repeat but don't need to be transformed so I'm asking around about how we should fix it.

In the mean time, we can either keep the repeat and the transformer (though the transformer will not be used to transform any apps), or use a second test class (though we'll need an extra hack in there so that you can tell the two sets of results apart).

Azquelt:
@njr-11 I've updated the FAT project to enable the existing test class written for 2.0 and have renamed everything to either 1_3_EE9 or 1_3_EE10. This is ready for re-review.

For the moment I've left the transformer libraries in the TCK project, see my other comment for details about that. I can have another go at taking it out if you're not happy about it.

Edit: The new tests require Fault Tolerance so I had to base this PR on #22063 which makes mpFaultTolerance-4.0 compatible with EE10. Only the last few commits are really part of this PR, the others will disappear once the fault tolerance PR is merged.

Azquelt:
#libby #build

njr-11 (approved):
Thanks for making all of these updates!

posmith:

Azquelt:
Rebased to resolve merge conflict
#build
#libby

Azquelt:
#build
#libby

Azquelt:
Rebased again to avoid a merge conflict
#build
#libby

Azquelt:
One of the child builds failed to start resulting in not all tests running. I'm trying to re-run it here: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_B6xuACLWEe2b2O1gXXKb4w

Azquelt:
Child build passed, the only tests which didn't pass are known issues.

PR #22110, opened 2022-08-15
com.ibm.ws.ui_rest_fat: Add call to ignore expected error in CatalogPersistenceTest

lavenal:
The tests in CatalogPersistenceTest were running in a different order as written. As a result, the last test ran before tearing down was an error scenario with an invalid server.xml. The tear down method needs to include errors to be ignored when stopping the server.


lavenal:
#build

amyreit (approved):

PR #22117, opened 2022-08-15
align with jnosql removal of stereotype

njr-11:
Stereotype was removed under pull 111 so the copied annotations should do the same.


njr-11:
#build

tabishop (approved):

PR #22152, opened 2022-08-17
Fix Admin Center Explore tool error dialog text layout problem

lavenal:
When the error dialog includes a continuous long string of text without a breaking character in it (eg. hyphen), the text will overflow the content box in the dialog. To fix this problem, add css to break long continuous text in the dialog to avoid overflow.

Without the fix:

image

With the fix:

image


lavenal:
#build

k8vance88 (approved):

PR #22171, opened 2022-08-18
Switch to BatchSpanProcessor

### LINKED ISSUE #22151: mpTelemetry-1.0 Throughput Performance
I am looking into the performance of mpTelemetry-1.0 performance - compared to mpOpentracing-3.0.

I am hitting a very simple REST endpoint with JMeter with 50 threads concurrently.
```
@Path("/")
public class RestServer {
	
  @GET
  @Produces(MediaType.TEXT_PLAIN)
  @Path("/ping")
  public String ping() {
    return "pong";
  }
}
```

Liberty is running on 2 cpus and I also have a a jaeger all-in-one container (https://www.jaegertracing.io/docs/1.37/getting-started/) running on 6 cpus.

Initial performance results are really bad. As a baseline, with mpOpenTracing-3.0 and jaeger client 1.3.1 I can do 18,000 requests per second. With mpOpenTelemetry-1.0 and `-Dotel.traces.exporter=otlp` or `-Dotel.traces.exporter=jaeger`, I can only do 2000 req/sec, and throughput becomes slower and slower, eventually getting below 100 reqs/sec and lower. If I set `-Dotel.traces.exporter=zipkin` throughput is constant at 4000 req/sec, but there are occasional errors sending spans to jaeger.

I have tried moving the io.opentelemetry code up to 1.17 (from 1.14), but that has not seemed to help.

I noticed that open-telemetry creates a okhtt3.Dispatcher that is used to connect to jaeger over grpc. 

https://github.com/open-telemetry/opentelemetry-java/blob/80ee9f4bcb3de7f3e3f98755195d4a5bdfb2c3e3/exporters/common/src/main/java/io/opentelemetry/exporter/internal/okhttp/OkHttpUtil.java#L24

By default, the Dispatcher has a maxRequestsPerHost set to 5. By increasing this to 64, I can get throughput up to over 8000 req/sec, but eventually hit memory problems - running out of heap caused by many of okhttp3 objects, even with a 3GB heap. I am investigating this now, wondering if requests are just getting backed up, as it seems to recover if I take off the load.

I plan to try running this scenario outside of liberty to if I see the same issues.

More data here: https://ibm.box.com/s/mk0mkxf3urhxt62dh8zxrala9vvoj18c

jdmcclur:
It looks like we are using a SimpleSpanProcessor for mpOpentelemetry, which does not seem recommended for performance based on the documentation. Switching to BatchSpanProcessor seems to be much better for performance. Hoping this works.

https://github.com/open-telemetry/opentelemetry-java/blob/734b13e9c5a57d9cf7bc269ad801b1e7a357049f/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/SimpleSpanProcessor.java#L28

fixes #22151


jdmcclur:
#build

fmhwong (approved):
LGTM

Yas-Franco (approved):
Looks good- Thank you Joe

PR #22175, opened 2022-08-18
Skip Concurrency TCK InheritedAPITests

KyleAure:
Need to skip InheritedAPITests because of a defect in the TCK where a singleton EJB was annotated with `@Stateless` which causes conflicting counts. 
See: https://github.com/jakartaee/concurrency/issues/261


njr-11 (approved):
Makes sense to disable this for now given the bug.  It can be re-enabled once 3.0.2 is picked up.

KyleAure:
#build

PR #22192, opened 2022-08-19
Java 19 methods on managed scheduled future

njr-11:
Implement Java 19 methods on the scheduled future for managed scheduled executor tasks.
While adding this, I spotted a bug in the code that was added to policy executor futures and am fixing that as well.  In that case aborted tasks should not be returning a completed task failure exception because the tasks did not run and did not complete.  Instead, it should raise IllegalStateException to follow the exceptionNow JavaDoc.


njr-11:
#build

KyleAure (approved):

njr-11:
#build

njr-11:
#build

PR #22196, opened 2022-08-19
Add an INACTIVE CheckpointPhase

tjwatson:
Add an INACTIVE CheckpointPhase
    
Change CheckpointPhase.getPhase to never return null.
If no checkpoint is being done then getPhase will return
the INACTIVE phase.  Always register the CheckpointPhase
service.  This service is internal so is not beta gaurded from
registration.
    
Updates to all places that checked for null instead of
the INACTIVE phase.

### DIFF
--- a/dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
+++ b/dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
@@ -260,16 +260,21 @@ public class EJBRuntimeImpl extends AbstractEJBRuntime implements ApplicationSta
 
     private final CheckpointPhase checkpointPhase;
 
-    @Activate
-    public EJBRuntimeImpl(@Reference(cardinality = ReferenceCardinality.OPTIONAL) CheckpointPhase checkpointPhase) {
-        if (checkpointPhase != null || !ProductInfo.getBetaEdition()) {
-            this.checkpointPhase = checkpointPhase;
+    public EJBRuntimeImpl() {
+        CheckpointPhase phase = CheckpointPhase.getPhase();
+        if (!ProductInfo.getBetaEdition()) {
+            this.checkpointPhase = phase;
         } else {
-            this.checkpointPhase = CheckpointPhase.getPhase(System.getProperty("io.openliberty.ejb.checkpoint.phase", ""));
+            CheckpointPhase testPhase = CheckpointPhase.getPhase(System.getProperty("io.openliberty.ejb.checkpoint.phase", ""));
+            if (phase == CheckpointPhase.INACTIVE && testPhase != null) {
+                this.checkpointPhase = testPhase;
+            } else {
+                this.checkpointPhase = phase;
+            }
         }
 
         // For any Checkpoint phase, pause all non-persistent timers until checkpoint restored
-        if (checkpointPhase != null && !checkpointPhase.restored()) {
+        if (!checkpointPhase.restored()) {
             TimerNpRunnable.pause();
         }
     }
@@ -281,7 +286,9 @@ public class EJBRuntimeImpl extends AbstractEJBRuntime implements ApplicationSta
                unbind = "ignoreCheckpointRestored")
     protected final void checkpointRestored(ServiceReference<?> checkpoint) {
         // Resume all non-persistent timers on checkpoint restore
-        TimerNpRunnable.resume();
+        if (checkpointPhase != CheckpointPhase.INACTIVE) {
+            TimerNpRunnable.resume();
+        }
     }
 
     protected final void ignoreCheckpointRestored(ServiceReference<?> checkpoint) {
--- a/dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
@@ -20,8 +20,8 @@ import java.util.function.Supplier;
 import com.ibm.websphere.ras.annotation.Trivial;
 
 /**
- * Phase which a checkpoint is being taken. This enum is registered as a
- * service when a checkpoint operation is being done.
+ * Phase which a checkpoint of the running process is being taken.
+ * This enum is registered as a service when running in an OSGi framework.
  *
  */
 public enum CheckpointPhase {
@@ -39,12 +39,21 @@ public enum CheckpointPhase {
      * Phase of startup after the applications have been discovered but no
      * code from the application has been executed.
      */
-    DEPLOYMENT;
+    DEPLOYMENT,
+    /**
+     * Phase that indicates there is no checkpoint being done.
+     */
+    INACTIVE(true);
 
     /**
      * The checkpoint phase service property used to store the checkpoint phase when registered as a service.
      */
     public static final String CHECKPOINT_PROPERTY = "io.openliberty.checkpoint";
+
+    /**
+     * A service filter to lookup an active checkpoint phase.
+     */
+    public static final String CHECKPOINT_ACTIVE_FILTER = "(!(" + CHECKPOINT_PROPERTY + "=INACTIVE))";
 
     /**
      * The checkpoint phase service property used to store if the checkpoint process has been restored.
@@ -66,6 +75,21 @@ public enum CheckpointPhase {
         phases = new HashMap<String, CheckpointPhase>();
         for (CheckpointPhase p : CheckpointPhase.values()) {
             phases.put(p.toString(), p);
+        }
+    }
+
+    /**
+     *
+     */
+    private CheckpointPhase() {
+        this(false);
+    }
+
+    private CheckpointPhase(boolean inactive) {
+        if (inactive) {
+            // if inactive then set restored and noMoreAddHooks to true;
+            restored = true;
+            noMoreAddHooks = true;
         }
     }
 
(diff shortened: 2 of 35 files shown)


tjwatson:
#build

tjwatson:
#build

dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
```diff
@@ -39,13 +39,22 @@ public enum CheckpointPhase {
      * Phase of startup after the applications have been discovered but no
      * code from the application has been executed.
      */
-    DEPLOYMENT;
+    DEPLOYMENT,
+    /**
+     * Phase that indicates there is no checkpoint being done.
+     */
+    INACTIVE(true);
```
Azquelt, line 46: Update Javadoc for `CheckpointPhase` which currently says
> This enum is registered as a service _when a checkpoint operation is being done_.

dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
```diff
@@ -124,7 +149,7 @@ final public boolean restored() {
      *
      * @param hook the hook to be used to prepare and restore
      *                 a checkpoint process.
-     * @throws IllegalStateException if the phase is in progress.
+     * @throws IllegalStateException if the phase is not in progress.
      * @see CheckpointHook#MULTI_THREADED_HOOK
```
Azquelt: This javadoc also needs to be updated to say what the return value means (same for `addMultiThreadedHook()`)

dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
tjwatson: I've rebased to latest integration branch so you comment is not included in the change view.  But I updated the class doc to:

```
Phase which a checkpoint of the running process is being taken. 
This enum is registered as a service when running in an OSGi framework.
```

dev/com.ibm.ws.kernel.boot.core/src/io/openliberty/checkpoint/spi/CheckpointPhase.java
tjwatson: Thanks for pointing that out.  Here is the update I made which is applied to both methods:

```java
    /**
     * Adds a {@code CheckpointHook} for this checkpoint phase.
     * The hook will be run to {@link CheckpointHook#prepare() prepare}
     * and {@link CheckpointHook#restore() restore} a checkpoint
     * process while the JVM is in multi-threaded mode. If the
     * checkpoint process is already {@link #restored() restored}
     * then the hook is not added and {@code false} is returned,
     * indicating that the hook will not be called.
     * A return value of {@code true} indicates that the hook will
     * be called for both {@link CheckpointHook#prepare() prepare}
     * and {@link CheckpointHook#restore() restore}.
     *
     * @param hook the hook to be used to prepare and restore
     *                 a checkpoint process.
     * @throws IllegalStateException if the phase is not in progress.
     * @return true if the hook is successfully added; otherwise false
     *         is returned.
     * @see CheckpointHook#MULTI_THREADED_HOOK
     */
```

tjwatson:
#build

tjwatson:
@tkburroughs I included you to review because you were an early adopter of the original service design.  I changed the EJB code to no longer consume the phase as a service because it simplifies the code a bit.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
```diff
@@ -281,7 +286,9 @@ public EJBRuntimeImpl(@Reference(cardinality = ReferenceCardinality.OPTIONAL) Ch
                unbind = "ignoreCheckpointRestored")
     protected final void checkpointRestored(ServiceReference<?> checkpoint) {
         // Resume all non-persistent timers on checkpoint restore
-        TimerNpRunnable.resume();
+        if (checkpointPhase != CheckpointPhase.INACTIVE) {
```
tkburroughs, line 289: While the code is not wrong, is it unnecessary? Curious why checkpointRestored() would ever be called if INACTIVE?

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
```diff
@@ -260,16 +260,21 @@ private static RuntimeException rethrow(Throwable t) {
...
+            if (phase == CheckpointPhase.INACTIVE && testPhase != null) {
+                this.checkpointPhase = testPhase;
+            } else {
+                this.checkpointPhase = phase;
+            }
         }
 
         // For any Checkpoint phase, pause all non-persistent timers until checkpoint restored
-        if (checkpointPhase != null && !checkpointPhase.restored()) {
+        if (!checkpointPhase.restored()) {
```
tkburroughs, line 277: This is a change in behavior, as previously the code was using the `checkpointPhase` parameter and not `this.checkpointPhase`. Therefore, the code would have ignored the test property (now testPhase).  The issue is that when just using the test property, there is no mechanism to call `checkpointRestored`, so timers will never run.

tjwatson: The check for `phase != CheckpointPhase.INACTIVE` can be thought of as a replacement for `phase != null`.  

So I think the same code could have been:

```java
        CheckpointPhase checkpointPhase = CheckpointPhase.getPhase();
        if (checkpointPhase != CheckpointPhase.INACTIVE || !ProductInfo.getBetaEdition()) {
            this.checkpointPhase = checkpointPhase;
        } else {
            CheckpointPhase testPhase = CheckpointPhase.getPhase(System.getProperty("io.openliberty.ejb.checkpoint.phase", ""));
            this.checkpointPhase = testPhase != null ? testPhase : checkpointPhase;
        }

        // For any Checkpoint phase, pause all non-persistent timers until checkpoint restored
        if (checkpointPhase != CheckpointPhase.INACTIVE && !checkpointPhase.restored()) {
            TimerNpRunnable.pause();
        }
    }
```
I can change it to that, but I think it is equivalent to the code I have in the PR

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tjwatson: We now always register the `CheckpointPhase` that is returned by the static method `CheckpointPhase.getPhase()`.  For the case where there is no `checkpoint` action being done this phase is going to be the `INACTIVE` phase which reports to already be `restored`.  So the target filter will match and this method will get called during activation.

We could instead update the target filter to be:

```
target = "(&(" + CheckpointPhase.CHECKPOINT_RESTORED_PROPERTY + "=true)" +
         "(!(" + CheckpointPhase.CHECKPOINT_PROPERTY + "=INACTIVE)))"
```
That should allow us to call `TimerNpRunnable.resume()` unconditionally here.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tkburroughs: The old behavior would not call `pause()` if checkpoint phase was obtained from the property, rather than the real checkpoint service.  Because `checkpoint` is no longer passed in as a parameter, `checkpoint` is the same as `this.checkpoint`, whereas in the original code it was not.  In the original code, `checkpoint` could have been `null`, whereas `this.checkpoint` may have been APPLICATION (set from the property).

The above code is still not the same behavior as before.... instead, it would need to be something like

```
        boolean checkpointFromProperty = false;
        CheckpointPhase checkpointPhase = CheckpointPhase.getPhase();
        if (checkpointPhase != CheckpointPhase.INACTIVE || !ProductInfo.getBetaEdition()) {
            this.checkpointPhase = checkpointPhase;
        } else {
            checkpointFromProperty = true;
            this.checkpointPhase = CheckpointPhase.getPhase(System.getProperty("io.openliberty.ejb.checkpoint.phase", ""));
        }

        // For any Checkpoint phase, pause all non-persistent timers until checkpoint restored
        if (!checkpointFromProperty && !checkpointPhase.restored()) {
            TimerNpRunnable.pause();
        }
    }
```

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tkburroughs: Ah, got it; the match is on the `restored` setting in the phase.  Now that I understand, I'm fine with the code as is; or with the target filter change.

dev/com.ibm.ws.ejbcontainer/src/com/ibm/ws/ejbcontainer/osgi/internal/EJBRuntimeImpl.java
tjwatson: Ah, now I get it.  But with the change to how `INACTIVE` is always registered then we always do call `checkpointRestored` and this constructor would have set `this.checkpointPhase` to the phase specified by the property which the tests would have set it to FEATURES, DEPLOYMENT or APPLICATIONS so the check for INACTIVE would pass making it call the `TimerNpRunnable.resume()`

This is a bit obscure.  I'll try to make this more straightforward.

sebratton (approved):

tjwatson:
#libby

PR #22226, opened 2022-08-22
Update MicroProfile FATs

tevans78:
Remove MicroProfileActions.LATEST and the repeatAll methods
Allow existing basic MP FATs to be repeated against MP6
Add a new MP6 FAT bucket

#build


tevans78:
#build

dev/io.openliberty.microprofile.config.2.0.internal_fat/fat/src/io/openliberty/microprofile/config/internal_fat/Config20ExceptionTests.java
```diff
@@ -49,7 +48,7 @@ public class Config20ExceptionTests extends FATServletClient {
     public static final String SERVER_NAME = "Config20ExceptionServer";
 
     @ClassRule
-    public static RepeatTests r = MicroProfileActions.repeat(SERVER_NAME, MicroProfileActions.LATEST, MicroProfileActions.MP50);
```
Azquelt, line 52: Do we want these config tests to run against MP60 now?

dev/io.openliberty.microprofile6.internal_fat/bnd.bnd
```diff
@@ -0,0 +1,31 @@
+#*******************************************************************************
+# Copyright (c) 2019, 2021 IBM Corporation and others.
```
Azquelt: I assume the files in this project have been mostly copied unchanged which is why the copyright dates are in the past?

I'm guessing this file has changes to update the java level to 11 so I think it's date should have changed. Were there any other files in this project which also needed changes?

dev/io.openliberty.microprofile6.internal_fat/fat/src/io/openliberty/microprofile6/internal/test/suite/MP6EE8CompatibleTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    @Mode(TestMode.FULL)
+    @ExpectedFFDC("java.lang.IllegalArgumentException")
+    public void testMP60andEE8() throws Exception {
+        server.startServerAndValidate(true, //preClean
+                                      false, //cleanStart
+                                      false, //validateApps
+                                      false, //expectStartFailure
+                                      false); //validateTimedExit
+    }
```
Azquelt: The expected FFDC appears to be the only assertion in this test? Should the test assert that the expected errors are emitted. It doesn't look like `MPCompatibilityTestUtils.cleanUp()` asserts that they occur either.

Azquelt: Same for the test with EE9

dev/io.openliberty.microprofile.config.2.0.internal_fat/fat/src/io/openliberty/microprofile/config/internal_fat/Config20ExceptionTests.java
tevans78: Perhaps but that wasn't the aim of this PR. Repeating Config FATs against MP6 would be a separate PR I think.

dev/io.openliberty.microprofile6.internal_fat/bnd.bnd
tevans78: Yes, it was copied, yes the date should be updated.

dev/io.openliberty.microprofile.config.2.0.internal_fat/fat/src/io/openliberty/microprofile/config/internal_fat/Config20ExceptionTests.java
Azquelt: Most of the config FATs have been updated for MP6 - I wasn't sure of the intention of these ones.

dev/io.openliberty.microprofile6.internal_fat/fat/src/io/openliberty/microprofile6/internal/test/suite/MP6EE8CompatibleTest.java
tevans78: The cleanUp method does expect the errors and fails if they do not show up ... but maybe that assertion should be in the test rather than the cleanUp?

tevans78:
#build

dev/io.openliberty.microprofile6.internal_fat/fat/src/io/openliberty/microprofile6/internal/test/suite/MPCompatibilityTestUtils.java
```diff
@@ -0,0 +1,131 @@
...
+            for (String msg : expectedFailuresRegExps) {
+                assertServerMessage(server, msg);
+            }
+            if (server.isStarted()) {
+                server.stopServer(expectedFailuresRegExps);
+            }
+        } else {
+            fail("Server was null");
+        }
+    }
```
Azquelt, line 79: This will not stop the server if the messages were not found.

Azquelt: Ah, but you're still calling cleanup in the `AfterClass` method, so you won't leave it around, even if this method throws an error.

Azquelt (approved):

tevans78:
#libby

PR #22236, opened 2022-08-23
Restructure the OP claim propagation FATs

c00crane:
I was trying to keep all of the claim propagation FATs under one project, but realized that it would be better to have a repository for the shared tests, then separate projects for oidc client and social client tests.


c00crane:
#run-libby-bot
#build

jimmy1wu (approved):

PR #22241, opened 2022-08-23
disallow Java 19 methods on persistent executor

njr-11:
With the persistent executor only being used internally for persistent EJB timers, there is no reason to provide working implementations of the new Java 19 methods.  But we also don't want the default implementations doing things that might not be correct in case persistent EJB timers or other function ever starts using them.  The most straightforward option is to force them to raise errors.  We can do this for all of the new methods except for getState, which cannot compile against Java 11 due to its return type.  For that one, we will need to rely on the default implementation hopefully doing the right thing in some cases and indirectly causing an exception to be raised when it doesn't.


njr-11:
#build

gjwatts (approved):

PR #22242, opened 2022-08-23
Change CWWKL0060W to unsupported usage error

dazavala:
Modify message CWWKL0060W to indicate that it is an unsupported usage **error** for a BELL configuration to reference the liberty global shared library and also enable SPI visibility. 

This PR also removes beta message CWWKL0061W as we are not going to implement/release the corresponding feature runtime function.

### DIFF
--- a/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
+++ b/dev/com.ibm.ws.classloading.bells/resources/com/ibm/ws/classloading/bells/internal/resources/BellConfigurationMessages.nlsprops
@@ -87,16 +87,9 @@ bell.spi.visibility.enabled.explanation=To prevent applications and resources fr
 bell.spi.visibility.enabled.useraction=Disable SPI visibility in the BELL configuration whenever the BELL must reference a library through a class loader that is shared with applications that reference the same library. Otherwise, no action is required.
 
 # BETA
-bell.spi.visibility.disabled.libref.global=CWWKL0060W: BELL SPI visibility is not enabled for the Liberty global shared library.
-bell.spi.visibility.disabled.libref.global.explanation=BELL SPI visibility is not supported for the Liberty global shared library.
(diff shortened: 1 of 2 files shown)


dazavala:
#build

dazavala:
#build

dmuelle (commented):
Hi @dazavala - the change itself looks good, however now the explanation just repeats the message text, which is not ideal. Is there any more info/ context we could provide to help the user in the explanation?  Would it be helpful to concisely explain why SPI visibility is incompatible w/ the Liberty global library? Or is that an internal detail that wouldn't concern the user?

dazavala:
Hello @dmuelle.  I can modify the explanation to describe the resulting behavior -- BELL SPI visibility was not enabled for the global library -- but I don't want to explain this arbitrary restriction in further detail other than the use-case is not supported. Technically, the global library is for EE applications, not OSGi services. The capability to configure a BELL to reference a global library may have been an oversight IMHO. The capability potentially exposes BELL services and the Liberty runtime to any class/resource the user dumps into the global library, which e.g. would easily cause integration problems with open source software packages. The capability has little value, because a BELL extension developer would/should package into a library only the artifacts necessary to support the provided services (post edit: and the developer expects class loading isolation).  If users request BELL SPI visibility for the global library or suffer a zero-migration regression, we can revisit the feature design and possibly alleviate the restriction.

dazavala:
@dmuelle fyi... The message is updated. Note that I cannot simply say "SPI visibility cannot be enabled for ...", because SPI visibility is not a configurable capability for a `<library/>` -- that is, a user cannot deploy a shared library and enable it to see SPI packages. The SPI visibility capability applies strictly to libraries referenced by a `<bell/>`. So I must say "BELL SPI visibility cannot be enabled...". --Thanks again.

dazavala:
#build

dmuelle (approved):
LGTM, thanks

tjwatson (approved):
LGTM

posmith:
L2 message review completed

dazavala:
#libbybot

dazavala:
Thanks everyone. Merged.

PR #22245, opened 2022-08-24
Added testcase to check JNDI behavior with checkpoint and restore

KeenanMcCarthy:
Added two test cases:
1. Test checkpoint behavior when JNDI entry is modified between checkpoint and restore
2. Check behavior when a new jndi entry is added between checkpoint and restore

### DIFF
--- a/dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/JNDITest.java
+++ b/dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/JNDITest.java
@@ -1,12 +1,17 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2022 IBM Corporation and others.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * (c) Copyright IBM Corporation 2022.
  *
- * Contributors:
- *     IBM Corporation - initial API and implementation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  *******************************************************************************/
 package io.openliberty.checkpoint.fat;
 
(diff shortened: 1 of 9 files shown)


KeenanMcCarthy:
#build

KeenanMcCarthy:
#libby

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/JNDITest.java
```diff
@@ -0,0 +1,106 @@
...
+        server.startServer();
+
+        ServerConfiguration config = server.getServerConfiguration();
+        JNDIEntry defaultEntry = new JNDIEntry();
+        defaultEntry.setJndiName("jndi/value");
+        defaultEntry.setValue("default value");
+        config.getJndiEntryElements().add(defaultEntry);
+        server.updateServerConfiguration(config);
+
+        server.stopServer();
```
tjwatson: You should stop the server before doing the config update.  It probably doesn't effect the actual test execution because you take the checkpoint and restore the server process before doing the config update.  But when you have a server running while you do a config update it will dynamically react to the config change on running (restored) server.

We backup the workarea on checkpoint so when you restore the server again in the test it will start from the state when you did the checkpoint (before the config change) and then reprocess the config change.  That is technically still testing what we need it to, but it may confuse others looking at this test and wondering why the config is changed while the server is running.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/JNDITest.java
```diff
@@ -0,0 +1,106 @@
...
+
+        server.startServer();
+        HttpUtils.findStringInUrl(server, "jndiApp/servlet", "default value");
+        server.stopServer();
+
+        ServerConfiguration config = server.getServerConfiguration();
+        config.getJndiEntryElements().getById("jndiEntry").setValue("alternate value");
+        server.updateServerConfiguration(config);
+
+        server.setCheckpoint(CheckpointPhase.APPLICATIONS, false, null);
```
tjwatson: You should set the checkpoint before starting the server the first time where we validate that the "default value" is being used.  That way wen you start the server again after the config change it will test that it detects the updated "alternate value" on restore.

dev/io.openliberty.checkpoint_fat/publish/servers/jndiServer/bootstrap.properties
```diff
@@ -0,0 +1,13 @@
...
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
+###############################################################################
+bootstrap.include=../testports.properties
+com.ibm.ws.logging.trace.specification=*=info:checkpoint=all
+io.openliberty.checkpoint.allowed.features=timedexit-1.0
```
tjwatson: I have a PR #22286 that adds the following to all our test servers, can you add this to the bootstrap.properties?

```
io.openliberty.checkpoint.dump.threads=true
```

KeenanMcCarthy:
#build

tjwatson (approved):

KeenanMcCarthy:
#libby

PR #22255, opened 2022-08-25
Fix more eclipse errors

tkburroughs:
- add missing .classpath entry


tkburroughs:
Change already merged under different PR; cancelling.

PR #22274, opened 2022-08-25
AdminCenter and its tools: change locale selection based on the browser language settings

lavenal:
The current implementation of all dojo based AdminCenter codes does not look at the browser locale list to select the first supported locale in the list. Instead it uses English as the default locale if the first locale in the list is not supported. This PR changes the logic to select the first supported locale in the list. If none is supported, then it would default to English.

Example, if the browser has the following languages in the order listed:

Dutch (nl) - not supported
Portuguese (pt) - not supported unless it is pt-br
Chinsee (zh) - supported
English (en) - supported

With the current implementation, Dutch is set as the locale. Since it is not supported, it defaults to English.
With the code changes, Chinese would be the selected locale.

### DIFF
--- a/dev/com.ibm.ws.ui.shared/resources/WEB-CONTENT/jsShared/bidiConfig.jsp
+++ b/dev/com.ibm.ws.ui.shared/resources/WEB-CONTENT/jsShared/bidiConfig.jsp
@@ -1,5 +1,5 @@
 <%--
-    Copyright (c) 2014 IBM Corporation and others.
+    Copyright (c) 2014, 2022 IBM Corporation and others.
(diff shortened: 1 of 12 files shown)


lavenal:
#build

lavenal:
#build

lavenal:
Confirmed that batch scan problem is not caused by my changes with the latest PB that doesn't include the OL changes.
With that, Amy also found the same batch scan problem in actual builds but no issue has opened for it.

lavenal:
close this and use https://github.com/OpenLiberty/open-liberty/pull/22379 to deliver instead.

Issue #22284, opened 2022-08-26
Check all features new in MP 6.0 require Java 11


Azquelt:
- [x] mpOpenAPI-3.1 #22288 
- [x] mpTelemetry-1.0 #22623

PR #22288, opened 2022-08-26
Make mpOpenAPI-3.1 bundles + tests require Java 11

Azquelt:
mpOpenAPI-3.1 already won't start with Java <11 due to a transitive dependency on eeCompatible-10.0, but we need to mark the bundles and tests as well.

For #22284


Azquelt:
#build

tevans78 (changes requested):
Would like to see the eclipse java versions changed as well

Azquelt:
I updated the execution environment on the project build path in eclipse and delivered the resulting changes to the .classpath and settings files.

I've also added the `Require-Capability` header to the API jar.

Azquelt:
#libby
#build

jhanders34 (approved):

tevans78 (approved):

PR #22299, opened 2022-08-26
experiment with repository update methods

njr-11:
The JNoSQL repository only has findBy and deleteBy.
This pull experiments with what an updateBy might look like.

Here is what I tried:
updateBy[conditions][updates]

where [conditions] is the same conditions you could have on a findBy or deleteBy and [updates] is one or more attributes to update following the pattern [Set|Add|Multiply|Divide]AttributeName,

so you end up with something like
updateByProductIdMultiplyPrice(productId, 1.02f);


njr-11:
#build

jmstephensgit (approved):

PR #22327, opened 2022-08-30
featureUtility repo connection failure error code

jjiwooLim:
In case of repository connection failure, throw install exception and exit with rc=33.

### DIFF
--- a/dev/com.ibm.ws.install.featureUtility/src/com/ibm/ws/install/featureUtility/FeatureUtility.java
+++ b/dev/com.ibm.ws.install.featureUtility/src/com/ibm/ws/install/featureUtility/FeatureUtility.java
@@ -72,6 +72,7 @@ public class FeatureUtility {
     private final static String OPEN_LIBERTY_PRODUCT_ID = "io.openliberty";
     private final static String WEBSPHERE_LIBERTY_GROUP_ID = "com.ibm.websphere.appserver.features";
     private final static String BETA_EDITION = "EARLY_ACCESS";
+    private final static String CONNECTION_FAILED_ERROR = "CWWKF1390E";
     private static String to;
     private boolean isInstallServerFeature = false;
 
@@ -482,7 +483,7 @@ public class FeatureUtility {
         try {
 
             for (File esaFile : artifacts) {
-				double increment = ((progressBar.getMethodIncrement("installFeatures")) / (artifacts.size()));
+		double increment = ((progressBar.getMethodIncrement("installFeatures")) / (artifacts.size()));
             	String featureName = extractFeature(esaFile.getName());
                 fine(Messages.INSTALL_KERNEL_MESSAGES.getLogMessage("STATE_INSTALLING", featureName));
                 map.put("license.accept", true);
@@ -590,6 +591,9 @@ public class FeatureUtility {
         if (map.get("action.error.message") != null) {
             fine("action.exception.stacktrace: " + map.get("action.error.stacktrace"));
             String exceptionMessage = (String) map.get("action.error.message");
+	    if (exceptionMessage.contains(CONNECTION_FAILED_ERROR)) {
+		throw new InstallException(exceptionMessage, InstallException.CONNECTION_FAILED);
+	    }
             throw new InstallException(exceptionMessage);
         }
         return result;
--- a/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
+++ b/dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
@@ -715,6 +715,18 @@ public class InstallKernelMap implements Map {
             } else {
                 throw new IllegalArgumentException();
             }
+        } else if (ACTION_ERROR_MESSAGE.equals(key)) {
+            if (value instanceof String) {
+                data.put(ACTION_ERROR_MESSAGE, value);
+            } else {
+                throw new IllegalArgumentException();
+            }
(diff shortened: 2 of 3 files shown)


jjiwooLim:
#build

jjiwooLim:
#build

jjiwooLim:
#build

dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/ReturnCode.java
```diff
@@ -49,6 +49,8 @@ public enum ReturnCode implements ExitCode {
     ERROR_BAD_JAVA_BITMODE(ExitCode.ERROR_BAD_JAVA_BITMODE), // 31
     // Used by the server scripts when jvm.options has an invalid line.
     ERROR_BAD_JVM_OPTION(32),
+    // Repository connection failed
+    CONNECTION_FAILED(33),
```
jjiwooLim, line 53: `InstallException.java` already has CONNECTION_FAILED=33, but `featureUtilityExectuor.java` checks if the return code is defined in `ReturnCode.java`. Adding rc=33 for connection failure. I will need to update featureUtility doc.

jjiwooLim:
#libby-bot

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -723,7 +723,7 @@ else if (DOWNLOAD_ARTIFACT_LIST.equals(key)) {
                 throw new IllegalArgumentException();
             }
         }
-        return data.get(key);
+        return data.put(key, value);
```
jacobwdv: What this method used to do is :
1. populate HashMap `data` after validating the key.
2. Return the value to which the specified key is now mapped, or null if this map contains no mapping for the key (the key wasn't added)

This changes the method to
1. populate HashMap `data` after validating the key.
2. If the key doesn't meet any criteria in the if block, add the key anyways. If the key did meet criteria in the if block, add it again.
3. Return the previous value associated with key, or null if there was no mapping for key.

Returning the value from `put()` is a significant change in behaviour, I don't think this is what is intended. Instead of making this change we need to add the key / value pair in the if block or in an `else` block after line 725.

jacobwdv: if we are only concerned about the key `action.error.message` then we should add an else if and check for that exact key: 
```
else if (key.equals("action.error.message")) {
            if (value instanceof String) {
                data.put(key, value);
            } else {
                throw new IllegalArgumentException();
            }
        }
```

dev/com.ibm.ws.install.featureUtility/src/com/ibm/ws/install/featureUtility/FeatureUtility.java
```diff
@@ -590,6 +590,9 @@ private List<File> downloadFeatureEsas(Collection<String> resolvedFeatures) thro
         if (map.get("action.error.message") != null) {
             fine("action.exception.stacktrace: " + map.get("action.error.stacktrace"));
             String exceptionMessage = (String) map.get("action.error.message");
+	    if (exceptionMessage.contains("CWWKF1390E")) {
```
jacobwdv: `CWWKF1390E` should be a String constant.

jacobwdv (changes requested):
A few comments. Feel free to ping me about this. I'm fairly sure we don't want to add the put the way we did in InstallKernelMap.java. I left details in the PR.

dev/com.ibm.ws.install.featureUtility/src/com/ibm/ws/install/featureUtility/FeatureUtility.java
```diff
@@ -590,6 +591,9 @@ private List<File> downloadFeatureEsas(Collection<String> resolvedFeatures) thro
         if (map.get("action.error.message") != null) {
```
jacobwdv, line 591: Can we use constants for `action.error.message` and `action.error.stacktrace`?

dev/com.ibm.ws.install/src/com/ibm/ws/install/internal/InstallKernelMap.java
```diff
@@ -1758,10 +1770,12 @@ public List<File> getJsonsFromMavenCentral(Set<String> jsonsRequired) throws Ins
             this.put(DOWNLOAD_ARTIFACT_LIST, jsonCoord);
 
             Object downloaded = this.get(DOWNLOAD_RESULT);
-            if (this.get("action.error.message") != null) {
-                fine("action.exception.stacktrace: " + this.get("action.error.stacktrace"));
-                String exceptionMessage = (String) this.get("action.error.message");
-                throw new InstallException(exceptionMessage);
+            String exceptionMessage = (String) this.get("action.error.message");
```
jacobwdv, line 1773: constants for `action.error.message` / `action.error.stacktrace`

jacobwdv (approved):
minor comment, looks good otherwise

jjiwooLim:
#libby-bot

jjiwooLim:
#build

dev/com.ibm.ws.install.featureUtility/src/com/ibm/ws/install/featureUtility/FeatureUtility.java
jjiwooLim: I will create a seperate PR for this

jjiwooLim:
#build

PR #22329, opened 2022-08-30
PR for Issue 22300: Fix issue with empty strings returned from provider metadata

### LINKED ISSUE #22300: Jakarta handling of null and empty string
The default values in providerMetadata are "" and our methods are checking if values were set by checking for null.  The checks will fail and we'll go down paths in the belief that we have valid values.
I found one instance in io.openliberty.security.oidcclientcore.internal/src/io/openliberty/security/oidcclientcore/authentication/JakartaOidcAuthorizationRequest.java
in
getAuthorizationEndpointFromProviderMetadata
....
if (authzEndpoint != null) {

and replaced that line with 
if (authzEndpoint != null && !authzEndpoint.isEmpty()) {

The code is behaving better now.
I'm sure that similar cases exist for the other values that we're getting out of providerMetadata.

ayoho:
Resolves #22300


ayoho:
#build

teddyjtorres (approved):

ayoho:
#libby
#build

PR #22332, opened 2022-08-30
updates to entity annotations for records

njr-11:
Update the annotations that we copied over from Jakarta NoSQL to include updates that went in there to allow for entity Id/Column annotations to be applied to Java record constructor parameters.


mswatosh (approved):

njr-11:
#build

PR #22333, opened 2022-08-30
Test setting CRIU_LOG_LEVEL for checkpoint

### LINKED ISSUE #22297: CRIU log level is not configurable for server checkpoint
Within `CRIU` support, the method `org.eclipse.openj9.criu.CRIUSupport.setLogLevel(int)` is never called from `io.openliberty.checkpoint.internal.openj9.ExecuteCRIU_OpenJ9.dump(Runnable,Runnable,File,String,File,File,boolean)`. The server checkpoint command cannot set (configure) the `CRIU` log level in order to gather debug diagnostics necessary to service `CRIU` failures during checkpoint. 

The server script defines environment variable `CRIU_LOG_LEVEL`, which is used to configure the `CRIU` log level for the restore command, only. The commands log `CRIU` diagnostics to their namesake files -- `checkpoint.log` and `restore.log` -- in the `${server.output.dir}/logs/checkpoint directory`. 

We need the same capability for checkpoint. And we can use the `CRIU_LOG_LEVEL` environment variable just as we do for restore, provided the key `CRIU_LOG_LEVEL` is added to the default list of immutable keys in the server script:
```
# Default list of immutable env vars for checkpoint
X_WLP_IMMUTABLE_VARS="-Dorg.eclipse.openj9.criu.ImmutableEnvVars=INVOKED,WLP_USER_DIR,WLP_OUTPUT_DIR,LOG_DIR,X_LOG_DIR,LOG_FILE,X_LOG_FILE,VARIABLE_SOURCE_DIRS,X_CMD"
```
Otherwise the JVM will return null when method `ExecuteCRIU_OpenJ9.dump` attempts to read the env var during checkpoint.

tjwatson:
fixes #22297

### DIFF
--- a/dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/CRIULogLevelTest.java
+++ b/dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/CRIULogLevelTest.java
@@ -1,0 +1,145 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.checkpoint.fat;
+
+import static io.openliberty.checkpoint.fat.FATSuite.configureEnvVariable;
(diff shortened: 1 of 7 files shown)


tjwatson:
#build

dazavala (approved):
More than comprehensive verification of checkpoint log level inputs. Without looking I presume the `@Skip...` annotation skips the test class on non-beta, non-lux images. You mentioned clarity concerns. IMO any complexity in the test's structure made sense after understanding that the information provided by the `TestMethod` `enum` is used to setup (`@Before`) each test server's checkpoint+restore operation and to provide checkpoint's expected outcome. Intuitive use of lambda to pass a test behavior.

PR #22343, opened 2022-08-31
Skip testNegativeoTimeOut

volosied:
Test verifies that a negatuve timeout does not set any timeout, but this connection to the server times out in jetty. 

This behavior is also tested via testSSCNegativeoTimeOut, so it should be okay to skip it.


volosied:
#build

pnicolucci (approved):

PR #22353, opened 2022-08-31
Wait for mongo to be ready before registering clients

c00crane:
Add a loop to list registered clients in a provider before trying to add a new client.
We're having problems registering clients every so often and I think that it's because the mongodb may not be ready.
I'm adding calls to list the clients in a provider before trying to add a new one.
If that list returns anything but 200, we'll sleep and try again after 5 seconds.  We'll do that multiple times - hopefully, it'll start working when mongo is ready and at that point, we'll be able to add the new client.


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.oidc.client_fat.backchannelLogout,com.ibm.ws.security.oidc.server_fat.backchannelLogout,com.ibm.ws.security.social_fat.LibertyOP.backchannelLogout

ayoho (approved):

PR #22370, opened 2022-09-01
Issue #19456: Configure infinispan to run with and without near-side …

### LINKED ISSUE #19456: SEC 380: Add authentication cache JCache tests for near-side / client-side cache
The tests should verify, where supported, that JCache can use the near-side cache not only to reduce the amount of requests to retrieve cache entries from going out on the network, but to also reuse the same object instance when possible as this will reduce the amount of deserialization required and increase performance.

This should be done for all of the JCache providers that we test in FVT, if they support it.

jvanhill:
…cache randomly each run. Update FAT tests to not run if skip.test system property is set. This will help with the eXtreme Scale fat in WS-CD-Open which can't run with Java 11 or on zOS at this time.

fixes #19456

### DIFF
--- a/dev/io.openliberty.jcache.internal_fat.hazelcast/fat/src/io/openliberty/jcache/internal/fat/BaseTestCase.java
+++ b/dev/io.openliberty.jcache.internal_fat.hazelcast/fat/src/io/openliberty/jcache/internal/fat/BaseTestCase.java
@@ -15,6 +15,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
 import org.junit.After;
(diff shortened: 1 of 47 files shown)


jvanhill:
#build

ryanesch (approved):

jvanhill:
#build

jvanhill:
#build

jvanhill:
#build

jvanhill:
#build
#libby

jvanhill:
Docker environment appeared unavailable. Respin the build.

#build

jvanhill:
#build

jvanhill:
#build

Issue #22198, opened 2022-08-20
CTS: sigtest: Faces 4.0 failures


pnicolucci:
**Bold** failures still need issues opened

[javatest.batch] Missing Superclasses or Superinterfaces
[javatest.batch] ---------------------------------------
[javatest.batch] 
[javatest.batch] jakarta.faces.component.html.HtmlDoctype:                   interface @ jakarta.faces.component.Doctype
[javatest.batch] 
[javatest.batch] Missing Classes
[javatest.batch] ---------------
[javatest.batch] 
[javatest.batch] jakarta.faces.component.behavior.FacesBehavior$Literal
[javatest.batch] jakarta.faces.convert.FacesConverter$Literal
[javatest.batch] jakarta.faces.event.WebsocketEvent$Closed$Literal
[javatest.batch] jakarta.faces.event.WebsocketEvent$Opened$Literal
[javatest.batch] jakarta.faces.flow.builder.FlowBuilderParameter$Literal
[javatest.batch] jakarta.faces.flow.builder.FlowDefinition$Literal
[javatest.batch] jakarta.faces.model.FacesDataModel$Literal
[javatest.batch] jakarta.faces.push.Push$Literal
[javatest.batch] jakarta.faces.validator.FacesValidator$Literal
[javatest.batch] 
[javatest.batch] Added Classes
[javatest.batch] -------------
[javatest.batch] 
[javatest.batch] jakarta.faces.view.facelets.FaceletsResourceResolver
[javatest.batch] 
[javatest.batch] Missing Nested Classes
[javatest.batch] ----------------------
[javatest.batch] 
[javatest.batch] jakarta.faces.component.behavior.FacesBehavior:             nested public final static jakarta.faces.component.behavior.FacesBehavior$Literal
[javatest.batch] jakarta.faces.convert.FacesConverter:   nested public final static jakarta.faces.convert.FacesConverter$Literal
[javatest.batch] jakarta.faces.event.WebsocketEvent$Closed:                  nested public final static jakarta.faces.event.WebsocketEvent$Closed$Literal
[javatest.batch] jakarta.faces.event.WebsocketEvent$Opened:                  nested public final static jakarta.faces.event.WebsocketEvent$Opened$Literal
[javatest.batch] jakarta.faces.flow.builder.FlowBuilderParameter:            nested public final static jakarta.faces.flow.builder.FlowBuilderParameter$Literal
[...]

PR #22387, opened 2022-09-05
Myfaces 4453

pnicolucci:
for #22198

https://issues.apache.org/jira/browse/MYFACES-4453

The only difference between this PR and what was delivered in the open source is we're missing MYFACES-4445 so in EndpointImpl we don't need to call `get` on the BeanManager. We should work to pull in MYFACES-4445 next so we have the updates made there as well for additional testing.

### DIFF
--- a/dev/io.openliberty.jakarta.faces.4.0/bnd.bnd
+++ b/dev/io.openliberty.jakarta.faces.4.0/bnd.bnd
@@ -46,6 +46,9 @@ publish.wlp.jar.suffix: dev/api/spec
 # Use this once we have a non RC1 release
 # org.apache.myfaces.core:myfaces-api;strategy=exact;version=${Implementation-Version},\
 -buildpath: \
+    io.openliberty.jakarta.cdi.4.0;version=latest, \
     io.openliberty.jakarta.expressionLanguage.5.0;version=latest, \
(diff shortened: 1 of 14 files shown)


pnicolucci:
#libby #build

pnicolucci:
#libby #build

pnicolucci:
#libby #build

volosied (approved):
I compared to open source. and it looks good to me. 

Thomas also deleted some files, and I think these still exist in the liberty bundles. However, I don't think they should cause any problems while we wait for the next release.

otherwise, the only other changes pulled in where the array changes replacing c style to java style

pnicolucci:
> I compared to open source. and it looks good to me.
> 
> Thomas also deleted some files, and I think these still exist in the liberty bundles. However, I don't think they should cause any problems while we wait for the next release.
> 
> otherwise, the only other changes pulled in where the array changes replacing c style to java style

Thanks @volosied  for the review. Yea, I think the array changes are fine here and not deleting the files is also ok. I couldn't think of a good way to easily delete them within the bundle. We will need an RC2 once we get closer to a beta and should be able to clean a lot of this up at that point in time.

pnicolucci:
#libby #build

Issue #19867, opened 2022-01-20
Enable more existing CDI FAT buckets to run with EE10


tevans78:
Part 1: https://github.com/OpenLiberty/open-liberty/pull/19865

CDI FAT buckets:

- [x] annotations_fat
- [x] api_fat
- [x] beansxml_fat
  - Looks done except `BeanDiscoveryModeNoneTest` #22607
- [x] beansxml.implicit_fat
- [x] ejb_fat
- [x] extension_fat
- [x] jee_fat
  - Done except `InjectIntoPathTest` failing on EE10 #22607
- [x] jndi_fat
- [x] lifecycle_fat #22391
- [x] noncontextual_fat #22406 
- [x] third.party_fat #22414 
- [x] visibility_fat #22585
- [x] 3.0_fat #22607
- [x] 4.0_fat
- [x] core_fat

tevans78:
All done

PR #22391, opened 2022-09-06
Update CDI Lifecycle FAT and repeat on EE10

Azquelt:
For #19867


dev/com.ibm.ws.cdi.lifecycle_fat/test-applications/ObservesInitializedInJars.war/src/cdi12/observersinjarsbeforebean/WarBeforeBeansObserver.java
```diff
@@ -1,60 +0,0 @@
...
-import javax.enterprise.inject.spi.Extension;
-
-import cdi12.observersinjars.SomeClass;
-
-public class WarBeforeBeansObserver implements Extension {
-
-    private static String beforeBeanDiscoveryCalled = "Method not called";
-
-    private ClassLoader TCCL = null;
-    private ClassLoader myCL = null;
```
Azquelt, line 24: This class is a bit odd.

It was registered as an extension and `beforeBeanDiscovery` logs whether certain classes are loadable from the app class loader and the TCCL.

However, nothing was asserting what the values should be. I took it out to simplify the test but if we decide that we do want to assert something here we could add it back in.

Azquelt:
In general there were a few cases where I simplified the application packaging (e.g. switching an .ear for a .war or removing a web.xml which didn't declare anything) and removed features from the server.xml because it didn't seem relevant to the test.

Azquelt:
#build

tevans78 (approved):

PR #22393, opened 2022-09-06
Asynchronous on REST application method

njr-11:
A user who was trying out the Concurrency 3.0 / Jakarta EE 10 specification challenged whether an `@Asynchronous` returning `void` could be used on a REST application method.  Given that its a CDI managed bean, this should work.  This pull adds an automated test case demonstrating that it does.


njr-11:
#build

jmstephensgit (approved):

PR #22402, opened 2022-09-06
ignore reg change timeout error msg

tabishop:
Fix for defect 291876. Ignore reg timeout error which does not cause an actual problem.


tabishop:
#build

mswatosh (approved):

PR #22409, opened 2022-09-07
21186 tck fat2.1 updates

simysc:
#21186


simysc:
#build

teddyjtorres (approved):
Thank you very much for these changes. We will deliver these changes in first and then do the necessary updates to make it work with mpJwt-2.1 and the 2.1-RC2 or 2.1-RC3 artifacts.

PR #22412, opened 2022-09-07
EJB Persistent timer test - ignore shut down error

olendvcook:
Some of the tests start up a server and shut it down soon after, which can cause slow components to not fully come up before the server starts shutting down causing error printouts unrelated to the test, adding an ignore for one.


olendvcook:
#build

brideck (approved):

PR #22417, opened 2022-09-07
like keyword and contains for non-collections

njr-11:
Behavior of the JNoSQL Like keyword isn't well documented, and this pull is speculating how it might apply to JPQL when wildcards are used.  Also, I've updated the examples with a Contains keyword, which was previously only for collection types to also apply to string types, giving it the previous usage that we had in place for Like.  This will make both more intuitive given their names.


njr-11:
#build

jmstephensgit (approved):

njr-11:
#build

PR #22424, opened 2022-09-07
Add initial test for checkpoint and WebSocket

KeenanMcCarthy:
Created a test application that queries a WebSocket endpoint based on the "clientEndpoint" variable in the server.xml file. After checkpoint, this variable is modified to an alternate endpoint, which is then queried on restore.


KeenanMcCarthy:
#build
#libby

tjwatson (approved):

KeenanMcCarthy:
#libby

Issue #21157, opened 2022-05-17
Security 3.0: Create OidcClientCore Client Discovery Subsystem


teddyjtorres:
Task: Create OidcClientCore Client Discovery Subsystem. Discovery to be performed on first call to perform Authentication Request.

Spec section: https://github.com/jakartaee/security/blob/master/spec/src/main/asciidoc/authenticationMechanism.adoc#metadata-configuration

> The OpenID Connect authentication mechanism needs metadata about the OpenID Connect Provider to function properly. The OpenID Connect specification defines that this data can be read from the well known openid configuration endpoint which resides at [OpenID Connect Provider base URL].well-known/openid-configuration.

...

> The OpenIdAuthenticationMechanismDefinition.providerURI defines the base URL of the OpenID Connect Provider where the /.well-known/openid-configuration is appended to (or used as-is when it is the well known configuration URL itself). Reading the well known openid configuration endpoint can be done eagerly when the application is deployed or lazily at the time a secured URL is accessed for the first time.

ayoho:
The following FATs should cover this:
- ConfigurationTests_minimumAnnotation
    - `providerURI`: Configured
    - `providerMetadata`: Not configured
- ConfigurationTests_maximumAnnotation
    - `providerURI`: Configured
    - `providerMetadata`: Configured, but with small subset of values
- ConfigurationTests_noProviderURI_withoutProviderMetadata
    - `providerURI`: Not configured
    - `providerMetadata`: Not configured
- ConfigurationTests_noProviderURI_withProviderMetadata
    - `providerURI`: Not configured
    - `providerMetadata`: Configured with valid values

Other FATs throughout the buckets have a mix of `providerMetadata` being configured or not configured for various purposes.

PR #22427, opened 2022-09-07
PR for Issue 21157: Perform discovery for every OIDC request, not just for missing authz endpoint

ayoho:
Updates the Jakarta Security code to perform OIDC Discovery right at the start of each flow. Before, we were only performing discovery when the authorization endpoint URL wasn't configured in the provider metadata in the servlet annotation. This will ensure we always have the OP's metadata and that values in the provider metadata can be used to override values from the discovery data.

For #21157


ayoho:
#build

teddyjtorres (approved):

PR #22432, opened 2022-09-08
specify bc provider package version

arunavemulapalli:
specify version in the dynamic import for bouncycastle.jce.provider package. 
I see that only test code is using the init()  where we are initializing BC's JCE provider, but somewhere else I saw that wss4j is using some bc utilities. So, specifying the version in the import here to make sure that we are using the level that we think are using.


arunavemulapalli:
#build

PR #22448, opened 2022-09-09
remove token from logs

tevans78:
Make sure that secure token data is not output to test logs


tevans78:
#build

websterc87 (approved):
Approved pending successful completion of the build and verification that the token is no longer in the output.

PR #22472, opened 2022-09-13
entity type always inferred

njr-11:
Investigate whether it is possible to always infer the entity type and not require it on the data repository annotation.
Based on how discussions are going in Jakarta Data, it seems that a base repository interface will almost always be extended, which includes the entity type as a parameter.


njr-11:
#build

nmittles (approved):

PR #22495, opened 2022-09-14
Updates for running tests on ARM

mswatosh:
Sets the default to using remote containers when running on ARM, due to decreased performance and failure to start with some x86 containers.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
+++ b/dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
@@ -79,8 +79,9 @@ public class ArtifactoryImageNameSubstitutor extends ImageNameSubstitutor {
 
     static String getPrivateRegistry() {
         String artifactoryServer = System.getProperty(artifactoryRegistryKey);
-        if (artifactoryServer == null || artifactoryServer.isEmpty() || artifactoryServer.startsWith("${"))
-            throw new IllegalStateException("No private registry configured. System property '" + artifactoryRegistryKey + "' was: " + artifactoryServer);
+        if (artifactoryServer == null || artifactoryServer.isEmpty() || artifactoryServer.startsWith("${") || artifactoryServer.equals("null"))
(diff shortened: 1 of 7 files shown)


mswatosh:
#build

mswatosh:
#build

mswatosh:
#build

mswatosh:
Test failure matches a known issue

dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
```diff
@@ -79,8 +79,9 @@ private static boolean isSyntheticImage(DockerImageName dockerImage) {
 
     static String getPrivateRegistry() {
         String artifactoryServer = System.getProperty(artifactoryRegistryKey);
-        if (artifactoryServer == null || artifactoryServer.isEmpty() || artifactoryServer.startsWith("${"))
-            throw new IllegalStateException("No private registry configured. System property '" + artifactoryRegistryKey + "' was: " + artifactoryServer);
+        if (artifactoryServer == null || artifactoryServer.isEmpty() || artifactoryServer.startsWith("${") || artifactoryServer.equals("null"))
+            throw new IllegalStateException("No private registry configured. System property '" + artifactoryRegistryKey + "' was: " + artifactoryServer + "  "
+                                            + "Ensure artifactory properties are set in gradle.startup.properties");
```
KyleAure, line 84: Was there a reason this line was removed?

KyleAure (approved):
The changes look good. Just a clarification question above.

dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
mswatosh: Unless I'm missing something nothing was removed here? I added an additional check for the string "null" because I hit that from not having artifactory properties in my gradle.startup.properties file, and then added the message about checking gradle.startup.properties in the Exception.

PR #22500, opened 2022-09-14
Speed up SAML Fats in LITE mode - plus clean up extra error msgs

c00crane:
The ldap server that these tests use was down (was still down today) and it was taking 15 - 20 minutes for consul to ping the server, fail, retry, ...  Then the transformer was running because this is an EE9 instance and that slowed things down.
The LDAP server is back up, I'm working on making this project (and as many SAML projects as possible) run using the in-memory ldap servers instead of real servers.
The build of the shibboleth war was resulting in a war with 2 copies of status.jsp (in the same location - the bad copy is being used and that resulted in a "operator is not allowed for source level below 1.7" error in the log (it wasn't causing anything to fail, but it's ugly and we're replacing the file to fix that issue).  I've made a change to only have 1 copy of the file (the updated jsp).
I'm also going to add a timeout to the consul query of the available ldap servers - for the cases where some may be down, it won't hang for so long.
Finally, I'm updating most of the SAML Fats to only use EE9 in full mode (except on Windows).


c00crane:
#build

c00crane:
#run-libby-bot
#build

ayoho (approved):

c00crane:
#run-libby-bot
#build

Issue #22368, opened 2022-09-01
Faces 4.0: Enable com.ibm.ws.jsf.2.3_fat tests for EE10


pnicolucci:
Init delivery of tests that are currently passing: PR:[ 22385](https://github.com/OpenLiberty/open-liberty/pull/22385)

1) Enable the remaining tests

Commented out tests in EE10-enabled test classes:

1) ~~The below tests need to be updated as they are looking in the Session map for the Session beans (ManagedBeans) but when using `@Named` beans the Session beans are no longer in the Session map so we need to work on a solution:
a. JSF23CDIConfigByACPTests  (testNavigationHandlerInjection_CDIConfigByACP)
b. JSF23CDIFacesInMetaInfTests  (testNavigationHandlerInjection_CDIFacesInMetaInf)
c. JSF23CDIFacesInWebXMLTests  (testNavigationHandlerInjection_CDIFacesInWebXML)
Initial work for these was done: https://github.com/OpenLiberty/open-liberty/pull/22488
**Enabled using this PR**: #23775~~

2) ~~AJAX request not working:
JSF23SelectOneRadioGroupTests.java (testSelectOneRadioGroup_AjaxRequest) : https://github.com/OpenLiberty/open-liberty/pull/22501 Enabling in this PR: https://github.com/OpenLiberty/open-liberty/pull/24377. This one will need to stay commented out now that RC5 is integrated. I'll update the comment here: #24553 and we'll test it manually for now here: https://github.com/OpenLiberty/open-liberty/issues/24387~~

3) ~~JSF23CDIGeneralTests.java (testInjectableELImplicitObjects (enabled as part of https://github.com/OpenLiberty/open-liberty/pull/22642), testSpecIssue790Test1, testSpecIssue1404And1423_JSF23NewAPIMethods(enabled again here: https://github.com/OpenLiberty/open-liberty/pull/24152)) : https://github.com/OpenLiberty/open-liberty/pull/22505~~

4) ~~JSF23UIRepeatConditionTests  (testUIRepeatCondition): https://github.com/OpenLiberty/open-liberty/pull/22525 This is another case where ajax looks to be working manually but HtmlUnit doesn't like it for some reason. -> https://github.com/OpenLiberty/open-liberty/pull/24345~~

[...]

PR #22501, opened 2022-09-14
Enable JSF23SelectOneRadioGroupTests.java for Jakarta EE10 Faces 4.0

pnicolucci:
for #22368

One failing test case is still not enabled. Looks to be passing manually but the second AJAX request does not look to be happening in the automation using HtmlUnit.


pnicolucci:
#libby #build

volosied (approved):

Issue #21166, opened 2022-05-17
Security 3.0: Create OidcClientCore Token Subsystem


teddyjtorres:
Task: Create OidcClientCore Token Subsystem. Update OidcHttpAuthenticationMechanism to use OidcClientCore Client to perform Token Request (Continues AuthorizationCodeFlow up to the point of performing a Token Request and returns a TokenResponse, no token validation since this will be done in a new IdentityStore bean)

Spec section: https://github.com/jakartaee/security/blob/master/spec/src/main/asciidoc/authenticationMechanism.adoc#authentication-dialog

> For step (4) the authentication mechanism itself must call the Token endpoint to retrieve an Access Token and ID Token. This constitutes a so-called server to server call, as the end-user (caller) MUST NOT be involved here.
> The call to the token endpoint must include the following parameters (as specified by the OpenID Connect specification):
> 
> - The ClientId value as taken from OpenIdAuthenticationMechanismDefinition.clientId
> - The ClientSecret value as taken from OpenIdAuthenticationMechanismDefinition.clientId
> - The grant_type value set to the constant authorization_code
> - The RedirectURI value as taken from OpenIdAuthenticationMechanismDefinition.redirectURI
> - The code received from the OpenID Connect Provider in the callback request as the code request parameter

> If the call to the Token endpoint is successful, it should return a "Token Response" in JSON format.

ayoho:
Specific tests for error scenarios or corner cases:
- ConfigurationTests_badClientSecret
    - Ensures the token request fails
- ConfigurationTests_omittedClientSecret
    - Ensures the token request fails

Pretty much every other FAT executes the token request and tests the token subsystem.

PR #22513, opened 2022-09-15
Initial implementation of token and claims interfaces for issue #21166

teddyjtorres:
For issue #21166

Creates initial implementation for the tokens and claims APIs defined in https://github.com/jakartaee/security/tree/master/api/src/main/java/jakarta/security/enterprise/identitystore/openid.

### DIFF
--- a/dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/AccessTokenImpl.java
+++ b/dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/AccessTokenImpl.java
@@ -1,0 +1,108 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.security.jakartasec.tokens;
+
+import java.io.Serializable;
+import java.time.Instant;
+import java.util.Collections;
+import java.util.Map;
+
+import jakarta.security.enterprise.identitystore.openid.AccessToken;
+import jakarta.security.enterprise.identitystore.openid.JwtClaims;
+import jakarta.security.enterprise.identitystore.openid.Scope;
+
+public class AccessTokenImpl implements AccessToken, Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private final String tokenString;
+    private final Long expirationTimeInSeconds;
+    private final Instant responseGenerationTime;
+    private final Long tokenMinValidityInMillis;
+    private final Type type;
+
+    public AccessTokenImpl(String tokenString, Instant responseGenerationTime, Long expirationTimeInSeconds, Long tokenMinValidityInMillis) {
+        this.tokenString = tokenString;
+        this.expirationTimeInSeconds = expirationTimeInSeconds;
+        this.responseGenerationTime = responseGenerationTime;
+        this.tokenMinValidityInMillis = tokenMinValidityInMillis;
+        // TODO: Detect if a JWT access token
+        type = Type.MAC;
+    }
+
+    @Override
+    public String getToken() {
+        return tokenString;
+    }
+
+    @Override
+    public boolean isJWT() {
+        // TODO Auto-generated method stub
+        return false;
+    }
+
+    @Override
+    public JwtClaims getJwtClaims() {
+        if (Type.BEARER.equals(type)) {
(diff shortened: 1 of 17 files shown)


teddyjtorres:
#build

dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/AccessTokenImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public Long getExpirationTime() {
+        return expirationTimeInSeconds;
+    }
+
+    @Override
+    public boolean isExpired() {
+        if (expirationTimeInSeconds != null && expirationTimeInSeconds > 0 && tokenMinValidityInMillis > 0) {
```
ayoho, line 85: If the access token's expiration time is set to 0, wouldn't this method always return false, and the token would never be considered expired? Similar question for if `tokenMinValidityInMillis == 0` - wouldn't that skip this check and (assuming type isn't BEARER) always return false as well?

dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/ClaimsImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    private static final long serialVersionUID = 1L;
+    private final Map<String, Object> claims = Collections.synchronizedMap(new HashMap<>());
+
+    public ClaimsImpl(Map<String, Object> claimsMap) {
+        claims.putAll(claimsMap);
+    }
+
+    @Override
+    public Optional<String> getStringClaim(String name) {
+        String value = getClaim(name);
```
ayoho, line 39: Just curious: why only use the private `getClaim()` method here and not in the other methods as well?

dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/IdentityTokenImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private final String tokenString;
+    private final Map<String, Object> idTokenClaimsMap;
+    private final JwtClaims jwtClaims;
+    private final Long tokenMinValidityInMillis;
+
+    public IdentityTokenImpl(String tokenString, Map<String, Object> idTokenClaimsMap, Long tokenMinValidityInMillis) {
+        this.tokenString = tokenString;
+        this.idTokenClaimsMap = idTokenClaimsMap;
```
ayoho, line 33: I can see it being an issue if working with JWE tokens, but is the `idTokenClaimsMap` arg necessary if we could just do essentially what you have in the unit test?
```suggestion
        String[] parts = tokenString.split(("\\."));
        String tokenClaimsJsonString = new String(Base64.getDecoder().decode(parts[1]), "UTF-8");
        this.idTokenClaimsMap = org.jose4j.jwt.JwtClaims.parse(tokenClaimsJsonString).getClaimsMap();
```
I'm wondering if there'd be any room for discrepancy between the `tokenString` and the `idTokenClaimsMap`.

ayoho (approved):

dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/ClaimsImpl.java
teddyjtorres: getClaim() works well for String, but had some issues when it was a numeric type, like integers and longs, so decided to read the value from the claims map instead.

dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/IdentityTokenImpl.java
teddyjtorres: The TokenResponseValidator already parses the id token and gets jose4j's JwtClaims object, 

```
JwtClaims jwtClaims = null;
        if (jwtcontext != null && jwtcontext.getJwtClaims() != null) {
            jwtClaims = jwtcontext.getJwtClaims();
```

so instead of parsing the token again here, the validator can create this IdentityTokenImpl object with the map from the JwtClaims object after successful token verification. That would improve performance.

I agree it might be an issue for JWE is the validator is not able to handle JWE.

dev/io.openliberty.security.jakartasec.3.0.internal/src/io/openliberty/security/jakartasec/tokens/AccessTokenImpl.java
teddyjtorres: I agree. For tokenMinValidity, the requirement is that it cannot be a negative value. The same should apply to expires_in. I will also modify this method in a follow up PR to return true by default.

PR #22516, opened 2022-09-15
Added testcase for OpenApi

KeenanMcCarthy:
This testcase modifies the mp.openapi.servers value in the server config between checkpoint and restore, then verifies that this change is reflected at the /openapi endpoint.


KeenanMcCarthy:
#libby
#build

anjumfatima90 (approved):

PR #22551, opened 2022-09-16
Fix test bug 199875

dazavala:
The BELL fat case `BellFATTest` breaks z/OS builds because the server fails to start during `@Before` w/ no clues, for each test. The test case does not fail other builds.  All tests recently introduced for epic issue 19368 do not suffer the failure, which implies an issue with server startup timing or the server configuration files are problematic on z/OS. Modify the BellFATTest case to use the working test pattern in order to fix the tests or obtain better diagnostics.


dazavala:
#build

dev/com.ibm.ws.classloading.bells_fat/fat/src/com/ibm/ws/classloading/BellFatTest.java
```diff
@@ -101,9 +106,21 @@ static void stopServer() {
...
+        sysProps.put(BETA_EDITION_JVM_OPTION, Boolean.FALSE.toString());
+        try {
+            setSysProps(server, sysProps);
+            server.startServer();
+
+            assertNotNull("Should warn that no services were found to export from the 'testNoEntry' library.",
+                    server.waitForStringInLog(".*CWWKL0055W: .*testNoEntry"));
+        } finally {
+            stopServer();
+            removeSysProps(server, sysProps);
```
tjwatson, line 115: Can you removeSysProps before stopServer here?  Otherwise you risk no calling removeSysProps if stopServer fails.

dev/com.ibm.ws.classloading.bells_fat/fat/src/com/ibm/ws/classloading/BellFatTest.java
```diff
@@ -69,7 +74,7 @@ public static void tearDown() throws Throwable {
 
     @Before
     public void beforeTest() throws Throwable {
-        server.startServer();
+        // nada
```
tjwatson: Why keep this method?  Is it overriding something?

tjwatson (approved):
I think this is fine to deliver.  Just have a couple of suggestions/questions?  But if we need this delivered promptly you can fix up later.

dev/com.ibm.ws.classloading.bells_fat/fat/src/com/ibm/ws/classloading/BellFatTest.java
dazavala: The test works as intended when `removeSysProps()` is called before `stopServer()`. The call to `stopServer()` is vestigial as the method is also called by `@After afterTest()`. So, I removed the call to `stopServer()` from all tests that manage server system props, except test `BellPropertiesTest.testPropertiesInvalid()`. BTW, this change was introduced to ensure the test rewrites the jvm.options file before `bells_fat_server` is started for the first time.

dev/com.ibm.ws.classloading.bells_fat/fat/src/com/ibm/ws/classloading/BellFatTest.java
dazavala: I removed the method from all three test cases in the suite.

dazavala:
#build

dazavala:
#libbybot

PR #22570, opened 2022-09-19
initial steps before creating NoSQL repository

njr-11:
Set up a test case where a NoSQL repository (currently unusable) will be injected, along with the code to create it.
This also updates the existing test bucket for relational data access repository to transition more toward using the copy of interfaces from the nonship feature so that we can delete the duplicates from the test bucket.


njr-11:
#build

mswatosh (approved):

PR #22571, opened 2022-09-19
Fix issue with states getting changes in two new getPolicyConfig methods

emilytee:


emilytee:
Update the PolicyConfigurationFactory's two new getPolicyConfiguration methods to not change the state as per the spec: 
If at the time of the call, the identified policy context does not exist in the provider, then the policy context
     * will <strong>not</strong> be created in the provider and a null will be returned. No state transition of any kind
     * is allowed to occur, the <code>PolicyConfiguration</code> instance is to be returned as-is.

emilytee:
#run-libby-bot

emilytee:
#build

jhanders34 (approved):

emilytee:
#build

emilytee:
#run-libby-bot

PR #22605, opened 2022-09-21
[TS010666588] skip illegal characters in jvm.options. (release bug)

bmarwell:
This resolves IBM case TS010666588.

Reading a jvm.options file with an invalid character, e.g. `0xfc` will make Open Liberty and IBM Liberty stop reading the file on encountering the character in certain situations. This will make the server start without the JVM options after that character.

Circumstances:
* jvm.options is in ASCII/UTF-8
* jvm.options contains a non-standard char like `0xfc` which is invalid for ASCII and UTF-8.
* OS is Mac/AIX
* Language is set to anything ending with `.UTF-8`.

This bug does not appear, when:

* on Linux, regardless of language settings. It will always pass through invalid chars.
* On AIX/Mac when LANG=C is set _OR_ LANG= does not end in UTF-8 OR the file `jvm.options` is encoded properly.

### DIFF
--- a/dev/com.ibm.ws.appclient.boot.ws-client/publish/bin/client
+++ b/dev/com.ibm.ws.appclient.boot.ws-client/publish/bin/client
@@ -450,7 +450,7 @@ mergeJVMOptions()
       saveIFS=$IFS
       IFS=$newline
       
-      for option in `readNativeFile "$jvmOptions" '[#-]' | tr -d '\r'`; do
+      for option in `readNativeFile "$jvmOptions" '[#-]' | LC_ALL=C tr -d '\r'`; do
         if [ -n "$option" ]; then
           case $option in
           \#*);;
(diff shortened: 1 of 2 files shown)


bmarwell:
Code was tested with a modified version of the `server` script and with an excerpt of the server script: https://gist.github.com/bmarwell/fb026966b72a0e1ffd947ada6b480b48

* Without my changes, the output is empty (unexpected).
* With my  changes applied, the script will properly output `-Xms1024m`.

Real example attached to the IBM case.

Adding LANG=C is a known workaround for such issues on BSD-based systems.

bmarwell:
Both CLA and iCLA are already signed, as proven by this commit: https://github.com/OpenLiberty/open-liberty/pull/8956/commits/e159898d1c9a230f5d71b82414874204a1a66715 from this PR: https://github.com/OpenLiberty/open-liberty/pull/8956

jimblye (approved):

bmarwell:
Hey @jimblye @cbridgha I did re-sign the CLA (due to changed github name). Any news on this?

PR #22607, opened 2022-09-21
Update remaining CDI FAT tests for EE10

Azquelt:
- 3.0 FAT
- InjectIntoPathTest in jee FAT
- BeanDiscoveryModeNoneTest in beansxml FAT

For #19867

### DIFF
--- a/dev/io.openliberty.cdi.3.0.internal_fat/fat/src/io/openliberty/cdi30/internal/fat/tests/CDI30BeansXMLTest.java
+++ b/dev/io.openliberty.cdi.3.0.internal_fat/fat/src/io/openliberty/cdi30/internal/fat/tests/CDI30BeansXMLTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 IBM Corporation and others.
+ * Copyright (c) 2021, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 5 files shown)


Azquelt:
#build

dev/io.openliberty.cdi.3.0.internal_fat/fat/src/io/openliberty/cdi30/internal/fat/tests/CDI30BeansXMLTest.java
```diff
@@ -39,10 +41,11 @@ public class CDI30BeansXMLTest extends FATServletClient {
 
     private static final String BEANS_XML_APP_NAME = "CDI30BeansXMLApp";
 
+    @ClassRule
+    public static RepeatTests r = EERepeatActions.repeat(SERVER_NAME, EERepeatActions.EE9, EERepeatActions.EE10);
```
tevans78, line 45: This would run EE9 in LITE and EE10 in FULL ... which is the opposite to what we do elsewhere? Should it be switched?

tevans78 (approved):

dev/io.openliberty.cdi.3.0.internal_fat/fat/src/io/openliberty/cdi30/internal/fat/tests/CDI30BeansXMLTest.java
tevans78: we'll review LITE vs FULL later

PR #22610, opened 2022-09-21
Fix CDI12ExtensionTest on EE10

### LINKED ISSUE #19900: Investigate why CDI12ExtensionTest.testHelloWorldExtensionServlet does not work on EE10
It seems that with CDI 4.0, if the application extension does not get scanned properly for beans if the beans.xml either contains no version information (CDI v1.0) or has bean discovery mode set to annotated (CDI v1.1+). If the discovery mode is set to ALL then CDI 4.0 is happy.

https://github.com/OpenLiberty/open-liberty/blob/66284ce3c6684229f07abb4272bc46b55aab5e13/dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/tests/CDI12ExtensionTest.java#L87

Azquelt:
Previously a CDI 1.0 beans.xml was added. This results in a bean-discovery-mode of "all" on 3.0 and earlier but "annotated" on 4.0.

Instead, use a CDI 1.1 beans.xml which explicitly states a bean-discovery-mode of "annotated" and annotate all bean classes appropriately.

Fixes #19900


Azquelt:
#build

tevans78 (approved):

PR #22613, opened 2022-09-21
Fix one of the build failure captured in comment 4 of RTC 292562. Gen…

jvanhill:
…erate new cluster names for each test to ensure we don't join an already running cluster.


jvanhill:
#build

kristip17 (approved):

Issue #22572, opened 2022-09-19
MP Primitive performance improvements review of RegisterRequestInterceptor.notify


uberskigeek:
Reviewing need to reverse order of services if 1 or fewer services are in the service list

PR #22633, opened 2022-09-23
Skip reversal of service list

uberskigeek:
Skip the reversal of the service list in the event the list has only 1 entry.


uberskigeek:
#build

jhanders34 (approved):

uberskigeek:
Errors Found in the RTC build are not related to the changes in this PR. 
One error is the result of being. up bale to dowload a docker image
One is a timing issue between server startup and test execution
and the Third is an issue with the format of the automation UI where the format of the window obscures the widget that needs to be clicked.

uberskigeek:
Fixes #22572

PR #22636, opened 2022-09-23
use explicit version for hamcrest

benjamin-confino:
This test works with hamcrest 2.1 and throws a class not found exception on hamcrest 1.1

Without an explicit dependency it uses whatever it finds in .m2 and falls back to 1.1


benjamin-confino:
#build

benjamin-confino:
#build

Azquelt:
This shouldn't be necessary, the TCK already has a dependency on hamcrest which lists the version: https://github.com/eclipse/microprofile-fault-tolerance/blob/3.0/tck/pom.xml#L108-L113

Azquelt (changes requested):
This shouldn't be needed

benjamin-confino:
Andrew is going to fix this by fixing the build environment instead.

PR #22640, opened 2022-09-23
Update TCK Results output

tevans78:
#build

When the TCK adoc files were uploaded to RTC, they all had similar names and the content was getting mixed up. If you look at the TCK Results available in the Downloads tab of a build, the description given does not appear to match up with the contents of the file. e.g. A file that stated com.ibm.ws.concurrent.mp in the description could end up being the TCK results for MicroProfile Rest Client 1.2.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -258,6 +258,8 @@ public class LibertyServer implements LogMonitorClient {
 
     protected static final String SERVER_CONFIG_FILE_NAME = "server.xml";
     protected static final String JVM_OPTIONS_FILE_NAME = "jvm.options";
+    protected static final String OPENLIBERTY_PROPERTIES_FILE_NAME = "openliberty.properties";
+    protected static final String COM_IBM_WEBSPHERE_PRODUCTVERSION_KEY = "com.ibm.websphere.productVersion";
 
     protected static final String EBCDIC_CHARSET_NAME = "IBM1047";
 
@@ -311,6 +313,10 @@ public class LibertyServer implements LogMonitorClient {
     private boolean logOnUpdate = true;
 
     protected boolean debuggingAllowed = true;
+
+    private Properties openLibertyProperties;
+
+    private String openLibertyVersion;
 
     /**
      * This returns whether or not debugging is "programatically" allowed
@@ -6928,5 +6934,42 @@ public class LibertyServer implements LogMonitorClient {
             }
         }
         return checkpointSupported;
+    }
+
+    protected String getOpenLibertyPropertiesFilePath() {
+        return this.getInstallRoot() + "/lib/versions/" + OPENLIBERTY_PROPERTIES_FILE_NAME;
+    }
+
+    protected RemoteFile getOpenLibertyPropertiesFile() throws Exception {
+        return LibertyFileManager.createRemoteFile(this.machine, this.getOpenLibertyPropertiesFilePath());
+    }
+
+    public Properties getOpenLibertyProperties() {
+        if (this.openLibertyProperties == null) {
+            this.openLibertyProperties = new Properties();
+            InputStream is = null;
+            try {
+                is = getOpenLibertyPropertiesFile().openForReading();
+                this.openLibertyProperties.load(is);
+            } catch (Exception e) {
+                LOG.warning("Unable to read openliberty.properties file: " + e.getMessage());
+            } finally {
+                if (is != null) {
+                    try {
+                        is.close();
+                    } catch (IOException e) {
+                        LOG.warning("Unable to close input stream for openliberty.properties file: " + e.getMessage());
+                    }
+                }
+            }
+        }
+        return this.openLibertyProperties;
+    }
+
+    public String getOpenLibertyVersion() {
+        if (this.openLibertyVersion == null) {
+            this.openLibertyVersion = (String) getOpenLibertyProperties().get(COM_IBM_WEBSPHERE_PRODUCTVERSION_KEY);
+        }
+        return this.openLibertyVersion;
     }
 }
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -6944,6 +6944,9 @@ public class LibertyServer implements LogMonitorClient {
         return LibertyFileManager.createRemoteFile(this.machine, this.getOpenLibertyPropertiesFilePath());
     }
 
+    /**
+     * @return the contents of the wlp/lib/versions/openliberty.properties file
+     */
     public Properties getOpenLibertyProperties() {
         if (this.openLibertyProperties == null) {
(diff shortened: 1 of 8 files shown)


tevans78:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -6929,4 +6935,41 @@ public boolean getCheckpointSupported() {
...
+        }
+        return this.openLibertyProperties;
+    }
+
+    public String getOpenLibertyVersion() {
+        if (this.openLibertyVersion == null) {
+            this.openLibertyVersion = (String) getOpenLibertyProperties().get(COM_IBM_WEBSPHERE_PRODUCTVERSION_KEY);
+        }
+        return this.openLibertyVersion;
+    }
```
Azquelt: These need javadoc

It's particularly unclear what "getOpenLibertyProperties" means without reading the method and tracking through to the file name.

`getOpenLibertyVersion` should also state that it's reading from the product properties file and that the server doesn't need to be started.

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKResultsWriter.java
```diff
@@ -77,43 +78,42 @@ public static void preparePublicationFile(TCKResultsInfo resultInfo, String[] sp
...
-                                          "+\nDependency Injection\n",
+                                          "\nDependency Injection\n",
                                           "link:https://download.eclipse.org/jakartaee/dependency-injection/", "SHA-256:\n``\n",
-                                          "+\nDebugging\n",
+                                          "\nDebugging\n",
                                           "link:https://download.eclipse.org/jakartaee/debugging/", "SHA-256:\n``\n",
-                                          "* Java runtime used to run the implementation:\n+\n", javaVersion,
-                                          "\n\n* Summary of the information for the certification environment, operating system, cloud, ...:\n+\n",
+                                          "* Java runtime used to run the implementation:\n\n", javaVersion,
+                                          "\n\n* Summary of the information for the certification environment, operating system, cloud, ...:\n\n",
```
Azquelt: You've removed a bunch of `+` characters in this section.

I think those were necessary to keep the text from multiple lines attached to the same bullet point. See [list continuations](https://docs.asciidoctor.org/asciidoc/latest/lists/continuation/#list-continuation)

As an aside, this whole section with an array of document sections is nasty. It would probably be more readable if it was multiple calls to a `StringBuilder`.

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKResultsWriter.java
```diff
@@ -0,0 +1,135 @@
...
+
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.parsers.SAXParser;
+import javax.xml.parsers.SAXParserFactory;
+
+import org.xml.sax.SAXException;
+
+public class TCKResultsWriter {
+
+    public static void preparePublicationFile(TCKResultsInfo resultInfo, String[] specParts) {
```
Azquelt: This class and method could use javadoc.

The method in particular should document that it outputs a file and what that file will be named. Alternatively, you could pass in the output file as a parameter.

`specParts` as an array of strings also needs documentation (or refactoring to include the information into `TCKResultsInfo`)

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKResultsWriter.java
tevans78: When I ran this locally and looked at a preview view for these files, those `+` chars were showing up all over the place. I was intending to get rid of these String[]s in a follow on PR anyway.

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKResultsInfo.java
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package componenttest.topology.utils.tck;
+
+/**
+ *
+ */
+public class TCKResultsInfo {
```
Azquelt: Populate or remove Javadoc block (a one line description is probably all that's needed)

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKResultsWriter.java
tevans78: I was planning to merge that specParts stuff into the TCKResultsInfo object. So many things that could be improved here.

Azquelt (changes requested):
I think the removal of `+` characters is wrong and there are a few places which would benefit from Javadoc.

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKResultsWriter.java
Azquelt: Odd. They are in [existing published files](https://raw.githubusercontent.com/OpenLiberty/certifications/prod/microprofile/metrics/3.0/21.0.0.1-cl210120201202-1100-TCKResults.adoc) and do [appear to render correctly](https://openliberty.io/certifications/microprofile/metrics/3.0/21.0.0.1-cl210120201202-1100-TCKResults.html).

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKResultsWriter.java
```diff
@@ -0,0 +1,135 @@
...
+                                          ".\n\n== Open Liberty ", openLibertyVersion, " - MicroProfile ", specName, " ", specVersion,
+                                          " Certification Summary (Java ", javaMajorVersion, ")\n\n",
+                                          "* Product Name, Version and download URL (if applicable):\n",
+                                          "\nhttps://repo1.maven.org/maven2/io/openliberty/openliberty-runtime/",
+                                          openLibertyVersion, "/openliberty-runtime-", openLibertyVersion, ".zip[Open Liberty ", openLibertyVersion, "]\n",
+                                          "* Specification Name, Version and download URL:\n\n",
+                                          "link:https://download.eclipse.org/microprofile/microprofile-", MPSpecLower, "-", specVersion, rcVersion,
+                                          "/microprofile-", MPSpecLower, "-spec-", specVersion, rcVersion, ".html[MicroProfile ", specName, " ", specVersion,
+                                          rcVersion, "]\n\n* Public URL of TCK Results Summary:\n\n", "link:", filename, "[TCK results summary]\n\n",
```
Azquelt: In our existing files, the link here ends with `.html`, whereas with this change, `filename` will end with `.adoc`.

I'm unsure whether this difference will be handled when the file is rendered, but we need to check and change it if necessary.

Azquelt: Possibly we should use `xref` rather than `link`: https://docs.asciidoctor.org/asciidoc/latest/macros/inter-document-xref/

tevans78:
#build

Azquelt (approved):

tevans78:
#libby

PR #22645, opened 2022-09-23
Disable webserver plugin generation for checkpoint unless prop is true

### LINKED ISSUE #22390: Restore time regression observed when running pingperf in container environment
Between builds cl220820220629-1100 and cl220820220630-1100, I'm seeing a regression in containers when running pingperf with checkpoint restore. The first request time goes up from about 275ms on 0629-1100 to 350ms on 0630-1100. I do not see this same regression running on bare metal with no containers.

dettmoney:
In an InstantOn scenario, we don't need to generate the webserver plugin. It can take over 4x the time to do this when we're running in container as compared to running outside of a container. Added a property io.openliberty.checkpoint.generate.plugin to allow user to still generate plugin if desired.

### DIFF
--- a/dev/com.ibm.ws.webserver.plugin.runtime/src/com/ibm/ws/webserver/plugin/runtime/listeners/GeneratePluginConfigListener.java
+++ b/dev/com.ibm.ws.webserver.plugin.runtime/src/com/ibm/ws/webserver/plugin/runtime/listeners/GeneratePluginConfigListener.java
@@ -67,6 +67,8 @@ public class GeneratePluginConfigListener implements RuntimeUpdateListener, Appl
     private WsLocationAdmin locationService;
 
     private static GeneratePluginConfigListener theListener = null;
+    private static final String CHECKPOINT_GEN_PLUGIN = "io.openliberty.checkpoint.generate.plugin";
+    private boolean enablePlugin;
 
     public static GeneratePluginConfigListener getGeneratePluginConfigListener() {
         return theListener;
@@ -81,6 +83,7 @@ public class GeneratePluginConfigListener implements RuntimeUpdateListener, Appl
         if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled())
             Tr.debug(this, tc, "GPCL: activate called.");
         theListener = this;
+        enablePlugin = (CheckpointPhase.getPhase() == CheckpointPhase.INACTIVE || Boolean.valueOf(bc.getProperty(CHECKPOINT_GEN_PLUGIN)));
     }
 
     @Deactivate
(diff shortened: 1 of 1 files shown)


dettmoney:
#build

dev/com.ibm.ws.webserver.plugin.runtime/src/com/ibm/ws/webserver/plugin/runtime/listeners/GeneratePluginConfigListener.java
```diff
@@ -274,7 +277,9 @@ private void submitGeneratePluginTask() {
             Tr.debug(this, tc, "submitGeneratePluginTask : FrameworkState.isStopping() = " + FrameworkState.isStopping());
 
         ExecutorService currentExecutor = executorSrvc;
-        if (!FrameworkState.isStopping() && gpc != null && currentExecutor != null) {
+        final CheckpointPhase checkpoint = CheckpointPhase.getPhase();
```
tjwatson: While this is technically a smaller execution path, I find it confusing to put all this logic into a single if block here.

The code would be more clear to me if the original `if` statement was unchanged and instead you short-circuit below in the `if (checkpoint != CheckpointPhase.INACTIVE)` block below like this:

```java
            if (checkpoint != CheckpointPhase.INACTIVE) {
                if (Boolean.valueOf(bc == null ? null : bc.getProperty(CHECKPOINT_GEN_PLUGIN))) {
                    CheckpointHook restoreHook = new CheckpointHook() {
                        @Override
                        public void restore() {
                            currentExecutor.submit(generatePluginTask);
                        }
                    };
                    if (!checkpoint.addMultiThreadedHook(restoreHook)) {
                        currentExecutor.submit(generatePluginTask);
                    }
                }
            }
```

dev/com.ibm.ws.webserver.plugin.runtime/src/com/ibm/ws/webserver/plugin/runtime/listeners/GeneratePluginConfigListener.java
```diff
@@ -67,6 +67,8 @@ public class GeneratePluginConfigListener implements RuntimeUpdateListener, Appl
     private WsLocationAdmin locationService;
 
     private static GeneratePluginConfigListener theListener = null;
+    private static final String CHECKPOINT_GEN_PLUGIN = "io.openliberty.checkpoint.generate.plugin";
+    private BundleContext bc;
```
tjwatson: Instead of storing `bc` as an instance variable I would store
```java
private boolean generateOnRestore;
```
Then in `activate`
```java
    generateOnRestore=Boolean.valueOf(bc.getProperty(CHECKPOINT_GEN_PLUGIN));
```

Then use the `generateOnRestore` in the code.

tjwatson (approved):

dettmoney:
#build

tjwatson:
#libby

PR #22702, opened 2022-09-28
22701 simplicity rename

### LINKED ISSUE #22701: Simplicity 'LibertyFileManager.moveLibertyFile' attemps rename onto file on windows.
The implementation of 'moveLibertyFile' attempts to do a direct rename as its preferred implementation.

This is performed with full retries.

On windows, when the target file already exists, this always fails, leading to a 10s delay before moving to the secondary implementation, which does a file copy.

The file copy takes about 1s.

The implementation should be changed to skip the rename step when the destination file is on windows and already exists.

For an example of the impact of this change, when running single feature startup tests, approximately 260 startups are performed.
Each does a feature update, which invokes rename.  With the original implementation, that would add 2600s to running the entire FAT.  Even split among 4 buckets (65 feature startups per bucket), that is still an extra 650s.

(Note: The feature startups now use a different API.  However, the problem occurs for any rename attempt on windows, and may impact other FAT tests.)

tbitonti:
This is a correction to a bug in the test infrastructure (fattest.simplicity), and is tested by rerunning existing FATs.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyFileManager.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyFileManager.java
@@ -20,6 +20,7 @@ import java.util.regex.Pattern;
 
 import com.ibm.websphere.simplicity.LocalFile;
 import com.ibm.websphere.simplicity.Machine;
+import com.ibm.websphere.simplicity.OperatingSystem;
 import com.ibm.websphere.simplicity.RemoteFile;
 import com.ibm.websphere.simplicity.log.Log;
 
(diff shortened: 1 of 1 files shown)


tbitonti:
#build

tbitonti:
Note: The first two commits are expected to be added by a separate issue.  Only the third commit matters.

tbitonti:
#build

tbitonti:
Test failures:

- com.ibm.ws.automation.healing_fat FAT tests
  - com.ibm.ws.automation.test.HealingTest
    - testLibertyRunbook
- com.ibm.ws.jsf.2.3_fat FAT tests
  - com.ibm.ws.jsf23.fat.tests.JSF23WebSocketTests
    - testOpenAndCloseWebsocket_EE10_FEATURES

These test failures match RTC defects 289998 and 292790.

tbitonti:
#libby

jimblye (approved):

Issue #22396, opened 2022-09-06
FFDCIgnore when using `try-with-resources`


KyleAure:
This design issue is related to a regression where product code was changed from using a `try-catch` block to a `try-with-resources` block and FFDCIgnore no longer ignored the specified exception.

When this component is compiled the FFDC component does bytecode manipulation so that FFDC processing can happen at runtime. However, the FFDC component does not take into account what results from a `try-with-resources` block in java bytecode.  The `try-with-resources` block is replaced with `catch (Throwable)` blocks which are then reported as FFDCs.  

Here is the native method: 
```java
    @FFDCIgnore(JAXBException.class)
    public static Object parseRaDeploymentDescriptor(Entry ddEntry)
        throws JAXBException, SAXException, ParserConfigurationException, UnableToAdaptException {

        String ddName = ddEntry.getName();

        try {
            try ( InputStream ddStream = ddEntry.adapt(InputStream.class) ) {
                return parseResourceAdapterXml(ddStream, ddName, !IS_VERSION_10);
            } catch ( IOException e ) {
                throw new UnableToAdaptException(e);
            }

        } catch ( JAXBException jax ) {
            boolean isVersion10;
            try ( InputStream ddStream = ddEntry.adapt(InputStream.class) ) {
                isVersion10 = isVersion10ResourceAdapter(ddStream);
            } catch ( IOException e ) {
                throw new UnableToAdaptException(e);
            }
            if ( !isVersion10 ) {
                throw jax;
            }

            try ( InputStream ddStream = ddEntry.adapt(InputStream.class) ) {
                return parseResourceAdapterXml(ddStream, ddName, IS_VERSION_10);
            } catch ( IOException e ) {
                throw new UnableToAdaptException(e);
            }
        }
    }
```

Here is a decompiled method: 
```java
@FFDCIgnore({JAXBException.class})
  public static Object parseRaDeploymentDescriptor(Entry ddEntry) throws JAXBException, SAXException, ParserConfigurationException, UnableToAdaptException {
    String ddName = ddEntry.getName();
    try {
      InputStream ddStream = (InputStream)ddEntry.adapt(InputStream.class);
      try {
        Object object = parseResourceAdapterXml(ddStream, ddName, false);
        if (ddStream != null)
          ddStream.close(); 
        return object;
      } catch (Throwable throwable) {
[...]

PR #22705, opened 2022-09-28
Update FFDCIgnore Javadoc

Azquelt:
For #22396

During the discussion of `@FFDCIgnore` and try-with-resources, an update to the Javadoc was requested.

However, I noted while writing it that `@FFDCIgnore` is part of the logging SPI so I'm not sure if that changes anything about how we can change its behaviour or whether I should include a link back to the issue about the limitation with using try-with-resources.

### DIFF
--- a/dev/com.ibm.ws.logging.core/src/com/ibm/ws/ffdc/annotation/FFDCIgnore.java
+++ b/dev/com.ibm.ws.logging.core/src/com/ibm/ws/ffdc/annotation/FFDCIgnore.java
@@ -31,30 +31,61 @@ import java.lang.annotation.Target;
  * For example:<br>
  * 
  * <pre>
- * &#064;FFDCIgnore({ InterruptedException.class, NameAlreadyBoundException.class })
- * public void doSomeWork()
- * {
- * try
- * {
- * Thread.sleep(900);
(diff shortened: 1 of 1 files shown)


Azquelt:
I couldn't find a test for FFDC instrumentation. There's [SimpleTryCatchTest](https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.logging_test/test/com/ibm/ws/ras/instrumentation/test/SimpleTryCatchTest.java) which looks like it should be testing this, but when I tried running it the class isn't actually instrumented and no FFDCs are produced.

I manually tested the scenarios added to the Javadoc by inspecting the instrumented class file with `javap -c`.

wraschke:
#libby

jhanders34 (approved):

frowe (approved):
Probably should add a reminder to issue 22396 to update this javadoc when/if the issue is resolved

Azquelt:
#build

Azquelt:
#build

Azquelt:
Unusual failure in `com.ibm.ws.jmx.connector.bvt.JMXConnectorBVTTest` - don't think it's related to this PR but will run the build again to check.

#build

PR #22750, opened 2022-09-30
Don't expose 3rd party Weld packages that don't exist

### LINKED ISSUE #22748: Clean up 3rd party packages for CDI 2.0, 3.0 and 4.0
From @jhanders34

While looking at the output from the feature checker while working on a feature checker update for one of my PRs, I noticed that there were many informational messages about third party IBM-API-Packages that are not included in the product.  The two CDI ones I see are org.jboss.weld.context.beanstore and org.jboss.weld.context.conversation.  It looks like at some point I believe between weld 2 and 3 (CDI 1.2 and 2.0) these packages changed to have an s after the context subpackage.  Should those be removed from the IBM-API-Packages list for the cdi features?  Or possibly we need to update the third party jars to include the contexts version of the package and update the IBM-API-Packages where it changed to be contexts instead of context. If so can you put something on your backlog to get this cleaned up?

- [x] CDI-4.0
- [x] CDI-3.0
- [x] CDI-2.0

tevans78:
`org.jboss.weld.context.beanstore` and `org.jboss.weld.context.conversation` do not exist any more. They were renamed in Weld 3 but since no one has requested them, there is no reason to expose the newer packages.

fixes #22748 
#build


tevans78:
#build

tevans78:
resovled by https://github.com/OpenLiberty/open-liberty/pull/23383

PR #22761, opened 2022-09-30
infer more from parameterized types

njr-11:
With the decision not to have an annotation for limit, it needs to be possible to infer when there is a singular result from the result type, which might be parameterized.  This pull experiments with that, as well as with a number of other scenarios where we could use the parameterized type information to correctly infer how to form the result that the application wants to receive.


njr-11:
#build

nmittles (approved):

Issue #22369, opened 2022-09-01
Faces 4.0: Enable com.ibm.ws.jsf.2.2_fat tests for EE10


pnicolucci:
We need to investigate the following:
1) testMyFaces4433 -> https://github.com/OpenLiberty/open-liberty/pull/23845/files
    - Fixed in https://github.com/OpenLiberty/open-liberty/pull/24557
2) [testAjaxZeroDelay](https://github.com/OpenLiberty/open-liberty/blob/9185922ddd82d1b5a4f9300d7f71ecd6d32b1f81/dev/com.ibm.ws.jsf.2.2_fat/fat/src/com/ibm/ws/jsf22/fat/tests/JSF22ResetValuesAndAjaxDelayTests.java#L179) -> https://github.com/OpenLiberty/open-liberty/pull/24076/files
    - Ajax fails in with HTMLUnit, so this has been moved to https://github.com/openliberty/open-liberty/issues/24387

volosied:
## **testAjaxZeroDelay** 
This test looks at whether an ajax request is sent (with a delay of zero) for the "onkeyup" event. Each request is logged in the messages log. 

This may be a HTMLUnit bug as manual tests show the expected behavior.
My guess is that the "onkeyup" is somehow not registered in MyFaces using HTMLUnit.

Manual testing shows: 

http://localhost:9080/TestJSF22Ajax/ajaxZeroDelayTest.jsf

Input is “joh”

<input id=“form1:name” name=“form1:name” type=“text” value=“” onkeyup=“myfaces.ab(this,event,’keyup’,’’,’form1:employeeMatches’,{‘delay’:’0’})”>

```
XHRPOST http://localhost:9080/TestJSF22Ajax/ajaxZeroDelayTest.jsf
XHRPOST http://localhost:9080/TestJSF22Ajax/ajaxZeroDelayTest.jsf
XHRPOST http://localhost:9080/TestJSF22Ajax/ajaxZeroDelayTest.jsf
```

But with the FAT test, Faces 4.0 only sees it as one key press from what I can tell?

## **testMyFaces4433**  - JIRA CREATED: https://issues.apache.org/jira/projects/MYFACES/issues/MYFACES-4563

There appears to be a memory leak with our ViewScope logic (which is what the test looks for). This looks like a legitimate bug, but I only see it occur when the JSF trace is enabled.  I will make a JIRA and continue to investigate it.

pnicolucci:
[...]

PR #22773, opened 2022-10-02
Update copyright date ranges

pnicolucci:
for #22368 and #22369


pnicolucci:
#libby #build

volosied (approved):
Looking at the history, I don't see any other files missing a copyright update. 

I think you got them all.

PR #22779, opened 2022-10-03
align with Jakarta Data sorting

njr-11:
jakarta.data.repository.Sort and jakarta.data.repository.Direction were recently added to Jakarta Data.
Update our experimentation to align with the method signatures of those classes rather than Jakarta NoSQL's.


njr-11:
#build

tabishop (approved):

PR #22801, opened 2022-10-04
Refactored OpenAPIUIBundleUpdater to use ServiceEvents instead of pol…

KeenanMcCarthy:
…ling

Added a ServiceListener implementation that alerts OpenAPIUIBundleUpdater when all the expected OpenAPIUI bundles are ready


KeenanMcCarthy:
#build

PR #22802, opened 2022-10-04
address issue passing wrong x5c params object while getting ec jwk instance

### LINKED ISSUE #22298: fails to create Jose4jEllipticCurveJWK instance
OpenID Connect Client fails to validate JWT due to this error :
CWWKS1737E: The OpenID Connect client [clientid123] failed to validate the JSON Web Token. The cause of the error was: [CWWKS1739E: A signing key required by signature algorithm [ES256] was not available. ] …

Jose4jEllipticCurveJWK class is not storing the correct object type for “x5c” parameter, JWK therefore does not get successfully parsed and runtime fails to find the key to verify the JWT signature

arunavemulapalli:
- change the x5c params object from String to the expected type.
fixes #22298


arunavemulapalli:
#build

teddyjtorres (approved):

arunavemulapalli:
fixes #22298

PR #22812, opened 2022-10-04
verify idtoken nonce only when OpenIdAuthenticationMechanism useNonce is true

arunavemulapalli:
code update - verify nonce in the idtoken only when the config says it's required.


arunavemulapalli:
#build

teddyjtorres (approved):

arunavemulapalli:
#libby

PR #22834, opened 2022-10-05
align with Jakarta Data Limit parameter and fix package name

njr-11:
Update our experimentation code to switch Limit to a repository method parameter to match what is currently in Jakarta Data.
Replace MappingException which we copied from Jakarta NoSQL and instead use DataException that was recently added to Jakarta Data.
Also, our copy of jakarta.data.repository package classes is currently in the wrong package, so I'm also correcting that.


njr-11:
#build

njr-11:
#build

nmittles (approved):

njr-11:
#libby

PR #22837, opened 2022-10-05
PR for Issue 22726: Update custom cache key used in Jakarta Security 3.0

### LINKED ISSUE #22726: Ensure that we get a unique "state" value for each request
Testing @OpenIdAuthenticationMechanismDefinition:
My test class makes requests to unique applications.  I'm getting the same state for these requests.
We need to ensure that we get a unique value for the state on each request.

[npe.zip](https://github.com/OpenLiberty/open-liberty/files/9669216/npe.zip)

ayoho:
We were finding that the custom cache key we were generating as part of the Jakarta Security 3.0 flow was not unique between two different test scenarios. These changes update the custom cache key to be the hash code of the ID token since the ID token should be unique.

Resolves #22726


ayoho:
#build

arunavemulapalli (approved):

teddyjtorres (approved):

PR #22838, opened 2022-10-05
MicroProfile Metrics 5.0: Update to latest SmallRye Metrics and MP Metric dependencies  + FAT Tck

Channyboy:
#build
Related to https://github.com/OpenLiberty/open-liberty/issues/20018
Also https://github.com/OpenLiberty/open-liberty/issues/22808

Issue #22046, opened 2022-08-10
Update MicroProfile Fault Tolerance to work with MP Metrics 5.0


donbourne:
We don't expect any code changes will be needed for MP FT, since the metric types MP FT uses are still the same.

- [x] update to tolerate MP Metrics 5.0
- [x] test to ensure metrics function as expected

This relates to epic #20018

Azquelt:
#22063 has MP Fault Tolerance 4.0 tolerating MP60 and running tests with MP60 features (not passing yet because metrics is missing)

I'm waiting for the `mpMetrics-5.0` feature to be available to test with.

Channyboy:
In MicroProfile Metrics 5.0, changes include removal of the `MetricType` class and the `MetricRegistry.register(..)` methods.

See (RC2 is current as of this post) https://download.eclipse.org/microprofile/microprofile-metrics-5.0.0-RC2/microprofile-metrics-spec-5.0.0-RC2.html#release_notes_5_0

This would affect the `MetricRecorder30Impl`.
- `MetricType` is used during `Metadata` creation 
- Gauges whos value are provided by a `Supplier` are registered with the `MetricRegistry.register(...)`. The [gauge supplier](https://download.eclipse.org/microprofile/microprofile-metrics-5.0.0-RC2/apidocs/org/eclipse/microprofile/metrics/MetricRegistry.html#gauge-org.eclipse.microprofile.metrics.Metadata-java.util.function.Supplier-org.eclipse.microprofile.metrics.Tag...-) can be used instead.

Azquelt:
Further breaking changes
- `MetricRegistry.Type` has been removed, replaced by constants on `MetricRegistry`
- The `@RegistryType` qualifier has been replaced by `@RegistryScope`
  - This is used by the Fault Tolerance 4.0 TCK

Channyboy:
In MP Metrics 5.0 PR for noship deliveryhttps://github.com/OpenLiberty/open-liberty/pull/22715

- Removed `mpMetrics-5.0` from the `io.openliberty.mpFaultTolerance4.0-metrics.feature` auto-feature 
See: https://github.com/OpenLiberty/open-liberty/pull/22715/commits/3d40d96bd9744f24001a0890e06412fce4164a2e

- Temporarily disabling MP 6 repeat for the FT TCKs in `io.openliberty.microprofile.faulttolerance.4.0.internal_fat_tck`
[...]

PR #22839, opened 2022-10-06
Add mpMetrics 5.0 integration with Fault Tolerance

Azquelt:
Builds on #22838 and #22857

Add a new MetricRecorder implementation for the MP Metrics 5.0 API. Most of the code is shared with the existing implementation for MP Metrics 3.0+.

Use 4.0.2-RC1 TCK and enable repeat against MP 6.0.

For #22046


Azquelt:
#build

Azquelt:
Rebased on #22857 which should fix the high memory usage when using Histograms which caused an OOM error in the personal build.

Azquelt:
#libby
#build

Azquelt:
Updated the FT 4.0 + Metrics 5.0 autofeature to be `noship` - it should be set to `beta` at the same time as `mpMetrics-5.0`

Azquelt:
#libby
#build

Channyboy (approved):

Azquelt:
The 1017 personal build was good.

The earlier 1007 personal build failed due to an RTC connection error and not due to any test failures.

Issue #22477, opened 2022-09-13
Ensure work done by OpenAPIUIBundlesUpdater is not ongoing when doing a checkpoint


anjumfatima90:
The thread continually sleeps for 500 ms in a loop until the expected number of services is available in these three implementations of a method `waitForBundlesToStart(Set<Bundle> openAPIUIBundles)` in the following classes.
```
com.ibm.ws.openapi31.customization.OpenAPIUIBundlesUpdater
com.ibm.ws.microprofile.openapi.impl.OpenAPIUIBundlesUpdater
io.openliberty.microprofile.openapi20.internal.css.OpenAPIUIBundlesUpdater
```
The JVM Process looks like below:
```
4XESTACKTRACE                at java/lang/Thread.sleepImpl(Native Method)
4XESTACKTRACE                at java/lang/Thread.sleep(Thread.java:991)
4XESTACKTRACE                at java/lang/Thread.sleep(Thread.java:974)
4XESTACKTRACE                at io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesUpdater.waitForBundlesToStart(OpenAPIUIBundlesUpdater.java:336)
4XESTACKTRACE                at io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesUpdater.updateResources(OpenAPIUIBundlesUpdater.java:70)
5XESTACKTRACE                   (entered lock: io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesUpdater@0x0000000612293DF0, entry count: 1)
4XESTACKTRACE                at io/openliberty/microprofile/openapi20/internal/css/CustomCSSWABUpdater.restoreDefaults(CustomCSSWABUpdater.java:41)
4XESTACKTRACE                at io/openliberty/microprofile/openapi20/internal/css/CustomCSSProcessor.lambda$processCSSUpdates$0(CustomCSSProcessor.java:241)
5XESTACKTRACE                   (entered lock: java/lang/Object@0x00000006121D30F8, entry count: 1)
4XESTACKTRACE                at io/openliberty/microprofile/openapi20/internal/css/CustomCSSProcessor$$Lambda$220/0x0000000000000000.run(Bytecode PC:8)
4XESTACKTRACE                at java/util/concurrent/Executors$RunnableAdapter.call(Executors.java:515(Compiled Code))
4XESTACKTRACE                at java/util/concurrent/FutureTask.run(FutureTask.java:264)
[...]

PR #22878, opened 2022-10-07
Refactored OpenAPIUIBundlesUpdater in microprofile openapi 2.0

KeenanMcCarthy:
Modified OpenAPIUIBundlesUpdater to use a ServiceListener and ServiceEvents rather than polling, for #22477

### DIFF
--- a/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesListener.java
+++ b/dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesListener.java
@@ -1,0 +1,81 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.microprofile.openapi20.internal.css;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceEvent;
+import org.osgi.framework.ServiceListener;
+import org.osgi.framework.ServiceReference;
+
+/**
+ * ServiceListener implementation that alerts OpenAPIUIBundleUpdater when all
+ * the expected bundles are ready
+ */
+public class OpenAPIUIBundlesListener implements ServiceListener {
+
+    private final ConcurrentHashMap<String, Boolean> expectedBundleNames;
+    private final CountDownLatch countDownLatch;
+    private final BundleContext bundleContext;
+
+    public OpenAPIUIBundlesListener(Set<Bundle> openAPIUIBundles) throws InvalidSyntaxException {
+        this.bundleContext = FrameworkUtil.getBundle(OpenAPIUIBundlesUpdater.class).getBundleContext();
+        ConcurrentHashMap<String, Boolean> expectedBundleNames = new ConcurrentHashMap<>();
+        for (Bundle bundle : openAPIUIBundles) {
+            expectedBundleNames.put(bundle.getSymbolicName(), Boolean.TRUE);
+        }
+        this.expectedBundleNames = expectedBundleNames;
+        this.countDownLatch = new CountDownLatch(expectedBundleNames.size());
+    }
+
+    private void removeIfExpectedBundle(ServiceReference<Bundle> ref) {
+        String bundleKey = (String) ref.getProperty("web.module.key");
+        if (bundleKey != null) {
+            String bundleName = bundleKey.substring(0, bundleKey.indexOf('#'));
+            if (expectedBundleNames.remove(bundleName) != null) {
+                countDownLatch.countDown();
+            }
+        }
+    }
+
+    private void checkExistingServices(ServiceReference<Bundle>[] refs) {
+        if (refs != null) {
+            for (ServiceReference<Bundle> ref : refs) {
+                removeIfExpectedBundle(ref);
+            }
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public synchronized void serviceChanged(ServiceEvent event) {
+        ServiceReference<Bundle> ref = (ServiceReference<Bundle>) event.getServiceReference();
+        removeIfExpectedBundle(ref);
+    }
+
+    @SuppressWarnings("unchecked")
+    public void await() throws InterruptedException, InvalidSyntaxException {
+        bundleContext.addServiceListener(this, "(&(objectClass=org.osgi.framework.Bundle)(installed.wab.contextRoot=*))");
+        try {
+            ServiceReference<Bundle>[] refs = (ServiceReference<Bundle>[]) bundleContext.getServiceReferences(Bundle.class.getName(), "(installed.wab.contextRoot=*)");
+            checkExistingServices(refs);
+            countDownLatch.await();
+        } finally {
+            bundleContext.removeServiceListener(this);
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


KeenanMcCarthy:
#build

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesUpdater.java
```diff
@@ -313,28 +302,17 @@ private static void writeStreamToZip(InputStream stream, ZipOutputStream zos) th
...
-                }
-                refs = (ServiceReference<Bundle>[]) bundleContext.getServiceReferences(Bundle.class.getName(), "(installed.wab.contextRoot=*)");
-                waitComplete = verifyRefs(refs, expectedBundleNames);
-                if (!waitComplete) {
-                    //sleep UI bundle update thread to prevent getting services refs all the time
-                    Thread.sleep(500);
-                }
+            bundleContext.addServiceListener(new OpenAPIUIBundlesListener(expectedBundleNames));
+            synchronized (threadLock) {
+                threadLock.wait();
```
tjwatson: I suggest you use a CountDownLatch here and initialize it with the size of the expectedBundleNames.  Then pass that latch to the constructor of `OpenAPIUIBundlesListener`.  That way the listener can `countDown()` on the latch each time an expected bundle is found. and you can `await()` for it here.

There is also an issue with discovery of existing service registrations that match.  The listener is only going to see new events, it will not be able to detect any existing services here.  I suggest you add a method to the listener class (called `checkExistingServices` or similar) and that gets the existing services to check:

```
ServiceReference<Bundle>[] existingRefs = (ServiceReference<Bundle>[]) bundleContext.getServiceReferences(Bundle.class.getName(), "(installed.wab.contextRoot=*)");
```
And it checks reach existing ref and also does a `countDown()` when they are for one of the expected bundles.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesListener.java
```diff
@@ -0,0 +1,45 @@
...
+public class OpenAPIUIBundlesListener implements ServiceListener {
+
+    private final Set<String> expectedBundleNames;
+
+    public OpenAPIUIBundlesListener(Set<String> expectedBundleNames) {
+        this.expectedBundleNames = expectedBundleNames;
+    }
+
+    @Override
+    public void serviceChanged(ServiceEvent event) {
```
tjwatson: `serviceChanged` needs to be thread safe.  Right now I expect it is not with respect to manipulating the `expectedBundleNames` set.  Once you move to using a `CountDownLatch` you no longer are going to need to modify the set to remove the bundle names.  It will just be:

```java
            if (expectedBundleNames.contains(bundleName))
                expectedBundleNameLatch.countDown();
```
And you will not need any `unlockThread` method on the updater.

tjwatson: Actually you probably do need to still remove from a set otherwise you could risk doing a `countDown` multiple times for the same bundleName.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesListener.java
```diff
@@ -0,0 +1,62 @@
...
+        if (refs == null) {
+            return;
+        }
+        for (ServiceReference<Bundle> ref : refs) {
+            String bundleKey = (String) ref.getProperty("web.module.key");
+            String bundleName = bundleKey.substring(0, bundleKey.indexOf('#'));
+            if (expectedBundleNames.contains(bundleName)) {
+                expectedBundleNames.remove(bundleName);
+                countDownLatch.countDown();
+            }
```
tjwatson: Separate this out to a method that can be called from here and `serviceChanged`.

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesListener.java
```diff
@@ -0,0 +1,62 @@
...
+    private Set<String> expectedBundleNames;
+    private CountDownLatch countDownLatch;
+
+    public OpenAPIUIBundlesListener(Set<String> expectedBundleNames, CountDownLatch countDownLatch, ServiceReference<Bundle>[] refs) {
+        this.expectedBundleNames = expectedBundleNames;
+        this.countDownLatch = countDownLatch;
+        checkExistingServices(refs);
+    }
+
+    private void checkExistingServices(ServiceReference<Bundle>[] refs) {
```
tjwatson: This is just a style thing for me.  I would rather see this code inlined into the constructor:

```java
        if (refs == null) {
            for (ServiceReference<Bundle> ref : refs) {
                checkExpectedNames(ref);
            }
        }
```

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesListener.java
```diff
@@ -0,0 +1,62 @@
...
+import org.osgi.framework.ServiceListener;
+import org.osgi.framework.ServiceReference;
+
+/**
+ * ServiceListener implementation that alerts OpenAPIUIBundleUpdater when all
+ * the expected bundles are ready
+ */
+public class OpenAPIUIBundlesListener implements ServiceListener {
+
+    private Set<String> expectedBundleNames;
```
tjwatson: To make this thread safe this can be a `ConcurrentHashMap<String,Boolean>`.  Then use `expectedBundleNames.put(bundle.getSymbolicName(), Boolean.TRUE)` to add names and use the following to remove/countdown:

```java
            if (expectedBundleNames.remove(bundleName) != null) {
                countDownLatch.countDown();
            }
```

tjwatson: This is still an issue with the latest commit

KeenanMcCarthy:
#build
#libby

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesUpdater.java
```diff
@@ -310,39 +291,16 @@ private static void writeStreamToZip(InputStream stream, ZipOutputStream zos) th
...
-                    return false;
-                }
-                refs = (ServiceReference<Bundle>[]) bundleContext.getServiceReferences(Bundle.class.getName(), "(installed.wab.contextRoot=*)");
-                waitComplete = verifyRefs(refs, expectedBundleNames);
-                if (!waitComplete) {
-                    //sleep UI bundle update thread to prevent getting services refs all the time
-                    Thread.sleep(500);
-                }
-            }
+            new OpenAPIUIBundlesListener(openAPIUIBundles).await();
```
tjwatson, line 296: Need to return `true` here if await returns witihout exceptoin.

KeenanMcCarthy:
#libby

tjwatson (dismissed):

tjwatson:
#build

dev/io.openliberty.microprofile.openapi.2.0.internal/src/io/openliberty/microprofile/openapi20/internal/css/OpenAPIUIBundlesListener.java
```diff
@@ -0,0 +1,81 @@
...
+    @SuppressWarnings("unchecked")
+    @Override
+    public synchronized void serviceChanged(ServiceEvent event) {
+        ServiceReference<Bundle> ref = (ServiceReference<Bundle>) event.getServiceReference();
+        removeIfExpectedBundle(ref);
+    }
+
+    @SuppressWarnings("unchecked")
+    public void await() throws InterruptedException, InvalidSyntaxException {
+        bundleContext.addServiceListener(this);
```
tjwatson: This should be using a filter like `"(&(objectClass=org.osgi.framework.Bundle)(installed.wab.contextRoot=*))"`

Azquelt:
I am happy with this change once @tjwatson's last comment is addressed. Thanks.

tjwatson (approved):

tjwatson:
#build
#libby

PR #22929, opened 2022-10-12
Force @Interceptors instances to never be CDI Beans

tevans78:
When an interceptor class is declared via the @Interceptors annotation on an EJB, it should never be treated as a CDI Bean ... even if discovery-mode=all

#build

### DIFF
--- a/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/interceptors/InterceptorA.java
+++ b/dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/interceptors/InterceptorA.java
@@ -1,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.cdi.ejb.apps.interceptors;
+
+import javax.interceptor.AroundConstruct;
+import javax.interceptor.InvocationContext;
+
+public class InterceptorA extends Interceptor9 {
+
+    @AroundConstruct
+    private void aroundConstruct(InvocationContext ic) {
+        System.out.println(">InterceptorA aroundConstruct - " + this.hashCode());
+        try {
+            ic.proceed();
+        } catch (Exception e) {
+            // TODO Auto-generated catch block
+            // Do you need FFDC here? Remember FFDC instrumentation and @FFDCIgnore
+            // https://websphere.pok.ibm.com/~liberty/secure/docs/dev/API/com.ibm.ws.ras/com/ibm/ws/ffdc/annotation/FFDCIgnore.html
+            e.printStackTrace();
+        }
+        System.out.println("<InterceptorA aroundConstruct - " + this.hashCode());
+    }
+
+}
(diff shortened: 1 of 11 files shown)


dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/interceptors/InterceptorBase.java
```diff
@@ -0,0 +1,47 @@
...
+        Object res;
+        try {
+            res = ic.proceed();
+        } catch (Exception e) {
+            res = e;
+            e.printStackTrace();
+        }
+        System.out.println("<InterceptorBase aroundInvoke - " + this.hashCode());
+        return res;
+    }
```
Azquelt, line 45: Should probably rethrow these exceptions so that the test fails if an unexpected exception occurs.

Same in the other interceptor classes.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/interceptors/InterceptorsTestServlet.java
```diff
@@ -0,0 +1,31 @@
...
+@WebServlet("/")
+public class InterceptorsTestServlet extends FATServlet {
+
+    @Inject
+    private RequestScopedBean bean;
+
+    @Test
+    public void testEJB() {
+        bean.message();
+    }
```
Azquelt, line 29: Nothing is asserted.

I realise the the reported problem was that CDI threw an exception trying to find the interceptor, but it would be good to assert that the interceptors were called (i.e. the expected code path actually ran) or at least add a comment saying what this test is testing for.

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/tests/InterceptorsTest.java
```diff
@@ -0,0 +1,73 @@
...
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.rules.repeater.EERepeatActions;
+import componenttest.rules.repeater.RepeatTests;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+
+/**
+ * Testing that interceptors bound with an @Interceptors annotation on an ejb can be resolved
+ * even when they have a common type. In this case we force them not to be CDI beans.
+ */
```
Azquelt, line 38: I think a bit more detail about the important points of the test scenario would be helpful here.

```suggestion
/**
 * Testing that interceptors bound with an @Interceptors annotation on an ejb can be resolved
 * even when they have a common type.
 * <p>
 * In this test, although the interceptors are beans (due to bean-discovery-mode=all), we must not
 * look them up as beans because the classes are named directly in the {@code @Interceptors}
 * annotation.
 * <p>
 * In particular, type-safe resolution for {@code Interceptor9} would fail because both
 * {@code InterceptorA} and {@code Interceptor9} are beans with that type.
 */
```

dev/com.ibm.ws.cdi.ejb_fat/fat/src/com/ibm/ws/cdi/ejb/apps/interceptors/InterceptorA.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.cdi.ejb.apps.interceptors;
+
+import javax.interceptor.AroundConstruct;
+import javax.interceptor.InvocationContext;
+
+public class InterceptorA extends Interceptor9 {
```
Azquelt, line 16: These interceptor names seem a bit arbitrary.

Azquelt (approved):
Review comments to be addressed in #22962

tevans78:
#libby

PR #22935, opened 2022-10-12
Ensure tracer error doesn't prevent DataSource injection

KyleAure:
As part of our trace enablement, we attempt to create a proxy to the DataSource object so that we can trace method entries and exits.  However, it is possible for jdbc drivers to restrict access to certain classes/methods which would result in an IllegalArgumentException.

As it is today, the IllegalArgumentException is never caught and therefore will prevent DataSource injection.  
We should instead handle this exception case and log it as this should not prevent DataSource useage. 

This happens when using `WAS.database=all` trace.


jmstephensgit (approved):

KyleAure:
#build

PR #22945, opened 2022-10-13
check the hostname value before assigning to the dnsName

acdemyers:
On certs created during server startup check the hostname for characters not allowed in the dnsName part of the subject alternative name.


acdemyers:
#build

acdemyers:
#build

acdemyers:
#build

una-tapa (approved):
I chat with Alaine and agree with her evaluation that this this code area is rarely called 
  No performance concerns.  Approving. Thank you! 

        try {
            addr = InetAddress.getByName(hostname);
        } catch (Exception e) {
            //hostname likely does not resolve to an address
        }


Alaine >> Only when a cert is created and they specify a defaultHostName variable in the configuration.  That code path is only seen when a customize hostname value is set in the configuration.  I don't know how often that happens.

acdemyers:
#libbybot

PR #22948, opened 2022-10-13
Use https to access DHE

tevans78:
#build


jhanders34 (approved):

tevans78:
Merging without PB becuase this only affects local builds that are not using artifactory.

PR #22951, opened 2022-10-13
Remove BND Builder since there is no bnd file

tevans78:
The generated stub projects do not currently include a bnd file. Newer versions of the bnd builder fail to build anything if that file is missing. So for now (until a bnd file is added), remove the bnd builder from the template for the stub projects.
#build


jhanders34 (approved):

tevans78:
Merging without PB becuase this change only affects eclipse workspaces.

PR #22957, opened 2022-10-13
align with proposed page

njr-11:
Updates to more closely align with the current proposal for Page in Jakarta Data.


njr-11:
#build

tabishop (approved):
Some more comments on some of these classes would be ideal, but not holding up approval over it.

Issue #22960, opened 2022-10-14
trailing white-space in jvm.options files should be removed


urisimchoni:
**Describe the bug**  
The <wlp>/bin/server script processes jvm.options files, and creates a JVM argument from each line in those files. If a line in jvm.options file contains trailing spaces, then the space is considered to be part of the option. For example, if I include a line "-Dfoo=bar   " then System.getProperty("foo") yields "bar   ".

Depending on impact, this can create a problem that is difficult to troubleshoot. On the other hand, there is hardly any use for _trailing_ spaces.

The current behavior is to strip CR only.

**Expected behavior**  
The option should be stripped of trailing white space (CR, space, tab).

**Additional context** 
Whitespace is somewhat of a grey area when it comes to server configuration. The shell command line disregards them by default (unless they are explicitly quoted), so if the JVM arguments are configured by a shell script (such as with Tomcat for example), trailing whitespace is ignored. That is why admins might have this expectation here - that trailing whitespace be removed by default.

PR #22961, opened 2022-10-14
Remove trailing white space from jvm.options lines

urisimchoni:
This fixes issue #22960


urisimchoni:
The change is tested in an added FAT test

PR #22977, opened 2022-10-17
Update server.xml

### LINKED ISSUE #22965: Generating ssl key for FilterServer, when running FilterConfigTest takes too long
It takes too long to generate the ssl key, and as a result when stopping the FilterServer of the FilterConfigTest, the key is still being created. Consequently the process to delete server folder, which contains the key is blocked, by the key creation process.

mbuamuh:
Removed the keyStore entry from server.xml, as the test cases do not require this. Also it was unable to delete when repeating test cases, as when deleting the liberty server, the keystore creation process had not yet finished


benjamin-confino (approved):

mbuamuh:
#build

PR #23057, opened 2022-10-21
Keep uniqueId in URBridgeEntity

ryanesch:
When populating a DataObject in URBridgeEntity, there was a time when uniqueId was removed conditionally. At some point, the conditional part was removed and so uniqueId is always removed. This creates an issue when the uniqueId property is mapped to a different property, eg. uniqueName. 
In Traditional WebSphere, there was a way to set the mappings via custom properties. That is not yet possible in Liberty.


ryanesch:
#build

Issue #19172, opened 2021-10-28
Support MP Telemetry in Open Liberty


Emily-Jiang:
## Description of the high level feature, including any external spec links:  
With the merge of OpenTracing and OpenCensus, here comes [OpenTelemetry](https://opentelemetry.io/). OpenTracing is stabilised. All of the new activities are happening on OpenTelemetry. OpenTelemetry tracing is quite popular and got great support from the communities. 

At the moment, in MP, we are discussing different approaches at the weekly MP OT calls to get OpenTelemetry into MicroProfile. We need to start experiment in Open Liberty so that we can move onto the new technology asap. 

Sandbox proposal:
https://github.com/eclipse/microprofile-sandbox/blob/master/proposals/microprofile-telemetry/microprofile-telemetry-tracing.asciidoc

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Documents

When available, add links to required feature documents. Use "N/A" to mark particular documents which are not required by the feature.

- UFO: https://ibm.box.com/s/m6gabclow0ezgsar7zplpt91q5nfmhji
- FTS: #23143
- Beta Blog: #21847
- GA Blog: #24579

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Process Overview

- [Prioritization](#prioritization)
- [Design](#design)
- [Implementation](#implementation)
- [Legal and Translation](#legal-and-translation)
- [Beta](#beta)
- [GA](#ga)
  - [Focal Point Approvals](#focal-point-approvals-complete-by-feature-complete-date)
- [Other Deliverables](#other-deliverables)

### General Instructions

The process steps occur roughly in the order as presented. Process steps occasionally overlap.

Each process step has a number of tasks which must be completed or must be marked as not applicable ("N/A").

Unless otherwise indicated, the tasks are the responsibility of the Feature Owner or a Delegate of the Feature Owner.

If you need assistance, reach out to the [OpenLiberty/release-architect](https://github.com/orgs/OpenLiberty/teams/release-architect).

**Important: Labels are used to trigger particular steps and must be added as indicated.** 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## **Prioritization** (Complete Before Development Starts)

[...]

PR #23054, opened 2022-10-21
Updates to run MP Telemetry TCK RC5

Azquelt:
Updates and fixes to MP Telemetry to get the TCK running.

These changes should run the RC5 TCK, since then there have been changes to update to opentelemetry-java 1.19.0 which will require further changes in our implementation.

RC5 is not in maven central, so this PR also disables running the TCK in the build and is set up to run a TCK snapshot for local testing.

PR #23089, opened 2022-10-24
Update Telemetry to 1.19

Yas-Franco:
For issue #19172


Yas-Franco:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -65,11 +65,11 @@ public OpenTelemetry getOpenTelemetry() {
...
         if (oTelConfigs.get(ENV_ENABLE_PROPERTY) != null) {
-            if (oTelConfigs.get(ENV_ENABLE_PROPERTY).equals("true")) {
+            if (oTelConfigs.get(ENV_ENABLE_PROPERTY).equals("false")) {
                 return true;
             }
         } else if (oTelConfigs.get(CONFIG_ENABLE_PROPERTY) != null) {
-            if (oTelConfigs.get(CONFIG_ENABLE_PROPERTY).equals("true")) {
+            if (oTelConfigs.get(CONFIG_ENABLE_PROPERTY).equals("false")) {
                 return true;
             }
```
Azquelt: These checks need reviewed. Setting `otel.sdk.disabled = False` would result in telemetry being disabled.

These are supposed to be loaded from MP Config, so we should follow MP Config's rules for converting to a boolean.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat_tck/publish/tckRunner/tck/pom.xml
```diff
@@ -58,42 +58,42 @@
         <dependency>
             <groupId>org.eclipse.microprofile.telemetry.tracing</groupId>
             <artifactId>microprofile-telemetry-tracing-tck</artifactId>
-            <version>1.0-SNAPSHOT</version>
+            <version>1.0-RC6</version>
```
Azquelt, line 61: With this change you should also be able to enable the TCK running in the build. (Change the test mode from `EXPERIMENTAL` back to `LITE`)

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
Azquelt: This whole config bit should be looked at, let's raise a new issue for that.

Yas-Franco: https://github.com/eclipse/microprofile-telemetry/blob/main/tracing/spec/src/main/asciidoc/microprofile-telemetry-tracing-spec.asciidoc#tracing-enablement

The spec states that it would be enabled with `otel.sdk.disabled = False`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -28,8 +28,8 @@
 public class OpenTelemetryProducer {
 
     private static final String INSTRUMENTATION_NAME = "io.openliberty.microprofile.telemetry";
-    private static final String ENV_ENABLE_PROPERTY = "OTEL_EXPERIMENTAL_SDK_ENABLED";
-    private static final String CONFIG_ENABLE_PROPERTY = "otel.experimental.sdk.enabled";
+    private static final String ENV_ENABLE_PROPERTY = "OTEL_SDK_DISABLED";
+    private static final String CONFIG_ENABLE_PROPERTY = "otel.sdk.disabled";
```
Azquelt: The constant names should change to match the meaning of the property. I.e. `ENV_DISABLE_PROPERTY`.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
Azquelt: Yep, and that's the problem, with the current code it would be _disabled_ because `"False"` is not equal to `"false"`.

We'll raise a follow up issue to look again at this config part as we should be able to have MP Config do the conversion for us according to [it's rules](https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#_built_in_converters).

For now, can you change this code to use [Boolean.valueOf(String s)](https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#valueOf(java.lang.String)) to do the conversion?

Azquelt:
Now that #23054 is merged, this can be rebased on integration.

Yas-Franco:
#build

Azquelt (approved):

Yas-Franco:
#libby

PR #23090, opened 2022-10-24
Limit checkpoint cpus to avoid higher restore times

dettmoney:
If the number of cpus available at checkpoint time is high, it affects restore times. Since we don't care as much about how long checkpoint takes, it would be best to limit the cpus at checkpoint time, then allow the code to dynamically adjust at restore time to the number of cpus available at that time.


dettmoney:
#build

tjwatson (approved):

jhanders34 (approved):

gjdeval (approved):

dettmoney:
#build

PR #23093, opened 2022-10-24
Ensure DB Rotation replaces Username and Password

KyleAure:
Database rotation cannot assume where a test wants the username/password to be modified. 
Therefore, these servers need to be updated to have a username and password set for derby.


KyleAure:
#build

jgrassel (approved):
lgtm

KyleAure:
The personal build was successful. 
This is a fix for the DB Rotation build, wait to merge until buildcon is > 1

PR #23102, opened 2022-10-25
MP Telemetry: Current Baggage and span FATs

Azquelt:
- Add tests for getting the current Baggage and Span
- Initialise `Context` early to avoid SecurityManager exceptions
- Tidy up the FAT project a bit


Yas-Franco (approved):
Thanks Andrew- All looks good

Azquelt:
#build
#libby

PR #23106, opened 2022-10-25
Add issuerIdentifier to OIDC OP config to fix BCL beta behavior

ayoho:
Back-channel logout tokens that are created based on the ID token provided in the end_session request in some FATs will have an `"iss"` claim value that will not match the expected issuer. The ID tokens are created with `"http://localhost:8910/oidc/endpoint/OidcConfigSample"` as the issuer (HTTP), but the logout token creation code expects the issuer to be `"https://localhost:8920/oidc/endpoint/OidcConfigSample"` (with HTTPS).

This updates the code so the issuer will always be the HTTP version, making the GA and beta versions of the tests happy.

**Note: ** As the comment suggests, this should be removed or at least revisited when back-channel logout code is moved to GA.


ayoho:
#build

teddyjtorres (approved):

PR #23108, opened 2022-10-25
Allow social login FAT to find WASOidcSession cookie in beta mode

ayoho:
Beta mode currently creates a WASOidcSession cookie for OIDC clients as part of the back-channel logout support that's still in beta. That cookie happens to be left around until the client performs a logout, which the `LibertyOP_CookieVerificationTests.test_verifyDefaultCookiesDeleted` test doesn't do. So in the beta run of the test, there will be an extra WASOidcSession cookie at the end of the login flow that isn't there in the GA version of the product.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat.LibertyOP.1

jimmy1wu (approved):

PR #23125, opened 2022-10-26
Allow thread pool controller to shrink pool under heavy load

gjdeval:
In an OCP environment, with a small CPU allocation and under heavy load, the thread pool does not find the optimal size, staying larger than it should. This PR is to address that issue.


jhanders34 (approved):

gjdeval:
#build

gjdeval:
#libby

PR #23166, opened 2022-10-28
Additional check for empty beans.xml and set version on internal beans.xml

tevans78:
1. Add extra check for an empty beans.xml in case the undocumented Weld SPI function is not enough
2. Make sure that the version is properly set on all internal bundle beans.xml files

#build

### DIFF
--- a/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
+++ b/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
@@ -733,14 +733,24 @@ public class WebSphereCDIDeploymentImpl implements WebSphereCDIDeployment {
                 continue;
             }
             //check if the file is really empty ... just whitespace like a single space would cause Weld to complain anyway
-            try (InputStream is = parsedURL.openStream()) {
+            InputStream is = null;
+            try {
+                is = parsedURL.openStream();
                 if (is.available() == 0) {
                     //file is empty
                     continue;
                 }
-            } catch (IOException e) {
+            } catch (IOException e1) {
                 //could not read the file, assume it is empty
                 continue;
+            } finally {
+                if (is != null) {
+                    try {
+                        is.close();
+                    } catch (IOException e) {
+                        //FFDC and ignore
+                    }
+                }
             }
 
             //if the beans.xml was not an empty file then check if the version was set or not
--- a/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
+++ b/dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package com.ibm.ws.cdi.impl.weld;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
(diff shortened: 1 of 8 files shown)


dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
```diff
@@ -725,11 +727,21 @@ public Collection<URL> getUnversionedBeansXmlURLs() {
...
+            //check if the file is really empty ... just whitespace like a single space would cause Weld to complain anyway
+            try (InputStream is = parsedURL.openStream()) {
+                if (is.available() == 0) {
+                    //file is empty
+                    continue;
+                }
+            } catch (IOException e) {
+                //could not read the file, assume it is empty
+                continue;
+            }
```
Azquelt: Need FFDCIgnore here?

Note that you need to ignore `Throwable.class` when using try-with-resources.

tevans78: I can't think of a legitimate reason why we would get an IOException at this point ... so I think we want the FFDC?

dev/com.ibm.ws.microprofile.faulttolerance.2.0.cdi/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+    xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_2_0.xsd"
+    version="2.0" bean-discovery-mode="all">
+
+    <scan>
+        <exclude
+            name="com.ibm.ws.microprofile.faulttolerance.cdi20.*" />
+    </scan>
```
Azquelt, line 10: It seems odd to be setting `bean-discovery-mode=all` and then excluding all the classes within it from scanning. Should this not be `bean-discovery-mode=none`?

Might even be able to remove it since there's an extension in there.

I hate our CDI extension mechanisms :disappointed:

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
Azquelt: You still want to ignore `Throwable.class` because you're using try-with-resources.

The compiler generates synthetic catch blocks and the instrumentation isn't smart enough to tell the difference yet.

dev/com.ibm.ws.microprofile.faulttolerance.2.0.cdi/resources/META-INF/beans.xml
tevans78: There are a great many bundles that do it this way. I didn't fancy changing all of those today. So I opted for consistency with FT 1.0

tevans78:
#build

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/WebSphereCDIDeploymentImpl.java
```diff
@@ -733,14 +733,24 @@ public Collection<URL> getUnversionedBeansXmlURLs() {
...
                 //could not read the file, assume it is empty
                 continue;
+            } finally {
+                if (is != null) {
+                    try {
+                        is.close();
+                    } catch (IOException e) {
+                        //FFDC and ignore
+                    }
+                }
```
Azquelt, line 753: Unsure why you've changed this, the older version was better, though the new one does not need an `@FFDCIgnore` annotation.

tevans78:
#build

tevans78:
#build

Azquelt (approved):

tevans78:
#libby

PR #23222, opened 2022-11-01
Move JSF 2.3 Spec Issue tests to their own test class

### LINKED ISSUE #23221: Move JSF 2.3 spec issue tests to their own test class
We're seeing a number of failures in the JSF23CDIGeneralTests failing to start all applications on slow hardware. As such I'm moving some of the tests, specifically the spec issue tests and the apps they use to a new test class to try and avoid the problem.

pnicolucci:
fixes #23221


pnicolucci:
#libby #build

volosied (approved):

PR #23240, opened 2022-11-02
Split out OIDC FAT test cases for discovery errors into their own classes

ayoho:
Splits up the `com.ibm.ws.security.openidconnect.client.fat.IBM.OidcClientDiscoveryErrorTests` class so that test cases that reconfigure the servers are put into their own classes. This should alleviate some build failures with FFDCs, timeouts, and error messages leaking between tests sometimes.


ayoho:
#build

c00crane (approved):

ayoho:
Child build in the personal build had a connection issue. The affected FAT ran in both lite and full mode and passed in other child builds, so this is good to go.

PR #23241, opened 2022-11-02
Fix scanning order of CDI archives

### LINKED ISSUE #23252: AmbiguousResolutionException when same class is present twice and certain features are used
**Describe the bug**  
When a class is present twice in an application and is used as a CDI managed bean and certain features are used, CDI may sometimes report an `AmbiguousResolutionException`.

```
Caused by: org.jboss.weld.exceptions.AmbiguousResolutionException: WELD-001318: Cannot resolve an ambiguous dependency between: 
  - Managed Bean [class com.ibm.wssvt.acme.annuity.common.business.jaxrs.MyJAXBResolver] with qualifiers [@Any @Default],
  - Managed Bean [class com.ibm.wssvt.acme.annuity.common.business.jaxrs.MyJAXBResolver] with qualifiers [@Any @Default]
```

**Steps to Reproduce**
- Enable the mpFaultTolerance feature (any version)
- Create an EAR containing a .jar in the lib directory and a .war.
- Place the same CDI managed bean class inside both the .jar and the .war
- Note that the class can be loaded fine (you will always load it from the .jar)
- Attempt to inject and call the CDI managed bean
- An exception is thrown, reporting that the bean cannot be resolved because it is ambiguous

Note: the exception depends on the exact order that CDI scans the .jar and the .war, so the error does not happen every time.

**Expected behavior**  
- Assuming normal parent first classloading, all parts of the application will always see the version of the class from the .jar
- Therefore CDI should ignore the class in the .war
- There should not be a bean resolution exception

**Diagnostic information:**  
 - OpenLiberty Version: all
 - Affected feature(s): `cdi-1.2` or later  along with one or more of `mpFaultTolerance` (all versions), `mpGraphQL` (allVersions), `mpReactiveMessaging-1.0`, `mpRestClient-3.0` or `appSecurity-5.0`
 - Java Version: n/a
 - server.xml configuration: n/a

**Additional context**  

There are assumptions in the CDI code that more widely visible parts of the application (e.g. ear libs) are scanned before less widely visible parts (e.g. war classes). When a feature which includes a CDI extension which has visibility of all application BDAs is included, these assumptions are violated and the same bean can be incorrectly included twice.

Azquelt:
When we scan CDI archives, we try to scan all the visible archives
first. This should make sure we scan things which are more widely
visible (like shared libs) before things which aren't (like wars).

However, runtime extensions can break this as some of them have
visibility of all BDAs.

Fixes #23252

### DIFF
--- a/dev/com.ibm.ws.cdi.visibility_fat/bnd.bnd
+++ b/dev/com.ibm.ws.cdi.visibility_fat/bnd.bnd
@@ -37,7 +37,9 @@ tested.features:\
 	cdi-4.0,\
 	connectors-2.1,\
 	jakartaee-10.0,\
-	servlet-6.0
+	servlet-6.0,\
+	mpConfig-3.0,\
+	mpFaultTolerance-4.0
 
 -buildpath: \
 	com.ibm.websphere.javaee.ejb.3.2;version=latest,\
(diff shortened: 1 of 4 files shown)


Azquelt:
#build

tevans78 (approved):

PR #23283, opened 2022-11-07
Streamable and page count for query

njr-11:
Add the Streamable interface from Jakarta Data and add/update examples using it with pagination.
Add the count query for `@Query` from Jakart Data and add/update example using it with page counts and total across pages.


njr-11:
#build

tabishop (approved):

njr-11:
Build has 1 failure which is a duplicate of an unrelated pre-existing issue.

PR #23286, opened 2022-11-07
Add daylight savings skip to JPA beanvalidation FAT

dazey3:
The JPA bean validation FAT tests `javax.validation.constraints.Future` by creating `java.time.Instant` values that are ahead of `java.time.Instant.now()` by 100000 seconds.

During the recent daylight savings switch, these bean validation tests failed on Hibernate as the future time was generated to be across the daylight savings boundary, even though the test was executing before daylight savings.

For now, we are going to make these tests skip if running across the daylight savings boundary on Hibernate.

Signed-off-by: Will Dazey <wadazey@us.ibm.com>


dazey3:
#build

jgrassel (approved):
lgtm

PR #23308, opened 2022-11-09
Updated websockets expected to allowed FFDC

### LINKED ISSUE #23307: Update newly added websocket test
Update testAddEndpointOutisdeDeploymentPhase to use AllowedFFDC instead of ExpectedFFDC due to the not on ZOS rule in the tests which causes the tests not to run but the annotations still run either way.

isaacrivriv:
Move newly added test `testAddEndpointOutisdeDeploymentPhase` to use allowed FFDC. This is due to the tests running on ZOS with a rule that skips the tests from running but the annotation still runs and fails.

Fixes #23307


isaacrivriv:
#build

isaacrivriv:
#build

pnicolucci (approved):

PR #23320, opened 2022-11-10
Need to make the file a different size so config update will be detected

acdemyers:
Test case update to make sure the configuration files change size so it will be detected for a dynamic config update.


acdemyers:
#build

acdemyers:
fixes #23321

wrodrig (approved):

PR #23323, opened 2022-11-10
Jakarta Security 3.0: Custom cache key fix for refresh token flow

ayoho:
Updates the `OidcHttpAuthenticationMechanism` class to use the access token hash code as the new custom cache key if an ID token isn't available, since the refresh token flow doesn't necessarily produce an ID token.


ayoho:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat,io.openliberty.security.jakartasec.3.0.internal_fat.config

jimmy1wu (approved):

PR #23324, opened 2022-11-11
Jakarta Security Refresh tests

c00crane:
Create Jakarta Security refresh tests.
Authenticate and access an app, wait for tokens to expire and try to access the app again.
When tokenAutoRefresh=true in the annotation and the OP includes a refresh token, refresh the token if either/both the access or id_token is expired.
Test that with tokenAutoRefresh=true and not refresh token from the OP, we'll logout, re-auth or request another login as appropriate.


c00crane:
#libby

c00crane:
#libby
#build

dev/io.openliberty.security.jakartasec.3.0.internal_fat.refresh/fat/src/io/openliberty/security/jakartasec/fat/refresh/tests/BasicRefreshTests.java
```diff
@@ -0,0 +1,865 @@
...
+    /**
+     * This test will expect to use a refreshed token to access the protected app the second time. tokenAutoRefresh is true, the OP includes refresh tokens
+     * and both the id and access tokens have expired. The notifyProvider and the expiry settings should not matter in this case.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void BasicRefreshTests_tokenAutoRefreshTrue_providerAllowsRefreshTrue_notifyProviderTrue_idTokenExpiryTrue_accessTokenExpiryTrue() throws Exception {
+
+        genericGoodRefreshTest(true, true, true, true, true, TokenWasRefreshedSecondTime);
```
ayoho, line 476:
```suggestion
        genericGoodRefreshTest(TokenAutoRefresh, true, NotifyProvider, IDTokenHonorExpiry, AccessTokenHonorExpiry, TokenWasRefreshedSecondTime);
```
Would make this a little clearer passing the constants you have defined so we don't get lost trying to figure out which boolean is supposed to mean what. Same goes for all calls.

ayoho (approved):

dev/io.openliberty.security.jakartasec.3.0.internal_fat.refresh/fat/src/io/openliberty/security/jakartasec/fat/refresh/tests/BasicRefreshTests.java
c00crane: will update on the next delivery

PR #23325, opened 2022-11-11
Tolerate null ID token in token response for refresh flows

ayoho:
Adds fix to avoid checks on an ID token that might have null claims.


ayoho:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat,io.openliberty.security.jakartasec.3.0.internal_fat.config

c00crane (approved):

PR #23330, opened 2022-11-11
Add tests using OIDC clients and annotations in one server

c00crane:
Tests use 2 servers
OP - with a couple providers
RP - with the openidConnectClient-1.0 feature, multiple clients and multiple test apps.
The RP will have multiple test apps that use the @OpenIdAuthenticationMechanismDefinition annotation to protect them and a couple test apps that are either not protected or are protected by openidconnect client configs in the server.xml.
These tests will show that the apps that should be protected by the openidconnect client are protected by it (validate that the provider is OP2 and that the clients used are client_98 and client_99)
For the tests that protected by the annotation, make sure that OP1 is the provider issuing the tokens.
Show that the annotation takes precedence over the openidconnect client config in the server.


c00crane:
#build

c00crane:
#libby
#build

dev/io.openliberty.security.jakartasec.3.0.internal_fat/fat/src/io/openliberty/security/jakartasec/fat/tests/BasicOIDCAnnotationWithOidcClientConfigTests.java
```diff
@@ -45,6 +55,88 @@ public static void setUp() throws Exception {
...
+        //show that we get to the test app without having to log in
+
+        Page response = invokeAppReturnLoginPage(webClient, url);
+
+        response = actions.doFormLogin(response, Constants.TESTUSER, Constants.TESTUSERPWD);
+        // confirm protected resource was accessed
+        Expectations expectations = new Expectations();
+        expectations.addSuccessCodeForCurrentAction();
+        expectations.addExpectation(new ResponseUrlExpectation(Constants.STRING_CONTAINS, url, "Did not land on the RP Protected test application."));
+        expectations.addExpectation(new ResponseFullExpectation(Constants.STRING_CONTAINS, "\"client_98\"", "Did not find the proper client which should be client_99."));
```
ayoho, line 136:
```suggestion
        expectations.addExpectation(new ResponseFullExpectation(Constants.STRING_CONTAINS, "\"client_98\"", "Did not find the proper client which should be client_98."));
```

ayoho (approved):

dev/io.openliberty.security.jakartasec.3.0.internal_fat/fat/src/io/openliberty/security/jakartasec/fat/tests/BasicOIDCAnnotationWithOidcClientConfigTests.java
c00crane: updating with the userinfo tests.

PR #23346, opened 2022-11-14
Update tests to use constants instead of true/false

c00crane:
The Jakarta Security logout and refresh tests use common test test methods when the expected results are "common".
This means we expect a logout, token refresh, access to an app without logging in, ....  The test cases will pass in a series of true/false values that the common method will use to build the app name to call.
Instead of passing in a series of true/false, we'll use values like:
`genericGoodRefreshTest(TokenAutoRefresh, ProvderAllowsRefresh, NotifyProvider, IDTokenHonorExpiry, AccessTokenHonorExpiry, TokenWasRefreshedSecondTime);`
instead of:
`genericGoodRefreshTest(true, true, true, true, true, true);`

That will be more descriptive.

```
    protected static final boolean TokenWasRefreshed = true;
    protected static final boolean TokenWasNotRefreshed = false;

    protected static final boolean NotifyProvider = true;
    protected static final boolean DoNotNotifyProvider = false;
    protected static final boolean IDTokenHonorExpiry = true;
    protected static final boolean IDTokenDoNotHonorExpiry = false;
    protected static final boolean AccessTokenHonorExpiry = true;
    protected static final boolean AccessTokenDoNotHonorExpiry = false;

    protected static final String goodRedirectUri = "goodRedirectUri";
    protected static final String badRedirectUri = "badRedirectUri";
    protected static final String emptyRedirectUri = "emptyRedirectUri";

```


dev/io.openliberty.security.jakartasec.3.0.internal_fat.refresh/fat/src/io/openliberty/security/jakartasec/fat/refresh/tests/BasicRefreshTests.java
```diff
@@ -787,64 +779,64 @@ public void BasicRefreshTests_tokenAutoRefreshFalse_providerAllowsRefreshTrue_no
     @Test
     public void BasicRefreshTests_tokenAutoRefreshFalse_providerAllowsRefreshTrue_notifyProviderFalse_idTokenExpiryFalse_accessTokenExpiryFalse() throws Exception {
 
-        genericGoodEndToEndReAccessTest(false, true, false, false, false);
+        genericGoodEndToEndReAccessTest(DoNotTokenAutoRefresh, ProvderDoesNotAllowsRefresh, DoNotNotifyProvider, IDTokenDoNotHonorExpiry, AccessTokenDoNotHonorExpiry);
```
ayoho:
```suggestion
genericGoodEndToEndReAccessTest(DoNotTokenAutoRefresh, ProvderAllowsRefresh, DoNotNotifyProvider, IDTokenDoNotHonorExpiry, AccessTokenDoNotHonorExpiry);
```

ayoho (changes requested):

dev/io.openliberty.security.jakartasec.3.0.internal_fat.refresh/fat/src/io/openliberty/security/jakartasec/fat/refresh/tests/BasicRefreshTests.java
c00crane: `BasicRefreshTests_tokenAutoRefreshFalse_providerAllowsRefreshTrue_notifyProviderFalse_idTokenExpiryFalse_accessTokenExpiryFalse `
and 
`BasicRefreshTests_tokenAutoRefreshFalse_providerAllowsRefreshFalse_notifyProviderFalse_idTokenExpiryFalse_accessTokenExpiryFalse`
both use genericGoodEndToEndReAccessTest, but to be complete/correct, it should be passing true/ProvderAllowsRefresh (but this does explain why the tests passed when I ran it locally with false)

c00crane:
#libby
#build

ayoho (approved):

PR #23371, opened 2022-11-15
add jaxrs trace for JAXRS20ClientAsyncInvokerTest

WhiteCat22:
I believe [this line](https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.jaxrs.2.0.client_fat/test-applications/bookstore/src/com/ibm/ws/jaxrs20/fat/bookstore/AsyncInvokerTestServlet.java#L454)  isn't timing out like it should. I have no idea why this only occurs on some platforms and not others.

Adding trace to the test server so that we can get trace from failing builds.


WhiteCat22:
#libby

jim-krueger (approved):

PR #23386, opened 2022-11-16
Deliver Jakarta userinfo tests

c00crane:
Create tests for jakarta userinfo function.
If subject or group values are not found in either the access or id token, look for them in the output from userinfo.
I've created tests for the following:
The callerNameClaim and callerGroupsClaim used are the "default" (Liberty) values - the OP should be including these claims in the tokens that it creates.  I have userinfo endpoints that will set either valid or invalid values.  The tests will show that in this case, we don't use the values from userinfo as we can get them from the tokens created by the OP.
The callerNameClaim or callerGroupsClaim used are something other than the "default" (Liberty) values - the OP will NOT include these claims in the tokens that it creates.  I have userinfo endpoints that will set either valid or invalid values.  The tests show that in this case, we do use the values from userinfo.  When the values from userinfo are valid, we should be granted access to the test app.  When the groups are invalid, we will not be granted access to the app.  When the Name/sub is not valid, the test will still be granted access because in this case, the groups are valid.


c00crane:
#libby
#build

dev/io.openliberty.security.jakartasec.3.0.internal_fat.config/test-applications/UserInfo.war/src/userinfo/servlets/UserInfoServlet.java
```diff
@@ -0,0 +1,161 @@
...
+
+        return "testuser";
+    }
+
+    protected JsonObject getMinimumClaims(String accessToken) {
+        JsonObjectBuilder claims = Json.createObjectBuilder();
+        claims.add(getSub(), createSub());
+        claims.add(getGroupId(), createGroups());
+        // TODO - hard coded port - not good.....
+        claims.add("iss", "https://localhost:8920/oidc/endpoint/OP1");
```
ayoho, line 125: The hard coded bit doesn't really matter; we're not validating any of claims in the User Info response beyond trying to look up the caller name and caller group. We'll verify the signature of the JWT if the User Info is in JWT format, but we don't care about any of the claims or their values.

dev/io.openliberty.security.jakartasec.3.0.internal_fat.config/build.gradle
```diff
@@ -85,7 +85,11 @@ autoFVT.doLast {
   
      def rpCopyServers = [
 	'jakartasec-3.0_fat.config.rp.claimsDef',
-	'jakartasec-3.0_fat.config.rp.ELOverride'
+	'jakartasec-3.0_fat.config.rp.ELOverride',
+	'jakartasec-3.0_fat.config.rp.userinfo.jwt.jwt',
+	'jakartasec-3.0_fat.config.rp.userinfo.jwt.opaque',
+	'jakartasec-3.0_fat.config.rp.userinfo.jose.jwt',
+	'jakartasec-3.0_fat.config.rp.userinfo.jose.opaque'
```
ayoho, line 92:
```suggestion
	'jakartasec-3.0_fat.config.rp.userinfo.json.jwt',
	'jakartasec-3.0_fat.config.rp.userinfo.json.opaque'
```
The User Info responses are in JSON format, so I'd use "json" in the server names. JOSE is a standard for signing and encryption, which is what the JWT formats would use. Using "jose" to refer to the JSON-formatted stuff would be confusing.

ayoho (approved):

dev/io.openliberty.security.jakartasec.3.0.internal_fat.config/build.gradle
c00crane: not sure why I did that - I really did mean to call them json.  
Will update in my next delivery.

dev/io.openliberty.security.jakartasec.3.0.internal_fat.config/test-applications/UserInfo.war/src/userinfo/servlets/UserInfoServlet.java
c00crane: I can update the comment - hate hard coded values.
I can look at adding a put method that gets passed the real port from the test setup and then uses that value to create the claim later.  But, that's not very high on my list at the moment.

PR #23404, opened 2022-11-17
23403: Http/2 Quiesce problem

### LINKED ISSUE #23403: HTTP/2 Intermittent server quiesce failure when stream is closed with an exception
Describe the bug
Intermittent issue where threads hang when an exception happens on an HTTP/2 connection.   Because the threads are hung, when the server is stopped, a quiesce warning will be issued.

Steps to Reproduce
Cannot be reproduced at will, but can happen when invalid frames are received on stream 0 of an HTTP/2 connection while other streams are open on that connection.

Expected behavior
HTTP/2 streams and connections and the threads managing them should stop without error, even in exception conditions.

Diagnostic information:
The following messages may be seen at server stop.  The number of hung threads will be a multiple of 2.

CWWKE1102W: The quiesce operation did not complete. The server will now stop.
CWWKE1106W: 1 shutdown operations did not complete during the quiesce period. 
CWWKE1107W: 2 threads did not complete during the quiesce period.

OpenLiberty Version: 18.0.0.2 (first release which suppoted http/2 ?) - current

Affected feature(s): Connections using HTTP/2.

Java Version: n/a

server.xml configuration: n/a

Additional context: n/a

loriadi:
fixes #23403 

There are threads working on stream 0 and stream 1 on a connection.  

Thread A is handling a bad frame on stream 0, it enters the linkStatusSync block in H2Connection.java and proceeds to shutdown the connection from there.  This block of code does a huge amount of work inside the sync block, including sending a goaway frame synchronously on the connection and closing out all the other streams and links from there.  As part of this, it tries to enter sync block (WebConnCanCloseSync) in HttpDispatcherLink, but it can't because...

Thread B is finishing up a response on stream 1.  It has sent headers for the response and is now trying to send the last data frame.  It is trading execution with thread A.  Thread B enters finish() which enters the WebConnCanCloseSynd block and then processNextFrame() which calls IsStreamClosed() which calls checkIfGoawaySending().  checkIfGoawaySending() should return true, but it can't enter because Thread A is holding the linkStatusSync block.

Deadlock.

### DIFF
--- a/dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/h2internal/H2InboundLink.java
+++ b/dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/channel/h2internal/H2InboundLink.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 1997, 2021 IBM Corporation and others.
+ * Copyright (c) 1997, 2022 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


loriadi:
#build #spawn.fullfat.buckets=com.ibm.ws.transport.http2_fat

loriadi:
#build

hutchig (approved):
The change here is well documented and straightforward to understand. LGTM.

PR #23431, opened 2022-11-21
23429: Update http/2 test

### LINKED ISSUE #23429: Update http/2 test
testInvalidStreamIdSequence() sent a headers frame on stream 7.  The response should be a HEADERS frame and a DATA frame.  Then the test sends the bad HEADERS frame for stream 5.

However, the test only waits to get the response HEADERS frame for stream 7 before it sends the bad HEADERS frame on stream 5, so there is a tiny window where the response DATA frame may not have been received before the stream 5 headers is sent.

The test either needs to make the response DATA frame not a required response frame, or it needs to wait to receive both the HEADERS and DATA frames before it sends the bad HEADERS frame on stream 5.

Option 1 is easier, the error that this test is trying to produce will still happen, so going for that option.

loriadi:
fixes #23429 

Update the test to not expect the asynchronously sent DATA frame.


loriadi:
#build

hutchig (approved):

PR #23436, opened 2022-11-21
Add debug to see if retry addPartition helps ldap krb5 FATs

kristip17:
Ran into an issue again where the addPartition for the ApacheDS ldap on the krb5 FATs fails on z/OS runs. I'd previously added a retry, but it recently failed on the single retry. Adding an exception if we find that we can retry and succeed the second time before adding additional retries.

See RTC 293765


kristip17:
#build

jvanhill (approved):

PR #23443, opened 2022-11-22
Test injection of OpenTelemetry bean

Yas-Franco:
For #23670

Tests the CDI bean for OpenTelemetry by building a new Tracer and Spans for issue #23670
Adds DoPrivileged method to mpTelemetry WithSpan interceptor
Tests WithSpan annotation for issue #23450 
Exposes `io.opentelemetry.sdk.trace` to access SpanAttributes in tests


Yas-Franco:
#build

Yas-Franco:
#libby-build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/OpenTelemetryServlet.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    public void tracerTest() {
+        assertNotNull(tracer);
+    }
+
+    @Test
+    public void testOpenTelemetryNotNull() {
+        assertNotNull(openTelemetry);
+    }
```
Azquelt: `testOpenTelemetryNotNull` doesn't test anything since the test will throw an exception during `postConstruct` if `openTelemetry` is `null`.

Generally we want to try to cause an exception in the test method rather than during startup if we can.

I'm not entirely sure what `testSpanContextPropagationExplicitParent()` is really here for. It might be helpful to test that we can do _something_ with the `tracer`, but this test looks like it's trying to assert some other behaviour which wouldn't really belong in this test.

I would move all of this code into one test method so that the first time we call the `openTelemetry` bean is in the test method.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/OpenTelemetryServlet.java
```diff
@@ -0,0 +1,72 @@
...
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import jakarta.annotation.PostConstruct;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import jakarta.servlet.annotation.WebServlet;
+
+@WebServlet("/opentelemetry")
+@ApplicationScoped
+public class OpenTelemetryServlet extends FATServlet {
```
Azquelt: Can we rename this class to `OpenTelemetryBeanServlet` to better describe what it's testing.

Azquelt (changes requested):
See comments

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/OpenTelemetryServlet.java
Yas-Franco: Done!

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/OpenTelemetryServlet.java
Yas-Franco: I have updated this test now

Yas-Franco:
#build
#libby

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/OpenTelemetryBeanServlet.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private static final String SPAN_NAME = "MySpanName";
+
+    @Test
+    public void testOpenTelemetryBean() {
+        assertNotNull(openTelemetry);
+        Tracer tracer = openTelemetry.getTracer("instrumentation-test", "1.0.0");
+        assertNotNull(tracer);
+        Span span = tracer.spanBuilder(SPAN_NAME).startSpan();
+        assertThat(span.getSpanContext().getSpanId(), not(equalTo("0000000000000000")));
```
Azquelt: Can we make a constant for `"0000000000000000"` please? It would be easy to miss off a zero and break the test.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/WithSpanServlet.java
```diff
@@ -0,0 +1,115 @@
...
+    private SpanBean spanBean;
+
+    private static final String PARAMETER_TEST = "testString";
+
+    @Test
+    public void callNotAnnotated() {
+        //Returned spanId from methodNotAnnotated should not be the spanID of 0000000000000000
+        //No span was created
+        String spanId = spanBean.methodNotAnnotated();
+        assertThat(spanId, equalTo("0000000000000000"));
```
Azquelt: In these tests where we're trying to make sure that a span is created or not created by a method call, I think it would be better to capture the current span at the start of the method and compare the other spanId with that.

E.g.
```java
Span originalSpan = Span.current();

String newSpanId = spanBean.methodNotAnnotated();
assertThat(newSpanId, equalTo(originalSpan.getSpanContext().getSpanId());
```

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/WithSpanServlet.java
```diff
@@ -0,0 +1,115 @@
...
+        //The currentId remains the same as the previousId - It's not clear if this is expected behaviour or not
+        @WithSpan
+        void methodRecursiveAnnotated(int calls, String previousId) {
+            Span span = Span.current();
+            String currentId = span.getSpanContext().getSpanId();
+            System.out.println(currentId + " " + previousId);
+            if (calls > 0) {
+                methodRecursiveAnnotated(calls - 1, currentId);
+            }
+        }
```
Azquelt: The reason this isn't working is that CDI can only call interceptors when you call the method on an object that has been injected by CDI.

When you're calling from outside the bean, you're generally doing this anyway and don't really think about it.

However, if you're calling a method on the same bean and you want interceptors to fire, you need to inject the bean into itself and call the method on the injected object.

E.g.
```java
    @ApplicationScoped
    public static class SpanBean {
        
        @Inject
        private SpanBean thisBean;
        
        @WithSpan
        void methodRecursiveAnnotated(int calls, String previousId) {
            Span span = Span.current();
            String currentId = span.getSpanContext().getSpanId();
            System.out.println(currentId + " " + previousId);
            if (calls > 0) {
                thisBean.methodRecursiveAnnotated(calls - 1, currentId);
            }
        }

    }
```

Azquelt: Also, for this method you might want to pass a list into the method and have each call append its spanId to the list. Then you can assert that all the SpanIds in the list are different.

Azquelt (changes requested):
This looks really good, just a few comments.

Yas-Franco:
The recursive annotation method did not work as expected due to CDI issues. Methods can create child spans  by calling other annotated methods as shown in the `nestedAnnotated` test.

Yas-Franco:
#libby
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/OpenTelemetryBeanServlet.java
Azquelt: This bit still needs a constant. Ideally we should put it somewhere common.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/WithSpanServlet.java
```diff
@@ -0,0 +1,135 @@
...
+            ids.add(secondSpanBean.methodAnnotated());
+            return ids;
+        }
+
+        //SpanAttribute is applied to the method parameter
+        @WithSpan
+        public String methodAnnotatedWithParameter(@SpanAttribute("testParameter") String testParameter) {
+            Span span = Span.current();
+
+            System.out.println(span.getSpanContext().getTraceFlags());
```
Azquelt: Is this still supposed to be here? Printing out useful information is fine, but this doesn't seem relevant to the test.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/bnd.bnd
```diff
@@ -33,4 +33,5 @@ tested.features:\
 	io.openliberty.jakarta.cdi.4.0;version=latest,\
 	com.ibm.ws.componenttest.2.0;version=latest,\
 	io.openliberty.io.opentelemetry.internal;version=latest,\
-        io.openliberty.org.eclipse.microprofile.rest.client.3.0;version=latest
+  io.openliberty.jakarta.annotation.2.1;version=latest,\
+  io.openliberty.org.eclipse.microprofile.rest.client.3.0;version=latest
```
Azquelt: The indentation here doesn't match the rest of the file.

dev/com.ibm.websphere.appserver.features/visibility/public/mpTelemetry-1.0/io.openliberty.mpTelemetry-1.0.feature
```diff
@@ -12,6 +12,7 @@ IBM-API-Package: \
   io.opentelemetry.api.trace;type="third-party",\
   io.opentelemetry.api.common;type="third-party",\
   io.opentelemetry.context;type="third-party",\
+  io.opentelemetry.sdk.trace;type="third-party",\
```
Azquelt, line 15: Why does this package need to be added?

Azquelt (changes requested):
Added a few questions

dev/com.ibm.websphere.appserver.features/visibility/public/mpTelemetry-1.0/io.openliberty.mpTelemetry-1.0.feature
yasmin-aumeeruddy: This is not required. It's been taken out now

Yas-Franco: I got this error: 

java.lang.NoClassDefFoundError: io.opentelemetry.sdk.trace.ReadableSpan at 
io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.WithSpanServlet$SpanBean.methodAnnotatedWithParameter(WithSpanServlet.java:111) at 
io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.WithSpanServlet$SpanBean$Proxy$_$$_WeldSubclass.methodAnnotatedWithParameter$$super(Unknown Source) at

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/bnd.bnd
```diff
@@ -33,4 +33,8 @@ tested.features:\
 	io.openliberty.jakarta.cdi.4.0;version=latest,\
 	com.ibm.ws.componenttest.2.0;version=latest,\
 	io.openliberty.io.opentelemetry.internal;version=latest,\
+<<<<<<< HEAD
         io.openliberty.org.eclipse.microprofile.rest.client.3.0;version=latest
+=======
+	io.openliberty.jakarta.annotation.2.1;version=latest
+>>>>>>> 4aaac27a7f (Test injection of OpenTelemetry bean)
```
Azquelt: This is a merge that hasn't been done properly

dev/com.ibm.websphere.appserver.features/visibility/public/mpTelemetry-1.0/io.openliberty.mpTelemetry-1.0.feature
Azquelt: Sure, but you can't just expose packages in the API because it makes testing easier, you need to justify why the package needs to be accessible by applications and usually the list of accessible packages is laid down in the design document.

In this case I think we do need it anyway since the autoconfigure SPI references `io.opentelemetry.sdk.trace.SdkTracerProviderBuilder` [here][reference], but otherwise we'd have to find a different way to do this test.

[reference]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-sdk-extension-autoconfigure-spi/1.19.0/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.html#addTracerProviderCustomizer(java.util.function.BiFunction)

Yas-Franco:
#libby
#build

Azquelt (approved):

PR #23486, opened 2022-11-23
Add a doPrivileged call to WithSpanInterceptor

Yas-Franco:
Fixes a Java 2 Security error from the mpTelemetry `@WithSpan`  annotation


Yas-Franco:
#build

PR #23496, opened 2022-11-26
return types when using Query annotation

njr-11:
Experiment with some additional return types that are possible when using the Query annotation.
Also includes experimentation with Null, True, and False repository keywords.


njr-11:
#build

njr-11:
#build
#libby

KyleAure (approved):

PR #23498, opened 2022-11-28
Fix TCK Regex on Windows

tevans78:
Update the regular expresssions that match the maven GAV and paths
Add some unit tests for the regex and also the checksum utils

#build


Yas-Franco (approved):
Looks good!

tevans78:
Well that could have gone better but probably going to merge anyway...
- The PB did not complete fully but that is almost certainly an unrelated problem
- All TCKs ran and passed
- The original problem where jakarta TCK versions were not parsed properly has come back
So I will need to make a follow on fix but this PR should at least resolve the issue on Windows which is high impact.

PR #23506, opened 2022-11-28
Increase token lifetime to handle slow test systems

c00crane:
Increase token lifetimes to account for slow test machines.
The logout tests need tokens that do not live long since they'll be invoking logout with/without expired tokens and we don't want to have to wait too long for them to expire.  But, we've hit some slow machines where the tokens have expired before they're returned to the clients during a normal login process.


c00crane:
#build

ayoho (approved):

PR #23508, opened 2022-11-28
Updated bootstrap.properties to allow jakarta EE 10 features with checkpoint

KeenanMcCarthy:


dev/io.openliberty.checkpoint_fat/publish/servers/webProfileJSP/bootstrap.properties
```diff
@@ -10,5 +10,5 @@
 ###############################################################################
 bootstrap.include=../testports.properties
 com.ibm.ws.logging.trace.specification=*=info:checkpoint=all
-io.openliberty.checkpoint.allowed.features=timedexit-1.0,jsp-2.2,servlet-3.1
+io.openliberty.checkpoint.allowed.features=timedexit-1.0,servlet-3.1,servlet-6.0,pages-3.1,expressionLanguage-5.0
```
sebratton: The features servlet-6.0, pages-3.1, expressionLanguage-5.0 feature are already marked as enabled for instantOn so should not be needed in the list - unless it's not working in which case there is a bug somewhere in calculation of allowed features. Also I'm unclear why servlet-3.1 is needed in the list as it's part of jsp-2.3 which is enabled for InstantOn. 
Can you run this with only timedExit-1.0 in the allowed features?

sebratton (dismissed):
Looks good other than the suggested change (if feasible).

KeenanMcCarthy:
#build
#libby

KeenanMcCarthy:
#build
#libby

sebratton (dismissed):
Looks good.

KeenanMcCarthy:
#build

KeenanMcCarthy:
#build
#libby

sebratton (approved):

PR #23520, opened 2022-11-29
Fix Jakarta version parsing again

tevans78:
The problem this time was that the scope of the jakarta TCK jars was "test" rather than "compile".
Fixed and added another test

#build


Yas-Franco (approved):
Looks good!

PR #23524, opened 2022-11-29
reset the mark in the trace.log between tests

c00crane:
The security test framework automatically resets the "mark" in messages.log between tests - this allows the tests to only search for messages issued during a particular tests execution.  The refresh tests need to search for a message only recorded in trace.log.  The trace log mark is not reset between tests and we're finding the message from a previous test.
I can't reproduce locally, but this is happening quite frequently in our automated builds.
I will add a setMark request for the trace log to the between tests tooling.


c00crane:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat.refresh

teddyjtorres (approved):

PR #23529, opened 2022-11-29
Remove WASJSFConfig as it is no longer necessary

### LINKED ISSUE #23528: Remove WASJSFConfig from io.openliberty.faces.4.0.internal
The only use of this class is for a IBM specific configuration `com.ibm.ws.jsf.delayManagedBeanPostConstruct` which is no longer valid as of Faces 4.0 since ManagedBeans are no longer a part of Faces 4.0.

pnicolucci:
fixes #23528


pnicolucci:
#libby #build

volosied (approved):
Makes sense to remove it.

PR #23536, opened 2022-11-30
Create test for span and baggage propagation with jax clients

### LINKED ISSUE #23448: Telemetry REST integration FAT tests
MP Telemetry needs to integrate with `restfulWS` and `mpRestClient`. Whenever a request is made or received by one of these features, MP Telemetry should create a span. Trace ID information should be sent via the HTTP headers so that the traces from both the client and server can be joined up. Baggage information should also be sent via the HTTP headers so that contextual information from the first service can be included in trace messages in the second service.

For each of these locations:
* JAX-RS server
* JAX-RS client
* MP Rest Client

We need to test:
* Creation of spans
* Propagation of spanIds
* Propagation of baggage
* synchronous and async execution
* Correct application of [Semantic convention attributes][semconv]
  * TCK `RestClientSpanTest` only checks HTTP_STATUS, HTTP_METHOD, HTTP_SCHEME, HTTP_TARGET, HTTP_URL, see if there are any others which should be present
  * There are [constants defined][semconv-javadoc] for each of the attributes in the spec

There are several different standards (selected [via configuration][propagation-config]) for propagating spanIds and baggage. We need to test:
  * W3C tracer (default for trace)
  * W3C baggage (default for baggage)
  * B3
  * Jaeger

[semconv]: https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/http-metrics/
[semconv-javadoc]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-semconv/1.18.0-alpha/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.html
[propagation-config]: https://github.com/open-telemetry/opentelemetry-java/tree/v1.18.0/sdk-extensions/autoconfigure#propagator

benjamin-confino:
For #23448

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/JaxIntegration.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/JaxIntegration.java
@@ -1,0 +1,88 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal_fat;
+
+import static com.ibm.websphere.simplicity.ShrinkHelper.DeployOptions.SERVER_ONLY;
+import static org.junit.Assert.assertEquals;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import com.ibm.websphere.simplicity.ShrinkHelper;
+
+import componenttest.annotation.Server;
+import componenttest.annotation.TestServlet;
+import componenttest.annotation.TestServlets;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+import componenttest.topology.utils.HttpRequest;
+
+import io.openliberty.microprofile.telemetry.internal_fat.apps.jaxpropagation.JaxEndpoints;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.jaxpropagation.InMemorySpanExporterProvider;
+
+import io.opentelemetry.sdk.autoconfigure.spi.traces.ConfigurableSpanExporterProvider;
+
+@RunWith(FATRunner.class)
+public class JaxIntegration extends FATServletClient {
+
+    public static final String SERVER_NAME = "Telemetry10Jax";
+    public static final String APP_NAME = "JaxPropagation";
+
+    private static int[] dontRetryOnFail = {500}; //The server will return a 500 response code if an Assert fails during the test. Passing 500 in as an allowed response code will prevent wasting time  and more importantly prevent extra spans from confusing anyone trying to debug.
+
+    @Server(SERVER_NAME)
+    public static LibertyServer server;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        WebArchive app = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                        .addPackage(JaxEndpoints.class.getPackage())
+                        .addAsServiceProvider(ConfigurableSpanExporterProvider.class, InMemorySpanExporterProvider.class)
+                        .addAsResource(new StringAsset("otel.sdk.disabled=false\notel.traces.exporter=in-memory\notel.bsp.schedule.delay=100"),
+                                                       "META-INF/microprofile-config.properties");
+
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+        server.startServer();
+    }
+
+    @Test
+    public void testIntegrationWithJaxClient() throws Exception {
+        HttpRequest pokeJax = new HttpRequest(server, "/" + APP_NAME + "/endpoints/jaxclient");
+        assertEquals("Test Passed", pokeJax.run(java.lang.String.class));
+
+        Thread.sleep(1000);
+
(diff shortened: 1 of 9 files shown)


benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/jaxrspropagation/InMemorySpanExporter.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public void assertSpanCount(int spanCount) {
+        int retries = 4;
+        while (retries > 0 && finishedSpanItems.size() != spanCount) {
+            try {
+                retries --;
+                Thread.sleep(3000);
+            } catch (Exception e) {}
+        }
```
Azquelt: Can we reduce the sleep time here? (and increase the number of retries accordingly)

We'd like to continue as soon as we have the right number of items (which should be within 100ms) rather than waiting a whole three seconds.

Doesn't matter too much when there are only two tests, but if we use this in more tests it will greatly increase test runtime.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/jaxrspropagation/InMemorySpanExporterProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.traces.ConfigurableSpanExporterProvider;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import jakarta.enterprise.inject.spi.CDI;
+
+public class InMemorySpanExporterProvider implements ConfigurableSpanExporterProvider {
+    @Override
+    public SpanExporter createExporter(final ConfigProperties config) {
+System.out.println("GREP create");
```
Azquelt: Did you mean to leave this sysout here? It's not a bad thing to log but should be indented properly and probably be a Logger call.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/jaxrspropagation/InMemorySpanExporter.java
Azquelt: Also, don't catch and ignore `InterruptedException` here, the test should fail immediately if we're interrupted. I would usually rethrow it wrapped in a RuntimeException.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/JaxIntegration.java
```diff
@@ -0,0 +1,83 @@
...
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+        server.startServer();
+    }
+
+    @Test
+    public void testIntegrationWithJaxClient() throws Exception {
+        HttpUtils.findStringInUrl(server, "/" + APP_NAME + "/endpoints/jaxclient", dontRetryOnFail, HttpUtils.HTTPRequestMethod.GET, "Test Passed");
+        try {
+            Thread.sleep(1000);
+        } catch (Exception e) {}
```
Azquelt: We shouldn't catch `Exception` here. If sleep throws an `InterruptedException` we should fail the test.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/jaxrspropagation/JaxEndpoints.java
```diff
@@ -0,0 +1,155 @@
...
+        return Response.ok("Test Passed").build();
+    }
+
+    @GET
+    @Path("/jaxclient")
+    public Response getJax(@Context UriInfo uriInfo) {
+        assertNotNull(Span.current());
+        try {
+            Thread.sleep(3000);
+        } catch (Exception e) {}
```
Azquelt: Don't catch `Exception` around your sleeps in this class either.

Azquelt (changes requested):
I've made a few comments.

I'm also a little concerned that there are so many sleeps in these tests. I think you're sleeping for a minimum of 4 seconds in each test which is likely to be unsustainable as we add more tests.

I've asked for the one in `InMemorySpanExporter` to be reduced since that code is likely to be reused in other tests but I'm happy leave the other parts and revisit later if bucket runtime becomes a problem.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/jaxrspropagation/JaxEndpoints.java
```diff
@@ -0,0 +1,159 @@
...
+        Baggage baggage = Baggage.current();
+        assertEquals("bar", baggage.getEntryValue("foo"));
+
+        String baseUrl = uriInfo.getAbsolutePath().toString().replace("/mpclient", ""); //The mpclient will add the final part of the URL for you, so we remove the final part.
+        URI baseUri = null;
+        try {
+            baseUri = new URI(baseUrl);
+        } catch (Exception e) {
+            fail("URI exception from " + baseUrl);
+        }
```
Azquelt: Probably not a huge deal here, but calling `fail()` like this will throw away the original exception, which may have useful information in it.

Rethrowing as a `RuntimeException` and keeping the cause is probably better here, though if you want it to show up as a test _failure_ (rather than a test _exception_), throwing `AssertionError` will do that while also letting you keep the original exception as the cause.

Azquelt: I also I think we should never be throwing away the cause when catching something general like `Exception`, even in a test. It's not quite as bad if you're catching something specific like `URISyntaxException`.

If you managed to get something weird like a `ClassNotFoundException`, you'd have no way of knowing that the cause wasn't a `URISyntaxException`.

Azquelt (approved):

benjamin-confino:
#build

Azquelt:
#libby

PR #23564, opened 2022-12-01
Ensure artifactory auth data is updated

KyleAure:
Previous Behavior:
```
Check if the repository exists in config.json:
  if True: do not modify config.json
  if False: do any auths exist:
    if True: insert new registry + auth
    if False: create auths element
```

New Behavior:
```
Check if the repository exists in config.json:
  if True: does the repository have the correct auth token:
    if True: do not modify config.json
    if False: search and replace auth token
  if False: do any auths exist:
    if True: insert new registry + auth
    if False: create auths element
```


KyleAure:
#build

amyreit (approved):

PR #23581, opened 2022-12-03
remove some experimental code that will not be in Jakarta Data 1.0

njr-11:
Removal of some mock code that was trying out publisher as a repository return type that doesn't work correctly and leads to occasional test failures.


njr-11:
#build

tabishop (approved):

PR #23587, opened 2022-12-05
IgnoreCase for repository methods

njr-11:
Experiment with IgnoreCase on repository methods:
- as a keyword within name-pattern conditions
- as a keyword within name-pattern OrderBy
- as an attribute of the `@OrderBy` annotation
- as an attribute of dynamic `Sort`, both when sorts are use on their own as well as with both types of pagination.


njr-11:
#build

njr-11:
#build
#libby

jmstephensgit (approved):

PR #23605, opened 2022-12-06
23451 Test injection of OTel CDI beans

### LINKED ISSUE #23451: Telemetry bean value changing tests
MP Telemetry requires that several objects are exposed as CDI beans.

- [ ] The current `Span`
- [ ] The current `Tracer`
- [ ] The current `Baggage`

Which instance is "current" for each of these beans can be changed. Test that the CDI beans still work if the instance which is "current" is updated.

For each one:
* Get the current object
* Assert something about it using the CDI bean
  * mostly it's important to call a method on the CDI bean here to force CDI to initialize it
* Change the current object
* Assert something about the new current object using the CDI bean to see whether the bean reflects the change in the current object

benjamin-confino:
fixes #23451

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
@@ -1,0 +1,63 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal.cdi;
+
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.baggage.BaggageBuilder;
+import io.opentelemetry.api.baggage.BaggageEntry;
+
+/**
+ * This proxy class redirects method calls to Baggage.current(), by doing so it allows people to use @Inject Baggage and get an object which will not become obsolete.
+ */
+public class BaggageProxy implements Baggage {
+
+    @Override
+    public int size() {
+        return Baggage.current().size();
+    }
+
+    @Override
+    public void forEach(BiConsumer<? super String, ? super BaggageEntry> consumer) {
+        Baggage.current().forEach(consumer);
+    }
+
+    @Override
+    public Map<String, BaggageEntry> asMap() {
+        return Baggage.current().asMap();
+    }
+
+    @Override
+    public String getEntryValue(String entryKey) {
+        return Baggage.current().getEntryValue(entryKey);
+    }
+
+    @Override
+    public BaggageBuilder toBuilder() {
+        return Baggage.current().toBuilder();
+    }
+    
+    @Override
+    public boolean equals(Object obj) {
+        return Baggage.current().equals(obj);
+    }
+
+    public int hashCode() {
+        return Baggage.current().hashCode();
+    }
+
+    @Override
+    public String toString() {
+        return Baggage.current().toString();
+    }
+}
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
@@ -31,7 +31,6 @@ import io.opentelemetry.sdk.logs.SdkLoggerProvider;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.enterprise.context.RequestScoped;
 import jakarta.enterprise.inject.Disposes;
 import jakarta.enterprise.inject.Produces;
 import jakarta.inject.Inject;
@@ -127,14 +126,14 @@ public class OpenTelemetryProducer {
     }
 
     @Produces
-    @RequestScoped
+    @ApplicationScoped
     public Span getSpan() {
-        return Span.current();
+        return new SpanProxy();
     }
 
     @Produces
-    @RequestScoped
+    @ApplicationScoped
     public Baggage getBaggage() {
-        return Baggage.current();
+        return new BaggageProxy();
     }
 }
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
@@ -1,0 +1,90 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal.cdi;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.context.Context;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * This proxy class redirects method calls to Span.current(), by doing so it allows people to use @Inject Span and get an object which will not become obsolete.
+ */
+public class SpanProxy implements Span {
+
+    @Override
+    public <T> Span setAttribute(AttributeKey<T> key, T value) {
+        return Span.current().setAttribute(key, value);
+    }
+
+    @Override
+    public Span addEvent(String name, Attributes attributes) {
+        return Span.current().addEvent(name, attributes);
+    }
+
+    @Override
+    public Span addEvent(String name, Attributes attributes, long timestamp, TimeUnit unit) {
+        return Span.current().addEvent(name, attributes, timestamp, unit);
+    }
+
+    @Override
+    public Span setStatus(StatusCode statusCode, String description) {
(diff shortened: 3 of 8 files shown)


benjamin-confino:
#build

benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
```diff
@@ -0,0 +1,60 @@
...
+package io.openliberty.microprofile.telemetry.internal.cdi;
+
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.baggage.BaggageBuilder;
+import io.opentelemetry.api.baggage.BaggageEntry;
+
+public class BaggageProxy implements Baggage {
```
Azquelt, line 23: This class needs javadoc saying what it does and what it's for.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -127,14 +127,14 @@ public Tracer getTracer(OpenTelemetry openTelemetry) {
...
-        return Span.current();
+        return new SpanProxy();
     }
 
     @Produces
     @RequestScoped
     public Baggage getBaggage() {
-        return Baggage.current();
+        return new BaggageProxy();
     }
```
Azquelt: Should this be `ApplicationScoped` as well?

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
```diff
@@ -0,0 +1,87 @@
...
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.context.Context;
+import java.util.concurrent.TimeUnit;
+
+public class SpanProxy implements Span {
```
Azquelt, line 24: This class needs javadoc saying what it does and what it's for.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/CDIProxies.java
```diff
@@ -0,0 +1,60 @@
...
+import componenttest.annotation.Server;
+import componenttest.annotation.TestServlet;
+import componenttest.annotation.TestServlets;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.cdiproxies.CDIProxyServlet;
+
+@RunWith(FATRunner.class)
+public class CDIProxies extends FATServletClient {
```
Azquelt: Can we change the name of this test? CDI itself creates proxies and we're not testing them here. Maybe `TelemetryBeanTest`?

Names for the associated servlet and server should be updated as well.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/FATSuite.java
```diff
@@ -19,8 +19,9 @@
 @RunWith(Suite.class)
 @MinimumJavaLevel(javaLevel = 11)
 @SuiteClasses({
-                Telemetry10.class,
-                JaxIntegration.class
+                //Telemetry10.class,
+                //JaxIntegration.class,
+                CDIProxies.class
 })
```
Azquelt: Need to uncomment the original tests again.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/cdiproxies/CDIProxyServlet.java
```diff
@@ -0,0 +1,90 @@
...
+    @Inject
+    Baggage injectedBaggage;
+
+    @Test
+    public void testSpanUpdates() {
+        injectedSpan.toString(); //Just poke the object so CDI fetches the real object behind the proxy.
+
+        Span span = Span.current();
+        logger.info("Span.current at the start of hte method : " + span.toString());
+        assertEquals("The injected span was not the same as the current span before calling makeCurrent()", span.toString(), injectedSpan.toString()); //TODO - there's probably a better way to compare spans than toStrings (but just using (span, injectedSpan) will compare a weld proxy to a real Span object.
```
Azquelt: Testing `span.getSpanContext()` for equality here might be good. The [SpanContext][span-context] holds most of the data about the span and must have a well defined `equals()` method.

[span-context]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/1.18.0/io/opentelemetry/api/trace/SpanContext.html

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/cdiproxies/CDIProxyServlet.java
```diff
@@ -0,0 +1,90 @@
...
+        assertEquals("The injected span was not the same as the current span after calling makeCurrent()", spanTwo.toString(), injectedSpan.toString());
+    }
+
+    @Test
+    public void testBaggageUpdates() {
+        injectedBaggage.toString(); //Just poke the object so CDI fetches the real object behind the proxy.
+
+        Baggage baggage = Baggage.current();
+        logger.info("Baggage.current at the start of hte method : " + baggage.toString());
+        assertEquals("The injected baggage was not the same as the current baggage before calling makeCurrent()", baggage.toString(), injectedBaggage.toString()); //TODO - there's probably a better way to compare
```
Azquelt: `baggage.asMap()` is probably the best way to test this one since `Baggage` doesn't say whether it declares an `equals()` method.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/cdiproxies/CDIProxyServlet.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Test
+    public void testSpanUpdates() {
+        injectedSpan.toString(); //Just poke the object so CDI fetches the real object behind the proxy.
+
+        Span span = Span.current();
+        logger.info("Span.current at the start of hte method : " + span.toString());
+        assertEquals("The injected span was not the same as the current span before calling makeCurrent()", span.toString(), injectedSpan.toString()); //TODO - there's probably a better way to compare spans than toStrings (but just using (span, injectedSpan) will compare a weld proxy to a real Span object.
+
+        Span newSpan = tracer.spanBuilder("my span").setParent(Context.current().with(span)).startSpan();
```
Azquelt: There's no need to set the parent here, it will automatically use the current span as the parent.

If you were to do it manually, I think you'd just want `Context.current()` here, but in this case you should just take it out.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/cdiproxies/CDIProxyServlet.java
```diff
@@ -0,0 +1,90 @@
...
+        logger.info("Span created during the test: " + newSpan.toString());
+        newSpan.makeCurrent();
+
+        Span spanTwo = Span.current();
+        logger.info("Span.current() after creating a new span and making it current : " + spanTwo.toString());        
+
+        assertEquals("The current span was not the one we created and invoked makeCurrent() with", newSpan.toString(), spanTwo.toString());
+        assertEquals("The injected span was not the same as the current span after calling makeCurrent()", spanTwo.toString(), injectedSpan.toString());
```
Azquelt: We should obey the contract for `Span` here and make sure we end it.

We should use
```java
try (Scope s = newSpan.makeCurrent()) {
    // rest of test
} finally {
    newSpan.end();
}
```

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/cdiproxies/CDIProxyServlet.java
```diff
@@ -0,0 +1,90 @@
...
+        Span span = Span.current();
+        logger.info("Span.current at the start of hte method : " + span.toString());
+        assertEquals("The injected span was not the same as the current span before calling makeCurrent()", span.toString(), injectedSpan.toString()); //TODO - there's probably a better way to compare spans than toStrings (but just using (span, injectedSpan) will compare a weld proxy to a real Span object.
+
+        Span newSpan = tracer.spanBuilder("my span").setParent(Context.current().with(span)).startSpan();
+        logger.info("Span created during the test: " + newSpan.toString());
+        newSpan.makeCurrent();
+
+        Span spanTwo = Span.current();
+        logger.info("Span.current() after creating a new span and making it current : " + spanTwo.toString());
```
Azquelt: Whatever test we use here, could we also check that `spanTwo` is not equal to `span`? It would suck if we'd misconfigured something and the test was passing because otel was disabled and _all_ spans were equal to each other.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/CDIProxies.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static final String SERVER_NAME = "Telemetry10Proxies";
+    public static final String APP_NAME = "ProxiesApp";
+
+    @Server(SERVER_NAME)
+    @TestServlets({
+
+                    @TestServlet(servlet = CDIProxyServlet.class, contextRoot = APP_NAME)
+    })
```
Azquelt: There's no need to use `@TestServlets` if there's only one test servlet (though it doesn't do any harm either, so you can leave it in if you like).

Azquelt (changes requested):
Looks mostly good, I've left some comments.

Can you also update the title and description of the pull request to better reflect the work you've done, particularly as it goes beyond the scope of the issue it fixes.

Azquelt (approved):

benjamin-confino:
#build

Azquelt:
Failures in the build match known defects.
#libby

PR #23611, opened 2022-12-06
Enable and test compatibility between mpTelemetry-1.0 and the OTel Agent

### LINKED ISSUE #23206: Add FAT tests for using `mpTelemetry-1.0` alongside the instrumentation agent
We want mpTelemetry to work with the OTel auto-instrumentation agent.

Add tests to see whether it does and to find out any features which don't work in an intuitive way.

Test that:
 - [x] does everything normal still work?
   - injection of beans
   - rest instrumentation
   - manual instrumentation
 - [x] do applications included the user's application get instrumented?
   - rxJava
   - Java 11 HttpClient
   - non-bean @WithSpan
 - [x] [configuration of the agent][agent-config]
   - Options from autoconfigure are respected
   - config can be provided via file
   - config can be provided via environment variables
   - config can be provided via jvmOptions
   - config can/cannot be provided via bootstrap.properties
   - disable specific instrumentations via config
   - can disable instrumentation generally but enable specific instrumentations
 - [x] multiple-app behaviour
   - can we trace on both apps?
   - confirm that app config is ignored?
   

[agent-config]: https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/

Azquelt:
Ensure that mpTelemetry works with the OpenTelemetry auto-instrumentation agent.

We don't ship or support the agent, but we want to make sure that the mpTelemetry feature can work with it, even if there are some limitations.

Make the following product code changes:
 - If the agent is detected, use `GlobalOpenTelemetry.get()` to retrieve the `OpenTelemetry` object it created, rather than creating our own
 - If the agent is detected, disable our interceptor for `WithSpan` and our filters for JAX-RS and MP Rest Client
 
Add the following tests:
 - Test that instrumentation of JAX-RS, MP Rest Client and `@WithSpan` annotated bean methods still works
 - Test that we can still manually create spans using the API (including nested spans)
 - Test that the RxJava library is instrumented by the agent when it's packaged inside the application
 - Test that the Java 11 HTTP Client is instrumented by the agent
 - Test that agent config is read from jvm.options, environment variables and the agent config file, but not from bootstrap.properties or microprofile-config.properties
 - Test that the agent can be disabled via config and that our feature then takes over
 - Test that individual instrumentations can be disabled and that in this case our feature does not take over
 - Test that all instrumentations can be disabled and then specific ones re-enabled
 - Test that multiple apps can be deployed and all are instrumented
   - Note: All apps run with the same Open Telemetry config and there's no way to differentiate between traces from different apps, other than the URL used to make the initial request

Fixes #23206


Azquelt:
#build

Azquelt:
#build
#libby

Yas-Franco:
> The build Azquelt-23611-20221207-0324 https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_NXsCIHYYEe2ux-fzqy0jNQ completed and has errors or failures.
> 
> For help analyzing your personal build, go to https://cognitive.hursley.ibm.com/buildAnalysis.html?uuid=_NXsCIHYYEe2ux-fzqy0jNQ

It looks like the classes needed are from https://mvnrepository.com/artifact/io.grpc/grpc-stub/1.43.2 and https://mvnrepository.com/artifact/com.google.guava/failureaccess/1.0.1 which we don't import. I'm not sure how you got the tests working locally though

Azquelt:
@yasmin-aumeeruddy This works locally but fails in the build because dependencies for FAT tests are pulled in using the `build.gradle` file. Locally it pulls dependencies from Maven Central which has the correct metadata but in the build it pulls from the internal artifactory instance which doesn't.

We do include `io.grpc:grpc-stub:1.43.2` in the build.gradle and I'd assumed that `com.ibm.ws.com.google.guava` would include all of the necessary guava dependencies but it looks like this is not the case.

Azquelt:
#libby
#build

Azquelt:
#build

Yas-Franco (approved):
Looks good!

PR #23623, opened 2022-12-07
Removing bootstrap binary

cbridgha:
This is to remove the binary that was mistakenly included in a previous commit - This binary is generated during the build phase - so does not belong in the source tree.


cbridgha:
#build

jimblye (approved):

hlhoots (approved):
LGTM

PR #23571, opened 2022-12-02
add async tests

benjamin-confino:
This expands the tests to cover async invocations.

PR #23630, opened 2022-12-07
Telemetry: propagate context on async requests

### LINKED ISSUE #23572: Open Telemetry spans and baggage not correctly propagated when using asynchronous clients
When a REST Client is used in a method traced by OpenTelementry trace ID information should be sent via the HTTP headers so that the traces from both the client and server can be joined up. Baggage information should also be sent via the HTTP headers so that contextual information from the first service can be included in trace messages in the second service.

Currently this works fine in normal circumstances but fails if the REST client is set up to run asynchronously. This is true for both a JAX-RS client and a microprofile rest client. 

To reproduce run io.openliberty.microprofile.telemetry.1.0.internal_fat with experimental tests enabled once #23571 is merged.

Azquelt:
Ensure the telemetry context is propagated when an async request is made using JAX-RS Client or MP Rest Client.

This builds on the tests from #23571

Fixes #23572 
For  #23448


Azquelt:
#build

benjamin-confino (dismissed):

Azquelt:
This needs to be rebased due to a conflict with #23637 which also fixed the problem of package names not matching the folder names (but made the change the other way).

Azquelt:
#libby
#build

benjamin-confino (approved):

PR #23636, opened 2022-12-07
Handle server stop failures in EJB FAT

tkburroughs:
- add try/finally to ensure subsequent tests don't fail if server stop fails


tkburroughs:
#build

olendvcook (approved):

Issue #23175, opened 2022-10-28
MP Telemetry defaults for logging and metrics


Azquelt:
By default, the autoconfiguration enables logging and metrics. We should disable this if we're not supporting it.

PR #23646, opened 2022-12-09
Disable metrics and logs for MPTelemetry

Yas-Franco:
For issue #23175 

Ignores config properties that the user has set for the logs and metrics exporters from OpenTelemetry. The  properties are then set to "none"

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
@@ -42,6 +42,10 @@ public class OpenTelemetryProducer {
     private static final String INSTRUMENTATION_NAME = "io.openliberty.microprofile.telemetry";
     private static final String ENV_DISABLE_PROPERTY = "OTEL_SDK_DISABLED";
     private static final String CONFIG_DISABLE_PROPERTY = "otel.sdk.disabled";
+    private static final String ENV_METRICS_EXPORTER_PROPERTY = "OTEL_METRICS_EXPORTER";
+    private static final String CONFIG_METRICS_EXPORTER_PROPERTY = "otel.metrics.exporter";
+    private static final String ENV_LOGS_EXPORTER_PROPERTY = "OTEL_LOGS_EXPORTER";
+    private static final String CONFIG_LOGS_EXPORTER_PROPERTY = "otel.logs.exporter";
 
     @Inject
     Config config;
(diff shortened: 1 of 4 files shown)


Yas-Franco:
#libby
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -113,8 +117,16 @@ private HashMap<String, String> getTelemetryProperties() {
...
+                //Do not set metrics or logs exporter
+                if ((!propertyName.equals(ENV_METRICS_EXPORTER_PROPERTY)) && (!propertyName.equals(CONFIG_METRICS_EXPORTER_PROPERTY))
+                    && (!propertyName.equals(ENV_LOGS_EXPORTER_PROPERTY)) && (!propertyName.equals(CONFIG_LOGS_EXPORTER_PROPERTY))) {
+                    config.getOptionalValue(propertyName, String.class).ifPresent(
+                                                                                  value -> telemetryProperties.put(propertyName, value));
+
+                }
+                //Metrics and logs are disabled by default
+                telemetryProperties.put(CONFIG_METRICS_EXPORTER_PROPERTY, "none");
+                telemetryProperties.put(CONFIG_LOGS_EXPORTER_PROPERTY, "none");
```
Azquelt: This should be outside the `for` loop.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -113,8 +117,16 @@ private HashMap<String, String> getTelemetryProperties() {
         HashMap<String, String> telemetryProperties = new HashMap<>();
         for (String propertyName : config.getPropertyNames()) {
             if (propertyName.startsWith("otel.") || propertyName.startsWith("OTEL_")) {
-                config.getOptionalValue(propertyName, String.class).ifPresent(
-                                                                              value -> telemetryProperties.put(propertyName, value));
+                //Do not set metrics or logs exporter
+                if ((!propertyName.equals(ENV_METRICS_EXPORTER_PROPERTY)) && (!propertyName.equals(CONFIG_METRICS_EXPORTER_PROPERTY))
+                    && (!propertyName.equals(ENV_LOGS_EXPORTER_PROPERTY)) && (!propertyName.equals(CONFIG_LOGS_EXPORTER_PROPERTY))) {
```
Azquelt: We shouldn't need to do this check if we just set these properties after the `for` loop.

Azquelt:
We should really have a test for this, but I'm not sure what we can check for.

Azquelt:
We can maybe call `OpenTelemetry.getMeterProvider` and check that its class name is `io.opentelemetry.api.metrics.DefaultMeterProvider`.

Unsure if this will work when the `MeterProvider` interface isn't available to applications:

```java
Object meterProvider = openTelemetry.getMeterProvider();
assertEquals("io.opentelemetry.api.metrics.DefaultMeterProvider", meterProvider.getClass().getName());
```

If that causes a class not found exception, this might work instead:
```java
Method getMeterProvider = OpenTelemetry.class.getMethod("getMeterProvider");
Object meterProvider = getMeterProvider.invoke(openTelemetry);
assertEquals("io.opentelemetry.api.metrics.DefaultMeterProvider", meterProvider.getClass().getName());
```

Using either of these you _may_ need to grant the test application addition permissions since we're using reflection.

I don't think we can use the same trick to get the logging provider since it's not available on the `OpenTelemetry` interface and I don't think we can cast it to `OpenTelemetrySDK` because you can't generally cast an injected CDI bean to one of its subclasses.

yasmin-aumeeruddy:
#libby
#build

Yas-Franco:
#build

Yas-Franco:
#libby

Azquelt (approved):

PR #23697, opened 2022-12-14
Instanton server fixes

tjwatson:
Remove need for using the 'ps' command.

Add glibc.pthread.rseq=0 to the glibc tunables to make process images more portable.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -862,9 +862,9 @@ readServerEnv()
 criuEnvDefaults() {
   if [ -z "$GLIBC_TUNABLES" ]
   then
-    GLIBC_TUNABLES=glibc.cpu.hwcaps=-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load
+    GLIBC_TUNABLES=glibc.cpu.hwcaps=-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load:glibc.pthread.rseq=0
   else
-    GLIBC_TUNABLES=${GLIBC_TUNABLES}:glibc.cpu.hwcaps=-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load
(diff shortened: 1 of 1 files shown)


tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1387,8 +1385,9 @@ criuRestore()
     mkdirs "${X_PID_DIR}"
     rmIfExist "${X_PID_FILE}"
     criu restore --skip-file-rwx-check --cpu-cap=none --file-locks --tcp-established --images-dir=${SERVER_OUTPUT_DIR}/workarea/checkpoint/image \
-      --shell-job --verbosity=${CRIU_LOG_LEVEL} --log-file ${X_LOG_DIR}/checkpoint/restore.log --pidfile ${X_PID_FILE} \
-      --restore-detached ${CRIU_EXTRA_ARGS}
+      --shell-job --verbosity=${CRIU_LOG_LEVEL} --log-file ${X_LOG_DIR}/checkpoint/restore.log ${CRIU_EXTRA_ARGS} \
```
sebratton, line 1384: any reason why you moved ${CRIU_EXTRA_ARGS} from the end of the command to here? The doc states:
```
# CRIU_EXTRA_ARGS   - Pass extra arguments to `criu restore`. Extra arguments are
#                     appended to the end of the list of arguments and can therefore
#                     override existing arguments, if desired.
```

tjwatson: My reason was to get first two lines of the `criu restore` call to be identical to the `criu restore` in the forground:

https://github.com/OpenLiberty/open-liberty/blob/b1f9bb78736d05f95915fa721fd441ba8dc3d138/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server#L1417-L1418

And then only append the two options `--pidfile ${X_PID_FILE}` and `--restore-detached` to the end.  Then, in theory, making it easier to keep the two in-sync when we have to add options.

sebratton: Sounds reasonable. Maybe we should remove the header comment though where we document that CRIU_EXTRA_ARGS can be used to override values.

sebratton (approved):

tjwatson:
#libby

PR #23700, opened 2022-12-15
Adjust java11 FAT test to work with JDK20+

jgrassel:
Fix a java 11 test case which had a check limit of up to Java 19.


jgrassel:
#build

gjwatts (approved):

PR #23715, opened 2022-12-15
Iterator with keyset pagination

njr-11:
Experiment with keyset pagination on an Iterator, trying it out for next/previous direction and with sort criteria that comes from Pageable, the OrderBy keyword, and the OrderBy annotation.
Also in this pull:
Tests with KeysetAwareSlice
Remove examples for consumer/collector which isn't in the spec


njr-11:
#build

jmstephensgit (approved):

PR #23720, opened 2022-12-16
Ensure first app is stopped, before restarting

KyleAure:
In this test, we ensure that a ManagedThreadFactory is persisted when an application is restarted.
To do that we remove the app from the apps location, and then put it back.
We currently wait for the config update message: 
`CWWKG0017I: The server configuration was successfully updated`

Before attempting to reinstall the application. 
However, the application could still be running. 
Resulting in the following error: 
`CWWKZ0013E: It is not possible to start two applications called concurrent.`

We should additionally, wait until we see the message: 
`CWWKZ0009I: The application concurrent has stopped successfully.`

### DIFF
--- a/dev/com.ibm.ws.concurrent_fat_config/fat/src/com/ibm/ws/concurrent/fat/EEConcurrencyConfigTest.java
+++ b/dev/com.ibm.ws.concurrent_fat_config/fat/src/com/ibm/ws/concurrent/fat/EEConcurrencyConfigTest.java
@@ -440,7 +440,7 @@ public class EEConcurrencyConfigTest extends FATServletClient {
         // save
         server.setMarkToEndOfLog();
         server.updateServerConfiguration(config);
-        server.waitForConfigUpdateInLogUsingMark(Collections.emptySet());
+        server.waitForConfigUpdateInLogUsingMark(Collections.emptySet(), "CWWKZ0009I"); //CWWKZ0009I: The application concurrent has stopped successfully.
 
         // Start the application by putting it back
         config.getApplications().add(concurrentApp);


dev/com.ibm.ws.concurrent_fat_config/fat/src/com/ibm/ws/concurrent/fat/EEConcurrencyConfigTest.java
```diff
@@ -441,6 +441,7 @@ public void testManagedThreadFactoryUsableAcrossAppRestart() throws Exception {
         server.setMarkToEndOfLog();
         server.updateServerConfiguration(config);
         server.waitForConfigUpdateInLogUsingMark(Collections.emptySet());
+        server.waitForStringInTraceUsingMark("CWWKZ0009I"); //CWWKZ0009I: The application concurrent has stopped successfully.
```
njr-11: Nice work debugging the cause and identifying where to correct it.
Instead of scanning trace files for the message, you should be able to have `waitForConfigUpdateInLogUsingMark(Collections.emptySet(), "CWWKZ0009I")` wait for it in the messages file, which is much shorter than trace,

KyleAure: Thanks for the suggestion.  I forgot you could add messages to the end of the config update.

KyleAure:
#build

njr-11 (approved):

KyleAure:
#libby

PR #23727, opened 2022-12-16
stop command will wait for sCommand file when server starting

### LINKED ISSUE #23392: Stopping liberty Windows service immediately after starting results in hang condition
**Describe the bug**  
This applies when an open liberty server is registered as a Windows service.   So obviously it is Windows-only.  If you stop the service immediately after starting the service, this will result in a hang condition. 

**Steps to Reproduce**  
 - Method 1
Register a liberty server as a Windows service.
Start the service using the Windows services panel.
As soon as the service appears to be running, stop the service.
Windows services should hang for a few minutes and eventually display a popup window.
The state of the service will be stuck in stopping.
To stop the service at this point, you will need to kill the server process.

- Method 2
Register a liberty server as a Windows service
Enter the  following commands one right after the other.   
server startWinService
server stopWinService
If you entered the 2nd command quick enough, the second command will hang.

**Expected behavior**  
You should be able to start the Windows service and immediately stop it.

**Diagnostic information:**  
 - OpenLiberty Version: [ * - 22.0.0.10] - beginning version is the 1st version that allows running liberty as a Windows service 
 - Affected feature(s)  - no specific feature.
 - Java Version: [ all versions ]
 - server.xml configuration ( doesn't matter)

**Additional context**  
The apache program, prunsrv, is used to register, start, stop, and unregister a liberty server as a Windows service.  When you use the Windows services app to start the service, it invokes prunsrv, which in turn invokes server.bat.   The problem seems to be that prunsrv is not waiting for server.bat to finish.  So it is reporting back to Windows services that the service is up, but the server is not quite finished yet.   When the server is finished starting up, it creates a _serverDir_\workarea\.sCommand file which contains the TCP/IP port that is needed to send commands to the server.   If you try to stop the server too soon, the command will ultimately fail, but will hang for about a minute or two.  It fails because it cannot find the .sCommand file.

An issue has been opened with prunsrv since they are not waiting for the command to finish:
    https://issues.apache.org/jira/browse/DAEMON-449

This is also fixable in liberty code.  The stop command first checks to see if the server is running.   If it is not, it immediately stops.   If it is running, it then checks for the existence of .sCommand.   If it doesn't exist, there is either a problem, or the server is still starting (more likely.).   The problem can be solved by simply polling on the existence of .sCommand.  The timeout should probably be whatever the stop timeout is, which defaults to 30 seconds.

jimblye:
Fixes #23392.

If the stop command detects that the server is running, but yet there is no .sCommand file then assume that the server must still be starting and simply hasn't created the .sCommand file yet.

If we don't do this, then the stop command will fail, since it needs the .sCommand file to know how to communicate with the server.  In the case of running the server as a Windows service, this causes a hang in prunsrv.  I believe this is a bug in prunsrv.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/ProcessControlHelper.java
+++ b/dev/com.ibm.ws.kernel.boot.core/src/com/ibm/ws/kernel/boot/internal/commands/ProcessControlHelper.java
@@ -99,7 +99,11 @@ public class ProcessControlHelper {
                     stopRc = scc.stopServer(launchArgs.getOption("force") != null);
                 } else {
                     // we can't communicate to the server...
-                    stopRc = ReturnCode.ERROR_SERVER_STOP;
+                    stopRc = waitForCommandFile(scc, new Integer(BootstrapConstants.SERVER_STOP_WAIT_TIME_DEFAULT));
+
+                    if (stopRc == ReturnCode.OK) {
+                        stopRc = scc.stopServer(launchArgs.getOption("force") != null);
+                    }
                 }
             } else {
                 // nope: lock not held, we're already stopped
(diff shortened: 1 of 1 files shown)


jimblye:
#build

hlhoots (approved):
LGTM

jimblye:
#build
#libby

PR #23756, opened 2022-12-20
infer result for select query, and parameter combinations

njr-11:
Removes the Result annotation (not part of the spec) because the information can be inferred.
Covers additional parameter combinations for repository methods, such as Limit and Sort for pagination.


njr-11:
#build

jmstephensgit (approved):

PR #23761, opened 2022-12-21
CDI: Consider shared libs as "application" BDAs

### LINKED ISSUE #23748: CDI Shared Library bean visibility problems
**Describe the bug**  
There are a few places where we're mishandling common library Bean Deployment Archives (BDAs) which result in the following problems:

- Common lib BDAs can't see Runtime Extension BDAs (and vice-versa, if enabled for the runtime extension)
- Finding the BDA of a class in a common lib fails

**Steps to Reproduce**  
We expect that this breaks:

- Trying to inject a bean provided by a runtime extension into a bean from a shared lib
- Using `CDI.current()` from a class in a shared lib (doesn't fail until it tries to actually get the current `BeanManager`, as that requires looking up the BDA)
   - This may actually only be a problem if there's a class from a Runtime Extension on the stack and that Runtime Extension can't see the bean you're looking for as we do a stack walk and will find the Runtime Extension BDA
   - It may not fail if there's no other class in a BDA on the stack as we then fall back to the BDA of the current module which will usually work

**Expected behavior**  
The above scenarios should work.

**Diagnostic information:**  
 - Affected feature(s) `cdi-1.2`, `cdi-2.0`, `cdi-3.0`, `cdi-4.0` (beta)
 - Java Version: n/a (bug is present with any Java version)

**Additional context**  
When an application is deployed, CDI creates a Bean Deployment Archive (BDA) object for each archive within the application. The BDA object keeps track of what classes are available within that archive and which other BDAs it has visibility of. CDI Beans within a BDA must have visibility of another BDA to be able to inject or look up beans from that BDA.

For the purposes of CDI, common and private libraries should be considered part of the application.

"CDI Runtime Extensions" are CDI extensions or beans provided by another liberty feature (e.g. `mpFaultTolerance` adds a CDI runtime extension to add a CDI interceptor and process annotations using the CDI extension API). CDI creates an additional BDA for each Runtime Extension.

In general:
- All application BDAs should have visibility of other application BDAs if they can load classes from the archive of the other BDA
- All application BDAs should have have visibility of all Runtime Extension BDAs
- Some Runtime Extension BDAs should have visibility of all application BDAs (depending on the configuration of the Runtime Extension)

CDI was not considering common libraries as part of the application, causing two problems:
- the visibility rules were set up incorrectly so that common library BDAs could not see Runtime Extension BDAs (and vice-versa)
- when CDI would internally query for the BDA of a class in a common library, it would get the wrong result
  - often it would get a different application BDA and using that BDA would lead to the same result

**Fix plan:**
We need to add tests for these cases and update the code so that they work.

We think we need to:
 - change `WebSphereCDIDeploymentImpl.initializeOrderedBeanDeploymentArchives()` to include shared libs in `applicationBDAs`
 - update `WebSphereCDIDeploymentImpl.initializeInjectionServices()` to exclude shared libs

Azquelt:
This fixes two problems:
 - application BDAs are given visibility of runtime extensions. Shared libs should have this.
 - application BDAs are included in the the list of BDAs we search when looking up the BDA for a class. Shared libs should be included this list.

Add tests for:
- visibility of runtime extension BDAs
- calling CDI.current() from class in a shared library
- enablement of CDI in an application where the only visible CDI bean is in a shared library

Fixes #23748

### DIFF
--- a/dev/com.ibm.ws.cdi.api_fat/fat/src/com/ibm/ws/cdi/api/fat/apps/current/CDICurrentTestServlet.java
+++ b/dev/com.ibm.ws.cdi.api_fat/fat/src/com/ibm/ws/cdi/api/fat/apps/current/CDICurrentTestServlet.java
@@ -10,22 +10,25 @@
  *******************************************************************************/
 package com.ibm.ws.cdi.api.fat.apps.current;
 
+import static java.time.temporal.ChronoUnit.SECONDS;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
-import static java.time.temporal.ChronoUnit.SECONDS;
+import java.time.Duration;
+import java.util.logging.Logger;
 
 import javax.annotation.Resource;
 import javax.enterprise.concurrent.ManagedExecutorService;
 import javax.enterprise.inject.spi.CDI;
+import javax.inject.Inject;
 import javax.servlet.annotation.WebServlet;
 
-import java.time.Duration;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import org.junit.Test;
 
 import com.ibm.ws.cdi.api.fat.apps.current.extension.MyDeploymentVerifier;
+import com.ibm.ws.cdi.api.fat.apps.current.sharedLib.ISimpleBean;
+import com.ibm.ws.cdi.api.fat.apps.current.sharedLib.SharedLibBean;
 
 import componenttest.app.FATServlet;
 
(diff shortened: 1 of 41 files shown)


Azquelt:
#build
#spawn.fullfat.buckets=com.ibm.ws.cdi.2.0_fat,com.ibm.ws.cdi.annotations_fat,com.ibm.ws.cdi.api_fat,com.ibm.ws.cdi.beansxml_fat,com.ibm.ws.cdi.beansxml.implicit_fat,com.ibm.ws.cdi.ejb_fat,com.ibm.ws.cdi.extension_fat,com.ibm.ws.cdi.jee_fat,com.ibm.ws.cdi.jndi_fat,com.ibm.ws.cdi.lifecycle_fat,com.ibm.ws.cdi.noncontextual_fat,com.ibm.ws.cdi.third.party_fat,com.ibm.ws.cdi.visibility_fat,com.ibm.ws.clientcontainer.cdi_fat,com.ibm.ws.ejbcontainer.cdi_fat,com.ibm.ws.jaxrs.2.0.cdi.1.2_fat,com.ibm.ws.jaxrs.2.1.cdi.2.0_fat,com.ibm.ws.jaxws.cdi_fat,com.ibm.ws.jbatch.cdi_fat,com.ibm.ws.jpa.tests.spec21.cdi_fat,com.ibm.ws.microprofile.faulttolerance.cdi_fat,com.ibm.ws.microprofile.faulttolerance.cdi_fat_metrics,io.openliberty.cdi.3.0.internal_fat,io.openliberty.cdi.4.0.internal_fat,io.openliberty.cdi.core_fat,io.openliberty.restfulWS.3.0.cdi.3.0_fat

Requesting a personal build that runs all the CDI FATs in full mode since changing BDA visibility affects everything but if this breaks something it's likely to be quite obscure.

benjamin-confino (approved):

PR #23764, opened 2022-12-22
Security 3.0: Add error message for TokenResponseValidator

kristip17:
Addresses to-do to add an error message to TokenResponseValidator.

Created new unit test to test different failures of getJwtClaimsFromIdTokenContext()

Example message with variables: `CWWKS2425E: The oidcclientid OpenID Connect client token is invalid. The jwtContext field is null.`

In context when it's thrown to the user: `io.openliberty.security.oidcclientcore.token.TokenValidationException: CWWKS2415E: The oidcclientid OpenID Connect client encountered the following error during validation of the token that was received from the OpenID Connect provider: CWWKS2425E: The oidcclientid OpenID Connect client token is invalid. The jwtClaims value on the JwtContext object is null or empty.`

### DIFF
--- a/dev/io.openliberty.security.oidcclientcore.internal/resources/io/openliberty/security/oidcclientcore/internal/resources/OidcClientCoreMessages.nlsprops
+++ b/dev/io.openliberty.security.oidcclientcore.internal/resources/io/openliberty/security/oidcclientcore/internal/resources/OidcClientCoreMessages.nlsprops
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2022 IBM Corporation and others.
+# Copyright (c) 2022, 2023 IBM Corporation and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License 2.0
 # which accompanies this distribution, and is available at
@@ -121,3 +121,7 @@ OIDC_CLIENT_INVALID_RESPONSE_TYPE.useraction=Specify one of the valid response t
 TOKEN_CLAIM_VALUE_MISMATCH=CWWKS2424E: The [{0}] value for the [{1}] claim in the token does not match the [{2}] expected value.
 TOKEN_CLAIM_VALUE_MISMATCH.explanation=The claim value is incorrect or malformed. The configuration for the OpenID Connect client might be incorrect, or the token was created with incorrect values.
 TOKEN_CLAIM_VALUE_MISMATCH.useraction=Ensure the attribute in the OpenID Connect client configuration that is related to the claim that is specified in the message is configured correctly.
+
+TOKEN_MISSING_CLAIMS=CWWKS2425E: The {0} OpenID Connect client token is invalid. The jwtClaims value on the JwtContext object is null or empty.
+TOKEN_MISSING_CLAIMS.explanation=Information is missing from the token and validation of the token cannot be completed.
+TOKEN_MISSING_CLAIMS.useraction=Review the log for any earlier errors for details on the failure. Check the format and content of the token that is returned from the OpenID Connect provider.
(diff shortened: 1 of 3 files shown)


kristip17:
#build

dev/io.openliberty.security.oidcclientcore.internal/resources/io/openliberty/security/oidcclientcore/internal/resources/OidcClientCoreMessages.nlsprops
```diff
@@ -119,3 +119,7 @@ OIDC_CLIENT_INVALID_RESPONSE_TYPE.useraction=Specify one of the valid response t
 TOKEN_CLAIM_VALUE_MISMATCH=CWWKS2424E: The [{0}] value for the [{1}] claim in the token does not match the [{2}] expected value.
 TOKEN_CLAIM_VALUE_MISMATCH.explanation=The claim value is incorrect or malformed. The configuration for the OpenID Connect client might be incorrect, or the token was created with incorrect values.
 TOKEN_CLAIM_VALUE_MISMATCH.useraction=Ensure the attribute in the OpenID Connect client configuration that is related to the claim that is specified in the message is configured correctly.
+
+INVALID_TOKEN_INFO=CWWKS2425E: The {0} OpenID Connect client token is invalid. The {1} field is null.
+INVALID_TOKEN_INFO.explanation=Information is missing from the token and validation of the token cannot be completed.
+INVALID_TOKEN_INFO.useraction=Review the log for any earlier errors for details on the failure.
```
dmuelle: this looks good from an ID perspective, except I wonder about the user action- why do they need to review earlier log messages if the error is a null field? Don't they just need to provide a value for the field?

kristip17: Yes, this useraction is rather weak. The path to these variables is not necessarily direct, I plan to check in with the team after the holidays for more options and will request a re-review if we can add more.

kristip17:
If we don't need another PB -- the last PB had a single failure that's a known issue.

dev/io.openliberty.security.oidcclientcore.internal/resources/io/openliberty/security/oidcclientcore/internal/resources/OidcClientCoreMessages.nlsprops
ayoho: I'll need to look into this one a bit. It's possible the use case for this message (null `JwtContext` or `JwtClaims`) might not be possible and that we would have encountered an exception if something were wrong with the token.

dev/io.openliberty.security.oidcclientcore.internal/src/io/openliberty/security/oidcclientcore/token/TokenResponseValidator.java
```diff
@@ -101,8 +101,8 @@ JwtClaims getJwtClaimsFromIdTokenContext(JwtContext jwtContext) throws TokenVali
             jwtClaims = jwtContext.getJwtClaims();
         }
         if (jwtContext == null || jwtClaims == null) {
-            // TODO - NLS message
-            throw new TokenValidationException(this.clientConfig.getClientId(), "not a valid token to continue the flow");
+            String msg = Tr.formatMessage(tc, "INVALID_TOKEN_INFO", new Object[] { this.clientConfig.getClientId(), jwtContext == null ? "jwtContext" : "jwtClaims" });
+            throw new TokenValidationException(this.clientConfig.getClientId(), msg);
```
ayoho:
```suggestion
            String msg = Tr.formatMessage(tc, "TOKEN_MISSING_CLAIMS", new Object[] { this.clientConfig.getClientId());
            throw new TokenValidationException(this.clientConfig.getClientId(), msg);
```
It doesn't look like it's possible for `jwtContext` to be null, so we shouldn't need to check for that here. The `jwtClaims` variable might be null, or at least I can't easily confirm that it will never be null. So I think the best approach here is to throw an exception if we find a null or empty claims object, with the NLS message saying as much.

kristip17: I updated the check and the message.

kristip17:
#build

dmuelle (approved):
LGTM, thanks

ayoho (approved):

kristip17:
#build

kristip17:
#libby

PR #23768, opened 2022-12-23
Get issuer from IdentityToken and add debug on OidcHttpAuthMech

kristip17:
Resolve two ToDos in OidcHttpAuthenticationMechanism:

- Get the issuer from the IdentityToken, if we can't get it, leave it at the default realm.

- Add debug if the OpenIdContext is null


kristip17:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat,io.openliberty.security.jakartasec.3.0.internal_fat.config,io.openliberty.security.jakartasec.3.0.internal_fat.refresh

kristip17:
The PB failure is 1 test and a known issue.

wraschke:
#pb

ayoho (approved):

Issue #19911, opened 2022-01-25
Update ProxyServicesImpl after deprecated methods were removed by Weld 5


tevans78:
Multiple methods from org.jboss.weld.serialization.spi.ProxyServices were removed.

If you are an integrator, you probably implemented this interface at some point in time; these methods weren’t used by Weld anyway, so they are now gone for good.

Here is a link to the issue removing it.
https://issues.redhat.com/browse/WELD-2702

tevans78:
It may be that if the methods are never called then they can be removed. Need to check if they are being called elsewhere.

https://github.com/OpenLiberty/open-liberty/blob/e65a099ee72b33bb86b665f450e23642984a8578/dev/io.openliberty.cdi.4.0.internal.weld/src/com/ibm/ws/cdi/proxy/AbstractProxyServices.java#L87

https://github.com/OpenLiberty/open-liberty/blob/e65a099ee72b33bb86b665f450e23642984a8578/dev/io.openliberty.cdi.4.0.internal.weld/src/com/ibm/ws/cdi/proxy/AbstractProxyServices.java#L187

benjamin-confino:
The final part of this was completed with PR 23816

PR #23780, opened 2023-01-03
19911 - make getClassloader private

benjamin-confino:
This method was removed from the interface in EE10 but is still used by an method that remains in the interface

Part of #19911


benjamin-confino:
#build

Yas-Franco (approved):
Looks good!

PR #23816, opened 2023-01-05
Remove privileges from proxy services

benjamin-confino:
final commit for #19911

### DIFF
--- a/dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
+++ b/dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
@@ -28,8 +28,6 @@ import org.osgi.framework.BundleReference;
 import org.osgi.framework.Constants;
 import org.osgi.framework.wiring.BundleWiring;
 
-import java.util.Optional;
-
 import com.ibm.ws.cdi.CDIRuntimeException;
 
 /**
--- a/dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
+++ b/dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
@@ -11,6 +11,7 @@
 package com.ibm.ws.cdi.proxy;
 
 import java.security.AccessController;
+import java.security.AccessControlException;
 import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
@@ -26,6 +27,8 @@ import org.osgi.framework.BundleException;
 import org.osgi.framework.BundleReference;
 import org.osgi.framework.Constants;
 import org.osgi.framework.wiring.BundleWiring;
+
+import java.util.Optional;
 
 import com.ibm.ws.cdi.CDIRuntimeException;
 
@@ -53,58 +56,43 @@ public class ProxyServicesImpl implements ProxyServices {
 
     @Override
     public ClassLoader getClassLoader(final Class<?> proxiedBeanType) {
-        return AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
-            @Override
-            public ClassLoader run() {
-                // Must always use the bean type's classloader;
-                // Otherwise package private access does not work.
-                // Unfortunately this causes us issues for types from OSGi bundles.
+        // Must always use the bean type's classloader;
+        // Otherwise package private access does not work.
+        // Unfortunately this causes us issues for types from OSGi bundles.
 
-                // It would be nice if we could have a marking header that allowed for
-                // bundles to declare they provide CDI bean types, but this becomes
-                // problematic for interface types that beans may be using for
-                // injection types because the exporter may have no idea their types
-                // are going to be used for CDI.  Therefore we have no way of knowing
-                // ahead of time what bundles are providing CDI bean types.
+        // It would be nice if we could have a marking header that allowed for
+        // bundles to declare they provide CDI bean types, but this becomes
+        // problematic for interface types that beans may be using for
+        // injection types because the exporter may have no idea their types
+        // are going to be used for CDI.  Therefore we have no way of knowing
+        // ahead of time what bundles are providing CDI bean types.
 
-                // This makes it impossible to use weaving hooks to add new dynamic
-                // import packages.  The weaving hook approach requires
-                // a weaving hook registration that knows ahead of time what
-                // bundles provide CDI bean types and then on first class define using
-                // that bundle's class loader the weaving hook would add the necessary
-                // weld packages as dynamic imports.  We cannot and will
-                // not be able to know exactly which bundles are providing bean
-                // types until this getClassLoader method is called.  But by the time
-                // this method is called it is too late for a weaving hook to do
-                // anything because weld is going to use the returned class loader
-                // immediately to reflectively define a proxy class.  The class loader
-                // MUST have visibility to the weld packages before this reflective
-                // call to defineClass.
-                ClassLoader cl = proxiedBeanType.getClassLoader();
-                if (cl == null) {
-                    cl = CLASS_LOADER_FOR_SYSTEM_CLASSES;
-                } else if (cl instanceof BundleReference) {
-                    Bundle b = ((BundleReference) cl).getBundle();
-                    addWeldDynamicImports(b, WELD_PACKAGES);
-                }
-                return cl;
-            }
-        });
+        // This makes it impossible to use weaving hooks to add new dynamic
+        // import packages.  The weaving hook approach requires
+        // a weaving hook registration that knows ahead of time what
+        // bundles provide CDI bean types and then on first class define using
+        // that bundle's class loader the weaving hook would add the necessary
+        // weld packages as dynamic imports.  We cannot and will
+        // not be able to know exactly which bundles are providing bean
+        // types until this getClassLoader method is called.  But by the time
+        // this method is called it is too late for a weaving hook to do
+        // anything because weld is going to use the returned class loader
+        // immediately to reflectively define a proxy class.  The class loader
+        // MUST have visibility to the weld packages before this reflective
+        // call to defineClass.
+        ClassLoader cl = proxiedBeanType.getClassLoader();
+        if (cl == null) {
+            cl = CLASS_LOADER_FOR_SYSTEM_CLASSES;
+        } else if (cl instanceof BundleReference) {
+            Bundle b = ((BundleReference) cl).getBundle();
+            addWeldDynamicImports(b, WELD_PACKAGES);
+        }
+            return cl;
     }
-
+    
     @Override
     public Class<?> loadBeanClass(final String className) {
-        //This is tricky. Sometimes we need to use app classloader to load some app class
-        try {
-            return (Class<?>) AccessController.doPrivileged(new PrivilegedExceptionAction<Object>() {
-                @Override
-                public Object run() throws Exception {
-                    return Class.forName(className, true, getClassLoader(this.getClass()));
-                }
-            });
-        } catch (PrivilegedActionException pae) {
-            throw new CDIRuntimeException(pae.getException());
-        }
+        throw new UnsupportedOperationException("This method is not implemented");
     }
 
     protected void addWeldDynamicImports(Bundle b, ManifestElement[] dynamicImports) {
--- a/dev/com.ibm.ws.cdi.2.0.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
+++ b/dev/com.ibm.ws.cdi.2.0.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
@@ -12,6 +12,8 @@ package com.ibm.ws.cdi.proxy;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.ReflectPermission;
+
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException;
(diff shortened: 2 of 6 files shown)


benjamin-confino:
#build

benjamin-confino:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_R_HXwI2nEe2GjtfTRRcnxQ&tab=com.ibm.team.build.web.ui.internal.editors.result.test.TestProvider

With java2 security enabled.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -53,6 +53,14 @@ public void cleanup() {
 
     @Override
     public ClassLoader getClassLoader(final Class<?> proxiedBeanType) {
+
+        if (System.getSecurityManager() != null) {
+            String className = new Exception().getStackTrace()[1].getClassName();
+            if (! className.equals("org.jboss.weld.bean.proxy.ProxyFactory")) {
+                throw new RuntimeException("A potential security breach was detected");
```
Azquelt: This needs a better error message. It might be better to restructure so that if `ProxyFactory` is the caller we use `doPrivileged`, and if it isn't then we don't and then we don't have to worry about throwing an exception ourselves.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -53,6 +53,14 @@ public void cleanup() {
 
     @Override
     public ClassLoader getClassLoader(final Class<?> proxiedBeanType) {
+
+        if (System.getSecurityManager() != null) {
+            String className = new Exception().getStackTrace()[1].getClassName();
```
Azquelt: You can do `Thread.currentThread().getStackTrace()` instead of creating an exception.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -53,6 +53,14 @@ public void cleanup() {
...
+            if (! className.equals("org.jboss.weld.bean.proxy.ProxyFactory")) {
```
Azquelt: I think this class is exported so we should be able to use the class directly rather than its name?

dev/com.ibm.ws.cdi.2.0.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -40,149 +36,142 @@
...
-
-	private static enum ClassLoaderMethods {
-		;//No enum instances
-
-		private static final Method defineClass1, defineClass2, getClassLoadingLock;
-
-		static {
-			try {
-				Method[] methods = AccessController.doPrivileged(new PrivilegedExceptionAction<Method[]>() {
-					public Method[] run() throws Exception {
```
Azquelt, line 53: I think this part is more complicated.

We don't really want this static initializer to ever fail, regardless of the permissions of the code on the stack at the time of the call, since if it does fail then the class is completely unusable.

However, we do then need to make sure that no-one can abuse the `Method`s that we're pulling out, so we may need a check around where they're invoked.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
benjamin-confino: StackTraceElement doesn't have any reference to the Class, just a string for it's name. I could get the name from ProxyFactory.class but that seems unnecessary.

dev/io.openliberty.cdi.4.0.internal.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -47,11 +54,9 @@ private static enum ClassLoaderMethods {
 
         private static final Method defineClass1, defineClass2, getClassLoadingLock;
 
-        //This will be evaluated lazily when ClassLoaderMethods is first called. 
         static {
             try {
                 Method[] methods = AccessController.doPrivileged(new PrivilegedExceptionAction<Method[]>() {
-                    @Override
```
Azquelt: These can go back in

dev/io.openliberty.cdi.4.0.internal.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -91,9 +96,6 @@ public void cleanup() {
     }
 
     private ClassLoader getClassLoader(final Class<?> proxiedBeanType) {
-        return AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
-            @Override
-            public ClassLoader run() {
                 // Must always use the bean type's classloader;
                 // Otherwise package private access does not work.
```
Azquelt: Need to reformat the indented text here.

dev/com.ibm.ws.cdi.2.0.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -39,150 +41,159 @@
...
-	}
-
-	private Class<?> loadClass(String classBinaryName, ClassLoader cl) throws ClassNotFoundException {
-		return Class.forName(classBinaryName, true, cl);
-	}
-
-	public boolean supportsClassDefining() {
-		return true;
-	}
+    SecurityManager sm = System.getSecurityManager();
```
Azquelt: I think you need to call `getSecurityManager` every time as it _could_ change.

dev/com.ibm.ws.cdi.2.0.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -39,150 +41,159 @@
...
+    ReflectPermission reflectPermission = new ReflectPermission("suppressAccessChecks");
+    RuntimePermission declaredMembersPermission = new RuntimePermission("accessDeclaredMembers");
+    RuntimePermission getClassLoaderPermission = new RuntimePermission("getClassLoader");
```
Azquelt: These are constants so they can be `private static final` (and capitalized)

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
Azquelt: Ok, it doesn't matter in this case since tests will fail if the name is wrong.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -53,6 +54,14 @@ public void cleanup() {
 
     @Override
     public ClassLoader getClassLoader(final Class<?> proxiedBeanType) {
+
+        if (System.getSecurityManager() != null) {
+            String className = Thread.currentThread().getStackTrace()[1].getClassName();
+            if (! className.equals("org.jboss.weld.bean.proxy.ProxyFactory")) {
+                throw new AccessControlException("Someone other than weld attempted to access proxy services");
+            }
```
Azquelt: Do we need either a similar check or the removal of `doPrivileged` further down in `loadBeanClass`?

Also, it looks like `loadBeanClass` calls `getClassLoader` and that would break your current stack-walking logic since there'd be an extra call in the stack. Are we testing this code path?

dev/com.ibm.ws.cdi.2.0.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -39,150 +41,162 @@
...
-	}
-
-	private Class<?> loadClass(String classBinaryName, ClassLoader cl) throws ClassNotFoundException {
-		return Class.forName(classBinaryName, true, cl);
-	}
-
-	public boolean supportsClassDefining() {
-		return true;
-	}
+    private static final ReflectPermission REFLECT_PERMISSION = new ReflectPermission("suppressAccessChecks");
```
Azquelt: Can you change this to `SUPPRESS_ACCESS_CHECKS_PERMISSION` please? It would better describe what it does.

benjamin-confino:
Good build here: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_zN9gwJZDEe2YL83b0ePHDw&tab=com.ibm.team.build.web.ui.internal.editors.result.properties.PropertiesProvider

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -27,6 +28,8 @@
 import org.osgi.framework.Constants;
 import org.osgi.framework.wiring.BundleWiring;
 
+import java.util.Optional;
```
benjamin-confino: TODO. Remove this unused import.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -132,5 +139,32 @@ protected void addWeldDynamicImports(Bundle b, ManifestElement[] dynamicImports)
...
+            for (int i = 0; i < stackClasses.length; i++) {
+                if (! stackClasses[i].equals(this.getClass())) {
+                    continue;
+                }
+                if (stackClasses[i].equals(org.jboss.weld.bean.proxy.ProxyFactory.class)) {
+                    break;
+                } else {
+                    throw new AccessControlException("Someone other than weld attempted to access proxy services");
+                }
+            }
```
Azquelt: This logic doesn't look right.

```java
                if (! stackClasses[i].equals(this.getClass())) {
                    continue;
                }
```
Doesn't this skip examining any class that isn't the current class? I think the `!` should be removed?

Do we have tests for this? This code looks like it should fail in all cases - the first thing on the stack would be this class so the check on 158 would fail, then the check on 161 would fail because this class is not `ProxyFactory` and we'd throw an `AccessControlException`?

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -132,5 +139,32 @@ protected void addWeldDynamicImports(Bundle b, ManifestElement[] dynamicImports)
             throw new RuntimeException(e.getTargetException());
         }
     }
+
+    private void stackCheck() {
```
Azquelt: A javadoc comment for this method stating what it's intending to check would be helpful.

Maybe something like `checkCallerIsWeld()` would be a better name, but I'm not super concerned about that.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
benjamin-confino: I've been using the passivation tests to test for this since the weld docs suggested they'd trigger this method. Clearly they don't. Now I wonder if anything we're doing goes through here.

Azquelt: We should be able to tell from trace whether it does or not.

benjamin-confino:
#build

benjamin-confino:
#build

dev/com.ibm.ws.org.jboss.weld/src/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -0,0 +1,1001 @@
...
+ * Main factory to produce proxy classes and instances for Weld beans. This
+ * implementation creates proxies which forward non-static method invocations to
+ * a {@link BeanInstance}. All proxies implement the {@link Proxy} interface.
+ *
+ * @author David Allen
+ * @author Stuart Douglas
+ * @author Marius Bogoevici
+ * @author Ales Justin
+ */
+public class ProxyFactory<T> implements PrivilegedAction<T> {
```
Azquelt, line 94: As a requirement of the Apache 2.0 license (4b), this file needs a prominent notice saying that we have modified it.

In addition, we should outline how and why we modified it in the class doc and add comments around all of our changes.

dev/com.ibm.ws.cdi.1.2.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -53,58 +54,43 @@ public void cleanup() {
...
-                @Override
-                public Object run() throws Exception {
-                    return Class.forName(className, true, getClassLoader(this.getClass()));
-                }
-            });
-        } catch (PrivilegedActionException pae) {
-            throw new CDIRuntimeException(pae.getException());
-        }
+        throw new UnsupportedOperationException("This method is not implemented");
     }
```
Azquelt, line 94: It seems odd to change this version to throw `UnsupportedOperationException` but in the 2.0 version we removed the `doPrivileged` call.

I think I'd lean towards just removing the `doPrivileged` call in both, but I'd expect them to be the same unless there's some reason why they shouldn't be.

benjamin-confino:
#build

benjamin-confino:
#build

Azquelt (approved):

benjamin-confino:
#libby

PR #23822, opened 2023-01-05
ID on an embeddable

njr-11:
Be able to define the ID attribute on an embedded class.

### DIFF
--- a/dev/io.openliberty.data.internal.persistence/src/io/openliberty/data/internal/persistence/EntityDefiner.java
+++ b/dev/io.openliberty.data.internal.persistence/src/io/openliberty/data/internal/persistence/EntityDefiner.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022 IBM Corporation and others.
+ * Copyright (c) 2022,2023 IBM Corporation and others.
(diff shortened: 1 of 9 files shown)


njr-11:
#build

dev/io.openliberty.data.internal.persistence/src/io/openliberty/data/internal/persistence/EntityDefiner.java
```diff
@@ -90,7 +92,7 @@ else if ((id == null || id.length() != 2) && name.endsWith("ID"))
         }
 
         if (id == null)
-            throw new IllegalArgumentException(entityClass + " lacks public field with @Id or of the form *ID");
+            throw new MappingException(entityClass + " lacks public field with @Id or of the form *ID"); // TODO
```
KyleAure, line 95: Is the TODO here to translate this message in the future?

KyleAure (approved):

dev/io.openliberty.data.internal.persistence/src/io/openliberty/data/internal/persistence/EntityDefiner.java
njr-11: yes

PR #23824, opened 2023-01-05
Do not register monitor metrics when MetricRegistry is null

### LINKED ISSUE #23823: MP Metrics 5.0: Monitor metrics bundle attempts registers metric when metric runtime initialization fails.
When initialization the MP Metrics runtime (i.e. underlying SR and Micrometer implementation), failures can occur which stops the Metrics runtime from being initialized properly. The related monitor metrics bundle `io.openliberty.microprofile.metrics.5.0.monitor.internal` which will start as part of the feature set will continue to initialize and attempt to register vendor (and certain base) metrics. If the MP Metrics runtime fails to initialize, a `null` MetricRegistry is provided to the monitor metrics bundle. Currently, it does not check for a `null` value and proceeds to attempt a registration which leads to a `NullPointerException`.

Channyboy:
#build
fixes #23823

### DIFF
--- a/dev/io.openliberty.microprofile.metrics.5.0.monitor.internal/src/io/openliberty/microprofile/metrics/internal/monitor/MonitorMetricsHandler.java
+++ b/dev/io.openliberty.microprofile.metrics.5.0.monitor.internal/src/io/openliberty/microprofile/metrics/internal/monitor/MonitorMetricsHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2022 IBM Corporation and others.
+ * Copyright (c) 2020, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


Channyboy:
#build

Channyboy:
#libby

fmhwong:
Are there anything the user can do when `metricRegistry` is `null`?

Channyboy:
@fmhwong Nope. If the MetricRegistry retrieved from `SharedMetricRegistries` is null then the initiliazation of the SmallRye Metrics (or Micrometer related libraries) failed and the feature is in-operable.

fmhwong:
> @fmhwong Nope. If the MetricRegistry retrieved from `SharedMetricRegistries` is null then the initiliazation of the SmallRye Metrics (or Micrometer related libraries) failed and the feature is in-operable.

Since you just write a debug message only, should we write an info or warning message telling customer how to fix the configuration?

Channyboy:
Yup, that was part of : https://github.com/OpenLiberty/open-liberty/pull/23830

fmhwong (approved):
LGTM

PR #23825, opened 2023-01-05
Fix test failure for RTC 294304. The test will now wait for the appli…

jvanhill:
…cation to start before making calls to the application.


jvanhill:
#build

Zech-Hein (approved):

Issue #23452, opened 2022-11-22
Test telemetry configuration


Azquelt:
We must implement all the config properties relating to tracing listed in [OTel SDK Autoconfigure][autoconf], with the following changes:

* Config properties are read from MP Config
* Tracing is disabled by default

We must test:

- [x]  each applicable config property which isn't used elsewhere is respected
- [x]  the capitalisation rules for environment variables in MP Config are respected (if we do anything other than call `Config.getValue()`) 
- [x]  the ordering of MP Config sources is respected
- [x] tracing is disabled by default

[autoconf]: https://github.com/open-telemetry/opentelemetry-java/tree/v1.19.0/sdk-extensions/autoconfigure

Azquelt:
Also to check: should all config parameters be made available to OTel, or only those starting `otel.`?

PR #23848, opened 2023-01-10
Test Telemetry config

Yas-Franco:
For issue #23452 

Tests the MPTelemetry config properties and the expected overriding values.

Changes the AutoConfiguredOpenTelemetrySdk.builder() to use the `addPropertiesCustomizer` so that system properties no longer get priority


dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -131,20 +131,16 @@ private boolean checkDisabled(Map<String, String> oTelConfigs) {
     private HashMap<String, String> getTelemetryProperties() {
         HashMap<String, String> telemetryProperties = new HashMap<>();
         for (String propertyName : config.getPropertyNames()) {
-            if (propertyName.startsWith("otel.") || propertyName.startsWith("OTEL_")) {
-                config.getOptionalValue(propertyName, String.class).ifPresent(
-                                                                              value -> telemetryProperties.put(propertyName, value));
-
+            if (propertyName.startsWith("otel.")) {
+                telemetryProperties.put(propertyName,config.getValue(propertyName, String.class));
             }
```
Azquelt: I suspect this fails for an environment variable named `otel_service_name` if `otel.service.name` is not also defined in another config source.

In most scenarios you do get away with it:
- if `otel.service.name` is defined in any config source, then we'll look its value up from MP Config and we'll find `otel_service_name` in the environment.
- if `OTEL_SERVICE_NAME` is defined in the environment and nowhere else then we won't add it here, but OTel's own config logic will find it.

I think the nearest thing we can get to correct is:
```
If (propertyName starts with `otel` ignoring case) {
  look up the value and add it to the map
  lowercase the name and convert `_` to `.`
  look up that name, if it has a value then add that to the map
}
```

This assumes that all properties start `otel` and contain only lowercase alphanumeric characters and `.` (which I _think_ is good enough, I think some properties contain `-` but OTel will accept `.` instead)

Azquelt: I think you also have a problem here if a config property is set in one config source, but then unset by a config source with a higher priority.

E.g. `foo=bar` in the environment, and then `foo=` as a system property.

I think in this case, `config.getPropertyNames()` will include `foo`, but `config.getValue("foo")` will throw an exception.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryConfigEnvTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+        server.addEnvVar("OTEL_SERVICE_NAME", "overrideDone");
+        server.addEnvVar("OTEL_SDK_DISABLED", "false");
+        server.startServer();
+    }
+
+    @AfterClass
+    public static void tearDown() throws Exception {
+        server.stopServer("CWNEN0047W.*ConfigServlet");
```
Azquelt: What does this warning mean and why do we expect it? Please add a comment.

Looks like `CWNEN0047W: Resource annotations on the fields of the {0} class will be ignored. The annotations could not be obtained because of the exception : {1}` and I'm not sure why we'd expect that if the test is working correctly.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/OpenTelemetryBeanServlet.java
```diff
@@ -44,7 +45,6 @@ public void testOpenTelemetryBean() {
         Tracer tracer = openTelemetry.getTracer("instrumentation-test", "1.0.0");
         assertNotNull(tracer);
         Span span = tracer.spanBuilder(SPAN_NAME).startSpan();
-        assertThat(span.getSpanContext().getSpanId(), not(equalTo(INVALID_SPAN_ID)));
```
Azquelt, line 47: What's this change about? It doesn't seem related to this PR.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/publish/servers/Telemetry10/server.xml
```diff
@@ -7,9 +7,14 @@
...
+	<variable name="otel.service.name" value="overrideThisServerVar"/>
+	 
 	<application id="TelemetryApp" name="TelemetryApp" type="war" location="TelemetryApp.war">
     	   <classloader apiTypeVisibility="+third-party"/>
-        </application>
+    	   	<appProperties>
+        		<property name="otel.service.name" value="overrideDone"/>
+				<property name="otel.sdk.disabled" value="false"/>
+    		</appProperties>
+    </application>
```
Azquelt: You've got a mix of tabs and spaces here

Azquelt: Also in some of the other server.xml files.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryConfigSystemPropTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @BeforeClass
+    public static void setUp() throws Exception {
+        WebArchive app = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                        .addAsResource(ConfigServlet.class.getResource("microprofile-config.properties"), "META-INF/microprofile-config.properties")
+                        .addClasses(ConfigServlet.class);
+
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+        server.addEnvVar("OTEL_SERVICE_NAME", "overrideThisEnvVar");
+        server.addEnvVar("OTEL_SDK_DISABLED", "true");
+        server.startServer();
```
Azquelt, line 65: When you're putting in variables like this which are to be overridden elsewhere, please add a comment saying where the value is overridden.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryConfigEnvTest.java
```diff
@@ -0,0 +1,68 @@
...
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.ConfigServlet;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.MetricsDisabledServlet;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.OpenTelemetryBeanServlet;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.PatchTestApp;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.SpanCurrentServlet;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.Telemetry10Servlet;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.WithSpanServlet;
+
+@RunWith(FATRunner.class)
+public class TelemetryConfigEnvTest extends FATServletClient {
```
Azquelt, line 44: You can add `@Mode(FULL)` to this, `TelemetryConfigServerVarTest` and `TelemetryConfigSystemPropertyTest`.

They all have quite a bit of overhead to test non-basic scenarios.

Yas-Franco:
#build
#libby

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
Azquelt: For the second part, I see you've made the change to use `getOptionalValue`.

For the first part, I see you've added a test for an environment variable set on its own. I guess OTel is picking this up directly from config which is ok. If it's set in another config source, we will search MP Config for it and find the environment variable version.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryConfigSystemPropTest.java
```diff
@@ -56,13 +59,14 @@ public static void setUp() throws Exception {
                         .addClasses(ConfigServlet.class);
 
         ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+        //Set for testing purposes. The properties in the boostrap.properties
         server.addEnvVar("OTEL_SERVICE_NAME", "overrideThisEnvVar");
         server.addEnvVar("OTEL_SDK_DISABLED", "true");
```
Azquelt:
```suggestion
        // These should be overridden by the values in bootstrap.properties
        server.addEnvVar("OTEL_SERVICE_NAME", "overrideThisEnvVar");
        server.addEnvVar("OTEL_SDK_DISABLED", "true");
```

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryConfigNullTest.java
```diff
@@ -0,0 +1,72 @@
...
+    @BeforeClass
+    public static void setUp() throws Exception {
+        WebArchive app = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                        .addClasses(ConfigServlet.class);
+
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+
+        //These variables take priority and the app property in the server.xml should not throw an exception
+        server.addEnvVar("otel_service_name", "overrideDone");
+        server.addEnvVar("otel_sdk_disabled", "false");
```
Azquelt: This one maybe needs a better explanation.

```suggestion
        // These properties are set to the empty string in server.xml which should lead to there being no value set for these properties in MP Config
        // HOWEVER, OTel does its own reading of environment variables if there's no value in MP Config to override it, so it will see the values set here
        // This is not ideal and is a current limitation, but we still want to test the scenario to ensure we don't throw an exception when processing config
        server.addEnvVar("otel_service_name", "overrideDone");
        server.addEnvVar("otel_sdk_disabled", "false");
```

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryConfigEnvOnlyTest.java
```diff
@@ -0,0 +1,72 @@
...
+    @BeforeClass
+    public static void setUp() throws Exception {
+        WebArchive app = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                        .addClasses(ConfigServlet.class);
+
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+
+        //These variables take priority
+        server.addEnvVar("otel_service_name", "overrideDone");
+        server.addEnvVar("otel_sdk_disabled", "false");
```
Azquelt: These aren't actually taking priority over anything, but an explanation of what we're testing here would be good.

```suggestion
        // This tests that we do pick up config properties which are _only_ defined in the environment, using a format which MP Config only accepts for environment variables.
        server.addEnvVar("otel_service_name", "overrideDone");
        server.addEnvVar("otel_sdk_disabled", "false");
```

Yas-Franco:
#libby

Yas-Franco:
#build

Azquelt (approved):

Azquelt:
#libby

PR #23858, opened 2023-01-10
dropTables config property (internal)

njr-11:
Users who experiment with Jakarta Data will need a way to remove tables when they change the definition of an entity.  The specification hasn't defined a configuration model yet, so for now we will reuse databaseStore and add a dropTables to it.


nmittles (approved):

njr-11:
#build

PR #23884, opened 2023-01-11
Synchronously touch all JSPs during checkpoint

dettmoney:
JSPs by default are compiled at first request. In a checkpoint/restore scenario, let's precompile all of the JSPs at checkpoint time to reduce first response times.

### DIFF
--- a/dev/com.ibm.ws.jsp/bnd.bnd
+++ b/dev/com.ibm.ws.jsp/bnd.bnd
@@ -168,4 +168,5 @@ instrument.classesExcludes: com/ibm/ws/jsp/resources/*.class
 	com.ibm.ws.javaee.version;version=latest, \
 	io.openliberty.jakarta.servlet.jsp.jstl.2.0.internal;version=latest, \
 	io.openliberty.jakarta.servlet.jsp.tags.3.0.internal;version=latest, \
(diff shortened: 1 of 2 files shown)


dettmoney:
#build

dettmoney:
#build

dettmoney:
A better solution to this might be to change the JspOptions.isPrepareJSPsSet() method to return true during checkpoint, but  only if it wasn't set to false explicitly in the server.xml. This gives the customer a way to go back to the original behavior if they wanted to do that. The logic in startPreTouch would then slightly change.

dettmoney:
#build

dettmoney:
#build

volosied:
@dettmoney  Is this part of an epic or another issue? Can you give us more information?

epj:
I also wonder about the lack of any FAT enhancements to test this change.   In particular there should be tests of this change in conjunction with different settings of prepareJSPs (on the jspEngine element in server.xml).

volosied:
@dettmoney Could you rebase against origin/integration. This PR has over 700 commits.  

Are there any tests for these changes?

tjwatson:
@dettmoney I think you are going to have to cherry-pick only the commit you want for this PR and redo your PR branch.

Issue #23867, opened 2023-01-11
Automatically populate Resource attributes


Azquelt:
The resource attribute [otel.service.name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service) should be provided by the SDK. 

We should:

- [x] Provide an appropriate default for the service name
- [x] Test that the default is used when the user has not specified a service name manually

PR #23892, opened 2023-01-12
Set Telemetry default service name

Yas-Franco:
For issue #23867

Provides the app name as the default for the tracing service name and tests this is done when no configuration is given.

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
@@ -100,8 +100,8 @@ public class OpenTelemetryProducer {
     }
 
     //Uses application name if the user has not given configured service.name resource attribute
-    private String getServiceName() {
-        String appName = config.getOptionalValue(SERVICE_NAME_PROPERTY, String.class).orElse(null);
+    private String getServiceName(ConfigProperties c) {
+        String appName = c.getString(SERVICE_NAME_PROPERTY);
         ComponentMetaData cmd = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData();
         if (appName == null) {
             if (cmd != null) {
@@ -114,7 +114,7 @@ public class OpenTelemetryProducer {
     //Adds the service name to the resource attributes
     private Resource customizeResource(Resource resource, ConfigProperties c) {
         ResourceBuilder builder = resource.toBuilder();
-        builder.put(ResourceAttributes.SERVICE_NAME, getServiceName());
+        builder.put(ResourceAttributes.SERVICE_NAME, getServiceName(c));
         return builder.build();
     }
 
@@ -161,9 +161,9 @@ public class OpenTelemetryProducer {
         HashMap<String, String> telemetryProperties = new HashMap<>();
         for (String propertyName : config.getPropertyNames()) {
             if (propertyName.startsWith("otel.")) {
-               config.getOptionalValue(propertyName, String.class).ifPresent(
-                                                                            value -> telemetryProperties.put(propertyName, value));
-            }    
+                config.getOptionalValue(propertyName, String.class).ifPresent(
+                                                                              value -> telemetryProperties.put(propertyName, value));
+            }
         }
         //Metrics and logs are disabled by default
         telemetryProperties.put(CONFIG_METRICS_EXPORTER_PROPERTY, "none");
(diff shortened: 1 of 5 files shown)


Yas-Franco:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/ServiceNameServlet.java
```diff
@@ -0,0 +1,51 @@
...
+    @Inject
+    OpenTelemetry openTelemetry;
+
+    //Tests if otel.service.name is set to the application name by default
+    @Test
+    public void testServiceNameConfig() {
+        Tracer tracer = openTelemetry.getTracer("config-test", "1.0.0");
+        Span span = tracer.spanBuilder("testSpan").startSpan();
+        span.end();
+        assertThat(openTelemetry.toString(), containsString("service.name=\"TelemetryApp\""));
```
Azquelt: We should test the attributes on the span, not by doing `toString()` on the `openTelemetry` object. Use `InMemorySpanExporter`.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -129,7 +141,15 @@ private boolean checkDisabled(Map<String, String> oTelConfigs) {
...
 
     private HashMap<String, String> getTelemetryProperties() {
+
         HashMap<String, String> telemetryProperties = new HashMap<>();
+
+        String defaultServiceName = lookupAppName();
+
+        if (defaultServiceName != null) {
+            telemetryProperties.put("otel.service.name", defaultServiceName);
+        }
```
Azquelt: It's probably better to use [addResourceCustomizer](https://www.javadoc.io/static/io.opentelemetry/opentelemetry-sdk-extension-autoconfigure/1.22.0-alpha/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.html#addResourceCustomizer(java.util.function.BiFunction)) rather than setting default config.

It gives you control over whether you are overriding config or not and avoids config edge cases (e.g. if OTel checks `otel.service.name` before `OTEL_SERVICE_NAME` there's a chance that it picks up your default rather than `OTEL_SERVICE_NAME` defined in the server environment.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -89,6 +99,25 @@ public OpenTelemetry getOpenTelemetry() {
 
     }
 
+    //Uses application name if the user has not given configured service.name resource attribute
+    private String getServiceName() {
+        String appName = config.getOptionalValue(SERVICE_NAME_PROPERTY, String.class).orElse(null);
```
Azquelt: At this point, it probably makes more sense to use the `ConfigProperties` object to get the service name, rather than going back to MP Config.

We already put everything from MP Config into the OTel's `ConfigProperties` and we'd expect everything else at this point to use `ConfigProperties`.

Azquelt (changes requested):
Looks good, added one comment.

Also, you've built on top of #23903 which is good (to avoid conflicts where the InMemorySpanExporter was moved), but it looks like you've rebased the commits from that PR which you can't do.

Yas-Franco:
#libby
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
Azquelt: This part hasn't been changed.

Azquelt:
You've got a few stray commits on your branch. Now that the PRs you were rebasing on have been merged, you should be able to interactive rebase on integration and drop any commits which aren't part of your change.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
Yas-Franco: It's in the last commit here: 

https://github.com/OpenLiberty/open-liberty/pull/23892/commits/35a4d8eedd010926f93f4894c9aa786b1de9fbad

Azquelt: Ah, thank you, I see it now.

Azquelt (approved):
Looks great! Thank you.

Yas-Franco:
#libby
#build

PR #23898, opened 2023-01-12
Fix FragmentOrderTest timing problem

### LINKED ISSUE #23896: Test Failure: com.ibm.ws.tests.anno.caching.FragmentOrderTest.fragmentOrder_testCacheValidate
On some platform/JDK combinations (mostly windows) we do not have millisecond precision on files, so if you update a file within 1 second it won’t be noticed. In this case, it finishes a test and updates the application within .2 seconds so the update is never recognized.

jimblye:
Fixes #23896


jimblye:
#build

brenthdaniel (approved):
It would be slightly better to have the sleep before the EAR replacement just in case the previous update and server start all happened within the same second. That's pretty unlikely, though, so this is fine.

Issue #23758, opened 2022-12-20
RESTfulWs-3.0 - CDI only injects during a request


WhiteCat22:
CDI is only injecting during a RESTful request, not during Application initialization or Client filters when no Application subclass exists (REST client in a ServletFilter for example)

This means that CDI will not inject into your Client Filters if your client is ran outside of a RESTful application. In a Servlet for example.

PR #23903, opened 2023-01-13
OpenTelemetry multi-app tests and client filter injection fix

### LINKED ISSUE #23461: Test multiple application support with MP Telemetry
* Creation of separate `Resource` for each application with appropriate [attributes][resource-attributes]
* Have one application call the other and check that the created spans for each application have appropriate attributes

[resource-attributes]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.13.0/specification/resource/semantic_conventions/README.md

### LINKED ISSUE #23736: Telemetry NullPointerException in client filter
When an application uses the client part of JAX-RS but doesn't use the server part, the `TelemetryClientFilter` is not created by CDI, is not injected and does not have its `PostConstruct` method called. This results in a `NullPointerException` when the filter method tries to use one of the uninitialized fields.

In my test scenario this happens when the app uses a JAX-RS client and does not contain an `Application` subclass. I'm not sure if it can happen in other situations.

Azquelt:
This threw up a new problem where JAX-RS can get the current application wrong when using JAX-RS client without the JAX-RS server component and passes this wrong information onto CDI. MP Config, on the other hand, has the right application. There's a bug open for this #23758

This resulted in the `OpenTelemetry` bean for app2 being created using the config for app1, because JAX-RS thought it was in app2 when it was actually in app1.

To avoid that, we need to work around the way we use CDI in the OpenTelemetry client filter, which also fixes #23736 

Fixes #23736 
Fixes #23461

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2022 IBM Corporation and others.
+ * Copyright (c) 2022, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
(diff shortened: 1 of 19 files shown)


Azquelt:
#build
#libby

benjamin-confino (dismissed):

Azquelt:
#libby
#build

Azquelt:
#build

benjamin-confino (dismissed):

Azquelt:
#libby
#build

Azquelt:
Rebased on #23903 again

Azquelt:
Rebased on integration. Previous build request hasn't started yet so it will pick up the new change.

benjamin-confino (approved):

Azquelt:
#libby

Issue #23925, opened 2023-01-16
The io.opentelemetry.api.baggage.propagation package is not visible to the application


tevans78:
Trying to access the package results in an exception

```
junit.framework.AssertionFailedError: 2023-01-16-17:29:10:779 ERROR: Caught exception attempting to call test method testW3CBaggagePropagator on servlet io.openliberty.microprofile.telemetry.internal_fat.apps.api.BaggageAPIServlet
java.lang.NoClassDefFoundError: io.opentelemetry.api.baggage.propagation.W3CBaggagePropagator
at io.openliberty.microprofile.telemetry.internal_fat.apps.api.BaggageAPIServlet.testW3CBaggagePropagator(BaggageAPIServlet.java:60)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at componenttest.app.FATServlet.doGet(FATServlet.java:73)
at <unknown class>.service(HttpServlet.java:527)
at <unknown class>.service(HttpServlet.java:614)
at com.ibm.ws.webcontainer.servlet.ServletWrapper.service(ServletWrapper.java:1260)
at com.ibm.ws.webcontainer.servlet.ServletWrapper.handleRequest(ServletWrapper.java:748)
at com.ibm.ws.webcontainer.servlet.ServletWrapper.handleRequest(ServletWrapper.java:445)
at com.ibm.ws.webcontainer.filter.WebAppFilterManager.invokeFilters(WebAppFilterManager.java:1296)
at com.ibm.ws.webcontainer.webapp.WebApp.handleRequest(WebApp.java:5080)
at com.ibm.ws.webcontainer.osgi.DynamicVirtualHost$2.handleRequest(DynamicVirtualHost.java:318)
at com.ibm.ws.webcontainer.WebContainer.handleRequest(WebContainer.java:1009)
at com.ibm.ws.webcontainer.osgi.DynamicVirtualHost$2.run(DynamicVirtualHost.java:283)
at com.ibm.ws.http.dispatcher.internal.channel.HttpDispatcherLink$TaskWrapper.run(HttpDispatcherLink.java:1248)
[...]

PR #23931, opened 2023-01-17
Add baggage.propagation api to Telemetry feature

Yas-Franco:
For issue #23925


Azquelt:
Does this need a test?

I know @tevans78 found this while writing tests, does that mean we already have some ready to be added?

tevans78:
> Does this need a test?
> 
> I know @tevans78 found this while writing tests, does that mean we already have some ready to be added?

At the moment my new test is super basic and is only really checking that the API is accessible. It is not an exhaustive test of the API.
https://github.com/OpenLiberty/open-liberty/pull/23932/files#diff-b265c9191268e742a61494fee19cd7106aae6e94836ae4478c8a750edc6a6562

Azquelt (approved):

Azquelt:
ok, this can be merged as-is and the test from #23932 enabled later (either as part of that PR, or separately)

Yas-Franco:
#libby
#build

PR #23932, opened 2023-01-17
Create Basic OTel API test

### LINKED ISSUE #23447: Basic testing of the OpenTelemetry API
We integrate OpenTelemetry for Java and expose its API classes to applications.

Most of the API will be tested incidentally as part of other tests, but we should have something that explicitly checks that all parts of the API are present and functional.

For the following areas:

* Tracing API ([Javadoc][trace-javadoc])
* Baggage API ([Javadoc][baggage-javadoc])
* Context API ([Javadoc][context-javadoc])
* Resource SDK ([Javadoc][resource-javadoc])

We should:
- [ ] Check that all packages are exposed in the feature
- [ ] Conduct a basic check that the classes and interfaces in the API can be used

[trace-javadoc]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/1.18.0/io/opentelemetry/api/trace/package-summary.html
[baggage-javadoc]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/1.18.0/io/opentelemetry/api/baggage/package-summary.html
[context-javadoc]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-context/1.18.0/index.html
[resource-javadoc]: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-sdk-common/1.18.0/index.html

tevans78:
fixes https://github.com/OpenLiberty/open-liberty/issues/23447

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/BaggageAPIServlet.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/BaggageAPIServlet.java
@@ -1,0 +1,66 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal_fat.apps.api;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Collection;
+
+import org.junit.Test;
+
+import componenttest.app.FATServlet;
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.baggage.BaggageBuilder;
+import io.opentelemetry.api.baggage.BaggageEntryMetadata;
+import io.opentelemetry.api.baggage.propagation.W3CBaggagePropagator;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.servlet.annotation.WebServlet;
+
+@SuppressWarnings("serial")
+@WebServlet("/baggage")
+@ApplicationScoped // Make this a bean so that there's one bean in the archive, otherwise CDI gets disabled and @Inject doesn't work
+public class BaggageAPIServlet extends FATServlet {
+
+    /**
+     * Very simple test that we can use a BaggageBuilder
+     * {@link BaggageBuilder}
+     * {@link Baggage}
+     * {@link BaggageEntryMetadata}
+     */
+    @Test
+    public void testBaggageBuilder() {
+        BaggageBuilder builder = Baggage.builder();
+        BaggageEntryMetadata metadata1 = BaggageEntryMetadata.create("myMetaData1");
+        builder.put("myKey1", "myValue1", metadata1);
+        BaggageEntryMetadata metadata2 = BaggageEntryMetadata.create("myMetaData2");
+        builder.put("myKey2", "myValue2", metadata2);
+        BaggageEntryMetadata metadata3 = BaggageEntryMetadata.create("myMetaData3");
+        builder.put("myKey3", "myValue3", metadata3);
+        Baggage baggage = builder.build();
+
+        assertEquals(3, baggage.size());
+
+        String value = baggage.getEntryValue("myKey2");
+        assertEquals("myValue2", value);
+    }
+
+    /**
+     * Very simple test that we can use a W3CBaggagePropagator
+     * {@link W3CBaggagePropagator}
+     */
+    @Test
+    public void testW3CBaggagePropagator() {
+        W3CBaggagePropagator propagator = W3CBaggagePropagator.getInstance();
+        Collection<String> fields = propagator.fields();
+        assertTrue(fields.contains("baggage"));
+    }
+
+}
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/BaggageAPIServlet.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/BaggageAPIServlet.java
@@ -50,6 +50,9 @@ public class BaggageAPIServlet extends FATServlet {
 
         String value = baggage.getEntryValue("myKey2");
         assertEquals("myValue2", value);
+
+        BaggageEntryMetadata metadataValue = baggage.asMap().get("myKey2").getMetadata();
+        assertEquals("myMetaData2", metadataValue.getValue());
     }
 
     /**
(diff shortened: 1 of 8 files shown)


tevans78:
#pb

tevans78:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/ContextAPIServlet.java
```diff
@@ -0,0 +1,227 @@
...
+    private static final ContextKey<String> MY_KEY = ContextKey.named(MY_KEY_STRING);
+
+    private static class MyTextMapPropagator<T> implements TextMapPropagator {
+        private final ContextKey<T> key;
+        private final T rawKey;
+
+        MyTextMapPropagator(ContextKey<T> key, T rawKey) {
+            this.key = key;
+            this.rawKey = rawKey;
+        }
```
Azquelt: `key`, `rawKey` and `T` don't get used anywhere and can be removed

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/ContextAPIServlet.java
```diff
@@ -0,0 +1,227 @@
...
+        //check that it contains the field we expect
+        assertTrue(textPropagator.fields().contains(MY_KEY_STRING));
+        //create a second context
+        Context contextB = Context.current();
+        //extract the value from the carrier and put it in the context
+        contextB = textPropagator.extract(contextB, carrier, new MyTextMapGetter());
+        String value = contextB.get(MY_KEY);
+        //check that it still matches
+        assertEquals("myValue", value);
+    }
```
Azquelt: The comments are good, but breaking this up into sections would make it easier to read.

I think it would make sense to assert that `carrier` contains the expected key and value after calling `textPropagator.inject`.

The assertions for `ContextPropagators` don't seem to relate to the rest of the test and could be done in a different test or at the end of this test method, rather than in the middle.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/CommonSDKServlet.java
```diff
@@ -0,0 +1,133 @@
...
+    @Test
+    public void testClockNanoTime() throws InterruptedException {
+        Clock clock = Clock.getDefault();
+        long startNano = clock.nanoTime();
+        int sleepNano = 100000;
+        Thread.sleep(0, sleepNano);
+        long endNano = clock.nanoTime();
+        long durationNano = endNano - startNano;
+        assertTrue(durationNano >= sleepNano);
+        assertTrue(durationNano < (10 * sleepNano));
```
Azquelt: I think I'd prefer to remove this assert, we've definitely seen build systems pause for more than a second.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/CommonSDKServlet.java
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @throws InterruptedException
+     */
+    @Test
+    public void testClockNow() throws InterruptedException {
+        Clock clock = Clock.getDefault();
+        long clockNow = clock.now(); //nanos
+        long systemNow = System.currentTimeMillis() * 1000000;
+        System.out.println("Clock: " + clockNow + " System: " + systemNow);
+        assertTrue("Clock: " + clockNow + " System: " + systemNow, clockNow <= systemNow);
```
Azquelt: This may not be true if `clock.now()` has a resolution greater than 1ms.

Although `System.currentTimeMillis()` is called later than `clock.now()`, the time it returns will be rounded to the nearest (or possibly previous) millisecond, so `clockNow` may be later than `systemNow`.

Azquelt: We're more concerned here that the user can use these methods than that they work correctly. Checking that Clock.now() is within 24h of the current time is probably sufficient.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/CommonSDKServlet.java
Azquelt: I think I misread. 100000ns is only 0.1ms, so I think we're actually checking that elapsed duration is <1ms, which is way too short.

I would be tempted to replace any nanosecond time constants and conversions with `Duration`. (i.e. `sleepNano = Duration.of(100, MILLIS).toNanos();`

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/BaggageAPIServlet.java
```diff
@@ -0,0 +1,66 @@
...
+        builder.put("myKey2", "myValue2", metadata2);
+        BaggageEntryMetadata metadata3 = BaggageEntryMetadata.create("myMetaData3");
+        builder.put("myKey3", "myValue3", metadata3);
+        Baggage baggage = builder.build();
+
+        assertEquals(3, baggage.size());
+
+        String value = baggage.getEntryValue("myKey2");
+        assertEquals("myValue2", value);
+    }
```
Azquelt: Could also test retrieving the metadata here (`baggage.asMap().get().getMetadata()`)

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/api/CommonSDKServlet.java
```diff
@@ -42,13 +46,23 @@ public class CommonSDKServlet extends FATServlet {
...
+        long clockDurationNano = endNano - startNano;
+        Duration clockDuration = Duration.ofNanos(clockDurationNano);
+
+        //the duration according to the clock should be at least as long as the time we slept for (we might have slept longer)
+        assertTrue(clockDuration.compareTo(sleepDuration) >= 0);
+
+        //make sure that the clockDuration isn't vastly bigger than the intended sleep duration
+        Duration maxSleepDuration = sleepDuration.plus(Duration.ofSeconds(5));
+        assertTrue(clockDuration.compareTo(maxSleepDuration) < 0);
     }
```
Azquelt, line 66: This is great, the clock tests look much more robust to me.

As an FYI, you can use the matchers `greaterThan()`, `lessThan()`, `between()` etc. to compare `Comparables` which should automatically get you reasonable test failure messages.

Azquelt (approved):
Looks great, thank you

tevans78:
#build

tevans78:
#libby

fritze2:
Hi, I'm requesting a #pbbeta to compare results against the above build and make sure we get the same outcome.

PR #23936, opened 2023-01-17
Fix EE10 repeat rules for SAML

c00crane:
No tests were being run on Java 8 for many of the SAML Fats - the EE10 (and EE9) runs were the only repeats being run and they were skipped because they should only run on Java 11 and above.


c00crane:
!build

c00crane:
#build

c00crane:
#libby

teddyjtorres (approved):

Issue #23481, opened 2022-11-23
Support the OpenTelemetry logging exporter


Azquelt:
Add support and tests for the [OpenTelemetry logging exporter](https://github.com/open-telemetry/opentelemetry-java/tree/v1.18.0/sdk-extensions/autoconfigure#logging-exporter)

Currently, the SDK Autoconfigure will allow you to configure it, but then complains that it can't find the exporter. We need to check if we're actually including it and if so why it's not available when we try to configure open telemetry.

Yas-Franco:
The package was just missing from the bundle. I still need to add tests for this

PR #23963, opened 2023-01-18
Add Telemetry logging exporter

Yas-Franco:
For issue #23481

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLoggingExporterTest.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLoggingExporterTest.java
@@ -1,0 +1,77 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal_fat;
+
+import static com.ibm.websphere.simplicity.ShrinkHelper.DeployOptions.SERVER_ONLY;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import com.ibm.websphere.simplicity.ShrinkHelper;
+
+import componenttest.annotation.Server;
+import componenttest.annotation.TestServlet;
+import componenttest.annotation.TestServlets;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.LoggingServlet;
+
+@RunWith(FATRunner.class)
+public class TelemetryLoggingExporterTest extends FATServletClient {
+
+    public static final String SERVER_NAME = "Telemetry10ConfigEnv";
+    public static final String APP_NAME = "TelemetryApp";
+
+    @Server(SERVER_NAME)
+    public static LibertyServer server;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        WebArchive app = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                        .addClasses(LoggingServlet.class);
+
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+
+        server.addEnvVar("otel_sdk_disabled", "false");
+
+        //Use logging exporter
+        server.addEnvVar("otel_traces_exporter", "logging");
+        server.startServer();
+    }
+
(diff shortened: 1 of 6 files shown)


Yas-Franco:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLoggingExporterTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    public void testConsole() throws Exception {
+        runTest(server, APP_NAME + "/ConfigServlet", "testServiceNameConfig");
+
+        //The exporter should not give this error message
+        assertTrue(server.findStringsInLogs("Failed to export spans").isEmpty());
+
+        //Checks for span output in logs
+        assertFalse(server.findStringsInLogs("io.opentelemetry.exporter.logging.LoggingSpanExporter").isEmpty());
+        assertFalse(server.findStringsInLogs("'testSpan' : .* \\[tracer: config-test:1.0.0\\]").isEmpty());
```
Azquelt: Unless you can guarantee that these strings must be in the logs at this point, consider using `server.waitForStringsInLogs` instead. I think you can get a slight delay between when a span is created and when it's exported, I think you want to do that here to avoid intermittent defects.

Also, if you need to ensure that a particular operation outputs a certain string, you can use `server.setMarkToEndOfLog` first, then perform the operation under test, and then call `server.waitForStringInLogUsingMark` afterwards.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLoggingExporterTest.java
```diff
@@ -0,0 +1,80 @@
...
+public class TelemetryLoggingExporterTest extends FATServletClient {
+
+    public static final String SERVER_NAME = "Telemetry10ConfigEnv";
+    public static final String APP_NAME = "TelemetryApp";
+
+    @Server(SERVER_NAME)
+    @TestServlets({
+                    @TestServlet(servlet = ConfigServlet.class, contextRoot = APP_NAME),
+    })
+    public static LibertyServer server;
```
Azquelt: It's odd to reuse `ConfigServlet` here, that's not really testing the same thing. It's also odd that we're listing it here as a test servlet, and then also have `testConsole()` which is manually calling one of its methods.

I think you'd be better creating another test servlet just for this test.

If you don't want the methods in the test servlet to be called automatically, you can not annotate them with `@Test` and you don't have to list them here using `@TestServlet`. You can still use `FATServletClient.runTest` to call the method and get any result back.

If you don't reuse `ConfigServlet`, you also don't need to set the name to `overrideDone`.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLoggingExporterTest.java
```diff
@@ -0,0 +1,80 @@
...
+import componenttest.annotation.TestServlet;
+import componenttest.annotation.TestServlets;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.telemetry.ConfigServlet;
+
+//@Mode(TestMode.FULL)
+@RunWith(FATRunner.class)
+public class TelemetryLoggingExporterTest extends FATServletClient {
```
Azquelt: Should this be full mode or not?

If not, please remove the commented out annotation.

Azquelt (changes requested):
This looks good, I think the approach of having a test method which calls `runTest` and then verifies that the expected strings are in the logs is a good one.

I think you should create a new test servlet though, as you're not testing the same things as you were in ConfigServlet and there's a small tweak needed to the way you test for the log messages.

Yas-Franco:
#libby
#build

yasmin-aumeeruddy:
#build
#libby

Yas-Franco:
#libby
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLoggingExporterTest.java
```diff
@@ -0,0 +1,77 @@
...
+        server.setMarkToEndOfLog();
+
+        runTest(server, APP_NAME + "/LoggingServlet", "testLoggingExporter");
+        
+        //The exporter should not give this error message
+        assertNull(server.verifyStringNotInLogUsingMark("Failed to export spans",1000));
+
+        //Checks for span output in logs
+        assertFalse(server.waitForStringInLogUsingMark("io.opentelemetry.exporter.logging.LoggingSpanExporter").isEmpty());
+        assertFalse(server.waitForStringInLogUsingMark("'testSpan' : .* \\[tracer: logging-exporter-test:1.0.0\\]").isEmpty());
```
Azquelt, line 70: I'm a little dubious that 1000ms is long enough for the "Failed to export spans" error to be emitted, and it's also a fairly long time to wait for a test that should complete instantaneously.

However, the positive tests in this case are probably sufficient.

Azquelt (approved):

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLoggingExporterTest.java
Azquelt: We could potentially be more strict on checking for the log lines. It's plausible that both of those checks could be matched by some random debugging lines, though I think that's unlikely.

PR #23980, opened 2023-01-19
Add @SkipIfCheckpointNotSupported to the test

anjumfatima90:


tjwatson (approved):

tjwatson:
#build

tjwatson (approved):
Approved to add the annotations.  We should merge this as soon as we can.  Could you open a separate PR that does the following:

In `componenttest.topology.impl.LibertyServer.doCheckpoint()` could you update the method to not have `&& getCheckpointSupported()`:

```java
    private boolean doCheckpoint() {
        return (checkpointInfo != null);
    }
```
This would have caused the tests to fail if run on non criu JVMs.

tjwatson:
#libby

anjumfatima90:
> Approved to add the annotations. We should merge this as soon as we can. Could you open a separate PR that does the following:
> 
> In `componenttest.topology.impl.LibertyServer.doCheckpoint()` could you update the method to not have `&& getCheckpointSupported()`:
> 
> ```java
>     private boolean doCheckpoint() {
>         return (checkpointInfo != null);
>     }
> ```
> 
> This would have caused the tests to fail if run on non criu JVMs.

Added this new PR https://github.com/OpenLiberty/open-liberty/pull/24032

PR #23986, opened 2023-01-19
Create test for exporting spans via TLS

### LINKED ISSUE #23455: Telemetry exporter security
For OTLP, SDK Autoconfiguration includes properties to [configure transport security][configure] for the connection to the trace server. We should test that this works as expected.

[configure]: https://github.com/open-telemetry/opentelemetry-java/tree/v1.18.0/sdk-extensions/autoconfigure#otlp-exporter-span-metric-and-log-exporters

benjamin-confino:
Fixes #23455


benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.OperatorCreationException;
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
+import org.bouncycastle.openssl.jcajce.JcaPEMWriter;
+
+import com.ibm.websphere.simplicity.log.Log;
+
+public class SSLUtils {
```
Azquelt, line 82: We need Javadoc to explain how to use this class.

The class should have a javadoc comment saying what its responsibilities are and each method should have a comment saying what it does and explaining its parameters.

We should also have an example of how to use this class. This could either go in the class javadoc if it involves several method calls, or in the method javadoc if that makes more sense.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+        return pair;
+    }
+
+    public static Certificate selfSign(KeyPair keyPair, String subjectDN, List<String> genericNameList)
+            throws Exception
+    {
+        Log.info(c, "selfSign", "entering SSLUtils.selfSign");
+
+        final String DNS_PATTERN = "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$";
+        final String IPV4_PATTERN = "^(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])(\\.(?!$)|$)){4}$";
```
Azquelt: Did you write these yourself or did you copy it from somewhere?

I would go with something simpler. Use `\d{1-3}(\.\d{1-3}){3}` to match an IPv4 address and assume everything else is a host name.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(startDate);
+        // 1 Yr validity
+        calendar.add(Calendar.YEAR, 1);
+
+        Date endDate = calendar.getTime();
+
+        // Use appropriate signature algorithm based on your keyPair algorithm.
+        String signatureAlgorithm = "SHA256WithRSA";
```
Azquelt: Not sure how this comment matches the code.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+            if (name.matches(DNS_PATTERN)) {
+                Log.info(c, "selfSign", "adding dns GeneralName " + name);
+                altNames.add(new GeneralName(GeneralName.dNSName, name));
+            } else if (name.matches(IPV4_PATTERN)) {
+                Log.info(c, "selfSign", "adding ip GeneralName " + name);
+                altNames.add(new GeneralName(GeneralName.dNSName, name));
+            }
+        }
+
+        GeneralNames subjectAltNames = GeneralNames.getInstance(new DERSequence((GeneralName[]) altNames.toArray(new GeneralName[] {})));
```
Azquelt: This line looks weird, does it need the cast and the `DERSequence` in there?

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+        Log.info(c, "exportPublicKeyToFile", "entering SSLUtils.exportPublicKeyToFile");
+
+        PublicKey publicKey = keyPair.getPublic();
+        File file = new File(path);
+        file.getParentFile().mkdirs();
+        file.createNewFile();
+
+        FileWriter fw = new FileWriter(file, false);
+        JcaPEMWriter pemWriter = new JcaPEMWriter(fw);
+        pemWriter.writeObject(publicKey);
```
Azquelt: You're not closing the stream.

You _must_ always ensure that any kind of IO stream is closed. You should use try-with-resources where possible.

```suggestion
        try (FileWriter fw = new FileWriter(file, false);
             JcaPEMWriter pemWriter = new JcaPEMWriter(fw)) {
            pemWriter.writeObject(publicKey);
        }
```

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+        PrivateKey privateKey = keyPair.getPrivate();
+        File file = new File(path);
+        file.getParentFile().mkdirs();
+        file.createNewFile();
+
+        FileWriter fw = new FileWriter(file, false);
+        JcaPEMWriter pemWriter = new JcaPEMWriter(fw);
+        pemWriter.writeObject(privateKey);
+
+        pemWriter.close();
```
Azquelt: As above, use a try-with-resources block to ensure that the file stream is always closed.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+
+        File file = new File(path);
+        file.getParentFile().mkdirs();
+        file.createNewFile();
+
+        FileWriter fw = new FileWriter(file, false);
+        JcaPEMWriter pemWriter = new JcaPEMWriter(fw);
+        pemWriter.writeObject(cert);
+
+        pemWriter.close();
```
Azquelt: As above, use a try-with-resources block.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+
+        X509CertificateHolder certificateHolder = certificateBuilder.build(contentSigner);
+
+        Certificate selfSignedCert = new JcaX509CertificateConverter()
+                .getCertificate(certificateHolder);
+
+        return selfSignedCert;
+    }
+
+    public static void exportPublicKeyToFile(String path, KeyPair keyPair) throws Exception {
```
Azquelt: In general, when writing utility methods for working with files, we should pass in the `File` or `Path` object, rather than passing in a string.

Any requirements on the file passed in (like it must not exist, or its parent directory must exist) should be documented. If you're doing anything else in your method (like automatically creating parent directories), that must be documented.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/JaegerSecureOtlpTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private static String getKey() {
+        generateSSLStuff();
+        return key;
+    }
+
+    private static String getCertificate() {
+        generateSSLStuff();
+        return cert;
+    }
```
Azquelt: These need documentation about what the `String` that's returned means.

Possibly it should be a `File` instead.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/JaegerSecureOtlpTest.java
```diff
@@ -0,0 +1,135 @@
...
+        server.addEnvVar(TestConstants.ENV_OTEL_BSP_SCHEDULE_DELAY, "100"); // Wait no more than 100ms to send traces to the server
+        server.addEnvVar(TestConstants.ENV_OTEL_SDK_DISABLED, "false"); //Enable tracing
+
+        //enable TLS
+        //server.addEnvVar(TestConstants.ENV_COLLECTOR_OTLP_GRPC_TLS_ENABLED, "true"); //Don't think is required
+
+        server.addEnvVar(TestConstants.ENV_OTEL_EXPORTER_OTLP_CERTIFICATE, cert);
+
+        //server.addEnvVar(TestConstants.ENV_OTEL_EXPORTER_OTLP_CLIENT_KEY, key.getAbsolutePath());
+        //server.addEnvVar(TestConstants.ENV_OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE, CACert.getAbsolutePath());
```
Azquelt: Remove the commented out code.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/JaegerSecureOtlpTest.java
```diff
@@ -0,0 +1,135 @@
...
+    private synchronized static void generateSSLStuff() {
+        if (createdSSLStuff) {
+            return;
+        }
+
+        try {
+            KeyPair generatedKeyPair = SSLUtils.generateKeyPair();
+
+            String dockerIP = DockerClientFactory.instance().dockerHostIpAddress();
+            String dnName = "O=Evil Inc, CN="+dockerIP+", L=Toronto,C=CA";
```
Azquelt: In case the certificate escapes and ends up somewhere odd, the dnName should probably indicate that it's a test certificate.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/utils/jaeger/JaegerContainer.java
```diff
@@ -56,6 +67,34 @@ public JaegerContainer(DockerImageName imageName) {
...
+    public JaegerContainer(String tlsCert, String tlsKey) {
+        super(new ImageFromDockerfile().withDockerfileFromBuilder(builder -> builder.from(
+                                                                        ImageNameSubstitutor.instance()
+                                                                        .apply(IMAGE_NAME)
+                                                                        .asCanonicalNameString())
+                                                                        .copy("/etc/certificate.crt", "/etc/certificate.crt")
+                                                                        .copy("/etc/private.key", "/etc/private.key")
+                                                                        .build())
+                                                                        .withFileFromFile("/etc/certificate.crt", new File(tlsCert),0644)
+                                                                        .withFileFromFile("/etc/private.key", new File(tlsKey), 0644));
```
Azquelt: Pass in the cert and key as `File`s

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/utils/otelCollector/OtelCollectorContainer.java
```diff
@@ -74,4 +74,15 @@ public int getOtlpGrpcPort() {
...
+        /**
+     * Get the URL to use to send OLTP spans via gRPC over tls
+     * <p>
+     * Only valid when the container is started
+     *
+     * @return the OLTP gRPC URL
+     */
+    public String getSecureOtlpGrpcUrl() {
+        return "https://" + getHost() + ":" + getOtlpGrpcPort();
+    }
```
Azquelt: Is this bit used anywhere?

Azquelt (changes requested):
Well done for getting this working, it is quite complicated but what you've written is good and clear.

I have added comments with a few questions about where the certificate creation code came from and a few code commenting and tidying issues.

There are also a few formatting errors, please format the code from Eclipse.

Azquelt:
Please rebase this to remove the merge commit.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+
+import com.ibm.websphere.simplicity.log.Log;
+
+public class SSLUtils {
+
+    private static final Class<?> c = SSLUtils.class;
+
+    public static KeyPair generateKeyPair() throws Exception {
+        KeyPairGenerator generator = KeyPairGenerator.getInstance("RSA");
+        generator.initialize(2048, new SecureRandom());
```
Azquelt: I _think_ that `SecureRandom` can be slow to initialize, it might be preferable to put it in a static field and synchronize the methods which use it.

dev/fattest.simplicity/bnd.bnd
```diff
@@ -60,6 +61,10 @@ Private-Package: \
 	com.ibm.websphere.simplicity.configuration
 
 Include-Resource: \
+    @${repo;org.bouncycastle:bcpg-jdk15on;1.69;EXACT},\
+    @${repo;org.bouncycastle:bcpkix-jdk15on;1.69;EXACT},\
+    @${repo;org.bouncycastle:bcprov-jdk15on;1.69;EXACT},\
+    @${repo;org.bouncycastle:bcutil-jdk15on;1.69;EXACT},\
```
tevans78, line 67: Are you sure that this is the best version to use?
If you look here - https://www.bouncycastle.org/latest_releases.html - you will find that there is a 1.73 release and that instead of using `jdk15on`, it seems to suggest using `jdk18on`.
> Java Version Details With the arrival of Java 15. jdk15 is not quite as unambiguous as it was. The jdk18on jars are compiled to work with anything from Java 1.8 up. They are also multi-release jars so do support some features that were introduced in Java 9, Java 11, and Java 15. If you have issues with multi-release jars see the jdk15to18 release jars below.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
```diff
@@ -0,0 +1,175 @@
...
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
```
tevans78: This is no longer needed

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/JaegerSecureOtlpTest.java
```diff
@@ -0,0 +1,135 @@
+/*******************************************************************************
+ * Copyright (c) 2022 IBM Corporation and others.
```
tevans78: Is this the correct date?

dev/fattest.simplicity/bnd.bnd
Azquelt: This is the version that other components in liberty already use. It looks like `acmeCA-2.0` and install code both use it.

Given that it's used in the product and we're only using it for tests, it seems sensible to just use whatever version is currently in use and leave it to the product components to update the version.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
benjamin-confino: With a pemWriter.close() (required to make it flush) this will actually cause an error. PemWriter extends BufferedWriter and BufferedWriter.close() calls Writer.close() so we'd get an IOException from closing BufferedWriter twice. 

But using try-with-resources for the PemWriter would be a good idea.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
benjamin-confino: Copied of course.

I don't like the idea of `\d{1-3}(\.\d{1-3}){3}` its far more readable but its going to accept invalid addresses like 999.999.999.999

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
benjamin-confino: I would have to research to find out, the javadoc is terrible for bouncycastle. getInstance takes java.lang.Object and the javadoc says nothing about what it really wants.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/utils/otelCollector/OtelCollectorContainer.java
benjamin-confino: I don't think its used yet, but it will be if we write a similar test for Grpc.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/JaegerSecureOtlpTest.java
benjamin-confino: Evil Inc not enough? 

I'll change it to Evil Inc Test Certificate.

dev/fattest.simplicity/src/componenttest/security/utils/SSLUtils.java
Azquelt: Calling `Closable.close()` has no effect if the stream is already closed so it shouldn't cause an error.

Only closing the `JcaPEMWriter` should be fine though as long as the doc says that it will close the underlying writer.

benjamin-confino:
#build

benjamin-confino:
previous build hasn't started yet

benjamin-confino:
Execution failed for task ':com.ibm.ws.security.utility:test'.
> Timeout has been exceeded

benjamin-confino:
#build

benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/JaegerSecureOtlpTest.java
```diff
@@ -0,0 +1,133 @@
...
+        if (createdSSLStuff) {
+            return;
+        }
+
+        try {
+            KeyPair generatedKeyPair = SSLUtils.generateKeyPair();
+
+            String dockerIP = DockerClientFactory.instance().dockerHostIpAddress();
+            String dnName = "O=Evil Inc Test Certificate, CN="+dockerIP+", L=Toronto,C=CA";
+            List genericNameList = new ArrayList<String>();
```
Azquelt:
```suggestion
            List<String> genericNameList = new ArrayList<>();
```

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/JaegerSecureOtelCollectorTest.java
```diff
@@ -0,0 +1,147 @@
...
+            privateKeyFile = new File(pathToPrivateKey);
+            SSLUtils.exportPrivateKeyToFile(privateKeyFile, generatedKeyPair);
+
+            String pathToCertificate = server.getServerSharedPath() + "/certificate.crt";
+            certificateFile = new File(pathToCertificate);
+            SSLUtils.exportCertificateToFile(certificateFile, certificateObject);
+            createdSSLStuff = true;
+        } catch (Exception e) { //If we get an exception let the test fail and show the developer what went wrong
+            throw new RuntimeException("Exception doing SSLStuff", e);
+        }
```
Azquelt: This may break when we come to add repeats because the files are generated within the server directory which will get deleted when the server is cleaned up.

I don't think we need to change this now though.

benjamin-confino:
#build

Azquelt:
#libby

Azquelt (approved):

PR #24027, opened 2023-01-23
Add jpa java binary level spanning support fat

jgrassel:
Adding the com.ibm.ws.jpa.tests.javasupport_fat test bucket.  This tests all of the JPA features against builds of the same application from across java build level targets supported by Liberty.  (Note, openjpa only supports java 7 level binaries, all of eclipselink is tested against java 7-20 level binaries at the time of this commit).

Of course, the maximum java level that is tested in a given run is the JDK level running the server and test bucket.


jgrassel:
#build

dazey3 (approved):
lgtm

PR #24036, opened 2023-01-24
Clarify process for externally requested features, remove a11y and g11n approvals, add InstantOn approval

mbroz2:


.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -65,39 +68,39 @@ The ([OpenLiberty/chief-architect](https://github.com/orgs/OpenLiberty/teams/chi
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 ## **Design** (Complete Before Development Starts)
 
-Design preliminaries determine whether a formal design, which will be provided by an [Upcoming Feature Overview (UFO)](https://ibm.box.com/v/UFO-Template) document, must be created and reviewed.  A formal design is required if the feature requires any of the following: UI, Serviceability, SVT, Performance testing, or non-trivial documentation/ID.
+Design preliminaries determine whether a formal design, which will be provided by an [Upcoming Feature Overview (UFO)](https://ibm.box.com/v/UFO-Template) document, must be created and reviewed.  A formal design is required if the feature requires any of the following: UI, Serviceability, SVT, Performance testing, or non-trivial documentation/ID.  Furthermore, each identified item places a blocking requirment on another team so it must be identified early in the process.  The Feature owner may check-off the item if they know it doeosn't apply, but otherwise they should work with the focal point to determine what work, if any, will be necessary and make them aware of it.
```
mbroz2:
```suggestion
Design preliminaries determine whether a formal design, which will be provided by an [Upcoming Feature Overview (UFO)](https://ibm.box.com/v/UFO-Template) document, must be created and reviewed.  A formal design is required if the feature requires any of the following: UI, Serviceability, SVT, Performance testing, or non-trivial documentation/ID.  Furthermore, each identified item places a blocking requirement on another team so it must be identified early in the process.  The feature owner may check-off the item if they know it doesn't apply, but otherwise they should work with the focal point to determine what work, if any, will be necessary and make them aware of it.
```

cbridgha (approved):
Looks good

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -122,28 +126,27 @@ In order to avoid last minute blockers and significant disruptions to the featur
...
-- [ ] Licenses and Certificates of Originality (COOs) are updated, or N/A
-
-### **Translation** (Complete 1 week before Feature Complete Date)
-- [ ] PII updates are merged, or N/A. Note timing with translation shipments.
 
 ### **Innovation** (Complete 1 week before Feature Complete Date)
 - [ ] Consider whether any aspects of the feature may be patentable. If any identified, disclosures have been submitted.
 
+### **Translation** (Complete by Feature Complete Date)
+- [ ] PII (Program Integrated Information) updates are merged (i.e. all English strings due for translation have been delivered), or N/A.
```
tevans78, line 134: technically the changes need to be in a release build by the Feature Complete Date

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -169,36 +178,34 @@ A feature is ready to GA after it is Feature Complete and has obtained all neces
...
 
 ### **Design Approved Features**
-- [ ] **Accessibility** Accessibility testing completed or N/A. ([OpenLiberty/accessibility-approvers](https://github.com/orgs/OpenLiberty/teams/accessibility-approvers))
-  - Approver adds label `focalApproved:accessibility`.
-- [ ] **ID** Documentation is complete or N/A. ([OpenLiberty/id-approvers](https://github.com/orgs/OpenLiberty/teams/id-approvers))
+- [ ] **ID** - Documentation is complete or N/A. ([OpenLiberty/id-approvers](https://github.com/orgs/OpenLiberty/teams/id-approvers))
   - Approver adds label `focalApproved:id`.
   - > **_NOTE:_**  If only trivial documentation changes are required, you may reach out to the ID Feature Focal to request a `ID Required - Trivial` label.  Unlike features with regular ID requirement, those with `ID Required - Trivial` label do not have a hard requirement for a Design/UFO.
-- [ ] **Performance** Performance testing is complete or N/A. ([OpenLiberty/performance-approvers](https://github.com/orgs/OpenLiberty/teams/performance-approvers))
+- [ ] **InstantOn** - InstantOn capable or N/A. ([OpenLiberty/instantOn-approvers](https://github.com/orgs/OpenLiberty/teams/instantOn-approvers))
```
tevans78, line 193: All of these approvals say `or N/A` on them but is it clear that it is the approver who decides that it isn't applicable ... or rather, is it clear that Feature Owners should not make this decision themselves??

.github/ISSUE_TEMPLATE/feature.md
mbroz2: That's only if we needed the content to come back translated prior to the feature going GA.  Since that's no longer the case (for `Translation - Missing`), then as long as their strings are delivered (just like the rest of their code), they're ok.
Also, I'm hopeful that in the not-to-distant future, this process will be completely automated and done on each release build, rather then a select set of builds only.  

But @pnickoll in case he thinks the above needs to be changed

.github/ISSUE_TEMPLATE/feature.md
mbroz2: This has been this way for several years and I'm not aware of it leading to confusion in the past.  The top of the feature template (under General Instructions) states:

> Each process step has a number of tasks which must be completed or must be marked as not applicable ("N/A").
> Unless otherwise indicated, the tasks are the responsibility of the Feature Owner or a Delegate of the Feature Owner.

So in this case, it specifies that the focal approvers are responsible for these checkmarks.  Even if the feature owner (or whoever) were to check that box instead of the focal, as long as they didn't add the label, the process would still work because it relies on labels (whereas the checkboxes are there for the developers)

If necessary, I could make the 'Approver' in the bullet below another link to the responsible party.

PR #24042, opened 2023-01-24
Don't use redirect-query-check with the static resource list

WhiteCat22:
Backporting [this fix](https://github.com/apache/cxf/pull/1034) from CXF


WhiteCat22:
#Libby #build

jim-krueger (approved):

jim-krueger (commented):
Looks good

PR #24063, opened 2023-01-25
Added testcase to test scheduled task execution with checkpoint and r…

KeenanMcCarthy:
…estore

Set scheduled task at checkpoint, and tests behavior upon restore

### DIFF
--- a/dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/ConcurrencyTest.java
+++ b/dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/ConcurrencyTest.java
@@ -67,9 +67,14 @@ public class ConcurrencyTest {
         return Math.abs(timestamp2.getTime() - timestamp1.getTime());
     }
 
+    /*
+     * Tests scheduling a task at checkpoint with a seven second delay through the ManagedExecutor,
+     * then waiting ten seconds between checkpoint and restore and checking the timestamp to ensure
+     * that there is a seven second delay upon restore (inculding a two second margin of error)
+     */
     @Test
-    public void testSchduledTaskTenSec() throws Exception {
-        Integer scheduledDelay = 10000;
+    public void testSchduledTask() throws Exception {
+        Integer scheduledDelay = 7000;
 
         ServerConfiguration config = server.getServerConfiguration();
         config.getVariables().getById("scheduledTime").setValue(scheduledDelay.toString());
@@ -78,7 +83,7 @@ public class ConcurrencyTest {
         server.updateServerConfiguration(config);
 
         server.setCheckpoint(CheckpointPhase.APPLICATIONS, false, (server) -> {
-            String result = server.waitForStringInLog("Thread scheduled at startup", 20000);
+            String result = server.waitForStringInLog("Thread scheduled at startup");
             assertNotNull("Scheduled executor not called at checkpoint", result);
         });
         server.startServer();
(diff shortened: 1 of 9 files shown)


dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/ConcurrencyTest.java
```diff
@@ -34,6 +34,7 @@
 import componenttest.custom.junit.runner.FATRunner;
 import componenttest.topology.impl.LibertyServer;
 import concurrentApp.ConcurrentApp;
+import concurrentApp.RepeatedTrigger;
 import io.openliberty.checkpoint.spi.CheckpointPhase;
 
 @RunWith(FATRunner.class)
```
sebratton, line 40: Need to annotate this class with @SkipIfCheckpointNotSupported

KeenanMcCarthy: This annotation is already included

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/ConcurrencyTest.java
```diff
@@ -0,0 +1,223 @@
...
+        Integer scheduledDelay = 10000;
+
+        ServerConfiguration config = server.getServerConfiguration();
+        config.getVariables().getById("scheduledTime").setValue(scheduledDelay.toString());
+        config.getVariables().getById("repeatTrigger").setValue("false");
+        config.getVariables().getById("repeatManagedExec").setValue("false");
+        server.updateServerConfiguration(config);
+
+        server.setCheckpoint(CheckpointPhase.APPLICATIONS, false, (server) -> {
+            String result = server.waitForStringInLog("Thread scheduled at startup", 20000);
```
sebratton: The waitforStringInLog is called after the server is halted (due to the checkpoint) so the 20 sec wait is not necessary, athough it does not affect the correctness of the test. It just introduces and unnecessary 20 sec wait when the check will fail.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/ConcurrencyTest.java
```diff
@@ -0,0 +1,223 @@
...
+            assertNotNull("Scheduled executor not called at checkpoint", result);
+        });
+        server.startServer();
+        Thread.sleep(10000);
```
sebratton, line 90: should this ne `scheduleDelay` rather than hardcoded?

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/ConcurrencyTest.java
```diff
@@ -0,0 +1,225 @@
...
+        String scheduledTaskLog = server.waitForStringInLog("The Liberty server process resumed operation from a checkpoint", 15000);
+        String taskExecutionLog = server.waitForStringInLog("Scheduled thread completed", 20000);
+        long timeElapsed = getTimeElapsed(scheduledTaskLog, taskExecutionLog);
+        assertTrue("Expected and Observed delays differ. Expected: " + scheduledDelay +
+                   " ms | Observed: " + timeElapsed + " ms",
+                   Math.abs(timeElapsed - scheduledDelay) < MARGIN_OF_ERROR);
+    }
+
+    @Test
+    public void testRepeatedMangedExecTwentySec() throws Exception {
```
sebratton: I've not seeing the point of having both this test and `testRepeatedTriggerTwentySec` they have the same preconditions. The only diff I see is that you are checking for different strings in the pre-restore lambda where both strings are expected. If that was the intent, then this could be combined into one test.

dev/io.openliberty.checkpoint_fat/fat/src/io/openliberty/checkpoint/fat/ConcurrencyTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        server.setCheckpoint(CheckpointPhase.APPLICATIONS, false, (server) -> {
+            String result = server.waitForStringInLog("Thread scheduled at startup", 20000);
+            assertNotNull("Repeated task not called at checkpoint", result);
+        });
+        server.startServer();
+        Thread.sleep(10000);
+        server.checkpointRestore();
+        String scheduledTaskLog = server.waitForStringInLog("The Liberty server process resumed operation from a checkpoint", 15000);
+        String taskExecutionLog = server.waitForStringInLog("Scheduled thread completed", 20000);
```
sebratton, line 156: You may want to wait >20 secs on test where the task timer is 20 seconds to avoid false failures.

KeenanMcCarthy:
#build

KeenanMcCarthy:
#libby
#build

sebratton (approved):

PR #24064, opened 2023-01-26
Create Async Context propagation test

### LINKED ISSUE #23868: Check that the Context API can be used between threads
Check that the Context API can be used to transfer context between threads when submitting tasks to a ManagedExecutor, and child spans can be created within the async task

    Note: we don't have to transfer this context automatically, but the user should be able to transfer it manually using the Context API

tevans78:
fixes https://github.com/OpenLiberty/open-liberty/issues/23868

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/context/MultiThreadedContextServlet.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/context/MultiThreadedContextServlet.java
@@ -1,0 +1,181 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal_fat.apps.context;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.junit.Test;
+
+import componenttest.app.FATServlet;
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.baggage.BaggageBuilder;
+import io.opentelemetry.context.Context;
+import jakarta.annotation.Resource;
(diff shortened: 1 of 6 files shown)


tevans78:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/context/MultiThreadedContextServlet.java
```diff
@@ -0,0 +1,110 @@
...
+
+            //create a new baggage and make a new current context with it
+            //this new context should not affect the original one
+            addBaggage("myKey1", "myValue2");
+
+            //check that the new baggage is in the current context
+            assertBaggage("myKey1", "myValue2");
+
+            return value;
+        }
```
Azquelt: Given that you now have an assert in this method anyway, it might make sense to assert `value` here as well, rather than returning it and asserting in the original method.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/context/MultiThreadedContextServlet.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Test
+    public void testContextWrap() throws InterruptedException, ExecutionException, TimeoutException {
+        //add some baggage to the current context
+        addBaggage("myKey1", "myValue1");
+
+        // execute my async sub-task
+        Context context = Context.current();
+        ExecutorService executorService = context.wrap(managedExecutorService);
```
Azquelt, line 70: It would probably be better here to test without making `context` the current context.

Azquelt:
The tests we have are good, could add one more to test `Context.wrap(Callable<T>);`

tevans78:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/context/MultiThreadedContextServlet.java
```diff
@@ -0,0 +1,160 @@
...
+        // execute my async sub-task
+        ExecutorService executorService = context.wrap(managedExecutorService);
+        Future<String> future = executorService.submit(new GetBaggageValueCallable(MY_KEY_1));
+        String baggageValue = future.get(5, TimeUnit.SECONDS);
+        //check that the value for MY_KEY_1 was null
+        assertEquals(null, baggageValue);
+
+        //check that the baggage of the current context was not modified by the sub-task
+        assertBaggage(MY_KEY_1, MY_VALUE_1);
+    }
```
Azquelt: This test (and the following one) doesn't actually do anything that checks that the specified context is propagated. The asserted behaviour would occur if no context was propagated (though it does check that the current context is _not_ propagated).

Azquelt (dismissed):

tevans78:
#build

Azquelt (dismissed):

Yas-Franco (approved):
Looks good

tevans78:
#libby

PR #24083, opened 2023-01-26
Basic scenarios with EmbeddedId and intermixed name patterns

njr-11:
Allow embeddables as entity properties, including as the id (EmbeddedId). Test some basic scenarios. More scenarios will be needed to explore further.  Also, try out referring to embeddables with names that intermix the patterns.


njr-11:
#build

nmittles (approved):

PR #24100, opened 2023-01-30
Use https to access dhe

tevans78:
`http` access will shortly be switched off. Using `https` instead.

#build


cbridgha (approved):
looks good

Issue #23176, opened 2022-10-28
Test Telemetry with longer running applications.


Azquelt:
While looking at security exceptions in the TCK logs, I noticed that there were some actions which seem to run on a timer and were causing errors at regular intervals.

These weren't seen with very short-lived test apps so we should make sure we have a full mode test which runs for a little longer.

Azquelt:
I suspect what may be happening is that the configured collector is not available at app startup and the OTel client retries the connection to it in a background thread.

We should ensure this scenario is tested as I suspect this could cause J2S failures which aren't seen in the golden scenario where the exporter connects at startup and remains connected.

PR #24119, opened 2023-01-31
Add telemetry test for long running app

Yas-Franco:
For issue #23176

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLongRunningTest.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLongRunningTest.java
@@ -68,11 +68,8 @@ public class TelemetryLongRunningTest extends FATServletClient {
         long timeoutNanos = DEFAULT_TIMEOUT.toNanos();
         long startNanoTime = System.nanoTime();
         while (System.nanoTime() - startNanoTime < timeoutNanos) {
-            try {
-                Thread.sleep(1000);
-                runTest(server, APP_NAME + "/longrunning", "testLongRunningTask");
-            } catch (InterruptedException e) {
-            }
+            Thread.sleep(1000);
+            runTest(server, APP_NAME + "/longrunning", "testLongRunningTask");
         }
     }
 
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLongRunningTest.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLongRunningTest.java
@@ -1,0 +1,83 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal_fat;
+
+import static com.ibm.websphere.simplicity.ShrinkHelper.DeployOptions.SERVER_ONLY;
+
+import java.time.Duration;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import com.ibm.websphere.simplicity.ShrinkHelper;
+
+import componenttest.annotation.Server;
+import componenttest.custom.junit.runner.FATRunner;
+import componenttest.custom.junit.runner.Mode;
+import componenttest.custom.junit.runner.Mode.TestMode;
+import componenttest.topology.impl.LibertyServer;
+import componenttest.topology.utils.FATServletClient;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.longrunning.LongRunningTask;
+import io.openliberty.microprofile.telemetry.internal_fat.common.spanexporter.InMemorySpanExporter;
+import io.openliberty.microprofile.telemetry.internal_fat.common.spanexporter.InMemorySpanExporterProvider;
+import io.opentelemetry.sdk.autoconfigure.spi.traces.ConfigurableSpanExporterProvider;
+
+@Mode(TestMode.FULL)
+@RunWith(FATRunner.class)
+public class TelemetryLongRunningTest extends FATServletClient {
+
+    public static final String SERVER_NAME = "Telemetry10LongRunning";
+    public static final String APP_NAME = "LongRunningApp";
+    public static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(120); // 120 seconds
+
+    @Server(SERVER_NAME)
+    public static LibertyServer server;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        WebArchive app = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                        .addPackage(LongRunningTask.class.getPackage())
+                        .addPackage(InMemorySpanExporter.class.getPackage())
+                        .addAsServiceProvider(ConfigurableSpanExporterProvider.class, InMemorySpanExporterProvider.class)
+                        .addAsResource(new StringAsset("otel.sdk.disabled=false\notel.traces.exporter=in-memory"),
+                                       "META-INF/microprofile-config.properties");
+
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
+        server.startServer();
+    }
+
+    @Test
+    public void testLongLastingTask() throws Exception {
+        server.setMarkToEndOfLog();
+
+        long timeoutNanos = DEFAULT_TIMEOUT.toNanos();
+        long startNanoTime = System.nanoTime();
+        while (System.nanoTime() - startNanoTime < timeoutNanos) {
+            try {
+                Thread.sleep(1000);
+                runTest(server, APP_NAME + "/longrunning", "testLongRunningTask");
+            } catch (InterruptedException e) {
+            }
+        }
+    }
+
+    @AfterClass
+    public static void tearDown() throws Exception {
+        server.stopServer();
+    }
+}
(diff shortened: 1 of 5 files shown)


dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLongRunningTest.java
```diff
@@ -41,11 +41,9 @@ public class TelemetryLongRunningTest extends FATServletClient {
 
     public static final String SERVER_NAME = "Telemetry10LongRunning";
     public static final String APP_NAME = "LongRunningApp";
+    public static final int DEFAULT_TIMEOUT = 120 * 1000; // 120 seconds
```
Azquelt:
```suggestion
    public static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(120);
```

When storing a duration, usually we should use the `Duration` class, rather than an `int` or a `long` to make it easier to translate between units.

If we do want to store a duration as a number rather than a `Duration` object for some reason, it's important that we include the unit in the variable name so that we don't mix up units later. E.g. `DEFAULT_TIMEOUT_MILLIS`.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLongRunningTest.java
```diff
@@ -54,10 +52,30 @@ public static void setUp() throws Exception {
...
         ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);
         server.startServer();
     }
+
+    @Test
+    public void testConsole() throws Exception {
+        server.setMarkToEndOfLog();
+
+        int timeout = DEFAULT_TIMEOUT;
+        while (timeout > 0) {
```
Azquelt:
```suggestion
        long timeoutNanos = DEFAULT_TIMEOUT.toNanos();
        long startNanoTime = System.nanoTime();
        while (System.nanoTime() - startNanoTime < timeoutNanos) {
```

We want to measure the actual time elapsed, rather than counting how long we've slept for since just counting sleep time doesn't take account of how long the request may have taken (which, if we've made a mistake or there's an infrastructure issue, can be longer than we expect).

To measure durations, use `System.nanoTime()` and use the `Duration` class to do any conversions required.

Note: read the `System.nanoTime()` javadoc carefully, it can wrap around and can be negative but you can avoid all the pitfalls by always subtracting the end time from the start time to get a duration in nanoseconds.

If you prefer, you can do everything with `Duration`s here, but it gets a bit messy:
```java
        long startNanoTime = System.nanoTime();
        while (Duration.ofNanos(System.nanoTime() - startNanoTime).compareTo(DEFAULT_TIMEOUT) < 0) {
```
As well as being longer, having to call `compareTo` directly is harder to read.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/longrunning/LongRunningTask.java
```diff
@@ -0,0 +1,50 @@
...
+        span.end();
+    }
+
+    private String doWork() {
+        Span span = tracer.spanBuilder("doingWork").startSpan();
+        assertThat(span.getSpanContext().getSpanId(), not(equalTo(INVALID_SPAN_ID)));
+        span.end();
+        return SUCCESS_MESSAGE;
+    }
+
```
Azquelt: While not really important to what we're testing here, it looks like you mean to make `span` the current span in `testLongRunningTask()` so that it's the parent of the span created in `doWork()`?

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/longrunning/LongRunningTask.java
```diff
@@ -0,0 +1,50 @@
...
+    public static final String SUCCESS_MESSAGE = "COMPLETED SUCCESSFULLY";
+
+    @Test
+    public void testLongRunningTask() {
+        Tracer tracer = openTelemetry.getTracer("logging-exporter-test", "1.0.0");
+        Span span = tracer.spanBuilder("testSpan").startSpan();
+        assertThat(doWork(), equalTo(SUCCESS_MESSAGE));
+        assertThat(span.getSpanContext().getSpanId(), not(equalTo(INVALID_SPAN_ID)));
+        span.end();
```
Azquelt: `span.end()` should be in a `finally` block so that it gets called if one of the `assertThat` methods throws an `AssertionError`

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/longrunning/LongRunningTask.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Inject
+    OpenTelemetry openTelemetry;
+
+    private static final String INVALID_SPAN_ID = "0000000000000000";
+
+    public static final String SUCCESS_MESSAGE = "COMPLETED SUCCESSFULLY";
+
+    @Test
+    public void testLongRunningTask() {
```
Azquelt, line 33: I'm torn on whether we should change the name here as this is no longer a long-running task, but it's still the task that runs within the long-running app.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/longrunning/LongRunningTask.java
```diff
@@ -0,0 +1,50 @@
...
+        Span span = tracer.spanBuilder("testSpan").startSpan();
+        assertThat(doWork(), equalTo(SUCCESS_MESSAGE));
+        assertThat(span.getSpanContext().getSpanId(), not(equalTo(INVALID_SPAN_ID)));
+        span.end();
+    }
+
+    private String doWork() {
+        Span span = tracer.spanBuilder("doingWork").startSpan();
+        assertThat(span.getSpanContext().getSpanId(), not(equalTo(INVALID_SPAN_ID)));
+        span.end();
```
Azquelt: `span.end()` should be in a `finally` block so that it gets called if the `assertThat` method throws an `AssertionError`

Azquelt (changes requested):

Yas-Franco:
#libby
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/longrunning/LongRunningTask.java
```diff
@@ -0,0 +1,56 @@
...
+        Span workSpan = tracer.spanBuilder("workSpan")
+                        .setParent(Context.current().with(parentSpan))
+                        .startSpan();
+        try {
+            assertThat(workSpan.getSpanContext().getSpanId(), not(equalTo(INVALID_SPAN_ID)));
+            workSpan.end();
+            return SUCCESS_MESSAGE;
+        } finally {
+            workSpan.end();
+        }
```
Azquelt: You're calling `workSpan.end()` twice, you only need to do it in the `finally` block.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryLongRunningTest.java
```diff
@@ -0,0 +1,83 @@
...
+        server.setMarkToEndOfLog();
+
+        long timeoutNanos = DEFAULT_TIMEOUT.toNanos();
+        long startNanoTime = System.nanoTime();
+        while (System.nanoTime() - startNanoTime < timeoutNanos) {
+            try {
+                Thread.sleep(1000);
+                runTest(server, APP_NAME + "/longrunning", "testLongRunningTask");
+            } catch (InterruptedException e) {
+            }
```
Azquelt: Don't catch `InterruptedException` - unless you're deliberately interrupting your own task, you can assume this means that the application or server is trying to shut down (e.g. the user running the test pressed Ctrl-C).

In this case, we should fail the test, rather than attempt to keep running it.

Sometimes we'll catch `InterruptedException` to wrap it in a `RuntimeException`, but in general we shouldn't be catching and ignoring `InterruptedException` unless we're also the ones doing the interrupting.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/longrunning/LongRunningTask.java
```diff
@@ -0,0 +1,56 @@
...
+            assertThat(span.getSpanContext().getSpanId(), not(equalTo(INVALID_SPAN_ID)));
+        } finally {
+            span.end();
+        }
+    }
+
+    private String doWork(Span parentSpan) {
+        Span workSpan = tracer.spanBuilder("workSpan")
+                        .setParent(Context.current().with(parentSpan))
+                        .startSpan();
```
Azquelt: It seems odd to be passing through the parent span and setting it manually, rather than making it the active span and having OTel pick it up automatically, but this will work so I guess it's ok.

Azquelt (changes requested):
Couple of small problems

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/longrunning/LongRunningTask.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static final String INVALID_SPAN_ID = "0000000000000000";
+
+    public static final String SUCCESS_MESSAGE = "COMPLETED SUCCESSFULLY";
+
+    @Test
+    public void testLongRunningTask() {
+        Span span = tracer.spanBuilder("longRunningTask").startSpan();
+        try {
+            span.makeCurrent();
```
Azquelt:
```suggestion
        try (Scope scope = span.makeCurrent()) {
```

Need to do the `makeCurrent()` in a try-with-resources like this to make sure it gets made non-current, otherwise the context gets left on the thread.

We always have to ensure we end the span and the scope - the span so that it finishes and is exported, the scope so that we don't leave the wrong context hanging around.

Azquelt (dismissed):

Azquelt (dismissed):

Yas-Franco:
#libby
#build

Yas-Franco:
#build

Yas-Franco:
The io.opentelemetry.context.Scope import was missing so the build failed previously. 

#libby
#build

Yas-Franco:
#build

Azquelt (approved):

PR #24125, opened 2023-01-31
mpTelemetry: Test JAX-RS async resource method

Azquelt:
Test MP Telemetry JAX-RS integration when using an async resource method.

We already have tests for async JAX-RS and MP Rest clients.

For #23448


Azquelt:
#libby
#build
#pb

Azquelt:
#build
#pb

Azquelt:
Async resource methods currently do not work because the scope is not correctly closed on the original request method. This needs to be fixed before these tests can be delivered.

Azquelt:
The problems with async methods were fixed and the tests from this PR were delivered in #24229

PR #24128, opened 2023-01-31
Update non-empty beans.xml test

tevans78:
This is a slight update to the existing non-empty beans.xml test to make it just a "non-empty" test rather than a CDI 1.0 beans.xml test.

#build


benjamin-confino (approved):

tevans78:
#build

PR #24135, opened 2023-01-31
Qualify qsh command references on IBM i

gjwatts:
We are seeing an assortment of strange character translation issues (CCSID) on the IBM i platform in some of our testing.  In talking with one of my IBM i OSS co-workers, a likely cause of this is using an unqualified reference to the `qsh` shell.  In Machine.java we take IBM i commands that come through the execute() method and prefix them with `qsh -c`.  When that happens, the `qsh` that is used can either point to `/usr/bin/qsh` or `/QOpenSys/usr/bin/qsh`, depending on how the system PATH is specified.  The `/QOpenSys/usr/bin/qsh` version of qsh tends to do some extra character translations that could result in the unintelligible characters we are seeing.  By switching to `/usr/bin/qsh` we will use a consistent version that is less likely to perform extra character translations.  This may not fix any current issues, but provides for a more consistent test environment.

### DIFF
--- a/dev/fattest.simplicity/src/com/ibm/websphere/simplicity/Machine.java
+++ b/dev/fattest.simplicity/src/com/ibm/websphere/simplicity/Machine.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2017, 2022 IBM Corporation and others.
+ * Copyright (c) 2017, 2023 IBM Corporation and others.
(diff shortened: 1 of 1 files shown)


gjwatts:
#build

gjwatts:
#build

gjwatts:
#build

gjwatts:
#build

PR #24164, opened 2023-02-02
Use IBM and OpenJ9 specific env to pass criu options

tjwatson:


tjwatson:
#build

tjwatson:
This change to use the j9 specific ENV vars to set the CRIU options is not a good approach because it impacts any code that may be exec'ing another process because it will, by default, inherit the running process environment.  For example, if the `keytool` is exec'ed then it will think it is in checkpoint mode and not behave correctly.

Closing PR.

Issue #24115, opened 2023-01-30
Update Faces 4.0 to RC4


volosied:
Release Notes - MyFaces Core - Version 4.0.0-RC4 can be found in the following link:
https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=10600&version=12352692

PR #24170, opened 2023-02-03
Remove js files that were missed when updating to RC4

pnicolucci:
for #24115


pnicolucci:
#libby #build

volosied (approved):
Sorry! Thanks for find this :)

PR #24184, opened 2023-02-03
PR for Issue 24001: Update SAML metatype descriptions

### LINKED ISSUE #24001: Fix configuration attribute name used in CWWKS1738E message
**Describe the bug**  

When using an OIDC RP via the social login feature, it's possible for the wrong configuration attribute name to be included in the error message emitted when the expected user name claim is not in the ID token returned from the OP. An example of such an error message is below.

> [1/20/23, 12:26:44:281 CST] 00000044 .ws.security.openidconnect.clients.common.AttributeToSubject E CWWKS1738E: The OpenID Connect client [client01] failed to authenticate the JSON Web Token because the claim [someBadName] specified by the [userIdentifier] configuration attribute was not included in the token.

The error message refers to the `userIdentifier` configuration attribute. However, in the social login feature the equivalent configuration attribute is actually called `userNameAttribute`. The message needs to be updated to use the correct attribute name.

**Steps to Reproduce**  
1. Set up an OIDC RP using the social login feature (e.g. `<oidcLogin>`).
2. Set the `userNameAttribute` attribute in the RP configuration to a claim that will not be in the ID token.
3. Go through an OIDC login flow.
4. Observe the error message when the login fails because of the "missing" claim.

**Expected behavior**  
The right attribute name should be in the NLS message.

**Diagnostic information:**  
 - OpenLiberty Version: At least 18.0.0.3 and above. Probably even earlier versions.
 - Affected feature(s): socialLogin-1.0
 - Java Version: N/A
 
   ```XML
   <oidcLogin userNameAttribute="someBadName" ... />
   ```

ayoho:
Updates the descriptions for `reAuthnCushion` and `reAuthnOnAssertionExpire` in the SAML metatype.

Resolves #24001


ayoho:
#build

PR #24199, opened 2023-02-06
Clean up Telemetry bundles

tevans78:
Clean up bnd files;
- set exact versions on buildpath
- set tracegroup correctly
- remove remnants of transformation
- remove private API feature
- Don't import or export from thirdparty jar

#build

### DIFF
--- a/dev/io.openliberty.io.opentelemetry.internal/bnd.bnd
+++ b/dev/io.openliberty.io.opentelemetry.internal/bnd.bnd
@@ -6,12 +6,9 @@
 # http://www.eclipse.org/legal/epl-2.0/
 # 
 # SPDX-License-Identifier: EPL-2.0
-#
-# Contributors:
-#     IBM Corporation - initial API and implementation
 #*******************************************************************************
 
--include= ~../cnf/resources/bnd/bundle.props,~../cnf/resources/bnd/transform.props
+-include= ~../cnf/resources/bnd/bundle.props
 bVersion=1.0
 openTelemetryVersion=1.19.0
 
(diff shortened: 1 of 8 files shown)


dev/io.openliberty.io.opentelemetry.internal/bnd.bnd
```diff
@@ -59,29 +56,29 @@ Require-Capability: osgi.ee; filter:="(&(osgi.ee=JavaSE)(version=11))"
...
+	io.opentelemetry:opentelemetry-exporter-zipkin;version='1.19.0';strategy=exact,\
+	io.opentelemetry:opentelemetry-extension-trace-propagators;version='1.19.0';strategy=exact,\
+	io.opentelemetry:opentelemetry-sdk;version='1.19.0';strategy=exact,\
+	io.opentelemetry:opentelemetry-sdk-common;version='1.19.0';strategy=exact,\
+	io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi;version='1.19.0';strategy=exact,\
+	io.opentelemetry:opentelemetry-sdk-extension-autoconfigure;version='1.19.0.alpha';strategy=exact,\
+	io.opentelemetry:opentelemetry-sdk-logs;version='1.19.0.alpha';strategy=exact,\
+	io.opentelemetry:opentelemetry-sdk-metrics;version='1.19.0';strategy=exact,\
+	io.opentelemetry:opentelemetry-sdk-trace;version='1.19.0';strategy=exact,\
+	io.opentelemetry:opentelemetry-semconv;version='1.19.0.alpha';strategy=exact,\
```
Azquelt, line 80: Unsure that this is necessary, I think `-buildpath` should pick the lowest available matching version.

Azquelt (dismissed):
Thanks

tevans78:
#build

Azquelt:
Last build failed due to infrastructure problems.

#build
#libby

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/build.gradle
```diff
@@ -1,16 +0,0 @@
...
- * 
- * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *     IBM Corporation - initial API and implementation
- *******************************************************************************/
-
-dependencies {
-  requiredLibs project(':io.openliberty.mpTelemetry.1.0.thirdparty')
-}
```
Azquelt: I think this is still needed. Although we are only calling the OTel API on the server (where it's provided by liberty), the test process will need the API classes to be available so that the classes are loadable when building the archives using ShrinkWrap.

tevans78:
#build

dev/io.openliberty.io.opentelemetry.internal/bnd.bnd
Azquelt: I remember we discussed this and decided using exact here was not required but was preferable.

In the future we may add newer versions of these dependencies and we should get an error if the older versions are removed.

Azquelt (approved):

tevans78:
#libby

PR #24212, opened 2023-02-07
Simple transactions test for checkpoint

dazavala:
Verify basic transaction manager operation when restoring a server that was checkpointed at `applications`.


dazavala:
#build

dazavala:
No longer using this PR.

PR #24229, opened 2023-02-08
Telemetry: More JAX-RS tests and bug fixes

### LINKED ISSUE #24174: mpTelemetry context is not cleared when server-side async JAX-RS is used
When the user requests async processing in their jax-rs resource method (i.e. their method returns `CompletionStage` or has an argument of `@Suspended AsyncResponse`), the container request and response filters may not run on the same thread.

In this case, our request filter sets the current telemetry context on the thread, but we have no opportunity to remove it and it remains active after the request method has returned.

This can cause problems if the thread is re-used and nothing clears the telemetry context.
- If the thread next handles a JAX-RS request, it will clear the context
- If the thread next handles a servlet request or an async task from a managed executor, it will not clear the context

### LINKED ISSUE #24194: InMemorySpanExporter expects start time to be synchronized across requests
`InMemorySpanExporter` orders spans by their start time.

This assumes that later events have later start timestamps, even across different requests which is not true on some platforms.

On some platforms (at least on Windows), you can get higher resolution for relative time from `System.nanoTime()` than you can for the current time from `Clock.systemUTC().instant()`. Therefore OTel uses [AnchoredClock][anchored-clock] which gets the current time the first time it's used and then uses `System.nanoTime()` for getting subsequent times.

This means you get more accurate (and guaranteed increasing) timestamps within one request, but not across multiple requests.

We need to rethink how we order the spans returned from `InMemorySpanExporter`.

RTC294287

[anchored-clock]: https://github.com/open-telemetry/opentelemetry-java/blob/v1.19.0/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/AnchoredClock.java

### LINKED ISSUE #24235: ClientAsyncTaskWrapper not called when concurrent-3.0 is enabled
In my `mpTelemetry-1.0` FATs, if `concurrent-3.0` is enabled then my `TelemetryClientAsyncTaskWrapper` is not called when I make an async request with JAX-RS client which causes my tests to fail.

I think this is due to [AsyncClientExecutorService.supplyAsync][code-ref] passing the `Supplier` directly to the executor if it's a `CompletionStageExecutor`.

[code-ref]: https://github.com/OpenLiberty/open-liberty/blob/b29cdb7592dbc4d0470df3c70c32554719a15831/dev/io.openliberty.org.jboss.resteasy.common.ee10/src/io/openliberty/restfulWS/client/AsyncClientExecutorService.java#L199

Azquelt:
New Telemetry JAX-RS tests: (all for #23448)
- Test telemetry with JAX-RS server-side async (`CompletionStage` and `@Suspended AsyncResponse`)
- Test telemetry with all HTTP methods
- Test telemetry with a selection of different HTTP response codes

Bug fixes:
- Close the scope correctly for `SERVER` spans when JAX-RS server-side async is used (product bug) (fixes #24174)
- Wrap rest async client tasks correctly when `concurrent-3.0` is enabled (product bug in restfulWS) (fixes #24235)
- Don't rely on timestamps to order spans in `InMemorySpanExporter` (test bug) (fixes #24194)

Other improvements:
- Add `SpanDataMatcher` (based on existing `SpanMatcher` for Jaeger spans) to create better error messages when spans are not as expected
- Add `TestSpans` utility to create and close test spans to reduce test boilerplate.

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2022 IBM Corporation and others.
+ * Copyright (c) 2022, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -13,6 +13,7 @@
 package io.openliberty.microprofile.telemetry.internal.rest;
 
 import java.util.concurrent.Callable;
+import java.util.function.Supplier;
 
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.ConfigurationPolicy;
@@ -36,6 +37,11 @@ public class TelemetryClientAsyncTaskWrapper implements ClientAsyncTaskWrapper {
     @Override
     public <T> Callable<T> wrap(Callable<T> c) {
         return Context.current().wrap(c);
+    }
+
+    @Override
+    public <T> Supplier<T> wrap(Supplier<T> s) {
+        return Context.current().wrapSupplier(s);
     }
 
 }
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/bnd.bnd
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/bnd.bnd
@@ -32,7 +32,8 @@ src: src, resources
 -dsannotations: \
   io.openliberty.microprofile.telemetry.internal.cdi.SPIMetaData,\
   io.openliberty.microprofile.telemetry.internal.rest.TelemetryClientAsyncTaskWrapper,\
-  io.openliberty.microprofile.telemetry.internal.rest.TelemetryClientBuilderListener
+  io.openliberty.microprofile.telemetry.internal.rest.TelemetryClientBuilderListener,\
+  io.openliberty.microprofile.telemetry.internal.rest.TelemetryServletContainerInitializer
 
 -dsannotations-inherit: true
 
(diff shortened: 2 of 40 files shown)


Azquelt:
#libby
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -4,7 +4,7 @@
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- *
```
Yas-Franco, line 7: The date in this header needs updating

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -4,7 +4,7 @@
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
```
jim-krueger, line 7: Add 2023

Azquelt: bleh, thanks

jim-krueger (dismissed):

Azquelt:
Rebased on integration to fix a soft merge conflict with `TelemetryLongRunningTest` and have made fixes suggested in review comments.

Azquelt:
Squashed review changes into their appropriate commits. (Original changes were f037d84 and fa7b7df)

Yas-Franco (dismissed):
Looks good

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
Azquelt: It has been, the latest version reads `2022, 2023`

Azquelt:
#build
#libby

jim-krueger (dismissed):

Azquelt:
Found a mistake in the code where I moved the closing of the scope to a servlet listener which caused the agent tests to fail. Now fixed.

#libby
#build

Yas-Franco (approved):
Looks good

PR #24239, opened 2023-02-08
Fix JPA test case for RTC 290560

jgrassel:
Test creates its own EntityManager instances, instead of using the one wrapped by jpaResource, then tries to use EntityTransaction transaction wrapper from jpaResource (which is associated with the EntityManager wrapped by itself) instead of getting the EntityTransaction associated with the EntityManager instance created by the thread.  This results in the test calling EntityTransaction.isActive() and .rollback() on the wrong EntityManager instance.


jgrassel:
#build

dazey3 (approved):
LGTM! Thanks for finding this!

PR #24243, opened 2023-02-09
Telemetry: Don't set the current context in the client filter

### LINKED ISSUE #24227: Investigate whether telemetry client filters need to set current context
Currently, TelemetryClientFilter is calling `context.makeCurrent()` - I don't think that should be necessary as the filter is passing the context with the request by setting HTTP headers.

There shouldn't be anything else that actually needs to call `Span.current()` and get the client span.

We do still need to call `start` and `end`, but I don't think we need to call `makeCurrent()` and close the resulting `Scope`.

Azquelt:
Don't set the current context in the client filter. There isn't any
need, the client span is only supposed to cover the actual request and
its details are propagated via the request headers. There's nothing else
that should run within the client span so there's nothing that should
need to have the client span set as the current context.

Never setting the current span means we don't have to worry about
closing the scope, though we do still have to worry about ending the
span. This isn't an issue at the moment, but potentially could be if the
response was handled on a different thread to the request.

Fixes #24227


Azquelt:
Rebased on integration
#build
#libby

Azquelt:
#build

Yas-Franco (approved):
Looks like a sensible change- approved

PR #24257, opened 2023-02-10
IdClass parameters on queries

njr-11:
This covers IdClass as positional parameters to find/filter/update/exists/count/delete queries.  It does not cover named parameters, which will need to be looked at separately.


njr-11:
#build

jmstephensgit (approved):

PR #24263, opened 2023-02-12
Test tracing between opentracing and mpTelemetry

### LINKED ISSUE #24139: Test mpTelemetry with mpOpenTracing and Jaeger
We should be able to have one server running `mpOpenTracing` and another running `mpTelemetry`, tracing to the same server and having trace IDs propagate between them.

We have guides for running `mpOpenTracing` with both [Jaeger][jaeger-guide] and [Zipkin][zipkin-guide] and we should be able to test both of those scenarios.

We also already have tests which set up and use Zipkin and Jaeger trace servers.

[jaeger-guide]: https://openliberty.io/guides/microprofile-opentracing-jaeger.html
[zipkin-guide]: https://openliberty.io/guides/microprofile-opentracing.html

Azquelt:
Test tracing calls between a server using mpTelemetry and a server using mpOpenTracing.

Fixes #24139

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/bnd.bnd
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/bnd.bnd
@@ -1,5 +1,5 @@
 #*******************************************************************************
-# Copyright (c) 2022 IBM Corporation and others.
+# Copyright (c) 2022, 2023 IBM Corporation and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License 2.0
 # which accompanies this distribution, and is available at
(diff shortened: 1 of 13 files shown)


Azquelt:
#build

Yas-Franco (dismissed):
Looks good!

Azquelt:
The new tests passed locally but failed in the build, need to investigate why.

Azquelt:
The OpenTracing server is failing with `java.lang.ClassNotFoundException: okhttp3.Interceptor`

I suspect this is due to maven artifacts in the internal artifactory not having the full dependency metadata. We will need to work out the full list of dependencies required and list them explicitly in the build.gradle.

Azquelt:
Added the missing libraries for the jaeger library.

#build
#libby

Yas-Franco (approved):
Looks good

PR #24266, opened 2023-02-13
JAX-WS: Add ignore for `CWWKO0801E` in JAX-WS negative test case. 

neuwerk:
This pull request ignores the following error in LibertyCXFNegativePropertiesTest's tear down method:

```CWWKO0801E: The SSL connection cannot be initialized from the 127.0.0.1 host and 61,320 port on the remote client to the 127.0.0.1 host and 8,020 port on the local server. Exception: javax.net.ssl.SSLHandshakeException: Received fatal alert: certificate_unknown```

This is caused by a change in messaging on certain JDKs and since the test case is a collection of negative tests, and all tests are passing, this error can be safely ignored.


neuwerk:
#build

sawadood (approved):

PR #24269, opened 2023-02-13
Telemetry: test JAX-RS redirect and invalid

Azquelt:
Test requesting a JAX-RS method which redirects to another JAX-RS method.

Test requesting a URL within a JAX-RS application but which does not map to a JAX-RS resource method.

For #23448


Azquelt:
#build

Yas-Franco (approved):
Looks good!

PR #24271, opened 2023-02-13
Telemetry: document tests via JAX-RS as deliberate

### LINKED ISSUE #24226: Test mpTelemetry rest clients with/without a jax-rs request
RESTful WS client and MP Rest Client behave slightly differently depending on whether the server is handling a JAX-RS request when the client is called.

I.e. the same code behaves differently when it's called from a servlet vs. when it's called from a JAX-RS resource method.

Ensure we have adequate testing for both scenarios.

Azquelt:
Document that we want to keep the structure of the tests in JaxRsEndpoints.

Although these tests would be clearer if re-written to use FATServlet, keeping them in this style means that we have test coverage using the JAX-RS and MP Rest Clients when called via JAX-RS as well as when called via servlet.

Fixes #24226


Azquelt:
#build

Yas-Franco (approved):
Looks good

PR #24147, opened 2023-02-01
Fix OL FATs that fail on IBM i due to incorrect shell usage

gjwatts:
Fixes: #24258

PR #24285, opened 2023-02-14
Fix FAT shells on IBM i

### LINKED ISSUE #24258: Some OL FATs fail on IBM i due to incorrect shell usage
On IBM i, some FATs hit untranslated EBCDIC characters that cause the test to fail.  This is because they are calling the `/bin/sh` shell by default, which is an ILE program (EBCDIC) and should be avoided since FATs run in PASE (ASCII).  

Using the non-ILE shell `/QOpenSys/usr/bin/sh` was the only shell I could find where we did not run into the untranslated EBCDIC characters in our FAT testing.

gjwatts:
Fixes #24258


gjwatts:
#build

jim-krueger (approved):
LGTM

gjwatts:
Closing.  Original issue #24258 will be fixed by PR #24147 instead.

PR #24296, opened 2023-02-14
22470: Reduce trace strings

### LINKED ISSUE #22470: Follow on to PR 22426
Add unit tests to address comments on PR 22426

The two comments that were made on 22426 were that some trace strings seemed excessive/were redundant.  This is addressed with this issue.

The other was that unit tests could be added for some new methods.  But, the methods are private and not easily added to any unit testing.  These methods are already covered by FAT tests added with the original PR.

Only test files were updated, so removed the release bug label.

loriadi:
Reduce trace string and remove redundancy.
fixes #22470


loriadi:
#build

loriadi:
#libbybot

tevans78 (approved):

PR #24326, opened 2023-02-15
purge futures list when application is stopped

### LINKED ISSUE #24293: Scheduled Futures leak resources from Managed Executor Services on application stop
**Describe the bug**  
The [`futures` queue](https://github.com/OpenLiberty/open-liberty/blob/aa6a9f874b88a4940fc58fdb5e6a0184f148fbe4/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ManagedScheduledExecutorServiceImpl.java#L63) in ManagedScheduledExecutorServiceImpl holds references to scheduled futures, even once they have completed.

The queue is periodically cleaned when new tasks are scheduled, by the private [`purgeFutures()` method](https://github.com/OpenLiberty/open-liberty/blob/aa6a9f874b88a4940fc58fdb5e6a0184f148fbe4/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ManagedScheduledExecutorServiceImpl.java#L89), but otherwise they are not actively removed, and it isn't called when applications are shutdown. As purgeFutures is private, applications can't call it themselves.

There's also a [deactivate method](https://github.com/OpenLiberty/open-liberty/blob/aa6a9f874b88a4940fc58fdb5e6a0184f148fbe4/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ManagedScheduledExecutorServiceImpl.java#L70-L82) that aggressively cleans up when the server is shutdown.

Our application does a lot of work via scheduled tasks, and in a test/dev environment we want to stop and start the application without restarting the server. We do have other memory leaks that we're tracking down, but this one is making it harder to see what else is going on.

**Steps to Reproduce** 

1. Create a web.xml with a context listener and a scheduled executor service
```
<?xml version="1.0" encoding="UTF-8"?><web-app xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd" id="WebApp_ID" version="3.0">
    <listener>
        <listener-class>com.example.FooContextListener</listener-class>
    </listener>

    <resource-env-ref>
        <resource-env-ref-name>wm/fooScheduler</resource-env-ref-name>
        <resource-env-ref-type>java.util.concurrent.ScheduledExecutorService</resource-env-ref-type>
    </resource-env-ref>
</web-app>
```

2. Create a simple Context Listener class that will submit a number of scheduled tasks to a managed executor
```
package com.example;

import java.lang.reflect.InvocationTargetException;
import java.lang.reflect.Method;
import java.util.concurrent.ScheduledExecutorService;
import java.util.concurrent.TimeUnit;
import javax.naming.InitialContext;
import javax.naming.NamingException;
import javax.servlet.ServletContextEvent;
import javax.servlet.ServletContextListener;

public class FooContextListener implements ServletContextListener {

  private ScheduledExecutorService executorService;

  @Override
  public void contextInitialized(ServletContextEvent sce) {
    try {
      executorService =
          (ScheduledExecutorService) new InitialContext().lookup("java:comp/env/wm/fooScheduler");
    } catch (NamingException ex) {
      ex.printStackTrace();
    }

    for (int i = 0; i < 30; i++) {
      final Bar bar = new Bar();
      executorService.schedule(bar, 1L, TimeUnit.MILLISECONDS);
    }

    ServletContextListener.super.contextInitialized(sce);
  }

  @Override
  public void contextDestroyed(ServletContextEvent sce) {
    ServletContextListener.super.contextDestroyed(sce);
  }

  private static class Bar implements Runnable {

    @Override
    public void run() {
      System.out.println("Hello");
    }
  }
}
```
3. Start Liberty, wait for the Tasks to run (print "Hello" 30 times), then stop the application (not the server). A leak of 30 references to the `$Bar` class is observable, e.g. in JProfiler:
image
image

You can see that the GC Root is via the `futures` queue in https://github.com/OpenLiberty/open-liberty/blob/aa6a9f874b88a4940fc58fdb5e6a0184f148fbe4/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ManagedScheduledExecutorServiceImpl.java#L63

4. In the contextDestroyed() method insert a call to `purgeFutures()` with reflection
```
    try {
      final Method method = executorService.getClass().getDeclaredMethod("purgeFutures");
      method.setAccessible(true);
      method.invoke(executorService);
    } catch (NoSuchMethodException
        | SecurityException
        | IllegalAccessException
        | IllegalArgumentException
        | InvocationTargetException ex) {
      ex.printStackTrace();
    }
```
5. Repeat the test and there is now no leak:
image

**Expected behavior**  
Resources should be released automatically when an application stops, as managed executors cannot be shutdown by application code, or there needs to be an explicit method to clean up. Waiting for 20 new tasks to be scheduled is unreliable.

**Diagnostic information:**  
 - OpenLiberty Version: I'm using 22.0.0.12 currently, but AFAIK it affects all recent versions

njr-11:
fixes #24293

### DIFF
--- a/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ConcurrencyService.java
+++ b/dev/com.ibm.ws.concurrent/src/com/ibm/ws/concurrent/internal/ConcurrencyService.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2020,2022 IBM Corporation and others.
+ * Copyright (c) 2020,2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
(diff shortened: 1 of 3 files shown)


njr-11:
#build

njr-11:
The build showed a ClassCastException in a test where a different implementation is used.  A commit has been added to correct this.
#build
#libby

njr-11:
Build had 1 failure, which was an intermittent timing issue in an unrelated unit test.

mswatosh (approved):

PR #24356, opened 2023-02-16
fix redirect to original resource callback error and caller-initiated paths

### LINKED ISSUE #24205: 501 status code with securityContext.authenticate(request, response, AuthenticationParameters.withParams().newAuthentication(true));
securityContext.authenticate(request, response, AuthenticationParameters.withParams().newAuthentication(true)); 
is returning a status code of 501.
(the @OpenIdAuthenticationMechanismDefinition annotation specifies redirectToOriginalResource=true)
[securityContext.auth.zip](https://github.com/OpenLiberty/open-liberty/files/10669779/securityContext.auth.zip)

### LINKED ISSUE #24284: redirectToOriginalResource does not handle auth endpoint error case properly
while testing with `redirectToOriginalResource` set to `true`, it appears that the redirect does not handle the case where the authentication endpoint returns with an error:

currently, if the authentication endpoint returns with an error, the browser will be redirected to:
`{originalResourceUrl}?code=&state={state}` <-- no auth code in error case

the proper behaviour would be to redirect to:
`{originalResourceUrl}?error={error}&state={state}` <-- may optionally include `error_description` and `error_uri`

jimmy1wu:
fixes #24284, fixes #24205

- redirect to original resource only if the callback from the op's auth endpoint is valid
    - previously, the rp would always redirect to the original resource if `redirectToOriginalResource` is true, before performing any of the callback checks. this lead us to redirect to the original resource even if the callback was invalid, which we shouldn't do.
    - now, after coming back from the op, the rp will:
        1. validate the callback
            - error on the callback if one of the checks fail
        2. redirect to original resource if `redirectToOriginalResource` is true
            - validates the callback again on the original resource
        3. removes the stored state
        4. make the call to the token endpoint
        ...
- do not restore the original resource request when the callback is successful if the original resource request data was not previously stored. this fixes a bug where the runtime would try to restore a request (even though it never was stored in the first place) in the case of a caller-initiated flow. this bug occurred because `HttpMessageContext#isAuthenticationRequest` loses its data on the callback.


jimmy1wu:
#build

jimmy1wu:
#build

jimmy1wu:
#libby

teddyjtorres (approved):

PR #24357, opened 2023-02-17
Create Jakarta securityContext.authenticate tests

c00crane:
Create tests to validate the behavior of SecurityContext.authenticate


c00crane:
changes were delivered using https://github.com/OpenLiberty/open-liberty/pull/24446 so the token validation tests could be delivered first.

PR #24385, opened 2023-02-20
include beta jar in dev/api/spec

njr-11:
Jakarta Data beta API jar is accidentally being left out of dev/api/spec


njr-11:
#build

nmittles (approved):

PR #24388, opened 2023-02-20
Add checks for unexpected app restart in checkpoint FATS

sebratton:


sebratton:
#build

sebratton:
#build

dev/io.openliberty.checkpoint_fat_persistence/fat/src/io/openliberty/checkpoint/fat/FATSuite.java
```diff
@@ -22,8 +22,9 @@
 @RunWith(Suite.class)
 @SuiteClasses({
                 AlwaysPassesTest.class,
-                DB2Test.class,
-                JPATest.class
+                DB2Test.class
+// Comment out JPA pending upcoming changes to support JPA with checkpoint
+// JPATest.class
```
tjwatson, line 27: Why not call setAssertNoAppRestartOnRestore(false) for `JPATest`?

tjwatson (approved):

dev/io.openliberty.checkpoint_fat_persistence/fat/src/io/openliberty/checkpoint/fat/FATSuite.java
sebratton: The intended use of that setting is to signal that the FAT is exhibiting passing behavior even when the app restarts after restore. I think it's clearer to comment out the test since it's not working ATM.

sebratton:
#libby-bot

sebratton:
The FAT changes were verified by manually running the checkpoint FAT on a checkpoint enabled platform.

sebratton:
Test failures urrelated to the changes

PR #24390, opened 2023-02-20
RTC294172 RTC293474 RTC294307 Build break fixes

benjamin-confino:
This PR contains fixes for 

RTC293474 - This validates an app on startup. It is unlikely to fix the error, but tidies the test a little

RTC294307 - This fixes a test failure by merging two methods annotated AfterClass into one method. This fixes a bug that occurs if they run in the wrong order. 

RTC294172 a follow up to #23816 that ensures the updated file to weld is packaged in the jar we ship with liberty.

### DIFF
--- a/dev/com.ibm.ws.cdi.visibility_fat/fat/src/com/ibm/ws/cdi/visibility/tests/vistest/VisTest.java
+++ b/dev/com.ibm.ws.cdi.visibility_fat/fat/src/com/ibm/ws/cdi/visibility/tests/vistest/VisTest.java
@@ -731,10 +731,7 @@ public class VisTest extends FATServletClient {
         if (server != null) {
             server.stopServer();
         }
-    }
 
-    @AfterClass
-    public static void cleanupFeatures() throws Exception {
         server.uninstallSystemFeature("visTest-1.2");
         server.uninstallSystemFeature("visTest-3.0");
     }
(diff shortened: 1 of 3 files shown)


benjamin-confino:
#build

dev/com.ibm.ws.cdi.visibility_fat/fat/src/com/ibm/ws/cdi/visibility/tests/vistest/VisTest.java
```diff
@@ -731,10 +731,7 @@ public static void afterClass() throws Exception {
         if (server != null) {
             server.stopServer();
         }
-    }
 
-    @AfterClass
-    public static void cleanupFeatures() throws Exception {
```
Azquelt, line 737: Revert this change. Doing it as separate methods ensures the uninstall of features is still done even if the `stopServer()` throws an exception.

dev/com.ibm.ws.org.jboss.weld/bnd.overrides
```diff
@@ -76,5 +76,6 @@ Service-Component: \
 Include-Resource: \
   @${repo;org.jboss.weld:weld-osgi-bundle;[2.4.8.Final,3)}!/!META-INF/maven/*, \
   org/jboss/weld/config=${bin}/org/jboss/weld/config, \
-  org/jboss/weld/context=${bin}/org/jboss/weld/context
```
Azquelt, line 79: Please do this in a separate change.

Azquelt (changes requested):

Azquelt (approved):
Comments can be addressed in a follow up PR.

PR #24398, opened 2023-02-21
RTC294172 / 23816 - add updated ProxyFactory to bnd.overrides so its …

benjamin-confino:
…added to the jar


benjamin-confino:
#build

PR #24411, opened 2023-02-21
adding sort criteria to custom queries

njr-11:
Identify the entity variable name from custom queries and use it when generating additional portions of the query, such as the ORDER BY clause.


njr-11:
#build
#libby

nmittles (approved):
There are a couple double spaces in.  I'm not sure if you want to clean that up. Either way I'm approving.
`@Query("SELECT prime.name, prime.hex FROM  Prime  prime  WHERE prime.number <= ?1")`

njr-11:
> There are a couple double spaces in. I'm not sure if you want to clean that up. Either way I'm approving. `@Query("SELECT prime.name, prime.hex FROM Prime prime WHERE prime.number <= ?1")`

The spaces are added on purpose as a test case that the entity variable name still gets identified correctly with extra whitespace.

PR #24416, opened 2023-02-22
Improved exception handling

benjamin-confino:
A follow up to the RTC294307 commit in #24390

This improves the exception handling


benjamin-confino:
#build

Azquelt (approved):

PR #24424, opened 2023-02-22
Remove server reconfiguration from wsoc SecureTest

### LINKED ISSUE #24423: Remove server reconfiguration in wsoc SecureTest
There is no need to reconfigure the server for this test. It looks like this was initially done because the server was started and the server.xml specifies an application that causes errors if the application wasn't available on server startup.

I'm removing the server reconfiguration and deploying the application and updating the server.xml to point to the apps directory.

pnicolucci:
fixes #24423


pnicolucci:
#libby #build

loriadi (approved):
Good clean up, changes look fine.

isaacrivriv (approved):

PR #24428, opened 2023-02-22
fix and re-enable jakarta security 3.0 tokenMinValidity fat tests

jimmy1wu:
- increased the token lifetime to 60s to ensure tokens are valid for long enough when running on slower machines
- made the `tokenMinValidity` tests less flaky by using the `exp` claim to determine the remaining lifetime rather than trying to figure it out using by subtracting the code execution times from the total lifetime


jimmy1wu:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat.config

jimmy1wu:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat.config

jimmy1wu:
#libby

c00crane (approved):

PR #24376, opened 2023-02-17
Jakarta Security 3.0: Check for required parameters in token response

ayoho:
Resolves #24350
Resolves #24354

Issue #24338, opened 2023-02-16
NPE when access_token missing from token endpoint response


c00crane:
I've omitted the access_token from the response from the token endpoint and received an NPE.

`[2/16/23, 10:18:39:103 CST] 00000052 id=00000000 nliberty.security.jakartasec.identitystore.OidcIdentityStore 3 validate 
                                                                                                               Exception occurred for client client_1
                                                                                                               java.lang.NullPointerException
	at io.openliberty.security.jakartasec.identitystore.OidcIdentityStore.createAccessTokenFromTokenResponse(OidcIdentityStore.java:167)
	at io.openliberty.security.jakartasec.identitystore.OidcIdentityStore.validate(OidcIdentityStore.java:82)
	at org.jboss.weld.generated.proxies.security.enterprise.identitystore.IdentityStore$329146295$Proxy$_$$_WeldClientProxy.validate(Unknown Source)
	at com.ibm.ws.security.javaeesec.identitystore.IdentityStoreHandlerImpl.validate(IdentityStoreHandlerImpl.java:70)
	at com.ibm.ws.security.javaeesec.identitystore.IdentityStoreHandlerImpl.validate(IdentityStoreHandlerImpl.java:58)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.jboss.weld.bean.proxy.AbstractBeanInstance.invoke(AbstractBeanInstance.java:38)
	at org.jboss.weld.bean.proxy.ProxyMethodHandler.invoke(ProxyMethodHandler.java:106)
	at org.jboss.weld.generated.proxies.security.enterprise.identitystore.IdentityStoreHandler$1302446337$Proxy$_$$_Wel`

c00crane:
[npe.zip](https://github.com/OpenLiberty/open-liberty/files/10759915/npe.zip)

c00crane:
Tests are no longer getting the NPE.

PR #24429, opened 2023-02-22
check to prevent npe in case access token is missing in the token response

arunavemulapalli:
address #24338

we may not run into this once the token response validation is done to make sure that the required parameters exist in the response ( #24376 )


arunavemulapalli:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat,io.openliberty.security.jakartasec.3.0.internal_fat.config,io.openliberty.security.jakartasec.3.0.internal_fat.logout,io.openliberty.security.jakartasec.3.0.internal_fat.refresh

ayoho (approved):

PR #24430, opened 2023-02-22
Test sessionDatabase-1.0 with InstantOn

tjwatson:


tjwatson:
#build

tjwatson:
#libby

tjwatson:
#build

dazavala (dismissed):
Hello.  The solution and FAT look great.  Some copyrights need updating.  And one question: Is it necessary to verify the session store db table does not initialize during checkpoint?  This is a specific case of a more general question, actually.

tjwatson:
> Hello. The solution and FAT look great. Some copyrights need updating. And one question: Is it necessary to verify the session store db table does not initialize during checkpoint? This is a specific case of a more general question, actually.

I didn't change anything WRT to db table initialization.  My rational was that each and every server startup would go through the same path as this as far as DB access is concerned.  So here I am just make that so as well.  Only it is ensured to happen on the restore side.  But if you consider a containerized liberty application without InstantOn each time you launch such an application the same path in the code would be run each time you launch that container.  This change makes that also happen for InstantOn.

tjwatson:
#build

tjwatson:
#build

dazavala (approved):
Improvements to thread safety and field simplification in the session invalidator look good.

tjwatson:
#libby

tjwatson:
Fixes #23704

PR #24439, opened 2023-02-23
Fix defect in RTC 294713. Certificates used by was-ldap11 and was-lda…

jvanhill:
…p21 were using old signature algorithms that were no longer accepted by newer versions of Java.


jvanhill:
#build

ryanesch (approved):

PR #24441, opened 2023-02-23
Check that restore succeeded

tjwatson:


tjwatson:
#build

tjwatson:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -1823,6 +1823,19 @@ public void run() {
...
+            // The return code is 0 indicating the server started, likely recovered
+            // set as started but then stop the server
+            setStarted();
+            try {
+                stopServer();
+            } catch (Exception e) {
+                Log.error(c, method, e);
+                // we don't want to fail if stop fails
+            }
+            fail("The server did not restore successfully");
```
sebratton, line 1837: I wonder if failing with an assert like this will still allow for collection of the server log to diagnose the restore from checkpoint which failed? It's similar to the fail assertion for `if (output.getReturnCode() != 0)` above but I don't recall seeing it fail in that code path, so not sure if it will always collect logs there either. We _may_ need a `postStopServerArchive()` call

sebratton (approved):

tjwatson:
#libby

PR #24442, opened 2023-02-23
Archive fat logs when checkpoint operation fails.

sebratton:
Some checkpoint FAT failure scenarios did not save the server logs.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -1840,20 +1840,32 @@ public class LibertyServer implements LogMonitorClient {
      *
      * @param output
      */
-    private void checkpointValidate(ProgramOutput output, boolean expectStartFailure) throws Exception {
-        if (!expectStartFailure) {
-            assertEquals("Checkpoint operation return code should be zero", 0, output.getReturnCode());
+    private void checkpointValidate(ProgramOutput output, boolean expectCheckpointFailure) throws Exception {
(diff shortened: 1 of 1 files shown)


sebratton:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -1840,20 +1840,32 @@ public void run() {
...
-        }
-        // validate server not started
-        resetStarted();
-        if (isStarted) {
-            Exception fail = new Exception("Server should not be started after a checkpoint operation");
-            Log.error(c, "Server should not be started after a checkpoint operation", fail);
-            throw fail;
+    private void checkpointValidate(ProgramOutput output, boolean expectCheckpointFailure) throws Exception {
+        String method = "checkpointValidate";
+        Log.info(c, method, method);
```
tjwatson, line 1845: Did you intend to use `Log.entering` here instead?

tjwatson (approved):

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
sebratton: I initially tried log.entering but it didn't result in any message with the default trace settings.  I really only wanted that to help with understaning the change when I was testing it. But I'd just as soon leave it there. It may be useful in following when the new code path is taken.

sebratton:
#build

sebratton:
#libby-bot

PR #24458, opened 2023-02-24
data entity annotations are removed from the spec

njr-11:
Entity annotations specific to Jakarta Data were removed from the spec in favor of using entity annotations from Jakarta Persistence or Jakarta NoSQL.  This pull removes our copies of those and some related ones we were experimenting with and updates test cases accordingly.


tabishop (approved):

njr-11:
#build

PR #24466, opened 2023-02-27
Avoid assertions of what is logged by Oracle JDBC driver

KyleAure:
When creating the JDBCDriverService in Liberty each DataSource is processed asynchronously.
The Custom Oracle configuration is processed when the first JDBCDriverService.create() method is called. 

But, we are not guaranteed to have finished configuring the oracle logger before any other DataSource is created. 
Thus, we cannot assert that certain output will be contained in the first log where we would expect to see output from the first use of the JDBC driver. 

I thought I had made the test simple enough to make these assertions but over a large number of runs, we can expect this to fail.


tabishop (approved):

KyleAure:
#build

PR #24489, opened 2023-02-28
Avoid timing issues with Oracle custom logging

KyleAure:
Depending on the timing of this test suite it is possible that when we check for log files, the first log file has not been filled yet so we cannot assert that both Oracle logs exist.


KyleAure:
#build

tabishop (approved):

PR #24497, opened 2023-02-28
Avoid resource injection for inferred datasources on IBMi

KyleAure:
When running tests on IBMi we cannot use resource injection when a DataSource configuration depends on inferred data source classes.  
This is because IBMi systems have a version of the Java JDK that includes the DB2 driver on the classpath.  
Therefore, if we use another driver with a nondescript driver name, and generic `properties` we have no way of deciding which DataSource classes to use.


KyleAure:
#build

tabishop (dismissed):

KyleAure:
Rerequested IBMi build at: https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewDefinition&id=_JWdEUD7dEemKFI0xRbWJ_w

The previous IBMi build had an addition test error that wasn't corrected. 

KyleAure:
#libby check headers
@tabishop - rereview the latest commit.

tabishop (approved):

PR #24498, opened 2023-02-28
Refactor of io.openliberty.jaxws.security fat

### LINKED ISSUE #24463: JAX-WS: Refactor io.openliberty.jaxws.security_fat to address chronic timeout failures
The `io.openliberty.jaxws.security_fat` needs to be restructured to address extraordinarily long test execution time, as well as numerous timeout failures. 

- [ ] The tests that repeat the same configuration but against different endpoints should be collapsed into a single test method. Reconfiguration would happen once, but the test could then check expected response against all three endpoints. This should improve test duration due to not re-configuring the application and forcing app restart for each endpoint test. Example:
``` 	// Valid name and valid plain password
	@Test
	public void testValidNameAndValidPlainPasswordPOJO() throws Exception {
		updateClientBndFile("bindings/validNameAndValidPlainPwd.xml");

		RequestParams params = new RequestParams("employee", "pojo", "http", server.getHttpDefaultPort(),
				"/employee/employPojoService");
		runTest(params, "Hello, employee from SayHelloPojoService", null);
	}

	@Test
	public void testValidNameAndValidPlainPasswordStateless() throws Exception {
		updateClientBndFile("bindings/validNameAndValidPlainPwd.xml");

		RequestParams params = new RequestParams("employee", "stateless", "http", server.getHttpDefaultPort(),
				"/employee/employStatelessService");
		runTest(params, "From other bean: Hello, employee from SayHelloSingletonService", null);
	}

	@Test
	public void testValidNameAndValidPlainPasswordSingleton() throws Exception {
		updateClientBndFile("bindings/validNameAndValidPlainPwd.xml");

		RequestParams params = new RequestParams("employee", "singleton", "http", server.getHttpDefaultPort(),
				"/employee/employSingletonService");
		runTest(params, "From other bean: Hello, employee from SayHelloStatelessService", null);
	}

```
The separate methods for POJO, Stateless, Singleton endpoints can be collapsed into a single method like so:

```
	@Test
	public void testValidNameAndValidPlainPassword() throws Exception {
		updateClientBndFile("bindings/validNameAndValidPlainPwd.xml");
                // POJO Test
		RequestParams params = new RequestParams("employee", "pojo", "http", server.getHttpDefaultPort(),
				"/employee/employPojoService");
		runTest(params, "Hello, employee from SayHelloPojoService", null);
                // Stateless  Singleton Test
		RequestParams params = new RequestParams("employee", "stateless", "http", server.getHttpDefaultPort(),
				"/employee/employStatelessService");
		runTest(params, "From other bean: Hello, employee from SayHelloSingletonService", null);
                // Singleton Test
		RequestParams params = new RequestParams("employee", "singleton", "http", server.getHttpDefaultPort(),
				"/employee/employSingletonService");
		runTest(params, "From other bean: Hello, employee from SayHelloStatelessService", null);
	}
```
There is probably an opportunity to abstract this further, but at the very least combining to eliminate as much app restart as possible should be our approach.
- [ ] Most of the test classes do server reconfiguration requiring server start stop every test. We want to avoid that as much as possible. We can split classes up into additional smaller classes to prevent server restart between every test method. 
- [ ] Due to the large number of classes in this project, we should break the test classes up in to multiple test projects. As a start we should move the SSL related classes to a project called `io.openliberty.jaxws.security_fat.ssl` then depending on fat runtime for both projects. We should separate the classes into additional projects as needed. Example moving some tests from `io.openliberty.jaxws.security_fat` to `io.openliberty.jaxws.security_fat.1` and `io.openliberty.jaxws.security_fat.ssl.1`. We can break them into as many numbered projects as we need from there.

BerksanAtes:
Fixes #24463


BerksanAtes:
Updated conflicting class with newest changes. However, since I moved this class to another FAT bucket, git wouldn't track my correction.

BerksanAtes:
Updated conflicting class with newest changes. However, since I moved this class to another FAT bucket, git wouldn't track my correction.

BerksanAtes:
#build

BerksanAtes:
#run-libby-bot

BerksanAtes:
#build

dev/io.openliberty.jaxws.security_fat.1/bnd.bnd
```diff
@@ -0,0 +1,81 @@
...
+# Copyright (c) 2023 IBM Corporation and others.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License 2.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-2.0/
+# 
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
```
neuwerk, line 11: I found out that we aren't adding the `Contributors` section of the header to bnd files any more so you should remove line number 10 and 11
```suggestion
```

neuwerk: This applies to all new files in this PR, so all the java files as well that have this Contributors section as are found below.

dev/io.openliberty.jaxws.security_fat.1/fat/src/com/ibm/ws/jaxws/security/fat/ClientCertificateTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+    }
+
+    // private static final String WITH_CLIENT_ALIAS_CONFIG =
+    // "withAliasClientCertConfiguration.xml";
+
+    // private static final String PATCHY_SERVER_TRUST_STORE_CONFIG =
+    // "patchyServerTrustStoreConfiguration.xml";
+
+    private static final int WAIT_TIME_OUT = 10 * 1000;
```
neuwerk: Would you mind moving this field to the start of the class with the other fields?

dev/io.openliberty.jaxws.security_fat.1/publish/files/JaxWsTransportSecurityServer/bindings/validNameButInvalidPwd.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
neuwerk: Can you verify that all these binding files are needed by this test project, if their are binding files not being used by this particular project, can you delete the ones that aren't being used so we don't have any unnecessary duplicates?

dev/io.openliberty.jaxws.security_fat.1/publish/files/JaxWsTransportSecurityServer/serverConfigs/withAliasClientCertConfiguration.xml
```diff
@@ -0,0 +1,66 @@
+<server>
```
neuwerk: Can you do the same with the server configuration files? Verify that only the ones used by the tests in this project are in this project, so we don't have unnecessary duplicates?

dev/io.openliberty.jaxws.security_fat.ssl/.classpath
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" path="fat/src"/>
+	<classpathentry kind="src" path="test-applications/WsdlLocationHttpsServer/src"/>
+	<classpathentry kind="src" path="test-applications/WsdlLocationHttpsClient/src"/>
```
neuwerk: These class path entries for the test applications look wrong, as they don't seem to match what's actually being used by the tests here. I can't find any reference to the some of these apps in the test classes. You should remove the entries to the tests apps that aren't actually being used in this project.

dev/io.openliberty.jaxws.security_fat.ssl/publish/files/JaxWsTransportSecurityServer/bindings/certInClientKSButNotInServerTS.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
neuwerk: Same comment about verifying that all these binding files are used in this test project, and if any aren't delete the ones not used to avoid duplicates.

dev/io.openliberty.jaxws.security_fat.ssl/publish/files/JaxWsTransportSecurityServer/serverConfigs/basicAuthWithSSL.xml
```diff
@@ -0,0 +1,58 @@
+<server>
```
neuwerk, line 1: Same comment about the server.xml files, make sure this project is using whatever files are here, and delete the ones not in use in this project.

dev/io.openliberty.jaxws.security_fat/fat/src/com/ibm/ws/jaxws/security/fat/TransportSecurityUsingDispatchClientCertTest.java
```diff
@@ -30,58 +34,38 @@
...
+    }
 
-	@AllowedFFDC({ "java.lang.NullPointerException" })
-	@Test
-	public void testValidClientCertDispatchStateless() throws Exception {
-		updateProviderWEBXMLFile("clientCert_provider_web.xml");
-		updateClientBndFile("bindings/validCertAlias.xml");
+    @AllowedFFDC({ "java.lang.NullPointerException" })
+    @Test
+    public void testValidClientCertDispatch() throws Exception {
```
neuwerk: Can you actually move this test in to the ClientCerticateTest? I don't think there is a reason to have this as it's own separate test class, when it's doing the same thing as the ClientCertificateTest. We can just combine this test method into that class, and it should be good to go.

dev/io.openliberty.jaxws.security_fat/fat/src/com/ibm/ws/jaxws/security/fat/TransportSecurityUsingDispatchTest.java
```diff
@@ -28,135 +32,70 @@
 @Mode(Mode.TestMode.FULL)
 public class TransportSecurityUsingDispatchTest extends AbstractJaxWsTransportSecurityTest {
```
neuwerk: Can you do the same thing with this class? I don't think there is a reason to have this as it's own separate test class, when essentially it's doing the same tests as the BasicAuthWithoutSSL and BasicAuthWithSSL. The tests that apply to the BasicAuthWithoutSSL can be moved to that class, and the ones that apply to the BasicAuthWithSSL can move to that class. And then we can just remove this test class entirely.

neuwerk (changes requested):

dev/io.openliberty.jaxws.security_fat.1/fat/src/com/ibm/ws/jaxws/security/fat/ClientCertificateTest.java
BerksanAtes: It was like this before my changes. But it would look tidier.

dev/io.openliberty.jaxws.security_fat.1/bnd.bnd
BerksanAtes: Hmm that's new.

BerksanAtes:
#build

neuwerk (approved):
LGTM! Thanks @BerksanAtes

neuwerk:
#run-libby-bot

Issue #24505, opened 2023-03-01
Update SSO FATs to generate results when tests can't be run


c00crane:
1. Update the FATSuites for all SSO projects to include AlwaysPassesTest.class
2. update CommonAltRemoteLDAPServerSuite.java, CommonRemoteLDAPServerSuite.java and CommonLocalLDAPServerSuite.java to catch failures/exceptions from LocalLDAPServerSuite.setUpUsingServers 
and either
- invoke junit's "fail" the project will fail, but have results (pros - project completes faster, problem may be more obvious, cons - tests that could have possibly succeeded won't be run)
- just log the failure and continue on - some tests may run ok, but others should fail (pros - tests that can succeed will run and succeed, cons - runs longer, root of the problem may be harder to determine, no tests may succeed

c00crane:
#24518

PR #24518, opened 2023-03-02
Make SSO FATs handle infrastructure issues better wrt junit reporting

c00crane:
When the SSO FATs run into problems obtaining LDAP servers they don't report junit results.  This yeilds no logs when the results are reported.  That makes it hard to debug.
Recently, there were issues with artifactory and there were no logs to pin down the issue.
The FATSuite sets up LDAP server references for the tests to use - this was failing causing the FATSuite to exit without running tests.
I'm adding a catch around "LocalLDAPServerSuite.setUpUsingServers".  We'll log the error and continue running tests.  Some FATs will be able to run some of the test classes that use a BasicRegistry and others will all fail.  But, we'll get results.
I'm also adding AlwaysPassesTest to FATSuites that do not already have it.  This will always produce junit results.


c00crane:
#24505

c00crane:
I also replaced use of AlwaysRunAndPassTest (from the common security fat project) with AlwaysPassesTest which is available to all FATs from fattest.simplicity

c00crane:
#libby
#build

c00crane:
Retry build - first attempt failed due to infrastructure issues
#build

ayoho (approved):

c00crane:
#libby
#build

PR #24544, opened 2023-03-03
log error if jwks read or connect timeout is negative

jimmy1wu:
log the existing `CWWKS2507W` warning message if `jwksConnectTimeout` or `jwksReadTimeout` (or their EL counterparts) are negative. the same warning message is already being issued when `tokenMinValidity`/`tokenMinValidityExpression` is negative.

    /**
     * Optional. Sets the connect timeout(in milliseconds) for Remote JWKS
     * retrieval. Value must not be negative and if value is zero then infinite
     * timeout.
     *
     * @return
     */
    int jwksConnectTimeout() default 500;

    /**
     * Optional. Sets the read timeout(in milliseconds) for Remote JWKS
     * retrieval. Value must not be negative and if value is zero then infinite
     * timeout.
     *
     * @return
     */
    int jwksReadTimeout() default 500;

https://github.com/jakartaee/security/blob/master/api/src/main/java/jakarta/security/enterprise/authentication/mechanism/http/OpenIdAuthenticationMechanismDefinition.java


jimmy1wu:
#build

jimmy1wu:
#build

jimmy1wu:
#libby

ayoho (approved):

PR #24545, opened 2023-03-03
Add time for Leap Day on Renew Acme FAT

kristip17:
Our test, AcmeValidityAndRenewTest, that sets a renewBeforeExpiration set longer than the certificate validity period started to fail because the time was no longer long enough -- add extra days to adjust for leap year. Bumped from +1 day to +10 days so we should haven't an issue in the future.

For RTC 294975


kristip17:
#build

kristip17:
The test had a connection timeout launching some of the 10 FAT bucket repeates. But 4 Lites did run and 1 Full ran successfully, since this is a Full bucket test, I'm not worried about the extra Lite buckets that didn't get launched.

jvanhill (approved):

kristip17:
#libby

PR #24557, opened 2023-03-06
Enable MyFaces-4433 test for ee10

pnicolucci:
for #22369

PR #24560, opened 2023-03-06
Enable testMyFaces4433 since MYFACES-4563 was fixed in 4.0.0

volosied:
for #22369


volosied:
Closed in favor of https://github.com/OpenLiberty/open-liberty/pull/24557

PR #24564, opened 2023-03-06
Update jwks read timeout in ProviderMetadata app

c00crane:
On slow test systems the jwks read can time out if the default timeout of 500 milliseconds is used.  Most of the test applications override the value.  This one instance did not have the update.  I just need to add jwksReadTimeoutExpression to the @OpenIdAuthenticationMechanismDefinition annotation.


c00crane:
#build

teddyjtorres (approved):

PR #24568, opened 2023-03-06
Update java requirement for mpJwt 2.1 FATs

c00crane:
Add javac.source: 11 to the bnd.bnd for mpJwt 2.1 FATs since they can't run with Java 8.  The repeat rules will ship the one and only repeat so even the AlwaysPassesTest won't run.


c00crane:
#build

c00crane:
#libby

teddyjtorres (approved):

PR #24581, opened 2023-03-07
Increase timeoutCushion for TCK timeout tests to account for slower h…

jim-krueger:
Increased the org.eclipse.microprofile.rest.client.tck.timeoutCushion jvm option to 60000 to avoid build-break defects on slower hardware.  This cushion is still well below the default connection timeout of 120 seconds so the tests are still valid.

I've attempted to reproduce this problem locally using the same OS and JDK versions, but have not been able to.


jim-krueger:
#libby #build

jhanders34 (approved):

jim-krueger:
#build

jim-krueger:
Build failure unrelated to this change.  Merging.

Issue #24589, opened 2023-03-07
Investigate timing when issuing first token after OP startup


c00crane:
We have tests that are using short lived tokens (lifetimes of 20 seconds).  The tokens that we get the first time that we try to access a protected app after starting the OP are expired by the time the client runtime tries to use them.
We're seeing this when using @OpenIdAuthenticationMechanismDefinition to protect apps, but, I think this is an issue with the OP.

In the attached logs, search for test case: BasicRefreshTests_tokenAutoRefreshFalse_providerAllowsRefreshFalse_notifyProviderTrue_idTokenExpiryTrue_accessTokenExpiryFalse_jwt
[fatoutput-io.openliberty.security.jakartasec.3.0.internal_fat.refresh.zip](https://github.com/OpenLiberty/open-liberty/files/10912652/fatoutput-io.openliberty.security.jakartasec.3.0.internal_fat.refresh.zip)

jimmy1wu:
these are the largest time gaps between logs i found in the OP's trace logs

9.6 seconds:
```
[2/10/23, 17:05:29:720 PST] 00000091 id=00000000 com.ibm.ws.channel.ssl.internal.SSLHandshakeIOCallback       < complete(read) Exit 
[2/10/23, 17:05:39:363 PST] 0000011f id=00000000 com.ibm.ws.channel.ssl.internal.SSLConnectionLink            > init, vc=1797138549 Entry 
```

4.5 seconds:
```
[2/10/23, 17:05:41:367 PST] 000000ff id=00000000 com.ibm.ws.channel.ssl.internal.SSLHandshakeIOCallback       < complete(read) Exit 
[2/10/23, 17:05:45:849 PST] 00000090 id=00000000 com.ibm.ws.channel.ssl.internal.SSLConnectionLink            > init, vc=868043983 Entry 
```

10.8 seconds:
```
[2/10/23, 17:05:48:881 PST] 00000090 id=00000000 com.ibm.ws.channel.ssl.internal.SSLReadServiceContext        < error Exit 
[2/10/23, 17:05:59:724 PST] 0000011e id=00000000 com.ibm.ws.channel.ssl.internal.SSLReadServiceContext        > error, vc=1609202224 Entry 
```

Issue #24592, opened 2023-03-07
Feature resolver processing is too slow when used from Liberty Maven Plugin feature install, for already-installed features


scottkurz:
**Describe the bug**  

First, calling this a "bug" is subjective but read on.

When using Liberty Maven Plugin "dev mode", most of the startup sequence is pretty quick and not much longer than the startup of the Liberty server started within..... **EXCEPT** that is, for the feature install.    

Using the timing of the liberty-maven-plugin console log messages as a close-enough proxy for the feature install execution time, when installing these features:
> [INFO] Installing features: [mpconfig-2.0, mpmetrics-3.0, mphealth-3.1, cdi-2.0, jaxrs-2.1]

it takes about 45 seconds for the features to install initially.   That's fine.  Not trying to raise a concern when the features actually need to be installed upon the first time dev mode starts after Liberty is installed.

**However**, it takes about 15 seconds to get through this step every time dev mode is started after the first.     From the user perspective, a noticeable delay "installing features x,y,z" that have already been installed starts to get annoying and makes the whole dev mode experience less pleasant. 

**Call stack details**  

By running dev mode in a Java debugger (e.g. `mvnDebug liberty:dev`) I narrowed down the routine taking a long time to the **FeatureResolverImpl.processRoots()** call within this call stack:

```
FeatureResolverImpl.processRoots(Collection<String>, Set<String>, SelectionContext) line: 305    
FeatureResolverImpl.processCurrentPermutation(Collection<String>, Set<String>, SelectionContext) line: 296    
FeatureResolverImpl.doResolveFeatures(Collection<String>, Set<String>, SelectionContext) line: 259    
FeatureResolverImpl.resolveFeatures(Repository, Collection<ProvisioningFeatureDefinition>, Collection<String>, Set<String>, Set<String>, EnumSet<ProcessType>) line: 182    
FeatureResolverImpl.resolveFeatures(Repository, Collection<ProvisioningFeatureDefinition>, Collection<String>, Set<String>, boolean) line: 114    
RepositoryResolver.resolveFeatures(ResolutionMode) line: 468    
RepositoryResolver.resolve(Collection<String>, ResolutionMode) line: 364    
RepositoryResolver.resolveAsSet(Collection<String>) line: 351    
InstallKernelMap.singleFileResolve() line: 929    
InstallKernelMap.get(Object) line: 273    
InstallMap.get(Object) line: 275    
[...]

PR #24593, opened 2023-03-07
Update short token lifetime tests to avoid tokens expiring too soon

c00crane:
Update Jakartasec FAT that use tokens with short lifetimes.
The OP is taking extra time to issue the first token after the server starts.
Adding a request from the test class @BeforeClass method to make a request of the OP - that should get the OP moving.
We've only seen a problem with the first request to the OP - this should resolve the test problem and I've opened issue 
#24589 to investigate the runtime further.

#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat,io.openliberty.security.jakartasec.3.0.internal_fat.config,io.openliberty.security.jakartasec.3.0.internal_fat.config.1,io.openliberty.security.jakartasec.3.0.internal_fat.config.2,io.openliberty.security.jakartasec.3.0.internal_fat.logout,io.openliberty.security.jakartasec.3.0.internal_fat.refresh


jimmy1wu (approved):

scottkurz:
Looks like you meant to link to #24589 (not #24592 )

c00crane:
there were some failures in the build, but all were un-related to this change.  Unrelated fat projects failed and the failure in io.openliberty.security.jakartasec.3.0.internal_fat is an un-related issue which has been fixed by #24623

PR #24602, opened 2023-03-07
OneToOne entities

njr-11:
Initial set of updates to allow OneToOne entity mappings.
At this point, you can define a OneToOne mapping and not otherwise have to declare the entity that it maps to, and you can have a repository method return type be an entity that was mapped as OneToOne.


njr-11:
#build

jmstephensgit (approved):

PR #24610, opened 2023-03-08
Objectclass in filter should be case insensitive

ryanesch:
Using 'objectclass' in userFilter and groupFilter should be case insensitive. This is hard to diagnose as a user, but you may be hitting this issue if you are making user registry calls for display name and nothing is returned, but a user registry call for security name does return the user.
This is because the filters are parsed for objectclasses that are later used to identify entities as Groups and PersonAccounts.


ryanesch:
#build

ryanesch:
#build

jvanhill (approved):

ryanesch:
#build

wraschke:
#libby

PR #24623, opened 2023-03-08
Remove hard-coded port in test app annotation

c00crane:
The RedirectToOriginalResourceTruePromptNoneServlet app has a hard coded port in the @OpenIdAuthenticationMechanismDefinition annotation.  We can't guarantee the port that we'll get when the tests are running.  This instance of the annotation should use the variable value set by the test class at runtime.

There is a test that is using a hard coded value for the providerURI, but that test is skipped if we don't get the default port of 8920.  This test does not need a hard coded value.


c00crane:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat

jimmy1wu (approved):

teddyjtorres (approved):

Issue #24631, opened 2023-03-08
Fix ClassCastException during the de-serialization of CDI Injected Event


volosied:
**Describe the bug**  
A passivated (i.e. serializable) bean which has an injected jakarta.enterprise.event.Event (or javax equivalent) will not be properly restored.  This can be encountered when session persistence is enabled and session data is serialized and de-serialized from a database.

The FFDC: 
```
Stack Dump = java.lang.ClassCastException: cannot assign instance of org.jboss.weld.event.EventImpl$SerializationProxy to field org.apache.myfaces.flow.cdi.FlowScopeContextualStorageHolder.flowDestroyedEvent of type jakarta.enterprise.event.Event in instance of org.apache.myfaces.flow.cdi.FlowScopeContextualStorageHolder
	at java.base/java.io.ObjectStreamClass$FieldReflector.setObjFieldValues(ObjectStreamClass.java:2076)
	at java.base/java.io.ObjectStreamClass$FieldReflector.checkObjectFieldValueTypes(ObjectStreamClass.java:2039)
	at java.base/java.io.ObjectStreamClass.checkObjFieldValueTypes(ObjectStreamClass.java:1293)
	at java.base/java.io.ObjectInputStream.defaultCheckFieldValues(ObjectInputStream.java:2512)
```
 

**Steps to Reproduce**  

If a passivated bean contains code as such below, and this bean is then serialized, such as via sessionDatabase-1.0 , the ClassCastException will occur: 

    @Inject
    private Event<someType> someEvent;

**Expected behavior**  
The CDI event objects should be restored without errors. 

**Diagnostic information:**  
 - OpenLiberty Version: All. (Fix targeted for 22.0.0.3) 
 - Affected feature(s) cdi-1.2, cdi-2.0, cdi-3.0, cdi-4.0, (sessionDatabase-1.0, if following the example above) 
 - Java Version: n/a
 - server.xml configuration: n/a 
 
**Additional context**  
This problem was discovered via faces-4.0 (currently in beta as of this post) for this following event: https://github.com/apache/myfaces/blob/myfaces-core-module-4.0.0/impl/src/main/java/org/apache/myfaces/flow/cdi/FlowScopeContextualStorageHolder.java#L65-L67

volosied:
Fixed, so I'll close it out. 

Tests will be added under #24679

PR #24636, opened 2023-03-09
Fix passivation of CDI beans with injected Event

benjamin-confino:
Add two packages to the list of packages to search when deserializing objects from a passivated bean.

`org.jboss.weld.event` contains the implementation for an injected `Event` and after some debugging we found that `org.jboss.weld.annotated.slim.unbacked` was needed in some cases as well, though this isn't shown by the test in this PR.

for #24631


benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
Previous build didn't fail, but its still running and way overdue.

benjamin-confino:
#libby

volosied (approved):
Changes are good for MyFaces! Thanks! Please merge when you can :)

Issue #21992, opened 2022-08-05
Update Resolver so that it can better handle tolerates


pnickoll:
Currently the resolver (when using resolve to install a single feature) creates a single set of prereqs. For features that have tolerates we could (easily) end up in a situation where the list of features are not compatible with each other as resolve was not designed with tolerates in mind and does not know the list of features a user has in their server.xml file (they could use multiple server.xml files too each with different dependencies, so installing one prereq chain is unlikely to work for both servers).

Quote from email chain: "It used to be that you could install any feature and get all of its dependencies and be sure it would start with any other feature, but that changed when we added tolerates. It got way worse with EE9 because we started using tolerates a lot more so that we could have one version of a management-type feature (like admin center) which works with both EE8 and EE9. I don't think it's even possible for a user to manually install all possible dependencies of a feature since we have public features which tolerate multiple versions of a private feature and we don't document the existence of the private features."

### Proposal
To fix this we could instead install every feature in the prereq list instead of just 1 set of prereqs. This will end up installing more features, quite possibly more than required, but without knowing the server.xml we can't tell which exact features are required.

This work could either be done in the kernel resolver (which already does a lot of the resolver work) or in the repository resolver layer (which does a lot less work but does know about install lists which is something we could take advantage of).

Check the existing docs and guides to see if we need to update them regarding this change

### Case 1 (most common scenario).

User installs features independently and starts the server

`./installUtility install jakartaee-9.1 localConnector-1.0 microProfile-5.0`

OR

`./featureUtility installfeature jakartaee-9.1 localConnector-1.0 microProfile-5.0`

server.xml configuration

```
<server description="new server">

   <!-- Enable features -->
   <featureManager>
        <feature>jakartaee-9.1</feature>
        <feature>localConnector-1.0</feature>
        <feature>microProfile-5.0</feature>
  </featureManager>
```

then start server:

./server start <server>

```
….
[...]

PR #24638, opened 2023-03-09
Address repository resolver problems

Azquelt:
Basic resolve (i.e. not resolveAsSet) is now expected to return all
tolerated dependencies of all features to be installed.

~ResolveAsSet is now expected to return all auto-features which would be
activated if already installed features which are not in the requested
set are considered.~ This requirement was removed, see discussion in the issue.

For #21992

### DIFF
--- a/dev/com.ibm.ws.repository.resolver/src/com/ibm/ws/repository/resolver/FeatureTreeWalker.java
+++ b/dev/com.ibm.ws.repository.resolver/src/com/ibm/ws/repository/resolver/FeatureTreeWalker.java
@@ -1,0 +1,221 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package com.ibm.ws.repository.resolver;
+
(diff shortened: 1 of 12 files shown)


Azquelt:
Now that the question about installing all enabled autofeatures when doing resolveAsSet has been answered, I've updated the tests accordingly.

However, while doing this I noticed a few more problems with these changes:
- When doing basic resolve, we were finding and installing any features which could enable any auto-feature which was already installed
  - This is the wrong behaviour, we only want to make sure that all tolerated dependencies of installed features get installed
- When doing basic resolve we would return an install list for every autofeature, even those already installed (which would usually be empty)

And a few problems which existed before this change:
- If an autofeature name was passed to resolveAsSet(), we would try to add any features which enabled that autofeature to the list of features to resolve
  - this is different to what the kernel resolver would do
  - this could result in us asking the kernel resolver to resolve a private feature, which will fail
  - I don't think there are any actual cases where this could happen anyway
- `KernelResolverRepository` wasn't preferring installed features over repository features if the repository feature was added first
  - In practise, we were always adding the installed feature first so this wasn't breaking anything, but it's a good idea to fix it in case we change the surrounding code later

Azquelt:
#build
#libby

I expect that some install tests may fail in the personal build but I expect all repository tests to pass.

pnickoll (dismissed):

Azquelt:
Review change commits squashed, code is identical.

Azquelt:
Rebased on `release`

#build
#libby

Azquelt:
#build

Azquelt:
These changes were delivered along with the install FAT test updates in #25106

PR #24671, opened 2023-03-14
Add Jakarta-ee API pom deps

cbridgha:
The major part of these changes is to compensate for the bnd subprojects that are used for our API spec jar's.
We don't use a bnd buildpath - but directly "include" the libraries in our bundles...    For these jar's we will start setting direct maven dependency properties.

Also included perf improvement to only grab the specific jar entry for the pom.xml, and stop traversing the entire contents...


cbridgha:
#build

cbridgha:
#build

cbridgha:
#build

hlhoots (approved):
Seems good to me.

cbridgha:
#libby

Issue #24578, opened 2023-03-07
Application can't recover from exceptions thrown during startup


inad9300:
I am opening this ticket as a follow-up to https://github.com/OpenLiberty/ci.maven/issues/1588, as it was determined that the issue belonged to the Open Liberty runtime rather than the Maven plugin.

**Describe the bug**  

Exceptions thrown during application startup lead to the application crashing and being unable to restart even when the source of the problem is fixed. The following message is shown on any application restart attempts thereafter:

> During application startup, the application failed to start because multiple applications were found to have the name of \<my-application-name>.

**Steps to Reproduce**  

From https://github.com/OpenLiberty/ci.maven/issues/1588#issuecomment-1412725373:

> [create] this simple class and [enable] the CDI feature:
> 
> ```java
> @ApplicationScoped
> public class MyObserver {
> 
>     public void onInit(@Observes @Initialized(ApplicationScoped.class) Object bean ) {
>     	Class clazz = this.getClass();
>         // Uncomment to cause failure
>     	//clazz = null;
>     	System.out.println("From onInit(), in " + clazz.toString());    	
>     }
> }
> ```
> [...] All you'd need to do is:
>
> 1. `mvn liberty:run` # Wait for server and app to complete startup. The 'run' goal includes the "all-in-one" aspect of 'dev' (install + configure Liberty, deploy app) but it doesn't run in the "watcher" loop. It just runs the server in the foreground.
> 2. Make your app source change
> 3. `mvn compile` # watch app update
> 4. Make another source change
> 5. `mvn compile` # watch app update again
> [...]

**Expected behavior**  

The application should be able to restart after the code that throws an exception at startup is fixed.

**Diagnostic information:**  
- OpenLiberty Version: I am unable to determine. Actually, I would like help to do so. I believe the Maven plugin keeps it up to date automatically, but I don't know how to control this or how to check what's the current version of Open Liberty that I'm using.
- Affected feature(s): possibly all, don't know specifically
- Java Version: I'm using Docker image `eclipse-temurin:17.0.4.1_1-jdk-jammy`
- server.xml configuration: Please find it in https://github.com/OpenLiberty/ci.maven/issues/1587#issuecomment-1242933252

pmd1nh:
@inad9300 Hi,  can you please collect a trace so I can debug this problem?
In server.xml set,
`
[...]

PR #24723, opened 2023-03-17
Clear application metadata on init failure

brenthdaniel:


brenthdaniel:
#build

brenthdaniel:
@hutchig I think you maintain grpc? The changes here were needed because we are now fully uninstalling applications that fail to start. In the case where a GRPC service fails to start because of a duplicate service name it was removing the legitimate service with the same name from the grpc runtime. I fixed that by scoping the services by the j2eeName, but if you prefer another solution let me know.

hlhoots (approved):
LGTM

brenthdaniel:
#build

hutchig (approved):
Looks good to me, and thank you for the test material too.

brenthdaniel:
#run-libby-bot

brenthdaniel:
Fix for #24578

Issue #24590, opened 2023-03-07
Test Failure: io.openliberty.microprofile.metrics.internal.monitor_fat.TestEnableDisableFeaturesTest.testEDF5


fmhwong:
RTC 294706

```
testEDF5:java.lang.Throwable: 2023-02-11-10:21:17:074 There were 2 errors:
  java.io.IOException(Server returned HTTP response code: 500 for URL: https://localhost:8020/metrics?scope=vendor)
  java.lang.Exception(Errors/warnings were found in server TestEDF5 logs:
<br>[2/11/23, 10:21:11:341 PST] 0000002c com.ibm.ws.webcontainer.util.ApplicationErrorUtils           E SRVE0777E: Exception thrown by application class 'com.ibm.ws.session.http.AbstractHttpSession.isNew:218'
<br>[2/11/23, 10:21:11:572 PST] 0000002c com.ibm.ws.webcontainer.webapp                               E SRVE0315E: An exception occurred: java.lang.Throwable: java.lang.IllegalStateException: The method is called on an invalidated session: 58W_3izBZhmfEjRlEGi-Qg2)
at componenttest.custom.junit.runner.FATRunner.newThrowableWithTimeStamp(FATRunner.java:338)
at componenttest.custom.junit.runner.FATRunner.access$700(FATRunner.java:70)
at componenttest.custom.junit.runner.FATRunner$1.evaluate(FATRunner.java:304)
at componenttest.custom.junit.runner.FATRunner$2.evaluate(FATRunner.java:364)
at componenttest.custom.junit.runner.FATRunner.run(FATRunner.java:178)
  java.io.IOException(Server returned HTTP response code: 500 for URL: https://localhost:8020/metrics?scope=vendor)
  java.lang.Exception(Errors/warnings were found in server TestEDF5 logs:
<br>[2/11/23, 10:21:11:341 PST] 0000002c com.ibm.ws.webcontainer.util.ApplicationErrorUtils           E SRVE0777E: Exception thrown by application class 'com.ibm.ws.session.http.AbstractHttpSession.isNew:218'
<br>[2/11/23, 10:21:11:572 PST] 0000002c com.ibm.ws.webcontainer.webapp                               E SRVE0315E: An exception occurred: java.lang.Throwable: java.lang.IllegalStateException: The method is called on an invalidated session: 58W_3izBZhmfEjRlEGi-Qg2)
[...]

PR #24760, opened 2023-03-21
 Debug if a request with session header is sent to /metrics

Channyboy:
#build

Debugging the metrics servlet to see if the request is passing in a sessions request.
This can potentially explain the failure as the FAT test is programmatically connecting to the metrics endpoint with no stateful information. However, this may bring up questions regarding about what is issuing the request.

Debug for #24590


fmhwong (approved):
LGTM

PR #24769, opened 2023-03-22
move JPA entities to Data JPA tests

njr-11:
Refactor the separation between the io.openliberty.data.internal_fat and io.openliberty.data.internal_fat_jpa test buckets such that the former does not include JPA entity annotations and can run with Jakarta Data plus JDBC only whereas the latter has JPA entity annotations and runs with Jakarta Data plus Jakarta Persistence.


njr-11:
#build

KyleAure (approved):

PR #24803, opened 2023-03-24
switch from data provider to CDI extension step

njr-11:
Align with spec update to have each Jakarta Data provider plug in via its own CDI extension, rather than a provider model with central handling from a single CDI extension.


njr-11:
#build

mswatosh (dismissed):

njr-11:
#build
#libby

mswatosh (approved):

Issue #23947, opened 2023-01-17
Deadlock reported when creating SIP Listening endpoint


TiborBeres:
Deadlock reported when creating SIP Listening endpoint

PR #24832, opened 2023-03-27
remove synchronized from createSIPListenningConnection method to prevent deadlock

TiborBeres:
Custumer has reported a deadlock when starting up sipcontainer where the sipcontainer and the channel framework both placed a lock on the thread. 

Converting createSIPListenningConnection from synchronized to not syncronized should prevent this deadlock.


TiborBeres:
Fixes #23947

TiborBeres:
pull request for #23947

TiborBeres:
#build

epj (approved):
okay since CF synchronizes at its level it should be safe to remove the sync at SIP level

PR #24862, opened 2023-03-28
Disable CRIU HW breakpoints for restore

ymanton:
HW breakpoints allow CRIU to stop process threads at arbitrary locations but are slower to initialize. The first thread that is stopped will take much longer than subsequent threads, due to HW and kernel setup. Setting CRIU_FAULT=130 disables HW breakpoints on x86 and ARM and falls back to stopping threads at syscall transitions instead.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -867,6 +867,9 @@ criuEnvDefaults() {
   fi
   export GLIBC_TUNABLES
   export LD_BIND_NOT=on
+  # Disable CRIU's use of HW breakpoints on restore.
+  # HW breakpoints are slower in some environments.
+  export CRIU_FAULT=130 # 130=FI_NO_BREAKPOINTS
 }
 
 ##


ymanton:
For reference, the change in CRIU is https://github.com/checkpoint-restore/criu/pull/1957.

tjwatson:
#build

tjwatson (approved):

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -867,6 +867,9 @@ criuEnvDefaults() {
   fi
   export GLIBC_TUNABLES
   export LD_BIND_NOT=on
+  # Disable CRIU's use of HW breakpoints on restore.
+  # HW breakpoints are slower in some environments.
+  export CRIU_FAULT=130 # 130=FI_NO_BREAKPOINTS
```
tjwatson, line 872: Should we be conditionally setting this in case there is some scenario where they want to set it to something else in the environment?

ymanton: For our current purposes only meaningful states for this var are 1) to be set to `130` (to disable HW breakpoints) and 2) to be unset (defined and empty will be treated as an error by CRIU) to get the default behaviour.

(The variable itself is overloaded, so the other possible numeric values control things unrelated to HW breakpoints.)

If we want someone to be able to override our intention they'll have to be able to tell us that they want `CRIU_FAULT` unset (i.e. HW breakpoints enabled), so we'll need a new variable (e.g. `UNSET_CRIU_FAULT`) since tests like `-z` and `-n` don't distinguish between empty and unset.

We could use a Bash-specific test like `[ -v CRIU_FAULT ]` to avoid the need for a separate variable, but `-v` is Bash-specific and recent, so maybe not for this script?

tjwatson: I think we will simply hardcode the value for now.  We can always enhance it later to allow others to define values, if needed.

tjwatson:
The build tjwatson-24873-20230329-1637
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_-t_ONs5FEe2yDpyHOcT1OQ
completed successfully!

Issue #24534, opened 2023-03-02
Add an info message when the use of javax common annotations are detected with EE9/10?


tkburroughs:
The Java EE Common Annotations APIs are included in Java SE through Java 8 and can easily be imported into class files instead of the Jakarta equivalent.  Customers developing Jakarta applications may be confused when these annotations do not work with Jakarta features; no errors or warnings appear in the logs.

Proposal is to add an informational message (Tr.audit) like below when such annotations are detected using the annotation scanner. An informational message would cause the least disruption to an existing EE 9 application that happened to have one or more of these annotation, but worked regardless of the annotation not working.

**_CWWKM0483I: The javax.annotation.Resource annotation is ignored on classes in the WrongPackageEJB module of the WrongPackageApp application. Jakarta EE applications must use the jakarta.annotation.Resource annotation. The affected classes are: io.openliberty.injection.jakarta.ejb.JakartaSingletonBean, io.openliberty.injection.jakarta.ejb.JakartaSingletonResourcesBean, io.openliberty.injection.jakarta.ejb.JakartaStatelessBean_**

This would occur at most once per annotation type per module if any of the following annotations are found:
```
javax.annotation.PostConstruct - method level
javax.annotation.PreDestroy - method level
javax.annotation.Resource - class, field, or method level
javax.annotation.Resources - class level
```

Note: the following annotations are also in this package, but will be ignored:
```
javax.annotation.Generated - not used by the Liberty runtime (retention=source)
javax.annotation.ManagedBean - never included in the JDK
javax.annotation.Priority - never included in the JDK
```

** Notes: **
These classes were removed in Java 11. 
The problem occurs when Java SE 8 is on the classpath of the compiler, but developer is writing EE9+ code (jakarta namespace)
The transformer handles this properly for existing apps.  This issue is for new code.

tkburroughs:
All work complete; closing.

PR #24865, opened 2023-03-28
Issue #24534: Remove beta guard for message CWWKM0483I

tkburroughs:
The message is now translated and may be logged.

for #24534


tkburroughs:
#build

olendvcook (approved):

PR #24866, opened 2023-03-28
add ability to run test buckets against different databases

njr-11:
Give the main Jakarta Data test bucket and the Jakarta Data JPA test bucket the ability to run against different databases.  The intention here is to enable for running locally, not to automate these buckets to run with other databases in builds yet.  Most of the combinations with other databases have at least a couple of issues that would need to be resolved before that happens.

### DIFF
--- a/dev/io.openliberty.data.internal_fat/fat/src/test/jakarta/data/DataTest.java
+++ b/dev/io.openliberty.data.internal_fat/fat/src/test/jakarta/data/DataTest.java
@@ -59,7 +59,7 @@ public class DataTest extends FATServletClient {
         server.addEnvVar("DB_PASSWORD", testContainer.getPassword());
 
         // Set up server DataSource properties
-        DatabaseContainerUtil.setupDataSourceProperties(server, testContainer);
+        DatabaseContainerUtil.setupDataSourceDatabaseProperties(server, testContainer);
 
         WebArchive war = ShrinkHelper.buildDefaultApp("DataTestApp", "test.jakarta.data.web");
         ShrinkHelper.exportAppToServer(server, war);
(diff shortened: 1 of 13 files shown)


dev/io.openliberty.data.internal_fat/fat/src/test/jakarta/data/DataTest.java
```diff
@@ -37,13 +42,24 @@
...
     @BeforeClass
     public static void setUp() throws Exception {
+        // Get driver type
+        DatabaseContainerType type = DatabaseContainerType.valueOf(testContainer);
+        server.addEnvVar("DB_DRIVER", type.getDriverName());
+        server.addEnvVar("DB_USER", testContainer.getUsername());
+        server.addEnvVar("DB_PASSWORD", testContainer.getPassword());
+
+        // Set up server DataSource properties
+        DatabaseContainerUtil.setupDataSourceProperties(server, testContainer);
```
KyleAure: There are two setup method options: 

```java
DatabaseContainerUtil.setupDataSourceProperties(server, testContainer);
DatabaseContainerUtil.setupDataSourceDatabaseProperties(server, testContainer);
```

The first one will replace your `<properties.derby.embedded ... />` with `<properties .../>`.  This is for tests that want to specifically test generic properties. 

The second one will replace your `<properties.derby.embedded .../>` with `<properties.[the database being tested] .../>`.  

Either should work, but I figured I would point it out just in case you didn't know. 
(the method names are admittedly a bit confusing)

KyleAure (changes requested):
You will need to modify the following files: 

io.openliberty.data.internal_fat/fat/src/test/jakarta/data/FATSuite.java
io.openliberty.data.internal_fat_jpa/fat/src/test/jakarta/data/jpa/FATSuite.java

To extend TestContainerSuite
```java
import componenttest.containers.TestContainerSuite;
public class FATSuite extends TestContainerSuite 
```

This will make sure that our test infrastructure properly configures docker and testcontainers when testing against remote docker hosts, or artifactory.

njr-11:
#build
#libby

KyleAure (approved):

PR #24874, opened 2023-03-29
Skip dirs on clean

jonhawkes:
#build


jonhawkes:
#build

brenthdaniel:
This change would definitely need to be considered as a design issue before being implemented. Opening up certain workarea locations to not being cleaned would be a huge change. There are many considerations -- for example, we clean the workarea when service is applied. Would we still not delete this directory in that case? For this particular use case, I think we would also want to understand why the information can't be recalculated or stored somewhere else.

jimblye:
Instead of having a list of files that cannot be deleted, maybe consider having a single top level directory in the workarea that is never cleaned.     So the log file might be here:  workarea/**durable**/rrs/tx/rmData.log
If anyone else ever needs to do this, they just store their files under workarea/**durable**

PR #24882, opened 2023-03-29
Reduce dependency on Hostname/addr in jwt builder tests

c00crane:
Our iSeries test machines seem to have multiple address/names and we're not getting the proper information at times.
The issuer is created using the host info by default and this is causing problems.
I'm going to configure an issuer value into each builder and client so that we can control the value and not have to rely upon the host lookups.


c00crane:
#libby
#build
#spawn.fufflat.buckets=com.ibm.ws.security.jwt_fat.builder

dev/com.ibm.ws.security.jwt_fat.builder/fat/src/com/ibm/ws/security/jwt/fat/builder/utils/BuilderHelpers.java
```diff
@@ -52,13 +52,13 @@ public class BuilderHelpers {
...
+        settings.put(PayloadConstants.ISSUER, "testIssuer");
         return setDefaultClaims(settings);
     }
 
     public static JSONObject setDefaultClaims(String builderId) throws Exception {
         JSONObject settings = new JSONObject();
-        settings.put(PayloadConstants.ISSUER, builderId);
+        settings.put(PayloadConstants.ISSUER, "testIssuer");
         return setDefaultClaims(settings);
     }
```
ayoho: These are now the exact same method, and neither of them need a method argument.

dev/com.ibm.ws.security.jwt_fat.builder/publish/servers/com.ibm.ws.security.jwt_fat.builder/configs/server_minimumConfig.xml
```diff
@@ -4,6 +4,7 @@
 	<featureManager>
 		<feature>jwt-1.0</feature>
 		<feature>jsonp-1.0</feature>
+		<feature>servlet-3.1</feature>
```
ayoho, line 7: Why was this needed now? I'm assuming it must be related to switching out features between EE8, EE9, and EE10?

ayoho (dismissed):

dev/com.ibm.ws.security.jwt_fat.builder/publish/servers/com.ibm.ws.security.jwt_fat.builder/configs/server_minimumConfig.xml
c00crane: yeah - we need it for EE9 and EE10 - forcing servlet 5 & 6 through the transform ends up adding it in places that we don't need it (and results in multiple copies of each)

dev/com.ibm.ws.security.jwt_fat.builder/fat/src/com/ibm/ws/security/jwt/fat/builder/utils/BuilderHelpers.java
c00crane: I'll remove the first put in another pull. (as usual, another cut/paste error)  - you'll need to re-review.

c00crane:
#libby
#build
#spawn.fufflat.buckets=com.ibm.ws.security.jwt_fat.builder

ayoho (approved):

Issue #20132, opened 2022-02-14
Update JMS Client to use Netty


isaacrivriv:
Go through the `com.ibm.ws.messaging.comms.client` bundle to switch channelfw usage to Netty

pnicolucci:
This has been implemented, closing.

Issue #20131, opened 2022-02-14
Update JMS Server to use Netty


isaacrivriv:
Go through the `com.ibm.ws.messaging.comms.server` bundle to switch channelfw usage to Netty

pnicolucci:
This work is implemented, closing

PR #24885, opened 2023-03-29
Netty jms update

### LINKED ISSUE #20134: Work on Netty JMS prototype
Transform JMS bundles necessary to build a functional prototype of JMS client and server using Netty

isaacrivriv:
for #20132
for #20131

### DIFF
--- a/dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/framework/NetworkConnectionFactory.java
+++ b/dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/framework/NetworkConnectionFactory.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2003, 2006 IBM Corporation and others.
+ * Copyright (c) 2003, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -11,8 +11,6 @@
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
 package com.ibm.ws.sib.jfapchannel.framework;
-
-import com.ibm.wsspi.channelfw.VirtualConnectionFactory;
 
 /**
  * A factory for implementations of the NetworkConnection interface.
@@ -42,10 +40,11 @@ public interface NetworkConnectionFactory
      * @throws FrameworkException if the connection cannot be created.
      */
     NetworkConnection createConnection() throws FrameworkException;
-
+    
     /**
-     * 
-     * @return an VirtualConnectionFactory
+     * Destroys the connection object stored
+     * @return if connection was successfully destroyed
      */
-    VirtualConnectionFactory getOutboundVirtualConFactory();
+    void destroy() throws FrameworkException;
+    
 }
--- a/dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/framework/NetworkConnectionFactory.java
+++ b/dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/framework/NetworkConnectionFactory.java
@@ -42,8 +42,8 @@ public interface NetworkConnectionFactory
     NetworkConnection createConnection() throws FrameworkException;
     
     /**
-     * Destroys the connection object stored
-     * @return if connection was successfully destroyed
+     * Destroys objects related to starting connections.
+     * @throws FrameworkException if objects were NOT successfully destroyed
      */
     void destroy() throws FrameworkException;
     
--- a/dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
+++ b/dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
@@ -60,7 +60,8 @@ import io.openliberty.netty.internal.tls.NettyTlsProvider;
 @Component(
         configurationPid = "com.ibm.ws.messaging.comms.wasJmsOutbound",
         configurationPolicy = REQUIRE,
-        property = "service.vendor=IBM")
+        property = { "nettyTlsProvider.cardinality.minimum=1", "service.vendor=IBM" }
+)
 public class CommsOutboundChain implements ApplicationPrereq {
     private static final TraceComponent tc = Tr.register(CommsOutboundChain.class, JFapChannelConstants.MSG_GROUP, JFapChannelConstants.MSG_BUNDLE);
     private static final TraceNLS nls = TraceNLS.getTraceNLS(JFapChannelConstants.MSG_BUNDLE);
@@ -104,16 +105,19 @@ public class CommsOutboundChain implements ApplicationPrereq {
             ChannelConfiguration tcpOptions,
             @Reference(name="commsClientService")
             CommsClientServiceFacade commsClientService,
+            @Reference(name = "nettyTlsProvider", cardinality = OPTIONAL)
+            NettyTlsProvider nettyTlsProvider,
             /* Require SingletonsReady so that we will wait for it to ensure its availability at least until the chain is deactivated. */ 
             @Reference(name="singletonsReady")
             SingletonsReady singletonsReady,
             Map<Object, Object> properties) {
 
         if (isAnyTracingEnabled() && tc.isEntryEnabled())
-            entry(this, tc, "<init>", tcpOptions, commsClientService, properties);
+            entry(this, tc, "<init>", tcpOptions, commsClientService, nettyTlsProvider, properties);
 
         this.tcpOptions = tcpOptions;
         this.commsClientService = commsClientService;
+        this.nettyTlsProvider = nettyTlsProvider;
 
         isSecureChain = MetatypeUtils.parseBoolean(OUTBOUND_CHAIN_CONFIG_ALIAS, "useSSL", properties.get("useSSL"), false);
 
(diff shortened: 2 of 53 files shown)


isaacrivriv:
#build

dmuelle:
I don't see any messages that need ID review in this PR. Adding `ID review not needed` label. If that's not correct let me know. Thanks

posmith:

isaacrivriv:
#build

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/framework/NetworkConnectionFactory.java
```diff
@@ -42,10 +40,11 @@ public interface NetworkConnectionFactory
      * @throws FrameworkException if the connection cannot be created.
      */
     NetworkConnection createConnection() throws FrameworkException;
-
+    
     /**
-     * 
-     * @return an VirtualConnectionFactory
+     * Destroys the connection object stored
+     * @return if connection was successfully destroyed
```
joe-chacko: What happens if the connection was not successfully destroyed?
Maybe add an `@throws` line.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
```diff
@@ -127,89 +159,130 @@ void unbindSecureFacet(OutboundSecureFacet oldFacet) {
...
 
-    private Map<String, Object> getTcpOptions() {
+    public Map<String, Object> getTcpOptions() {
         if (isAnyTracingEnabled() && tc.isEntryEnabled()) Tr.entry(this, tc, "getTcpOptions");
         Map<String, Object> tcpProps = tcpOptions.getConfiguration();
         if (isAnyTracingEnabled() && tc.isEntryEnabled()) exit(this, tc, "getTcpOptions", tcpProps);
         return tcpProps;
     }
 
+    @Reference(name = "nettyTlsProvider", cardinality = OPTIONAL, policyOption = GREEDY, unbind = "unbindTlsProviderService")
```
joe-chacko: This reference should not be dynamic: it belongs in the constructor parameter list.
Bind/unbind are unnecessary complications and extra code.
Also, we should use metatype to ensure we wait for this when it is expected.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
```diff
@@ -91,12 +116,19 @@ public CommsOutboundChain(
         this.commsClientService = commsClientService;
 
         isSecureChain = MetatypeUtils.parseBoolean(OUTBOUND_CHAIN_CONFIG_ALIAS, "useSSL", properties.get("useSSL"), false);
+        
+        useNettyTransport = ProductInfo.getBetaEdition() ?
```
joe-chacko: A trigraph operator does not make this clearer. Use 
`Product.getBetaEdition() && MetatypeUtils.parseBoolean(OUTBOUND_CHAIN_CONFIG_ALIAS, "useNettyTransport", properties.get("useNettyTransport"), true)` instead.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
```diff
@@ -127,89 +159,130 @@ void unbindSecureFacet(OutboundSecureFacet oldFacet) {
...
+        if (isAnyTracingEnabled() && tc.isEntryEnabled()) exit(this, tc, "getNettyTlsProvider", nettyTlsProvider);
+        return nettyTlsProvider;
+    }
+
     private synchronized void createBasicJFAPChain() {
         if (isAnyTracingEnabled() && tc.isEntryEnabled()) entry(this, tc, "createBasicJFAPChain", chainName);
-        try {
-            ChannelFramework cfw = commsClientService.getChannelFramework();
-            cfw.registerFactory("JFapChannelOutbound", JFapChannelFactory.class);
+        if(!useNetty()) {
```
joe-chacko, line 190: Avoid using `if (!condition) {} else {}` — flip the condition and swap the if/else blocks. 
This should be a refactoring in your IDE.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
```diff
@@ -127,89 +159,130 @@ void unbindSecureFacet(OutboundSecureFacet oldFacet) {
...
+                // /com.ibm.ws.messaging_fat/fat/src/com/ibm/ws/messaging/fat/CommsWithSSL/CommsWithSSLTest.java
+                // Checks for the presence of this string in the trace log, in order to ascertain that the chain has been enabled.
+                if (isAnyTracingEnabled() && tc.isDebugEnabled()) debug(this, tc, "JFAP Outbound chain"+ chainName + " successfully started ");
+                chainList.put(chainName, this);
+            } catch (ChannelException | ChainException exception) {
+                if (isAnyTracingEnabled() && tc.isDebugEnabled()) debug(this, tc, "JFAP Outbound chain " + chainName + " failed to start, exception ="+exception);
+            }
+        }else {
+        	// Use Netty Framework for transport
+            chainList.put(chainName, this);
```
joe-chacko, line 213: How is this different from the similar looking code in the if block?
If they are meant to be the same, move it outside the else block.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/Connection.java
```diff
@@ -634,7 +683,11 @@ private long send(JFapByteBuffer buffer,
 
             // Proddle the write callback to get it to leap into action and start
             // sending data from the priority queue.
-            writeCompletedCallback.proddle();
+            // TODO: Question ..this makes me think about a Proddable or WriteCompletedCallback interface type abstraction (not clear if it would be better but maybe)
```
joe-chacko: Definitely create the interface and get rid of this checking and duplication.

joe-chacko (changes requested):
See comments and DMs for details.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
isaacrivriv: Sounds good, I'll make this change.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
isaacrivriv: In this case they are meant to be the same but can't move the code out of the if/else block to be common because of the try catch block while setting up channelfw which would not add to the chainlist if `ChannelException|ChainException` occurs

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/Connection.java
isaacrivriv: Will work on it.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
isaacrivriv: I also believe in part it should be in the constructor, originally tried it this way but ran into issues because the TLS bundles is loaded with the SSL feature used for creating Netty SSL configurations. So if a server is configured with just unencrypted traffic, the server will never call the constructor parameter list for the bundle and nothing will ever load. Maybe we can figure something out that'd be better for this, I'll DM you.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/framework/NetworkConnectionFactory.java
isaacrivriv: Added a throws line for this and more clarification. In this case, it's not a specific connection but the object necessary to start up connections for a factory. For chfw there is a virtual connection factory for managing connections which has its own destroy method but for Netty our bootstrap defines how connections are started so the description was not much accurate. Updated to match this.

isaacrivriv:
#libby

isaacrivriv:
#build

isaacrivriv:
#build

isaacrivriv:
#libby

joe-chacko:
#build #run-libby-bot

isaacrivriv:
#libby

isaacrivriv:
#build

isaacrivriv:
#build

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
```diff
@@ -304,4 +391,29 @@ private void removeChannel(String channelName) {
...
+	 */
+	public boolean isSecureChain() {
+	    return isSecureChain;
+	}
+	
+	/**
+	 * Get the SSL Options for this Chain
+	 * @return the OutboundSecureFacet wrapper of SSL Options for the chain
+	 */
+	public OutboundSecureFacet getSecureFacet() {
```
joe-chacko: Do we need to expose the OutboundSecureFacet here, or could we just return the SSL Options?
Returning the facet is tighter coupling because it is an internal object, whereas the SSL Options comes from another component's SPI.

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/Connection.java
```diff
@@ -108,9 +110,9 @@ public abstract class Connection implements ConnectionInterface
     protected ConversationTable conversationTable = null;
 
     // Callbacks notified when a read or write completes for this connection
-    private ConnectionWriteCompletedCallback writeCompletedCallback = null;
+    private BaseConnectionWriteCallback writeCompletedCallback = null;
```
joe-chacko: It looks like this can be final since it is only ever assigned inside a constructor.

dev/com.ibm.ws.messaging.comms.server/src/com/ibm/ws/jfap/inbound/channel/CommsServerServiceFacade.java
```diff
@@ -62,27 +64,35 @@
...
     private static final String SECURE_PORT = "wasJmsSSLPort";
     private static final String BASIC_PORT = "wasJmsPort";
     private static final String CONFIG_ALIAS = "wasJmsEndpoint";
     private static final TraceComponent tc = register(CommsServerServiceFacade.class, JFapChannelConstants.MSG_GROUP, JFapChannelConstants.MSG_BUNDLE);
     private String endpointName = null;
-
-    private final CommsInboundChain inboundBasicChain = new CommsInboundChain(this, false);
-    private final CommsInboundChain inboundSecureChain = new CommsInboundChain(this, true);
+    
+    private InboundChain inboundBasicChain;
```
joe-chacko: Can we make this final?

dev/com.ibm.ws.messaging.comms.server/src/com/ibm/ws/jfap/inbound/channel/CommsServerServiceFacade.java
```diff
@@ -62,27 +64,35 @@
...
+    private InboundChain inboundSecureChain;
```
joe-chacko: Can we make this final?

dev/com.ibm.ws.messaging.comms.server/src/com/ibm/ws/jfap/inbound/channel/InboundChain.java
```diff
@@ -0,0 +1,20 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package com.ibm.ws.jfap.inbound.channel;
+
+public interface InboundChain {
+	
+	public void enable(boolean enabled);
```
joe-chacko: don't need the word public here — it's implicit in an interface that all its members are public

joe-chacko (changes requested):

dev/com.ibm.ws.messaging.comms.server/src/com/ibm/ws/jfap/inbound/channel/InboundChain.java
isaacrivriv: Did not know this, I will make the change!

dev/com.ibm.ws.messaging.comms.client/src/com/ibm/ws/sib/jfapchannel/impl/CommsOutboundChain.java
isaacrivriv: Sounds good, made this change

isaacrivriv:
#libby

isaacrivriv:
#build

joe-chacko (approved):
Good work

isaacrivriv:
#build

isaacrivriv:
Errors in build are related to com.ibm.ws.security.utility_fat FAT tests nothing JMS related. Should be safe to merge in

PR #24910, opened 2023-03-31
experiment with multiple databases

njr-11:
Experimentation with how Jakarta Data could handle multiple databases if a String attribute were added to the Repository annotation that could be used to select a databaseStore, dataSource, DataSourceDefinition, or resource reference.


njr-11:
#build

mswatosh (approved):

PR #24911, opened 2023-03-31
Add permissions for Custom UserRegistry running as a Bell

kristip17:
For RTC 294158

Add permissions for the user and group files when running the Custom UserRegistry as a Bell.


kristip17:
#build

emilytee (approved):

PR #24927, opened 2023-04-04
Switch to posix sed command for server.env

cbridgha:
On the Turkish language on RHEL8, the existing sed command that reads the variable values from the server.env file doesn't allow all characters it should.

moving from:
sed -e 's/^\([a-zA-Z0-9_][a-zA-Z0-9_]*\)=.*/\1/' 
to
sed -e 's/^\([_[:alnum:]][_[:alnum:]]*\)=.*/\1/'


cbridgha:
#build

jimblye (approved):
This line is probably still too permissive.  It allows an environment variable to start with a number.  But that's a different problem which already exists.
if var=`safeEcho "$line" | sed -e 's/^\([_[:alnum:]][_[:alnum:]]*\)=.*/\1/'`;

PR #24928, opened 2023-04-04
resource ref bindings and UUID

njr-11:
Add resource reference binding to Jakarta Data test with resource reference.
Add coverage for UUID as the id on an unannotated entity, and update code to autodetect this as the id.


njr-11:
#build

jmstephensgit (approved):

PR #24940, opened 2023-04-05
Convert to units defined by parameter

### LINKED ISSUE #24939: `requestTiming-1.0` causes elevated (or spiking) CPU performance due to the `SlowRequestManager`
**Describe the bug**  
When using the `requestTiming-1.0` feature in OpenLiberty, the CPU usage is elevated. CPU impact correlates to CPU capacity. 

This is more obvious when a lower threshold is set for the "slow request" threshold (i.e. e.g. <= 15s. Even so, may not encounter a noticeable impact depending on CPU capacity. 

This is also more obvious if the request has a high hung threshold or if the request is hung indefinitely and can not be terminated by by the `interruptHungRequst` attribute (leading to an indefinite hang). This allows for a bigger window of oppurtunity to witness any CPU spikes/elevation.

No stack trace as the consequence is an uptick in CPU usage.

**Steps to Reproduce**  
- enabled `requestTiming-1.0` feature
- Set `slowRequestThreshold` attribute to <=10s. (impact varies with CPU capacity).
- If CPU is being monitored, elevated CPU usage until request is finished.

**Expected behavior**  
Not to experience CPU elevation/spikes.

**Diagnostic information:**  
 - OpenLiberty Version: 23.0.0.3 and below
 - Affected feature(s) : `requestTiming-1.0`
 - Java Version: _No definitive boundaries. Can apply to any Java_
 - server.xml configuration: `No special configuration, as long as slowRequestThreshold is <= 15s. This is a subjective value.

**Additional context**

Channyboy:
#build
fixes #24939

### DIFF
--- a/dev/com.ibm.ws.request.timing/src/com/ibm/ws/request/timing/queue/QueueableRequest.java
+++ b/dev/com.ibm.ws.request.timing/src/com/ibm/ws/request/timing/queue/QueueableRequest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2021 IBM Corporation and others.
+ * Copyright (c) 2019, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


Channyboy:
#build

pgunapal (approved):
LGTM

PR #24950, opened 2023-04-07
Optimize update poms

cbridgha:
Thanks to Tom for assisting in creating an efficient set of utilities that processes the built jar's, and now that it is don right before packaging, only uses 1  java process to achieve the same results with dramatic time saving during the build.


cbridgha:
#build

cbridgha:
#build

tbitonti (approved):

cbridgha:
#libby

PR #24961, opened 2023-04-10
Add classloading trace to the JPA container ModifyConfig FAT

dazavala:
For RTC-295221.  Get classloading trace to help determine whether the updated fileset reference used by the new library classloader supplies the expected classpath entries or was not injected.

Problem: During the library update (i.e. after the test replaces the `fileset` jar of the library referenced by the BELL), the BELL runtime registers the javax...PersistenceProvider service, and immediately, the persistence runtime consumes the service (i.e. triggers OSGi to invoke the service’s factory to get the service instance). Nothing strange about that behavior. The service factory attempts to load the PersistenceProvider service impl class using the new library classloader, which fails unexpectedly with a CNFE indicating the class is not visible on its classpath. Typically, a faulty or unavailable fileset reference(s) will cause a library classloader to malfunction in this manner.


dazavala:
#build

jgrassel (approved):
LGTM

dazavala:
Thanks @jgrassel.

PR #24975, opened 2023-04-11
Fix race condition with derby during tests

mswatosh:
Fix for a test bug where booting two derby instances simultaneously causes an mbean error resulting in one of the derby instances not starting, and test failures.


njr-11 (approved):

mswatosh:
#build

PR #24996, opened 2023-04-12
implement maxInUseTime and tests

tabishop:
Implement maxInUseTime for connections, and tests. Create a separate maintenance thread to handle cleaning up maxInUseTime connections so that the maint thread can run in millisecond intervals


tabishop:
#build

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -36,6 +36,9 @@ connTimeout.desc=Amount of time after which a connection request times out. A va
 maxIdleTime=Maximum idle time
 maxIdleTime.desc=Amount of time a connection can be unused or idle until it can be discarded during pool maintenance, if doing so does not reduce the pool below the minimum size. A value of -1 disables this timeout.
 
+maxInUseTime=Remove long running in use connections
```
dmuelle:
```suggestion
maxInUseTime=Remove long running in-use connections
```

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
```diff
@@ -36,6 +36,9 @@ connTimeout.desc=Amount of time after which a connection request times out. A va
...
+maxInUseTime.desc=This property is a temporary workaround for application code that does not close connections. It configures an amount of time after which pool maintenance can invoke Connection.abort on a connection that is being used by the application. Stopping connections that are in use can result in unexpected and unpredictable threading errors. The proper solution is to update the application to always close connections and not use this property. If you must configure it as a workaround, choose a value of 20 minutes or more depending on the normal application transaction times. If transaction times are high, 10 minutes or more, then double your longest running transaction time for this value. A value of -1 disables this timeout.
```
dmuelle:
```suggestion
maxInUseTime.desc=This property is a temporary workaround for application code that does not close connections. It configures an amount of time after which pool maintenance can invoke the Connection.abort method on a connection that the application is using. Stopping connections that are in use might result in unexpected and unpredictable threading errors. The proper solution is to update the application to always close connections and not use this property. If you must configure this property as a workaround, choose a value of 20 minutes or more, depending on the normal application transaction times. If transaction times are high, 10 minutes or more, then double your longest running transaction time for this value. A value of -1 disables this timeout.
```

not sure if Connection.abort is a method or property, adjust as needed. Providing a noun identifier for each code entity is helpful for translations.

dmuelle (changes requested):
ID review

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
tabishop: good suggestion

dev/com.ibm.ws.jca.cm/resources/OSGI-INF/l10n/metatype.properties
tabishop: yep, it is a method. Thanks for the suggestions.

tabishop:
#build

dmuelle (dismissed):
LGTM, thanks

tabishop:
#build

tabishop:
The latest build failures are unrelated to the changes in this PR. #libby

jmstephensgit (approved):

PR #24999, opened 2023-04-13
Issue #20368: Stop pre-creating EJBs for Checkpoint

tkburroughs:
To avoid global transactions during checkpoint, the following changes will be made to "undo" changes originally intended to improve checkpoint performance:

1 - Non-startup Singleton beans will no longer be pre-created. They will be created
    on first use, which has always been the normal behavior prior to checkpoint.
    Singleton bean postConstruct methods will use a global transaction by default.

2 - Stateless bean pool will not be automatically pre-loaded for checkpoint.
    Stateless bean postConstruct method may use a global transaction. Also,
    preloading the Stateless pools is blocked until after @Startup beans are
    created, so in the future preloading the bean pools would occur on restore.

Basically, both options are reverted to their normal behavior without checkpoint and both behaviors could still be enabled via other configuration options, though enabling them would likely cause the checkpoint to fail due to use of transactions.

for #20368


tkburroughs:
#build

dev/io.openliberty.ejbcontainer.checkpoint_fat/fat/src/io/openliberty/ejbcontainer/checkpoint/fat/tests/EjbStartCheckpointTest.java
```diff
@@ -198,6 +238,28 @@ private void assert_12_BeanMetaDataInitilzedAtStart() throws Exception {
         }
     }
 
+    private void assert_24_BeanMetaDataInitilzedAtStart() throws Exception {
```
brideck, line 241: Note for future PR: Initilzed -> Initialized in 3 method names.

brideck (approved):

tjwatson (commented):
A late LGTM.  Thanks

PR #25001, opened 2023-04-13
Automated OpenAPI UI testing with Selenium

### LINKED ISSUE #20512: Expanded OpenAPI UI FAT testing
We need to look at whether we can/should do better FAT testing for the OpenAPI UI.

Currently we have a fairly quick manual testing procedure which is run when we make changes to the UI, but that won't catch any problems caused by a change which wasn't intended to change the UI.

As we're mostly just including Swagger UI within Open Liberty, which has its own tests, we don't need to rigorously test it, but we do need to make sure that it loads correctly and displays some operations.

I do not want to remove the manual testing, as it's very hard to cover every possible deviation from the expected behaviour and particularly to catch unwanted additions to the page.

We should talk to the Admin Center team to see how they do their UI testing and determine whether it makes sense for us to use a similar approach.

Azquelt:
Conduct a basic test of the OpenAPI UI test to catch accidental breakage.

Fixes #20512


Azquelt:
@KyleAure This PR uses the Testcontainers webdriver module which uses the `testcontainers/vnc-recorder:1.2.0` image.

From the code in `ImageVerifier` it looks like we treat image names which start `testcontainers` specially. I've tried to add the `vnc-recorder` image as a "known" image so that it gets cached and we don't get a warning about the FAT using an unknown image version.

Could you review the second commit on this PR and check I've done this correctly? Thanks

Azquelt:
#build
#libby

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    /** Wait for "short" tasks which shouldn't need to wait for anything, like selecting a drop down */
+    private static final Duration SHORT_WAIT = Duration.ofSeconds(5);
+
+    public static final String APP_NAME = "app";
+
+    @Server("openapi-ui-test")
+    public static LibertyServer server;
+
+    @Rule
```
KyleAure, line 67: If you make this a `@Rule` instead of a `@ClassRule` your testcontainer will be stopped, and removed after each test.  Obviously in this case you only have one test.  But it might be worthwhile switching this to a `@ClassRule` just in case more tests are added in the future.

KyleAure (dismissed):
All the testcontainer changes look good.  Just one minor comment.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
Azquelt: I did initially have it as a `@ClassRule`, but `@Rule` seems to be the recommended way to do it for the web driver containers. If I use `@ClassRule`, the option to save a recording only for failing tests does not work.

Unfortunately the recording is pretty essential for working out what went wrong when a test fails. Maybe I should try adding a second test to check what the performance is like. If it's bad it might be preferable to always save the recording.

tevans78 (dismissed):

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
amyreit: For our tests, we don't do a recording - we take screenshots at critical points and on failure to use for debug.  (TakesScreenshot) Our tests are much larger and longer so a recording would be enormous. Curious where you saw that the @Rule is the recommended way? We are using classrule in most places.

amyreit (dismissed):
The actual test looks pretty straightforward. Hopefully using the cssSelector will not bite you. I know you don't have control over the ui to add ids, so you don't have a lot of options.

Azquelt:
In discussion after the review, I think this test needs changes to not run on Java 8.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
Azquelt: [Selenium recommends a fresh browser per test](https://www.selenium.dev/documentation/test_practices/encouraged/fresh_browser_per_test/)

All of the [testcontainers examples use `@Rule`](https://www.testcontainers.org/modules/webdriver_containers/) and it claims that creation of browser containers is "fast".

I am still a little concerned about the performance of using `@Rule` with our infrastructure. I'll be sure to keep an eye on it when I add more tests.

Taking a screenshot does seem more practical than recording every test, I will also look into that if we have performance or space issues. The recording was surprisingly small but I still wouldn't want to record passing tests.

Azquelt:
#libby
#build

dev/io.openliberty.microprofile.openapi.ui.internal_fat/bnd.bnd
```diff
@@ -17,7 +17,7 @@ src: \
 	fat/src
 
 fat.project: true
-
+fat.minimum.java.level: 11
```
tevans78, line 20: Do you need
```
javac.source: 11
javac.target: 11
```
as well ... and does setting those actually implicitly set `fat.minimum.java.level`?

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/FATSuite.java
```diff
@@ -0,0 +1,24 @@
...
+ *******************************************************************************/
+package io.openliberty.microprofile.openapi.ui.internal.fat.tests;
+
+import org.junit.runner.RunWith;
+import org.junit.runners.Suite;
+import org.junit.runners.Suite.SuiteClasses;
+
+import componenttest.containers.TestContainerSuite;
+
+@RunWith(Suite.class)
```
tevans78, line 18: Should this have `@MinimumJavaLevel(javaLevel = 11)`?

tevans78 (approved):

dev/io.openliberty.microprofile.openapi.ui.internal_fat/bnd.bnd
Azquelt: Going by the documentation at https://github.com/OpenLiberty/open-liberty/wiki/FAT-Tests#skipping-tests-based-on-java-level either is sufficient on its own.

In this case, I might expand this FAT bucket later to cover earlier versions of MP OpenAPI which are compatible with Java 8, so I opted not to set `javac.source` to ensure the code remains compatible with Java 8.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/FATSuite.java
Azquelt: Going by the documentation at https://github.com/OpenLiberty/open-liberty/wiki/FAT-Tests#skipping-tests-based-on-java-level this isn't necessary when using `fat.minimum.java.level` in the bnd.bnd.

Azquelt:
The failures in the build are not related to this change.

Issue #19029, opened 2021-10-19
PKCE does not work with openidConnectClient-1.0


SvenSuessenguth:
## Description

Implement PKCE for openidConnectClient-1.0 and oidcLogin-1.0 per https://datatracker.ietf.org/doc/html/rfc7636.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Documents

When available, add links to required feature documents. Use "N/A" to mark particular documents which are not required by the feature.

- Aha: Externally raised RFE ([Aha](https://cloud-platform.ideas.ibm.com/ideas/LIBERTY-I-75))
- UFO: https://ibm.box.com/s/o151t7o2v7j16t1hiwwmf0nm8fbi8pwk
- FTS: https://github.com/OpenLiberty/open-liberty/issues/25150
- Beta Blog: https://github.com/OpenLiberty/open-liberty/issues/25332
- GA Blog: https://github.com/OpenLiberty/open-liberty/issues/25739

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Process Overview

- [Prioritization](#prioritization)
- [Design](#design)
- [Implementation](#implementation)
- [Legal and Translation](#legal-and-translation)
- [Beta](#beta)
- [GA](#ga)
  - [Focal Point Approvals](#focal-point-approvals-complete-by-feature-complete-date)
- [Other Deliverables](#other-deliverables)

### General Instructions

The process steps occur roughly in the order as presented. Process steps occasionally overlap.

Each process step has a number of tasks which must be completed or must be marked as not applicable ("N/A").

Unless otherwise indicated, the tasks are the responsibility of the Feature Owner or a Delegate of the Feature Owner.

If you need assistance, reach out to the [OpenLiberty/release-architect](https://github.com/orgs/OpenLiberty/teams/release-architect).

**Important: Labels are used to trigger particular steps and must be added as indicated.** 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## **Prioritization** (Complete Before Development Starts)

The ([OpenLiberty/chief-architect](https://github.com/orgs/OpenLiberty/teams/chief-architect)) and area leads are responsible for prioritizing the features and determining which features are being actively worked on.

### **Prioritization**
- [x] Feature added to the "New" column of the [Open Liberty project board](https://github.com/orgs/OpenLiberty/projects/2)
[...]

PR #25004, opened 2023-04-13
Issue 19029: Initial metatype for OIDC client PKCE support

ayoho:
Adds the first pass of metatype for PKCE support for the openidConnectClient-1.0 and socialLogin-1.0 features. Attribute name and approach is subject to change, but currently designed as a `pkceCodeChallengeMethod` attribute with two options:
- `plain`
- `S256`

For #19029


ayoho:
#build

jimmy1wu (approved):

dmuelle:
it looks as though there are no actual messages here that need ID review. Adding the review not needed label. If that's incorrect just let me know.

PR #25023, opened 2023-04-17
Remove unused duplicate MP Config messages

### LINKED ISSUE #20269: Fix duplicate MP Config messages
The messages `CWMCG0301E` and `CWMCG9996E` exist in two different files and aren't identical
* `CWMCG9996E` has differring explanation text
* `CWMCG0301E` has slightly different message text, explanation, and action

tevans78:
Fixes https://github.com/OpenLiberty/open-liberty/issues/20269

There were two NLS bundles which contained CWMCG0300E, CWMCG0301E and CWMCG9996E. However, the code which references the second NLS bundle, never uses them. The first two messages can simply be removed from the second bundle but we do still need an emergency message in there. It seems likely that emergency message code should have been CWMCG9995E... so that one is updated.

#build


dmuelle:
Adding `ID review not needed` as there don't seem to be ay new messages to review here. If that's incorrect just let me know.

posmith:
L2 message review completed

pnickoll (approved):

PR #25029, opened 2023-04-17
Adding mdb to message ignore list

mswatosh:
Adding an mdb to the list of acceptable CNTR4015W messages to ignore. 

From the build results this defect was opened for, all of the tests passed, indicating that the Endpoint does come up, so ignoring this will help prevent build breaks on slow systems


mswatosh:
#build

gjwatts (approved):

PR #25040, opened 2023-04-18
ElementCollection of embeddable

njr-11:
Tests for an entity attribute that is a collection of an Embeddable type and annotated ElementCollection.
Tests for an entity attribute that is a collection but not annotated ElementCollection.


njr-11:
#build

mswatosh (approved):

PR #25064, opened 2023-04-20
Do a resolveBundles call after installing all kernel bundles

tjwatson:
This avoids having to resolve bundles while starting the kernel bundles.  There is complications with resolving bundles in parallel from the start-level.  While that should work, it is a complication that we can easily avoid and this is better behavior anyway to resolve all the kernel bundles in one call.


tjwatson:
#build

tjwatson:
I'm going to rework the fix a bit to avoid calling `resolveBundles` when all the kernel bundles are already resolved (from a cached state).

tjwatson:
#build

tjwatson:
#libby

sebratton (dismissed):

dazavala (approved):
LGTM

tjwatson:
#libby

PR #25081, opened 2023-04-21
array attribute on entity

njr-11:
Include an array attribute on an entity and define methods that query for single and multiple results of the array attribute.


njr-11:
#build

jmstephensgit (approved):

Issue #25082, opened 2023-04-23
Session timeout or invalidation stopped processing after deserialization exceptions


zhang-ca:
The session invalidation processes stopped after deserialization exceptions in JCache persistence scenarios. The deserialization exceptions are expected, because serialVersionUID is different between a SessionData class loaded in JVM and a SessionData deserialized in JCache.

However, the stop processes after deserialization exception does not happen with database persistence scenarios.

The expected result is that the session meta data and session attributes data are continue to remove from JCache after invalidation timeout exceptions to avoid data growing in JCache.

PR #25083, opened 2023-04-23
Allow invalidation processes continue after deserialization exceptions

zhang-ca:
Problem Description :
The customer found session data leak at JCache session persistent server (Redis). It seems to be caused by session timeout invalidation stopped processing after error.
The customer tried the same with DB session persistence. Problem does not happen with Db. So, this might be related to the httpSessionCache.

Problem Scenario :
　Step1. Create session A (contains object with serialVersionUID=1L)
　Step2. Update application and deploy it to the liberty (Update serialVersionUID to 2L from 1L)
　Step3. Create session B and session C. (contains object with serialVersionUID=2L)
　Step4. Wait for session time out for all A, B and C... Invalidated order is various.  
          If that order is 1 B, 2 A, 3 C, 
                B will be invalidated. 
                Invalidation of A will be failed with error. 
                C will not be invalidated because invalidation process stopped after error of A's invalidation. << The changes allow this to continue for invalidation to avoid data leak.


zhang-ca:
fixes #25082

zhang-ca:
for #25082

zhang-ca:
#build

bensonlatibm (approved):
lgtm!

PR #25084, opened 2023-04-24
handle host resolution issue on iSeries

c00crane:
Disable one more test that has problems on the iSeries systems - those systems have multiple host names/addresses.  The tests get varying names/addresses and whats gets registered isn't what the test always gets to use.  Most of our tests use localhost and we don't have a problem, but a few tests do use the real host info.  We'll just skip those few tests on iSeries since they will fail randomly.


c00crane:
#build

c00crane:
#libby
#build

teddyjtorres (approved):

PR #25103, opened 2023-04-24
2nd fix for Derby race condition

mswatosh:
The original fix synchronized the method, but because this is happening with multiple instances of DatabaseStoreImpl we need to synchronize on a static object to prevent other instances of DatabaseStoreImpl from starting Derby at the same time


njr-11 (approved):

mswatosh:
#build

PR #25048, opened 2023-04-19
Update All TestNG installs to custom 7.5.0 build

abutch3r:
Updated Arquillian related variables to improve consistency and findability across individual projects when updates are to be made.

Update arquillian versions to latest known tested versions that should be compatible with TestNG 7.5.0

7.5.0 of TestNG now includes SLF4J, so update ArquillianLoadableExtension to include SLF4J.

fixes #23781

PR #25108, opened 2023-04-25
Revert "Update All TestNG installs to custom 7.5.0 build"

abutch3r:
Reverts OpenLiberty/open-liberty#25048


websterc87 (approved):
Backout looks good to me, we need to fix these buckets running on CI Orch before redelivery

PR #25110, opened 2023-04-25
remove as alias for delete

njr-11:
This is a simple/trivial change to be more flexible. While writing tests we have often accidentally written "remove" rather than "delete", which is also supported by some vendors.


gjwatts (approved):

njr-11:
#build

PR #25114, opened 2023-04-25
custom repository interface

njr-11:
Test case that specifies a generic custom repository interface with entity type/key type parameters.  This generic interface is inherited by another repository interface that adds the entity class and key type as type parameters.


KyleAure (approved):

njr-11:
#build

njr-11:
Previous build had 1 failure which happened before any tests were attempted looks like a test framework issue because the logs show the server came up cleanly.
#build

PR #25115, opened 2023-04-25
Fix test defect in CheckpointSPITest

sebratton:
Fix a race condition on restore between two threads making changes to config hashMap.


tjwatson (approved):

sebratton:
#build

PR #25118, opened 2023-04-26
IBM API/SPI should not generate pom dependencies

cbridgha:
Although the jar's do go through a "cleanup" routine...  these jar's shouldn't create the pom dependencies in any case - they will never contain 3rd party resources...   And the pom's are extracted for the maven repo zip prior to the cleanup task - Stopping these from being created in the first place...


cbridgha:
#build

hlhoots (approved):
LGTM

PR #25126, opened 2023-04-27
Disable entry exit debug on methods

sebratton:
The methods in question are invoked via callback from the jvm running in a special single threaded mode where the entry trace is undesirable since it can lead to a deadlock.

### DIFF
--- a/dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
+++ b/dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
@@ -493,6 +493,7 @@ public class CheckpointImpl implements RuntimeUpdateListener, ServerReadyStatus 
     }
 
     @FFDCIgnore(Exception.class)
+    @Trivial
     private void callHooks(String operation, List<CheckpointHook> checkpointHooks,
                            Consumer<CheckpointHook> perform,
                            Function<Exception, CheckpointFailedException> failed) throws CheckpointFailedException {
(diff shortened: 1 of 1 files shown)


sebratton:
#build

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -507,6 +508,7 @@ private void callHooks(String operation, List<CheckpointHook> checkpointHooks,
         }
     }
 
+    @Trivial
```
tjwatson, line 511: Should also add `@Trivial` to the `restore` method.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
```diff
@@ -493,6 +493,7 @@ List<CheckpointHook> getHooks(Object[] hooks) {
     }
 
     @FFDCIgnore(Exception.class)
+    @Trivial
```
tjwatson, line 496: I believe it is good to have the `@Trivial` added here.  I will point out any deadlock we are trying to avoid with the logging framework in single-threaded mode is going to be hit by the calls to `debug` below when debug trace is enabled for `checkpoint`, which is the case for our FAT servers.  The call to `Tr.debug` will try to acquire the contended logging lock.

dev/io.openliberty.checkpoint/src/io/openliberty/checkpoint/internal/CheckpointImpl.java
sebratton: Done

tjwatson (approved):

sebratton:
#libby
#build

sebratton:
NO error/failures related to the changes in this PR.

PR #25005, opened 2023-04-13
Use mvn wrapper2

tevans78:
Redelivery of https://github.com/OpenLiberty/open-liberty/pull/24754

Plus some fixes for zos and Windows

PR #25191, opened 2023-05-03
Use Maven Wrapper to run TCK FATs

### LINKED ISSUE #24289: Update TCKs to use mvn wrapper scripts
The TCKs are currently set up to use a copy of mvn which is pre-installed on the build machine. It would be better if they used the mvnw wrapper script which downloads mvn as and when it is needed.

https://github.ibm.com/websphere/WS-CD-Open/issues/26032

Azquelt:
Fixes #24289

Builds on previous work in #25005.

- Add the maven wrapper script and properties file to the fattest.simplicity project
- When running a TCK, copy these files to the test directory and use them to launch the TCK test
- When running in CI, download the maven distribution from an internal repo instead
- Refactor some methods out of TCK Runner
- Improve the timeout implementation when running a TCK (computation of the timeout value may still need some work)

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKRunner.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKRunner.java
@@ -12,12 +12,11 @@
  *******************************************************************************/
 package componenttest.topology.utils.tck;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
-import java.io.InputStreamReader;
+import java.net.Authenticator;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
@@ -28,6 +27,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -56,6 +56,7 @@ import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
 import com.ibm.websphere.simplicity.PortType;
+import com.ibm.websphere.simplicity.ProgramOutput;
 import com.ibm.websphere.simplicity.log.Log;
 import com.ibm.ws.fat.util.Props;
 
@@ -63,6 +64,7 @@ import componenttest.custom.junit.runner.RepeatTestFilter;
 import componenttest.topology.impl.LibertyServer;
 import componenttest.topology.utils.tck.TCKResultsInfo.TCKJarInfo;
 import componenttest.topology.utils.tck.TCKResultsInfo.Type;
+import componenttest.topology.utils.tck.TCKUtilities.ArtifactoryAuthenticator;
 import junit.framework.AssertionFailedError;
 
 /**
@@ -100,9 +102,13 @@ public class TCKRunner {
     private final String suiteFileName;
     private final Map<String, String> additionalMvnProps;
     private final boolean isTestNG;
-    private final String relativeTckRunner;
     private final Type type;
     private final String specName;
+    private final File mavenUserHome;
+    private final File settingsFile;
+    private final File tckRunnerDir;
+
+    private final Authenticator artifactoryAuthenticator;
 
     /**
      * runs "mvn clean test" in the tck folder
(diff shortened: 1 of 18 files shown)


dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -207,4 +220,542 @@ public static TCKJarInfo parseTCKDependencies(Type type, String[] dependencyOutp
...
+            int exitValue = output.getExitCode();
+            Log.info(c, "zosTagASCII", "chtag RC = " + exitValue);
+
+            if (exitValue != 0) {
+                Log.info(c, "zosTagASCII", "OUTPUT:");
+                for (String line : output.getOutput()) {
+                    Log.info(c, "zosTagASCII", line);
+                }
+
+                throw new Exception("Process failure <chtag> see log for output.");
```
tevans78: Maybe include output in exception message

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -207,4 +220,542 @@ public static TCKJarInfo parseTCKDependencies(Type type, String[] dependencyOutp
...
+            }
+        } catch (Exception e) {
+            Log.error(c, "Could not tag zos file as ASCII", e);
```
tevans78: Re-throw the exception ... no way it will recover

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -207,4 +220,542 @@ public static TCKJarInfo parseTCKDependencies(Type type, String[] dependencyOutp
...
+     * and replaces the public repository URL with the artifactory URL.
+     *
+     * @param  wrapperPropertiesFile the wrapper properties file to update
+     * @return                       the new set of properties
+     * @throws IOException
+     */
+    public static Properties updateWrapperPropertiesFile(File wrapperPropertiesFile) throws IOException {
+        Properties props = new Properties();
+        //get the artifactory server
+        String artifactoryServer = System.getProperty(FAT_TEST_PREFIX + ARTIFACTORY_SERVER_KEY);
```
tevans78: Pass in server and update doc to be clear this is only run when using artifactory

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -207,4 +220,542 @@ public static TCKJarInfo parseTCKDependencies(Type type, String[] dependencyOutp
...
+    }
+
+    public static boolean useArtifactory() {
+        Log.info(c, "useArtifactory", "Force external: " + System.getProperty(FAT_TEST_PREFIX + ARTIFACTORY_FORCE_EXTERNAL_KEY, ""));
+        boolean forceExternal = Boolean.getBoolean(FAT_TEST_PREFIX + ARTIFACTORY_FORCE_EXTERNAL_KEY);
+        if (forceExternal) {
+            return false;
+        }
+
+        return !System.getProperty(FAT_TEST_PREFIX + ARTIFACTORY_SERVER_KEY, "").isEmpty();
```
tevans78: rewrite to make more readable ... single return?

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -207,4 +220,542 @@ public static TCKJarInfo parseTCKDependencies(Type type, String[] dependencyOutp
...
+        return System.getProperty(FAT_TEST_PREFIX + ARTIFACTORY_SERVER_KEY);
+    }
+
+    public static String getArtifactoryUser() {
+        return System.getProperty(FAT_TEST_PREFIX + ARTIFACTORY_USER_KEY);
+    }
+
+    public static String getArtifactoryToken() {
+        return System.getProperty(FAT_TEST_PREFIX + ARTIFACTORY_TOKEN_KEY);
+    }
```
tevans78: Use these methods where appropriate

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -207,4 +220,542 @@ public static TCKJarInfo parseTCKDependencies(Type type, String[] dependencyOutp
...
+
+    /**
+     * Get the location within the dev directory to create a temporary maven home directory.
+     *
+     * @return             {@code .m2} within the dev directory
+     * @throws IOException
+     */
+    public static File getTemporaryMavenHomeDir() throws IOException {
+        // user.dir is set to the test project directory when running a FAT
+        String userDir = System.getProperty("user.dir");
```
tevans78: consider using a different starting point

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -207,4 +220,542 @@ public static TCKJarInfo parseTCKDependencies(Type type, String[] dependencyOutp
...
+     *
+     * @param  directory   the file to check
+     * @throws IOException if the file does not exist or is not a directory
+     */
+    public static void requireDirectory(File directory) throws IOException {
+        requireExists(directory);
+        if (!directory.isDirectory()) {
+            throw new IOException(directory + " is not a directory");
+        }
+    }
```
tevans78, line 770: Use this method where appropriate

tevans78 (dismissed):
No major changes required ... small issues can be done as a follow up.

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
Azquelt: On re-review I think this is actually correct, `user.dir` is the current working directory.

However, `File` will automatically resolve against this directory when an absolute file path is needed so I'm removing the manual system property lookup and adding some validation of the final directory.

Azquelt:
#libby
#build

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKRunner.java
```diff
@@ -880,71 +911,45 @@ private static String jarPathInDir(String jarNameFragment, String dir) {
...
-            int numOfLinesToInclude = 7; // We will include the last 7 lines of the output file in the timeout message
-            int lineNum = output.size() - numOfLinesToInclude;
-            if (lineNum < 0) {
-                lineNum = 0;
-            }
-            while (lineNum < output.size()) {
-                lastLines.add(output.get(lineNum));
-                lineNum++;
+        if (mavenUserHome != null) {
+            String artifactoryServer = TCKUtilities.getArtifactoryServer();
```
tevans78, line 921: I think we looked at this before but should there be a `TCKUtilities.useArtifactory()` guard around these blocks?

tevans78: `mavenUserHome` will always be null if `useArtifactory()` returned false ... so the code is correct, just perhaps needs a comment

tevans78 (approved):

Azquelt:
Failures in the build are not related to this issue.

PR #25218, opened 2023-05-05
Ensure the loader has access to checkpoint before transforming

tjwatson:


tjwatson:
#build

dazavala (approved):
LGTM. No chance of classloader being null; new noClassFound() look w.r.t. arguments and exception handling+ffdc.

PR #25240, opened 2023-05-09
Rework the mapCache fat

sebratton:
The fat was subject to intermittent failures due to the distributed map beingsetup prior to restore. We consider the timeout of values in the map to be indeterminate in that scenario and so have moved the distributed map setup after the restore, making the test a simple sannity check on the map behavior after checkpoint/restore.


tjwatson (dismissed):

sebratton:
#build

sebratton:
#libby

sebratton:
#build

sebratton:
#build

tjwatson (approved):

sebratton:
#libby

PR #25241, opened 2023-05-09
Check for errors in log after a server checkpoint

sebratton:
Check for errors in log after a server checkpoint. Fail test if errors found. Allow exceptions set of string.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -513,25 +513,30 @@ public class LibertyServer implements LogMonitorClient {
         checkpointInfo.criuRestoreDisableRecovery = value;
     }
 
+    public LibertyServer addCheckpointRegexIgnoreMessage(String regEx) {
+        checkpointInfo.checkpointRegexIgnoreMessages.add(regEx);
+        return this;
+    }
+
     public static class CheckpointInfo {
-        final Consumer<LibertyServer> defaultCheckpointLambda = (LibertyServer s) -> {
-            Log.debug(c, "No beforeCheckpointLambda supplied.");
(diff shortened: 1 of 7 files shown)


sebratton:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -513,48 +513,53 @@ public void setCriuRestoreDisableRecovery(boolean value) {
         checkpointInfo.criuRestoreDisableRecovery = value;
     }
 
+    public LibertyServer addCheckpointRegexIgnoreMessage(String regEx) {
```
tjwatson, line 516: Why is this method not directly on `CheckpointInfo` like `setAssertNoAppRestartOnRestore`?

sebratton: No reason. We have a mix of APIs now to affect the behavior at checkpoint. Some are on the actual checkpointInfo object, most recently others have been added on the server object e.g. `public boolean isCriuRestoreDisableRecovery()`. If you'd like to move them all into the checkpoint object then I'l like to do that in a follow on fix. Perhaps using that to move checkpoint info into it's own class to cut down on the gargantuan size of LibertyServer

tjwatson: I prefer we keep the logic and method in `CheckpointInfo` for these things.  I am fine with that happening in a separate PR

sebratton:
#Libby 
#build

sebratton:
#Libby

sebratton:
Build failure due to known issue unrelated to this PR.

tjwatson (approved):

sebratton:
#libbybot

Issue #25072, opened 2023-04-21
Make the UI endpoint location configurable


Azquelt:
Make it so that the OpenAPI UI endpoint can be configured with the `mp.openapi.extensions.liberty.ui.path` MP Config property.

Add a FAT test to ensure that the UI page can be reached at the configured path. Ensure this test is repeated for all versions of `mpOpenAPI` in full mode.

Add a FAT test to ensure there's a log message telling you where the endpoint is located. (I think we get a message anyway, so you should just have to write a test to check the message is there).

Make sure all functionality is gated behind a call to `ProductInfo.getBetaEdition()`. Include a test that the endpoint config property is not respected if we're not running in beta mode.

Since this functionality would appear automatically in the next beta release, no code may be delivered until #25011 has been approved.

PR #25261, opened 2023-05-12
Enable User customization of OpenAPI UI Endpoint

abutch3r:
Allow user to define a property for the MP OpenAPI UI Endpoint such that the UI is available on that path e.g. `/custom/location` 

Given package is used by all OpenAPI feature versions, it will always default to `/openapi/ui` and is currently beta guarded

includes validation to prevent conflict with the OpenAPI 

Expands UI testing to include test for updating of UI - runs the same tests against the UI as the basic test and basic test covers the reverting to default 

#25072

### DIFF
--- a/dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
+++ b/dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
@@ -37,8 +37,8 @@ CWWKO1753W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT.explanation=The OpenAPI UI path w
 CWWKO1753W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT.useraction=Specify a different path for the OpenAPI UI path.
 
 CWWKO1754W_OPEN_API_UI_PATH_INVALID=CWWKO1754W: The OpenAPI UI path is invalid or contains invalid characters. The URL must be constructed using Unicode alphanumeric characters A-Za-z0-9, the underscore (_), dash (-), forward slash (/) and period (.).
-CWWKO1754W_OPEN_API_UI_PATH_INVALID.explanation=The value assigned to the OpenAPI UI path must not include any characters that are considered invalid.
-CWWKO1754W_OPEN_API_UI_PATH_INVALID.useraction=The OpenAPI UI path cannot be set to a value that contains invalid characters.
+CWWKO1754W_OPEN_API_UI_PATH_INVALID.explanation=The OpenAPI UI Path cannot be set to a value that contains invalid characters.
+CWWKO1754W_OPEN_API_UI_PATH_INVALID.useraction=Ensure the OpenAPI Path contains only A-Za-z0-9, _, -, ., /.
 
 CWWKO1755W_OPEN_API_UI_PATH_SEGMENT_INVALID=CWWKO1755W: OpenAPI UI path segments "/." or "/.." are invalid.
 CWWKO1755W_OPEN_API_UI_PATH_SEGMENT_INVALID.explanation=OpenAPI UI path segments "/." or "/.." are invalid.
--- a/dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
+++ b/dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
@@ -1,0 +1,45 @@
+###############################################################################
+# Copyright (c) 2023 IBM Corporation and others.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License 2.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
+###############################################################################
+# -------------------------------------------------------------------------------------------------
+#CMVCPATHNAME N/A
+#COMPONENTPREFIX CWWKO
+#COMPONENTNAMEFOR Open API UI
+#ISMESSAGEFILE TRUE
+#NLS_MESSAGEFORMAT_NONE
+#NLS_ENCODING=UNICODE
+# -------------------------------------------------------------------------------------------------
+
+# Message prefix block: CWWKO1750 - CWWKO1799
+CWWKO1750I_OPEN_API_PATH_UPDATE=CWWKO1750I: The OpenAPI UI path has been set to {0}.
+CWWKO1750I_OPEN_API_PATH_UPDATE.explanation=The OpenAPI UI path has been set to a value that is not the default.
+CWWKO1750I_OPEN_API_PATH_UPDATE.useraction=No action is required.
+
+CWWKO1751W_OPEN_API_PATH_UPDATE_FAILED=CWWKO1751W: The OpenAPI UI path is invalid, the default value will be used instead. Check previous warnings for cause of validation failure.
+CWWKO1751W_OPEN_API_PATH_UPDATE_FAILED.explanation= The OpenAPI UI path is invalid and will use the default value instead.
+CWWKO1751W_OPEN_API_PATH_UPDATE_FAILED.useraction=Check earlier warning messages for causes of validation failures.
+
+CWWKO1752W_OPEN_API_PATH_CONFLICT=CWWKO1752W: The OpenAPI UI path cannot be set to {0}. This path is used to retrieve the OpenAPI documents.
+CWWKO1752W_OPEN_API_PATH_CONFLICT.explanation=The OpenAPI document and OpenAPI UI paths cannot be the same.
+CWWKO1752W_OPEN_API_PATH_CONFLICT.useraction=Specify a unique path for the OpenAPI UI.
+
+CWWKO1753W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT=CWWKO1753W: The OpenAPI UI path was set to the same path as the {0} feature, which might prevent startup of itself or the other feature.
+CWWKO1753W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT.explanation=The OpenAPI UI path was set to the same path as another feature, which can prevent startup or access.
+CWWKO1753W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT.useraction=Specify a different path for the OpenAPI UI path.
+
+CWWKO1754W_OPEN_API_UI_PATH_INVALID=CWWKO1754W: The OpenAPI UI path is invalid or contains invalid characters. The URL must be constructed using Unicode alphanumeric characters A-Za-z0-9, the underscore (_), dash (-), forward slash (/) and period (.).
+CWWKO1754W_OPEN_API_UI_PATH_INVALID.explanation=The value assigned to the OpenAPI UI path must not include any characters that are considered invalid.
+CWWKO1754W_OPEN_API_UI_PATH_INVALID.useraction=The OpenAPI UI path cannot be set to a value that contains invalid characters.
+
+CWWKO1755W_OPEN_API_UI_PATH_SEGMENT_INVALID=CWWKO1755W: OpenAPI UI path segments "/." or "/.." are invalid.
+CWWKO1755W_OPEN_API_UI_PATH_SEGMENT_INVALID.explanation=OpenAPI UI path segments "/." or "/.." are invalid.
+CWWKO1755W_OPEN_API_UI_PATH_SEGMENT_INVALID.useraction=Remove any path "/." or "/.." segments from the path.
--- a/dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
+++ b/dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
@@ -56,9 +56,9 @@ public class OpenAPIUIEndpointManager {
         uiPath = OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH;
 
         if (ProductInfo.getBetaEdition()) {
-            Config config = ConfigProvider.getConfig(OpenAPIUIEndpointManager.class.getClassLoader());
             //check for system property `open_api_path_enabled` as additional guide - getBoolean returns `true` if the value exists and is set to `true`, if the value is `false`
             if (Boolean.getBoolean("open_api_path_enabled")) {
+                Config config = ConfigProvider.getConfig(OpenAPIUIEndpointManager.class.getClassLoader());
                 resolvePathFromConfig(config);
             }
         }
--- a/dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
+++ b/dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
@@ -48,7 +48,7 @@ public class OpenAPIUIEndpointManager {
     private String uiPath = null;
     private String dataPath = null;
 
-    private OpenAPIWABConfigManager openAPIWABConfigManager;
+    private WABConfigManager wabConfigManager;
 
     @Activate
     protected void activate(ComponentContext context, Map<String, Object> properties) {
@@ -57,23 +57,18 @@ public class OpenAPIUIEndpointManager {
 
         if (ProductInfo.getBetaEdition()) {
             Config config = ConfigProvider.getConfig(OpenAPIUIEndpointManager.class.getClassLoader());
-            if (config.getOptionalValue("open_api_path_enabled", Boolean.class).orElse(false)) {
+            //check for system property `open_api_path_enabled` as additional guide - getBoolean returns `true` if the value exists and is set to `true`, if the value is `false`
+            if (Boolean.getBoolean("open_api_path_enabled")) {
                 resolvePathFromConfig(config);
-                if (!validatePath(uiPath, dataPath)) {
-                    uiPath = dataPath + OPEN_API_UI_PATH;
-                    Tr.warning(tc,"CWWKO1751W_OPEN_API_PATH_UPDATE_FAILED");
-                } else if(!uiPath.equals(OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH)){
-                    Tr.info(tc,"CWWKO1750I_OPEN_API_PATH_UPDATE",uiPath);
-                }
             }
         }
-        openAPIWABConfigManager = new OpenAPIWABConfigManager(context, OPEN_API_UI_VAR_NAME, uiPath, "OpenAPI UI");
-        openAPIWABConfigManager.activate();
+        wabConfigManager = new WABConfigManager(context, OPEN_API_UI_VAR_NAME, uiPath, "OpenAPI UI");
+        wabConfigManager.register();
     }
 
     @Deactivate
     protected void deactivate(ComponentContext context, int reason) {
-        openAPIWABConfigManager.deactivate();
+        wabConfigManager.unregister();
     }
 
     /**
@@ -148,6 +143,12 @@ public class OpenAPIUIEndpointManager {
         //Process dataPath first as the value is used by UI Path if value does not exist
         dataPath = resolvePath(config.getOptionalValue(OPEN_API_ENDPOINT_CONFIG_NAME, String.class).orElse(OPEN_API_ENDPOINT_PATH));
         uiPath = resolvePath(config.getOptionalValue(OPEN_API_UI_CONFIG_NAME, String.class).orElse(dataPath + OPEN_API_UI_PATH));
+        if (!validatePath(uiPath, dataPath)) {
+            uiPath = dataPath + OPEN_API_UI_PATH;
+            Tr.warning(tc,"CWWKO1751W_OPEN_API_PATH_UPDATE_FAILED");
+        } else if(!uiPath.equals(OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH)){
+            Tr.info(tc,"CWWKO1750I_OPEN_API_PATH_UPDATE",uiPath);
+        }
     }
 
 }
--- a/dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
+++ b/dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
@@ -1,0 +1,153 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ******************************************************************************/
+
+package com.ibm.ws.microprofile.openapi.ui;
+
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.ws.kernel.productinfo.ProductInfo;
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+import org.osgi.service.component.annotations.Deactivate;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE, immediate = true, property = { "service.vendor=IBM" })
+public class OpenAPIUIEndpointManager {
+
+    private static final TraceComponent tc = Tr.register(OpenAPIUIEndpointManager.class);
+
+    private static final String OPEN_API_ENDPOINT_CONFIG_NAME = "mp.openapi.extensions.liberty.path";
+    private static final String OPEN_API_ENDPOINT_PATH = "/openapi";
+
+    private static final String OPEN_API_UI_VAR_NAME = "openAPIUIPATH";
+    private static final String OPEN_API_UI_CONFIG_NAME = "mp.openapi.extensions.liberty.ui.path";
+    private static final String OPEN_API_UI_PATH = "/ui";
+
+    private static final Pattern PATH_PATTERN = Pattern.compile("^(/[\\w./_-]*)?$");
+    private static final Map<String, String> CONFLICTING_PATHS_MAP = Stream.of(
+                                                                               new AbstractMap.SimpleEntry<>("/ibm/api", "OpenAPI"),
+                                                                               new AbstractMap.SimpleEntry<>("/health", "mpHealth"),
+                                                                               new AbstractMap.SimpleEntry<>("/metrics", "mpMetrics")).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    private String uiPath = null;
+    private String dataPath = null;
+
+    private OpenAPIWABConfigManager openAPIWABConfigManager;
+
+    @Activate
+    protected void activate(ComponentContext context, Map<String, Object> properties) {
+        //Default UI Path is /openapi/ui
+        uiPath = OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH;
+
+        if (ProductInfo.getBetaEdition()) {
+            Config config = ConfigProvider.getConfig(OpenAPIUIEndpointManager.class.getClassLoader());
+            if (config.getOptionalValue("open_api_path_enabled", Boolean.class).orElse(false)) {
+                resolvePathFromConfig(config);
+                if (!validatePath(uiPath, dataPath)) {
+                    uiPath = dataPath + OPEN_API_UI_PATH;
+                    Tr.warning(tc,"CWWKO1751W_OPEN_API_PATH_UPDATE_FAILED");
+                } else if(!uiPath.equals(OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH)){
+                    Tr.info(tc,"CWWKO1750I_OPEN_API_PATH_UPDATE",uiPath);
+                }
+            }
+        }
+        openAPIWABConfigManager = new OpenAPIWABConfigManager(context, OPEN_API_UI_VAR_NAME, uiPath, "OpenAPI UI");
+        openAPIWABConfigManager.activate();
+    }
+
+    @Deactivate
+    protected void deactivate(ComponentContext context, int reason) {
+        openAPIWABConfigManager.deactivate();
+    }
+
+    /**
+     * Validate the OpenAPI path to ensure it is Valid, contains only valid characters and mets certain rules.
+     *
+     * Reserved Path is a related entitiy where the path and the reserved path cannot be the same*
+     *
+     * @param path
+     * @param reservedPath
+     * @return
+     */
+    public static boolean validatePath(String path, String reservedPath) {
+        boolean valid = true;
+        try {
+            //temporarily construct a URL using provided path to use the URL validation code.
+            new URL("https://localhost" + path);
+        } catch (MalformedURLException e) {
+            Tr.warning(tc, "CWWKO1754W_OPEN_API_UI_PATH_INVALID",e.getLocalizedMessage());
+            return false;
+        }
+        if (path.equals(reservedPath)) {
+            Tr.warning(tc, "CWWKO1752W_OPEN_API_PATH_CONFLICT", path);
+            valid = false;
+        }
+        //check that path is made up of valid characters
+        if (!PATH_PATTERN.matcher(path).matches()) {
+            Tr.warning(tc, "CWWKO1754W_OPEN_API_UI_PATH_INVALID", path);
+            valid = false;
+            //check against other feature paths to see if a potential conflict might occur
+        }
+
+        // check that no segment is just `/.` or `/..`, `/.....` should be left as is as it has no special meaning
+        ArrayList<String> segments = new ArrayList<>(Arrays.asList(path.split("/")));
+        if (segments.contains(".") || segments.contains("..")) {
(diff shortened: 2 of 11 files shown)


abutch3r:
#libby 
#build

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
```diff
@@ -0,0 +1,34 @@
...
+#COMPONENTPREFIX CWWKO
+#COMPONENTNAMEFOR Open API UI
+#ISMESSAGEFILE TRUE
+#NLS_MESSAGEFORMAT_NONE
+#NLS_ENCODING=UNICODE
+# -------------------------------------------------------------------------------------------------
+
+# Message prefix block: CWWKO1750 - CWWKO1799
+
+CWWKO1750W_OPEN_API_UI_PATH_RESERVED=CWWKO1750W: The OpenAPI UI Path cannot be set to {0}. This Path is reserved for use by the application server.
```
ramkumar-k-9286:
```suggestion
CWWKO1750W_OPEN_API_UI_PATH_RESERVED=CWWKO1750W: The application server reserves the OpenAPI UI Path, so it cannot be set to {0}.
```

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
```diff
@@ -0,0 +1,34 @@
...
+CWWKO1750W_OPEN_API_UI_PATH_RESERVED.explanation=The OpenAPI UI Path cannot be set to a value that has been reserved for use by the application server.
```
ramkumar-k-9286:
```suggestion
CWWKO1750W_OPEN_API_UI_PATH_RESERVED.explanation= The application server reserves the use of certain values, preventing the OpenAPI UI Path from being set to these values.
```

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
```diff
@@ -0,0 +1,34 @@
...
+CWWKO1750W_OPEN_API_UI_PATH_RESERVED.useraction=Specify a valid value for the extensions attribute in the Micro Profile configuration.
+
+CWWKO1751W_OPEN_API_UI_PATH_INVALID=CWWKO1751W: The OpenAPI UI Path cannot be set to {0}. The Path is invalid or contains invalid characters.
+CWWKO1751W_OPEN_API_UI_PATH_INVALID.explanation=The OpenAPI UI Path cannot be set to a value containing invalid characters.
```
ramkumar-k-9286:
```suggestion
CWWKO1751W_OPEN_API_UI_PATH_INVALID.explanation= The value assigned to the OpenAPI UI Path must not include any characters that are considered invalid.

File: dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+CWWKO1751W_OPEN_API_UI_PATH_INVALID.useraction=Specify a valid value for the extensions attribute in the configuration.
+
+CWWKO1752W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT=CWWKO1752W: The OpenAPI UI Path has been set to the same path as the {0} feature and may prevent startup of itself or the other feature.
```
ramkumar-k-9286:
```suggestion
CWWKO1752W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT=CWWKO1752W: Setting the OpenAPI UI Path to the same path as the {0} feature might disrupt the startup process of itself or the other feature.
```

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
```diff
@@ -0,0 +1,34 @@
...
+CWWKO1752W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT.explanation=The OpenAPI UI Path has been set to the same path as another feature and prevent startup or access.
```
ramkumar-k-9286:
```suggestion
CWWKO1752W_OPEN_API_UI_PATH_POTENTIAL_CONFLICT.explanation= If the OpenAPI UI Path shares the same path as another feature, it can prevent startup or access.
```

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+        // given first pass will go to //foo/bar
+        while(path.contains("//")){
+            path = path.replace("//","/");
+        }
+        // Remove trailing slash if the path contains one.
+        if (path.endsWith("/") && !path.equals("/")) {
+            path = path.substring(0, path.length() - 1);
+        }
+        return path;
+    }
```
tevans78: Should these three "processing" methods be static and have separate unit tests to check all the parsing edge cases and invalid path values?

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
```diff
@@ -0,0 +1,34 @@
+###############################################################################
+# Copyright (c) 2017 IBM Corporation and others.
```
Azquelt: Date

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
```diff
@@ -0,0 +1,34 @@
...
+#ISMESSAGEFILE TRUE
+#NLS_MESSAGEFORMAT_NONE
+#NLS_ENCODING=UNICODE
+# -------------------------------------------------------------------------------------------------
+
+# Message prefix block: CWWKO1750 - CWWKO1799
+
+CWWKO1750W_OPEN_API_UI_PATH_RESERVED=CWWKO1750W: The OpenAPI UI Path cannot be set to {0}. This Path is reserved for use by the application server.
+CWWKO1750W_OPEN_API_UI_PATH_RESERVED.explanation=The OpenAPI UI Path cannot be set to a value that has been reserved for use by the application server.
+CWWKO1750W_OPEN_API_UI_PATH_RESERVED.useraction=Specify a valid value for the extensions attribute in the Micro Profile configuration.
```
Azquelt: This message is actually used when the OpenAPI document and UI paths are set to the same value but that isn't what it says.

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
```diff
@@ -0,0 +1,34 @@
...
+
+CWWKO1751W_OPEN_API_UI_PATH_INVALID=CWWKO1751W: The OpenAPI UI Path cannot be set to {0}. The Path is invalid or contains invalid characters.
+CWWKO1751W_OPEN_API_UI_PATH_INVALID.explanation=The OpenAPI UI Path cannot be set to a value containing invalid characters.
+CWWKO1751W_OPEN_API_UI_PATH_INVALID.useraction=Specify a valid value for the extensions attribute in the configuration.
```
Azquelt: We should list the valid characters in the message, otherwise the user won't know what they've done that's not allowed.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE, immediate = true, property = { "service.vendor=IBM" })
+public class OpenAPIUIEndpointManager {
+
+    private static final TraceComponent tc = Tr.register(OpenAPIUIEndpointManager.class);
+
+    private static final String OPEN_API_UI_VAR_Name = "openAPIUIPATH";
```
Azquelt: all caps for a constant

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+    private static final Pattern PATH_PATTERN = Pattern.compile("^(/[\\w./_-]*)?$");
+    private static final String[] INVALID_PATHS = {"/.","/.."};
+
+    private OpenAPIWABConfigManager openAPIWABConfigManager;
+
+    @Activate
+    protected void activate(ComponentContext context, Map<String, Object> properties){
+
+        String path = OPEN_API_UI_PATH;
+        if(ProductInfo.getBetaEdition()) {
```
Azquelt: We need an additional guard, since we're not even ready to beta this functionality yet.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
abutch3r: `validatePath` and `resolvePath` certainly should have unit tests. Though both were going to end up in common code when I did a bit of refactoring for the data endpoint as both have largely the same requirements
`validatePath` would need a small refactor to make it not reliant on config existing and that would allow it to also be `static`

The initial processing potentially could do, but would say moving the warning for potential conflicts out so that was more testable would be a better way to go.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+        openAPIWABConfigManager = new OpenAPIWABConfigManager(context, OPEN_API_UI_VAR_Name, path);
+        openAPIWABConfigManager.pushConfiguration();
+    }
+
+    @Deactivate
+    protected void deactivate(ComponentContext context, int reason){
+        openAPIWABConfigManager.deactivate();
+    }
+
+    protected String processOpenAPIUIPATH(final ComponentContext context, final Config config) {
```
Azquelt: "process" is not very descriptive of what this method does

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        public void modifiedEvent(Dictionary<String, String> props) {
+            Tr.event(tc, "OpenAPI UI web app bundle modified, WAB config=" + wabConfigReg + " using props=" + props);
+        }
+
+
+        public void deactivateEvent() {
+            Tr.event(tc, "Unregistered UI web app bundle, WAB config=" + wabConfigReg + " for contextName=" + contextName + " and contextPath=" + contextPath);
+        }
```
Azquelt: I think the code would be clearer with these methods inlined, they're all only called from one place.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+        return path;
+    }
+
+    public static <T> T getOptionalValue(Config config, String propertyName, Class<T> propertyType, T defaultValue) {
+        Optional<T> optional = config.getOptionalValue(propertyName, propertyType);
+        return optional.orElse(defaultValue);
+    }
+
+    final static class OpenAPIWABConfigManager{
```
Azquelt: I'm not sure what the point of having this as a separate inner class is.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+        if (path.endsWith("/") && !path.equals("/")) {
+            path = path.substring(0, path.length() - 1);
+        }
+        return path;
+    }
+
+    public static <T> T getOptionalValue(Config config, String propertyName, Class<T> propertyType, T defaultValue) {
+        Optional<T> optional = config.getOptionalValue(propertyName, propertyType);
+        return optional.orElse(defaultValue);
+    }
```
Azquelt: I think the code would be clearer if this method was inlined.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
Azquelt: Why is this method `protected`?

Actually why are all these methods protected?

I think that `@Activate` and `@Deactivate` methods are generally supposed to be `protected` for some reason I can't remember, but I can't see why the others should be.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UICustomPathTest.java
```diff
@@ -0,0 +1,159 @@
...
+    private RemoteWebDriver driver;
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        WebArchive war = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                .addClass(TestResource.class);
+
+        ShrinkHelper.exportDropinAppToServer(server, war, DeployOptions.SERVER_ONLY);
+
+        ServerFileUtils serverFileUtils = new ServerFileUtils();
```
Azquelt: I wouldn't use this class, it doesn't seem widely used or maintained.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UICustomPathTest.java
```diff
@@ -0,0 +1,159 @@
...
+        // append to bootstrap.properties
+        FileWriter writer = new FileWriter(bootProps, true);
+        writer.append(PROPERTY_NAME + "=" + PATH_VALUE);
+        writer.append(System.getProperty("line.separator"));
+        writer.close();
+
+        String jvmOptions = serverFileUtils.getServerFileLoc(server)+"/jvm.options";
+        writer = new FileWriter(jvmOptions, true);
+        writer.append("-Dcom.ibm.ws.beta.edition" + "=" + "true");
+        writer.close();
```
Azquelt: Usually we would just write the bootstrap.properties file ourselves and put it in `/publish/servers/<servername>` if the content doesn't need to be dynamic.

Here it might make sense for it to be dynamic if we write more tests, so we could do it this way, but in that case we'll also have to be a little bit careful to avoid the config from one test contaminating another test.

If we do want to write a properties file directly, I would generally use `Properties.load()` and `.store()` and you should always use the try-with-resources syntax to close writers where possible.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UICustomPathTest.java
```diff
@@ -0,0 +1,159 @@
...
+     * @param by the locating mechanism
+     * @param timeout the timeout
+     * @return the located element
+     * @throws TimeoutException if the element is not found within the timeout
+     */
+    private static WebElement waitForElement(SearchContext searchRoot, By by, Duration timeout) {
+        return new FluentWait<>(searchRoot).ignoring(NoSuchElementException.class)
+                .withTimeout(timeout)
+                .until(c -> c.findElement(by));
+    }
```
Azquelt: These methods should move to a common class.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UICustomPathTest.java
```diff
@@ -0,0 +1,159 @@
...
+        WebElement headerbarWrapper = waitForElement(headerbar, By.cssSelector("div.headerbar-wrapper"));
+        assertThat("Headerbar image", headerbarWrapper.getCssValue("background-image"), startsWith("url(\"data:image/png"));
+
+        // Check we can see and open the operation
+        WebElement testGetOpBlock = waitForElement(driver, By.id("operations-default-get_test"));
+        WebElement testGetButton = testGetOpBlock.findElement(By.tagName("button"));
+        testGetButton.click();
+        WebElement testGet200Response = waitForElement(testGetOpBlock, By.cssSelector("tr.response[data-code=\"200\"]"));
+        assertNotNull("200 response line", testGet200Response);
+    }
```
Azquelt: We could probably make this code common as well if we're going to want to repeat it in multiple places.

After calling `driver.get`, the only thing that would need to be passed in is the `driver`.

Azquelt (changes requested):
I've left a few comments, we also need to format the code according to the formatting rules in the project.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,192 @@
...
+        String dataEndpoint = getOptionalValue(config, OPEN_API_ENDPOINT_CONFIG_NAME, String.class, OPEN_API_ENDPOINT_RESERVED_PATH);
+        if (path.equals(dataEndpoint)) {
+            Tr.warning(tc, "CWWKO1750W_OPEN_API_UI_PATH_RESERVED", OPEN_API_ENDPOINT_RESERVED_PATH);
+            valid=false;
+        }
+        //check against known invalid paths that contain valid characters
+        else if (Arrays.asList(INVALID_PATHS).contains(path)){
+            Tr.warning(tc, "CWWKO17510W_OPEN_API_UI_PATH_INVALID", path);
+            valid=false;
+        }
```
Azquelt: This check only looks for the path being `/..` or `/.`, we should probably reject the path if it has any segment which is `..` or `.`

Azquelt (changes requested):

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
posmith: L2 reviewed and agrees that we should list the valid characters in the message.

posmith:
L2 message review completed

abutch3r:
#libby #build

abutch3r:
#libby

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
helyarp: I prefer the original message, with lowercase p unless path needs to be Path:
The OpenAPI UI Path cannot be set to {0}. This path is reserved for use by the application server.

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
helyarp: Although we try not to have passive voice in messages, I prefer the original message with "has been reserved" changed to "is reserved":

The OpenAPI UI Path cannot be set to a value that is reserved for use by the application server.

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
helyarp: Suggested message explanation:
The OpenAPI UI Path cannot be set to a value that contains invalid characters.

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
helyarp: Suggested change:
The OpenAPI UI Path was set to the same path as the {0} feature, which might prevent startup of itself or the other feature.

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
helyarp: Suggested change:
The OpenAPI UI Path was set to the same path as another feature, which can prevent startup or access.

abutch3r:
#libby #build

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
abutch3r: This message is used solely for one other endpoint - so have focused the message on that conflict instead of a generic message for any potential endpoint value

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
abutch3r: This message is used solely for one other endpoint - so have focused the message on that conflict instead of a generic message for any potential endpoint value

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
abutch3r: Changes made as suggested

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
abutch3r: suggested changes made

dev/com.ibm.ws.microprofile.openapi.ui/resources/com/ibm/ws/microprofile/openapi/ui/resources/OpenAPIUIMessages.nlsprops
abutch3r: Updated as suggested

abutch3r:
#libby

helyarp (commented):
Looks good!

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private static final String OPEN_API_UI_VAR_NAME = "openAPIUIPATH";
+    private static final String OPEN_API_UI_CONFIG_NAME = "mp.openapi.extensions.liberty.ui.path";
+    private static final String OPEN_API_UI_PATH = "/ui";
+
+    private static final Pattern PATH_PATTERN = Pattern.compile("^(/[\\w./_-]*)?$");
+    private static final Map<String, String> CONFLICTING_PATHS_MAP = Stream.of(
+                                                                               new AbstractMap.SimpleEntry<>("/ibm/api", "OpenAPI"),
+                                                                               new AbstractMap.SimpleEntry<>("/health", "mpHealth"),
+                                                                               new AbstractMap.SimpleEntry<>("/metrics", "mpMetrics")).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
Azquelt, line 47: I'm still not convinced this is necessary. I'll open an issue to review it again before release.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,153 @@
...
+     */
+    public static boolean validatePath(String path, String reservedPath) {
+        boolean valid = true;
+        try {
+            //temporarily construct a URL using provided path to use the URL validation code.
+            new URL("https://localhost" + path);
+        } catch (MalformedURLException e) {
+            Tr.warning(tc, "CWWKO1754W_OPEN_API_UI_PATH_INVALID",e.getLocalizedMessage());
+            return false;
+        }
```
Azquelt, line 91: Not convinced this is needed. In addition, the message produced here is likely to be less useful than those produced by the following lines.

We'll address it under another issue.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,153 @@
...
+            valid = false;
+            //check against other feature paths to see if a potential conflict might occur
+        }
+
+        // check that no segment is just `/.` or `/..`, `/.....` should be left as is as it has no special meaning
+        ArrayList<String> segments = new ArrayList<>(Arrays.asList(path.split("/")));
+        if (segments.contains(".") || segments.contains("..")) {
+            Tr.warning(tc, "CWWKO1755W_OPEN_API_UI_PATH_SEGMENT_INVALID");
+            valid = false;
+        }
```
Azquelt, line 108: These should be errors, not warnings. To be addressed in another issue.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,153 @@
...
+        if (ProductInfo.getBetaEdition()) {
+            Config config = ConfigProvider.getConfig(OpenAPIUIEndpointManager.class.getClassLoader());
+            if (config.getOptionalValue("open_api_path_enabled", Boolean.class).orElse(false)) {
+                resolvePathFromConfig(config);
+                if (!validatePath(uiPath, dataPath)) {
+                    uiPath = dataPath + OPEN_API_UI_PATH;
+                    Tr.warning(tc,"CWWKO1751W_OPEN_API_PATH_UPDATE_FAILED");
+                } else if(!uiPath.equals(OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH)){
+                    Tr.info(tc,"CWWKO1750I_OPEN_API_PATH_UPDATE",uiPath);
+                }
```
Azquelt: I don't like that `resolvePathFromConfig` is setting fields which this method then modifies.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIWABConfigManager.java
```diff
@@ -0,0 +1,95 @@
...
+        this.context = context;
+        this.contextName = contextName;
+        this.contextPath = contextPath;
+        this.name=name;
+    }
+
+    /**
+     * Active Web Bundle
+     */
+    public void activate() {
```
Azquelt: I don't think activate and deactivate are not good terms for what this class is doing. Those words already have a specific meaning for bundles and components in OSGi and this class is doing something different.

This class is setting a value for a WAB context path variable.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIWABConfigManager.java
```diff
@@ -0,0 +1,95 @@
...
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
+import org.osgi.service.component.ComponentContext;
+
+import java.util.Dictionary;
+import java.util.Hashtable;
+
+public class OpenAPIWABConfigManager {
+
+    private static final TraceComponent tc = Tr.register(OpenAPIUIEndpointManager.class);
```
Azquelt: The class on this line needs updating.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIWABConfigManager.java
```diff
@@ -0,0 +1,95 @@
...
+
+    final ComponentContext context;
+    final String contextName;
+    final String contextPath;
+    final String name;
+
+    ServiceRegistration<WABConfiguration> wabConfigReg;
```
Azquelt: I think these should all be private

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,153 @@
...
+    private OpenAPIWABConfigManager openAPIWABConfigManager;
+
+    @Activate
+    protected void activate(ComponentContext context, Map<String, Object> properties) {
+        //Default UI Path is /openapi/ui
+        uiPath = OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH;
+
+        if (ProductInfo.getBetaEdition()) {
+            Config config = ConfigProvider.getConfig(OpenAPIUIEndpointManager.class.getClassLoader());
+            if (config.getOptionalValue("open_api_path_enabled", Boolean.class).orElse(false)) {
```
Azquelt: This should read from system properties, not MP Config

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIWABConfigManager.java
```diff
@@ -0,0 +1,95 @@
...
+import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.wsspi.wab.configure.WABConfiguration;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
+import org.osgi.service.component.ComponentContext;
+
+import java.util.Dictionary;
+import java.util.Hashtable;
+
+public class OpenAPIWABConfigManager {
```
Azquelt: Class needs javadoc to say what it does.

Possibly it could have a better name too.

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
Azquelt: Need to update both the message codes and the call to `Tr.warning`.

Azquelt (dismissed):

Azquelt:
Raised #25367 for the message issues

abutch3r:
#libby #build

dev/com.ibm.ws.microprofile.openapi.ui/src/com/ibm/ws/microprofile/openapi/ui/OpenAPIUIEndpointManager.java
```diff
@@ -0,0 +1,154 @@
...
+    protected void activate(ComponentContext context, Map<String, Object> properties) {
+        //Default UI Path is /openapi/ui
+        uiPath = OPEN_API_ENDPOINT_PATH + OPEN_API_UI_PATH;
+
+        if (ProductInfo.getBetaEdition()) {
+            Config config = ConfigProvider.getConfig(OpenAPIUIEndpointManager.class.getClassLoader());
+            //check for system property `open_api_path_enabled` as additional guide - getBoolean returns `true` if the value exists and is set to `true`, if the value is `false`
+            if (Boolean.getBoolean("open_api_path_enabled")) {
+                resolvePathFromConfig(config);
+            }
```
Azquelt: Please move the `ConfigProvider.getConfig()` call inside the guard.

abutch3r:
No Docs changes since the last review

Azquelt (approved):

Azquelt:
Personal build does include the most recent changes.

abutch3r:
#libby

abutch3r:
While build only ran ~33% of test cases it did include the UI tests which did pass

abutch3r:
Being replaced by #25436 as this undoes 95% of the changes that were done here so to keep the commits cleaner this is being closed

PR #25268, opened 2023-05-15
Checkpoint remove beta check for InstantOn

tjwatson:


tjwatson:
#build

tjwatson:
#build

tjwatson:
#build

tevans78 (approved):
Looks good from a high level point of view.

dev/io.openliberty.checkpoint_fat/publish/servers/checkpointFaces-backendServices/jvm.options
```diff
@@ -1,2 +1 @@
-#-Dcom.ibm.ws.beta.edition=true
--Xdump:java:events=throw+systhrow,filter=org/eclipse/openj9/criu/JVMCheckpointException,request=exclusive+preempt
\ No newline at end of file
+#-Xdump:java:events=throw+systhrow,filter=org/eclipse/openj9/criu/JVMCheckpointException,request=exclusive+preempt
```
sebratton, line 1: Any reason you commented this out here but left it un-commented in other FATS?

tjwatson: This server must not be having a checkpoint being done on it.  Otherwise I would expect that the `com.ibm.ws.beta.edition` property must be set, but the original had it commented out.

The reason my change commented out the `-Xdump` is because I removed all instances of `-Dcom.ibm.ws.beta.edition=true\n"` from all of our FAT's `jvm.options`.  But because of the leading `#` here that stayed and became the first char of the `-Xdump`.  I suspect this particular server shouldn't have either option set at all.

sebratton (approved):

tjwatson:
#libby

PR #25273, opened 2023-05-15
PR for Issue 25247: Do not require client secret for OIDC clients using private_key_jwt token auth method

### LINKED ISSUE #25247: Social login should not require the client_secret with private key jwt
The oidc and social clients will be using a Private Key JWT instead of the secret, so we should no longer require the secret in the social login client configs.
The runtime needs to be updated and the clientSecret needs to be removed from the configurations used to test social login.

ayoho:
Resolves #25247

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/AuthorizationCodeHandler.java
+++ b/dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/AuthorizationCodeHandler.java
@@ -46,6 +46,7 @@ public class AuthorizationCodeHandler {
     private final HttpServletRequest request;
     private final HttpServletResponse response;
     private final ConvergedClientConfig clientConfig;
+    private final String configId;
     private final String clientId;
 
     private OidcClientUtil oidcClientUtil = null;
@@ -59,6 +60,7 @@ public class AuthorizationCodeHandler {
         this.response = response;
         this.clientConfig = clientConfig;
         clientId = clientConfig.getClientId();
+        configId = clientConfig.getId();
 
         oidcClientUtil = getOidcClientUtil();
         authenticatorUtil = getOIDCClientAuthenticatorUtil();
@@ -113,11 +115,11 @@ public class AuthorizationCodeHandler {
         try {
             oidcResult = sendTokenRequestAndValidateResult(oidcClientRequest, sslSocketFactory, authzCode, responseState, redirectUrl);
         } catch (BadPostRequestException e) {
-            Tr.error(tc, "OIDC_CLIENT_TOKEN_REQUEST_FAILURE", new Object[] { e.getMessage(), clientId, clientConfig.getTokenEndpointUrl() });
+            Tr.error(tc, "OIDC_CLIENT_TOKEN_REQUEST_FAILURE", new Object[] { e.getMessage(), configId, clientConfig.getTokenEndpointUrl() });
             sendErrorJSON(e.getStatusCode(), "invalid_request", e.getMessage());
             return new ProviderAuthenticationResult(AuthResult.FAILURE, e.getStatusCode());
         } catch (Exception e) {
-            Tr.error(tc, "OIDC_CLIENT_TOKEN_REQUEST_FAILURE", new Object[] { e.getLocalizedMessage(), clientId, clientConfig.getTokenEndpointUrl() });
+            Tr.error(tc, "OIDC_CLIENT_TOKEN_REQUEST_FAILURE", new Object[] { e.getLocalizedMessage(), configId, clientConfig.getTokenEndpointUrl() });
             return new ProviderAuthenticationResult(AuthResult.SEND_401, HttpServletResponse.SC_UNAUTHORIZED);
         }
         return oidcResult;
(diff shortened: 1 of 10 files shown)


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat.delegated,com.ibm.ws.security.social_fat.LibertyOP.1,com.ibm.ws.security.social_fat.LibertyOP.2,com.ibm.ws.security.social_fat.LibertyOP.3,com.ibm.ws.security.social_fat.LibertyOP.backchannelLogout,com.ibm.ws.security.social_fat.LibertyOP.claimPropagation,com.ibm.ws.security.social_fat.multiProvider,com.ibm.ws.security.social_fat.okdServiceLogin

dev/com.ibm.ws.security.social/resources/com/ibm/ws/security/social/resources/SocialMessages.nlsprops
```diff
@@ -655,3 +652,6 @@ BACKCHANNEL_REQUEST_NOT_SUPPORTED_CONFIG=CWWKS2350E: The back-channel logout req
 BACKCHANNEL_REQUEST_NOT_SUPPORTED_CONFIG.explanation=The social media login configuration is not an OpenID Connect client, so it does not support back-channel logout.
 BACKCHANNEL_REQUEST_NOT_SUPPORTED_CONFIG.useraction=Use the logout endpoint that is appropriate for the social media login configuration.
 
+CLIENT_SECRET_MISSING_BUT_REQUIRED_BY_TOKEN_AUTH_METHOD=CWWKS2351E: The {0} OpenID Connect client uses the {1} token endpoint authentication method which requires a client secret, but a client secret is not configured.
```
ramkumar-k-9286:
```suggestion
CLIENT_SECRET_MISSING_BUT_REQUIRED_BY_TOKEN_AUTH_METHOD=CWWKS2351E: The {0} OpenID Connect client uses the {1} token endpoint authentication method. This authentication method requires a client secret, which is not configured.
```

dev/com.ibm.ws.security.social/resources/com/ibm/ws/security/social/resources/SocialMessages.nlsprops
```diff
@@ -655,3 +652,6 @@ BACKCHANNEL_REQUEST_NOT_SUPPORTED_CONFIG=CWWKS2350E: The back-channel logout req
...
+CLIENT_SECRET_MISSING_BUT_REQUIRED_BY_TOKEN_AUTH_METHOD.explanation=The token endpoint authentication method that is specified in the message requires a client secret to authenticate with the token endpoint of the OpenID Connect provider.
+CLIENT_SECRET_MISSING_BUT_REQUIRED_BY_TOKEN_AUTH_METHOD.useraction=Either configure a client secret for the OpenID Connect client, or change the token endpoint authentication method used by the client to one that does not require a client secret.
```
ramkumar-k-9286:
```suggestion
CLIENT_SECRET_MISSING_BUT_REQUIRED_BY_TOKEN_AUTH_METHOD.useraction=Configure a client secret for the OpenID Connect client, or switch the token endpoint authentication method of the client to one that does not require a client secret.

**Comment by ayoho** [comment]
#libby
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat.delegated,com.ibm.ws.security.social_fat.LibertyOP.1,com.ibm.ws.security.social_fat.LibertyOP.2,com.ibm.ws.security.social_fat.LibertyOP.3,com.ibm.ws.security.social_fat.LibertyOP.backchannelLogout,com.ibm.ws.security.social_fat.LibertyOP.claimPropagation,com.ibm.ws.security.social_fat.multiProvider,com.ibm.ws.security.social_fat.okdServiceLogin

File: dev/com.ibm.ws.security.openidconnect.clients.common/src/com/ibm/ws/security/openidconnect/clients/common/AuthorizationCodeHandler.java [hunk 3]
```diff
@@ -59,6 +60,7 @@ public AuthorizationCodeHandler(HttpServletRequest request, HttpServletResponse
         this.response = response;
         this.clientConfig = clientConfig;
         clientId = clientConfig.getClientId();
+        configId = clientConfig.getId();
```
jimmy1wu, line 63: why are we using `configId` instead of `clientId` in these messages now? do the NLS messages have to be updated too since they use language like `The OpenID Connect client [{1}] is...`? i feel like this may possibly be confusing in the future.

ayoho: I noticed this when I was looking at the new error message we were emitting. We can have multiple `<openidConnectClient>` elements configured that all use the same client ID, i.e. the same OAuth client that's registered with the OP. However each `<openidConnectClient>` element has its own unique ID that's distinct from whatever client ID is being used by that configuration. We want the unique ID to be in these messages so that administrators know which `<openidConnectClient>` configuration is actually causing the problem.

We've got tests, for example, that have all of these configurations in the same server.xml:
```XML
    <oidcLogin clientId="client01" id="oidcLogin1" authFilterRef="authFilter1" ... />
    <oidcLogin clientId="client01" id="oidcLogin_badClientSecret" authFilterRef="authFilter_badClientSecret" ... />
    <oidcLogin clientId="client01" id="oidcLogin_blankClientSecret" authFilterRef="authFilter_blankClientSecret" ... />
    <oidcLogin clientId="client01" id="oidcLogin_emptyClientSecret" authFilterRef="authFilter_emptyClientSecret" ... />
    <oidcLogin clientId="client01" id="oidcLogin_XOR_Secret" authFilterRef="authFilter_xorSecret" ... />
   ...
```

They all use the same client ID, so before my change the error message would end up being `The OpenID Connect client [client01] is unable to contact the OpenID Connect provider...`, which doesn't point to which configuration actually had the problem. Now the message will be something like `The OpenID Connect client [oidcLogin_badClientSecret] is unable to contact the OpenID Connect provider...`, which points to the actual config being used.

jimmy1wu (approved):

PR #25281, opened 2023-05-16
persistence service updates and record-like class on Java 11

njr-11:
Persistence service updates from #25248
Also, I was able to copy addition parts of 25248, temporarily hard-coding them to use a simulated record-like class on Java 11.  This allows us to get more of the code in to ensure it won't be broken by other changes.


njr-11:
#build

KyleAure (approved):

njr-11:
#build
#libby

PR #25305, opened 2023-05-18
CrudRepository methods with record

njr-11:
Get the remaining CrudRepository methods working when the entity is a record.
This requires getting JPA to return query results as record instances rather than the generated entity and identifying the correct accessor from the record instead of the metamodel of the generated entity when attempting deletes based on specific entity instances rather than by their primary keys.


njr-11:
#build

mswatosh (approved):

PR #25315, opened 2023-05-22
Fix clientSecret checks for OIDC in social login

### LINKED ISSUE #25247: Social login should not require the client_secret with private key jwt

ayoho:
Fixes to ensure the `clientSecret` attribute is only required for OIDC clients in social login that don't use the `private_key_jwt` token endpoint authentication method.

Fixes #25247

### DIFF
--- a/dev/com.ibm.ws.security.social/resources/OSGI-INF/metatype/metatype.xml
+++ b/dev/com.ibm.ws.security.social/resources/OSGI-INF/metatype/metatype.xml
@@ -20,7 +20,7 @@
         <AD id="clientId" name="%clientId" description="%clientId.desc"
             required="true" type="String" />
         <AD id="clientSecret" name="%clientSecret" description="%clientSecret.desc"
-            type="String" ibm:type="password" />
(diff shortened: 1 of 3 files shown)


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat.delegated,com.ibm.ws.security.social_fat.LibertyOP.1,com.ibm.ws.security.social_fat.LibertyOP.2,com.ibm.ws.security.social_fat.LibertyOP.3,com.ibm.ws.security.social_fat.LibertyOP.backchannelLogout,com.ibm.ws.security.social_fat.LibertyOP.claimPropagation,com.ibm.ws.security.social_fat.multiProvider,com.ibm.ws.security.social_fat.okdServiceLogin

jimmy1wu (approved):

PR #25342, opened 2023-05-24
Ensure tx recovery at restore runs exactly once

dazavala:
A subtle change to ensure that recoverOnStartup is checked, and when enabled, that tx recovery runs exactly once after the transaction manager service starts (executes doStartup()) and transaction services.


dazavala:
#build

Issue #25353, opened 2023-05-25
mpTelemetry performance issue when mpTelemetry is disabled.


jdmcclur:
While @gjdeval was running the rest-crud app (which is a pretty simple primitive) with microProfile-6.0 we noticed that the TelemetryContainerFilter filter is doing work even though mpTelemetry is not enabled.

```
  Parent      0   0.00  12.04           0         983    J:org/jboss/resteasy/core/interception/jaxrs/PreMatchContainerRequestContext.filter()Lorg/jboss/resteasy/specimpl/BuiltResponse;
 
    Self      0   0.00  12.04           0         983    J:io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter$Proxy$_$$_WeldClientProxy.filter(Ljakarta/ws/rs/container/ContainerRequestContext;)V
 
   Child      0   0.06  11.34           5         926    J:io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter.filter(Ljakarta/ws/rs/container/ContainerRequestContext;)V
   Child      0   0.00   0.56           0          46    J:org/jboss/weld/bean/proxy/InterceptionDecorationContext.startIfNotEmpty()Lorg/jboss/weld/bean/proxy/InterceptionDecorationContext$Stack;
   Child      0   0.00   0.13           0          11    J:org/jboss/weld/bean/proxy/ProxyMethodHandler.getInstance()Ljava/lang/Object;
``` 

I am hoping we can avoid registering the filter if telemetry is disabled or at least add a check here to leave the instrumenter as null if telemetry is disabled, so all the filter methods are mostly skipped below (or something like that that would skip the work in the filters).  [https://github.com/OpenLiberty/open-liberty/blob/f553d73ff8feeb42374d37ae84444fbf9[…]roprofile/telemetry/internal/rest/TelemetryContainerFilter.java](https://github.com/OpenLiberty/open-liberty/blob/f553d73ff8feeb42374d37ae84444fbf9c0d9fa3/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter.java#L74)

PR #25366, opened 2023-05-30
Improve telemetry filters

Yas-Franco:
For issue #25353 

Adds a check to leave the instrumenter as null if telemetry is disabled, so all the filter methods are skipped.


Yas-Franco:
#libby
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
```diff
@@ -71,6 +71,11 @@ public void init() {
             if (instrumenter != null) {
                 return;
             }
+            
+            if(openTelemetry.toString().equals(OpenTelemetry.noop().toString())){
```
Azquelt: This doesn't feel like a good way to check whether we're using the noop `OpenTelemetry` object. In general `toString()` is not fully defined so there's no guarantee that it will be the same for two equal objects, or different for two different objects.

I suspect `openTelemetry.equals(OpenTelemetry.noop())` would work (though `OpenTelemetry.noop().equals(openTelemetry)` may not), or we should find a different way to check if we're enabled.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
```diff
@@ -71,6 +71,11 @@ public void init() {
...
+                this.instrumenter = null;
+                return;
+            }
```
Azquelt: I don't think this works. If `instrumenter` is `null`, won't the `filter` methods throw a `NullPointerException`?

For `TelemetryClientFilter`, can we do the check in `TelemetryClientBuilderListener` and `TelemetryRestClientBuilderListener` to avoid adding the filter in the first place if telemetry is not enabled?

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter.java
```diff
@@ -72,6 +72,11 @@ public TelemetryContainerFilter() {
 
     @Inject
     public TelemetryContainerFilter(final OpenTelemetry openTelemetry) {
+        //Instrumenter is null if OpenTelemetry is not enabled
+        if(openTelemetry.toString().equals(OpenTelemetry.noop().toString())){
+            this.instrumenter = null;
+            return;
+        }
```
Azquelt: This works a bit better here than in `TelemetryClientFilter` because the filter methods already check for the `instrumenter` being `null`. However, since we're now expecting `instrumenter` to be `null` much more often, we should check for that before calling `Context.current();`.

We register the `TelemetryContainerFilter` using a service loader, so there's no opportunity to dynamically skip registering it in the first place, so setting `instrumenter` to `null` here makes sense, but we should still find a better way to check if telemetry is disabled.

Azquelt (changes requested):

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
jdmcclur: In my experimenting openTelemetry.equals(OpenTelemetry.noop()) did not work. I think because openTelemetry is proxied via CDI.

Yas-Franco: Unfortunately you are right @jdmcclur which is why I had to use `toString()`. I have pushed up changes now which checks if Telemetry is disabled before registering the filters but was unsure if checking the properties multiple times would be a different performance issue. What do you both think?

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/helper/EnabledHelper.java
```diff
@@ -0,0 +1,53 @@
...
+    public boolean checkDisabled() {
+        HashMap<String, String> oTelConfigs = getTelemetryProperties();
+        //In order to enable any of the tracing aspects, the configuration otel.sdk.disabled=false must be specified in any of the configuration sources available via MicroProfile Config.
+        if (oTelConfigs.get(ENV_DISABLE_PROPERTY) != null) {
+            return Boolean.valueOf(oTelConfigs.get(ENV_DISABLE_PROPERTY));
+        } else if (oTelConfigs.get(CONFIG_DISABLE_PROPERTY) != null) {
+            return Boolean.valueOf(oTelConfigs.get(CONFIG_DISABLE_PROPERTY));
+        }
+        return true;
+    }
```
Azquelt: This class isn't needed any more, right?

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
```diff
@@ -82,48 +78,57 @@ public void init() {
...
-                } finally {
-                    request.removeProperty(configString + "context");
-                    request.removeProperty(configString + "parentContext");
-                }
-                return null;
-            }
-        });
+    public boolean isEnabled() {
+        return instrumenter != null;
     }
```
Azquelt, line 139: This method needs Javadoc

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
```diff
@@ -82,48 +78,57 @@ public void init() {
...
                             .addAttributesExtractor(NetClientAttributesExtractor.create(NET_CLIENT_ATTRIBUTES_GETTER))
                             .buildClientInstrumenter(new ClientRequestContextTextMapSetter());
+        } else {
+            this.instrumenter = null;
         }
     }
 
+    TelemetryClientFilter() {
+        this.instrumenter = null;
+    }
```
Azquelt, line 91: Add a comment to say this is the no-arg constructor required by CDI

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/FATSuite.java
```diff
@@ -21,17 +21,17 @@
...
+//                TelemetrySpiTest.class,
+//                TelemetryConfigEnvTest.class,
+//                TelemetryConfigServerVarTest.class,
+//                TelemetryConfigSystemPropTest.class,
+//                TelemetryConfigEnvOnlyTest.class,
+//                TelemetryConfigNullTest.class,
+//                TelemetryServiceNameTest.class,
+//                TelemetryLoggingExporterTest.class,
+//                TelemetryAPITest.class,
+//                MultiThreadedContextTest.class,
```
Azquelt: This change needs removed.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryInfo.java
```diff
@@ -0,0 +1,54 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal.cdi;
+
+import io.opentelemetry.api.OpenTelemetry;
+
+public class OpenTelemetryInfo {
```
Azquelt, line 19: Better javadoc on this class would be useful.

Azquelt (changes requested):

Yas-Franco:
#libby
#build

Yas-Franco:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
Azquelt: We found a way around this by creating `OpenTelemetryInfo` to hold both the `OpenTelemetry` object and whether or not telemetry is enabled.

Azquelt (approved):

PR #25371, opened 2023-05-30
Add betaVersion to liberty-versions.props

fritze2:
- betaVersion is used to create beta images of Open Liberty, ie. `openliberty-23.0.0.7-beta-cl230620230521-1900.zip`. I've included the property into the liberty-versions.props so downstream Liberty builds can publish a URL to that beta image.

### DIFF
--- a/dev/cnf/build.gradle
+++ b/dev/cnf/build.gradle
@@ -428,12 +428,25 @@ task zipGradleBootstrap(type: Zip) {
     }
 }
 
+def getBetaVersion() {
+    def (year, major, minor, month) = "${bnd.libertyRelease}".tokenize('.')
+    int new_month = month.toInteger() + 1
+    if (new_month > 12) {
+        new_month = 1
+        int new_year = year.toInteger() + 1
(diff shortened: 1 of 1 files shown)


fritze2:
#build

dev/cnf/build.gradle
```diff
@@ -428,12 +428,25 @@ task zipGradleBootstrap(type: Zip) {
     }
 }
 
+def getBetaVersion() {
+    def (year, major, minor, month) = "${bnd.libertyRelease}".tokenize('.')
+    int new_month = month.toInteger() + 1
+    if (new_month > 12) {
+        new_month = 1
+        int new_year = year.toInteger() + 1
+        return "" + new_year + "." + major + "." + minor + "." + new_month
```
wraschke, line 437: If current release is something like 23.0.0.12, then flow will go into this if statement and return 24.0.0.1?

wraschke (approved):

dev/cnf/build.gradle
fritze2: Right

PR #25395, opened 2023-06-01
Performance of ServerStartTest in kernel.boot_fat on macOS

jimblye:
ServerStartTest.java uses "jcmd" to find the pid for a java process.  jcmd sometimes hangs on MAC which adds about an hour to ServerStartTest.   The problem was only observed on MAC and is intermittent.


cbridgha (approved):
Looks good - so this doesn't affect linux OS?  so its not going to speed up remote run tests....

tbitonti (approved):

jimblye:
#build

PR #25408, opened 2023-06-03
version conflict on save

njr-11:
Add test coverage for version conflicts that occur during save.
Also, add coverage for a version that is a Timestamp rather than a numeric type.


njr-11:
#build

mswatosh (approved):

PR #25468, opened 2023-06-12
Find-and-delete on a couple of basic scenarios

njr-11:
This pull experiments with find-and-delete for two very basic scenarios, both involving annotatively-defined queries.  One returns at most a single removed entity.  The other returns a list of removed entities.

Also in this pull, I removed ReactiveRepository and updated a test class that inherited from it but didn't do anything reactive.  The ReactiveRepository interface (which had no methods defined yet) was recently deferred from the 1.0 release of Jakarta Data.


njr-11:
#build

jmstephensgit (approved):

PR #25469, opened 2023-06-12
SpringBootThinUtility filter 3.1 starter embedded server dependencies

jimblye:
Very slight change.  One less JAR dependency for jetty starter.  asm-analysis:jar is removed in the 3.1 list.


jimblye:
#build

PR #25505, opened 2023-06-15
ordering for find and delete

njr-11:
Various ways of specifying ordering when using find and delete:
- with Sort parameters
- with OrderBy in query-by-method name
- with OrderBy annotation


njr-11:
#build

jmstephensgit (approved):

PR #25512, opened 2023-06-16
get Data JPA tests working on Postgres

njr-11:
Get the io.openliberty.data.internal_fat_jpa test bucket full passing on PostgreSQL.
The only error appears to be on an exception path that we were trying out with Derby - it gets surfaced as a different DataException.  The tests really shouldn't be attempting that path at all because it isn't supported.


njr-11:
#build

mswatosh (approved):

PR #25517, opened 2023-06-19
Reduce Bouncy castle jar for install kernel

jjiwooLim:
Reduced the bc provider jar to only include packages needed by featureUtility signature verification process.
com.ibm.ws.install.jar went from 6.4MB to 3.3MB


jjiwooLim:
#build

bensonlatibm (approved):

PR #25551, opened 2023-06-22
Better clean up after ServerEnvTest fat

### LINKED ISSUE #25493: File default.log exists after ServerEnvTest runs.
The ServerEnvTest sets the LOG_FILE variable in server.env and changes the name of the log file.   It attempts to rename that to console.log at the end of each test.   The rename is failing because the server is still running at that point.
There are several other tests that are failing because console.log does not exist.   I don't think ServerEnvTest is the cause of the other tests failing, because console.log should be created every time the server is started.    But to remove any doubt, the ServerEnvTest can be modified to use another means of testing besides changing the log file name.

jimblye:
Test bug.   Failures during cleanup phase of ServerEnvTest are causing other tests to fail.
Fixes #25493


jimblye:
#build

hlhoots (approved):
LGTM

jimblye:
#build

jimblye:
#libby

jimblye:
#build

jimblye:
#build

jimblye:
#build

PR #25570, opened 2023-06-26
Fix defect rtc-296420

dazavala:
The TransactionLogTest and TransactionManagerTest FATs fail during `Repeat.withoutModification()`, only, when the checkpoint transactions FAT suite is run in FULL mode.  These test can suffer two problems, both of which will cause a mismatch between the EE API used by the test apps and the test servers:

1) The expected application fails to export (deploy) during test setup. **Fix:** Harden the test setup to first clean-up the exported app archives and then export the test app archives using the OVERWRITE deploy option.  

2) The test server contains a non-EE8 feature configuration.  **Fix:** Modify the `withoutModification()` part of the Repeat rule to use the EE8FeatureReplacementAction rather than rely on `LibertyServerFactory.getLibertyServer(test-server, null, true)` to copy the expected version of file `publish/servers/{test-server}/server.xml` to `${server.config.dir}`.

### DIFF
--- a/dev/io.openliberty.checkpoint_fat_transaction/fat/src/io/openliberty/checkpoint/fat/TransactionLogTest.java
+++ b/dev/io.openliberty.checkpoint_fat_transaction/fat/src/io/openliberty/checkpoint/fat/TransactionLogTest.java
@@ -26,24 +26,26 @@ import java.io.UncheckedIOException;
 import java.util.function.Consumer;
 
 import org.junit.After;
-import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


dazavala:
#build

dazavala:
#build

dazavala:
#libby

dazavala:
All personal builds pass.  Here's the [InstantOn result](https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_U933IBUbEe65IbOhZCpF2w&tab=com.ibm.team.build.web.ui.internal.editors.result.overview.OverviewPage). The break in `Simple2PCCloudTest` is a recent problem involving WebContainer, which fails when processing work because it lacks a reference to the ManagedObjectService.

sebratton (approved):

PR #25572, opened 2023-06-26
Update weld to 5.1.1.Final

### LINKED ISSUE #25558: Update to Weld 5.1.1.Final
https://repo1.maven.org/maven2/org/jboss/weld/weld-osgi-bundle/5.1.1.Final/

Also update weld classfile writer to 1.3 as the weld update depends on it.

benjamin-confino:
fixes #25558


benjamin-confino:
#build

tevans78 (approved):

tevans78 (changes requested):
I'm afraid that you need to update the package exports as well. Feel free to work out a more efficient way of doing this.
e.g. this file
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/io.openliberty.org.jboss.weld5/bnd.overrides

tevans78 (dismissed):
Please add a readme somewhere to explain how someone should update Weld and what the gradle script actually does

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

dev/io.openliberty.org.jboss.weld5/build.gradle
```diff
@@ -0,0 +1,161 @@
...
+
+task bndMagic {
+
+    doFirst{
+        copy {   
+            configurations.weld.asFileTree.each {
+               from(zipTree(it))
+            }
+            into unzippDir
+        }
```
Azquelt: Given that we only want the manifest, we should change this to only extract the manifest

dev/io.openliberty.org.jboss.weld5/build.gradle
```diff
@@ -0,0 +1,161 @@
...
+       }
+       index++
+    }
+
+    return returnArray
+}
+
+compileJava {
+    dependsOn bndMagic
+}
```
Azquelt: This doesn't run early enough for bnd to pick up the modified bnd file.

I'm not sure we _can_ run a gradle task early enough for bnd to see the changes, unfortunately.

Might be better as a manually run task.

dev/io.openliberty.org.jboss.weld5/build.gradle
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        // Compile the full exports string. Consisting of: 
+        // 1) The new exports we add - a trailing new line char is included in the string
+        // A single whitespace charachter because String.join won't do a prefix to the first entry.
+        // Every entry followed by a ',\' a line break, and a whitespace to start the next line.
+        String exportsBndFormat = additionalExports + " " + String.join(",\\\\" + System.lineSeparator() + " ", upstreamExports)
+
+        // The input string uses the same format
+        String importsBndFormat = additionalImports + " " + String.join(",\\\\" + System.lineSeparator() + " ", upstreamImports)
```
Azquelt: I don't _think_ we want `System.lineSeparator()` here. We don't want the bnd file to have `\r\n` if it was generated on Windows.

dev/io.openliberty.org.jboss.weld5/build.gradle
```diff
@@ -0,0 +1,161 @@
...
+  weld {
+    transitive = false
+  }
+}
+
+dependencies {
+  weld "org.jboss.weld:weld-osgi-bundle:${weldVersion}"
+}
+
+task bndMagic {
```
Azquelt, line 62: Could we have a more descriptive name which describes the type of bnd magic that we're doing, like `updateWeldPackages`

Azquelt: This isn't a well behaved gradle task as it doesn't declare its inputs and outputs so Gradle can't work out whether it needs to be re-run or not which may result in this bundle being rebuilt every time even if there are no changes.

I'm not sure that that actually matters at the moment since our build is so bad anyway and it's a bit difficult to get right, so I'm ok leaving it for the moment.

dev/io.openliberty.org.jboss.weld5/bnd.bnd
```diff
@@ -18,7 +18,7 @@ javac.target: 11
 
 Require-Capability: osgi.ee; filter:="(&(osgi.ee=JavaSE)(version=11))"
 
--include= jar:${fileuri;${repo;org.jboss.weld:weld-osgi-bundle;[5.1,5.2)}}!/META-INF/MANIFEST.MF,bnd.overrides
+-include= jar:${fileuri;${repo;org.jboss.weld:weld-osgi-bundle;[5.1,5.2)}}!/META-INF/MANIFEST.MF,check-me-in-if-i-change.bnd
```
Azquelt, line 21: If this means we're no longer using overrides.bnd, then that file should be deleted.

Azquelt: Ah, you have done this. It shows up as a rename which is why I don't see the deletion of overrides.bnd in the file list.

dev/io.openliberty.org.jboss.weld5/check-me-in-if-i-change.bnd
```diff
@@ -7,97 +7,124 @@ bFullVersion=${version;==;${bVersion}}.${libertyBundleMicroVersion}
 Build-Identifier: SNAPSHOT-${now}
 Bundle-Version: ${bFullVersion}.${if;${driver;gradle};${version.qualifier};eclipse}
 
-# The first 8 exported packages are additions, the rest come from the original manifest
+Service-Component: \
+  com.ibm.ws.weld.deserializationClassProvider;\
```
Azquelt, line 16: It would be helpful for this file to have a header that says something like "Changes to this file should be made by modifying bnd.template"

I think that's general enough that you can just put it in bnd.template. We could do a bigger warning like "THIS FILE IS GENERATED", but we'd then have to generate the header in build.gradle.

benjamin-confino: That's in the readme but I can put it there too.

benjamin-confino:
#build

benjamin-confino:
#build

Azquelt (dismissed):
Looks much better, thanks.

benjamin-confino:
#build

benjamin-confino:
#build

dev/io.openliberty.org.jboss.weld5/build.gradle
```diff
@@ -0,0 +1,196 @@
...
+        File cnfDepFile = cnf.file('oss_dependencies.maven')
+        Path pathToCnfDepFile = Paths.get(cnfDepFile.getAbsolutePath()) //toPath() causes a compile error
+        String searchString = "org.jboss.weld:weld-osgi-bundle:${weldVersion}"
+
+        List<String> foundMatch = Files.lines(pathToCnfDepFile)
+        .filter(line -> line.contains(searchString))
+        .collect(Collectors.toList())
+
+        if (foundMatch.isEmpty()) {
+            throw new GradleException("The weld version ${weldVersion} was not found in the cnf/oss_dependencies.maven. Did you change the version in cnf/oss_dependencies.maven but not in this project's build.gradle?")
```
Yas-Franco, line 88: This is the only change from yesterday's approval. Happy to approve

Yas-Franco (approved):

benjamin-confino:
#run-libby-bot

Issue #25528, opened 2023-06-20
Confusing output to STDERR when a checkpoint is taken


sebratton:
On RHEL (but not ubuntu) when taking a checkpoint, there is some unexpected output at the end that looks like an error message. It's not an error message, the script proceeds, and a checkpoiont image is produced.  A following server restore is successful. 

This is likely the shell generating a "diagnostic" message about the process being terminated after checkpoint. This may be happening on RHEL and not ubuntu because they have different defaults for "#/bin/sh". Ubuntu defaults to "dash" which apparently is a more "posix-compliant shell to bourne". 

The errant output is the last line of output below. The pid listed is the pid of the checkpointed java process.

[jazz_build@impinger1 wlp]$ bin/server checkpoint server1

Launching server1 (WebSphere Application Server 23.0.0.6/wlp-1.0.78.cl230620230612-1100) on Eclipse OpenJ9 VM, version 17.0.7+7 (en_US)
[AUDIT   ] CWWKE0001I: The server server1 has been launched.
[AUDIT   ] CWWKE0100I: This product is licensed for development, and limited production use. The full license terms can be viewed here: https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/license/base_ilan/ilan/23.0.0.6/lafiles/en.html
[AUDIT   ] CWWKZ0058I: Monitoring dropins for applications.
[AUDIT   ] CWWKC0451I: A server checkpoint "beforeAppStart" was requested. When the checkpoint completes, the server stops.
**bin/server: line 951: 118749 Killed                  "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null**
[jazz_build@impinger1 wlp]$

PR #25584, opened 2023-06-27
Avoid distracting output on stderr after a checkpoint

sebratton:
On bash, when criu kills the server process, bash writes a message to standard error containing 'Killed' and the line number and line of the script where java was launched. This is redirected to /dev/null with this change. NOTE this does not affect execution on bourne shell where the message is coming from the STDERR of the java process itself. The message is not as distracting in that case since it's simply the word Killed.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -1046,7 +1046,12 @@ javaCmd()
     # trap on the signals here to ensure the background tail process ends.
     trap "exit" INT TERM
     trap "killSilent $TAIL_PID" EXIT
-    "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+    #On some shells (bash) the shell will issue an ugly message to STDERR when criu kills the process.
+    # We direct STDERR to /dev/null to avoid that noise.
+    {
+      #  Launch with stdin set to /dev/null to avoid file-handle complications on criu restore
+      "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+    } 2> /dev/null
     rc=$?
     if [ $rc -eq 20 ]; then
       # bad checkpoint arg; sleep to allow the message to go to tail


sebratton:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1046,7 +1046,12 @@ javaCmd()
     # trap on the signals here to ensure the background tail process ends.
     trap "exit" INT TERM
     trap "killSilent $TAIL_PID" EXIT
-    "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+    #On some shells (bash) the shell will issue an ugly message to STDERR when criu kills the process.
+    # We direct STDERR to /dev/null to avoid that noise.
+    {
+      #  Launch with stdin set to /dev/null to avoid file-handle complications on criu restore
+      "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+    } 2> /dev/null
```
tjwatson, line 1054: We had discussed piping both 1 and 2 to /dev/null here

```
    } 2>&1 > /dev/null
```

sebratton: The reason for piping both stderr and stdout was that I had incorrectly stated that on ubuntu  it was emitting the message to stdout. But I was mistaken. It is emitting it to stderr as on RHEL. The reason the "Killed" is still appearing on ubuntu is that it seems to be emitted to the same stderr stream that the forked java process is writting to and appears in the `${CHECKPOINT_CONSOLE_LOG}`. Whereas on RHEL it seems to be emitted by the STDERR of bash itself and so does NOT appear in the `${CHECKPOINT_CONSOLE_LOG}`. So I'm unable to block the message on ubuntu without doing somting more aggressive like filtering. However since we are targetting container using RHEL-UBI this release, it seems a valid fix. Also the message as formatted by bash looks more like something went wrong than the one on ubuntu, which is just a single word, "Killed".

The distinction between behavior on RHEL/Ubuntu is probably down to the fact that  `#!/bin/sh` on RHEL invokes bash by default whereas on ubuntu it invokes dash which is supposed to be a POSIX compliant stand-in for /bin/sh

tjwatson (approved):

sebratton:
Failures in the PB FATs were know or occurring in other builds without the commits here.

sebratton:
Fixes issue #25528

PR #25585, opened 2023-06-27
find and remove record entities

njr-11:
Combined find and remove capability when a record is used as the entity, and tests.
This also includes a couple of other minor fixes, such as clarifying the generated entity class name.  It was previously misleadingly named as a Record, which it is not.


KyleAure (dismissed):

njr-11:
#build

njr-11:
#build
#libby

KyleAure (approved):

PR #25596, opened 2023-06-29
find-and-delete returning IdClass

njr-11:
Find-and-delete repository operation that returns one or more IdClass results.
Also includes deleteFirst[#] with IdClass results, and tests for both.


njr-11:
#build

KyleAure (approved):

PR #25621, opened 2023-07-03
Only generate anyAttribute in XSD for types with extraProperties

### LINKED ISSUE #25579: server.xml schema (XSD) generated from metatype (e.g. with schemaGen) uses  'xsd:anyAttribute' too liberally, preventing validation of correct attribute names
**Describe the bug**  

In working with server.xml in IDE tooling like: WDT, LDT, or Liberty Tools, it would be nice to get assistance on an incorrect attribute name.

So e.g. if instead of

```
<webApplication contextRoot="ctx" ...
```

I type:  `<webApplication context-root="ctx"`, I should get some kind of diagnostic or error indication.

Though the tools above do provide XSD-based validation, the usefulness is limited because the XSD they validate against is not generated from Liberty metatype in a helpful way (for this use case).

**Steps to Reproduce**  
Run `serverSchemaGen defaultServer`, the `<webApplication>` element is defined by type:

```xml
    <xsd:complexType name="com.ibm.ws.app.manager.webappcfg"> 
        ...
        <xsd:anyAttribute processContents="skip"> 
        </xsd:anyAttribute> 
    </xsd:complexType>
```

The `<xsd:anyAttribute>` at the end means there is no way to distinguish the correct spelling of 'contextRoot', (or to generally detect invalid values),

**Expected behavior**  

Looking at the code [here](https://github.com/OpenLiberty/open-liberty/blob/fab751a0ced3c747a5bd0509977150d9c8de7206/dev/com.ibm.ws.config/src/com/ibm/ws/config/schemagen/internal/SchemaWriter.java#L1043-L1049):

```java
        if (type.getHasExtraProperties()) {
            writeInternalPropertiesType(type);
        }


        if (generateWildcards) {
            writeAttributeWildcard();
        }
```

perhaps a better approach would be something more like:

```java
        if (type.getHasExtraProperties()) {
            writeAttributeWildcard();
        }
```

That is, if a metatype is marked as allowing extra properties, like, e.g. some of the DB/JDBC-related config which has driver-defined properties, then generate the xsd:anyAttribute.  BUT, do NOT generate this anyAttribute otherwise.

**Diagnostic information:**  
 - OpenLiberty Version: 23.0.0.6 and as far back as I can see:
 - Java Version: [i.e. full output of `java -version`]
 - server.xml configuration: see above

**Additional context**  

I'm not sure to what degree this would represent an external change, or if this can be considered purely an implementation detail.    It seems it has external-type of impacts.  Labeling as a design issue to discuss further.

E.g. would this require a new "outputVersion" value for the [schemaGen](https://www.openliberty.io/docs/latest/reference/command/schemaGen.html) command?

scottkurz:
Fixes #25579

### DIFF
--- a/dev/com.ibm.ws.config/src/com/ibm/ws/config/schemagen/internal/SchemaWriter.java
+++ b/dev/com.ibm.ws.config/src/com/ibm/ws/config/schemagen/internal/SchemaWriter.java
@@ -1041,10 +1041,6 @@ class SchemaWriter {
         }
 
         if (type.getHasExtraProperties()) {
-            writeInternalPropertiesType(type);
-        }
-
-        if (generateWildcards) {
             writeAttributeWildcard();
         }
 


scottkurz:
#build

PR #25626, opened 2023-07-04
never throw an otel exception

### LINKED ISSUE #25059: Ensure MP Telemetry can never throw exceptions
OpenTelemetry itself never throws exceptions, it just logs a problem and continues. This ensures that a problem in the telemetry code doesn't impact the normal function of the application.

Our MP Telemetry code needs to provide similar guarantees. The easiest way to do this is to ensure that all of our entry points have a try-catch block around them which will catch and log any exception. We might need a new generic "a telemetry error occurred" message if we don't already have one.

Where we have filters with request and response methods, we must make sure that the response method gracefully handles the case where the request method failed.

benjamin-confino:
fixes #25059

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package io.openliberty.microprofile.telemetry.internal.cdi;
 
+import java.util.Collections;
 import java.util.Map;
 import java.util.function.BiConsumer;
 
@@ -19,47 +20,91 @@ import io.opentelemetry.api.baggage.Baggage;
 import io.opentelemetry.api.baggage.BaggageBuilder;
 import io.opentelemetry.api.baggage.BaggageEntry;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
 /**
  * This proxy class redirects method calls to Baggage.current(), by doing so it allows people to use @Inject Baggage and get an object which will not become obsolete.
  */
 public class BaggageProxy implements Baggage {
 
+    private static final TraceComponent tc = Tr.register(BaggageProxy.class);
+
     @Override
     public int size() {
-        return Baggage.current().size();
+        try {
+            return Baggage.current().size();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return 0;
+        }
     }
 
     @Override
     public void forEach(BiConsumer<? super String, ? super BaggageEntry> consumer) {
-        Baggage.current().forEach(consumer);
+        try {
+            Baggage.current().forEach(consumer);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+        }
     }
 
     @Override
     public Map<String, BaggageEntry> asMap() {
-        return Baggage.current().asMap();
+        try {
+            return Baggage.current().asMap();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return Collections.emptyMap();
+        }
     }
 
     @Override
     public String getEntryValue(String entryKey) {
-        return Baggage.current().getEntryValue(entryKey);
+        try {
+            return Baggage.current().getEntryValue(entryKey);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return "";
+        }
     }
 
     @Override
     public BaggageBuilder toBuilder() {
-        return Baggage.current().toBuilder();
+        try {
+            return Baggage.current().toBuilder();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return Baggage.builder();
+        }
     }
     
     @Override
     public boolean equals(Object obj) {
-        return Baggage.current().equals(obj);
+        try {
+            return Baggage.current().equals(obj);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return false;
+        }
     }
 
     public int hashCode() {
-        return Baggage.current().hashCode();
+        try {
+            return Baggage.current().hashCode();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return -1;
+        }
     }
 
     @Override
     public String toString() {
-        return Baggage.current().toString();
+        try {
+            return Baggage.current().toString();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return "";
+        }
     }
 }
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
@@ -15,6 +15,7 @@ package io.openliberty.microprofile.telemetry.internal.cdi;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -68,41 +69,44 @@ public class OpenTelemetryProducer {
     @ApplicationScoped
     @Produces
     public OpenTelemetryInfo getOpenTelemetryInfo() {
-
-        if (AgentDetection.isAgentActive()) {
-            // If we're using the agent, it will have set GlobalOpenTelemetry and we must use its instance
-            // all config is handled by the agent in this case
-            return new OpenTelemetryInfo(true, GlobalOpenTelemetry.get());
-        }
-
-        HashMap<String, String> telemetryProperties = getTelemetryProperties();
-
-        //Builds tracer provider if user has enabled tracing aspects with config properties
-        if (!checkDisabled(telemetryProperties)) {
-            OpenTelemetry openTelemetry = AccessController.doPrivileged((PrivilegedAction<OpenTelemetry>) () -> {
-                return AutoConfiguredOpenTelemetrySdk.builder()
-                                .addPropertiesCustomizer(x -> telemetryProperties) //Overrides OpenTelemetry's property order
-                                .addResourceCustomizer(this::customizeResource)//Defaults service name to application name
-                                .setServiceClassLoader(Thread.currentThread().getContextClassLoader())
-                                .setResultAsGlobal(false)
-                                .registerShutdownHook(false)
-                                .build()
-                                .getOpenTelemetrySdk();
-            });
-
-            if (openTelemetry != null) {
-                return new OpenTelemetryInfo(true, openTelemetry);
+        try {
+            if (AgentDetection.isAgentActive()) {
+                // If we're using the agent, it will have set GlobalOpenTelemetry and we must use its instance
+                // all config is handled by the agent in this case
+                return new OpenTelemetryInfo(true, GlobalOpenTelemetry.get());
             }
+
+            HashMap<String, String> telemetryProperties = getTelemetryProperties();
+
+            //Builds tracer provider if user has enabled tracing aspects with config properties
+            if (!checkDisabled(telemetryProperties)) {
+                OpenTelemetry openTelemetry = AccessController.doPrivileged((PrivilegedAction<OpenTelemetry>) () -> {
+                    return AutoConfiguredOpenTelemetrySdk.builder()
+                                    .addPropertiesCustomizer(x -> telemetryProperties) //Overrides OpenTelemetry's property order
+                                    .addResourceCustomizer(this::customizeResource)//Defaults service name to application name
+                                    .setServiceClassLoader(Thread.currentThread().getContextClassLoader())
+                                    .setResultAsGlobal(false)
+                                    .registerShutdownHook(false)
+                                    .build()
+                                    .getOpenTelemetrySdk();
+                });
+
+                if (openTelemetry != null) {
+                    return new OpenTelemetryInfo(true, openTelemetry);
+                }
+            }
+            //By default, MicroProfile Telemetry tracing is off.
+            //The absence of an installed SDK is a “no-op” API
+            //Operations on a Tracer, or on Spans have no side effects and do nothing
+            ComponentMetaData cData = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData();
+            String applicationName = cData.getJ2EEName().getApplication();
+            Tr.info(tc, "CWMOT5100.tracing.is.disabled", applicationName);
+
+            return new OpenTelemetryInfo(false, OpenTelemetry.noop());
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return new OpenTelemetryInfo(false, OpenTelemetry.noop());
         }
-        //By default, MicroProfile Telemetry tracing is off.
-        //The absence of an installed SDK is a “no-op” API
-        //Operations on a Tracer, or on Spans have no side effects and do nothing
-        ComponentMetaData cData = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData();
-        String applicationName = cData.getJ2EEName().getApplication();
-        Tr.info(tc, "CWMOT5100.tracing.is.disabled", applicationName);
-
-        return new OpenTelemetryInfo(false, OpenTelemetry.noop());
-
     }
 
     //Uses application name if the user has not given configured service.name resource attribute
@@ -114,7 +118,6 @@ public class OpenTelemetryProducer {
                 appName = cmd.getModuleMetaData().getApplicationMetaData().getName();
             }
         }
-
         return appName;
     }
 
@@ -126,31 +129,34 @@ public class OpenTelemetryProducer {
     }
 
     public void disposeOpenTelemetry(@Disposes OpenTelemetry openTelemetry) {
-
-        if (AgentDetection.isAgentActive()) {
-            return;
-        }
-
-        if (openTelemetry instanceof OpenTelemetrySdk) {
-            OpenTelemetrySdk sdk = (OpenTelemetrySdk) openTelemetry;
-            List<CompletableResultCode> results = new ArrayList<>();
-
-            SdkTracerProvider tracerProvider = sdk.getSdkTracerProvider();
-            if (tracerProvider != null) {
-                results.add(tracerProvider.shutdown());
+        try {
+            if (AgentDetection.isAgentActive()) {
+                return;
             }
 
-            SdkMeterProvider meterProvider = sdk.getSdkMeterProvider();
-            if (meterProvider != null) {
-                results.add(meterProvider.shutdown());
-            }
+            if (openTelemetry instanceof OpenTelemetrySdk) {
+                OpenTelemetrySdk sdk = (OpenTelemetrySdk) openTelemetry;
+                List<CompletableResultCode> results = new ArrayList<>();
 
-            SdkLoggerProvider loggerProvider = sdk.getSdkLoggerProvider();
-            if (loggerProvider != null) {
-                results.add(loggerProvider.shutdown());
-            }
+                SdkTracerProvider tracerProvider = sdk.getSdkTracerProvider();
+                if (tracerProvider != null) {
+                    results.add(tracerProvider.shutdown());
+                }
 
-            CompletableResultCode.ofAll(results).join(10, TimeUnit.SECONDS);
+                SdkMeterProvider meterProvider = sdk.getSdkMeterProvider();
+                if (meterProvider != null) {
+                    results.add(meterProvider.shutdown());
+                }
+
+                SdkLoggerProvider loggerProvider = sdk.getSdkLoggerProvider();
+                if (loggerProvider != null) {
+                    results.add(loggerProvider.shutdown());
+                }
+
+                CompletableResultCode.ofAll(results).join(10, TimeUnit.SECONDS);
+            }
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
         }
     }
 
@@ -165,24 +171,34 @@ public class OpenTelemetryProducer {
     }
 
     private HashMap<String, String> getTelemetryProperties() {
-        HashMap<String, String> telemetryProperties = new HashMap<>();
-        for (String propertyName : config.getPropertyNames()) {
-            if (propertyName.startsWith("otel.")) {
-                config.getOptionalValue(propertyName, String.class).ifPresent(
-                                                                              value -> telemetryProperties.put(propertyName, value));
+        try {
+            HashMap<String, String> telemetryProperties = new HashMap<>();
+            for (String propertyName : config.getPropertyNames()) {
+                if (propertyName.startsWith("otel.")) {
+                    config.getOptionalValue(propertyName, String.class).ifPresent(
+                                                                                  value -> telemetryProperties.put(propertyName, value));
+                }
             }
+            //Metrics and logs are disabled by default
+            telemetryProperties.put(CONFIG_METRICS_EXPORTER_PROPERTY, "none");
+            telemetryProperties.put(CONFIG_LOGS_EXPORTER_PROPERTY, "none");
+            telemetryProperties.put(ENV_METRICS_EXPORTER_PROPERTY, "none");
+            telemetryProperties.put(ENV_LOGS_EXPORTER_PROPERTY, "none");
+            return telemetryProperties;
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return new HashMap<String, String>();
         }
-        //Metrics and logs are disabled by default
-        telemetryProperties.put(CONFIG_METRICS_EXPORTER_PROPERTY, "none");
-        telemetryProperties.put(CONFIG_LOGS_EXPORTER_PROPERTY, "none");
-        telemetryProperties.put(ENV_METRICS_EXPORTER_PROPERTY, "none");
-        telemetryProperties.put(ENV_LOGS_EXPORTER_PROPERTY, "none");
-        return telemetryProperties;
     }
 
     @Produces
     public Tracer getTracer(OpenTelemetry openTelemetry) {
-        return openTelemetry.getTracer(INSTRUMENTATION_NAME);
+        try {
+            return openTelemetry.getTracer(INSTRUMENTATION_NAME);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return OpenTelemetry.noop().getTracerProvider().get("");
+        }
     }
 
     @Produces
@@ -200,6 +216,11 @@ public class OpenTelemetryProducer {
     @ApplicationScoped
     @Produces
     public OpenTelemetry getOpenTelemetry(OpenTelemetryInfo openTelemetryInfo) {
-        return openTelemetryInfo.getOpenTelemetry();
+        try {
+            return openTelemetryInfo.getOpenTelemetry();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return OpenTelemetry.noop();
+        }
     }
 }
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
@@ -17,74 +17,142 @@ import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.api.trace.StatusCode;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
 /**
  * This proxy class redirects method calls to Span.current(), by doing so it allows people to use @Inject Span and get an object which will not become obsolete.
  */
 public class SpanProxy implements Span {
 
+    private static final TraceComponent tc = Tr.register(SpanProxy.class);
+
     @Override
     public <T> Span setAttribute(AttributeKey<T> key, T value) {
-        return Span.current().setAttribute(key, value);
+        try {
+            return Span.current().setAttribute(key, value);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return this;
+        }
     }
 
     @Override
     public Span addEvent(String name, Attributes attributes) {
-        return Span.current().addEvent(name, attributes);
+        try {
+            return Span.current().addEvent(name, attributes);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return this;
+        }
     }
 
     @Override
     public Span addEvent(String name, Attributes attributes, long timestamp, TimeUnit unit) {
-        return Span.current().addEvent(name, attributes, timestamp, unit);
+        try {
+            return Span.current().addEvent(name, attributes, timestamp, unit);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return this;
+        }
     }
 
     @Override
     public Span setStatus(StatusCode statusCode, String description) {
-        return Span.current().setStatus(statusCode, description);
+        try {
+            return Span.current().setStatus(statusCode, description);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return this;
+        }
     }
 
     @Override
     public Span recordException(Throwable exception, Attributes additionalAttributes) {
-        return Span.current().recordException(exception, additionalAttributes);
+        try {
+            return Span.current().recordException(exception, additionalAttributes);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return this;
+        }
     }
 
     @Override
     public Span updateName(String name) {
-        return Span.current().updateName(name);
+        try {
+            return Span.current().updateName(name);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return this;
+        }
     }
 
     @Override
     public void end() {
-        Span.current().end();
+        try {
+            Span.current().end();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+        }
     }
 
     @Override
     public void end(long timestamp, TimeUnit unit) {
-        Span.current().end(timestamp, unit);
+        try {
+            Span.current().end(timestamp, unit);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+        }
     }
 
     @Override
     public SpanContext getSpanContext() {
-        return Span.current().getSpanContext();
+        try {
+            return Span.current().getSpanContext();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return null;
+        }
     }
 
     @Override
     public boolean isRecording() {
-        return Span.current().isRecording();
+        try {
+            return Span.current().isRecording();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return false;
+        }
     }
 
     @Override
     public boolean equals(Object obj) {
-        return Span.current().equals(obj);
+        try {
+            return Span.current().equals(obj);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return false;
+        }
     }
 
     @Override
     public int hashCode() {
-        return Span.current().hashCode();
+        try {
+            return Span.current().hashCode();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return -1;
+        }
     }
 
     @Override
     public String toString() {
-        return Span.current().toString();
+        try {
+            return Span.current().toString();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return "";
+        }
     }
 
 }
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/helper/InstrumenterActivator.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/helper/InstrumenterActivator.java
@@ -19,7 +19,13 @@ import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
 import io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
 public class InstrumenterActivator implements BundleActivator {
+
+    private static final TraceComponent tc = Tr.register(InstrumenterActivator.class);
+
     //The J2S exception occurs due to a call during the initialization of Instrumenter and InstrumenterBuilder
     @Override
     public void start(BundleContext ctx) throws Exception {
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
@@ -21,27 +21,47 @@ import org.osgi.service.component.annotations.ConfigurationPolicy;
 import io.openliberty.restfulWS.client.ClientAsyncTaskWrapper;
 import io.opentelemetry.context.Context;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
 /**
  * Ensures that the OTel context is used when either JAX-RS client or MP Rest Client makes an async request.
  */
 @Component(configurationPolicy = ConfigurationPolicy.IGNORE)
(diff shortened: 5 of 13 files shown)


benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/resources/io/openliberty/microprofile/telemetry/internal/resources/MPTelemetry.nlsprops
```diff
@@ -29,3 +29,7 @@
 CWMOT5100.tracing.is.disabled=CWMOT5100I: The MicroProfile Telemetry Tracing feature is enabled but not configured to generate traces for the {0} application.
 CWMOT5100.tracing.is.disabled.explanation=The MicroProfile Telemetry Tracing feature needs to be configured to create traces and export them to a tracing system.
 CWMOT5100.tracing.is.disabled.useraction=To enable tracing, specify otel.sdk.disabled=false in any of the configuration sources accessible through MicroProfile Config. Further configuration might be required to allow the MicroProfile Telemetry Tracing feature to export traces to your tracing system.
+
+CEMOT5101.telemetry.error=CEMOT5101: An exception was thrown from MicroProfile Telemetry. See the assocaited exception: {0}.
```
helyarp: Suggestion: Remove "See the assocaited exception:"

If you want to include the sentence:
- Either remove `assocaited` or change it to `associated`.
- Consider replacing "exception" with "error". We try to use "error" rather than "exception" to simplify message text. But, if "exception" is needed to clarify what was thrown, then keep "exception".

dev/io.openliberty.microprofile.telemetry.1.0.internal/resources/io/openliberty/microprofile/telemetry/internal/resources/MPTelemetry.nlsprops
```diff
@@ -29,3 +29,7 @@
...
+CEMOT5101.telemetry.error.explanation=An exception was thrown from within MicroProfile Telemetry and logged.
+CEMOT5101.telemetry.error.useraction=Examine the wrapped exception and perform any appropriate actions.
```
helyarp: If possible, explain what the user needs to examine in the wrapped exception. Will the info provided to the user explain the "appropriate action"?

helyarp (commented):
Suggested changes

dev/io.openliberty.microprofile.telemetry.1.0.internal/resources/io/openliberty/microprofile/telemetry/internal/resources/MPTelemetry.nlsprops
benjamin-confino: Hello @helyarp thank you for the suggestions. 

Happy to replace "exception" with "error" or if you prefer you can completely rewrite the message to whatever you like.

dev/io.openliberty.microprofile.telemetry.1.0.internal/resources/io/openliberty/microprofile/telemetry/internal/resources/MPTelemetry.nlsprops
benjamin-confino: This one might not be possible, the goal of this PR is to ensure that an exception in open telemetry never stops the app from running. As a result this message needs to be valid for any possible exception that open telemetry might throw

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -68,41 +68,44 @@ public class OpenTelemetryProducer {
...
+            //Operations on a Tracer, or on Spans have no side effects and do nothing
+            ComponentMetaData cData = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData();
+            String applicationName = cData.getJ2EEName().getApplication();
+            Tr.info(tc, "CWMOT5100.tracing.is.disabled", applicationName);
+
+            return new OpenTelemetryInfo(false, OpenTelemetry.noop());
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
         }
```
Azquelt: We need to return a no-op implementation here, rather than `null` (i.e. the same behaviour as if tracing is disabled).

You're not allowed to return `null` from an `ApplicationScoped` producer method anyway, so if we did return `null` the user would just get a more confusing error message.

dev/io.openliberty.microprofile.telemetry.1.0.internal/resources/io/openliberty/microprofile/telemetry/internal/resources/MPTelemetry.nlsprops
Azquelt: @benjamin-confino @helyarp How about something like this?

This is similar to what we wrote for MP Reactive Messaging.

```suggestion
CEMOT5101.telemetry.error=CEMOT5101: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
CEMOT5101.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
CEMOT5101.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
```

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter.java
```diff
@@ -124,6 +139,8 @@ public void filter(final ContainerRequestContext request, final ContainerRespons
 
         try {
             instrumenter.end(spanContext, request, response, null);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
         } finally {
```
Azquelt: We're more worried about unexpected exceptions from our own code than we are about unexpected exceptions from OTel. This try-catch should go around the whole method.

Same for any other methods where you do this, the try-catch should be around the whole method.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
```diff
@@ -19,47 +19,91 @@
...
+    private static final TraceComponent tc = Tr.register(BaggageProxy.class);
+
     @Override
     public int size() {
-        return Baggage.current().size();
+        try {
+            return Baggage.current().size();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return -1;
```
Azquelt: If we can't get the current baggage, the most appropriate thing to do is probably to pretend we have an empty `Baggage`, so I'd return `0` here.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
```diff
@@ -19,47 +19,91 @@
...
     }
 
     @Override
     public Map<String, BaggageEntry> asMap() {
-        return Baggage.current().asMap();
+        try {
+            return Baggage.current().asMap();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: Return `Collections.emptyMap()`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
```diff
@@ -19,47 +19,91 @@
...
     }
 
     @Override
     public BaggageBuilder toBuilder() {
-        return Baggage.current().toBuilder();
+        try {
+            return Baggage.current().toBuilder();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: `return Baggage.builder()`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
```diff
@@ -19,47 +19,91 @@
...
     }
 
     @Override
     public String toString() {
-        return Baggage.current().toString();
+        try {
+            return Baggage.current().toString();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: `null` is likely to cause further problems. It's probably preferable to return an empty string here.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -165,24 +170,34 @@ private boolean checkDisabled(Map<String, String> oTelConfigs) {
...
             }
+            //Metrics and logs are disabled by default
+            telemetryProperties.put(CONFIG_METRICS_EXPORTER_PROPERTY, "none");
+            telemetryProperties.put(CONFIG_LOGS_EXPORTER_PROPERTY, "none");
+            telemetryProperties.put(ENV_METRICS_EXPORTER_PROPERTY, "none");
+            telemetryProperties.put(ENV_LOGS_EXPORTER_PROPERTY, "none");
+            return telemetryProperties;
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: `return Collections.emptyMap();`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -165,24 +170,34 @@ private boolean checkDisabled(Map<String, String> oTelConfigs) {
...
     }
 
     @Produces
     public Tracer getTracer(OpenTelemetry openTelemetry) {
-        return openTelemetry.getTracer(INSTRUMENTATION_NAME);
+        try {
+            return openTelemetry.getTracer(INSTRUMENTATION_NAME);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: Return a no-op `Tracer`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -200,6 +215,11 @@ public Baggage getBaggage() {
     @ApplicationScoped
     @Produces
     public OpenTelemetry getOpenTelemetry(OpenTelemetryInfo openTelemetryInfo) {
-        return openTelemetryInfo.getOpenTelemetry();
+        try {
+            return openTelemetryInfo.getOpenTelemetry();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: Return a no-op `OpenTelemetry`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SPIMetaData.java
```diff
@@ -27,24 +27,34 @@
...
+            beans.add(OpenTelemetryProducer.class);
+            if (!AgentDetection.isAgentActive()) {
+                beans.add(WithSpanInterceptor.class);
+            }
+            beans.add(TelemetryClientFilter.class);
+            beans.add(TelemetryContainerFilter.class);
+            return beans;
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return new HashSet();
```
Azquelt: I don't think it's necessary to add it for this class. We always return the same things and this error would occur at startup.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
```diff
@@ -17,74 +17,142 @@
...
+    private static final TraceComponent tc = Tr.register(SpanProxy.class);
+
     @Override
     public <T> Span setAttribute(AttributeKey<T> key, T value) {
-        return Span.current().setAttribute(key, value);
+        try {
+            return Span.current().setAttribute(key, value);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: `return this` is probably better, since we are a `Span` and the point of returning `Span` is to allow chaining.

Returning `null` is likely to cause knock-on problems.

This needs fixed throughout this class.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
```diff
@@ -17,74 +17,142 @@
...
     }
 
     @Override
     public String toString() {
-        return Span.current().toString();
+        try {
+            return Span.current().toString();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt: Return empty string here.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
```diff
@@ -55,14 +58,22 @@ public WithSpanInterceptor() {
...
                                                                                                                            MethodRequest::getMethod,
                                                                                                                            new WithSpanParameterAttributeNamesExtractor(),
                                                                                                                            MethodRequest::getArgs);
 
-        this.instrumenter = builder.addAttributesExtractor(attributesExtractor)
+            instrumenter = builder.addAttributesExtractor(attributesExtractor)
                         .buildInstrumenter(methodRequest -> spanKindFromMethod(methodRequest.getMethod()));
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            instrumenter = null;
```
Azquelt: Other parts of this class need updating to cope with `instrumenter` possibly being `null`.

Obviously it shouldn't happen, but if it did then we'd bombard the user with lots of `NullPointerException` messages.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
```diff
@@ -75,52 +86,63 @@ private static SpanKind spanKindFromMethod(Method method) {
...
+            } finally {
+                if (scope != null) {
+                    scope.close();
+                }
             }
-        }
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
+        }
```
Azquelt: ~This gets a bit tricky. We're an interceptor, so we should return the result of `invocationContext.proceed()` - but at this point we don't know if we've already called it or not.~

~I'd be tempted to make it several different try blocks so that we can put all the important parts apart from the call to `invocationContext.proceed()` in a try-catch.~

Actually this is all broken
- `invocationContext.proceed()` will throw an exception if the user's method (or a later interceptor) throws an exception
- Your catch block will catch that exception and log it as an internal telemetry error rather than rethrowing it
- If an exception is thrown by `proceed()`, the existing code won't call `instrumentation.end()`
  - I've opened #25640 as a separate defect for this, so you can close that if you fix it here

I think the right thing to do is probably to catch any exception from `proceed()` separately (without FFDC), pass it to `instrumentation.end()` and rethrow it at the end of the method.

Then you can put everything other than the rethrow and return lines inside a try-catch as you have done here. You'll probably need to manually call `FFDCFilter` to get the FFDC since both catch blocks will be catching `Exception` and you can't ignore one without ignoring both.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
```diff
@@ -75,52 +86,63 @@ private static SpanKind spanKindFromMethod(Method method) {
...
+                    String spanName = annotation.value();
+                    if (spanName.isEmpty()) {
+                        spanName = SpanNames.fromMethod(methodRequest.getMethod());
+                    }
+                    return spanName;
+                });
+            } catch (Exception e) {
+                Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+                return null;
+            }
```
Azquelt: Don't need to do this method since it's called within the try-catch added for `span()`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/helper/InstrumenterActivator.java
```diff
@@ -19,14 +19,24 @@
...
     public void start(BundleContext ctx) throws Exception {
         Class.forName(Instrumenter.class.getName());
         Class.forName(InstrumenterBuilder.class.getName());
         // Eagerly initialize context classes to avoid J2S exceptions
-        Context.root();
+        try {
+            Context.root();
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+        }
```
Azquelt: I don't think we should add a try-catch here. This would occur at startup, and I don't think we can do anything sensible if this fails.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -21,27 +21,47 @@
...
+
     /** {@inheritDoc} */
     @Override
     public Runnable wrap(Runnable r) {
-        return Context.current().wrap(r);
+        try {
+            return Context.current().wrap(r);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt, line 42: `return r;`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -21,27 +21,47 @@
...
 
     /** {@inheritDoc} */
     @Override
     public <T> Callable<T> wrap(Callable<T> c) {
-        return Context.current().wrap(c);
+        try {
+            return Context.current().wrap(c);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt, line 53: `return c;`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -21,27 +21,47 @@
...
     }
 
     @Override
     public <T> Supplier<T> wrap(Supplier<T> s) {
-        return Context.current().wrapSupplier(s);
+        try {
+            return Context.current().wrapSupplier(s);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
```
Azquelt, line 63: `return s;`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -21,27 +21,47 @@
...
     @Override
     public Runnable wrap(Runnable r) {
-        return Context.current().wrap(r);
+        try {
+            return Context.current().wrap(r);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
+        }
     }
```
Azquelt: `return r` rather than `null`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -21,27 +21,47 @@
...
     @Override
     public <T> Callable<T> wrap(Callable<T> c) {
-        return Context.current().wrap(c);
+        try {
+            return Context.current().wrap(c);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
+        }
     }
```
Azquelt: Return `c` rather than `null`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientAsyncTaskWrapper.java
```diff
@@ -21,27 +21,47 @@
...
     @Override
     public <T> Supplier<T> wrap(Supplier<T> s) {
-        return Context.current().wrapSupplier(s);
+        try {
+            return Context.current().wrapSupplier(s);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return null;
+        }
     }
```
Azquelt: Return `s` rather than `null`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
```diff
@@ -96,13 +108,18 @@ public void filter(final ClientRequestContext request) {
...
+                        if (instrumenter.shouldStart(parentContext, request)) {
+                            Context spanContext = instrumenter.start(parentContext, request);
+                            request.setProperty(configString + "context", spanContext);
+                            request.setProperty(configString + "parentContext", parentContext);
+                        }
+                        return null;
+                    } catch (Exception e) {
+                        Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+                        return null;
                     }
```
Azquelt: Makes more sense to have a single `return null;` outside the catch block.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryClientFilter.java
```diff
@@ -114,17 +131,22 @@ public void filter(final ClientRequestContext request, final ClientResponseConte
...
+                            instrumenter.end(spanContext, request, response, null);
+                        } finally {
+                            request.removeProperty(configString + "context");
+                            request.removeProperty(configString + "parentContext");
+                        }
+                        return null;
+                    } catch (Exception e) {
+                        Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+                        return null;
                     }
```
Azquelt: Can move this `return null;` outside the `catch` as well

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter.java
```diff
@@ -73,17 +78,22 @@ public TelemetryContainerFilter() {
     @Inject
     public TelemetryContainerFilter(final OpenTelemetryInfo openTelemetry) {
         if (openTelemetry.getEnabled() && !AgentDetection.isAgentActive()) {
-            InstrumenterBuilder<ContainerRequestContext, ContainerResponseContext> builder = Instrumenter.builder(
+            try {
+                InstrumenterBuilder<ContainerRequestContext, ContainerResponseContext> builder = Instrumenter.builder(
```
Azquelt: Probably better to put the `if` inside the `try`

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryServletRequestListener.java
```diff
@@ -23,13 +26,20 @@
...
-            scope.close();
+            try {
+                scope.close();
+            } catch (Exception e) {
+                Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+                return;
+            }
             sre.getServletRequest().removeAttribute(TelemetryContainerFilter.SPAN_SCOPE);
         }
     }
```
Azquelt: Expand this `try-catch` to cover the whole method.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/package-info.java
```diff
@@ -13,7 +13,7 @@
 /**
  * @version 1.0
  */
-@TraceOptions(traceGroup = "TELEMETRY")
+@TraceOptions(traceGroup = "TELEMETRY", messageBundle = "io.openliberty.microprofile.telemetry.internal.resources.MPTelemetry")
 @org.osgi.annotation.versioning.Version("1.0")
 package io.openliberty.microprofile.telemetry.internal.rest;
```
Azquelt: I think you need this change in the `helper` package as well.

dev/io.openliberty.microprofile.telemetry.1.0.internal/resources/io/openliberty/microprofile/telemetry/internal/resources/MPTelemetry.nlsprops
helyarp: @Azquelt Much better. I like the proposed changes.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
benjamin-confino: The only other ref to instrumenter is inside a try/catch already, I'm not sure what you mean by this.

Do you want an 
if(instrumenter == null) {
            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
            return null;
 }
 
 To avoid an NPE?

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
benjamin-confino: > I think the right thing to do is probably to catch any exception from proceed() separately (without FFDC), pass it to instrumentation.end() and rethrow it at the end of the method.

Why not just call instrumentation.end() from the finally block? There already is a try block right where we'd want it.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/OpenTelemetryProducer.java
```diff
@@ -68,41 +69,44 @@ public class OpenTelemetryProducer {
...
+            //The absence of an installed SDK is a “no-op” API
+            //Operations on a Tracer, or on Spans have no side effects and do nothing
+            ComponentMetaData cData = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData();
+            String applicationName = cData.getJ2EEName().getApplication();
+            Tr.info(tc, "CWMOT5100.tracing.is.disabled", applicationName);
+
+            return new OpenTelemetryInfo(false, OpenTelemetry.noop());
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            return OpenTelemetry.noop();
```
Azquelt: The return type of this method is `OpenTelemetryInfo`, so I think you need to wrap the noop object before returning it.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/BaggageProxy.java
Azquelt: This still needs changed.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/SpanProxy.java
```diff
@@ -17,74 +17,142 @@
...
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
 /**
  * This proxy class redirects method calls to Span.current(), by doing so it allows people to use @Inject Span and get an object which will not become obsolete.
  */
 public class SpanProxy implements Span {
 
+    private static final TraceComponent tc = Tr.register(SpanProxy.class);
+
```
Azquelt, line 29: Not related to this PR, but this class should probably override `makeCurrent()` as a no-op.

Setting this class as the current span would be bad as it would redirect to itself and cause a stack overflow.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
Azquelt: Ah, sorry, there's a separate bug in the original code.

If the method returns a value, that value should be passed to `instrumentation.end()`. If the method throws an exception, that exception should be passed to `instrumentation.end()` instead. [Javadoc here](https://www.javadoc.io/doc/io.opentelemetry.instrumentation/opentelemetry-instrumentation-api/1.19.0/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.html)

This would be a release bug I think and needs a test, so maybe we should do it in a separate PR.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
Azquelt: This change still needs done.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter.java
```diff
@@ -95,16 +105,21 @@ public void filter(final ContainerRequestContext request) {
...
+
+                Context spanContext = instrumenter.start(parentContext, request);
+                Scope scope = spanContext.makeCurrent();
+                request.setProperty(SPAN_CONTEXT, spanContext);
+                request.setProperty(SPAN_PARENT_CONTEXT, parentContext);
+                request.setProperty(SPAN_SCOPE, scope);
+            }
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            this.instrumenter = null;
```
Azquelt: Why do we want to set the instrumenter to `null` here?

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryContainerFilter.java
```diff
@@ -113,17 +128,20 @@ public void filter(final ContainerRequestContext request, final ContainerRespons
...
+            }
 
-        try {
+        
             instrumenter.end(spanContext, request, response, null);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
         } finally {
             request.removeProperty(REST_RESOURCE_CLASS);
             request.removeProperty(REST_RESOURCE_METHOD);
```
Azquelt: Semantics are changed here. Previously the `removeProperty(` lines did not run if `spanContext` or `instrumenter` were `null` and now they do.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryServletRequestListener.java
```diff
@@ -23,15 +26,22 @@
...
-        Scope scope = (Scope) sre.getServletRequest().getAttribute(TelemetryContainerFilter.SPAN_SCOPE);
-        if (scope != null) {
-            scope.close();
+            Scope scope = (Scope) sre.getServletRequest().getAttribute(TelemetryContainerFilter.SPAN_SCOPE);
+            if (scope != null) {
+
+                scope.close();
+            }
             sre.getServletRequest().removeAttribute(TelemetryContainerFilter.SPAN_SCOPE);
+        }  catch (Exception e) {
```
Azquelt: Previously `sre.getServletRequest().removeAttribute(TelemetryContainerFilter.SPAN_SCOPE);` was inside the `if` block, now it is outside.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
Azquelt: If this part fails, we don't want to log further exceptions later on which are as a result of the same problem.

If `instrumenter` is `null`, we don't want the `span()` method to try to use it.

I would change this part (line 95):
```java
boolean shouldStart = instrumenter.shouldStart(parentContext, methodRequest);
```

to this:
```java
boolean shouldStart = instrumenter != null && instrumenter.shouldStart(parentContext, methodRequest);
```

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
```diff
@@ -55,14 +58,22 @@ public WithSpanInterceptor() {
...
 
-        this.instrumenter = builder.addAttributesExtractor(attributesExtractor)
+            instrumenter = builder.addAttributesExtractor(attributesExtractor)
                         .buildInstrumenter(methodRequest -> spanKindFromMethod(methodRequest.getMethod()));
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CEMOT5101.telemetry.error", e));
+            instrumenter = null;
+        } finally {
+            this.instrumenter = instrumenter;
+        }
```
Azquelt: I don't think there's a need for a `finally` block here (or for the local variable `instrumenter`)

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/cdi/WithSpanInterceptor.java
Azquelt: Though seeing the comment further down this file, maybe we leave this file alone for the moment since it has other issues.

helyarp (commented):
Looks good!

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/helper/InstrumenterActivator.java
```diff
@@ -19,7 +19,13 @@
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
 import io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
 public class InstrumenterActivator implements BundleActivator {
+
+    private static final TraceComponent tc = Tr.register(InstrumenterActivator.class);
+
```
Azquelt, line 28: Not needed, but not an issue either.

dev/io.openliberty.microprofile.telemetry.common/resources/io/openliberty/microprofile/telemetry/common/resources/MPTelemetry.nlsprops
```diff
@@ -37,4 +37,9 @@ CWMOT5000.cannot.get.globalopentelemetry.useraction=Use CDI to inject the OpenTe
 
 CWMOT5001.cannot.set.globalopentelemetry=CWMOT5001E: Setting the GlobalOpenTelemetry object is not supported.
 CWMOT5001.cannot.set.globalopentelemetry.explanation=The GlobalOpenTelemetry class is not supported because no global OpenTelemetry object exists. Instead, each application must use CDI to inject the OpenTelemetry object for that application.
-CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
\ No newline at end of file
+CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
+
+CEMOT5102.telemetry.error=CEMOT5102: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
+CEMOT5102.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
+CEMOT5102.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
```
Azquelt:
```suggestion
CWMOT5002.telemetry.error=CWMOT5002E: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
CWMOT5002.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
CWMOT5002.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
```

`CWMOT` is the message prefix block. We would put the `E` to signify an error on the end of the message ID.

In addition, the next free ID is 5002. (We could also use 5101 since we have a 5100, but I think extending the 500x series makes more sense.)

References to the message ID unfortunately need to be updated throughout.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/package-info.java
```diff
@@ -10,7 +10,8 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
-@TraceOptions(traceGroup = "TELEMETRY")
+@TraceOptions(traceGroup = "TELEMETRY", messageBundle = "io.openliberty.microprofile.telemetry.common.resources.MPTelemetry")
+@org.osgi.annotation.versioning.Version("1.0")
```
Azquelt: At the moment, the version is declared in the bnd.bnd. We should either declare it here or there, but not in both places.

Azquelt (changes requested):

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry/internal/rest/TelemetryServletRequestListener.java
```diff
@@ -23,15 +26,21 @@
...
-        Scope scope = (Scope) sre.getServletRequest().getAttribute(TelemetryContainerFilter.SPAN_SCOPE);
-        if (scope != null) {
-            scope.close();
-            sre.getServletRequest().removeAttribute(TelemetryContainerFilter.SPAN_SCOPE);
+            Scope scope = (Scope) sre.getServletRequest().getAttribute(TelemetryContainerFilter.SPAN_SCOPE);
+            if (scope != null) {
+                scope.close();
+                sre.getServletRequest().removeAttribute(TelemetryContainerFilter.SPAN_SCOPE);
+            }
+        }  catch (Exception e) {
```
Azquelt: Indentation and double space

dev/io.openliberty.microprofile.telemetry.common/resources/io/openliberty/microprofile/telemetry/common/resources/MPTelemetry.nlsprops
```diff
@@ -37,4 +37,9 @@ CWMOT5000.cannot.get.globalopentelemetry.useraction=Use CDI to inject the OpenTe
 
 CWMOT5001.cannot.set.globalopentelemetry=CWMOT5001E: Setting the GlobalOpenTelemetry object is not supported.
 CWMOT5001.cannot.set.globalopentelemetry.explanation=The GlobalOpenTelemetry class is not supported because no global OpenTelemetry object exists. Instead, each application must use CDI to inject the OpenTelemetry object for that application.
-CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
\ No newline at end of file
+CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
+
+CWMOT5102.telemetry.error=CWMOT5102: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
+CWMOT5102.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
+CWMOT5102.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
```
Azquelt:
```suggestion
CWMOT5002.telemetry.error=CWMOT5002E: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
CWMOT5002.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
CWMOT5002.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
```

Fix number and add `E` to the message itself.

benjamin-confino:
#build

Azquelt (approved):
Thanks!

Azquelt:
#libby

benjamin-confino:
Build is green, will merge when the branch is open.

PR #25638, opened 2023-07-06
Log stacks on trace

benjamin-confino:
Errors in this test will involve things being called out of order, having the stacks so we can check for common classes calling things in the wrong order will be useful.


benjamin-confino:
#build

tevans78 (approved):

benjamin-confino:
One failure:  <br>[7/6/23, 10:04:19:501 PDT] 0000002b com.ibm.ejs.container.BusinessExceptionMappingStrategy       E CNTR0020E: EJB threw an unexpected (non-declared) exception during invocation of method "ejbPreDestroy" on bean "BeanId(StartupEvents#StartupEventsEJB.jar#StartupSingletonEJB, null)". Exception data: java.lang.AssertionError: EjbApplicationScopedBean#-920701738: EJB PreDestroy was observed after Shutdown

But that's the failure this extra logging is here to debug so no surprise we see it in the build.

PR #25639, opened 2023-07-06
Do not check for undefined behaviour

benjamin-confino:
I can't find anything in the spec saying what order these events should fire in, so lets not test for either now.


benjamin-confino:
#build

PR #25666, opened 2023-07-07
Temp fix for jvm bug detecting criu support

sebratton:
The JVM bug is breaking the @SkipIfCheckpointNotSupported annotation and allows the checkpoint FATS to execute on systems where not supported. This can be backed out when a fix is provided in a release JVM

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/probe/CriuSupport.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/probe/CriuSupport.java
@@ -4,7 +4,7 @@
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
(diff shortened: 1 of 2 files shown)


tjwatson (dismissed):

sebratton:
#build

sebratton:
#build 
#libby

dev/fattest.simplicity/src/componenttest/topology/impl/probe/CriuSupport.java
```diff
@@ -70,7 +79,7 @@ public static void main(String[] args) {
      * @param string
      */
     private static void err(String string) {
-        System.err.println("criuSupport> " + string);
+        System.err.println("string");
```
tjwatson: You meant to use the var `string` here and not the literal value of "string" here.

sebratton: Good catch. Thanks.

sebratton:
#build

sebratton:
#libby

tjwatson (approved):

sebratton:
All RTC build errors were known issues.

PR #25671, opened 2023-07-10
id type results for find and delete

njr-11:
Results for find-and-delete operations can be a list of the id type or an `Optional` of the id type.
This also covers some other result types such as rejecting invalid and testing list/optional of Object/unparameterized.


mswatosh (approved):

njr-11:
#build

PR #25676, opened 2023-07-11
update comment

benjamin-confino:
Update comment


Yas-Franco (approved):
Now in line with our inclusivity policy

PR #25683, opened 2023-07-11
Always collect server logs when checkpointValidate() fails

dazavala:
Method `LibertyServer.checkpointValidate()` collects server logs only when failed assertions are detected. The method also detects unexpected server behavior/errors, but it does not collect the server logs in these cases. This can result in test breaks that lack server logs when `checkpointValidate()` fails.  Modify `checkpointValidate()` to collect server logs when detecting both failed assertions and unexpected server behavior/errors.


dazavala:
#build
#libby

dazavala:
Liberty and Liberty InstantOn personal builds are good.  @sebratton These changes are ready pending review.

sebratton (approved):
LGTM

PR #25722, opened 2023-07-13
create test bucket for validation

njr-11:
This is the start of a test bucket to cover scenarios where Jakarta Validation constraints are applied to various types of entities in Jakarta Data.  We expect to see Jakarta Persistence automatically apply validation to the entity in the case of `@Entity` and possibly Java class entity, but that isn't happening.  So currently the bucket manually attempts validation showing at least that much works.


njr-11:
#build

njr-11:
resubmitting build due to infrastructure issue:
#build

njr-11:
Rebased and added a commit to switch back to java:module now that the injection NullPointerException is fixed
#libby
#build

jmstephensgit (approved):

PR #25725, opened 2023-07-14
Resubmit update to Weld 5.1.1 but use SP1 instead of Final

benjamin-confino:
A test failure was burred behind ant going into an infinite recursive loop. Fortunately someone else had a problem from the same weld change, submitted a bug report, and weld SP1 was released to revert that change. Therefore we will update to 5.1.1.SP1 instead of final.


benjamin-confino:
#build

benjamin-confino:
Build has not yet started

Azquelt:
Includes the reverted changes from #25572 and #25676

benjamin-confino:
#libby

Azquelt (approved):

benjamin-confino:
Previous build failed because artifacts needed uploading.

benjamin-confino:
#build

PR #25774, opened 2023-07-20
Fix TCKUtilities.useArtifactory()

KyleAure:
Fixes the following issues with the method `TCKUtilities.useArtifactory()`

## String "null" from gradle

If I run a TCK test suite with an Artifactory repository I see the following in logs: 
```
[07/20/2023 14:39:03:799 CDT] 001 TCKUtilities                   useArtifactory                 I Use artifactory = true (artifactory.download.server=na-public.artifactory.swg-devops.com, artifactory.force.external.repo=null)
```

If I run a TCK test suite without an Artifactory repository I see the following in logs: 
```
[07/20/2023 17:13:32:238 CDT] 001 TCKUtilities                   useArtifactory                 I Use artifactory = true (artifactory.download.server=null, artifactory.force.external.repo=null)
[07/20/2023 17:13:32:636 CDT] 001 TCKUtilities                   downloadFile                   S Error downloading file https://null/artifactory/wasliberty-maven-remote/org/apache/maven/wrapper/maven-wrapper/3.2.0/maven-wrapper-3.2.0.jar: java.net.UnknownHostException: null
```

There was no change in behavior because gradle set `artifactory.download.server=null`
https://github.com/OpenLiberty/open-liberty/blob/683ef861e09d708fddb1cca8f6ef7d742ebd1b3f/dev/wlp-gradle/subprojects/fat.gradle#L436

## String ${env.artifactory.download.server} from ant

Once the above issue is fixed we then run into an issue from ant where it will look for the property `artifactory.download.server` which if it does not come from gradle, could have come from one of our various `.properties` files in autoFVT.

If it does not find the property from gradle or a `.properties` file. Ant will set the value to `${env.artifactory.download.server}`

https://github.com/OpenLiberty/open-liberty/blob/683ef861e09d708fddb1cca8f6ef7d742ebd1b3f/dev/fattest.simplicity/autoFVT-defaults/src/ant/launch.xml#L334-L338

If I run a TCK test suite without an Artifactory repository I see the following in logs: 
```
[07/20/2023 17:13:32:238 CDT] 001 TCKUtilities                   useArtifactory                 I Use artifactory = true (artifactory.download.server=${env.artifactory.download.server}, artifactory.force.external.repo=)
[07/20/2023 17:13:32:636 CDT] 001 TCKUtilities                   downloadFile                   S Error downloading file https://${env.artifactory.download.server}/artifactory/wasliberty-maven-remote/org/apache/maven/wrapper/maven-wrapper/3.2.0/maven-wrapper-3.2.0.jar: java.net.UnknownHostException: ${env.artifactory.download.server}
```

## Fuzzy logic from java

Once the above two issues are correct we have fuzzy logic in Java. 

https://github.com/OpenLiberty/open-liberty/blob/683ef861e09d708fddb1cca8f6ef7d742ebd1b3f/dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java#L675

| haveArtifactoryServer | forceExternal | expected | useArtifactory() |
| ---- | ---- | ---- | ---- |
| True | True | True | **False** |
| True | False | True | True |
| False | True | **????** | False |
| False | False | False | False |

If we both have an haveArtifactoryServer and  forceExternal we expect true, but get false. 
If we don't have haveArtifactoryServer, but request to forceExternal, we get false which is unexpected and should probably be an error.


KyleAure:
#libby
#build

Azquelt (changes requested):

dev/fattest.simplicity/test/componenttest/topology/utils/tck/TCKUtilitiesTest.java
```diff
@@ -0,0 +1,128 @@
...
+public class TCKUtilitiesTest {
+
+    /**
+     * Test method TCKUtilities.useArtifactory()
+     * Variables: artifactory.download.server, artifactory.force.external.repo
+     */
+    @Test
+    public void usingArtifactory() {
+        final String forceExternalProp = TCKUtilities.FAT_TEST_PREFIX + TCKUtilities.ARTIFACTORY_FORCE_EXTERNAL_KEY;
+        final String artifactoryServerProp = TCKUtilities.FAT_TEST_PREFIX + TCKUtilities.ARTIFACTORY_SERVER_KEY;
```
Azquelt: Use capitals for constants

dev/fattest.simplicity/test/componenttest/topology/utils/tck/TCKUtilitiesTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        logTest(2, artifactoryServerProp, "null", forceExternalProp, "true");
+        System.clearProperty(artifactoryServerProp);
+        System.setProperty(forceExternalProp, "true");
+        try {
+            TCKUtilities.useArtifactory();
+            fail("Should have thrown an IllegalStateException");
+        } catch (IllegalStateException e) {
+            //pass
+        }
```
Azquelt: Test should expect `false`: `useArtifactory()` should always return `false` if `forceExternal` is `true`.

dev/fattest.simplicity/test/componenttest/topology/utils/tck/TCKUtilitiesTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        logTest(4, artifactoryServerProp, "example.com", forceExternalProp, "null");
+        System.setProperty(artifactoryServerProp, "example.com");
+        System.clearProperty(forceExternalProp);
+        assertTrue(TCKUtilities.useArtifactory());
+
+        logTest(5, artifactoryServerProp, "example.com", forceExternalProp, "true");
+        System.setProperty(artifactoryServerProp, "example.com");
+        System.setProperty(forceExternalProp, "true");
+        assertTrue(TCKUtilities.useArtifactory());
```
Azquelt: Test should expect `false`

dev/fattest.simplicity/test/componenttest/topology/utils/tck/TCKUtilitiesTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        logTest(8, artifactoryServerProp, "", forceExternalProp, "true");
+        System.setProperty(artifactoryServerProp, "");
+        System.setProperty(forceExternalProp, "true");
+        try {
+            TCKUtilities.useArtifactory();
+            fail("Should have thrown an IllegalStateException");
+        } catch (IllegalStateException e) {
+            //pass
+        }
```
Azquelt: Test should expect `false`

dev/fattest.simplicity/test/componenttest/topology/utils/tck/TCKUtilitiesTest.java
```diff
@@ -0,0 +1,128 @@
...
+     * Test method: TCKUtilities.getArtifactoryServer()
+     * Variables: artifactory.download.server
+     */
+    @Test
+    public void getArtifactoryServer() {
+        final String artifactoryServerProp = TCKUtilities.FAT_TEST_PREFIX + TCKUtilities.ARTIFACTORY_SERVER_KEY;
+
+        System.clearProperty(artifactoryServerProp);
+        assertNull(TCKUtilities.getArtifactoryServer());
+        assertTrue(Objects.isNull(TCKUtilities.getArtifactoryServer()));
```
Azquelt, line 92: What is the point of this double check?

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -704,7 +723,27 @@ static String getArtifactoryUser() {
...
+     * @return        null if the string was invalid, itself otherwise.
+     */
+    static String validateStringProperty(String result) {
+        if (result == null)
+            return result;
+
+        if (result.startsWith("${"))
+            return null;
+
+        return result;
```
Azquelt: Validating usually means "check it and throw an exception if it's wrong", it seems weird that a method named "validate" is making changes to the data.

Suggest renaming it to something like `normalizeStringProperty` or `fixupStringProperty`

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
```diff
@@ -665,28 +665,47 @@ public static Properties updateWrapperPropertiesFile(File wrapperPropertiesFile,
...
+        boolean forceExternal = isArtifactoryForceExternal();
         String artifactoryServer = getArtifactoryServer();
-        boolean haveArtifactoryServer = (artifactoryServer != null && !artifactoryServer.isEmpty());
 
-        boolean useArtifactory = haveArtifactoryServer && !forceExternal;
+        boolean useArtifactory = (Objects.nonNull(artifactoryServer) && !artifactoryServer.isEmpty());
+
+        if (forceExternal && !useArtifactory) {
+            throw new IllegalStateException("Forced external artifactory usage, but no server was set");
+        }
```
Azquelt: This change is wrong. The point of `forceExternal` is to force the use of external servers (i.e. maven central, as opposed to an internal artifactory server), even if an artifactory server address is configured.

dev/fattest.simplicity/src/componenttest/topology/utils/tck/TCKUtilities.java
Azquelt: It might make sense for this method to also convert an empty string to `null`.

Azquelt (changes requested):

Azquelt:
I'm surprised we didn't see the issue with unset properties arriving as the string "null" rather than the value `null` during testing. I thought I'd checked that case but I must have made a mistake.

The changes to the gradle to avoid setting environment variables to the string "null" makes sense, as do the changes to `TCKUtilities` to recognise when the property consists of an ant property which hasn't been expanded correctly.

The change in the logic in `TCKUtilities.useArtifactory()` I think is wrong though. `forceExternal` being set indicates that we should ignore any artifactory server that might be configured and go straight to the external maven central servers. It makes it easier to test without artifactory without having to remove all the artifactory config properties.

The intended behaviour is:
| haveArtifactoryServer | forceExternal | useArtifactory() |
| ---- | ---- | ---- |
| True | True | False |
| True | False | True |
| False | True | False |
| False | False | False |

KyleAure:
@Azquelt Thanks for the explanation of the property `forceExternal` In my mind Artifactory was the external repo to Maven Central.  

I think overall this logic is hard to follow would you mind if I made a suggestion?  We could instead follow a similar logic to what I use for artifactory usage for docker.... something like: 

```java
     /**
     * Checks whether we are configured to use artifactory or not
     *
     * @return {@code true} if we are configured to use artifactory, {@code false} if not
     */
    static boolean useArtifactory() {
        boolean result;
        String reason;
        
        do {
            //State 1: fat.test.use.artifactory.substitution should always be honored first
            if( System.getProperty("fat.test.use.artifactory.substitution") != null ) {
                result = Boolean.getBoolean("fat.test.use.artifactory.substitution");
                reason = "fat.test.use.remote.docker set to " + result;
                break;
            }
            
            //State 2: Artifactory host was configured, assume user wants to use artifactory
            if( System.getProperty("fat.test.artifactory.download.server") != null ) {
                result = true;
                reason = "fat.test.artifactory.download.server set to " + System.getProperty("fat.test.artifactory.download.server");
                break;
            }

            //DEFAULT: User Maven Central for local builds, and Artifactory for RTC builds
            result = !FATRunner.FAT_TEST_LOCALRUN;
            reason = "fat.test.localrun set to " + FATRunner.FAT_TEST_LOCALRUN;

        } while (false);


        if ( result = true && System.getProperty("fat.test.artifactory.download.server") != null ) {
            throw new IllegalStateException("We need to use Artifactory for Maven dependencies, but no Artifactory server was configured.");
        }
        
        reason = result ? //
                        "Artifactory will be used. Reason: " + reason : //
                        "Maven Central will be used. Reason: " + reason;

        Log.info(c, "useArtifactory", reason);
        return result;
    }
```

`fat.test.use.artifactory.substitution`: Is a flag to force the decision True/False
`fat.test.artifactory.download.server`: If the server is configured then use it
Otherwise defer to where the test is being run

Obviously I would switch to using constants.

dev/fattest.simplicity/test/componenttest/topology/utils/tck/TCKUtilitiesTest.java
KyleAure: Test coverage to since we use `Objects.isNull` in the `useArtifactory` method.
🤷🏼 You never know it could return false.

Azquelt: Yep, missed the use of `isNull` in the other file.

Azquelt:
Had a discussion on this as the structure of the suggested change wasn't something I was familiar with:

1. Yes, it would probably be clearer if we changed it to do several checks with a comment on each
2. We should stick with `artifactory.force.external.repo` over `fat.test.use.artifactory.substitution` as that property is already in use to control the use of artifactory or not for downloading dependencies when compiling Open Liberty: https://github.com/OpenLiberty/open-liberty/blob/dbed0ecbdaf8c2ef62e1fcda25c3b4cd161ab609/dev/settings.gradle#L44

Azquelt (approved):

KyleAure:
#libby
#build

PR #25780, opened 2023-07-21
Build compatible extensions

### LINKED ISSUE #25977: Allow the SPI to register Build Compatible Extensions
**Describe the use case that you want to enable:**
We have an SPI to allow liberty components to easily register SPI extensions (as well as syntatic sugar for adding new BDAs and Beans). In CDI 4.0 a new type of extensions, build compatible extensions, has been introduced so the SPI should be expanded to support them. 

**Describe why this is important to you:**
We were contacted by another integrator who were only able to use Build Compatible Limitations on their end.

**Additional context**
Regular extensions have better performance than BCE, so for now we will be discouraging BCEs by making the SPI internal only, but should more requests like the above occur the code will be ready.

### LINKED ISSUE #25978: The SPI for registering CDI extensions and Beans will scan the entire archive without an extension
**Describe the bug**  
If the SPI for registering CDI extensions is used, the archive it is in will be treated as a normal CDI archive, which defaults to bean scanning mode "annotated" unless there is an extension in the bundle. This means that a class in the bundle with a bean defining annotation will become a bean even if it is not registered via the SPI.

**Steps to Reproduce**  
- Create a bundle that uses CDIExtensionMetadata
- Put a class in that bundle with a default CDI annotation like @ApplicationScoped
- Do not include that class in the set returned by addBeans()
- Do not include an extension in the bundle.

**Expected behavior**  
I would expect CDI to be unable to see the Bean

**Diagnostic information:**  
 - OpenLiberty Version: [e.g. 21.0.0.6 - 23.0.0.9]
 - Affected feature(s) [e.g. CDI-1.2+]

benjamin-confino:
Fixes #25977
Fixes #25978

What this does: 

- Moves the code that reads the CDI extension SPI out of CDIContainerImpl and into implementations of ExtensionArchiveProvider (EE7 through 9 redirect to common code in a Util class, EE10 does not)
- Modifies LiteExtensionProvider so it also processes SPIs. The end result is an extension archive that can contain both regular and build compatible extensions; as well as extra classes and annotations. Everything co-inhabits nicely. 
- Creates a new interface ExtensionContainer, if an ExtensionArchive contains one it will treat the extensions within like any extensions in an ExtensionArchive. This is a hack because the ExtensionArchive interface uses javax Extensions but LiteExtensionTranslator uses jakarta Extensions. 
- Splits the FAT test for the SPI into two. This prevents the original from using `cdi.internals-3.0` which would change what is being tested. `cdi.internals-3.0` is used on EE10 because the SPI for BCE currently is limited to our internal interface. (This was a surprisingly cumbersome process).


dev/io.openliberty.cdi.4.0.internal.weld/src/io/openliberty/cdi40/internal/weld/LiteExtensionArchive.java
```diff
@@ -51,6 +52,10 @@ public LiteExtensionArchive(CDIRuntime cdiRuntime, ClassLoader bceClassLoader, L
         this.buildCompatibleExtensions = buildCompatibleExtensions;
     }
 
+    public LiteExtensionArchive(CDIRuntime cdiRuntime, ClassLoader bceClassLoader, Set<Class<? extends BuildCompatibleExtension>> buildCompatibleExtensions) {
```
benjamin-confino: Remove

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ExtensionArchiveImpl.java
```diff
@@ -97,6 +102,29 @@ public Set<Supplier<Extension>> getSPIExtensionSuppliers() {
...
+
+        for (ExtensionContainer e : extensionContainers) {
+            for (Supplier<Object> supplier : e.getSPIExtensionSuppliers()) {
+                result.add(() -> {
+                    Object maybeExtension = null;
+                    try {
+                        maybeExtension = supplier.get();
+                        if (! (maybeExtension instanceof Extension)) {
+                            throw new IllegalArgumentException(maybeExtension.getClass().getCanonicalName()
+                                                           + " was registered as an extension via the WebSphereCDIExtensionMetaData interface. But it does not implement javax.enterprise.inject.spi.Extension");
```
benjamin-confino: Change this and previous to javax/jakarta

benjamin-confino:
#build

Azquelt:
The last two commits appear to be a separate change which needs its own issue and its own PR, particularly as it should be much simpler and easier to review and merge than this one.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchiveProvider.java
```diff
@@ -29,6 +38,11 @@ public interface ExtensionArchiveProvider {
      * @return a collection of additional extension archives to add
      * @throws CDIException if an unexpected exception is encountered
      */
-    Collection<ExtensionArchive> getArchives(CDIRuntime cdiRuntime, WebSphereCDIDeployment deployment) throws CDIException;
+    default Collection<ExtensionArchive> getDiscoveredArchives(CDIRuntime cdiRuntime, WebSphereCDIDeployment deployment) throws CDIException {
+        return Collections.emptySet();
+    }
```
Azquelt: This name change doesn't make sense. The archives added by implementors of this interface are not "discovered" - they're created anew.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchiveProvider.java
```diff
@@ -29,6 +38,11 @@ public interface ExtensionArchiveProvider {
...
+
+    ExtensionArchive newSPIExtensionArchive(CDIRuntime cdiRuntime, ServiceReference<CDIExtensionMetadata> sr,
+                                                    CDIExtensionMetadata webSphereCDIExtensionMetaData, WebSphereCDIDeployment applicationContext) throws CDIException;
```
Azquelt: This method doesn't make sense on this class. Please put it on its own interface.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchive.java
```diff
@@ -43,4 +43,7 @@ public interface ExtensionArchive extends CDIArchive {
      * @return a set of suppliers of extension objects
      */
     Set<Supplier<Extension>> getSPIExtensionSuppliers();
+
+    default void addLiteExtensionTranslator(Extension liteExtensionTranslator, Set<String> buildCompatibleExtensionClassNames) {};
```
Azquelt: This method is weird and takes a completely different approach to everything else on this interface. Every other option is set via the constructor, why would this one have a setter?

Also, like every method on an interface like this, it should have Javadoc.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ExtensionArchiveImpl.java
```diff
@@ -70,6 +78,7 @@ public boolean isExtClassesOnly() {
     public Set<String> getExtensionClasses() {
         Set<String> extensionClasses = super.getExtensionClasses();
         extensionClasses.addAll(spiExtensions);
+        extensionClasses.addAll(spiBuildCompatibleExtensions);
         return extensionClasses;
     }
```
Azquelt: What is `getExtensionClasses()` used for?

Is it correct to have build compatible extension classes in here?

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchive.java
Azquelt: Initially, I thought there was only one implementation of this interface so I was confused as to why we have two: the general purpose `ExtensionArchiveImpl` and the special case `ProbeExtensionArchive`.

This method would only work on `ExtensionArchiveImpl`, so I'm becoming more and more convinced it shouldn't be on this interface.

Azquelt: Furthermore, it doesn't strictly need to be a Lite Extension Translator extension, right? You're just adding a list of class names to be reported as "extension classes" and an extension instance you'd like to be added to the extension archive.

dev/io.openliberty.cdi.4.0.internal.weld/src/io/openliberty/cdi40/internal/archive/liberty/package-info.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+/**
+ * @version 1.0
+ */
+@org.osgi.annotation.versioning.Version("1.0")
+package io.openliberty.cdi40.internal.archive.liberty;
```
Azquelt: Why is this added?

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ExtensionArchiveImpl.java
Azquelt: Confusingly, there is also a method `CDIArchive.getBuildCompatibleExtensionClasses()`. However, it's currently used to determine that we need to create a lite extension translator for an application bundle.

dev/io.openliberty.jakarta.cdi.3.0/src/com/ibm/ws/cdi/impl/CDI30ExtensionSPIProvider.java
```diff
@@ -0,0 +1,38 @@
...
+import com.ibm.ws.cdi.internal.interfaces.WebSphereCDIDeployment;
+
+import io.openliberty.cdi.spi.CDIExtensionMetadata;
+
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+import org.osgi.framework.ServiceReference;
+
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE)
+public class CDI30ExtensionSPIProvider implements ExtensionArchiveProvider {
```
Azquelt: This shouldn't go in the API bundle

dev/com.ibm.ws.cdi.interfaces/src/com/ibm/ws/cdi/extension/CDIExtensionMetadataInternal.java
```diff
@@ -12,6 +12,9 @@
  *******************************************************************************/
 package com.ibm.ws.cdi.extension;
 
+import java.util.Collections;
+import java.util.Set;
+
 import io.openliberty.cdi.spi.CDIExtensionMetadata;
```
Azquelt: No real changes in this file

dev/io.openliberty.cdi.4.0.internal.weld/src/io/openliberty/cdi40/internal/archive/liberty/package-info.java
benjamin-confino: Remember the previous version with container classes. A left over from that which can be safely removed.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchive.java
benjamin-confino: You're correct that it doesn't strictly need to be a Lite Extension Translator. 

We could call it something like addConstructedExtension instead. As this method exists because we need to pass in Lite Extension Translator objects rather than just a class name.

benjamin-confino: It uses a setter simply because I thought it would be less ugly to call a setter from cdi 4.0 ExtensionArchiveProvider than to have the other three pass in nulls. 

If you'd prefer to use the constructor that would be an easy change.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchiveProvider.java
benjamin-confino: I was thinking of the fact we LiteExtensionProvider was creating archives for BCEs discovered through the classloader. I have no objection to changing the name though.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchiveProvider.java
benjamin-confino: Do we really need two interfaces for version specific code for generating extension archives? Perhaps rather than splitting the interface we should find a framing where it makes sense to have both methods?

benjamin-confino:
> The last two commits appear to be a separate change which needs its own issue and its own PR, particularly as it should be much simpler and easier to review and merge than this one.

They're in the same commit because I want to make sure the big changes in the earlier commits pass the new tests. And I wanted to write the new tests in the same branch to avoid a merge conflict with the tests from the earlier commit.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ExtensionArchiveImpl.java
benjamin-confino: `getExtensionClasses` is used by BDAFactory to pass in the extensionClassNames to BeanDeploymentArchiveImpl's constructor. After that there's code in scanForBeanClassNames that can include it in `beanClasses`. Which gets passed up to weld via `getBeanClasses`

I've read both the javadoc and the integration guide for what weld expects `getBeanClasses` and it was pretty terrible, but it looks like it expects results based on the bean scanning mode which is defined in the CDI spec, and checking the 4.0 spec I didn't see any reason to treat BCE and regular Es differently.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchive.java
Azquelt: I think `addConstructedExtension` is a good name.

As discussed, I think this does need to come off the interface because we have three implementations (I forgot `LiteExtensionArchive`) and only one will ever implement it.

I don't mind whether we move this to the constructor (which is a bit unwieldy because it has two arguments) or cast to the implementation to call the method.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchiveProvider.java
Azquelt: As discussed, I think we do, because we need exactly one implementation of this method, whereas we could have any number of the other method.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ExtensionArchiveImpl.java
Azquelt: As discussed, we don't think the spiBuildCompatibleExtensions should be added here.

However, we might need the `LiteExtensionTranslator` class added. I think some parts of BeanDeploymentArchiveImpl assume we can tell if an archive is an extension archive by seeing if it has either an Extension class or a BCE class.

dev/io.openliberty.jakarta.cdi.3.0/src/com/ibm/ws/cdi/impl/CDI30ExtensionSPIProvider.java
Azquelt: Turns out we don't need this class at all because the `com.ibm.ws.cdi.2.0.weld` bundle is also included in the CDI 3.0 feature.

dev/com.ibm.ws.cdi.extension_fat/test-bundles/cdi.spi.extension/src/com/ibm/ws/cdi/extension/spi/test/bundle/buildcompatible/BCExtraBean.java
```diff
@@ -0,0 +1,25 @@
...
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.cdi.extension.spi.test.bundle.buildcompatible;
+
+//Even though this is only run in EE10 we use javax otherwise the MANIFEST.MF imports both javax and jakarta.enterprise.context; then blows up when the transformer doesn't filter out the duplicate
+import javax.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class BCExtraBean {
```
Azquelt: I don't think this class is used anywhere

dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/apps/spi/SPIExtensionServlet.java
```diff
@@ -23,6 +23,7 @@
 import org.junit.Test;
 
 import com.ibm.ws.cdi.extension.spi.test.bundle.UnregisteredBean;
+import com.ibm.ws.cdi.extension.spi.test.bundle.buildcompatible.BCExtraBean;
```
Azquelt: No real changes in this file.

dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/apps/spi40/SPIExtensionServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    MyBeanInjectionString beanInjectedString;
+
+    @Inject
+    AppBean appBean;
+
+    @Inject
+    BCExtraBean bcExtraBean;
+
+    @Inject
+    CustomBDABean customBDABean;
```
Azquelt: Should really make all of these private

dev/io.openliberty.cdi.4.0.interfaces/bnd.bnd
```diff
@@ -0,0 +1,30 @@
...
+# 
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
+#*******************************************************************************
+-include= ~../cnf/resources/bnd/bundle.props
+
+bVersion=1.0
+	com.ibm.ws.cdi.extension
```
Azquelt: The bnd file is missing the bundle name and symbolic name and the project is missing the eclipse project files.

I'll make a commit to fix both of these.

dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/apps/spi40/SPIExtensionServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    @Test
+    public void testGetBeanClasses() {
+        assertEquals("An Interceptor registered via getBeanClasses in the SPI intercepted a normal scoped class registered via getBeanClasses Injection of a normal scoped class that was registered via getBeanClasses",
+                     beanInjectedString.toString());
+    }
+
+    @Test
+    public void testSPIInterceptorNormal() {
+        assertEquals("Produced injection", classString.toString());
+    }
```
Azquelt: This test looks like it tests a producer, not an interceptor.

dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/tests/CDI12NoExtensionSPITest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @AfterClass
+    public static void cleanup() throws Exception {
+        final String METHOD_NAME = "cleanup";
+        Log.info(CDI12ExtensionTest.class, METHOD_NAME, "Stopping the server.");
+        if (server.isStarted()) {
+            server.stopServer();
+        }
+        Log.info(CDI12ExtensionTest.class, METHOD_NAME, "Removing spi extension test user feature files.");
+        CDIExtensionRepeatActions.uninstallUserExtension(server, CDIExtensionRepeatActions.CDI_SPI_WITH_NO_EXTENSION_BUNDLE_ID);
```
Azquelt: Should go in a try-finally block to ensure we always remove the user feature

dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/tests/CDI40ExtensionSPITest.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Test the runtime extension to function correctly
+ *
+ * This is split from CDI12ExtensionSPITest because the test-bundle needs access to CDI internals. 
+ * getBuildCompatibleExtensions() is only avaialbe to internal components.
+ * Rather than enable access to internals for everything I split the test so we can test with and without that access.
+ */
+@RunWith(FATRunner.class)
+@Mode(TestMode.LITE)
+public class CDI40ExtensionSPITest extends FATServletClient {
```
Azquelt: As discussed, this can be a FULL mode test.

Azquelt:
The other thing missing was copyright dates on some modified files.

dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/tests/CDIExtensionRepeatActions.java
```diff
@@ -82,7 +84,7 @@ public static String getFeatureName(String bundleID, EEVersion eeVersion) {
         } else if (eeVersion == EEVersion.EE9) {
             name = bundleID + "-3.0";
         } else if (eeVersion == EEVersion.EE10) {
-            name = bundleID + "-3.0";
+            name = bundleID + "-4.0";
```
Azquelt, line 87: Update copyright date on this file.

dev/com.ibm.ws.cdi.extension_fat/fat/src/com/ibm/ws/cdi/extension/tests/FATSuite.java
```diff
@@ -28,10 +28,11 @@
                 CDI12ExtensionTest.class,
                 CDI12ExtensionSPIConstructorExceptionTest.class,
                 CDI12ExtensionSPITest.class,
+                CDI12NoExtensionSPITest.class,
+                CDI40ExtensionSPITest.class,
                 DynamicBeanExtensionTest.class,
                 ObserverTest.class,
-                WeldDevModeTest.class
-})
+                WeldDevModeTest.class})
```
Azquelt, line 35: Update copyright date on this file

dev/com.ibm.ws.cdi.extension_fat/test-bundles/cdi.spi.extension/src/com/ibm/ws/cdi/extension/spi/test/bundle/SPIMetaData.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020. 2021 IBM Corporation and others.
```
Azquelt: Undo date change, file has not changed.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ExtensionArchiveProvider.java
```diff
@@ -29,6 +33,7 @@ public interface ExtensionArchiveProvider {
      * @return a collection of additional extension archives to add
      * @throws CDIException if an unexpected exception is encountered
      */
-    Collection<ExtensionArchive> getArchives(CDIRuntime cdiRuntime, WebSphereCDIDeployment deployment) throws CDIException;
-
+    default Collection<ExtensionArchive> getArchives(CDIRuntime cdiRuntime, WebSphereCDIDeployment deployment) throws CDIException {
+        return Collections.emptySet();
+    }
```
Azquelt: This change is no longer needed since it's the only method on the interface.

The rest of the changes to this file (copyright date, imports) should also be reverted.

dev/io.openliberty.cdi.4.0.internal.weld/src/io/openliberty/cdi40/internal/weld/CDI40ExtensionSPIProvider.java
```diff
@@ -12,35 +12,54 @@
...
 import jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension;
+import jakarta.enterprise.inject.spi.Extension;
 
 /**
  * Provides the weld-lite-extension-translator extension when the deployed application has build compatible extensions
  */
 @Component(configurationPolicy = ConfigurationPolicy.IGNORE)
-public class LiteExtensionProvider implements ExtensionArchiveProvider {
-    private static final TraceComponent tc = Tr.register(LiteExtensionProvider.class);
+public class CDI40ExtensionSPIProvider implements ExtensionArchiveProvider, ExtensionArchiveFactory {
```
Azquelt: Can you move the class javadoc to the `getArchives()` method please? It only really applies to that method.

dev/com.ibm.ws.cdi.interfaces/src/io/openliberty/cdi/spi/CDIExtensionMetadata.java
```diff
@@ -68,6 +68,8 @@ public interface CDIExtensionMetadata {
      * All classes returned by this method will be will be found by CDI during type discovery
      * so that they can then be used as beans (or interceptors etc. if annotated as such) by the application.
      * All classes must be in the same archive as your CDIExtensionMetadata.
+     *
+     * If a bean registered via this does not have a bean defining annotation it will default to a scope of @Dependent
      */
```
Azquelt: As this is SPI, the Javadoc is published so we should format it properly.
```suggestion
     * All classes must be in the same archive as your CDIExtensionMetadata.
     * <p>
     * If a bean registered via this does not have a bean defining annotation it will default to a scope of {@code @Dependent}
     */
```

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ExtensionArchiveImpl.java
```diff
@@ -29,11 +29,19 @@
 
 public class ExtensionArchiveImpl extends CDIArchiveImpl implements ExtensionArchive {
 
-    private static final TraceComponent tc = Tr.register(ExtensionArchiveImpl.class);
+    protected static final TraceComponent tc = Tr.register(ExtensionArchiveImpl.class);
```
Azquelt: Why does this need to be `protected`? Usually a subclass would have its own `TraceComponent`.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ExtensionArchiveImpl.java
```diff
@@ -34,6 +34,14 @@ public class ExtensionArchiveImpl extends CDIArchiveImpl implements ExtensionArc
     private final ExtensionContainerInfo extensionContainerInfo;
     private Set<String> spiExtensions = null;
 
+    //Store these seperately so we don't try to make regular extensions out of them.
+    private Set<String> preConstructedExtensionNames = new HashSet<String>();
```
Azquelt, line 38: We add class names to this, but we never retrieve them.

benjamin-confino:
#build

Azquelt (dismissed):

benjamin-confino:
#build

benjamin-confino:
#build

tevans78 (approved):
@Azquelt already approved the first 8 commits. I'm approving the last one.

benjamin-confino:
previous build was at commit b77e43bbb379399040d2aca2a87e8d423ac73d8c, before the feature update to export the new interface

benjamin-confino:
#build

benjamin-confino:
Only failures were in com.ibm.ws.jpa.tests.container_modifyconfig_fat FAT tests. I see injection but it looks like this bucket has no CDI injection.

benjamin-confino:
#libby

PR #25787, opened 2023-07-21
experiment with bean for default instance

njr-11:
Supplies a bean via the CDI extension to attempt to provide the default ManagedExecutorService instance for `@Inject ManagedExecutorService`, following along the pattern that will be needed for configured instances.  The test case remains disabled because it isn't working yet.

### DIFF
--- a/dev/io.openliberty.concurrent.internal.cdi/src/io/openliberty/concurrent/internal/cdi/ConcurrencyExtension.java
+++ b/dev/io.openliberty.concurrent.internal.cdi/src/io/openliberty/concurrent/internal/cdi/ConcurrencyExtension.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2021 IBM Corporation and others.
+ * Copyright (c) 2021,2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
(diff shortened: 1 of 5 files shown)


njr-11:
#build
#libby

njr-11:
#build

dev/io.openliberty.concurrent.internal.cdi/src/io/openliberty/concurrent/internal/cdi/ConcurrencyExtension.java
```diff
@@ -1,34 +1,93 @@
...
+            Set<Annotation> qualifiers = injectionPoint.getQualifiers();
+            executorQualifiers.add(qualifiers);
+        }
+    }
+
+    public void afterBeanDiscovery(@Observes AfterBeanDiscovery event) {
+        if (ConcurrencyExtensionMetadata.eeVersion.getMajor() >= 11) {
+            for (Iterator<Set<Annotation>> it = executorQualifiers.iterator(); it.hasNext();) {
+                Set<Annotation> qualifiers = it.next();
+                it.remove();
```
KyleAure, line 78: I'm curious as to why you remove the element here?
Doesn't seem necessary if the iterator only exists in the scope of the for loop.
But if it is I'd like to learn why :)

KyleAure (approved):

dev/io.openliberty.concurrent.internal.cdi/src/io/openliberty/concurrent/internal/cdi/ConcurrencyExtension.java
njr-11: Removing from the iterator also removes from the Set that is iterated over, so it's no longer holding references to it in memory.  It might have been simpler to just do a .clear operation at the end.

PR #25789, opened 2023-07-21
Standardize artifactory logic between maven and docker

KyleAure:
replace `use.artifactory.substitution` with `artifactory.force.external.repo`

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
+++ b/dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
@@ -32,7 +32,7 @@ public class ArtifactoryImageNameSubstitutor extends ImageNameSubstitutor {
     /**
      * Manual override that will allow builds or users to pull from the default registry instead of Artifactory.
      */
-    private static final String substitutionOverride = "fat.test.artifactory.force.external.repo";
+    private static final String forceExternal = "fat.test.artifactory.force.external.repo";
 
     /**
      * Artifactory keeps a cache of docker images from DockerHub within
@@ -79,10 +79,20 @@ public class ArtifactoryImageNameSubstitutor extends ImageNameSubstitutor {
             }
 
             // Priority 5: System property artifactory.force.external.repo (NOTE: only honor this property if set to true)
-            if (Boolean.getBoolean(substitutionOverride)) {
+            if (Boolean.getBoolean(forceExternal)) {
                 result = original;
                 needsArtifactory = false;
                 reason = "System property [ fat.test.artifactory.force.external.repo ] was set to true, must use original image name.";
+                break;
+            }
+
+            // Priority 6: If Artifactory registry is available use it to avoid rate limits on other registries
+            if (ArtifactoryRegistry.instance().isArtifactoryAvailable()) {
+                result = DockerImageName.parse(mirror + '/' + original.asCanonicalNameString())
+                                .withRegistry(ArtifactoryRegistry.instance().getRegistry())
+                                .asCompatibleSubstituteFor(original);
+                needsArtifactory = true;
+                reason = "Artifactory was available.";
                 break;
             }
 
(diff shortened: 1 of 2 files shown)


dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
```diff
@@ -30,9 +30,9 @@ public class ArtifactoryImageNameSubstitutor extends ImageNameSubstitutor {
     private static final Class<?> c = ArtifactoryImageNameSubstitutor.class;
 
     /**
-     * Manual override that will allow builds or users to pull from Artifactory instead of DockerHub.
+     * Manual override that will allow builds or users to pull from the default registry instead of Artifactory.
      */
-    private static final String substitutionOverride = "fat.test.use.artifactory.substitution";
+    private static final String substitutionOverride = "fat.test.artifactory.force.external.repo";
```
Azquelt: I would suggest renaming this variable to `forceExternal` or `forceNoArtifactory` or `forceNoSubstitution`

dev/fattest.simplicity/src/componenttest/containers/ArtifactoryImageNameSubstitutor.java
```diff
@@ -56,33 +56,33 @@ public DockerImageName apply(final DockerImageName original) {
...
-                                .asCompatibleSubstituteFor(original);
-                needsArtifactory = true;
-                reason = "Using a remote docker host, must use Artifactory registry";
+            // Priority 5: System property artifactory.force.external.repo (NOTE: only honor this property if set to true)
+            if (Boolean.getBoolean(substitutionOverride)) {
+                result = original;
+                needsArtifactory = false;
+                reason = "System property [ fat.test.artifactory.force.external.repo ] was set to true, must use original image name.";
                 break;
             }
```
Azquelt: I think this is saying "If we're using a remote docker host, we must use Artifactory" and so we ignore the `force.external` option?

What prompts us to use a remote docker host? Should that decision also respect the `force.external` flag?

Alternatively, is it better for this to remain a separate if you want to be able to enable substituting the repository name without using a remote docker server?

I guess the other issue here is that docker hub is quite strict on their rate limits (unlike maven central) so not using an artifactory proxy is more of an problem.

KyleAure: > I think this is saying "If we're using a remote docker host, we must use Artifactory" and so we ignore the force.external option?

Yes, because above all else we don't want to get rate limited on our remote docker hosts.

> What prompts us to use a remote docker host? Should that decision also respect the force.external flag?

Another complex hierarchy of properties and environmental variables.  However, forcing the use of Artifactory when using a remote docker host should be considered before `forceExternal` due to the above reason.

However, I think I do need to add in a state for if Artifactory could be used then we should and have the `forceExternal` be before that decision.

KyleAure:
#build

Azquelt (approved):
I _think_ this is right.

The logic ends up being a little more complicated than when we're using artifactory with maven because there are some extra considerations:

- If we're using a remote docker host, then we assume that needs to use artifactory as well
- If the image specifies a particular source then we always use that source and never use artifactory

After those two conditions, the logic is now the same as [we use for maven](https://github.com/OpenLiberty/open-liberty/blob/ecdac855de663f48bca8b92c1062583aadd33af2/dev/settings.gradle#L44-L46):
- If `force.external.repo` is set, don't use artifactory
- Otherwise, if an artifactory server is set, use it
- Otherwise don't use artifactory

The change in behaviour is that we'll now use artifactory if an artifactory server is configured, whereas previously we would only do so if `fat.test.use.artifactory.substitution` was set.

Azquelt:
#libby

KyleAure:
Waiting to build

KyleAure:
#build

KyleAure:
#build

KyleAure:
Personal build had one test bucket timeout and produce javacores.  

Re-running child build to verify failure was caused by this PR: 
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_t3xdgDYYEe6tLJXjfwlpJg

KyleAure:
Re-run passed.  Assume it was just a one off issue.

PR #25801, opened 2023-07-24
Added initial delay for Netty quiesce verifier

### LINKED ISSUE #25800: Fix Race Condition
Ran into the following exception within builds

> java.lang.NullPointerException: Cannot invoke "java.util.concurrent.ScheduledFuture.isDone()" because "this.quiesceVerifierTask" is null
> > at io.openliberty.netty.internal.impl.NettyQuiesceListener.lambda$startTasks$1(NettyQuiesceListener.java:92)
> > at com.ibm.ws.threading.internal.SchedulingRunnableFixedHelper.run(SchedulingRunnableFixedHelper.java:292)
> > at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
> > at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
> > at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
> > at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
> > at java.base/java.lang.Thread.run(Thread.java:858)

We believe this is a possible race condition with the Netty Quiesce listener where the method handling Quiesce check for Netty is referencing itself from the executor. On an initial turn, it may be running separately where the method already starts running on a separate thread while the NettyQuiesceListener has not yet set the return value of the variable causing the NPE. A possible solution is to add an initial delay to make sure the variable is set before running the executor thread.

isaacrivriv:
fixes #25800

### DIFF
--- a/dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
+++ b/dev/io.openliberty.netty.internal.impl/src/io/openliberty/netty/internal/impl/NettyFrameworkImpl.java
@@ -219,6 +219,12 @@ public class NettyFrameworkImpl implements ServerQuiesceListener, NettyFramework
     		long timeout = getDefaultChainQuiesceTimeout();
 
     		if(timeout > 0) {
+                if(activeChannelMap.isEmpty() && outboundConnections.isEmpty()){
+                    if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


isaacrivriv:
#build

pnicolucci (approved):

isaacrivriv:
Error on build related to unit test `com.ibm.ws.security.utility` timing out which is not related to this change. Should be safe to merge in.

PR #25802, opened 2023-07-24
Allow a longer connection timeout

c00crane:
On some platforms the connection is taking longer than expected - I'm increasing the default connection timeout that the tests use.  We'll still test the default value, but, will generally use a longer timeout.


c00crane:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat,io.openliberty.security.jakartasec.3.0.internal_fat.config.1,io.openliberty.security.jakartasec.3.0.internal_fat.config.2,io.openliberty.security.jakartasec.3.0.internal_fat.logout,io.openliberty.security.jakartasec.3.0.internal_fat.refresh

c00crane:
Personal build was good (failure was in an unrelated ui project)

arunavemulapalli (approved):

PR #25828, opened 2023-07-27
Set content-length header to match size of updated content

abutch3r:
The Content-length header was not being updated when the document path was changed. such that only paths of 7 characters could work.

Remove super.flushBuffer() from wrapper to prevent early commit of content before the filter updates the response.

Update test to not have a doc path value that is the same length of path as `/openapi`


abutch3r:
#libby

abutch3r:
#build

Azquelt (approved):

abutch3r:
#build

abutch3r:
#build

Issue #25830, opened 2023-07-27
Liberty Error: CloneNotSupportedException caught while trying to clone MessageDigest with algorithm SHA-512


simysc:
**Describe the bug:**  
 An exception called "CloneNotSupportedException" is caught while trying to clone MessageDigest with algorithm SHA-512 on Liberty.

**Trace Log:**

[7/18/23 14:52:49:072 UTC] 0000003f id=00000000 cation.internal.cache.keyproviders.BasicAuthCacheKeyProvider 3 CloneNotSupportedException caught while trying to clone MessageDigest with algorithm SHA-512. This is pretty unlikely, and we need to get details about the JDK which is in use. 

               java.lang.CloneNotSupportedException

               at java.security.MessageDigest$Delegate.clone(MessageDigest.java:595)

               at com.ibm.ws.security.authentication.internal.cache.keyproviders.BasicAuthCacheKeyProvider.getMessageDigest(BasicAuthCacheKeyProvider.java:206)

**Steps to Reproduce**  
TBD

**Expected behavior**  
No exception should be logged as Liberty recovers by itself. There is no functional issue due to this.

**Diagnostic information:**  
 - Open Liberty Version: 22.0.0.9
 - Affected feature(s) : IBM JCEPlus provider
 - Java Version: 8.0.7.20

**Additional context**  
Related slack messages: https://ibm-cloud.slack.com/archives/C324NP6H5/p1689784419360069

simysc:
Closing issue.

PR #25832, opened 2023-07-27
Suppressed CloneNotSupportedException For Issue #25830

simysc:
This PR is to suppress the exception called "CloneNotSupportedException" which is caught while trying to clone MessageDigest with algorithm SHA-512 on Liberty.

### DIFF
--- a/dev/com.ibm.ws.security.authentication.builtin/src/com/ibm/ws/security/authentication/internal/cache/keyproviders/BasicAuthCacheKeyProvider.java
+++ b/dev/com.ibm.ws.security.authentication.builtin/src/com/ibm/ws/security/authentication/internal/cache/keyproviders/BasicAuthCacheKeyProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2022 IBM Corporation and others.
+ * Copyright (c) 2011, 2023 IBM Corporation and others.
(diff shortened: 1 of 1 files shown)


simysc:
#build

una-tapa (changes requested):
This code is checking the debug but doing nothing. 

if (tc.isDebugEnabled()) {
                /*
                 * commented out to suppress CloneNotSupportedException
                 * Tr.debug(tc, "CloneNotSupportedException caught while trying to clone MessageDigest with algorithm " + SHA_512
                 * + ". This is pretty unlikely, and we need to get details about the JDK which is in use.",
                 * cnse);
                 */
            }

Can we change it to something like below? 
No check, and add reasons in the comment. 

catch (CloneNotSupportedException cnse) {
           //Implementation of clone is optional in Java. When clone does not work, just return a new instance. 
            return MessageDigest.getInstance(SHA_512);
        }

simysc:
#build

una-tapa (dismissed):
The code change looks good. Thanks, Simy.

simysc:
#Libby

simysc:
#build

una-tapa (approved):
copyright looks good.

simysc:
#Libby

simysc:
#build

simysc:
#Libby

simysc:
#libby

PR #25837, opened 2023-07-28
Use containsKey rather than the Groovy specific hasProperty

tevans78:
Environment variables were not being set properly on a local run. Appears that the Groovy specific hasProperty method does not do what we want and always returns false. Switched to the regular Java containsKey method.

#build


Azquelt (approved):

Azquelt:
FYI @KyleAure

tevans78:
Note that this code path is only used for local FATs. The main build runs FATs remotely and so the properties are set in a different way. Therefore, there is very little value in waiting for the PB to complete. I have tested this locally so will merge without the PB.

PR #25839, opened 2023-07-28
sync up spec API with Jakarta Data 1.0 beta 3

njr-11:
Now that beta 3 for Jakarta Data 1.0 is out, we should update our copy of the API to match it.  Specifically this concerns 3 classes that are switched over to Java records.

### DIFF
--- a/dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Limit.java
+++ b/dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Limit.java
@@ -15,38 +15,27 @@ package jakarta.data.repository;
 /**
  * Method signatures are copied from the jakarta.data.repository.Limit from the Jakarta Data repo.
  */
-public class Limit {
-    private final int max;
-    private final long start;
+public record Limit(int maxResults,
+                long startAt) {
 
-    private Limit(long startAt, int maxResults) {
-        start = startAt;
-        max = maxResults;
-        if (start < 1)
-            throw new IllegalArgumentException("startAt: " + start);
-        if (max < 1)
-            throw new IllegalArgumentException("maxResults: " + max);
-    }
-
-    public int maxResults() {
-        return max;
+    public Limit {
+        if (startAt < 1)
+            throw new IllegalArgumentException("startAt: " + startAt);
+        if (maxResults < 1)
+            throw new IllegalArgumentException("maxResults: " + maxResults);
     }
 
     public static Limit of(int maxResults) {
-        return new Limit(1L, maxResults);
+        return new Limit(maxResults, 1L);
     }
 
     public static Limit range(long startAt, long endAt) {
-        if (startAt >= endAt)
+        if (startAt > endAt)
             throw new IllegalArgumentException("startAt: " + startAt + ", endAt: " + endAt);
 
         if (Integer.MAX_VALUE <= endAt - startAt)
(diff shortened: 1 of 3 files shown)


dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Pagination.java
```diff
@@ -14,104 +14,55 @@
...
+record Pagination(long page,
+                int size,
+                List<Sort> sorts,
+                Mode mode,
+                Cursor cursor)
+                implements Pageable {
+
+    Pagination {
+        if (page < 1)
+            throw new IllegalArgumentException("pageNumber: " + page);
```
gjwatts, line 35: Since the `pageNumber` parameter was renamed to `page`, I would suggest changing this line to:
```
throw new IllegalArgumentException("page: " + page);
```
to be consistent.

dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Pagination.java
```diff
@@ -14,104 +14,55 @@
...
+        if (size < 1)
+            throw new IllegalArgumentException("maxPageSize: " + size);
```
gjwatts, line 37: Same here:
```
throw new IllegalArgumentException("size: " + size);
```

dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Limit.java
```diff
@@ -15,38 +15,27 @@
...
-        max = maxResults;
-        if (start < 1)
-            throw new IllegalArgumentException("startAt: " + start);
-        if (max < 1)
-            throw new IllegalArgumentException("maxResults: " + max);
-    }
-
-    public int maxResults() {
-        return max;
+public record Limit(int maxResults,
```
gjwatts, line 18: Just pointing out that the order of the parameters was switched, but I assume that was intentional.

gjwatts (approved):

dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Limit.java
njr-11: Yes, the order needs to match what is in the Jakarta Data beta 3 driver.

dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Pagination.java
njr-11: The message needs to match the Jakarta Data beta 3 driver. Possibly it should change there, but we need to remain consistent with it until it does.

dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/repository/Pagination.java
njr-11: The message needs to match the Jakarta Data beta 3 driver. Possibly it should change there, but we need to remain consistent with it until it does.

njr-11:
#build

njr-11:
#build

njr-11:
#build

njr-11:
Build was successful with all tests passing, although it did not generate a comment for this github issue.

PR #25840, opened 2023-07-28
Incorrect default for session cache id property

zhang-ca:
If the user's browser session is moving back and forth across multiple web applications, you might see extra persistent store activity as the in-memory sessions for a web module are refreshed from the persistent store.  The default behavior for cache identifiers increasing is enabled. That is, the property to prevent cache identifiers increasing is disabled.
optimizeCacheIdIncrements = false


zhang-ca:
#build

zhang-ca:
#build

bensonlatibm (approved):
lgtm!

Issue #25216, opened 2023-05-05
InstantOn Problems restoring when using a DB for the transaction log.


jdmcclur:
I am testing InstantOn with rest-crud and using a DB for the transaction log, based on info here: https://openliberty.io/docs/latest/transaction-service.html.

I added this to the server.xml:
```
    <transaction transactionLogDBTableSuffix="${HOSTNAME}${wlp.server.name}"
        recoverOnStartup="false"
        recoveryGroup="peer-group-name"
        recoveryIdentity="${HOSTNAME}${wlp.server.name}">
        <dataSource transactional="false">
        <jdbcDriver libraryRef="postgres"/>
          <properties databaseName="rest-crud" user="restcrud" password="restcrud" serverName="checkers06.rtp.raleigh.ibm.com" portNumber="5432"/>
        </dataSource>
    </transaction>
```

Things work fine in a normal case (not using instantOn).
```
[AUDIT   ] WTRN0108I: Server with identity 16808a2c779fdefaultServer is monitoring its peers for Transaction Peer Recovery
[AUDIT   ] WTRN0108I: Insert New Lease for server with recovery identity 16808a2c779fdefaultServer
```

But using instantOn, after a restore on the first request I see.
```
[5/5/23, 14:32:43:731 EDT] 000000d2 com.ibm.ws.logging.internal.impl.IncidentImpl                I FFDC1015I: An FFDC Incident has been created: "java.lang.IllegalStateException: Required attribute is null,AtomicServiceReference[name=jpaRuntime,ctx=null,ref=null,svc=null] com.ibm.wsspi.injectionengine.InjectionBinding.getInjectionObject 408" at ffdc_23.05.05_14.32.43.0.log
[5/5/23, 14:32:43:824 EDT] 000000d2 com.ibm.ws.logging.internal.impl.IncidentImpl                I FFDC1015I: An FFDC Incident has been created: "com.ibm.wsspi.injectionengine.InjectionException: CWNEN0030E: The server was unable to obtain an object instance for the java:comp/env/com.acme.crud.FruitResource/entityManager reference.  The exception message was: Required attribute is null,AtomicServiceReference[name=jpaRuntime,ctx=null,ref=null,svc=null] com.ibm.ws.cdi.impl.weld.injection.WebSphereInjectionServicesImpl 201" at ffdc_23.05.05_14.32.43.1.log
[...]

PR #25841, opened 2023-07-30
Replace TM service reference with TX Mgr factory in JPAComponentImpl

dazavala:
For issue #25216 

The `JPAComponent` service component does not tolerate dynamic updates to the `TransactionManagerService`. Doing so causes `JPAComponentImpl` to reactivate, which is problematic whenever applications do not also restart. Alleviate the problem by modifying `JPAComponentImpl` to use the `EmbeddableTransactionManagerFactory (ETMF)`, instead of the `TransactionManagerService`, to access the `EmbeddableWebSphereTransactionManager` singleton and the `UOWCurrent` implementations. Notice the `TransactionManagerService` impl wraps the `ETMF` and exports the same behaviors.

This solution is preferable to modifying the `TMS` reference from `static+mandatory` to `dynamic+optional`, because the `TMS` is required for the Persistence/JPA and Persistence/JPA Container feature runtimes. However, doing so would solve the problem as declarative services would simply inject the updated service reference into the existing JPAComponentImpl rather than reactivate it (i.e. deactivate > unbind svc refs > rebind svc refs > reactivate.) Also, note that a `dynamic+mandatory` reference does not solve the problem.

The ideal solution would modify the `JPAComponentImpl` to tolerate service updates w/o requiring applications to restart, which IMHO, would require managing application state within some other class. Also, the solution must ensure the service component activates before it is used. This behavior can be enabled by 1) configuring the `JCAComponent` service component to `immediate`; 2) by tracking the `TMS` and consuming the `JPAComponent` service (i.e. getService(JPAC)) when the updated `TMS` service registers with OSGi; or 3) by changing the `JPAComponent` service component's dependency graph to ensure some service always consumes the `JPAComponent` service after deactivation, before it used to process work.

### DIFF
--- a/dev/com.ibm.ws.jpa.container/src/com/ibm/ws/jpa/container/osgi/internal/JPAComponentImpl.java
+++ b/dev/com.ibm.ws.jpa.container/src/com/ibm/ws/jpa/container/osgi/internal/JPAComponentImpl.java
@@ -47,6 +47,7 @@ import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
 
+import com.ibm.tx.jta.embeddable.EmbeddableTransactionManagerFactory;
 import com.ibm.websphere.csi.J2EEName;
 import com.ibm.websphere.ras.ProtectedString;
 import com.ibm.websphere.ras.Tr;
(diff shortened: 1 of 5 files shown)


dazavala:
#build

#libby

dazavala:
#build

dazavala:
@tkburroughs  Review concerns: These changes introduce import `com.ibm.tx.jta.embeddable.EmbeddableTransactionManagerFactory` into JPAComponentImpl, which requires the server to resolve the new import package in bundle `com.ibm.ws.jpa.container.jar`. Package `com.ibm.tx.jta.embeddable` is exported by `com.ibm.ws.tx.embeddable.jar`. Bundles that export `...tx.jta...` packages, including `com.ibm.ws.tx.embeddable.jar`, are provisioned by the transaction-x.y features, and these bundle manifests declare the necessary package exports. Since Persistence (JPA) features include transaction-x.y, this solution does not introduce potential packing wiring failures when the server resolves installed feature bundles.

dazavala:
#build

tkburroughs (approved):

dazavala:
#libby

Liberty personal builds passed. The InstantOn personal build request remains pending, so I manually verified this branch in my workspace by running the FULL checkpoint_FAT. Look good. Ready to merge pending libby's checks AND devOps authorization.

dazavala:
FYI, the InstantOn personal build completed with no breaks related to these changes.
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_nPoRADDXEe6cYpJ4GUjoAA&tab=com.ibm.team.build.web.ui.internal.editors.result.overview.OverviewPage

This PR is ready to merge pending devOps approval.

PR #25847, opened 2023-07-31
Add support for PBKDF2withHmacSHA256 to wlp.encoding.key

NottyCode:
Currently Liberty uses PBKDF2withHmacSHA1 which isn't a good practice anymore since SHA1 is considered weak. Need to implement support for PBKDF2withHmacSHA256, but without breaking the ability to read previously encrypted AES encrypted passwords.


NottyCode:
#build

NottyCode:
#build

PR #25844, opened 2023-07-31
Springboot websocket fat test

rsherget:
Fixes the websocket SpringBoot FAT test

PR #25858, opened 2023-08-01
Fix for Trace SB3 test and other start failure issues

hlhoots:
#build 

This Fixes the SB test error for the Trace test case as mentioned in 25764.  This also fixes the beta = true for the test bucket that was causing the server to not start properly due to thinning.


cbridgha (approved):
Looks good

hlhoots:
Last build had a timeout that we've not encountered and cannot produce locally.  Rerunning the build to see if there really is a problem.

#build

hlhoots:
Closing this out.  Changes within are added to https://github.com/OpenLiberty/open-liberty/pull/25844

PR #25875, opened 2023-08-02
Remove unnecessary metatype from app manager lifecycle

joe-chacko:
#build


joe-chacko:
Does #build work in the initial comment?
Or #run-libby-bot either?

Azquelt (approved):
This looks good to me on the basis that the metatype elements were marked internal use only.

PR #25901, opened 2023-08-07
validation of entities on update and return types for saveAll

njr-11:
Adds test cases for updates to JPA entities with validation constraints, for cases where the constraints are violated and not violated.  While working on this, I noticed that the save/saveAll operation doesn't all the full range of return types that are required by the Jakarta Data spec, so this pull also corrects that an adds testing.


jmstephensgit (approved):

njr-11:
#build

PR #25908, opened 2023-08-08
Remove service.vendor from OSGi Components

tevans78:
We do not think that the `service.vendor=IBM` is appropriate or necessary for this Open Liberty project.


tevans78:
#build

PR #25920, opened 2023-08-09
Update OpenShiftUserApiUtils to get groups from API response

ayoho:
The `OpenShiftUserApiUtils` class only looks for groups in a subset of the response from the OpenShift User API. The response back from the API should look something like this:

```JSON
{
    "kind": "User",
    "apiVersion": "user.openshift.io/v1",
    "metadata": {
        "name": "<namespace>:<user-name>",
        "creationTimestamp": null
    },
    "groups": [
        "system:authenticated",
        "group1",
        "group2"
    ]
}
```

Before this change, the populated groups would only contain the namespace extracted from the username in the `"metadata"` entry. These changes update the code to also add all of the entries from `"groups"` (if any exist) into the result.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.social_fat.okdServiceLogin

jimmy1wu (approved):

PR #25925, opened 2023-08-10
injection of default managed executors and context service

njr-11:
Automatically allow injection of the default ManagedExecutorService, ManagedScheduledExecutorService, and ContextService.
Also includes some code to try to identify matching pre-existing ManagedExecutorService/ManagedScheduledExecutorService/ContextService beans so that we only add if one isn't already there.


njr-11:
#build
#libby

mswatosh (approved):

PR #25942, opened 2023-08-14
JAX-WS: Refactor LibertyCXFNegativePropertiesTest

neuwerk:
This PR refactors the `LIbertyCXFNegativePropertiesTest` suite in order to address repeated timeout failures.

The test methods have been moved to the `LIbertyCXFNegativePropertiesTestServlet.java` class

-  LIbertyCXFNegativePropertiesTestServlet now contains the WebServiceRef clients to prevent connection issues
-  The test application has been moved into a separate test-application directory libertyCXFProperites to clean up overlaps with WebServiceRefFeatureTest
-  Cleaned up code overlap and reused code from other applications
-   Addressed the cause of the timeouts by fixing the server.xml configuration


neuwerk:
#build

sawadood (approved):
I think we need to review all the assertion messages and make sure they convey some error condition,  some of the current assertion messages sound more like informational messages.
For example,
assertNull("Since cxf.ignore.unsupported.policy is not enabled, checkEffectivePolicy will be called",
server.waitForStringInTraceUsingMark("WARNING: checkEffectivePolicy will not be called"));
I would replace this with a msg something like "Error: checkEffectivePolicy is not called even though   cxf.ignore.unsupported.policy property is not set.

PR #25951, opened 2023-08-15
Expect an additional error message for WasReqURLTests

c00crane:
Several WasReqURLTests expected error message CWWKS1532E.  They will now also expect error message CWWKS1554E.  The 2 messages are very similar, but, each does contain unique information.
I've updated the tests to expect/require both messages be issued now.


c00crane:
#build

ayoho (approved):

PR #25960, opened 2023-08-16
Fix Origin header behavior in OIDC client

ayoho:
The OIDC client token request code has a quirk where the Origin HTTP header will be added to _all_ token endpoint requests once it has been sent initially. That was the effect of me quickly adding the Origin header to a static `commonHeaders` variable in the `OidcClientHttpUtil` class to get the functionality initially working.

I updated an existing `setupPost()` method to now be public, allowing me to create an `HttpPost` object local to the `TokenRequestor.postToTokenEndpoint()` method. That now lets me add the Origin header to only that particular request (if a header value is configured) instead of all subsequent requests.


ayoho:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.oidc.client_fat.3,com.ibm.ws.security.social_fat.LibertyOP.3

c00crane (approved):

PR #25976, opened 2023-08-17
Improve portability of checkpoint for power and z/OS systems

tjwatson:
Additional GLIBC_TUNABLES settings are needed.

 - For z/OS on s390x arch "z14" is needed and DFLTCC=0
 - For power "arch_2_07" is needed

This change also adds -XX:+IgnoreUnrecognizedRestoreOptions JVM option so we can avoid failing to restore if an unsupported JVM restore option is specified.


tjwatson:
#build

dazavala (approved):
LGTM.  Changes exactly satisfy their description.

PR #25980, opened 2023-08-17
Update server script to fix sed command

### LINKED ISSUE #25958: sed command in server script returning incorrect value on Solaris
Hi,  We download and include the Liberty wlp-base-embeddable-x.x.x.x.zip into IBM MQ. We upgrade to 23.0.0.6 and that with this version it break on Solaris as it fails to correctly process the "./etc/server.env", and so uses a default value for "WLP_USER_DIR".
The server.env has the following entry:
WLP_USER_DIR=/var/mqm/web/installations/Installation1
The liberty binaries are located "/opt/mqm/web".
Looking further I can there was a change to "/bin/server" script being:
previously:
	if var=`safeEcho "$line" | sed -e 's/^\([a-zA-Z0-9_][a-zA-Z0-9_]*\)=.*/\1/'`; then
changed to:
	if var=`safeEcho "$line" | sed -e 's/^\([_[:alnum:]][_[:alnum:]]*\)=.*/\1/'`; then
This syntax works find on Linux & AIX, but does not on solaris.
I test AIX & Linux with the following:
echo "WLP_USER_DIR=/var/mqm/web/installations/Installation1" | sed -e 's/^\([_[:alnum:]][_[:alnum:]]*\)=.*/\1/'
Linux & AIX returns:
	WLP_USER_DIR
Solaris returns:
	WLP_USER_DIR=/var/mqm/web/installations/Installation1
Using the old sed command give the same output as Linux & AIX when run on solaris. So we only have the issue with the solaris(5.11 11.4.10.3.0) ?
It looks like solaris sed does not understand the ":alnum:" ?

jimblye:
fixes #25958 

Modify sed command to run on Solaris.   Other operating systems continue to use previous sed command.
The sed command extracts the variable from an assignment statement in server.env.  

I did make one other minor change to the sed command which applies to all operating systems on which the script runs.  Previously, the command allowed the variable name to start with a number.  I cannot find any operating system on which a variable name starting with a number is valid.  So I disallowed that.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -829,12 +829,25 @@ readServerEnv()
               fi
            fi ;;
       *=*)
-        # Only accept alphanumeric variable names to avoid eval complexities.
-        if var=`safeEcho "$line" | sed -e 's/^\([_[:alnum:]][_[:alnum:]]*\)=.*/\1/'`; then
-           case $var in
-           *=*)
-               SERVER_ENV_SETUP_FAILURE="${var} ${SERVER_ENV_SETUP_FAILURE}" ;;
-           *)
+          # The sed commands extracts the variable from the line but only if it contains alpha-numeric
+          # characters. Only accept alphanumeric variable names to avoid eval complexities.
+          if [ "`uname -s`" != "SunOS" ]; then
+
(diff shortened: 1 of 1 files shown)


jimblye:
#build

hlhoots (approved):
LGTM

jimblye:
#libby

jimblye:
#libby

PR #25984, opened 2023-08-17
Add SpringBoot3 application tag with War test

rsherget:
Adds a SpringBoot 3 test for starting a SpringBoot War app with the springbootApplication tag in the server.xml


cbridgha (approved):
Looks good

PR #25988, opened 2023-08-18
Update Fats that use the kafka-clients jar to latest

### LINKED ISSUE #25645: Update Reactive Messaging Tests to use latest Kafka Client version
Update Reactive Messaging Tests to use latest Kafka Client version. This should be done for mpReactiveMessaging-1.0 and 3.0 to see if anything is broken

abutch3r:
Upgrade from Kafka-clients jar from 2.3.0 to 3.5.1

MP Health 2.0 fat also has a kafka test that uses the client so update that as well

2.3.0 kept in oss_source_depenencies.maven given that it is still used by the kafka.adapter.impl bundle, while 3.5.1 currently is used solely for testing purposes as such added to the oss_dependencies.maven file.

Update kafka-clients dependencies to align with versions used by 3.5.1

Fixes #25645


abutch3r:
#build

abutch3r:
#build

abutch3r:
#libby #build

Azquelt (approved):

abutch3r:
Build failure was in unrelated openidconnect test cases which would be unaffected by anything kafka that the kafka clients would of caused as the failures look more low level relating to signing keys.

PR #25991, opened 2023-08-18
Harden OpenAPI HTTPS Tests

abutch3r:
Especially on Windows were certificate creation can be exceedingly slow. >30 seconds has been regularly observed in failed OpenAPI Selenium tests. We should wait for the server to state that the HTTPS endpoint is ready to accept traffic, before connecting the selenium driver to server.

As it does take more 30 seconds on occasion, allow up to 2 minute for the log message to appear. as server will start without the HTTPS Endpoint being ready.

Allow time for JS to execute having logged in for the OAuth tests for the change in button text

Replace hard-coding of the redirect port for the OAuth tests with server defined port.

For RTC295995
For RTC296869
For RTC297048

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
+++ b/dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
@@ -1111,6 +1111,21 @@ public class LibertyServer implements LogMonitorClient {
     }
 
     /**
+     * Wait for the server to state that it is listening on its SSL port
+     *
+     * @throws Exception
+     */
+    public void waitForSSLStart() throws Exception {
+        //wait for "CWWKO0219I: TCP Channel defaultHttpEndpoint-ssl has been started and is now listening for requests on host"
+        String sslStartMsg = waitForStringInLogUsingMark("CWWKO0219I:.*defaultHttpEndpoint-ssl.*");
+        if (sslStartMsg == null) {
+            RuntimeException rx = new RuntimeException("Timed out waiting for the server to initialize defaultHttpEndpoint-ssl");
+            Log.error(c, "waitForSSLStart", rx);
+            throw rx;
+        }
+    }
+
+    /**
      * Start the server and validate that the server was started:
      * prepares/cleans the server directory, then performs a clean start
      *
--- a/dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
+++ b/dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
@@ -101,6 +101,10 @@ public class UIBasicTest {
 
     @Test
     public void testHttpsUI(){
+        //Reduce possibility that Server is not listening on its HTTPS Port
+        //Especially for Windows if certificates are slow to create
+        server.waitForStringInLog("CWWKO0219I: TCP Channel defaultHttpEndpoint-ssl", 60000);
+
         driver.get("https://host.testcontainers.internal:" + server.getHttpDefaultSecurePort() + "/openapi/ui");
         testUI();
     }
--- a/dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
+++ b/dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
@@ -100,10 +100,10 @@ public class UIBasicTest {
     }
 
     @Test
-    public void testHttpsUI(){
+    public void testHttpsUI() throws Exception{
         //Reduce possibility that Server is not listening on its HTTPS Port
         //Especially for Windows if certificates are slow to create
-        server.waitForStringInLog("CWWKO0219I: TCP Channel defaultHttpEndpoint-ssl", 60000);
+        server.waitForSSLStart();
 
         driver.get("https://host.testcontainers.internal:" + server.getHttpDefaultSecurePort() + "/openapi/ui");
         testUI();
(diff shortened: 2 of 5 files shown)


abutch3r:
#build

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIBasicTest.java
```diff
@@ -101,6 +101,10 @@ public void testHttpUI(){
 
     @Test
     public void testHttpsUI(){
+        //Reduce possibility that Server is not listening on its HTTPS Port
+        //Especially for Windows if certificates are slow to create
+        server.waitForStringInLog("CWWKO0219I: TCP Channel defaultHttpEndpoint-ssl", 60000);
```
Azquelt: We don't need to set a manual timeout here. The standard timeout is two minutes to account for slow build machines and I don't think we need to reduce it for these tests.

In general I would use `CWWKO0219I:.*defaultHttpEndpoint-ssl` as the string to search for to avoid relying on the server running with the English locale, and have a comment with an example of the log line we're looking for.

Azquelt:
From the logs, it looks like the updates made to the server.xml for `customPathOAuthTest` are still present when `defaultPathOAuthTest` runs.

It looks like in most builds `defaultPathOAuthTest` runs first, so we don't see this problem. In general the execution order of tests within a class is not fixed, so we need to make it work either way around.

The easiest solution here might be to store the server config in the `@Before` method and restore it in an `@After` method so that we know any changes made by tests can't leak.

abutch3r:
#build

dev/fattest.simplicity/src/componenttest/topology/impl/LibertyServer.java
```diff
@@ -1110,6 +1110,19 @@ public void waitForSSLRestart() throws Exception {
...
+     * Wait for the server to state that it is listening on its SSL port
+     *
+     * @throws Exception
+     */
+    public void waitForSSLStart() throws Exception {
+        //wait for "CWWKO0219I: TCP Channel defaultHttpEndpoint-ssl has been started and is now listening for requests on host"
+        String sslStartMsg = waitForStringInLogUsingMark("CWWKO0219I:.*defaultHttpEndpoint-ssl.*");
+        if (sslStartMsg == null) {
+            Log.warning(c, "SSL may not have started properly - future failures may be due to this");
+        }
```
Azquelt: It would be much more helpful to throw an exception here so that we don't even try to run the test if the SSL port didn't open. That way we get the actual problem that occurred in the test failure exception, rather than a later consequence of the problem.

That said, I do note that the sibling `waitForSSLRestart` method merely logs.

I don't want to make a method that just logs (because it's not useful) but I also don't want to have a similar looking method that does something different (because it's confusing). I would move this method back into our project where we can do what's sensible for our tests and make it throw an exception.

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIOauthTest.java
```diff
@@ -218,11 +228,14 @@ public void OAuthTest(String uiPath) {
         allowOnceBtn.click();
 
         // should now have a single tab which is the original openapi page
-        new WebDriverWait(driver, Duration.ofSeconds(3)).until(driver -> driver.getWindowHandles().size() == 1);
+        new WebDriverWait(driver, Duration.ofSeconds(3)).until(d -> driver.getWindowHandles().size() == 1);
```
Azquelt, line 235: I don't _think_ this changes anything. Was this change (and the one on line 210) just testing to see if it made a difference?

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIOauthTest.java
```diff
@@ -218,11 +228,14 @@ public void OAuthTest(String uiPath) {
...
         // Switch back to original window
         driver.switchTo().window(originalWindow);
 
-        //check button text has changed to Logout
+        //The javascript may not have completed the update on the auth button from `Authorize` to `Logout`, so we wait for an element that should be a logout button
+        waitForElement(driver, By.xpath("//button[text()='Logout']"));
```
Azquelt: Should we be searching beneath `authContainer` rather than `driver` here?

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIOauthTest.java
```diff
@@ -118,6 +125,9 @@ public void teardownTest() throws Exception {
         //close the browser before stopping the server to reduce changes of open connections
         driver.quit();
         server.stopServer();
+
+        // Restore Config to original after test to prevent potential config bleeding between tests
+        server.updateServerConfiguration(baseConfig);
```
Azquelt: This should probably go in a try-finally to ensure we still reset the config if there were errors in the server logs (which will cause `stopServer()` to throw an exception).

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIOauthTest.java
abutch3r: This is now in line with the Selenium documentation for waiting, this was to see if this would improve some of the other behaviors we have seen on the build systems. though we typically saw the second one have issues, not the first.

abutch3r:
#libby #build

dev/io.openliberty.microprofile.openapi.ui.internal_fat/fat/src/io/openliberty/microprofile/openapi/ui/internal/fat/tests/UIOauthTest.java
Azquelt: Huh, fair enough

Azquelt (dismissed):

abutch3r:
#build #libby

Azquelt (approved):

abutch3r:
previous build looked very sick #build

abutch3r:
Latest build ran on 3 Windows(11,19 and 22) + Java combinations(11, 17 and 20) with all openapi ui tests passing.

PR #25994, opened 2023-08-18
generic array return type

njr-11:
When inheriting repository methods from an interface and the methods have a generic array return type, the result types aren't being processed correctly and methods with multiple results fail with an error that says only 1 result is allowed. This failed the TCK when I switched a few methods over to inheriting from a common interface, and I realized we don't have a test of our own for it, so this pull adds a test as well as fixing the problem.


njr-11:
#build

jmstephensgit (approved):

Issue #25986, opened 2023-08-18
TelemetryClientFilter will throw an exception for an app which is not CDI enabled


benjamin-confino:
TelemetryClientFilter uses CDI to acquire a `OpenTelemetryInfo`. This works fine if its invoked from an app which is CDI enabled but will throw an exception if the app is not CDI enabled. 

We need to fix this and to create a test for it.

benjamin-confino:
It threw an exception, we will also need a fix.

benjamin-confino:
Fixed by #26177

PR #26003, opened 2023-08-21
Create test that uses TelemetryClientFilter without CDI

benjamin-confino:
This creates a test for #25986 but is disabled until the fix is delivered.

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/clientnocdi/ClientInvokedServlet.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/clientnocdi/ClientInvokedServlet.java
@@ -1,0 +1,33 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal_fat.apps.clientnocdi;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+
+import jakarta.servlet.annotation.WebServlet;
+import jakarta.servlet.http.HttpServlet;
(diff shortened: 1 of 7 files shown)


dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/clientnocdi/ClientInvokedServlet.java
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
```
tevans78: You don't need to include the Contributors section any more

benjamin-confino:
#build

benjamin-confino:
#libby

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/clientnocdi/ClientTriggeringServlet.java
```diff
@@ -0,0 +1,85 @@
...
+
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Scope;
+
+/**
+ * Tests MP Telemetry class TelemetryClientFilter on an app that is not CDI enabled
+ */
+@WebServlet("/ClientTriggeringServlet")
+public class ClientTriggeringServlet extends HttpServlet {
```
Azquelt, line 42: This class is listed in the test class as a test servlet, but it isn't written as a test servlet.

I think this class should extend `FATTestServlet` and have a `@Test` annotated method rather than a `doGet` method.

benjamin-confino: Good catch, I'll remove the test servlet list. 

This class used to be a FatTestServlet but I switched it when I modified the test to check trace.log for proof TelemetryClientFilter was used.

benjamin-confino:
#build

PR #26010, opened 2023-08-21
startWinService stopWinService default timeouts too short

### LINKED ISSUE #23732: startWinService & stopWinService default timeouts in server.bat script too short
**Windows only**  - the problem is not applicable to other platforms

**Describe the bug**
The default timeouts in the server.bat script for starting and stopping liberty running **as a windows service** are both 5 seconds.  This is way below the comparable timeouts for starting and stopping the server in the java code where the default is 30 seconds.   In fact, for server stop, the quiesce timeout is 30 seconds, meaning the server might not even begin to shut down the runtime for up to 30 seconds.

The result of the default being too short is that the server start and server stop commands will frequently finish before the server is actually started or stopped.  

In the case of the stopWinService command, the result can be an unnecessary second attempt to stop the service which results in a confusing message:
  
```
C:\temp\wlp\bin>server startWinService

C:\temp\wlp\bin>server stopWinService
Stopping server defaultServer.
Server defaultServer is not running.
```

The two messages, "Stopping ..." and "... not running" are displayed because the stopWinService timed out and then another stop command was automatically issued, but in this example, the first stop actually worked!   If the timeout were longer, those messages would not have been displayed.

**Steps to Reproduce**  
Run these two commands back-to-back on Windows with no delay between the commands.   Assumes you have already registered the defaultServer as a Windows service.
```
server startWinService
server stopWinService
```
You probably will not be able to reproduce this problem in this manner in any release prior to 2023, and that is because starting and stopping a Windows service like this prior to 23.0.0.1 results in a hang condition, which is fixed by #23392 .

However, the problem could be reproduced by anything that causes the server shutdown to take more than default 5 seconds.

**Expected behavior**  
There should be no messages.

**Diagnostic information:**  
 - OpenLiberty Version: [e.g. 21.0.0.9 and up ] _(There was no timeout at all below 21.0.0.9 as there was no checking as to whether the server actually stopped or started in the case of a Windows service, and there was no retry in the case of a failure to stop)_
 - Affected feature(s)  starting and stopping liberty as a Windows service
 - Java Version: [N/A]
 - server.xml configuration (N/A)

**Additional context**  
Customer can adjust these timeouts on their own by setting the environment variables:
```
set WINDOWS_SERVICE_START_TIMEOUT=30
set WINDOWS_SERVICE_STOP_TIMEOUT=30
```

jimblye:
Fixes #23732

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server.bat
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server.bat
@@ -57,11 +57,11 @@
 @REM
 @REM WINDOWS_SERVICE_START_TIMEOUT - Use when liberty is registered as a Windows
 @REM              service.  The value specifies the number of seconds the script 
-@REM              waits for the service to start before continuing. Default is 5.
+@REM              waits for the service to start before continuing. Default is 30.
 @REM
 @REM WINDOWS_SERVICE_STOP_TIMEOUT - Use when liberty is registered as a Windows
 @REM              service.  The value specifies the number of seconds the script 
-@REM              waits for the service to stop before continuing.  Default is 5.
+@REM              waits for the service to stop before continuing.  Default is 30.
 @REM
 @REM SERVER_WORKING_DIR - The directory containing output files from the JVM.  For 
 @REM              example the javadump files.  The default value is the location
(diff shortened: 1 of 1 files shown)


cbridgha (approved):
Looks good

jimblye:
#libby

PR #26007, opened 2023-08-21
Update security client fat timeout issue

adriankalafut:
Fix RTC 296796
Update the repeat rule for EE levels for com.ibm.ws.security.client_fat.

PR #26016, opened 2023-08-21
improve security.client_FAT robustness on windows

Zech-Hein:
This PR aims to resolve build break 295409.

This FAT is one of many that are prone to having timeout failures when running on Windows VMs where processes tend to run slower.


Zech-Hein:
closing as this is a duplicate of #26007

PR #26018, opened 2023-08-22
For quality of life, output the EE level in more places

benjamin-confino:
This will output what EE level is being tested in common build break error messages so you can immediately go to the logs for the correct server.

### DIFF
--- a/dev/fattest.simplicity/src/componenttest/exception/TopologyException.java
+++ b/dev/fattest.simplicity/src/componenttest/exception/TopologyException.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011 IBM Corporation and others.
+ * Copyright (c) 2011, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


dev/fattest.simplicity/src/componenttest/exception/TopologyException.java
```diff
@@ -12,25 +12,36 @@
...
 public class TopologyException extends Exception {
 
     public TopologyException() {
         super();
     }
 
     public TopologyException(String message, Throwable cause) {
-        super(message, cause);
+        super(getRepeatAction() + message, cause);
     }
```
Azquelt, line 23: This class will need a copyright date update.

Azquelt (commented):
These changes all look sensible :+1:

The `TopologyException` is a bit odd in that we throw it on most problems in the FAT framework and in many cases it would make sense for it to have more context in the message which would need to be provided by the thrower, but as a quick fix, I think adding the repeat actions here will still be valuable.

Azquelt (approved):

benjamin-confino:
#build

benjamin-confino:
#libby

PR #26025, opened 2023-08-22
Flush and close the outputstream

anjumfatima90:


dev/io.openliberty.springboot.fat30_fat/fat/src/com/ibm/ws/springboot/support/fat/CommonWebFluxTests.java
```diff
@@ -84,12 +89,17 @@ public void testBlockingIO() throws IOException, InterruptedException {
...
+            InputStream in = con.getInputStream();
+            byte[] buf = new byte[size];
+            int cnt;
+            while ((cnt = in.read(buf)) > 0) {
+                response.write(buf, 0, cnt);
+                Thread.sleep(sleep);
+            }
+        } finally {
+            response.flush();
+            response.close();
```
tjwatson, line 102: This is a local `ByteArrayOutputStream`.  I don't think we need a flush/close here.  But it doesn't hurt I guess.

dev/io.openliberty.springboot.fat30_fat/fat/src/com/ibm/ws/springboot/support/fat/CommonWebFluxTests.java
```diff
@@ -72,9 +72,14 @@ public void testBlockingIO() throws IOException, InterruptedException {
...
-            Thread.sleep(sleep);
-            out.write(data);
+        try {
+            for (int i = 0; i < numIterations; i++) {
+                Thread.sleep(sleep);
+                out.write(data);
+            }
+        } finally {
+            out.flush();
+            out.close();
```
tjwatson, line 82: It is unclear to me that we should close the OutputStream on the `HttpURLConnection` while it is still alive.  But I think we can give it a try to see if it solves the intermittent issue.  Alternatively, I would have suggested only to do a `flush` call here.

tjwatson (approved):

anjumfatima90:
#build

anjumfatima90:
#build

PR #26040, opened 2023-08-23
experiment with validation conditional on valid annotation

njr-11:
Update code to try out a proposal to opt-in to bean validation with the Valid annotation.  Given that this isn't in the spec, this code path is not currently used, but can be switched on by removing the immediate return from the isValidatable method near the TODO comment.


njr-11:
#build

wraschke:
Test comment only. Hello.

jmstephensgit (approved):

PR #26048, opened 2023-08-24
Remove unnecessary call to uname for performance

### LINKED ISSUE #26046: InstantOn Startup Regression
It looks like PR #25980 has caused a ~10% regression (20ms) in startup in some instantOn scenarios that have many values in  the server.env.

This is because the code is dong a `uname -s` for every entry in the server.env, 
```
if [ "`uname -s`" != "SunOS" ]; then
```

This adds up and slows down startup. I think we already know the uname from above, so wondering if we can change the check to the following (or just check once before the for loop so it doesn't do it every time)?
```
 if [ "${uname}" != "SunOS" ]; then
 ```

jimblye:
Fixes #26046


hlhoots (approved):
LGTM

jimblye:
#build

jimblye:
#libby

jdmcclur:
LGTM

PR #26064, opened 2023-08-25
XML-WS 4.0: Add Client Container Test for StAX Provider

neuwerk:
This PR does the following:

- Add FFDCIgnores for expected ClassNofFoundExceptions to StAXUtils
- Uses existing tests, to add a new simple tests that makes sure that on EE10, the StAX Impl is correctly picking up Woodstox on EE10 and xmlWS-4.0.

Note: The test framework currently lacks the ability to check the Client trace.log file after a client has run, and the logs have been copied to the output directory. If this were corrected, this test could benefit from being updated to check the trace.log output for the correct debugging statement from StAXUtils.


neuwerk:
#build

BerksanAtes (approved):
Looks good to me!

PR #26073, opened 2023-08-28
Skip more short timeout tests on Windows

c00crane:
Tests that use a negative timeout are expecting access to protected apps after messages are logged in the server logs.  The tests will verify that error messages are logged that state that the timeout value was negative, that's not allowed, and the default timeout will be used.  On slow systems (our Windows test systems), the default timeout is NOT sufficient and we get the message about the negative value, but them we timeout.  So, we'll skip these tests when running on Windows.


c00crane:
#build
#spawn.fullfat.buckets=io.openliberty.security.jakartasec.3.0.internal_fat.config.1,io.openliberty.security.jakartasec.3.0.internal_fat.config.2

ayoho (approved):

PR #26086, opened 2023-08-29
Report app started after all hooks run

tjwatson:
A more accurate time to print that the application has started on restore is after all the restore hooks have run.  Here we listen for the ServerStarted service to print the app started messages.


tjwatson:
#build

sebratton (approved):

PR #26099, opened 2023-08-29
Add fromEnv for Java 21 and set auto-download to false

fritze2:
- Adds an environment variable from which a Java 21 JDK should be located for compiling.


fritze2:
#build

fritze2:
#build

jhanders34 (approved):

PR #26109, opened 2023-08-30
Fix topic cleanup in Kafka Reactive tests

Azquelt:
Use a single cleanup method because the ordering is not predictable if there's more than one.

Deleting the topic before shutting down the server results in a warning if the server tries to commit the partition offset after the test completes (common because we try to batch up commits).

For RTC297036


abutch3r (approved):
LGTM

Azquelt:
#libby
#build

Azquelt:
Build ran out of disk space
#build

Azquelt:
Last build broke #build

PR #26111, opened 2023-08-30
Fix log message checking for KafkaBadConfigTest

Azquelt:
The number of log messages varies depending on what order the tests run as well as the feature version in use because the first time it happens there will be an FFDC but on later runs there will not be.

Just have the test check that at least one log message is output.

For RTC297037


abutch3r (approved):
Certainly cleaner

Azquelt:
#build
#libby

PR #26119, opened 2023-08-30
Sed command in server script does not work on Z

### LINKED ISSUE #26103: Sed command in server script does not appear to work properly on Z
The sed command in this statement the following statement in server script returns an empty string on Z.   The $line is referring to a line in server.env which should be an assignment statement, e.g. x=y.    The sed command should return the variable; in this example x.    

`var=$(safeEcho "$line" | sed -n -e 's/^\([_[:alpha:]][_[:alnum:]]*\)=.*/\1/p')`

Going back in time, prior to April, that said command looked like this (with back ticks around the command):

 if var=`safeEcho "$line" | sed -e 's/^\([a-zA-Z0-9_][a-zA-Z0-9_]*\)=.*/\1/'`; then
        
  Manually testing that sed command, I find that it returns the whole line; e.g. x=y.    So that is also wrong and it appears this never worked.
  
 Note however, that these old commands like sed and awk have not kept up with changes that have been made for these programs on other platforms, but I believe they can be updated from third parties.   So maybe on some Z systems where sed has been upgraded they work.

jimblye:
Fixes #26103

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -831,15 +831,15 @@ readServerEnv()
       *=*)
           # The sed commands extracts the variable from the line but only if it contains alpha-numeric
           # characters. Only accept alphanumeric variable names to avoid eval complexities.
-          if [ "$uname" != "SunOS" ]; then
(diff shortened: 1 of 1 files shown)


hlhoots (approved):
LGTM

jimblye:
#build

jimblye:
#build

PR #22985, opened 2022-10-17
Generate SHA256 of TCK Jar

tevans78:
Find out the exact tck jar that was run
Generate a SHA256 hash
Publish in the result page
Use the exact version string from maven rather than passing it in

#build

For #22749

PR #26122, opened 2023-08-30
Remove top-level TCK failure

### LINKED ISSUE #26091: Remove top level TCK tests failed failure
Currently, if a TCK fails, we emit a failure listing the test methods that failed.

This is unfortunate because it means all the TCK defects that get reported look the same.

We should take that top-level failure out and just report the tests that failed.

Azquelt:
Stop generating a test failure which lists the tests which failed.

It's not that helpful since we will have the test failures themselves anyway and it confuses our defect analysis tools because the failure message is always very similar, regardless of which tests actually failed.

We made this change before in #4904, but the top-level failure got re-added amongst the changes in #22985

Fixes #26091


Azquelt:
#build

Azquelt:
Build ran out of disk space
#build

tevans78 (approved):

Azquelt:
Failures in the build do not appear to be related to this change. This can be merged when the branch is reopened.

PR #26132, opened 2023-08-30
Allow checkpoint java command to have a prefix

### LINKED ISSUE #26131: Disable processor-specific Glibc libraries from being loaded in InstantOn checkpoints
On Linux the ELF loader supports loading processor-specific libraries, which Glibc uses to provide `libpthreads`, `libm`, etc.

This mechanism needs to be disabled in order for InstantOn checkpoints to be restorable on older processors.

Here's an example of how it might be done in the Liberty server script:

```
# diff -u server.orig server
--- server.orig 2023-08-30 16:48:13.344600855 -0500
+++ server      2023-08-30 12:25:53.736028425 -0500
@@ -1050,7 +1050,7 @@
     # We direct STDERR to /dev/null to avoid that noise.
     {
       #  Launch with stdin set to /dev/null to avoid file-handle complications on criu restore
-      "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+      ld.so --glibc-hwcaps-mask "" "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
     } 2> /dev/null
     rc=$?
     if [ $rc -eq 20 ]; then
```

Note that

1. This affects all architectures because we're specifying an empty mask, i.e. no processor-specific library directories will be searched. In my quick inspection of various x86, Power, and Z Docker images I only saw UBI 8 and 9 ppc64le images take advantage of this, but future updates could easy do the same for other architectures.
2. It only works on UBI. On Ubuntu the loader is not called `ld.so` and it's not on the `PATH`. On 22.04 one would have to invoke ` /usr/lib64/ld-linux-x86-64.so.2` on x86, `/usr/lib64/ld64.so.2` on Power, etc, or we could possibly establish our own symlink, add it to the path, and so on.
3. It only affects the immediate process. Child processes are not affected and will go back to the default behaviour. We would need a different delivery method to get the loader to do the above for the entire process tree. If we want to do it outside of the server script we'd have to allow callers to override the `JAVA_CMD` variable or at least prepend something to it.

tjwatson:
Fixes #26131

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -1071,8 +1071,10 @@ javaCmd()
     #On some shells (bash) the shell will issue an ugly message to STDERR when criu kills the process.
     # We direct STDERR to /dev/null to avoid that noise.
     {
+      # JAVA_CHECKPOINT_PREFIX_CMD is used externally to allow prefixing commands before the
+      # java invocation.  For example, to make calls to ld.so to fine tune glibc behavior.
       #  Launch with stdin set to /dev/null to avoid file-handle complications on criu restore
-      "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+      ${JAVA_CHECKPOINT_PREFIX_CMD}"${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
     } 2> /dev/null
     rc=$?
     if [ $rc -eq 20 ]; then


tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1072,7 +1072,7 @@ javaCmd()
     # We direct STDERR to /dev/null to avoid that noise.
     {
       #  Launch with stdin set to /dev/null to avoid file-handle complications on criu restore
-      "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+      ld.so --glibc-hwcaps-mask "" "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
```
sebratton: Won't this negate error handling on platforms that can get to this code but where  ld.so does not exist (like macOS)? Today they would get a checkpoint not supported message.

tjwatson:
#build

tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1072,7 +1072,7 @@ javaCmd()
     # We direct STDERR to /dev/null to avoid that noise.
     {
       #  Launch with stdin set to /dev/null to avoid file-handle complications on criu restore
-      "${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
+      ${JAVA_CHECKPOINT_PREFIX_CMD}"${JAVA_CMD}" "$@" >> "${CHECKPOINT_CONSOLE_LOG}" 2>&1 < /dev/null
```
sebratton, line 1077: Can you document the variable ${JAVA_CHECKPOINT_PREFIX_CMD}? At least in general that it's there for use by by container build scripts

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
tjwatson: Yes, this is a problem.  We decided to instead add a prefix variable that can be set externally to prefix the JAVA_CMD

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
tjwatson: Done

sebratton (approved):

sebratton:
#libby

PR #26137, opened 2023-08-31
Automatic method validation for parameters and return types

njr-11:
Method validation for parameters and return types when Jakarta Validation is enabled.
Initially this will always attempt validation.


njr-11:
#build

mswatosh (approved):

PR #26143, opened 2023-09-01
26141 java 17 error message

### LINKED ISSUE #26141: We need a better error message for running java 17 without java_home set

benjamin-confino:
Fixes #26141

### DIFF
--- a/dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
+++ b/dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
 unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
+
+MODULE_NOT_OPENED.CWOWB1019E=CWOWB1019E: Running on Java 17+ without exposing the modules needed for CDI to create proxies.
+MODULE_NOT_OPENED.CWOWB1019E.explanation=Unlike previous versions of Java, Java 17 does not allow calling setAccessible(true) on a field or method of an encapsulated class. This is necessary for CDI to create its proxies.
+MODULE_NOT_OPENED.CWOWB1019E.useraction=Set the JAVA_HOME environment variable to the path of a Java SDK, and then try the command again. If this is insufficient, add the following JVM argument: --add-opens=java.base/java.lang=ALL-UNNAMED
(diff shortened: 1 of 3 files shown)


dev/io.openliberty.cdi.4.0.internal.weld/src/com/ibm/ws/cdi/proxy/ProxyServicesImpl.java
```diff
@@ -148,6 +153,9 @@ public Class<?> defineClass(Class<?> originalClass, String className, byte[] cla
         Object classLoaderLock = null;
         try {
             classLoaderLock = ClassLoaderMethods.getClassLoadingLock.invoke(loader, className);
+        } catch (ExceptionInInitializerError e) {
+            Tr.error(tc, "MODULE_NOT_OPENED.CWOWB1019E");
+            throw new RuntimeException(e);
```
Azquelt: I'm unsure about this approach in general, but if we decide to try to catch it here, there are some issue with this code.

1. `ExceptionInInitializerError` is a very generic error, it just means that some static initialization code for a class threw an exception. There's no indication that the problem is caused by the missing JVM arguments. We would need to look up the cause chain to see if the actual exception is the one we know about.
2. We should rethrow `e` rather than wrapping it in a `RuntimeException` (this is allowed because the it's an error and therefore unchecked)

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
 unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
+
+MODULE_NOT_OPENED.CWOWB1019E=CWOWB1019E: Running on Java 17 without exposing the modules needed for CDI to create proxies.
```
abutch3r: Would the same issue apply to >17 as well? e.g. 21, as the message currently implies that only 17 is affected and the behavior has changed back in later versions, which I assume it hasn't. So if someone on Java21 hit this while the error is confusing given they would likely know they are on 21 and we are seemingly stating they are on 17. So saying `17+` would allow for going forward.

For the explanation saying `Java 17 onwards,` instead of `Unlike previous versions of Java, Java 17` again supports multiple versions 

For the user action I think this can be simplified and more in line with existing messages such as 
https://github.com/OpenLiberty/open-liberty/blob/6e1dca820be198e36a4c7f3e01eee9a9851e653f/dev/com.ibm.ws.kernel.cmdline/resources/com/ibm/ws/kernel/boot/cmdline/resources/CommandLineMessages.nlsprops#L36
so 
`Set the JAVA_HOME environment variable to the path of a Java SDK, and then try the command again. If this is insufficient, add the following JVM argument: --add-opens=java.base/java.lang=ALL-UNNAMED`

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
...
+MODULE_NOT_OPENED.CWOWB1019E.explanation=Unlike previous versions of Java, Java 17 does not allow calling setAccessible(true) on a field or method of an encapsulated class. This is nessacary for CDI to create its proxies.
```
abutch3r: Replace `nessacary` with `necessary`

Azquelt:
I'm not sure about doing this check here in CDI. Would it be possible to do the check at server startup and explicitly check (a) that the the required module opens are not present and (b) that we're running on a java version that requires them.

benjamin-confino:
As the kernel team have prioritised this CDI does not need its own error logging.

PR #26146, opened 2023-09-01
Gradle build.image change date format

jjiwooLim:
It looks like there was a change in Locale in the build definition which affected the date format. The expected output is `Fri Sep 1 2023` but it was `Fri Sept 1 2023`

The followings are formatted output for each locale.

Locale.UK:
`Fri Sept 1 2023`

Locale.CANADA:
`Fri. Sep. 1 2023`

Locale.US:
`Fri Sep 1 2023`

Locale.ENGLISH:
`Fri Sep 1 2023`

We changed the locale to use Locale.ENGLISH and updated Java date time API to use newer API.

### DIFF
--- a/dev/build.image/build.gradle
+++ b/dev/build.image/build.gradle
@@ -981,9 +981,14 @@ if (isAutomatedBuild && !isIFIXBuild) {
 
             def rpms = file("./packaging/rpmbuild/RPMS/noarch/openliberty-" + bnd.get('libertyServiceVersion') + "-1.noarch.rpm")
             def debs = file("./packaging/debuild/openliberty_" + bnd.get('libertyServiceVersion') + "-1ubuntu1_all.deb")
-            def packageDate = new Date()
-            def formattedRPMDate = packageDate.format("E MMM d y")
-            def formattedDEBDate = packageDate.format("E, d MMM y hh:mm:ss Z")
+
+            ZonedDateTime packageDate = ZonedDateTime.now();
+            def formatter = java.time.format.DateTimeFormatter.ofPattern("E MMM d y", Locale.ENGLISH);
+            def formattedRPMDate = packageDate.format(formatter)
+
+            formatter = java.time.format.DateTimeFormatter.ofPattern("E, d MMM y hh:mm:ss Z", Locale.ENGLISH);
+            def formattedDEBDate = packageDate.format(formatter);
+	    
 
             inputs.dir('packaging/tempPackagingDir')
             outputs.file(rpms)


cbridgha (approved):
Looks good

jacobwdv:
Sarah and I believe the build break was caused by a change to the build server environment that updated the default Locale within the JVM that gradle uses. This could have been a new JDK or a new build server with a different Locale. 

The **expected** string from our build on Debian systems is: `Fri, 1 Sep 2023 12:52:50 -0400`

The **actual** string that broke the build becuase it was non-conformant to the dpkg-buildpackage script was: `Fri, Sept 2023 12:52:50 -0400`

When the build broke, we believe the Locale detected by the JVM was Locale.UK. It was presumably `Locale.ENGLISH` or `Locale.US` before (assuming our builds passed before today on DEB systems). 

Sarah's change explicitly specifies Locale.ENGLISH when constructing the `DateTimeFormatter` object used to parse ZoneDateTime into the formatted string. 

Explicitly specifying the Locale _could_ cause issues in the future when building on non-English systems depending on how dates are expected to be formatted by `dpkg-buildpackage` on non-English systems, or on English systems with different locales for that matter. 

Originally, the `Date` class was used to parse the current date/time and we used the `format()` method to generate the appropriate String. We could have kept using `Date` with a `SimpleDateFormatter` specifying Locale.English in the constructor. This produced the same result as using a `ZonedDateTime` object with a `DateTimeFormatter` as was done in our fix. 

```
def currentDate = new Date()
def dateFormat = new java.text.SimpleDateFormat("E, d MMM y hh:mm:ss Z", Locale.ENGLISH)
def formattedDEBDate = dateFormat.format(currentDate)
```

However, while researching this issue there were warnings about the `Date` class being buggy and outdated: https://stackoverflow.com/questions/67089932/simpledateformat-format-month-september-jdk16

We decided to change our gradle script to use ZonedDateTime instead. This decision added some additional risk to our code change since we're now changing the class that captures the current date instead of re-using the one that worked previously. We decided to use the more modern API after reading recommendations from colleagues and forums online. If this causes issues in the future, we might want to go back to the Date class as shown in the example above.

### Additional considerations. 

Instead of specifying a hard coded Locale.ENGLISH, we could have also used custom hardcoded DateFormatSymbols. This was not  as elegant or easy to understand as our final solution but could be used in the future if 

1. Java changes the formatting of dates in Locale.ENGLISH and,
2.  dpkg-buildpackage doesn't accept them.

```
def currentDate = new Date()
def dateFormat = new java.text.SimpleDateFormat()
def dateSymbols = dateFormat.getDateFormatSymbols();
String[] months = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"]
dateSymbols.setShortMonths(months)
dateFormat.setDateFormatSymbols(dateSymbols)
dateFormat.applyPattern("E, d MMM y hh:mm:ss Z")
println(dateFormat.format(currentDate))
```

The code sample above shows how you could specify the exact strings used as short months in a String array. The same can be done for other date formatter attributes. If we were to go this route we would need to also change the short weekday strings as they can vary depending on the System Locale as well.

PR #26149, opened 2023-09-01
FeatureList cache product feature list manifest entries

cbridgha:
Discovered the featureList utility was not saving the initial scan / read of all feature MANIFEST's - and in some cases this is called thousands of time...  as the # of features has grown, this has become too expensive, so some optimization is warranted...


cbridgha:
#build

PR #26152, opened 2023-09-01
Fix timing issue in concurrent timer demo

tkburroughs:
Using an instance variable in a stateless EJB to count timers is unreliable as multiple instances may be used.  Switch the count variable to static so count is accurate no matter which instance is used to run the timers or report the count.


tkburroughs:
#build

njr-11 (approved):

mswatosh (approved):

PR #26157, opened 2023-09-04
26155 rework acquistion open telemetry info

benjamin-confino:
This creates a new path for acquring  OpenTelemetryInfo via static methods rather than CDI.


dev/io.openliberty.microprofile.telemetry.common/resources/io/openliberty/microprofile/telemetry/common/resources/MPTelemetry.nlsprops
```diff
@@ -37,4 +37,9 @@ CWMOT5000.cannot.get.globalopentelemetry.useraction=Use CDI to inject the OpenTe
 
 CWMOT5001.cannot.set.globalopentelemetry=CWMOT5001E: Setting the GlobalOpenTelemetry object is not supported.
 CWMOT5001.cannot.set.globalopentelemetry.explanation=The GlobalOpenTelemetry class is not supported because no global OpenTelemetry object exists. Instead, each application must use CDI to inject the OpenTelemetry object for that application.
-CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
\ No newline at end of file
+CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
+
+CWMOT5002.telemetry.error=CWMOT5002E: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
+CWMOT5002.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
```
ramkumar-k-9286:
```suggestion
CWMOT5002.telemetry.error.explanation=An error occurred, preventing the normal operation of MicroProfile Telemetry.
```

posmith:
L2 message review completed

benjamin-confino:
#build

benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}
+
+	public static Span getSpan() {
+		return new SpanProxy(); //TODO I cannot see a way to get no-op spans or baggage, check if we need code here to handle apps that have shut down.
+	}
+
+	public static Baggage getBaggage() {
+		return new BaggageProxy();
+	};
```
Azquelt: No-op `Span`: [Span.getInvalid()](https://www.javadoc.io/static/io.opentelemetry/opentelemetry-api/1.29.0/io/opentelemetry/api/trace/Span.html#getInvalid())

There isn't really a "no-op" `Baggage`, but we can get an empty one: [Baggage.empty()](https://www.javadoc.io/static/io.opentelemetry/opentelemetry-api/1.29.0/io/opentelemetry/api/baggage/Baggage.html#empty())

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/cdi/OpenTelemetryProducer.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private static final TraceComponent tc = Tr.register(OpenTelemetryProducer.class);
+
+    J2EEName j2EEName = null;
+    
+    private J2EEName getJ2EEName() {
+    	if (j2EEName == null) {
+    		j2EEName = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData().getJ2EEName();
+    	}
+    	return j2EEName;
```
Azquelt: Mix of tabs and spaces

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
Azquelt: That said, I'm not sure it's necessary here. `SpanProxy` and `BaggageProxy` will call `Span.current()` and `Baggage.current()` respectively which should do something sensible.

I'm not sure there's any need for these objects outside CDI though, where we need an object which will reflect the current `Span` or `Baggage` changing. I think most other code could just call `Span.current()` or `Baggage.current()` whenever they need it, so we could omit these methods and create the objects in the Producer.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/clientnocdi/ClientTriggeringServlet.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        LOGGER.info(">>> testClientWithNoCDIJaxClient");
+
+        //Verify that CDI is disabled.
+        try {
+          CDI.current().getBeanManager();
+          //fail("We found a bean manager. CDI is enabled");
+        } catch (Exception e) {
```
Azquelt: Fail line should be uncommented.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.resources.ResourceBuilder;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+
+// We want this to start before CDI so it clears the list of stopped apps before CDI gets to them.
+@Component(property = { "service.vendor=IBM",  "service.ranking:Integer=150" })
+public class OpenTelemetryInfoFactory implements ApplicationStateListener {
+	
+	/*
```
Azquelt: Mix of tabs and spaces in this class, should be 4 space indents throughout.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
benjamin-confino: I was thinking its good to make the CDI and static entry points to open-telemetry identical, but if you think calling `current()` here is better I can make the change.

benjamin-confino:
#build 

#spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

benjamin-confino:
#25780 #spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

benjamin-confino:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -0,0 +1,90 @@
...
+    public static void setUp() throws Exception {
+        PropertiesAsset appConfig = new PropertiesAsset()
+                        .addProperty("otel.sdk.disabled", "false")
+                        .addProperty("otel.traces.exporter", "in-memory")
+                        .addProperty("otel.bsp.schedule.delay", "100");
+
+        WebArchive noCDIApp = ShrinkWrap.create(WebArchive.class, NO_CDI_APP_NAME + ".war")
+                        .addPackage(ClientTriggeringServlet.class.getPackage());
+        CDIArchiveHelper.addBeansXML(noCDIApp, com.ibm.websphere.simplicity.beansxml.BeansAsset.DiscoveryMode.NONE);//Not strictly needed unless we expand this test to cover MP rest client.
+
```
Azquelt, line 67: As a follow up: could we have imports instead of fully qualified names please

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -0,0 +1,90 @@
...
+        ShrinkHelper.exportAppToServer(server, noCDIApp, SERVER_ONLY);
+        server.startServer();
+    }
+
+    //This test does not use open telemetry directly, but it does trigger io.openliberty.microprofile.telemetry.internal.rest.TelemetryClientFilter.java for an app with no CDI and ensures it does not throw an exception
+    @Test
+    public void testClientWithNoCDi() throws Exception {
+        HttpRequest httpRequest = new HttpRequest(server, "/" + NO_CDI_APP_NAME + "/ClientTriggeringServlet");
+        assertEquals(io.openliberty.microprofile.telemetry.internal_fat.apps.clientnocdi.ClientInvokedServlet.TEST_PASSED, httpRequest.run(String.class));
```
Azquelt, line 77: As a followup: If possible, can we make this call using FATServletClient and FATServlet on the server side so that any exceptions get reported nicely.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -0,0 +1,90 @@
...
+
+        assertFalse("The test did not use a TelemetryClientFilter object", 
+                     server.findStringsInLogsUsingMark("microprofile.telemetry\\d*.internal.rest.TelemetryClientFilter < isEnabled Exit", server.getDefaultTraceFile())
+                     .isEmpty());
```
Azquelt, line 81: As a followup: see if we can change `InMemorySpanExporter` so that it doesn't require CDI so that we can actually assert that correct spans are created. It's mostly static anyway so it shouldn't require too much change.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryMisconfigTest.java
```diff
@@ -63,76 +63,67 @@ public class TelemetryMisconfigTest extends FATServletClient {
...
+        doesNotExistEndpointApp = ShrinkWrap.create(WebArchive.class, DOES_NOT_EXIST_ENDPOINT_APP_NAME + ".war")
                         .addPackage(JaxRsMisConfigEndpoints.class.getPackage())
                         .addAsResource(new StringAsset("otel.sdk.disabled=false\notel.traces.exporter=jaeger\notel.exporter.jaeger.endpoint=" + DOES_NOT_EXIST_ENDPOINT),
                                        "META-INF/microprofile-config.properties");
 
-        ShrinkHelper.exportAppToServer(server, doesNotExistEndpointApp, SERVER_ONLY);
-
-        server.startServer();
+        //Don't validate the apps because they have not been deployed yet.
+        server.startServerAndValidate(LibertyServer.DEFAULT_PRE_CLEAN, LibertyServer.DEFAULT_CLEANSTART, false);
```
Azquelt, line 120: I thought this shouldn't be needed because we haven't deployed the apps yet, but Ben pointed out that they are listed in the server.xml since they use third-party so this might be needed.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryMisconfigTest.java
```diff
@@ -219,17 +206,21 @@ public void testDoesNotExistEndpoint() throws Exception {
...
         new HttpRequest(server, "/" + INVALID_JAEGER_TIMEOUT_APP_NAME + "/misconfig/jaxrsclient")
                         .expectCode(200)
                         .run(String.class);
 
         assertNotNull(server.waitForStringInLogUsingMark("Invalid duration property otel.exporter.jaeger.timeout"));
     }
 
+    private static void deployAndWaitForApp(WebArchive app, String appName) throws Exception {
+        server.setMarkToEndOfLog();
+        ShrinkHelper.exportAppToServer(server, app, SERVER_ONLY);//This will set the mark for us in
```
Azquelt, line 219: As a followup: we think this comment should be removed.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/clientnocdi/ClientTriggeringServlet.java
```diff
@@ -0,0 +1,80 @@
...
+            fail("We found a bean manager. CDI is enabled");
+        } catch (Exception e) {
+            //ignore
+        }
+
+        String url = request.getRequestURL().toString();
+        url = url.replace("ClientTriggeringServlet", "ClientInvokedServlet"); //The jaxrsclient will use the URL as given so it needs the final part to be provided.
+
+        String result = client.target(url).request(MediaType.TEXT_PLAIN).get(String.class);
+
```
Azquelt, line 72: As noted before, it would be good,in a followup, to assert the created span here if we can change `InMemorySpanExporter` to not require CDI.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/publish/servers/Telemetry10JaxWithLogging/server.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <featureManager>
+        <feature>servlet-6.0</feature>
+        <feature>mpTelemetry-1.0</feature>
+        <feature>componentTest-2.0</feature>
+        <feature>restfulWS-3.1</feature>
+        <feature>mpRestClient-3.0</feature>
+    </featureManager>
+
+    <javaPermission className="java.lang.RuntimePermission" name="getClassLoader"/>
```
Azquelt: We think this was needed for a call from feature code.

If that's the case, we should have a doPriv there and this permission should be removed.

We need this test to pass without this permission (unless we find that test code needs it).

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryClientBuilderListener.java
```diff
@@ -0,0 +1,34 @@
...
+import io.openliberty.restfulWS.client.ClientBuilderListener;
+import jakarta.ws.rs.client.ClientBuilder;
+
+/**
+ * Adds the Telemetry client filter when a rest client is built.
+ * <p>
+ * This gets called for both JAX-RS Client and MP Rest Client.
+ */
+@Component(configurationPolicy = IGNORE)
+public class TelemetryClientBuilderListener implements ClientBuilderListener {
```
Azquelt, line 25: As a followup: this appears to contradict `TelemetryRestClientBuilderListener` which appears to handle MP Rest Client. Check which is correct and fix the comments.

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryClientBuilderListener.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public void building(ClientBuilder clientBuilder) {
+        TelemetryClientFilter currentFilter = new TelemetryClientFilter();
+        if (currentFilter.isEnabled()) {
+            clientBuilder.register(currentFilter);
+        }
```
Azquelt, line 32: As a followup: can we add a try-catch here like the one in `TelemetryRestClientBuilderListener`.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+
+    public static Span getSpan() {
+        return new SpanProxy(); //TODO I cannot see a way to get no-op spans or baggage, check if we need code here to handle apps that have shut down.
+    }
+
+    public static Baggage getBaggage() {
+        return new BaggageProxy();
+    };
```
Azquelt: These should be removed. There's no reason for anyone to use the proxy objects, except as CDI beans.

Other code should just call `Span.current()` or `Baggage.current()`.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    public static Tracer getTracer(J2EEName j2EEName) {
+        try {
+            return getOpenTelemetryInfo(j2EEName).getOpenTelemetry().getTracer(INSTRUMENTATION_NAME);
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return OpenTelemetry.noop().getTracerProvider().get("");
+        }
+    }
```
Azquelt, line 108: Is this creating a new `Tracer` on every call?

Azquelt: [Javadoc](https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/1.29.0/io/opentelemetry/api/OpenTelemetry.html) says "gets or creates a named tracer" so it probably won't but it would be good to confirm as it could be a performance issue if we're creating a new tracer on every call.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+            }
+
+            final Map<String, String> telemetryProperties = getTelemetryProperties();
+
+            //Builds tracer provider if user has enabled tracing aspects with config properties
+            if (!checkDisabled(telemetryProperties)) {
+                OpenTelemetry openTelemetry = AccessController.doPrivileged( (PrivilegedAction<OpenTelemetry>) () -> {
+                    //This contains API calls that change between the upstream open telemetry version.
+                    //We use @Reference because the callers coming into  OpenTelemetryInfoFactory via
+                    //The static accessor do not know which version of mpTelemetry will be in use.
```
Azquelt, line 142: As a followup: Out of date comment here, we no longer use `@Reference`

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+    public static Baggage getBaggage() {
+        return new BaggageProxy();
+    };
+
+    //The key methods that provide access to telemetry objects end here
+
+    private static OpenTelemetryInfo createOpenTelemetryInfo(J2EEName j2EEName) {
+        //If the app has already shut down but something still tries to get OpenTelemetry objects return a no-op object
+        if (shutDownApplications.contains(j2EEName)){ //There is a low risk race condition here. Thread A gets past this check just as thread B shuts everything down.
+            Tr.warning(tc, Tr.formatMessage(tc, "CWMOT5003.factory.used.after.shutdown"));
```
Azquelt: Need to check message on this, should it be a general "internal problem warning"?

As a follow up: also, I think it should contain the app name since we have it.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+
+                if (openTelemetry != null) {
+                    return new OpenTelemetryInfo(true, openTelemetry);
+                }
+            }
+            //By default, MicroProfile Telemetry tracing is off.
+            //The absence of an installed SDK is a “no-op” API
+            //Operations on a Tracer, or on Spans have no side effects and do nothing
+            ComponentMetaData cData = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData();
+            String applicationName = cData.getJ2EEName().getApplication();
```
Azquelt, line 163: As a followup: You already have the J2EEName as an argument to this method, so you don't need to look it up again.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+                                    .addPropertiesCustomizer(x -> telemetryProperties) //Overrides OpenTelemetry's property order
+                                    .addResourceCustomizer(OpenTelemetryInfoFactory::customizeResource) //Defaults service name to application name
+                                    .setServiceClassLoader(Thread.currentThread().getContextClassLoader())
+                                    .build()
+                                    .getOpenTelemetrySdk();
+                });
+
+                if (openTelemetry != null) {
+                    return new OpenTelemetryInfo(true, openTelemetry);
+                }
```
Azquelt, line 157: This `null` check looks weird, but `getOpenTelemetrySdk()` will return `null` if the SDK is disabled.

`checkDisabled()` should stop is getting to this point if that's the case, but we probably still want it in case there's some difference between the way we read config and the way OTel reads it.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+        try {
+            if (AgentDetection.isAgentActive()) {
+                return;
+            }
+
+            OpenTelemetry openTelemetry = null;
+
+            if (appToInfo.containsKey(j2EEName)) {
+                openTelemetry = appToInfo.get(j2EEName).getOpenTelemetry();
+            }
```
Azquelt, line 185: As as followup: I think you want to call `get()` and check for `null` here.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,358 @@
...
+            return new OpenTelemetryInfo(false, OpenTelemetry.noop());
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return new OpenTelemetryInfo(false, OpenTelemetry.noop());
+        }
+
+    }
+
+
+    public synchronized void disposeOpenTelemetry(J2EEName j2EEName) {
```
Azquelt, line 175: As a followup: Don't think this needs to be `synchronized`, we only call it from the dispose method and there shouldn't be a problem calling it for two applications at the same time since `appToInfo` is a concurrent map.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/cdi/OpenTelemetryProducer.java
```diff
@@ -0,0 +1,72 @@
...
+    @ApplicationScoped
+    @Produces
+    public OpenTelemetryInfo getOpenTelemetryInfo() {
+        return OpenTelemetryInfoFactory.getOpenTelemetryInfo(getJ2EEName());
+    }
+
+    @Produces
+    public Tracer getTracer() {
+        return OpenTelemetryInfoFactory.getTracer(getJ2EEName());
+    }
```
Azquelt, line 53: As a follow up: we're not sure why this isn't an `ApplicationScoped` bean. Might be an idea to check back in the history and work out if there's a reason or if this is an oversight.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
Azquelt: We didn't see an nlsprops change in this PR. Check the message exists!

Azquelt: Also, can we dump the stack trace in the trace log if this happens so we can help users diagnose why they saw this warning.

benjamin-confino:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

dev/io.openliberty.microprofile.telemetry.internal.common/resources/io/openliberty/microprofile/telemetry/internal/common/resources/MPTelemetry.nlsprops
```diff
@@ -40,6 +40,9 @@ CWMOT5001.cannot.set.globalopentelemetry.explanation=The GlobalOpenTelemetry cla
 CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
 
 CWMOT5002.telemetry.error=CWMOT5002E: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
-CWMOT5002.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
+CWMOT5002.telemetry.error.explanation=An error occurred, preventing the normal operation of MicroProfile Telemetry.
```
dmuelle:
```suggestion
CWMOT5002.telemetry.error.explanation=An error occurred that prevented the normal operation of MicroProfile Telemetry.
```

dev/io.openliberty.microprofile.telemetry.internal.common/resources/io/openliberty/microprofile/telemetry/internal/common/resources/MPTelemetry.nlsprops
```diff
@@ -40,6 +40,9 @@ CWMOT5001.cannot.set.globalopentelemetry.explanation=The GlobalOpenTelemetry cla
...
 CWMOT5002.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
 
+CWMOT5003.factory.used.after.shutdown=CWMOT5003W: Application {0} attempted to acquire MicroProfile Telemetry after the application has shutdown.
```
dmuelle:
```suggestion
CWMOT5003.factory.used.after.shutdown=CWMOT5003W: The {0} application  attempted to acquire MicroProfile Telemetry after the application shut down. 
```

dev/io.openliberty.microprofile.telemetry.internal.common/resources/io/openliberty/microprofile/telemetry/internal/common/resources/MPTelemetry.nlsprops
```diff
@@ -40,6 +40,9 @@ CWMOT5001.cannot.set.globalopentelemetry.explanation=The GlobalOpenTelemetry cla
 CWMOT5001.cannot.set.globalopentelemetry.useraction=Do not call the GlobalOpenTelemetry.get method. Where the application code needs to obtain an OpenTelemetry object, use CDI to inject it.
 
 CWMOT5002.telemetry.error=CWMOT5002E: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
-CWMOT5002.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
+CWMOT5002.telemetry.error.explanation=An error occurred, preventing the normal operation of MicroProfile Telemetry.
 CWMOT5002.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
 
+CWMOT5003.factory.used.after.shutdown=CWMOT5003W: Application {0} attempted to acquire MicroProfile Telemetry after the application has shutdown. 
+CWMOT5003.factory.used.after.shutdown.explanation=After an application shuts down MicroProfile Telemetry returns objects that will silently perform no-op actions when invoked, this could lead to the absence of expected telemetry data.
```
dmuelle:
```suggestion
CWMOT5003.factory.used.after.shutdown.explanation=After an application shuts down, MicroProfile Telemetry returns objects that silently perform no-op actions when they are invoked. This process might lead to the absence of expected telemetry data.    
```

dev/io.openliberty.microprofile.telemetry.internal.common/resources/io/openliberty/microprofile/telemetry/internal/common/resources/MPTelemetry.nlsprops
```diff
@@ -40,6 +40,9 @@ CWMOT5001.cannot.set.globalopentelemetry.explanation=The GlobalOpenTelemetry cla
...
 
 CWMOT5002.telemetry.error=CWMOT5002E: An internal error occurred in MicroProfile Telemetry. The error is: {0}.
-CWMOT5002.telemetry.error.explanation=An error has occurred that prevents normal operation of MicroProfile Telemetry.
+CWMOT5002.telemetry.error.explanation=An error occurred, preventing the normal operation of MicroProfile Telemetry.
 CWMOT5002.telemetry.error.useraction=Review the server message.log file and FFDC logs to identify the problem.
 
+CWMOT5003.factory.used.after.shutdown=CWMOT5003W: Application {0} attempted to acquire MicroProfile Telemetry after the application has shutdown. 
+CWMOT5003.factory.used.after.shutdown.explanation=After an application shuts down MicroProfile Telemetry returns objects that will silently perform no-op actions when invoked, this could lead to the absence of expected telemetry data.    
+CWMOT5003.factory.used.after.shutdown.useraction=Review the application to see why it is attempting to use OpenTelemetry after it has shutdown. Actions that can trigger the use of OpenTelemetry include: Calling a method annotated with @WithSpan or making a request with JAX-RS Client or MP Rest Client.
```
dmuelle:
```suggestion
CWMOT5003.factory.used.after.shutdown.useraction=Review the application to see why it attempted to use MicroProfile Telemetry after  shutting down. Actions that can trigger the use of MicroProfile Telemetry include calling a method that is annotated with @WithSpan or making a request with a JAX-RS Client or MP Rest Client.
```

dmuelle (changes requested):
ID review to unblock PR

dmuelle (dismissed):
LGTM

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/publish/servers/Telemetry10JaxWithLogging/server.xml
Azquelt: This permission line needs to be removed (or justified) before this PR can be merged since if it's needed it represents a regression in mpTelemetry-1.1 with `SecurityManager` enabled

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/interfaces/OpenTelemetryAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+import com.ibm.ws.threadContext.ComponentMetaDataAccessorImpl;
+
+import io.openliberty.microprofile.telemetry.internal.common.OpenTelemetryInfo;
+import io.openliberty.microprofile.telemetry.internal.common.OpenTelemetryInfoFactory;
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+
+public class OpenTelemetryAccessor {
+
```
Azquelt, line 26: As a followup, this class and its methods should have Javadoc

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,357 @@
...
+        if (shutDownApplications.contains(j2EEName)){ //There is a low risk race condition here. Thread A gets past this check just as thread B shuts everything down.
+            Tr.warning(tc, Tr.formatMessage(tc, "CWMOT5003.factory.used.after.shutdown"));
+            if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
+                Exception e = new Exception();
+                ByteArrayOutputStream stackStream = new ByteArrayOutputStream();
+                PrintStream stackPrintStream = new PrintStream(stackStream);
+                e.printStackTrace(stackPrintStream);
+                
+                Tr.event(tc, "OpenTelemetryInfoFactory", "The stack that led to OpenTelemetryInfoFactory being called after {0} has shutdown is:.", j2EEName.getApplication());
+                Tr.event(tc, stackStream.toString());
```
Azquelt: As a followup: `Tr.event` doesn't do placeholder substitution on the message.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
Azquelt: Message has been added but we also need to pass in the application name as the placeholder value.

benjamin-confino:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
```diff
@@ -0,0 +1,357 @@
...
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return OpenTelemetry.noop().getTracerProvider().get("");
+        }
+    }
+
+    //The key methods that provide access to telemetry objects end here
+
+    private static OpenTelemetryInfo createOpenTelemetryInfo(J2EEName j2EEName) {
+        //If the app has already shut down but something still tries to get OpenTelemetry objects return a no-op object
+        if (shutDownApplications.contains(j2EEName)){ //There is a low risk race condition here. Thread A gets past this check just as thread B shuts everything down.
```
Azquelt, line 114: As a followup: fix this race condition.

We shouldn't allow a new `OpenTelemetry` to be created after the app has been shut down because at that point we won't clean it up again.

Azquelt (dismissed):

benjamin-confino:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

benjamin-confino:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
Azquelt: I did think of a way to make it clearer that this code is safe:

- Add a new class `OpenTelemetryInfoHolder`, with one field which is the `OpenTelemetryInfo` and another which is `boolean disposed`
- Change `appToInfo` to type `Map<J2EEName, OpenTelemetryInfoHolder>`
- On app start, add an empty holder to the map
- On `getOpenTelemetryInfo`, look up the holder from the map.
  - Sync on the holder.
  - If the holder is empty, check it hasn't been disposed, then create an `OpenTelemetryInfo` and store it in the holder
- On app shutdown, remove the holder from the map
  - Sync on the holder
  - If the removed holder is not empty, dispose of the `OpenTelemetryInfo`
  - Mark the holder disposed

This avoids having a separate map to track shut down apps and provides an easy object to synchronize on at the cost of an extra level of indirection when doing the lookup.

Azquelt: Tom also had the suggestion that we could store the `OpenTelemetryInfoHolder` in the app metadata itself, rather than using the app name as a key to a map.

Apparently the CDI code does this.

Azquelt (approved):

benjamin-confino:
#libby

benjamin-confino:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_xk7EYFJWEe6zvJP4IbpA3g&tab=com.ibm.team.build.web.ui.internal.editors.result.activity.ActivitiesProvider

Is the latest build and its green.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
benjamin-confino: Done

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/interfaces/OpenTelemetryAccessor.java
benjamin-confino: Done

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
benjamin-confino: Lets discuss this one.

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryClientBuilderListener.java
benjamin-confino: Done

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
benjamin-confino: Done

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
benjamin-confino: Done

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
benjamin-confino: Done

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactory.java
benjamin-confino: Done

PR #26160, opened 2023-09-04
Include Smallrye RM API as separate bundle

### LINKED ISSUE #25852: Resolve MP Reactive Messaging API Bundling
The SmallRye RM provider has its own version of the MP RM API which adds some experimental features. We need to use that version at runtime because the provider implementation requires it, but we need to package a jar in the `/dev` folder with the regular MP API.

Azquelt:
Smallrye Reactive Messaging has it's own version of the Reactive
Messaging API which includes extensions.

Use their version at runtime, but ensure that we package the spec
version in our dev directory.

Create a new bundle for the Smallrye API, change the existing API bundle
to use the spec API and update the feature file to use the right one at
runtime.

Fixes #25852


Azquelt:
#libby
#build

abutch3r (approved):
LGTM

PR #26178, opened 2023-09-05
More changes to handle iSeries test machine host name/address resolut…

c00crane:
Our iSeries test machines are having problems resolving their host name/addresses.  Most of the tests use localhost and run just fine.  The jwk endpoint tests need to use the real host info and on our iSeries machines, they sometimes run into problems. 
I've added logic to the test setup method that will try to use the resolved ip address.  If that works, we'll run the tests.  If this address is NOT usable, we'll skip starting the RS server (used in this test class) and then skip each of the tests.
We have enough coverage on all other platforms and in the cases where the name/address resolve correctly in iSeries.


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.jwt_fat.builder

ayoho (approved):

Issue #25697, opened 2023-07-12
Validation in Jakarta Data


njr-11:
User story:
"As a user of Jakarta Data, I would like Jakarta Data to integrate with Jakarta Validation so that entities can be validated for me."

The user can define validation constraints on entities.  Look into how validation could be automatically done, maybe triggered by `@Valid` on save operations or something else and how constraint violations are surfaced to the user.

```
@Repository
public interface Cars extends DataRepository<Car, Long> {
   @Valid 
   public void save(Car car);
}
```

```
@Repository
public interface Cars extends DataRepository<Car, Long> {
   public void save(@Valid Car car);
}
```

Micronaut does something similar to this:
```
@Repository
public interface Cars extends CrudRepository<@Valid Car, Long> {
   ...
}
```

mswatosh:
Jakarta Persistence Section on Bean Validation:
https://jakarta.ee/specifications/persistence/3.0/jakarta-persistence-spec-3.0.html#a2366

PR #26182, opened 2023-09-06
Use getters instead of public field

### LINKED ISSUE #26181: Weld Warnings on app startup about io.openliberty.data.internal.persistence.cdi.DataExtension
Weld throws warnings when a CDI extension has non-static public fields. To avoid this we can wrap those fields in getters. 

(This is because CDI extensions are also ApplicationScoped beans and a weld proxy cannot redirect access to a public field but can for a getter)

benjamin-confino:
fixes #26181


benjamin-confino:
#build

njr-11 (approved):

benjamin-confino:
#build

benjamin-confino:
Fixed by #25697

PR #26197, opened 2023-09-06
Resolve symbolic links when looking for java

### LINKED ISSUE #18412: If JAVA_HOME is not set the java9.options file is not always read by the server script
This bit of logic in the `server` script has issues:

```
  JPMS_MODULE_FILE_LOCATION="${JAVA_HOME}/lib/modules"
  if [ -z "${JAVA_HOME}" ]; then
    JPMS_MODULE_FILE_LOCATION=$(dirname $(dirname $(which java)))/lib/modules
  fi

  # If we are running on Java 9, apply Liberty's built-in java 9 options
  if [ -f "${JPMS_MODULE_FILE_LOCATION}" ]; then
    mergeJVMOptions "${WLP_INSTALL_DIR}/lib/platform/java/java9.options"
  fi

```

Typically on linux systems like ubuntu the java executable file is symbolically linked like this:

```
$ ls -l /usr/bin/java
lrwxrwxrwx 1 root root 22 Apr 10 20:27 /usr/bin/java -> /etc/alternatives/java
$ ls -l /etc/alternatives/java
lrwxrwxrwx 1 root root 57 Aug 19 10:17 /etc/alternatives/java -> /usr/lib/jvm/<java installation>/bin/java

```

Then the symbolic links are manipulated with something like `sudo update-alternatives --config java`

The problem is that $(dirname $(dirname $(which java))) is not following the symbolic links.  Where `which java` returns `/usr/bin/java` the result ends up being `/usr/lib/modules` which does not exist.  So we end up not reading in the java9.options.

Could look at using something like `readlink` to get the real path to where the `java` executable is.

jimblye:
Fixes #18412

Tested all 3 implementations on Solaris, Mac, and Ubuntu (in standalone script).   The **realpath** command works on all 3 platforms.  Unfortunately, it doesn't work on z/OS..  Hence the fallback methods. The **readlink -e** command doesn't work on mac, because readlink on mac does not have a "-e" option.   The more basic fallback implementation works on Ubuntu, Mac & Solaris and z/OS (after lots of fiddling on z/OS).  

The whole point of doing this is to find out if we are using version 9 or above.  I'm wondering why we can't just do this:

```
java_version=$("$JAVA_CMD" -version 2>&1 | awk -F '"' '/version/ {print $2}')

# Extract the major version (e.g., 1, 9, 10, 11). 
 # Java switched from using 1.x version numbers in version 9
major_version=$(echo "$java_version" | awk -F. '{print $1}')

 # If we are running on Java 9 and up, apply Liberty's built-in java 9 options 
if [ "$major_version" -ge 9 ]; then
    mergeJVMOptions "${WLP_INSTALL_DIR}/lib/platform/java/java9.options"
fi
```
JAVA_CMD has already been determined by the time we reach this point in the code.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
+++ b/dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
@@ -438,6 +438,10 @@ warning.unableToPackageLooseConfigFileMissingPath=CWWKE0095W: The server cannot 
 warning.unableToPackageLooseConfigFileMissingPath.explanation=The specified loose configuration application XML file contains invalid information about the location of the files. For example, it might specify a directory that cannot be found. Paths in this XML file should be relative to the configuration directory of the server. The archive represented by this file will not be included in the packaged server output.
 warning.unableToPackageLooseConfigFileMissingPath.useraction=Modify the paths in the loose configuration application XML file to point to the correct location that contains the application files. 
 
+warning.tooManySymbolicLinks=CWWKE0096W: Too many symbolic links encountered when trying to resolve a path ( {0} ).
+warning.tooManySymbolicLinks.explanation=Symbolic links are shortcuts that point to other files or directories. This error occurs when the system encounters too many symbolic links in a chain, making it unable to resolve the path.
+warning.tooManySymbolicLinks.useraction=To resolve this issue, consider checking the symbolic links in the path, and ensure that they do not create circular references. You may also want to use the 'readlink' command to inspect individual links in the chain.
+
 warn.packageRuntime.include.unknownOption=CWWKE0099W: Unable to use option --include={0}, will use --include=wlp instead.
 warn.packageRuntime.include.unknownOption.explanation=The option was ignored because it was not recognized.
 warn.packageRuntime.include.unknownOption.useraction=Retry the command using valid options. Try using --help for a list of valid options.
--- a/dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
+++ b/dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
@@ -438,8 +438,8 @@ warning.unableToPackageLooseConfigFileMissingPath=CWWKE0095W: The server cannot 
 warning.unableToPackageLooseConfigFileMissingPath.explanation=The specified loose configuration application XML file contains invalid information about the location of the files. For example, it might specify a directory that cannot be found. Paths in this XML file should be relative to the configuration directory of the server. The archive represented by this file will not be included in the packaged server output.
 warning.unableToPackageLooseConfigFileMissingPath.useraction=Modify the paths in the loose configuration application XML file to point to the correct location that contains the application files. 
 
-warning.tooManySymbolicLinks=CWWKE0096W: Too many symbolic links encountered when trying to resolve a path ( {0} ).
-warning.tooManySymbolicLinks.explanation=Symbolic links are shortcuts that point to other files or directories. This error occurs when the system encounters too many symbolic links in a chain, making it unable to resolve the path.
+warning.tooManySymbolicLinks=CWWKE0096W: Too many symbolic links were encountered when the system tried to resolve the following path: {0}.
+warning.tooManySymbolicLinks.explanation=Symbolic links are shortcuts that point to other files or directories. This error occurs when the system encounters too many symbolic links in a chain and it is unable to resolve the path.
 warning.tooManySymbolicLinks.useraction=To resolve this issue, check the symbolic links in the path and ensure that they do not create circular references. You might also use the 'readlink' command to inspect individual links in the chain.
 
 warn.packageRuntime.include.unknownOption=CWWKE0099W: Unable to use option --include={0}, will use --include=wlp instead.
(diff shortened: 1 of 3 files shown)


jimblye:
#build

jimblye:
#build

tjwatson:
> The whole point of doing this is to find out if we are using version 9 or above. I'm wondering why we can't just do this:

I think we avoid doing `java -version` to avoid invoking `java` as that takes extra time to do that.  But then again, that may be just as fast as the realpath implementation and the searching for the modules directory.  Besides that, this is a fallback when JAVA_HOME is not set.  If that is slow then the proper thing to do is set JAVA_HOME.  For our container images we always have JAVA_HOME set.

tjwatson:
I will say running `java -version` locally is not fast at all.  I think it takes 10s of millis.  That will not be acceptable for an InstantOn regression.  But then again this is for the case where JAVA_HOME is not set.  For InstantOn (in-container) we always have JAVA_HOME set.

covener:
Not working on zOS for me when running under /bin/sh -- on my system it falls back to `cd -P` and the builtin cd for /bin/sh doesn't know `-P`.

We have some similar stuff in IHS and we basically parse `ls -l` output.

```
\ls -l $F | sed -e "s#.*-> ##"

$ ls -l ~/.bash_profile
lrwxrwxrwx   1 WASTST1  WASUSER       28 Aug 29 13:01 /u/WASTST1/.bash_profile -> /u/WASTST1/bin/.bash_profile
WASTST1@BOSS0152:~/SRC/2.4.x (zformlogin) $ ls -l ~/.bash_profile | sed -e "s#.*-> ##"
/u/WASTST1/bin/.bash_profile
```

We do not have LANG=C or any locale stuff  --  it might be a risk.

jimblye:
#build

jimblye:
#build

jimblye:
Ready for another review.   This issue doesn't address Windows.  Probably because Windows doesn't pretend to use java9.options if JAVA_HOME is not set.  Well, until now.   I updated the server.bat script to derive JAVA_HOME from the PATH.   Unfortunately, I don't know if there is a way that we can resolve symbolic links on Windows without requiring PowerShell, and I'm not sure how easy it would be to implement even with PowerShell.    With this update, at least some of the time we'll use java9.options on Windows without having JAVA_HOME set.   Though, at this point, I suspect most people are used to setting JAVA_HOME with Liberty.

On the non-Windows side of things I added a counter to bail out if we follow > 100 symbolic links.   If you get that far, I'm pretty there's a circular link.

dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
```diff
@@ -438,6 +438,10 @@ warning.unableToPackageLooseConfigFileMissingPath=CWWKE0095W: The server cannot
 warning.unableToPackageLooseConfigFileMissingPath.explanation=The specified loose configuration application XML file contains invalid information about the location of the files. For example, it might specify a directory that cannot be found. Paths in this XML file should be relative to the configuration directory of the server. The archive represented by this file will not be included in the packaged server output.
 warning.unableToPackageLooseConfigFileMissingPath.useraction=Modify the paths in the loose configuration application XML file to point to the correct location that contains the application files. 
 
+warning.tooManySymbolicLinks=CWWKE0096W: Too many symbolic links encountered when trying to resolve a path ( {0} ).
```
dmuelle:
```suggestion
warning.tooManySymbolicLinks=CWWKE0096W: Too many symbolic links were encountered when the system tried to resolve the following path: {0}.
```

dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
```diff
@@ -438,6 +438,10 @@ warning.unableToPackageLooseConfigFileMissingPath=CWWKE0095W: The server cannot
...
+warning.tooManySymbolicLinks.explanation=Symbolic links are shortcuts that point to other files or directories. This error occurs when the system encounters too many symbolic links in a chain, making it unable to resolve the path.
```
dmuelle:
```suggestion
warning.tooManySymbolicLinks.explanation=Symbolic links are shortcuts that point to other files or directories. This error occurs when the system encounters too many symbolic links in a chain and it is unable to resolve the path.
```

dev/com.ibm.ws.kernel.boot/resources/com/ibm/ws/kernel/boot/resources/LauncherMessages.nlsprops
```diff
@@ -438,6 +438,10 @@ warning.unableToPackageLooseConfigFileMissingPath=CWWKE0095W: The server cannot
...
+warning.tooManySymbolicLinks.useraction=To resolve this issue, consider checking the symbolic links in the path, and ensure that they do not create circular references. You may also want to use the 'readlink' command to inspect individual links in the chain.
```
dmuelle:
```suggestion
warning.tooManySymbolicLinks.useraction=To resolve this issue, check the symbolic links in the path and ensure that they do not create circular references. You might also use the 'readlink' command to inspect individual links in the chain.
```

dmuelle (changes requested):
ID review

dmuelle (approved):
LGTM

jimblye:
#libby

hlhoots (approved):
LGTM

PR #26201, opened 2023-09-07
Test when the timer is scheduled to run and not when it actually runs.

anjumfatima90:
Certain background processes can slow things down. Additionally, there are situations where a virtual machine (VM) may not receive any CPU allocation for a brief period, potentially causing a 2 to 3-minute delay in timer execution. EJB timers schedule the next timeout based on the start time of the previous timeout, not its completion. Therefore, subsequent timers do not trigger exactly 1 second after the previous one finishes but instead 1 second after the previously scheduled start time (not the actual start time). If this results in a past timestamp, we execute the timer immediately and continue to do so until we catch up. Hence, checking the accuracy of when a timer runs is going to be very fragile.


anjumfatima90:
#build

tjwatson (approved):

anjumfatima90:
#libby

PR #26208, opened 2023-09-07
switch to resource accessor method approach

njr-11:
Switch to the approach defined by the Jakarta Data spec where you can define a repository interface method with no parameters and a return type of EntityManager, where the Jakarta Data provider implements the method to supply the EntityManager.


njr-11:
#build

KyleAure (approved):

Issue #23607, opened 2022-12-06
Enable verbose garbage collection by default on IBM Java/Semeru


kgibm:
## Description

By default, verbosegc is not enabled in Liberty (specifically, not enabled by default in Java). This is a problem if a performance or OutOfMemoryError issue occurs as the issue will often need to be reproduced with verbosegc, or users may simply overlook garbage collection performance issues (e.g. thread dumps may point to various application stacks but the underlying issue could be garbage collection). Verbosegc was enabled by default for new profiles in WAS traditional [9.0.0.3](https://www.ibm.com/support/pages/apar/PI69321) and [9.0.0.4](https://www.ibm.com/support/pages/apar/PI76705) (z/OS). This epic proposes to enable verbose garbage collection by default on IBM Java/Semeru. Initially discussed in design issue #23001.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Documents

When available, add links to required feature documents. Use "N/A" to mark particular documents which are not required by the feature.

- Aha: N/A
- UFO: https://ibm.box.com/s/o3isyhh62xixic925g8m7qto5ufgw4nk
- FTS: [Link to Feature Test Summary GH Issue](https://github.com/OpenLiberty/open-liberty/issues/27254)
- Beta Blog: Link to Beta Blog Post GH Issue
- GA Blog: [Link to GA Blog Post GH Issue](https://github.com/OpenLiberty/open-liberty/issues/27459)

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Process Overview

- [Prioritization](#prioritization)
- [Design](#design)
- [Implementation](#implementation)
- [Legal and Translation](#legal-and-translation)
- [Beta](#beta)
- [GA](#ga)
  - [Focal Point Approvals](#focal-point-approvals-complete-by-feature-complete-date)
- [Other Deliverables](#other-deliverables)

### General Instructions

The process steps occur roughly in the order as presented. Process steps occasionally overlap.

Each process step has a number of tasks which must be completed or must be marked as not applicable ("N/A").

Unless otherwise indicated, the tasks are the responsibility of the Feature Owner or a Delegate of the Feature Owner.

If you need assistance, reach out to the [OpenLiberty/release-architect](https://github.com/orgs/OpenLiberty/teams/release-architect).

**Important: Labels are used to trigger particular steps and must be added as indicated.** 

[...]

PR #26215, opened 2023-09-07
add verbose:gc as default

rsherget:
adds verbose:gc as default to server script

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -1118,10 +1118,10 @@ serverCmd()
   # Check for any verbose:gc variable set by user
   # By default we set this to be true unless user specifies otherwise
   # if [ ! jvmargs, javaoptions, jvmoptionsquoted contains verbose:gc, verbosegc, etc] then SERVER_IBM_JAVA_OPTIONS += -Xverbosegclog:verbosegc.%seq.log,10,1024
-  if ! [[ $JVM_OPTIONS_QUOTED == *"verbosegc"* || $JVM_OPTIONS_QUOTED == *"verbose:gc"* ]]
+  if ! [[ $JVM_OPTIONS_QUOTED == *"verbosegc"* || $JVM_OPTIONS_QUOTED == *"verbose:gc"* || ! -z "${VERBOSEGC}" ]]
   then
-    IBM_JAVA_OPTIONS+=" -Xverbosegclog:${LOG_DIR}/verbosegc.%seq.log,10,1024"
-    OPENJ9_JAVA_OPTIONS+=" -Xverbosegclog:${LOG_DIR}/verbosegc.%seq.log,10,1024"
+    IBM_JAVA_OPTIONS+=" -Xverbosegclog:${X_LOG_DIR}/verbosegc.%seq.log,10,1024"
+    OPENJ9_JAVA_OPTIONS+=" -Xverbosegclog:${X_LOG_DIR}/verbosegc.%seq.log,10,1024"
   fi
 
   if $os400lib; then
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -460,10 +460,6 @@ serverEnvDefaults()
     fi
   esac
 
-  # Check for any verbose:gc variable set by user
-  # By default we set this to be true unless user specifies otherwise
-  # if [ ! jvmargs, javaoptions, jvmoptionsquoted contains verbose:gc, verbosegc, etc] then SERVER_IBM_JAVA_OPTIONS += -Xverbosegclog:verbosegc.%seq.log,10,1024
-  SERVER_IBM_JAVA_OPTIONS+=" -Xverbosegclog:verbosegc.%seq.log,10,1024"
   # Add -Xquickstart -Xshareclasses:none for client JVMs only.  We don't want 
   # shared classes cache created for client operations.
   # Add -Dcom.ibm.tools.attach.enable=yes to allow self-attach on z/OS
@@ -1122,11 +1118,11 @@ serverCmd()
   # Check for any verbose:gc variable set by user
   # By default we set this to be true unless user specifies otherwise
   # if [ ! jvmargs, javaoptions, jvmoptionsquoted contains verbose:gc, verbosegc, etc] then SERVER_IBM_JAVA_OPTIONS += -Xverbosegclog:verbosegc.%seq.log,10,1024
-  if [ $IBM_JAVA_OPTIONS == *"verbosegc"* ] or [$IBM_JAVA_OPTIONS == *"verbose:gc"* ] then
-    IBM_JAVA_OPTIONS+=" -Xverbosegclog:verbosegc.%seq.log,10,1024"
-    OPENJ9_JAVA_OPTIONS+=" -Xverbosegclog:verbosegc.%seq.log,10,1024"
+  if ! [[ $JVM_OPTIONS_QUOTED = *"verbosegc"* ] || [ $JVM_OPTIONS_QUOTED = *"verbose:gc"* ]]
+  then
+    IBM_JAVA_OPTIONS+=" -Xverbosegclog:logs/verbosegc.%seq.log,10,1024"
+    OPENJ9_JAVA_OPTIONS+=" -Xverbosegclog:logs/verbosegc.%seq.log,10,1024"
   fi
-
 
   if $os400lib; then
     serverCmdOS400 "${SERVER_CMD_BACKGROUND_LOG}" "${SERVER_NAME}" "$@"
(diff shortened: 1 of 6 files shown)


rsherget:
#23607

rsherget:
It looks like the jvm.options are processed after the server default variables are made, so theres no way to check the jvm.options while creating the server defaults. Most likely need to check on server command (serverCMD for server. startServer, runServer, and debugServer for server.bat). something like if our JVM_OPTIONS variable does not contain verbosegc then add our verbosegc command

rsherget:
search for verbose:gc in batch file: https://stackoverflow.com/questions/7005951/batch-file-find-if-substring-is-in-string-not-in-a-file

search for verbose:gc in bash file: https://stackoverflow.com/questions/4467735/how-to-find-substring-inside-a-string-or-how-to-grep-a-variable

rsherget:
the start, run, debug, etc commands in the server script are found under 'case "$ACTION" in' line 1547

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1221,6 +1221,15 @@ serverCmd()
   SAVE_OPENJ9_JAVA_OPTIONS=${OPENJ9_JAVA_OPTIONS}
   OPENJ9_JAVA_OPTIONS=${SERVER_IBM_JAVA_OPTIONS}
 
+  # Check for any verbose:gc variable set by user
+  # By default we set this to be true unless user specifies otherwise
+  # if [ ! jvmargs, javaoptions, jvmoptionsquoted contains verbose:gc, verbosegc, etc] then SERVER_IBM_JAVA_OPTIONS += -Xverbosegclog:verbosegc.%seq.log,10,1024
+  if [[ $JVM_OPTIONS_QUOTED != *"gc"* ]]
```
kgibm: The problem with only checking for `gc` is that other options have `gc` in them: https://eclipse.dev/openj9/docs/xjit/. I suggest checking for both `verbosegc` and `verbose:gc`.

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1221,6 +1221,15 @@ serverCmd()
...
+  then
+    IBM_JAVA_OPTIONS+=" -Xverbosegclog:logs/verbosegc.%seq.log,10,1024"
```
kgibm: One potential issue I can see is that this configures it to write to the logs subdirectory of the current working directory. In principle, I guess it's possible a customer has configured the logs directory to be somewhere else, so you might want to double check what happens in that case and whether the JVM will create the logs subdirectory if it doesn't exist.

dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/VerboseLogTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        delJvmOptions();
+
+        String[] parms = new String[2];
+        parms[0] = "start";
+        parms[1] = SERVER_NAME;
+
+        Writer isw = new OutputStreamWriter(new FileOutputStream(jvmoptionsserverroot), "UTF-8");
+        BufferedWriter bw = new BufferedWriter(isw);
+        bw.write("-Dnogclog\n");
```
kgibm: What's `-Dnogclog`? I think the idea to disable it was with the envar `VERBOSEGC=false`

dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/VerboseLogTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        delJvmOptions();
+
+        String[] parms = new String[2];
+        parms[0] = "start";
+        parms[1] = SERVER_NAME;
+
+        Writer isw = new OutputStreamWriter(new FileOutputStream(jvmoptionsserverroot), "UTF-8");
+        BufferedWriter bw = new BufferedWriter(isw);
+        bw.write("-Dverbosegc=false\n");
```
kgibm, line 174: `-D` is a JVM system property. I think you'll need to write `VERBOSEGC=false` to `server.env` instead of `jvm.options` and then also check for this in the `server` scripts.

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1221,6 +1221,15 @@ serverCmd()
   SAVE_OPENJ9_JAVA_OPTIONS=${OPENJ9_JAVA_OPTIONS}
   OPENJ9_JAVA_OPTIONS=${SERVER_IBM_JAVA_OPTIONS}
 
+  # Check for any verbose:gc variable set by user
+  # By default we set this to be true unless user specifies otherwise
+  # if [ ! jvmargs, javaoptions, jvmoptionsquoted contains verbose:gc, verbosegc, etc] then SERVER_IBM_JAVA_OPTIONS += -Xverbosegclog:verbosegc.%seq.log,10,1024
+  if ! [[ $JVM_OPTIONS_QUOTED == *"verbosegc"* || $JVM_OPTIONS_QUOTED == *"verbose:gc"* || ! -z "${VERBOSEGC}" ]]
```
kgibm: Though unlikely, someone might specify `VERBOSEGC=true`, so it's better to explicitly check for `false` here instead of just the existence of the envar.

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -1221,6 +1221,15 @@ serverCmd()
   SAVE_OPENJ9_JAVA_OPTIONS=${OPENJ9_JAVA_OPTIONS}
   OPENJ9_JAVA_OPTIONS=${SERVER_IBM_JAVA_OPTIONS}
 
+  # Check for any verbose:gc variable set by user
+  # By default we set this to be true unless user specifies otherwise
+  # if [ ! jvmargs, javaoptions, jvmoptionsquoted contains verbose:gc, verbosegc, etc] then SERVER_IBM_JAVA_OPTIONS += -Xverbosegclog:verbosegc.%seq.log,10,1024
+  if ! [[ $JVM_OPTIONS_QUOTED == *"verbosegc"* || $JVM_OPTIONS_QUOTED == *"verbose:gc"* || "${VERBOSEGC}" == "false" ]]
```
kgibm: Another thing to double check is that if the customer specifies `-Xverbosegclog` in their jvm.options, that their options still take precedence. Alternatively, you could also check for `-Xverbosegclog` and skip adding it if it's already there... which might be cleaner anyway.

rsherget: JVM_OPTIONS_QUOTED contains all the jvm.options specified by the user so `JVM_OPTIONS_QUOTED == *"verbosegc"*` catches that, though it would be cleaner to check specifically for `Xverbosegclog` if theres not another way to activate it that contains `verbosegc`. The precedence is checked in the FAT tests, specifically the testJvmChangeVerbose test.

kgibm: Ahh, yes, you're right, apologies.

rsherget:
#build

rsherget:
#build

rsherget:
#build

rsherget:
#build

rsherget:
#build

rsherget:
#build

rsherget:
#build

rsherget:
#build

rsherget:
#build #using.java21=false

rsherget:
#libby

tbitonti (changes requested):
Some problems with the try/catch in 'before': No need to catch the exception if the file is expected.  Use try-with-resources form for the input streams.  Possible NPE because of a null JVM_VARIANT property value.

cbridgha (dismissed):
My biggest ask for these changes is making sure we are testing across many platforms - We used to have a "mini SOE" option....    Maybe asking for specific OS platforms

rsherget:
#build

rsherget:
Build testing com.ibm.ws.kernel.boot_fat with multiple OS platforms started https://libh-proxy1.fyre.ibm.com/cognitive-dev/pipelineAnalysis.html?uuid=921fcb8c-74f6-40ff-9cce-f66dc08e7aa4&tab=Steps

dev/com.ibm.ws.kernel.boot_fat/fat/src/com/ibm/ws/kernel/boot/VerboseLogTest.java
kgibm: @rsherget Just double checking if this is handled?

rsherget:
#build

rsherget:
Build testing com.ibm.ws.kernel.boot_fat with multiple OS platforms started part 2: https://libh-proxy1.fyre.ibm.com/cognitive-dev/pipelineAnalysis.html?uuid=08fcb2ff-f8a7-49bb-9505-084d3fd306cc

rsherget:
#libby

rsherget:
#build

tbitonti (approved):

tbitonti (approved):

fritze2:
#rosie-build

PR #26224, opened 2023-09-08
switch to using validation as defined by spec

njr-11:
Update to using Jakarta Validation in Jakarta Data as was recently defined in the Jakarta Data spec.  A couple of tests need to be disabled for now.  These will require a feature from Jakarta Validation to work.


jmstephensgit (approved):

njr-11:
#build

PR #26240, opened 2023-09-11
experiment with insert method

njr-11:
Try out an insert method that invoke entityManager.persist and raises an error if the entity already exists.


njr-11:
#build

KyleAure (approved):

PR #26247, opened 2023-09-12
Continue testing AWS advanced jdbc wrapper

KyleAure:
Ensure wrapper works correctly when used during ServletContextListener initialize method.


KyleAure:
This test doesn't add anymore value than the existing tests we have.  I don't think there is any reason to merge this, closing.

PR #26265, opened 2023-09-13
Fix Testcontainer unit testing bug

KyleAure:
Ensure the `~/.testcontainer.properties` file is deleted before unit testing to avoid test failures when the properties file is outdated.  For example, if the docker.cert.path points to a directory that no longer exists.


KyleAure:
#build

mswatosh (approved):

KyleAure:
Personal build failed with a single known defect.

Issue #25640, opened 2023-07-06
WithSpanInterceptor doesn't call instrumentation.end()


Azquelt:
`WithSpanInterceptor` will not call `instrumentation.end()` if `context.proceed()` throws an exception.

I think in practise this will mean that methods annotated with `@WithSpan` will not log a span if the method throws an exception. We might also leak something.

- [ ] Fix `WithSpanInterceptor` so that it always calls `end()` (and passes in the exception if one was thrown from the user's code)
- [ ] Add a test for a `@WithSpan` method which throws an exception.
- [ ] Work out the user-visible symptoms of this problem and re-write this issue using the [bug report template](https://github.com/OpenLiberty/open-liberty/blob/integration/.github/ISSUE_TEMPLATE/bug_report.md)

Azquelt:
Change to fix this was reverted

PR #26272, opened 2023-09-14
Telemetry with span

Yas-Franco:
For issue #25640


dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/WithSpanErrorServlet.java
```diff
@@ -0,0 +1,63 @@
...
+public class WithSpanErrorServlet extends FATServlet {
+
+    @Inject
+    private InMemorySpanExporter exporter;
+
+    @Inject
+    private SpanBean spanBean;
+
+    @Test
+    @ExpectedFFDC("java.lang.RuntimeException")
```
Azquelt: This isn't a scenario where we want to get an FFDC out. We should ignore the FFDC in the interceptor.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/WithSpanErrorServlet.java
```diff
@@ -0,0 +1,63 @@
...
+    public void callMethodException() {
+        try {
+            spanBean.methodException();
+
+        } catch (RuntimeException e) {
```
Azquelt: Add a `fail` here in case an exception isn't thrown.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/WithSpanErrorServlet.java
```diff
@@ -0,0 +1,63 @@
...
+            //Span should still be created with the error status present
+            SpanData spanData = exporter.getFinishedSpanItems(1).get(0);
+            assertThat(spanData, hasStatus(StatusCode.ERROR));
```
Azquelt: Add some additional checks here:

- Check the annotations which identify the method which was called
- Check the kind (should be `INTERNAL`)
- I think the actual exception should end up in the span as well. Use `SpanDataMatcher.withExceptionLog()`

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/telemetry/WithSpanErrorServlet.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    @ApplicationScoped
+    public static class SpanBean {
+
+        //Creates a span for this method that throws an exception
+        @WithSpan
+        public String methodException() {
+            throw new RuntimeException("runtime exception");
```
Azquelt: Create some kind of test exception which extends `RuntimeException` and check for that to ensure it's actually the expected exception being thrown and not some other kind of `RuntimeException`.

Yas-Franco:
#build

Yas-Franco:
#libby

Azquelt (approved):

Azquelt:
#libby
#build

PR #26278, opened 2023-09-14
HTTP Tracing

### LINKED ISSUE #26257: HTTP tracing development

fmhwong:
Fixes https://github.com/OpenLiberty/open-liberty/issues/26257

HTTP tracing is currently under beta fencing.


fmhwong:
#build

fmhwong:
#build

dev/io.openliberty.io.opentelemetry.1.29/bnd.bnd
```diff
@@ -51,7 +51,7 @@ Export-Package: \
   io.opentelemetry.exporter.sender.okhttp.internal;version=${openTelemetryVersion};thread-context=true,\
   io.opentelemetry.exporter.zipkin.internal;version=${openTelemetryVersion};thread-context=true,\
   io.opentelemetry.sdk.autoconfigure.internal;version=${openTelemetryVersion};thread-context=true,\
-  io.opentelemetry.*;version=${openTelemetryVersion}
+  io.opentelemetry.*;version=${openTelemetryVersion};thread-context=true
```
benjamin-confino, line 54: Felix confirms these are needed.

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryServletFilter.java
```diff
@@ -0,0 +1,311 @@
...
+        public String get(final ServletRequest request, final String key) {
+            if (request == null) {
+                return null;
+            }
+
+            if (request instanceof HttpServletRequest) {
+                HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+                return httpServletRequest.getHeader(key);
+            }
+            return null;
```
benjamin-confino, line 202: Wouldn't say no to some logging or guards, but I'm getting this is known to be safe by now.

benjamin-confino (approved):

fmhwong:
#libby

dev/io.openliberty.io.opentelemetry.1.29/bnd.bnd
Azquelt: There has been no adequate explanation of why this is needed. It shouldn't be needed and should be removed ASAP.

This should not have been merged with this in it. At the very least you can take out all of the lines above it because they only exist to put specific packages on the thread context classloader.

Azquelt: If there are _specific_ packages which need to be added to the TCCL, that's fine, but adding everything with no explanation is not.

Azquelt: Raised #26313 to do this.

Azquelt: It's also not clear why this wasn't needed before but is now.

If it was needed before, then this would have been a release bug and there needs to be an issue raised for that.

Azquelt: Although, in terms of raising a release bug, I'm more concerned with the change to the 1.0 project.

PR #26295, opened 2023-09-15
Fix concurrency tck signature testing on non-LTS java

KyleAure:
The concurrency TCK can only run Signature tests on LTS Java versions since minor differences in java byte code can affect the signatures produced.

### DIFF
--- a/dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
+++ b/dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
@@ -46,7 +46,7 @@ public class FATSuite {
                 return result = false;
             }
 
-            if (System.getProperty("os.name").contains("Windows")) {
+            if (System.getProperty("os.name", "unknown").toLowerCase().contains("windows")) {
                 reason = "signature test plugin not supported on Windows.";
                 return result = false;
             }
--- a/dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
+++ b/dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
@@ -16,7 +16,12 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 import org.junit.runners.Suite.SuiteClasses;
 
+import com.ibm.websphere.simplicity.log.Log;
+
 import componenttest.custom.junit.runner.AlwaysPassesTest;
+import componenttest.custom.junit.runner.Mode;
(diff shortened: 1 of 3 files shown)


KyleAure:
#build

dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
```diff
@@ -25,4 +30,39 @@
...
+                reason = testClass.getCanonicalName() + " is not run in full mode.";
+                return result = false;
+            }
+
+            if (testClass.getAnnotation(Mode.class).value() != TestMode.FULL) {
+                reason = testClass.getCanonicalName() + " is not run in full mode.";
+                return result = false;
+            }
+
+            if (System.getProperty("os.name").contains("Windows")) {
```
tkburroughs: Other places I'm familiar with perform this check as follows:
System.getProperty("os.name", "unknown").toLowerCase().contains("windows")

Not sure how reliable check is with case sensitivity?  Or if "os.name" might return null on some platforms?

dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
```diff
@@ -25,4 +30,39 @@
...
+                reason = "signature test plugin not supported on Windows.";
+                return result = false;
+            }
+
+            if (JavaInfo.JAVA_VERSION != 17 || JavaInfo.JAVA_VERSION != 21) {
```
tkburroughs, line 54: Would it be useful to add "&& JavaInfo.JAVA_VERSION < 26".  The current code would just silently never run the tests after Java 21.  Perhaps that is ok, or we could force someone to reconsider after say Java 25?  Would we never plan to certify Jakarta EE 11 on any LTS after 21?

tkburroughs (dismissed):
Approving, though I left a couple of comments to consider.

dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
KyleAure: Good suggestion, but the Concurrency TCK has generated signature files for specifically Java 17 & Java 21 which is why this logic is so strict.

dev/io.openliberty.jakarta.concurrency.3.1_fat_tck/fat/src/io/openliberty/jakarta/enterprise/concurrent/tck/FATSuite.java
KyleAure: Thanks, better safe than sorry I'll use the more strict check.

KyleAure:
#libby - re-check headers
#build - re-build with changes

tkburroughs (approved):

Issue #25526, opened 2023-06-20
Back Channel Logout - logout endpoint cleans up wrong client


c00crane:
If I login using a client from provider: bcl_mainPath_confClient, and then invoke https://localhost:8920/oidc/endpoint/OidcConfigSample_http_httpsRequired_true/logout
(notice it’s a different provider), BUT, I used the webclient from my original request.
I get a message about a successful logout and some of the cookies get cleaned up.  The access_token is still valid, but, some things are getting cleaned up when the logout should just fail.

[wrongProvider.zip](https://github.com/OpenLiberty/open-liberty/files/11805206/wrongProvider.zip)

ayoho:
I believe this is caused by my changes in https://github.com/OpenLiberty/open-liberty/pull/24811 when I was trying to fix the issuer checking. Since the OP doesn’t have an `issuerIdentifier` attribute configured, the issuer check passes. That’s why the logout proceeds even though this provider wasn’t the one that issued the token.

I’d probably put the logic back in that I took out to base the issuer check off the request URL if `issuerIdentifier` isn't configured, but make the check smarter to allow for either "providers" or "endpoint" in the URL.

jimmy1wu:
fixed in https://github.com/OpenLiberty/open-liberty/pull/26296

@c00crane please verify when you can.

PR #24811, opened 2023-03-24
PR for Issue 23588: Update how issuer claim is decided for OIDC back-channel logout

ayoho:
For #23588

PR #26296, opened 2023-09-15
add expected iss fallback and endpoint vs providers check

jimmy1wu:
for #25526

- puts back the code for the expected issuer to fallback to getting it from the request url (https://github.com/OpenLiberty/open-liberty/pull/24811)
- added code to check if the expected issuer from the request url doesn't match the iss in the id token, then see if switching `/oidc/endpoint/` to `/oidc/providers/` and vice-versa would match. if still not, then throw an exception.

**note:** i _did not_ undo the part in https://github.com/OpenLiberty/open-liberty/pull/24811 that sets the logout token `iss` as the id token `iss`. the reason for keeping this change is because if the id token was issued by `/oidc/endpoint` and the logout happened on `/oidc/providers`, the iss check on the OP will now pass due to the new code, but `/oidc/endpoint` should be used as the issuer in the logout token, otherwise the iss check will fail on the RP.

### DIFF
--- a/dev/com.ibm.ws.security.openidconnect.server/src/com/ibm/ws/security/openidconnect/web/OidcRpInitiatedLogout.java
+++ b/dev/com.ibm.ws.security.openidconnect.server/src/com/ibm/ws/security/openidconnect/web/OidcRpInitiatedLogout.java
@@ -222,7 +222,7 @@ public class OidcRpInitiatedLogout {
     }
 
     void sendBackchannelLogoutRequests(String userName, String idTokenString) {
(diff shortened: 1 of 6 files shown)


jimmy1wu:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.fat.common.backchannelLogout,com.ibm.ws.security.oidc.client_fat.backchannelLogout,com.ibm.ws.security.oidc.server_fat.backchannelLogout,com.ibm.ws.security.social_fat.LibertyOP.backchannelLogout

ayoho (approved):

PR #26297, opened 2023-09-15
If criu fails to restore the process display the restore.log

### LINKED ISSUE #26193: Improve error message when restore fails with recovery disabled
Restore checkpoint app image on OCP 4.13 failed with the following error in pod log
```
% oc get pods
NAME                           READY   STATUS             RESTARTS        AGE
ebuy-inston-55c874b886-wp8tz   0/1     CrashLoopBackOff   6 (4m32s ago)   10m

% oc logs ebuy-inston-55c874b886-wp8tz
Found mounted TLS certificates, generating keystore

CWWKE0964E: Restoring the checkpoint server process failed. Check the /logs/checkpoint/restore.log log to determine why the checkpoint process was not restored. The server did not launch because checkpoint restore recovery is disabled.
```
The checkpoint app image has recovery disabled (_ENV CRIU_RESTORE_DISABLE_RECOVERY=true in Containerfile_) so the pod exit and user can not view file /logs/checkpoint/restore.log

Need improvement to display _restore.log_ content when restore fails if it exists

If we want to see restore.log in this case, enable recovery

```
  env:
    - name: CRIU_RESTORE_DISABLE_RECOVERY
      value: false
```

tjwatson:
Fixes #26193

This fix always displays the `restore.log` if a failure to restore happened at the criu layer.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -1532,7 +1532,11 @@ restoreServer()
       issueMessage --message:error.checkpoint.restore.failed.no.recovery "${X_LOG_DIR}/checkpoint/restore.log"
     else
       issueMessage --message:info.checkpoint.restore.failed "${X_LOG_DIR}/checkpoint/restore.log"
-    fi    
+    fi
+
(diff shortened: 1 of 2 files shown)


tjwatson:
#build

sebratton (approved):
LGTM

PR #26300, opened 2023-09-15
We're unsure why this wasn passing before, after investigating the op…

benjamin-confino:
…en telemetry source code telemty 1.29 should be appending GET here

These tests have passed both locally and in personal builds, but when making unrelated changes they started failing for me. After analysing the source code I am confident that GET should be present. 

See https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.29.0/instrumentation-api-semconv/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanNameExtractor.java#L36

That method is the GET, which should be coming out in the span name we mark a regex against.


benjamin-confino:
#build

benjamin-confino:
#build

fmhwong (approved):
LGTM

benjamin-confino:
On Further investigation I think InMemorySpanExporter.shutdown() was not being called at all. This is very concerning for two reasons 

- It suggests there's a bug in integration.
- We didn't have any tests to catch that bug.

benjamin-confino:
I'll probably deliver this with other changes now I know why the behaviour changed.

benjamin-confino:
This was fixed in a different PR.

PR #26177, opened 2023-09-05
26162 merge common code

benjamin-confino:
fixes #26162
fixes #26158

PR #26308, opened 2023-09-16
26155 rework acquistion open telemetry info followup

benjamin-confino:
This implements the first batch of comments on #26177


benjamin-confino:
#build

benjamin-confino:
#build

Azquelt:
Are these follow-up changes for #26157, rather than #26177 ?

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/interfaces/OpenTelemetryAccessor.java
```diff
@@ -14,31 +14,60 @@
...
+    /**
+     * Gets or creates the instance of OpenTelemetry associated with this application and returns it wrapped inside an OpenTelemetryInfo.
+     *
+     * @return An instance of OpenTelemetryInfo containing the instance of OpenTelemetry associated with this application. This instance will be a no-op OpenTelemetry if telemetry
+     *         is disabled or the application has shut down.
+     */
     public static OpenTelemetryInfo getOpenTelemetryInfo() {
-        return OpenTelemetryInfoFactory.getOpenTelemetryInfo(ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData().getJ2EEName());
+        try {
+            OpenTelemetryInfoFactory factory = OSGIHelpers.getService(OpenTelemetryInfoFactory.class, OpenTelemetryProducer.class);
```
Azquelt:
```suggestion
            OpenTelemetryInfoFactory factory = OSGIHelpers.getService(OpenTelemetryInfoFactory.class, OpenTelemetryAccessor.class);
```
Javadoc on `getService` says the second argument should be the calling class.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/cdi/OpenTelemetryProducer.java
```diff
@@ -31,39 +33,65 @@ public class OpenTelemetryProducer {
...
 
-    J2EEName j2EEName = null;
-
-    private J2EEName getJ2EEName() {
-        if (j2EEName == null) {
-            j2EEName = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData().getJ2EEName();
-        }
-        return j2EEName;
-    }
+    ApplicationMetaData metaData = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData().getModuleMetaData().getApplicationMetaData();
```
Azquelt: 1. Field should be private
2. Because this class is a normal scoped bean, we should avoid doing things in the no-op constructor or in field initializers since that code will be called when the proxy is created.
   
   It would be better to initialize this field in a `@PostConstruct` method.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+public class OpenTelemetryInfoImpl implements OpenTelemetryInfo {
+
+    private static final TraceComponent tc = Tr.register(OpenTelemetryInfoImpl.class);
+    private static final String INSTRUMENTATION_NAME = "io.openliberty.microprofile.telemetry";
+
+    private final boolean enabled;
+    private final OpenTelemetry openTelemetry;
+    private final Tracer tracer;
+
+    private final String APP_NAME;
```
Azquelt: This field is not a constant.
```suggestion
    private final String appName;
```

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * No-args constructor for CDI
+     */
+    public OpenTelemetryInfoImpl() {
+        this.enabled = false;
+        this.openTelemetry = null;
+        tracer = null;
+        this.APP_NAME = "unkown";
+    }
```
Azquelt: Do we still need this?

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param enabled
+     * @param openTelemetry
+     */
+    public OpenTelemetryInfoImpl(boolean enabled, OpenTelemetry openTelemetry, String appName) {
+        super();
+        this.enabled = enabled;
+        this.openTelemetry = openTelemetry;
+        tracer = openTelemetry.getTracer(INSTRUMENTATION_NAME);
+        this.APP_NAME = appName;
+    }
```
Azquelt: When you look at the calling code, it's not immediately clear what the first argument means, or that it has to match up with the second argument.

Given that we already need to know whether we have a real `OpenTelemetry` or a dummy one to pass the right value for `enabled`, I'd suggest making this constructor private and having two static constructor methods instead:

```public static OpenTelemetryInfoImpl ofNoop(String appName)```

```public static OpenTelemetryInfoImpl of(OpenTelemetry oTel, String appName)```

File: dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/DisposedOpenTelemetryInfo.java [hunk 6]
```diff
@@ -0,0 +1,74 @@
...
+import io.opentelemetry.api.trace.Tracer;
+
+/**
+ *
+ */
+public class DisposedOpenTelemetryInfo implements OpenTelemetryInfo {
+
+    private static final TraceComponent tc = Tr.register(DisposedOpenTelemetryInfo.class);
+
+    private String appName = "Unkown";
```
Azquelt: "Unknown"

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/DisposedOpenTelemetryInfo.java
```diff
@@ -0,0 +1,74 @@
...
+    private void logWarning() {
+        Tr.warning(tc, Tr.formatMessage(tc, "CWMOT5003.factory.used.after.shutdown", appName));
+        if (TraceComponent.isAnyTracingEnabled() && tc.isEventEnabled()) {
+            Exception e = new Exception();
+            ByteArrayOutputStream stackStream = new ByteArrayOutputStream();
+            PrintStream stackPrintStream = new PrintStream(stackStream);
+            e.printStackTrace(stackPrintStream);
+
+            Tr.event(tc, "OpenTelemetryInfoFactory", "The stack that led to OpenTelemetryInfoFactory being called after " + appName + " has shutdown is:.");
+            Tr.event(tc, stackStream.toString());
```
Azquelt: Maybe pass `stackStream.toString()` as an argument to the first `event` call instead of a separate call.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,329 @@
...
+            //Run io.openliberty.microprofile41.internal_fat to reproduce examples of liberty internal apps reaching this
+            //for "applications" that never went through application started.
+            //
+            String j2EEName = metaData.getJ2EEName().toString();
+            if (j2EEName.startsWith("io.openliberty") || j2EEName.startsWith("com.ibm.ws")) {
+                return new OpenTelemetryInfoImpl(false, OpenTelemetry.noop(), "Liberty internals");
+            }
+
+            OpenTelemetryInfoReference atomicRef = (OpenTelemetryInfoReference) metaData.getMetaData(slotForOpenTelemetryInfoHolder);
+            OpenTelemetryInfoWrappedSupplier supplier = atomicRef.get();
```
Azquelt: As discussed, do the "is app internal check" after checking if `atomicRef` is null. Log a sensible exception if that occurs for a user app.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,329 @@
...
+                return openTelemetryInfo;
+            }
+        }
+
+        public boolean lock() {
+            synchronized (this) {
+                supplier = OpenTelemetryInfoFactoryImpl::createDisposedOpenTelemetryInfo;
+                return hasRun;
+            }
+        }
```
Azquelt: I think this would be simpler if this class handled disposal as well.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactoryImpl.java
benjamin-confino: Check

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/DisposedOpenTelemetryInfo.java
benjamin-confino: check

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/DisposedOpenTelemetryInfo.java
benjamin-confino: check

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoFactoryImpl.java
benjamin-confino: Check

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoImpl.java
benjamin-confino: Removed, that field is only actually used in the error message for Disposed.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoImpl.java
benjamin-confino: What about an EnabledOpenTelemetryInfo and DisabledOpenTelemetryInfo? 

I'm thinking that could make things in trace.log more immediately apparent, and we already have the interface.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/cdi/OpenTelemetryProducer.java
benjamin-confino: Check

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/interfaces/OpenTelemetryAccessor.java
benjamin-confino: Check

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -65,26 +61,40 @@ public static void setUp() throws Exception {
...
     }
 
     //This test does not use open telemetry directly, but it does trigger io.openliberty.microprofile.telemetry.internal.rest.TelemetryClientFilter.java for an app with no CDI and ensures it does not throw an exception
+
+    //This test also ensures that we actually call the started and shutdown methods in OpenTelemetryInfoFactoryImpl
+    //I'm putting it in here to avoid creating an extra server that enables trace. And in an existing method because
+    //Our version of junit doesn't allow fixed test ordering.
     @Test
-    public void testClientWithNoCDi() throws Exception {
+    public void testClientWithNoCDiAndFactoryStarupShutdownCalled() throws Exception {
```
Azquelt:
```suggestion
    public void testClientWithNoCDiAndFactoryStartupShutdownCalled() throws Exception {
```

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -65,26 +61,40 @@ public static void setUp() throws Exception {
...
+                      server.waitForStringInLog("OpenTelemetryInfoImpl > dispose Entry", server.getDefaultTraceFile()));
     }
 
     @AfterClass
     public static void tearDown() throws Exception {
-        server.stopServer();
+        if (server.isStarted()) {
+            server.stopServer("CWWKZ0059E", "CWWKZ0014W"); //CWWKZ0059E is complaining that we're deleting an app while its still configured. This is needed to test the shutdown code works properly. (Actually stopping the server archives the trace log). CWWKZ0014W is saying the application cannot be found because we just deleted it.
+        }
     }
```
Azquelt, line 101: Can avoid these errors by removing the app from server.xml instead of deleting the file.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/common/spanexporter/InMemorySpanExporter.java
```diff
@@ -227,8 +227,19 @@ public CompletableResultCode export(Collection<SpanData> spans) {
...
+             * Match 0 or 1 of the string "GET " (note whitespace) using a non-capturing group.
+             * Match a literal "/"
+             * Match anything
+             * Match a literal /
+             * match the string "test"
+             * Match anything until the end of the line.
+             */
+            if (span.getName().contains("readspans") || span.getName().matches("^(?:GET )?/.*/test.*$")) {
                 iter.remove();
             }
```
Azquelt, line 244: The commit message says you don't know why this was passing before. Have you checked for a build break defect, it may be able to be closed now.

Please also mention this in the PR description.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/DisposedOpenTelemetryInfo.java
```diff
@@ -0,0 +1,72 @@
...
+        logWarning();
+        return OpenTelemetry.noop();
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public Tracer getTracer() {
+        logWarning();
+        return OpenTelemetry.noop().getTracerProvider().get("");
+    }
```
Azquelt, line 54: These methods could potentially spam out warnings which doesn't seem particularly helpful. I'm still thinking more about how I think this should work.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,331 @@
...
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdkBuilder;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.resources.ResourceBuilder;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+
+// We want this to start before CDI so it clears the list of stopped apps before CDI gets to them.
+@Component(service = { ApplicationStateListener.class, OpenTelemetryInfoFactory.class }, property = { "service.vendor=IBM", "service.ranking:Integer=150" })
```
Azquelt: We no longer have a list of stopped apps, so this comment doesn't seem correct any more.

We may still want it to happen before CDI so that we know the producer won't call through to `getOpenTelemetry` before we've set up the slot in the app metadata?

Not sure if that's a concern. _Usually_ beans shouldn't be instantiated during app start, but I think it's possible for later extension lifecycle phases to look up beans.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,331 @@
...
+        public boolean closeAndDisposeIfCreated() {
+            synchronized (this) {
+                supplier = OpenTelemetryInfoFactoryImpl::createDisposedOpenTelemetryInfo;
+                if (openTelemetryInfo != null) {
+                    consumer.accept(openTelemetryInfo);
+                    return true;
+                }
+                return false;
+            }
+        }
```
Azquelt: I don't think this needs to return a boolean any more. If it does, the meaning of the return value should be documented.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * No-args constructor for CDI
+     */
+    public OpenTelemetryInfoImpl() {
+        this.enabled = false;
+        this.openTelemetry = null;
+        tracer = null;
+    }
```
Azquelt: I think we no longer need the no-arg constructor since this class isn't used directly as a CDI bean any more (the bean type is now the interface).

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private final boolean enabled;
+    private final OpenTelemetry openTelemetry;
+    private final Tracer tracer;
+
+    /**
+     * @param enabled
+     * @param openTelemetry
+     */
+    public OpenTelemetryInfoImpl(boolean enabled, OpenTelemetry openTelemetry, String appName) {
```
Azquelt: Doc needs updating.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,331 @@
...
+
+    /*
+     * There are two race conditions we need to protect against.
+     *
+     * On app startup/shutdown we need to swap the holder associated with the application.
+     * We protect against race conditions here by using an AtomicReference.
+     *
+     * Within the context of an application's lifecycle we need to ensure OpenTelemetryInfo
+     * is only created once. OpenTelemetryInfoReference handles this.
+     */
```
Azquelt, line 288: Looking at this again, it's not clear to me why we need to swap the holder associated with the application. 

We've ended up now with this chain of objects:
App Metadata -> OpenTelemetryInfoReference -> OpenTelemetryInfoWrappedSupplier -> OpenTelemetryInfo -> OpenTelemetry

Of these, I can explain the jobs of two:
- `OpenTelemetryInfoWrappedSupplier`: responsible for lazily initializing the `OpenTelemetryInfo` and disposing of it when the app shuts down.
- `OpenTelemetryInfo`: stores the `OpenTelemetry` and `Tracer` objects for when they need to be looked up. Also stores whether the `OpenTelemetry` is a real implementation or a no-op so that some integrations can be disabled if OTel is disabled.

I'm not sure why `OpenTelemetryInfoReference` is needed. At the moment we only update this when the app shuts down so that it will hold a different `OpenTelemetryInfoWrappedSupplier` which will create and return a `DisposedOpenTelemetryInfo` if asked.

I'm not sure why we couldn't instead update the `OpenTelemetryInfo` instance that `OpenTelemetryInfoWrappedSupplier` is holding when it disposes the old one.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,331 @@
...
+    private class OpenTelemetryInfoReference extends AtomicReference<OpenTelemetryInfoWrappedSupplier> {
+
+        private static final long serialVersionUID = -4884222080590544495L;
+    }
+
+    private class OpenTelemetryInfoWrappedSupplier {
+        private final Consumer<OpenTelemetryInfo> consumer;
```
Azquelt: rename to `disposer` - the name of a `Consumer` or `Supplier` should say what it's for, not just what it is.

In this case `supplier` is ok (it supplies us with an `OpenTelemetryInfo` instance), but `consumer` is not.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,331 @@
...
+        private Supplier<OpenTelemetryInfo> supplier;
+        private volatile OpenTelemetryInfo openTelemetryInfo = null;
+
+        public OpenTelemetryInfoWrappedSupplier(Consumer<OpenTelemetryInfo> consumer, Supplier<OpenTelemetryInfo> supplier) {
```
Azquelt: Swap the argument order please, it makes more sense for the creation function to be first and the disposal function to be second.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/OpenTelemetryInfoImpl.java
benjamin-confino: Check

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/common/spanexporter/InMemorySpanExporter.java
benjamin-confino: The commit message is out of date (I'll change it), it was the ASL problem I mentioned this morning.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
Azquelt: I guess you could say that the job of the `OpenTelemetryInfoWrappedSupplier` is just to handle lazily initializing and destroying the `OpenTelemetryInfo` - and then we need something else if we want to actually change it for a different object?

It does feel a bit like we're doing half the job in one place and half in another which makes it tricky to see what's going on.

Will think about it more tomorrow.

benjamin-confino: It would work if we updated the instance within OpenTelemetryInfoWrappedSupplier. 

However I think its a little neater this way. It gives OpenTelemetryInfoWrappedSupplier a clear lifecycle and takes fewer LOC.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
benjamin-confino: Alphabetical felt right, but I'm not going to argue with the man who has the power to reject a PR.

benjamin-confino:
#build

benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
Azquelt: ok

benjamin-confino:
https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemOid/com.ibm.team.build.BuildResult/_1ISFMFZuEe63MKOTK2E49A was green, but this will likely need merging after I deliver my other PR

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -36,18 +34,24 @@
...
     public static final String SERVER_NAME = "Telemetry10JaxWithLogging";
     public static final String NO_CDI_APP_NAME = "clientNoCDI";
 
+    @ClassRule
+    public static RepeatTests r = MicroProfileActions.repeat(SERVER_NAME, MicroProfileActions.MP60, MicroProfileActions.MP61)
+                    .andWith(FeatureReplacementAction.BETA_OPTION().fullFATOnly());
+
     @TestServlets({
                     @TestServlet(contextRoot = NO_CDI_APP_NAME, servlet = ClientTriggeringServlet.class),
     })
```
Azquelt: As a follow-up: this should be removed - `ClientTriggeringServlet` is not a test servlet.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -65,26 +69,40 @@ public static void setUp() throws Exception {
...
+
+        assertNotNull("OpenTelemetryInfoFactoryImpl.applicationStarting was not invoked",
+                      server.waitForStringInLog("OpenTelemetryInfoFactoryImpl > applicationStarting Entry", server.getDefaultTraceFile()));
+
+        assertNotNull("OpenTelemetryInfoFactoryImpl.applicationStopped was not invoked",
+                      server.waitForStringInLog("OpenTelemetryInfoFactoryImpl > applicationStopped", server.getDefaultTraceFile()));
 
+        assertNotNull("OpenTelemetryInfoImpl was not disposed",
+                      server.waitForStringInLog("EnabledOpenTelemetryInfo > dispose Entry", server.getDefaultTraceFile()));
     }
```
Azquelt: As a follow-up:

1. I don't like the way this test class sets up the app in the setup method, and then deletes the app in the test method. If you need to test something that happens when an app is removed, I think it makes more sense to deploy and then remove the app within the test method.

   Mostly, our tests deploy apps at startup and they remain running until the server is shut down. This test half-conforms to that pattern. It looks like a regular test but has a subtle line in the middle which undeploys the app. It would be easy for someone to add a new test to this class that may well work initially, depending on the execution order, but will fail if the ordering is different.

2. Don't use one test to test two things. You can use the same server, but if you need to test an app being removed, please use a separate app and deploy and remove it within the test method.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/publish/servers/Telemetry10Spi/jvm.options.orig
```diff
@@ -0,0 +1 @@
+-Dio.opentelemetry.context.enableStrictContext=true
```
Azquelt, line 1: As a follow-up: remove this file

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/DisposedOpenTelemetryInfo.java
```diff
@@ -0,0 +1,72 @@
...
+import io.opentelemetry.api.trace.Tracer;
+
+/**
+ * An impl of OpenTelemetryInfo that only occurs if an OpenTelemetryInfo was created after the relevant application has shut down.
+ */
+public class DisposedOpenTelemetryInfo implements OpenTelemetryInfo {
+
+    private static final TraceComponent tc = Tr.register(DisposedOpenTelemetryInfo.class);
+
+    private String appName = "Unkown";
```
Azquelt, line 28: As a follow-up: fix typo

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/EnabledOpenTelemetryInfo.java
```diff
@@ -0,0 +1,114 @@
...
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.logs.SdkLoggerProvider;
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+
+/**
+ * This produces an OpenTelemetry wrapper with info to state whether it is enabled or disabled
+ */
+public class EnabledOpenTelemetryInfo implements OpenTelemetryInfo {
```
Azquelt, line 34: As a follow-up: can we discuss why we have so many different `OpenTelemetryInfo` implementations?

I'm happy that it will work correctly and we need to get this merged to fix existing issues, but I don't like it and the naming of this class in particular could be improved.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,336 @@
...
+        } catch (Exception e) {
+            Tr.error(tc, Tr.formatMessage(tc, "CWMOT5002.telemetry.error", e));
+            return new ErrorOpenTelemetryInfo();
+        }
+    }
+
+    private OpenTelemetryInfo createOpenTelemetryInfo() {
+        try {
+
+            String appName = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor().getComponentMetaData().getJ2EEName().getApplication();
```
Azquelt, line 135: As a follow-up: please move this method and all of the private methods its calls out of this class.

This method is both critical and non-trivial - it ensures that we're configuring OTel correctly and I really don't like that it's mixed in the middle of this class which is otherwise concerned with ensuring that the `OpenTelemetryInfo` is created, stored and disposed of at the correct times.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,336 @@
...
+     *
+     * Within the context of an application's lifecycle we need to ensure OpenTelemetryInfo
+     * is only created once. OpenTelemetryInfoReference handles this.
+     */
+    private class OpenTelemetryInfoReference extends AtomicReference<OpenTelemetryInfoWrappedSupplier> {
+
+        private static final long serialVersionUID = -4884222080590544495L;
+    }
+
+    private class OpenTelemetryInfoWrappedSupplier {
```
Azquelt, line 294: As a follow-up: Add Javadoc to this class to explain its responsibilities and, if possible, come up with a better name for it.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/interfaces/OpenTelemetryAccessor.java
```diff
@@ -14,31 +14,58 @@
...
     }
 
+    /**
+     * Acquires the Baggage for the current Context
+     *
+     * @return Returns Baggage from the current Context, falling back to empty Baggage if none is in the current Context.
+     */
     public static Baggage getBaggage() {
         return Baggage.current();
     }
```
Azquelt, line 71: As a follow-up: remove these three methods.

- `Tracer` is now available from `OpenTelemetryInfo`
- Callers which need the current span or baggage should call the OTel API directly. Having our own method which just delegates to the OTel API is confusing because it looks like we're doing something special when we're not.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
benjamin-confino: For 2 I did it this way to avoid creating an extra server with trace.log enabled, but if you prefer it I could move it to a new test without trouble.

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/EnabledOpenTelemetryInfo.java
benjamin-confino: I was thinking about things from a PMR perspective. The different class names will make key state immediately obvious in the trace logs without introducing clutter.

benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java
```diff
@@ -0,0 +1,336 @@
...
+    public OpenTelemetryInfo getOpenTelemetryInfo(ApplicationMetaData metaData) {
+        try {
+            OpenTelemetryInfoReference atomicRef = (OpenTelemetryInfoReference) metaData.getMetaData(slotForOpenTelemetryInfoHolder);
+            if (atomicRef == null) {
+                //If this is triggered by internal code that isn't supposed to call ApplicationStateListener.applicationStarting() don't throw an error
+                String j2EEName = metaData.getJ2EEName().toString();
+                if (j2EEName.startsWith("io.openliberty") || j2EEName.startsWith("com.ibm.ws")) {
+                    Tr.info(tc, "CWMOT5100.tracing.is.disabled", j2EEName);
+                    return new DisabledOpenTelemetryInfo();
+                }
```
Azquelt, line 120: As a follow up, I don't think we want this INFO message emitted for internal applications.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ClientWithNoCdi.java
```diff
@@ -69,20 +69,35 @@ public static void setUp() throws Exception {
...
 
     //This test does not use open telemetry directly, but it does trigger io.openliberty.microprofile.telemetry.internal.rest.TelemetryClientFilter.java for an app with no CDI and ensures it does not throw an exception
+
+    //This test also ensures that we actually call the started and shutdown methods in OpenTelemetryInfoFactoryImpl
+    //I'm putting it in here to avoid creating an extra server that enables trace. And in an existing method because
+    //Our version of junit doesn't allow fixed test ordering.
     @Test
-    public void testClientWithNoCDi() throws Exception {
+    public void testClientWithNoCDIAndFactoryStarupShutdownCalled() throws Exception {
```
Azquelt, line 77: As a follow-up: fix typo "starup"

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/common/spanexporter/InMemorySpanExporter.java
Azquelt: ASL = ApplicationStateListener. In the current code, the listener isn't registered properly and doesn't fire.

Azquelt (approved):

Azquelt:
#libby

PR #26315, opened 2023-09-18
Always call setTestArray to avoid test ordering issue

### LINKED ISSUE #26314: Fix test ordering problem for io.openliberty.org.apache.jasper.expressionLanguage60.fat.tests.EL60ArrayLengthTest
If the `non_zero` test runs before the zero test then the `zero` test fails.

pnicolucci:
fixes #26314


pnicolucci:
#libby #build

volosied (approved):

PR #26317, opened 2023-09-18
Remove Telemetry packages from TCCL

### LINKED ISSUE #26313: Remove unnecessary OTel packages from thread context class loader
Remove this line from the BND of the bundles which provide the OTel packages:

```
  io.opentelemetry.*;version=${openTelemetryVersion};thread-context=true
```

There are probably some packages which do need to be on the TCCL, but we shouldn't need all of them.

Azquelt:
Fixes #26313


Azquelt:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.telemetry.1.0.internal_fat,io.openliberty.microprofile.telemetry.1.0.internal_fat_tck,io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat,io.openliberty.microprofile.telemetry.1.1.internal_fat_tck

benjamin-confino (approved):

Azquelt:
When #26278 was merged, we thought that additional packages were needed on the TCCL, but I wasn't convinced that we should add everything. I made this PR which reverts the change to the packages available via TCCL to run a personal build to see where the failures were, expecting to need to make further changes.

However, as seen above, the personal build ran without failure, so I'm simply merging the revert.

@fmhwong If there are actually still problems caused by this change, we need to create a test which will tell us that there's a problem.

PR #26324, opened 2023-09-19
Remove hardcoding of port 8020 from OAuthFlow for OpenAPI OAuth UI test

abutch3r:
Previous PR was a partial fix for Oauth Flow URL issue.

Remove host details from OAuth Flow URLs. The URLs can be relative to the application so this is a valid definition for the Urls.

Using the default Fat Test ports, when the new tab is created when clicking `Authorize` will go to the HTTP port causing an SSL error as OAuth flows can only be done via HTTPS. So by disabling HTTP port all redirects go to the https port.

fatTestPorts.xml is not included as on server definition restore the ordering can change meaning that it can appear are the `httpEndpoint` definition in the server.xml and therefore take precedence and re-enable `HTTP`. So fatTestCommon.xml is included with the explicit `httpEndpoint` definition, along with a cut down version of fatTestPorts

#25991 contained an incomplete fix for the issue were the new tab was being opened with the wrong port.

For RTC296869


abutch3r:
#build

Azquelt (approved):

abutch3r:
Latest parent build timed out due to Mac OS provisioning constraints. However once run, the two Mac children builds did both pass the tests running on OSX 13 with either Java 11 and Java 21

PR #26365, opened 2023-09-21
insert, update, and delete with entity parameters

njr-11:
Add and test repository methods with entity parameters.
This includes entities with versions.
It does not include updates to entities with embedded attributes.


njr-11:
#build

KyleAure (approved):

Issue #26342, opened 2023-09-20
ReactiveMessaging "CDI container is not available"


AdisNezirovic:
**Describe the bug**

When trying to inject `RestClient` into Bean which contains reactive messaging channel, I get "CDI container is not available", and application fails on startup.

```
[INFO] [WARNING ] CDI container is not available
[INFO] [ERROR   ] SRMSG00200: The method net.example.channels.KafkaEvents#consume has thrown an exception
[INFO] SRMSG00103: Exception thrown when calling the method net.example.channels.KafkaEvents#consume
[INFO] [ERROR   ] SRMSG00201: Error caught while processing a message in method net.example.channels.KafkaEvents#consume
```

The error seems to be related to OpenTelemetry feature (see trace bellow), but I'd really like to avoid removing that totally from the app, since we use it heavily.

**Steps to Reproduce**

From OpenLiberty starter template with MicroProfile 6.0, and following sample classes:

```
@RegisterRestClient(baseUri = "http://localhost:8080")
public interface ApiClient {
    @POST
    @Path("/")
    @Consumes(MediaType.APPLICATION_JSON)
    public CompletionStage<Response> postEvent(Event ev);
}
```

```
@ApplicationScoped
public class KafkaEvents {
    @Inject @RestClient ApiClient apiClient;

    @Incoming("in")
    public Uni<Void> consume(Message<List<Event>> stream) {
        return Uni.createFrom().voidItem();
    }
}
```
The channel is configured for Kafka in `microprofile-config.properties`

```
mp.messaging.incoming.in.bootstrap.servers=...
mp.messaging.incoming.in.connector=smallrye-kafka
mp.messaging.incoming.in.topic=...
mp.messaging.incoming.in.group.id=openliberty-bug-test
mp.messaging.incoming.in.auto.offset.reset=earliest
...
```

**Expected behavior**  

Application starting and being able to use RestClient from within `consume()` method (as you can see we haven't even come to that).
My current workaround is to use Java 11 async http client, which is slightly too involved for my taste.

**Diagnostic information:**  
 - OpenLiberty Version: 23.0.0.9 (microProfile-6.0 feature)
 - SmallRye: 4.9.0
 - Java Version: openjdk version "17.0.8.1" 2023-08-24
 
 Trace:
 ```
[...]

PR #26368, opened 2023-09-22
new test for using rest client within a reactive message

benjamin-confino:
A new FAT test to test the issue raised by #26342

In that issue it turned out the issue was in the customer's code, they made the recommended modifications and were satisfied it resolved the issue. In the process I noticed their scenario was not covered by the fat and wrote this test to recreate the working design we suggested to the customer.


benjamin-confino:
#build

abutch3r:
@benjamin-confino can you remove the telemetry files from the PR given this is for RM

Emily-Jiang (dismissed):
LGTM

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/ReactiveMessageThatTriggersClientTest.java
```diff
@@ -0,0 +1,118 @@
...
+@RunWith(FATRunner.class)
+@Mode(TestMode.FULL)
+public class ReactiveMessageThatTriggersClientTest extends FATServletClient {
+
+    public static final String APP_NAME = "ReactiveMessageThatTriggersRestClientApp";
+    public static final String REST_APP_NAME = "RestClientTriggeredByReactiveMessagApp";
+    public static final String SERVER_NAME = "Telemetry10ReactiveAndJaxClient";
+
+    @ClassRule
+    public static RepeatTests r = FATSuite.allMPRepeats(SERVER_NAME);
```
Azquelt: I don't think this repeat rule will work.

`allMPRepeats` will run with MP 1.4 (EE7), MP 4.1(EE8), MP 5.0, MP 6.0 and MP 6.1.

`mpReactiveMessaging-3.0` should only work with MP 5.0, MP 6.0 and MP 6.1.

`mpReactiveMessaging-1.0`, I don't think is compatible with any of those (it would only work with MP 2.0 - MP 3.3) and I suspect it does not output any traces anyway.

The `allMPRepeats` repeat does not include changing the version of `mpReactiveMessaging` (since it's a standalone spec, it's not included in any of the MP feature sets), so the test would always run with `mpReactiveMessaging-3.0`.

I think you probably need to run this only against MP 5.0, 6.0 and 6.1.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/bnd.bnd
```diff
@@ -47,11 +47,19 @@ tested.features:\
...
 -buildpath: \
 	com.ibm.ws.componenttest;version=latest,\
+	io.openliberty.jakarta.servlet.6.0;version=latest,\
+	io.openliberty.jakarta.restfulWS.3.1;version=latest,\
+	io.openliberty.jakarta.cdi.4.0;version=latest,\
 	io.openliberty.microprofile.telemetry.internal_fat.common,\
 	io.openliberty.io.opentelemetry;version=latest,\
+	io.openliberty.org.eclipse.microprofile.rest.client.3.0;version=latest,\
+	io.openliberty.jakarta.annotation.2.1;version=latest,\
+	io.openliberty.jakarta.concurrency.3.0;version=latest,\
```
Azquelt: Is it possible to write these tests using the EE7 APIs that the rest of the tests in this bucket depend on and let the transformer convert them to the jakarta ones?

While I think these tests should only run against EE9 and 10 and we can have both the javax and jakarta APIs on the buildpath, I'd prefer not to because it makes it easy to pull in the jakarta version of an annotation when the test needs the javax one.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/bnd.bnd
```diff
@@ -47,11 +48,19 @@ tested.features:\
...
 -buildpath: \
 	com.ibm.ws.componenttest;version=latest,\
+	com.ibm.websphere.javaee.servlet.4.0;version=latest,\
+	com.ibm.websphere.javaee.jaxrs.2.1;version=latest,\
+	com.ibm.websphere.javaee.cdi.1.2;version=latest,\
 	io.openliberty.microprofile.telemetry.internal_fat.common,\
 	io.openliberty.io.opentelemetry;version=latest,\
+	com.ibm.websphere.org.eclipse.microprofile.rest.client.1.3;version=latest,\
+	com.ibm.websphere.javaee.annotation.1.3;version=latest,\
+	com.ibm.websphere.javaee.concurrent.1.0;version=latest,\
```
Azquelt: You don't need any of these additional dependencies because they're already included further down the list.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/rest/rest/client/user/MessageAndRestClientTestServlet.java
```diff
@@ -0,0 +1,145 @@
...
+                sb.append("failed on index: " + index);
+                if (!foundSpanCallingJax) {
+                    sb.append(" Did not find a Span from SimpleReactiveMessagingBean");
+                }
+                if (!foundSpanInsideJax) {
+                    sb.append(" Did not find a Span from RestClientThatCapitalizes");
+                }
+                sb.append(System.lineSeparator());
+                failMessage = failMessage + sb.toString();
+            }
```
Azquelt: Kinda torn on this one. I guess you're ensuring that you're reporting all the results?

It's annoying that it's so much more code than just doing a basic check like this:

```java
assertThat("Failed to find spans for index " + index,
      listOfListOfResults.get(index),
      allOf(hasItem(spanInsideClientMatcher),
            hasItem(spanInsideJaxServiceMatcher()));
```

Or maybe more readably like this:
```java
List<SpanData> results = listOfListOfResults.get(index);
assertThat("Failed to find spans for index "+ index, results, hasItem(span().withName("SimpleReactiveMessagingBean.toUpperCase")));
assertThat("Failed to find spans for index "+ index, results, hasItem(span().withName("RestClientThatCapitalizes.jaxServiceThatCapitalizes")));
```

You could try to wrap up all the checks into one long assert but the error message there would probably be a bit incomprehensible so it's probably best left as is, even though it's not easy to see what the test is checking.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/rest/rest/client/user/MessageAndRestClientTestServlet.java
```diff
@@ -0,0 +1,145 @@
...
+        List<List<SpanData>> listOfListOfResults = new ArrayList<List<SpanData>>(5);
+
+        SpanDataMatcher spanInsideClientMatcher = SpanDataMatcher.hasName("SimpleReactiveMessagingBean.toUpperCase");
+
+        SpanDataMatcher spanInsideJaxServiceMatcher = SpanDataMatcher.hasName("RestClientThatCapitalizes.jaxServiceThatCapitalizes");
+
+        //Each message will end up with four spans being registered. And there are five messages.
+        for (Span s : spanAndExpectedNumber.keySet()) {
+            listOfListOfResults.add(new ArrayList(inMemorySpanExporter.getFinishedSpanItems(spanAndExpectedNumber.get(s), s)));
+        }
```
Azquelt: I don't think you want the code under test reporting how many spans are expected. 

I think if the code under test didn't run for some reason, the test wouldn't expect any spans and would pass without checking anything.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/rest/rest/client/user/MessageAndRestClientTestServlet.java
```diff
@@ -0,0 +1,145 @@
...
+        if (results.size() < 3) {
+            return results;
+        }
+
+        long maxNanos = System.nanoTime() + millisToNanos(MAX_WAIT);
+        //Reactive messaging will populate this on another thread
+        synchronized (results) {
+            while (results.size() < 3 && nanoTimeRemaining(maxNanos)) {
+                results.wait(WAIT_INTERVAL);
+            }
```
Azquelt: These checks look wrong - why are we checking `results.size() < 3`?

Shouldn't we be checking `results.size() == expectedResultCount`?

In other places we've used a lambda to pass in the wait condition and used that to separate the waiting logic from the logic to check the thing we're waiting for.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/rest/rest/client/user/MessageAndRestClientTestServlet.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return results;
+    }
```
Azquelt: If the wait times out, this method should throw an exception rather than returning the results it has.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/publish/servers/Telemetry10ReactiveAndJaxClient/server.xml
```diff
@@ -0,0 +1,32 @@
...
+        <feature>concurrent-3.0</feature>
+        <feature>jsonb-3.0</feature>
+        <feature>restfulWS-3.1</feature>
+        <feature>mpRestClient-3.0</feature>
+        <feature>mpReactiveMessaging-3.0</feature>
+    </featureManager>
+    
+    <library id="exporter" apiTypeVisibility="+third-party">
+		<file name="${server.config.dir}/shared/exporter.jar" commonLibraryRef="exporter" />
+	</library>
```
Azquelt: Indentation

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/jaxrspropagation/responses/JaxRsResponseCodeTestServlet.java
```diff
@@ -85,13 +85,10 @@ public void testRedirect() {
...
     }
 
-    // Method followRedirects() was unavailable before RestClient 2.0
     @Test
-    @SkipForRepeat({TelemetryActions.MP14_MPTEL11_ID, TelemetryActions.MP41_MPTEL11_ID})
     public void testRedirectWithRestClient() {
         JaxRsResponseCodeTestClient client = RestClientBuilder.newBuilder()
                         .baseUri(getBaseUri())
-                        .followRedirects(true)
                         .build(JaxRsResponseCodeTestClient.class);
```
Azquelt: Why have you changed this test?

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/rest/rest/client/user/MessageAndRestClientTestServlet.java
Azquelt: Hmm, I guess you need to know the traceId to query the `inMemorySpanExporter`.

I'd suggest just collecting a list of spans (or traceIds), and check it's the expected length before calling `getFinishedSpanItems`.

I would get rid of the expected number part, it seems to always be 4...

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/rest/rest/client/user/MessageAndRestClientTestServlet.java
```diff
@@ -0,0 +1,145 @@
...
+        List<String> results = getResultsWithWaitIfNeeded();
+
+        assertEquals(3, results.size());
+        assertEquals(3, results.size());
+        assertFalse(results.contains("LENGTH 8"));
+        assertFalse(results.contains("LENGTH 9!"));
+        assertTrue(results.contains("LENGTH 10!"));
+        assertTrue(results.contains("LENGTH 11!!"));
+        assertTrue(results.contains("LENGTH 12!!!"));
+    }
```
Azquelt: This part is just testing what that the reactive messaging app worked as expected, right?

I would argue that we don't even want to run the tests for the telemetry spans if these checks don't pass, so we should combine the two methods and do these checks first.

Also, these checks could be replaced with just `assertThat(results, contains("LENGTH10!", "LENGTH 11!!", "LENGTH 12!!!"));` which will check the length and the contents and report the actual value of `results` if it's not as expected.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/rest/rest/client/user/MessageAndRestClientTestServlet.java
benjamin-confino: If something goes wrong and jump from 2 spans to 4 in a single wait interval then `results.size() == expectedResultCount` will never be true and the the test will wait until timeout. 

So I check for `< 3` then assert the correct number is returned by the method. I could move the assert inside this method but I think `<` would be better practice than `==` here.

benjamin-confino:
#build

Azquelt (approved):

benjamin-confino:
#libby

Issue #26376, opened 2023-09-22
Session property ConnectionRetryCount not on Liberty


zhang-ca:
Session custom property ConnectionRetryCount is implemented on the current Liberty but is not available on the httpSessionDatabase configuration

https://www.ibm.com/docs/en/was-liberty/base?topic=SSEQTP_liberty/com.ibm.websphere.liberty.autogen.nd.doc/ae/rwlp_config_httpSessionDatabase.htm

zhang-ca:
For property detail https://www.ibm.com/docs/en/was-nd/9.0.5?topic=tracking-session-management-custom-properties#ConnectionRetryCount

PR #26377, opened 2023-09-22
Enable session property ConnectionRetryCount

zhang-ca:
Session property ConnectionRetryCount is not visible on  httpSessionDatabase configuration, enable the property in Liberty Kernel configuration metatype.xml.

WebSphere Application Server document for ConnectionRetryCount 
https://www.ibm.com/docs/en/was-nd/9.0.5?topic=tracking-session-management-custom-properties#ConnectionRetryCount

### DIFF
--- a/dev/com.ibm.ws.session.db/resources/OSGI-INF/l10n/metatype.properties
+++ b/dev/com.ibm.ws.session.db/resources/OSGI-INF/l10n/metatype.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2021 IBM Corporation and others.
+# Copyright (c) 2021, 2023 IBM Corporation and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


zhang-ca:
#build

zhang-ca:
fixes #26376

zhang-ca:
for #26376

bensonlatibm (approved):
LGTM!

zhang-ca:
The metatype.xml is for Liberty Kernel (OSGI) to pick up the default value (in this case, the default="2" ) at runtime, so this is not a required value for server.xml. It is also enable the documentation in https://www.ibm.com/docs/en/was-liberty/base?topic=SSEQTP_liberty/com.ibm.websphere.liberty.autogen.nd.doc/ae/rwlp_config_httpSessionDatabase.htm

Since the ConnectionRetryCount property was already implemented in the code during WAS migration (where default value of 2 is set in source code), therefore the property was a hidden property but was always available before this PR 26377 without documented in Liberty.

PR #26378, opened 2023-09-22
update where entity parameter has embeddables

njr-11:
Cover the scenario where the entity that is supplied as a parameter to a repository update operation has embeddables.


njr-11:
#build

mswatosh (approved):

PR #25836, opened 2023-07-28
refresh grpc libs to 1.57.2

hutchig:
Refreshing the grpc libraries.

ALL of the .java file updates are machine generated by the protobuf compiler.

The method exit trace change is just fixing a previous copy and paste error to the correct method name
matching the method entry trace.

PR #26400, opened 2023-09-26
Revert "refresh grpc libs to 1.57.2"

hutchig:
Reverts OpenLiberty/open-liberty#25836
There have been intermittent issue with these libraries so it needs to be reverted and revisited.

With a simple upgrade, unexpected EndOfStream errors occur
(believed to be due to the removal of the flush [here](https://github.com/grpc/grpc-java/pull/9177/files#diff-2de04da34f7e35c085ca26dc596410983f5ded55a8de11eb97811264dea011f2R111) )
and the patch in this PR in ServletServerStream has proved unreliable.


hutchig:
#build

hutchig:
#build
#spawn.fullfat.buckets=com.ibm.ws.grpc_fat

mrsaldana (approved):

hutchig:
The stream already closed error was due to stream 3 going half closed remote which can occur with
networking issues 
(not a local state/race condition error) and looks to be similar to
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=278461
Looking at all the server logs for a remote error all that I see is some
 javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
which is also something that pops up with infrastructure issues.
I have therefore, due to the green build and the nature of the only failure seen above merged the revert.

PR #15072, opened 2020-11-20
14935 Fix non-optimal FreePool invalid connection cleanup

gjwatts:
Fixes #14935

PR #26405, opened 2023-09-26
jms connections may leak when connection error event is used during matchManagedConnection.

jmstephensgit:
The connection error event should be used when the connection is being used by the application.   But in some cases, resource adapter detect connections are bad during matchManagedConnection processing and issue the connection error event instead of throwing a resource exception.   Even though a resource exception is the best way to inform the connection management code of a failing connection, code has been added to tolerate this behavior that can cause multithreaded access to the same connection resulting in some unexpected exceptions.

This fix is for connections that remain in the pool after the connection error event was issues from matchManagedConnection.

If WAS.j2c trace is enabled, the common way to identify this problem is if this is found in the log files.

[REMOVING] MCWrapper id x Managed connection com.ibm.commerce.messaging.adapters.jcaemail.JCAEmailManagedConnection@x State:STATE_ACTIVE_FREE

and after a minute or two of logging, this entry is still occurring with the same MCWrapper id and Managed connection.

This problem will only occur with resource adapters that are calling connection error event during the call to the resource adapters matchManagedConnection method.

### DIFF
--- a/dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/ConnectionEventListener.java
+++ b/dev/com.ibm.ws.jca.cm/src/com/ibm/ejs/j2c/ConnectionEventListener.java
@@ -228,7 +228,7 @@ public final class ConnectionEventListener implements javax.resource.spi.Connect
         }
 
         String tempString = "state " + mcWrapper.getStateString() + " ";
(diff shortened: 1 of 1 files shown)


jmstephensgit:
#build

gjwatts (approved):

gjwatts:
@jmstephensgit Hi Jim.  You can close this PR out.  I already made that change in PR #15072.

PR #26452, opened 2023-09-29
PR for Issue 26225: Fix OIDC social client emitting raw NLS message key instead of message text

### LINKED ISSUE #26225: Social Client WasReqUrl is logging message constant instead of the message String
When we're using a Social client and the WasReqURLOidc cookie doesn't contain host info specified in the webAppSecurity wasReqURLRedirectDomainNames, our request will fail.  We're just getting the message identifier logged in the client servers log.
We get: 
**OIDC_CLIENT_BAD_REQUEST_NO_COOKIE**

ayoho:
Updates the `RedirectionProcessor` class to have a local `Tr` variable that is tied to the NLS messages associated with its package. As the code was written, the `RedirectionProcessor` class accepted a `Tr` instance variable in its constructor and inadvertently used that object to look up an NLS message that wasn't in the associated message bundle. That caused the message to not be found, and the message key was emitted to the logs as-is. Now the local `Tr` variable will emit the correct error message.

Resolves #26225


c00crane:
The fix for this issue is being delivered via https://github.com/OpenLiberty/open-liberty/pull/27993

PR #26464, opened 2023-10-02
Use entire expire time to wait for lock on console.log

### LINKED ISSUE #26463: Server stop command reports error even though server stopped successfully
On Windows Liberty build systems, which are slow, we are seeing RC=21 which indicates that the server did not stop.

jimblye:
Fixes #26463 
Waits entire expire time for console.log which should indicate that the server has stopped.  Under normal conditions this is a Windows only thing, but the actual check is for pid==null.   So potentially that could happen on other platforms.  Not sure of the likelihood of that.

Potential issue is if the user has the console.log open in an editor while issuing the server stop command.   In that case, we'll wait the entire timeout period before eventually returning RC=21 indicating the server didn't stop, although it might have stopped successfully and probably did.


jimblye:
#build

cbridgha (approved):

PR #26468, opened 2023-10-02
JPA Postgres and Consul error message

mswatosh:
Adds Postgres as a platform that supports Unicode for JPA. This prevents an exception when unicode is used with Postgres.

This was manually tested in the https://github.com/o-s-expert/jakarta-data-2023 environment it was discovered in.

Also includes an improvement for the error message when Consul servers can't be reached


mswatosh:
#build

tkburroughs (approved):

mswatosh:
Build looks good., one test failure, which is unrelated.

PR #26474, opened 2023-10-03
Fix ppc64le GLIBC_TUNABLES to allow only P8 hwcaps

ymanton:
For Power (like x86) the hwcaps tunables specified via the environment are added to final set, they do not override the set. Therefore in order to prevent ISA 3.0 and 3.1 (Power 9, 10) features from being used we have to subtract arch_3_00 and arch_3_1 rather than setting arch_2_07.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -980,9 +980,9 @@ readServerEnv()
 criuEnvDefaults() {
   if [ -z "$GLIBC_TUNABLES" ]
   then
-    GLIBC_TUNABLES=glibc.cpu.hwcaps=z14,arch_2_07,-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load:glibc.pthread.rseq=0
+    GLIBC_TUNABLES=glibc.cpu.hwcaps=z14,-arch_3_00,-arch_3_1,-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load:glibc.pthread.rseq=0
   else
-    GLIBC_TUNABLES=${GLIBC_TUNABLES}:glibc.cpu.hwcaps=z14,arch_2_07,-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load:glibc.pthread.rseq=0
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
```diff
@@ -980,9 +980,9 @@ readServerEnv()
 criuEnvDefaults() {
   if [ -z "$GLIBC_TUNABLES" ]
   then
-    GLIBC_TUNABLES=glibc.cpu.hwcaps=z14,arch_2_07,-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load:glibc.pthread.rseq=0
+    GLIBC_TUNABLES=glibc.cpu.hwcaps=z14,-arch_3_00,-arch_3_1,-XSAVEC,-XSAVE,-AVX2,-ERMS,-AVX,-AVX_Fast_Unaligned_Load:glibc.pthread.rseq=0
```
tjwatson, line 983: I assume this is safe to apply now without any additional changes to Semeru or Liberty container images?

tjwatson:
#build

dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
ymanton: Yes, it should be safe and independent of any other changes.

ymanton:
Doing some testing now that suggests that the current patch may not be sufficient. I'm seeing an instruction that is not available on P8 being executed. That instruction has its own hwcap so we may need to subtract it as well. I'll update here when I know more.

tjwatson:
> Doing some testing now that suggests that the current patch may not be sufficient. I'm seeing an instruction that is not available on P8 being executed. That instruction has its own hwcap so we may need to subtract it as well. I'll update here when I know more.

I thought we were only targeting P9+.  If that is the case do we care if there is some instruction running that is not supported on P8?

ymanton:
>I thought we were only targeting P9+. If that is the case do we care if there is some instruction running that is not supported on P8?

True, but the instruction in question requires kernel support, so the real problem is that if we allow it to be used we're stuck only being able to restore on kernels that support it.

Unfortunately hwcaps tunables are not useful here, they don't prevent the use of this particular instruction because it's on a low-level path that I guess is outside the scope of what tunables are supposed to affect. Current patch is sufficient and we don't need to wait on anything from my POV.

tjwatson (approved):

PR #26504, opened 2023-10-03
add additional trace to the JAXRS21ClientCXFRxInvokerTest

WhiteCat22:
EE9 and EE10 fail with:
```
Real response is java.lang.reflect.InvocationTargetException and the expected response is one of [Timeout as expected]
```

The test is supposed to be printing the stack trace of the `java.lang.reflect.InvocationTargetException`, but it is not.

I'm modifying the tests to print the cause of the `Throwable` instead of the `Throwable` itself.


jim-krueger (dismissed):

jim-krueger (approved):

WhiteCat22:
#libby #build

PR #26508, opened 2023-10-04
Run ConfigRefresher.changesEnded in multi-threaded checkpoint hook

dazavala:
Invoke ConfigRefresher.changesEnded() within a multi-threaded hook when restoring from checkpoint. Restore may fail when changesEnded() is invoked in a single-threaded hook, because the method schedules a task to an executor that uses a blocking queue -- i.e. performs a blocking operation.

See RTC 297336.

### DIFF
--- a/dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
+++ b/dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
@@ -12,9 +12,12 @@
  *******************************************************************************/
 package com.ibm.ws.config.xml.internal;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Hashtable;
 import java.util.Map;
+import java.util.Queue;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
@@ -22,6 +25,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import org.osgi.framework.BundleContext;
+import org.osgi.framework.Constants;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.util.tracker.ServiceTracker;
 
 import com.ibm.websphere.config.ConfigUpdateException;
@@ -58,7 +63,8 @@ public class ConfigRefresher {
     private long configStartTime = 0;
     private Collection<Future<?>> futuresForChanges = null;
 
-    private ChangesEndedHook changesEndedHook = null;
+    private final ChangesEndedHook changesEndedHook;
+    private final ServiceRegistration<CheckpointHook> checkpointHookRegistration;
 
     ConfigRefresher(BundleContext bundleContext,
                     ChangeHandler changeHandler, ServerXMLConfiguration serverXMLConfig, ConfigVariableRegistry variableRegistry) {
@@ -76,11 +82,13 @@ public class ConfigRefresher {
         metatypeTracker = new ServiceTracker<MetaTypeRegistry, MetaTypeRegistry>(bundleContext, MetaTypeRegistry.class.getName(), null);
         metatypeTracker.open();
 
-        final CheckpointPhase checkpoint = CheckpointPhase.getPhase();
-        if (!checkpoint.restored()) {
-            changesEndedHook = new ChangesEndedHook();
-            checkpoint.addMultiThreadedHook(changesEndedHook);
-        }
+        changesEndedHook = new ChangesEndedHook();
+        Hashtable<String, Object> hookProps = new Hashtable<>();
+        // Lesser ranking ensures changesEnded() executes ASAP after the SystemConfiguration
+        // single-threaded restore hook
+        hookProps.put(Constants.SERVICE_RANKING, Integer.MIN_VALUE + 10000);
+        hookProps.put(CheckpointHook.MULTI_THREADED_HOOK, Boolean.TRUE);
+        checkpointHookRegistration = bundleContext.registerService(CheckpointHook.class, changesEndedHook, hookProps);
     }
 
     void start() {
@@ -90,6 +98,10 @@ public class ConfigRefresher {
     void stop() {
         configurationMonitor.stopConfigurationMonitoring();
         runtimeUpdateManagerTracker.close();
+
+        if (checkpointHookRegistration != null) {
+            checkpointHookRegistration.unregister();
+        }
     }
 
     public void refreshConfiguration() {
@@ -147,10 +159,14 @@ public class ConfigRefresher {
             // Let the notification show that we got an error while making the configuration changes
             configUpdatesDelivered.setResult(e);
         } finally {
-
             if (CheckpointPhase.getPhase().restored() && changesEndedHook.runMultiThreaded) {
-                // The hook will invoke changesEnded(configUpdatesDelivered)
-                changesEndedHook.configUpdatesDelivered = configUpdatesDelivered;
+                synchronized (changesEndedHook) {
+                    if (changesEndedHook.runMultiThreaded) {
+                        changesEndedHook.configUpdatesToDeliver.add(configUpdatesDelivered);
+                    } else {
+                        changesEnded(configUpdatesDelivered);
+                    }
+                }
             } else {
                 changesEnded(configUpdatesDelivered);
             }
@@ -283,16 +299,15 @@ public class ConfigRefresher {
 
     }
 
-    // Register as a multi-threaded hook to invoke changesEnded() in
-    // multi-threaded mode when restoring from checkpoint. Invoking
-    // changesEnded() in single-threaded mode may fail restore as it
-    // performs a blocking operation.
+    // Register as a multi-threaded hook to invoke changesEnded() when
+    // restoring from checkpoint. Invoking changesEnded() in single-threaded
+    // mode may fail restore as it performs a blocking operation.
     private class ChangesEndedHook implements CheckpointHook {
 
-        private boolean runMultiThreaded = false;
+        volatile boolean runMultiThreaded = false;
 
-        // Set before restore is called
-        private RuntimeUpdateNotification configUpdatesDelivered = null;
+        // FIFO queue of deferred config update notifications
+        Queue<RuntimeUpdateNotification> configUpdatesToDeliver = new ArrayDeque<RuntimeUpdateNotification>();
 
         @Override
         public void prepare() {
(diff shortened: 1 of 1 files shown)


dazavala:
#build

dazavala:
The previous build failed unitttest `NettyFrameworkImplTest.testTCPStartOutbound` (see RTC-[296690](https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=296690)). Rebased on current integration and trying again...

#build

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
```diff
@@ -136,7 +147,13 @@ private synchronized void doRefresh(Map<String, DeltaType> variableDelta) {
             // Let the notification show that we got an error while making the configuration changes
             configUpdatesDelivered.setResult(e);
         } finally {
-            changesEnded(configUpdatesDelivered);
+
+            if (CheckpointPhase.getPhase().restored() && changesEndedHook.runMultiThreaded) {
```
tjwatson: The `INACTIVE` returns true from `restored()`.  I am surprised this does not NPE on `changesEndedHook` when the phase is `INACTIVE` ...

But now I see that you would always create the `new ChangesEndedHook();` in the constructor in this case of `INACTIVE` phase.  Is that intended?  It is a bit confusing to me.

I suggest changing the constructor to always create the ChangesEndedHook and check the `INACTIVE` phase:

```java
        changesEndedHook = new ChangesEndedHook();
        if (checkpoint != CheckpointPhase.INACTIVE && !checkpoint.restored()) {
            checkpoint.addMultiThreadedHook(changesEndedHook);
        }
```
Make `changesEndedHook` final.

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
```diff
@@ -265,4 +282,30 @@ public void variableRefresh(Map<String, DeltaType> deltaMap) {
...
 
     }
+
+    // Register as a multi-threaded hook to invoke changesEnded() in
+    // multi-threaded mode when restoring from checkpoint. Invoking
+    // changesEnded() in single-threaded mode may fail restore as it
+    // performs a blocking operation.
+    private class ChangesEndedHook implements CheckpointHook {
+
+        private boolean runMultiThreaded = false;
```
tjwatson: Make package private to avoid synthetic method access here.  Also needs to be volatile since multiple threads may access it each config update..

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
```diff
@@ -265,4 +282,30 @@ public void variableRefresh(Map<String, DeltaType> deltaMap) {
...
+
+        @Override
+        public void prepare() {
+            runMultiThreaded = true;
+        }
+
+        @Override
+        public void restore() {
+            if (runMultiThreaded) {
+                changesEnded(configUpdatesDelivered);
```
tjwatson: set `runMultiThreaded` to false here so if other config updates happen after restore they will work.

tjwatson (changes requested):

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
dazavala: Done.  Yes, that was unintentional.  I had equated !restored() to (!phase.INACTIVE && !phase.restored).  Also, it was impossible to make the hook final in the original solution.

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
dazavala: Done

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
dazavala: Done

dazavala:
@tjwatson I would prefer this solution registers the multi-threaded hook as a service, where I can assign it a lower rank. That would reduce the window between when last single-threaded hook and the multi-threaded hook will run.  My understanding is that hooks added via CheckpointPhase.addHook() have rank 0, which isn't ideal for this solution.

dazavala:
#build
Build changes requested by @tjwatson.

dazavala:
Uploaded [Alternate hook registration with lesser ranking](https://github.com/OpenLiberty/open-liberty/pull/26508/commits/5497b7eadf401bb6226c536fb312ba21b147e2ff). Solution works as intended. The changesEnded restore hook runs almost immediately after the SystemConfiguration single-threaded restore hook.

#build

dazavala:
Latest builds passed 100%.  @tjwatson Ready for your review.

#libby

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
```diff
@@ -136,7 +157,12 @@ private synchronized void doRefresh(Map<String, DeltaType> variableDelta) {
             // Let the notification show that we got an error while making the configuration changes
             configUpdatesDelivered.setResult(e);
         } finally {
-            changesEnded(configUpdatesDelivered);
+            if (CheckpointPhase.getPhase().restored() && changesEndedHook.runMultiThreaded) {
+                // The hook will invoke changesEnded(configUpdatesDelivered)
+                changesEndedHook.configUpdatesDelivered = configUpdatesDelivered;
```
tjwatson: This could be a non-issue ...

It is not clear to me if multiple `RuntimeUpdateNotification` could be produced during a restore.  It would involve multiple calls to `com.ibm.ws.config.xml.internal.ConfigRefresher.doRefresh(Map<String, DeltaType>)`.  If that is possible then the "last one wins" and the earlier notifications would not get fired.

I question if `changesEndedHook.configUpdatesDelivered` should be some thread-safe queue that is drained in the `restore()` method.

dazavala:
Uploaded [Change hook registration and queue config update notifications](https://github.com/OpenLiberty/open-liberty/pull/26508/commits/11a3f1143769c33736cd4b820d45aa52fa8f8234).

#build

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
```diff
@@ -265,4 +295,31 @@ public void variableRefresh(Map<String, DeltaType> deltaMap) {
...
+
+    // Register as a multi-threaded hook to invoke changesEnded() when
+    // restoring from checkpoint. Invoking changesEnded() in single-threaded
+    // mode may fail restore as it performs a blocking operation.
+    private class ChangesEndedHook implements CheckpointHook {
+
+        volatile boolean runMultiThreaded = false;
+
+        // FIFO queue of deferred config update notifications
+        Queue<RuntimeUpdateNotification> configUpdatesToDeliver = new ArrayDeque<RuntimeUpdateNotification>();
```
tjwatson: type should be a `Deque`

dazavala:
#build

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
dazavala: I've uploaded the requested changes to [Change hook registration and queue config update notifications](https://github.com/OpenLiberty/open-liberty/pull/26508/commits/85de2e5cd7a82f44e204747c57657bb931c98e8b).

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
```diff
@@ -136,7 +159,17 @@ private synchronized void doRefresh(Map<String, DeltaType> variableDelta) {
...
+                synchronized (changesEndedHook) {
+                    if (changesEndedHook.runMultiThreaded) {
+                        changesEndedHook.configUpdatesToDeliver.add(configUpdatesDelivered);
+                    } else {
+                        changesEnded(configUpdatesDelivered);
+                    }
+                }
+            } else {
+                changesEnded(configUpdatesDelivered);
+            }
```
tjwatson: I think we should create a method `boolean ChangesEndedHook.queueNotification(RuntimeUpdateNotification)` method to call here instead of leaking `runMultiThreaded` and `configUpdatesToDeliver` out to here.  So it would end up being:

```java
            if (!changesEndedHook.queueNotification(configUpdatesDelivered) {
                changesEnded(configUpdatesDelivered);
            }
```

Then the queueNotifications could look like this:

```java
    private class ChangesEndedHook implements CheckpointHook {

        private final ThreadLocal<Boolean> checkpointThread = new ThreadLocal<Boolean>() {
             @Override protected Boolean initialValue() {
                 return Boolean.FALSE;
             }
         };

        // FIFO queue of deferred config update notifications
        final Deque<RuntimeUpdateNotification> configUpdatesToDeliver = new ArrayDeque<RuntimeUpdateNotification>();

        boolean queueNotification(RuntimeUpdateNotification notification) {
            if (!checkpointThreaded.get()) {
                return false;
            }
            configUpdatesToDeliver.add(notification);
        }

        @Override
        public void prepare() {
            checkpointThread.set(true);
        }

        @Override
        public void restore() {
            checkpointThread.set(false);
            RuntimeUpdateNotification notification = null;
            while (notification = configUpdatesToDeliver.pollFirst() != null) {
                changesEnded(notification);
            }
        }
    }
```

tjwatson (changes requested):

tjwatson (approved):

dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/ConfigRefresher.java
dazavala: Changes uploaded to  [Queue config update notifications](https://github.com/OpenLiberty/open-liberty/pull/26508/commits/dd933c0f7528a9c99c11333238bc90604504ae53). Two minor changes were added. Local testing and inspection look good.

dazavala:
#build
#libby

dazavala:
100% pass on [InstantOn build](https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_1u-4wGhJEe69i6b-9QSuMw&tab=com.ibm.team.build.web.ui.internal.editors.result.overview.OverviewPage). Waiting for [dazavala-26508-20231011-1024](https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_L2keIGhOEe69i6b-9QSuMw) or dazavala-26508-20231011-0942.

dazavala:
All personal builds passed at 100%. Merge!

PR #26548, opened 2023-10-10
Improve use of WSContextService in MP Reactive Streams Operators 3.0

### LINKED ISSUE #26352: Review asynchronous execution in Reactive Streams 3.0
Was: ~Replace ExecutorService in ReactiveStreams30 with ManagedExecutorService~

https://github.com/OpenLiberty/open-liberty/blob/51c0e42e906390c31db6bd58a8579010bf7dfc79/dev/io.openliberty.microprofile.reactive.streams.operators30.internal/src/io/openliberty/microprofile/reactive/streams/operators30/spi/impl/LibertyReactiveStreamsEngineImpl.java#L56 should accept a ManagedExecutorService instead of a standard ExecutorService

Azquelt:
~Use ManagedExecutorService when initially subscribing to streams in RSO 3.0.~

During review, some problems with the initial approach were pointed out:
- Using `ManagedExecutorService` requires a dependency on the `concurrent-x.x` feature, which we're not sure should be required here
- The user can create and configure any number of `ManagedExecutorService`s, and we weren't using a filter to ensure we get the right one.
- If we're using `ManagedExecutorService`, we should allow the user to configure _which_ `ManagedExecutorService` we use. While using the default one is fine in most cases, we don't want to back users into a corner where they want to configure the service one way for one feature and another way for another feature but can't.

So instead:

- Make sure we're using the internal WSContextService singleton (which does not require the concurrent feature and is not configurable by users) and that we're submitting tasks to liberty's global thread pool directly.
  - We may want to check if async is necessary here. If so, we may want to consider allowing the user to configure a different context service.
- Use the mutiny SPI ExecutionModel to instruct mutiny to use an executor which calls WSContextService. This replaces the previous code which wrapped one CompletionStage around another.

For #26352


Azquelt:
#libby
#build

dev/io.openliberty.microprofile.reactive.streams.operators30.internal/src/io/openliberty/microprofile/reactive/streams/operators30/spi/impl/RsoContextExecutor.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final Map<String, ?>[] THREAD_CONTEXT_PROVIDERS = new Map[] {
+                                                                                 Collections.singletonMap(WSContextService.THREAD_CONTEXT_PROVIDER,
+                                                                                                          "com.ibm.ws.classloader.context.provider"),
+                                                                                 Collections.singletonMap(WSContextService.THREAD_CONTEXT_PROVIDER,
+                                                                                                          "com.ibm.ws.javaee.metadata.context.provider"),
+                                                                                 Collections.singletonMap(WSContextService.THREAD_CONTEXT_PROVIDER,
+                                                                                                          "com.ibm.ws.security.context.provider"),
+    };
+
+    @Reference(target = "(service.pid=com.ibm.ws.context.manager)")
```
tevans78, line 49: Comments needed in this class to explain why these filters etc

tevans78 (dismissed):
Approving assuming that the comments will be updated

Azquelt:
Comments added
#libby
#build

abutch3r (approved):
Based on Tom's prior review, comments have been added to address his requested updates

PR #26571, opened 2023-10-11
Update to SmallRye Metrics 5.1.0-RC3 and disable optional test until …

Channyboy:
…MP Metrics 5.1.1

#build

RC3 pulls in a fix for `gc.time` to be a gauge instead of a counter. The TCK has a bug and expects it to be a counter. The optional test is disabled until the TCK can be fixed in `5.1.1`


Channyboy:
Error encountered in PR is not related to this build. There already exists BB item tracking it.

fmhwong (approved):
Can you document why we exclude `MPMetricBaseMetricsTest`?

PR #26585, opened 2023-10-12
A warning message should be logged on restore when a config property is updated after checkpoint even if the property had no default value.

anjumfatima90:
The scenario should occur only when the config property is accessed early during the checkpoint.

When config is injected as Optional<T>, the value doesn't get updated on restore.
But when the config is injected as Provider<T>, the value gets updated on restore.

In both the cases we would log the CWWKC0651W warning message.


anjumfatima90:
#build

tjwatson (approved):

anjumfatima90:
#libby

PR #26586, opened 2023-10-12
Wait for config update in concurrent persistent FAT

tkburroughs:
Test is updating configuration to disable running of persistent tasks but is not waiting for the update to complete, so tasks are running too early. Added a wait for the update to complete and ensure app is started (in case it was restarted).


brideck (approved):

tkburroughs:
#build

PR #26595, opened 2023-10-13
Update annotation on method with entity results

njr-11:
Experiment with various scenarios where the Update annotation is used on a method that returns entity results in various ways.  Include tests.


KyleAure (approved):

njr-11:
#build

PR #26624, opened 2023-10-16
lifecycle methods that return records

njr-11:
Experiment with Insert/Update/Save returning a record rather than a JPA entity.


njr-11:
#build

mswatosh (approved):

PR #26625, opened 2023-10-16
match the spec API for BasicRepository

njr-11:
Remove the `@Delete` annotation for a method of BasicRepository that doesn't match up with the spec.


njr-11:
#build

mswatosh (approved):

PR #26627, opened 2023-10-16
Fix config updates in RAR BeanValidation FAT

tkburroughs:
Test is not waiting for apps to stop when updating config resulting in multiple concurrent attempts to start the same app. Fixed by:
- Removed `@RunWith(FATRunner.class)` from abstract class, as it was preventing `@After` method from running, which waits for app stop and restores original config
- Moved `@Mode(TestMode.FULL)` to test classes, as it was ignored on abstract parent
- Increased application and configuration update wait times if set too small when running locally as some applications took longer than default to start


tkburroughs:
#build

jmstephensgit (approved):

PR #26635, opened 2023-10-17
test lifecycle methods returning array and iterables of records

njr-11:
Add tests for the Insert, Update, and Save lifecycle methods returning arrays and iterables of records.


njr-11:
#build

gjwatts (approved):

PR #26640, opened 2023-10-18
Add tests for mp metrics integration with reactive messaging 3.0

### LINKED ISSUE #26243: Add Tests for MP RM 3.0 with mpMetrics integration
While the TCK does have some metrics scenarios. This might not be sufficient for our requirements as the calls to metrics are internal.

We need to firstly determine if the existing TCK tests are sufficient for metrics integration. If not we need to add tests that call the metrics endpoint externally

### LINKED ISSUE #26665: Enable Metrics tests for Emitter channels
The current Smallrye implementation does not include metrics for channels that are associated with Emitters.

It does work for Outgoing and Incoming channels.

https://github.com/smallrye/smallrye-reactive-messaging/issues/2332

ryan-storey:
~~The assertions for the emitter messages on the outgoing channel are currently disabled as emitters currently don't report metrics per spec. These can be re-enabled when the issue is fixed. The response on the incoming channel is fine and can be tested against as this is picked up via a reception bean.~~
The above bug has been fixed in this PR

Closes #26243 
Fixes #26665

This PR is dependent on #26754


dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,151 @@
...
+            server.stopServer();
+        } finally {
+            KafkaUtils.deleteKafkaTopics(KafkaTests.getAdminClient());
+        }
+    }
+
+    private HashMap<String, Integer> getMetrics() throws Exception {
+        // Returns the MicroProfile Reactive Messaging metrics from the /metrics endpoint
+        HttpURLConnection response = HttpUtils.getHttpConnection(new URL("http", server.getHostname(), 8010, "/metrics"),
+                                                                 200, null, 30, HTTPRequestMethod.GET, null, null);
```
Azquelt: Using a hard-coded port is wrong here.

This should probably be changed to use methods on `componenttest.topology.utils.HttpUtils`

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTestServlet.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Inject
+    MetricsReceptionBean emitterReception;
+
+    @Inject
+    MetricsDeliveryBean deliveryBean;
+
+    @Inject
+    @Channel(EMITTER_OUTGOING_CHANNEL)
+    public Emitter<String> emitter;
```
Azquelt: In general, fields you inject like this should be `private`

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTestServlet.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public static final String EMITTER_OUTGOING_CHANNEL = "MetricsEmitter";
+    public static final String EMITTER_TOPIC = "EmitterTopic";
+    public static final String OUTGOING_TOPIC = "OutgoingTopic";
```
Azquelt: Usually constants go at the top of the class.

Occasionally it's more helpful to have them next to the thing that uses them, but I don' think that's the case here.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTestServlet.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public void deliverMessage() {
+        // deliver 5 messages via delivery and reception beans
+        for (int i = 1; i < 6; i++) {
+            deliveryBean.sendMessage("test");
+        }
+        KafkaReader<String, String> reader = kafkaTestClient.readerFor(OUTGOING_TOPIC);
+        reader.assertReadMessages(5, KafkaTestConstants.DEFAULT_KAFKA_TIMEOUT);
+    }
```
Azquelt: Add a comment to each of these that it's called manually from `MetricsTest`, so that it's clear you haven't just forgotten to add the `@Test` annotation.

Azquelt:
First PR comment needs a link back to the issue. If it's not fixing the whole issue then it should be clear, either from the PR or from the issue itself, what still needs to be done.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/bnd.bnd
```diff
@@ -1,5 +1,5 @@
 #*******************************************************************************
-# Copyright (c) 2019, 2022 IBM Corporation and others.
+# Copyright (c) 2023 IBM Corporation and others.
```
Azquelt:
```suggestion
# Copyright (c) 2019, 2023 IBM Corporation and others.
```

Copyright headers should generally have two dates, the year it was first published and the year it was last changed.

If those are the same, you just put one year, but you shouldn't be removing the first date from existing files.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,151 @@
...
+        server.startServer();
+    }
+
+    @Test
+    public void testDeliverPayload() throws Exception {
+        // Test that metrics are updated when a payload is delivered via an emitter
+        runTest(server, APP_NAME + "/MetricsTest", "emitterDeliverPayload");
+        HashMap<String, Integer> metrics = getMetrics();
+
+        assertThat("5 messages not detected on channel", metrics.get("metrics-incoming") == 5);
```
Azquelt: Either
```assertThat("5 messages not detected on channel", metrics.get("metrics-incoming"), equalTo(5));```
or
```assertEquals("5 messages not detected on channel", 5, metrics.get("metrics-incoming"));```

The existing code will not tell you how many messages were detected if it wasn't 5.

File: dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java [hunk 7]
```diff
@@ -0,0 +1,151 @@
...
+    @Test
+    public void testDeliverMessage() throws Exception {
+        // Test that metrics are updated when a message is delivered via delivery and reception beans
+        runTest(server, APP_NAME + "/MetricsTest", "deliverMessage");
+        HashMap<String, Integer> metrics = getMetrics();
+
+        assertThat("5 messages not detected on channel", metrics.get("metrics-outgoing") == 5);
+        assertThat("Outgoing and incoming channel counts are not the same",
+                   metrics.get("metrics-outgoing") == metrics
+                                   .get("metrics-message-incoming"));
```
Azquelt: As above, use either `assertThat(..., equalTo(...))` or `assertEquals(a, b)`

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        // filter the /metrics output for reactive messaging metrics
+        while ((currentLine = in.readLine()) != null)
+            if (currentLine.contains("mp_messaging_message_count_total{channel=")) {
+                String channelString = currentLine.split("channel=\"")[1].replace(",mp_scope=\"base\",", "");
+                // the metric counter is always a whole number (int)
+                // extract channel name and counter total from response line
+                metrics.put(channelString.split("\"} ")[0], (int) Float.parseFloat(channelString.split("\"} ")[1]));
+            }
+        in.close();
```
Azquelt: Don't use `while` without the `{}` braces. They're not technically needed here, but we always put them in because it makes it easier to read.

Azquelt: `close()` won't be called if an exception is thrown in the preceding code.

Use a [try-with-resources block instead](https://learning.oreilly.com/library/view/effective-java-3rd/9780134686097/ch2.xhtml#lev9) (or a `finally` block if there's some reason you can't use try-with-resources).

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,151 @@
...
+        String currentLine;
+
+        // filter the /metrics output for reactive messaging metrics
+        while ((currentLine = in.readLine()) != null)
+            if (currentLine.contains("mp_messaging_message_count_total{channel=")) {
+                String channelString = currentLine.split("channel=\"")[1].replace(",mp_scope=\"base\",", "");
+                // the metric counter is always a whole number (int)
+                // extract channel name and counter total from response line
+                metrics.put(channelString.split("\"} ")[0], (int) Float.parseFloat(channelString.split("\"} ")[1]));
+            }
```
Azquelt: This might be easier with a regex and capturing groups.

```java
Matcher m = Pattern.compile("mp_messaging_message_count_total{.*channel=\"([^\"])\".*} +(.*)").matcher(currentLine);
if (m.find()) {
    metrics.put(m.group(1), (int) Float.parseFloat(m.group(2)));
}
```

Azquelt (changes requested):
As a whole the new tests look good, but there are a few bits that need fixing and some others that could be improved.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/bnd.bnd
abutch3r: @Azquelt given this file is new to this year - it was a mistake on my part originally - it should just be 2023.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
ryan-storey: I had a play around with this but I can't seem to get it working with regex when `,mp_scope="base",` is present.

ryan-storey:
#build

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,153 @@
...
+    public void testDeliverPayload() throws Exception {
+        // Test that metrics are updated when a payload is delivered via an emitter
+        runTest(server, APP_NAME + "/MetricsTest", "emitterDeliverPayload");
+        HashMap<String, Integer> metrics = getMetrics();
+
+        assertThat("5 messages not detected on channel", metrics.get("metrics-incoming"), equalTo(5));
+//      Assertion disabled as emitters currently don't report metrics per spec
+//        assertTrue("Outgoing and incoming channel counts are not the same",
+//                   metrics.get("mp_messaging_message_count_total{channel=\"MetricsEmitter\",mp_scope=\"base\",}") == metrics
+//                                   .get("mp_messaging_message_count_total{channel=\"metrics-incoming\",mp_scope=\"base\",}"));
```
Azquelt: Is there an issue open for this?

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Test
+    public void testDeliverMessage() throws Exception {
+        // Test that metrics are updated when a message is delivered via delivery and reception beans
+        runTest(server, APP_NAME + "/MetricsTest", "deliverMessage");
+        HashMap<String, Integer> metrics = getMetrics();
+
+        assertThat("5 messages not detected on channel", metrics.get("metrics-outgoing"), equalTo(5));
+        assertThat("Outgoing and incoming channel counts are not the same", metrics.get("metrics-outgoing"), equalTo(metrics.get("metrics-message-incoming")));
```
Azquelt: I didn't notice this until you removed the unused injected `MetricsReceptionBean`, but you're checking that the metrics for the incoming channel have been updated before you confirm that the messages have been received. There's no wait here either, so it would be possible for this check to run before the incoming messages have been processed which would fail the test.

The easiest way to fix this is to put the MetricsReceptionBean back in and use `assertReceivedMessages` to wait for the messages to be received.

You also probably want a separate reception bean for each incoming channel since that will make it easier to assert the right number of messages are received.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTestServlet.java
```diff
@@ -0,0 +1,60 @@
...
+
+    // called manually from MetricsTest
+    public void emitterDeliverPayload() {
+        // deliver 5 payloads via an emitter
+        for (int i = 1; i < 6; i++) {
+            emitter.send("Message " + i);
+        }
+
+        KafkaReader<String, String> reader = kafkaTestClient.readerFor(EMITTER_TOPIC);
+        reader.assertReadMessages(5, KafkaTestConstants.DEFAULT_KAFKA_TIMEOUT);
```
Azquelt: As per the [other comment](https://github.com/OpenLiberty/open-liberty/pull/26640/files#r1365492208), replace this check with one that waits for the messages to be received by the reception bean.

ryan-storey:
#build

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/bnd.bnd
```diff
@@ -27,19 +27,22 @@ fat.test.container.images: confluentinc/cp-kafka:7.1.1
...
-    io.openliberty.org.testcontainers;version=latest,\
-    com.ibm.ws.componenttest.2.0;version=latest
+	io.openliberty.microprofile.reactive.messaging.internal_fat.common.jakarta;version=latest,\
+	io.openliberty.org.eclipse.microprofile.reactive.messaging.3.0;version=latest,\
+	io.openliberty.org.eclipse.microprofile.reactive.streams.operators.3.0;version=latest,\
+	com.ibm.websphere.org.reactivestreams.reactive-streams.1.0;version=latest,\
+	io.openliberty.org.eclipse.microprofile.config.3.0;version=latest,\
+	io.openliberty.jakarta.cdi.3.0;version=latest,\
+	io.openliberty.jakarta.servlet.5.0;version=latest,\
+	org.apache.kafka:kafka-clients;version='3.5.1',\
```
abutch3r: not sure where the `'` has come from, but should be put bkac

Azquelt: I've noticed that if you change stuff via the editor, it will add single quotes around the version like this. It doesn't look like there are any real changes in this part of the file, so ideally the change should be removed so as not to clutter the diff.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
abutch3r: https://github.com/OpenLiberty/open-liberty/issues/26665

dev/io.openliberty.io.smallrye.reactive.messaging-provider4/src/io/smallrye/reactive/messaging/providers/extension/AbstractEmitter.java
```diff
@@ -0,0 +1,210 @@
+package io.smallrye.reactive.messaging.providers.extension;
```
abutch3r: Need to include an appropriate copyright statement, use what we have put into the EmitterImpl class.

Also need to include a link back to the original source class given this is a chance to that. again EmitterImpl has the details

dev/io.openliberty.io.smallrye.reactive.messaging-provider4/src/io/smallrye/reactive/messaging/providers/extension/AbstractEmitter.java
```diff
@@ -0,0 +1,210 @@
...
+import io.smallrye.reactive.messaging.providers.helpers.NoStackTraceException;
+import io.smallrye.reactive.messaging.providers.metrics.MetricDecorator;
+import io.smallrye.reactive.messaging.providers.metrics.MicrometerDecorator;
+import jakarta.enterprise.inject.Instance;
+import jakarta.enterprise.inject.spi.CDI;
+
+public abstract class AbstractEmitter<T> implements MessagePublisherProvider<T> {
+    public static final NoStackTraceException NO_SUBSCRIBER_EXCEPTION = new NoStackTraceException("Unable to process message - no subscriber");
+    protected final AtomicReference<MultiEmitter<? super Message<? extends T>>> internal = new AtomicReference<>();
+    protected final Multi<Message<? extends T>> publisher;
```
Azquelt: This class, along with any others taken from an external open source project, should probably have a header indicating its license and where it came from.

I'm unsure if it's _strictly_ needed under the terms of the Apache 2.0 license since the original file has no copyright notice, we haven't made any modifications and we should already include the smallrye RM library in our NOTICES.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,154 @@
...
+    @Test
+    public void testDeliverPayload() throws Exception {
+        // Test that metrics are updated when a payload is delivered via an emitter
+        runTest(server, APP_NAME + "/MetricsTest", "emitterDeliverPayload");
+        HashMap<String, Integer> metrics = getMetrics();
+
+        assertThat("5 messages not detected on channel", metrics.get("metrics-incoming"), equalTo(5));
+        assertThat("Outgoing and incoming channel counts are not the same",
+                   metrics.get("MetricsEmitter"),
+                   equalTo(metrics.get("metrics-incoming")));
```
Azquelt: Use the channel name constants where you're referencing them in the tests.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTestServlet.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ *
+ */
+@SuppressWarnings("serial")
+@WebServlet("/MetricsTest")
+public class MetricsTestServlet extends AbstractKafkaTestServlet {
+
+    public static final String EMITTER_OUTGOING_CHANNEL = "MetricsEmitter";
+    public static final String EMITTER_TOPIC = "EmitterTopic";
+    public static final String OUTGOING_TOPIC = "OutgoingTopic";
```
Azquelt, line 30: More a note for the future than something that _needs_ changed: try to keep your names consistent.

For each test you have a test method, a method in the servlet, two channel names, two constant names, one or two helper beans and a topic name. Ideally, these should all be named consistently so that it's obvious at a glance which test it belongs to.

I know that these tests have gone through several revisions, and that's probably where the inconsistency has come from, but it's often worth going back through at the end and making everything match up. It will make it easier to read and debug later. (Eclipse's refactoring tools are helpful here)

ryan-storey:
Note that this PR is dependent on #26754 as it requires functionality from 4.10

abutch3r:
@ryan-storey can you look at resolving the above conflicts with integration and once pushed, kick off the builds

ryan-storey:
#build

abutch3r:
#libby

dev/io.openliberty.io.smallrye.reactive.messaging-provider4/src/io/smallrye/reactive/messaging/providers/wiring/Wiring.java
```diff
@@ -0,0 +1,910 @@
+/*
```
abutch3r: This class should no longer needed overlaying, so should be removed from the PR

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MultiAppMetricsTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public static final String APP_ONE_NAME = "MetricsTestOne";
+    public static final String APP_TWO_NAME = "MetricsTestTwo";
+    public static final String SERVER_NAME = "SimpleRxMessagingServer";
+
+    @Server(SERVER_NAME)
+    public static LibertyServer server;
+
+    @ClassRule
+    public static RepeatTests r = ReactiveMessagingActions.repeat(SERVER_NAME,
```
abutch3r, line 62: Given we are covering MP60_RM30 in the base metrics test, I would remove that repeat from here to reduce the overall FAT bucket time, given this already takes a while to run

ryan-storey:
#build #libby

abutch3r (dismissed):

abutch3r:
CI Build failures are either known fault tolerance test issues or issues with the underlying CI infrastructure. the tests in question are green in both CI and RTC 

@Azquelt can you take another look to see if there is anything else you think needs to be done

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,154 @@
...
+                if (currentLine.contains("mp_messaging_message_count_total{channel=")) {
+                    String channelString = currentLine.split("channel=\"")[1].replace(",mp_scope=\"base\",", "");
+                    // the metric counter is always a whole number (int)
+                    // extract channel name and counter total from response line
+                    metrics.put(channelString.split("\"} ")[0], (int) Float.parseFloat(channelString.split("\"} ")[1]));
+                }
+            }
+        } finally {
+            in.close();
+        }
```
Azquelt: Use try-with-resources

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/suite/KafkaTests.java
```diff
@@ -24,13 +24,17 @@
...
-@SuiteClasses({
-                KafkaNackTest.class,
-                KafkaValidationTests.class
-})
+@SuiteClasses({ 
+	MetricsTest.class, 
+	KafkaNackTest.class,
+    MultiAppMetricsTest.class,
+    KafkaValidationTests.class 
+	})
```
Azquelt: Fix inconsistent indentation.

Eclipse should be set to autoformat on save so I'm not sure how this happened.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MultiAppMetricsTestServletTwo.java
```diff
@@ -0,0 +1,53 @@
...
+        try (KafkaReader<String, String> reader = kafkaTestClient.readerFor(OUTGOING_TOPIC_NAME);
+                        KafkaWriter<String, String> writer = kafkaTestClient.writerFor(INCOMING_TOPIC_NAME)) {
+
+            writer.sendMessage("abc");
+            writer.sendMessage("xyz");
+            List<String> msgs = reader.assertReadMessages(2, KafkaTestConstants.DEFAULT_KAFKA_TIMEOUT);
+
+            assertThat(msgs, contains("cba", "zyx"));
+        }
+    }
```
Azquelt, line 52: This test servlet and the one next to it don't do anything that needs to happen on the server. It would probably have been simpler to do all this on the client side.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,154 @@
...
+    public static void teardown() throws Exception {
+        try {
+            server.stopServer();
+        } finally {
+            KafkaUtils.deleteKafkaTopics(KafkaTests.getAdminClient());
+        }
+    }
+
+    public static HashMap<String, Integer> getMetrics(LibertyServer server) throws Exception {
+        // Returns the MicroProfile Reactive Messaging metrics from the /metrics endpoint
```
Azquelt: This method should have more detailed Javadoc about what it's trying to do. The Javadoc should say what it returns and where it gets the information from.

I _think_, based on reading the code, that it's trying to extract the message counts for each channel. It does this by retrieving the `/metrics` endpoint and looking for lines that look like this:

```
mp_messaging_message_count_total{channel="channelname",mp_scope="base"} 7
```

and adding an entry to the result: `channelname -> 7`.

Azquelt (dismissed):
Several suggestions for minor fixes or improvements but nothing essential.

ryan-storey:
#build #libby

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -125,15 +125,14 @@ public static void teardown() throws Exception {
         }
     }
 
-    private HashMap<String, Integer> getMetrics() throws Exception {
```
abutch3r: Why is this now static and why are you now passing in the server? The server is a static entity within the class?

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
```diff
@@ -0,0 +1,151 @@
...
+    @AfterClass
+    public static void teardown() throws Exception {
+        try {
+            server.stopServer();
+        } finally {
+            KafkaUtils.deleteKafkaTopics(KafkaTests.getAdminClient());
+        }
+    }
+
+    public static HashMap<String, Integer> getMetrics(LibertyServer server) throws Exception {
```
abutch3r: Why is this now `public static` and why are you passing in the Liberty Server. the server is a static entity within the class so you should have no need to pass it to getMetrics

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
ryan-storey: It's static so that we can reuse the method in the MultiApp test cases. Good spot on the server though, I had not noticed that

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
ryan-storey: It's static so that we can reuse the method in the MultiApp test cases. Good spot on the server though, I had not noticed that

abutch3r (changes requested):
The changes to retrieve the metrics cannot go in as is

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsTest.java
abutch3r: If this is to be common across tests, it should be in its own common class, not in one test class and then reused by another. while we only have two tests at the moment here, if we ever expand this then it is likely to start to cause issues. So it being static in the common class taking the server as its argument is correct, this could lead to some unintended issues, even replacing server with URL

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import java.net.URL;
+import java.util.HashMap;
+
+import componenttest.topology.utils.HttpUtils;
+import componenttest.topology.utils.HttpUtils.HTTPRequestMethod;
+
+/**
+ *
+ */
+public class MetricsUtils {
```
abutch3r: This looks good, I would recommend including this in the `io.openliberty.microprofile.reactive.messaging.internal_fat.common` project given we already have common classes in there, so I would put this in a new `com.ibm.ws.microprofile.reactive.messaging.fat.metrics` Just in case we need to do anything with metrics in a new fat project in the future

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/metrics/MetricsUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static HashMap<String, Integer> getMetrics(URL url) throws Exception {
+        // Returns the MicroProfile Reactive Messaging metrics from the /metrics endpoint
+        HttpURLConnection response = HttpUtils.getHttpConnection(url, 200, null, 30, HTTPRequestMethod.GET, null, null);
```
Azquelt: Add Javadoc to this method. This is extra important when the method is in a utility class - you're expecting people to reuse this method, so you need to tell them what it's meant to do.

In particular, this method is called `getMetrics()`, but it doesn't return all metrics, and it doesn't return generic data about the metrics. It's specifically finding the metrics for the message counts for each channel. A more descriptive name _might_ be appropriate - I'm a bit on the fence on this one because it does return all the metrics which matter to the tests which are using it - but Javadoc explaining what it actually does and returns is definitely needed.

ryan-storey:
#build #libby

ryan-storey:
#build #libby

abutch3r (approved):

PR #26643, opened 2023-10-18
Publish a dist zip file per project to be extracted into the build.image project

fritze2:
- Currently, the building of open-liberty's image is coupled to compiling all 1000+ projects in the workspace. That causes very slow builds... up to 5 hours on EBC machines.
- Creating an image zip could be faster when only few projects have differences. We could package an image independently of having projects built in the workspace if the continuous builds uploaded what projects normally output into build.image.
- This PR generates a distribution zip file for each project that would copy files into build.image. Then instead of copying files from projects into build.image, the build for the build.image project will loop over all projects that do generate a distribution, and will extract that zip into build.image.
- This enables us to generate an image from distribution zips that could be located locally (when they are built on the machine) or on Artifactory (when they're pulled from a recent green continuous build).

### DIFF
--- a/dev/ant_build.js/bnd.bnd
+++ b/dev/ant_build.js/bnd.bnd
@@ -15,3 +15,5 @@
 -nobundles=true
 
 instrument.disabled: true
+
+publish.wlp.jar.disabled: true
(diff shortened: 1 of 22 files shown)


fritze2:
#pbbeta

Issue #26670, opened 2023-10-19
Remove excessive trace associated with SipHpelLogExtension


TiborBeres:
SipHpelLogExtension (on by default) is putting extensive amount of tracing into the server logs when full SIP tracing is enabled. We should reduce/eliminate tracing associated with this extension as it is simply not needed when debugging SIP issues. We can restore them later if needed.

PR #26671, opened 2023-10-19
changed tracepoints in SipLogExtension, SipApplicationSessionImpl

TiborBeres:
There is extensive tracing triggered by the SipHpelLogExtension, by default turned on when full SIP tracing is enabled. This level of trace is unnecessary and in most cases it eve hinders debugging. We need to change the tracing in order to improve the process of debugging SIP related issues.


TiborBeres:
#build

epj (approved):
makes sense

TiborBeres:
fixes #26670

PR #26679, opened 2023-10-20
mixed JPA entities on a repository

njr-11:
Add coverage for scenarios where:
 - a repository has no primary entity type and only has find methods that return various different entity classes
 - a repository that has lifecycle methods to save entities with different entity classes
 - a repository method that infers its primary entity type from the return type of find methods, which is possible because the find methods only return a single type of entity class.


njr-11:
#build

mswatosh (approved):

njr-11:
#build

PR #26684, opened 2023-10-22
set thread context metadata for startup event

### LINKED ISSUE #26683: Component metadata is not present during CDI Startup events
**Describe the bug**  
CDI does not set the component metadata or thread context classloader before calling observer methods annotated @Startup this can cause problems if application code running inside one of those observer methods depends on this data. E.G. JNDI

Example stack:

	at io.openliberty.cdi40.internal.fat.startupEvents.sharedLib.AbstractObserver.runAllJNDITests(AbstractObserver.java:147)
	at io.openliberty.cdi40.internal.fat.startupEvents.sharedLib.AbstractObserver.observeStartup(AbstractObserver.java:75)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.jboss.weld.injection.StaticMethodInjectionPoint.invoke(StaticMethodInjectionPoint.java:95)
	at org.jboss.weld.injection.StaticMethodInjectionPoint.invoke(StaticMethodInjectionPoint.java:85)
	at org.jboss.weld.injection.MethodInvocationStrategy$SimpleMethodInvocationStrategy.invoke(MethodInvocationStrategy.java:168)
	at org.jboss.weld.event.ObserverMethodImpl.sendEvent(ObserverMethodImpl.java:330)
	at org.jboss.weld.event.ObserverMethodImpl.sendEvent(ObserverMethodImpl.java:308)
	at org.jboss.weld.event.ObserverMethodImpl.notify(ObserverMethodImpl.java:286)
	at jakarta.enterprise.inject.spi.ObserverMethod.notify(ObserverMethod.java:142)
	at org.jboss.weld.util.Observers.notify(Observers.java:166)
	at org.jboss.weld.event.ObserverNotifier.notifySyncObservers(ObserverNotifier.java:285)
	at org.jboss.weld.event.ObserverNotifier.notify(ObserverNotifier.java:273)
	at org.jboss.weld.event.ObserverNotifier.fireEvent(ObserverNotifier.java:177)
	at org.jboss.weld.bootstrap.BeanDeploymentModule.fireEvent(BeanDeploymentModule.java:93)
	at org.jboss.weld.bootstrap.WeldStartup.endInitialization(WeldStartup.java:578)
	at org.jboss.weld.bootstrap.WeldBootstrap.endInitialization(WeldBootstrap.java:101)
	at com.ibm.ws.cdi.impl.CDIContainerImpl.endInitialization(CDIContainerImpl.java:208)
	at com.ibm.ws.cdi.liberty.CDIRuntimeImpl.applicationStarting(CDIRuntimeImpl.java:481)
	at com.ibm.ws.container.service.state.internal.ApplicationStateManager.fireStarting(ApplicationStateManager.java:53)
	at com.ibm.ws.container.service.state.internal.StateChangeServiceImpl.fireApplicationStarting(StateChangeServiceImpl.java:52)
	at com.ibm.ws.app.manager.module.internal.SimpleDeployedAppInfoBase.preDeployApp(SimpleDeployedAppInfoBase.java:549)
	at com.ibm.ws.app.manager.module.internal.SimpleDeployedAppInfoBase.installApp(SimpleDeployedAppInfoBase.java:510)
	at com.ibm.ws.app.manager.module.internal.DeployedAppInfoBase.deployApp(DeployedAppInfoBase.java:351)
	at com.ibm.ws.app.manager.ear.internal.EARApplicationHandlerImpl.install(EARApplicationHandlerImpl.java:79)
	at com.ibm.ws.app.manager.internal.statemachine.StartAction.execute(StartAction.java:184)
	at com.ibm.ws.app.manager.internal.statemachine.ApplicationStateMachineImpl.enterState(ApplicationStateMachineImpl.java:1369)
	at com.ibm.ws.app.manager.internal.statemachine.ApplicationStateMachineImpl.run(ApplicationStateMachineImpl.java:912)
	at com.ibm.ws.threading.internal.ExecutorServiceImpl$RunnableWrapper.run(ExecutorServiceImpl.java:247)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:858)

**Steps to Reproduce**  
Create a CDI extension or bean with a method that observes the Startup event like so:

`public void onStartup(@Observes Startup startupEvent)`

Inside this method perform a task that requires either the thread context classlodaer or module metadata, E.G.

`new InitialContext().lookup("java:comp/BeanManager");`

**Expected behavior**  
The thread context classloader and module metadata are available and used correctly.

**Diagnostic information:**  
 - OpenLiberty Version: [e.g. 21.0.0.8 - 21.0.0.10]
 - Affected feature(s) [e.g. cdi-4.0]

benjamin-confino:
fixes #26683

### DIFF
--- a/dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
+++ b/dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
 unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
+
+no.bda.for.module.CWOWB1019E=CWOWB1019E: WebSphere Application Server internal error occurred. Please contact WebSphere Application Server support with the following data: {0} {1}
+no.bda.for.module.CWOWB1019E.explanation=CMSG0001
+no.bda.for.module.CWOWB1019E.useraction=See the problem determination information on the WebSphere Application Server Support web page: http://www.ibm.com/software/webservers/appserv/was/support.
--- a/dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
+++ b/dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
@@ -105,6 +105,6 @@ unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI bea
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
 
-no.bda.for.module.CWOWB1019E=CWOWB1019E: WebSphere Application Server internal error occurred. Please contact WebSphere Application Server support with the following data: {0} {1}
+no.bda.for.module.CWOWB1019E=CWOWB1019E: WebSphere Application Server internal error occurred. Contact WebSphere Application Server support with the following data: {0} {1}
 no.bda.for.module.CWOWB1019E.explanation=CMSG0001
 no.bda.for.module.CWOWB1019E.useraction=See the problem determination information on the WebSphere Application Server Support web page: http://www.ibm.com/software/webservers/appserv/was/support.
--- a/dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ApplicationImpl.java
+++ b/dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ApplicationImpl.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2016 IBM Corporation and others.
+ * Copyright (c) 2015, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -52,6 +52,7 @@ public class ApplicationImpl implements Application {
     private final ApplicationType type;
     private final Collection<CDIArchive> moduleArchives;
     private final Collection<CDIArchive> libraryArchives;
+    private ClassLoader tccl = null; //A stored reference to the ClassLoader after it has been transformed into a TCCL.
 
     public ApplicationImpl(ApplicationInfo applicationInfo, RuntimeFactory factory) throws CDIException {
         this.applicationInfo = (ExtendedApplicationInfo) applicationInfo;
@@ -232,5 +233,17 @@ public class ApplicationImpl implements Application {
 
     public ExtendedApplicationInfo getApplicationInfo() {
         return applicationInfo;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void setTCCL(ClassLoader tccl) {
+        this.tccl = tccl;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public ClassLoader getTCCL() {
+        return tccl;
     }
 }
--- a/dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/CDIArchiveImpl.java
+++ b/dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/CDIArchiveImpl.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2018 IBM Corporation and others.
+ * Copyright (c) 2015, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -32,6 +32,9 @@ import com.ibm.ws.cdi.internal.interfaces.CDIArchive;
 import com.ibm.ws.cdi.internal.interfaces.CDIUtils;
 import com.ibm.ws.cdi.internal.interfaces.Resource;
 import com.ibm.ws.cdi.internal.interfaces.ResourceInjectionBag;
+import com.ibm.ws.container.service.annocache.Annotations;
+import com.ibm.ws.container.service.annocache.AnnotationsBetaHelper;
+import com.ibm.ws.container.service.annocache.CDIContainerAnnotations;
 import com.ibm.ws.container.service.app.deploy.ApplicationInfo;
 import com.ibm.ws.container.service.app.deploy.ClientModuleInfo;
 import com.ibm.ws.container.service.app.deploy.ContainerInfo;
@@ -39,7 +42,6 @@ import com.ibm.ws.container.service.app.deploy.ContainerInfo.Type;
 import com.ibm.ws.container.service.app.deploy.InjectionClassList;
 import com.ibm.ws.container.service.app.deploy.ModuleClassesContainerInfo;
 import com.ibm.ws.container.service.app.deploy.ModuleInfo;
-import com.ibm.ws.container.service.app.deploy.WebModuleInfo;
 import com.ibm.ws.container.service.app.deploy.extended.ExtendedModuleInfo;
 import com.ibm.ws.container.service.app.deploy.extended.ModuleContainerInfo;
 import com.ibm.ws.injectionengine.osgi.util.OSGiJNDIEnvironmentRefBindingHelper;
@@ -53,10 +55,6 @@ import com.ibm.wsspi.adaptable.module.Container;
 import com.ibm.wsspi.adaptable.module.Entry;
 import com.ibm.wsspi.adaptable.module.NonPersistentCache;
 import com.ibm.wsspi.adaptable.module.UnableToAdaptException;
-
-import com.ibm.ws.container.service.annocache.Annotations;
-import com.ibm.ws.container.service.annocache.AnnotationsBetaHelper;
-import com.ibm.ws.container.service.annocache.CDIContainerAnnotations;
 import com.ibm.wsspi.annocache.classsource.ClassSource_Factory;
 
 public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
@@ -83,15 +81,15 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
     // As module library archives:
     //
     // com.ibm.ws.cdi.internal.archive.liberty.CDIArchiveImpl.initModuleLibraryArchives()
-    
-    public CDIArchiveImpl(
-        ApplicationImpl application,
-        ContainerInfo containerInfo,
-        ArchiveType archiveType,
-        ClassLoader classLoader,
-        RuntimeFactory factory) {
 
-        super( containerInfo.getName(), factory.getServices() );
+    public CDIArchiveImpl(
+                          ApplicationImpl application,
+                          ContainerInfo containerInfo,
+                          ArchiveType archiveType,
+                          ClassLoader classLoader,
+                          RuntimeFactory factory) {
+
+        super(containerInfo.getName(), factory.getServices());
 
         this.factory = factory;
 
@@ -131,14 +129,14 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
     private Container getContainer(Entry entry) {
         try {
             return entry.adapt(Container.class); // throws UnableToAdaptException
-        } catch ( Throwable th ) {
+        } catch (Throwable th) {
             return null; // FFDC
         }
     }
 
     private Container getContainer(Container container, String path) {
         Entry startEntry = container.getEntry(path);
-        if ( startEntry == null ) {
+        if (startEntry == null) {
             return null;
         } else {
             return getContainer(startEntry);
@@ -160,8 +158,8 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
 
     @Override
     public String getPath() throws CDIException {
-        if ( path == null ) {
-            path = getPath( getContainer() );
+        if (path == null) {
+            path = getPath(getContainer());
         }
         return path;
     }
@@ -171,12 +169,12 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
 
         try {
             Entry entry;
-            while ( (entry = useContainer.adapt(Entry.class)) != null ) {
+            while ((entry = useContainer.adapt(Entry.class)) != null) {
                 // 'adapt' throws UnableToAdaptException
-                pathBuilder.insert(0,  entry.getPath() );
+                pathBuilder.insert(0, entry.getPath());
                 useContainer = entry.getRoot();
             }
-        } catch ( UnableToAdaptException e ) {
+        } catch (UnableToAdaptException e) {
             // FFDC
             throw new CDIException(e);
         }
@@ -192,7 +190,7 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
 
     @Override
     public boolean isModule() {
-        return ( containerInfo instanceof ModuleContainerInfo );
+        return (containerInfo instanceof ModuleContainerInfo);
     }
 
     //
@@ -225,15 +223,15 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
     private Collection<CDIArchive> initModuleLibraryArchives() {
         Map<ContainerInfo, CDIArchive> moduleLibraryArchives = new HashMap<ContainerInfo, CDIArchive>();
 
-        if ( containerInfo instanceof ModuleClassesContainerInfo ) {
+        if (containerInfo instanceof ModuleClassesContainerInfo) {
             ModuleClassesContainerInfo moduleClassesContainerInfo = (ModuleClassesContainerInfo) containerInfo;
             List<ContainerInfo> containerInfos = moduleClassesContainerInfo.getClassesContainerInfo();
-            for ( ContainerInfo containerInfo : containerInfos ) {
+            for (ContainerInfo containerInfo : containerInfos) {
                 Type containerType = containerInfo.getType();
 
-                if ( (containerType == Type.WEB_INF_LIB) ||
-                     (containerType == Type.MANIFEST_CLASSPATH) ||
-                     (containerType == Type.JAR_MODULE) ) {
+                if ((containerType == Type.WEB_INF_LIB) ||
+                    (containerType == Type.MANIFEST_CLASSPATH) ||
+                    (containerType == Type.JAR_MODULE)) {
                     ArchiveType archiveType = ContainerInfoTypeUtils.getType(containerType);
                     CDIArchive childArchive = factory.newArchive(application, containerInfo, archiveType, classLoader);
                     moduleLibraryArchives.put(containerInfo, childArchive);
@@ -250,18 +248,18 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
     private String appCallbackHandlerName;
 
     public ExtendedModuleInfo getModuleInfo() throws CDIException {
-        if ( !isModule() ) {
+        if (!isModule()) {
             return null;
         }
 
-        if ( extendedModuleInfo == null ) {
+        if (extendedModuleInfo == null) {
             NonPersistentCache moduleCache;
 
             ModuleContainerInfo moduleContainerInfo = (ModuleContainerInfo) containerInfo;
             Container moduleContainer = moduleContainerInfo.getContainer();
             try {
                 moduleCache = moduleContainer.adapt(NonPersistentCache.class);
-            } catch ( UnableToAdaptException e ) {
+            } catch (UnableToAdaptException e) {
                 // FFDC
                 throw new CDIException(e);
             }
@@ -274,25 +272,26 @@ public class CDIArchiveImpl extends AbstractCDIArchive implements CDIArchive {
 
     private ClientModuleInfo getClientModuleInfo() throws CDIException {
         ModuleInfo moduleInfo = getModuleInfo(); // throws CDIException
-        if ( !(moduleInfo instanceof ClientModuleInfo) ) {
+        if (!(moduleInfo instanceof ClientModuleInfo)) {
             return null;
         } else {
             return (ClientModuleInfo) moduleInfo;
         }
     }
-    
+
     public ModuleMetaData getModuleMetaData() throws CDIException {
         ExtendedModuleInfo useModuleInfo = getModuleInfo();
-        return ( (useModuleInfo == null) ? null : useModuleInfo.getMetaData() );
+        return ((useModuleInfo == null) ? null : useModuleInfo.getMetaData());
     }
 
     @Override
     public MetaData getMetaData() throws CDIException {
-        if ( isModule() ) {
+        if (isModule()) {
             return getModuleMetaData();
-        } else {
-            return application.getApplicationMetaData();
+        } else if (application == null) {
+            throw new IllegalArgumentException("getMetaData was invoked on a non-application archive"); //Putting this here rather than above because I'm not sure if anyone depended on the incorrect behaviour for modules, but if they got this far they'd NPE.
         }
+        return application.getApplicationMetaData();
     }
 
     @Override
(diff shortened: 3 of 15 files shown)


benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
I don't think the failures are related but double checking.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/ContextBeginnerEnderImpl.java
```diff
@@ -0,0 +1,117 @@
...
+import com.ibm.ws.runtime.metadata.ModuleMetaData;
+import com.ibm.ws.threadContext.ComponentMetaDataAccessorImpl;
+
+public class ContextBeginnerEnderImpl implements ContextBeginnerEnder {
+
+    private ClassLoader newTheadContextClassLoader = null;
+    private ClassLoader oldTheadContextClassLoader = null;
+    private JndiHelperComponentMetaData cmd = null;
+    private final boolean began = false;
+
```
Azquelt: There are several things I don't like about this class.

1. The scope is wrong: this shouldn't be a singleton that's stored in `CDIRuntime` because it holds state which relates to the latest extracted and applied context. Therefore you should be making a new object every time you extract context.
2. Two different jobs are conflated. This object holds both the new TCCL (i.e. the one to apply) and the old TCCL (i.e. the one to restore when the context is ended). This should be two classes, one which olds the new TCCL, created when context is extracted, and one which holds old TCCL, created when context is applied (and which can be used to remove the context and restore the original context).
    - The second class should implement `AutoClosable` and where possible you should call it with a try-with-resources block, ensuring that the context applied at the start of the block is always removed at the end.

Azquelt (changes requested):

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
```diff
@@ -224,7 +221,26 @@ public void applicationStarted(Application application) throws CDIException {
...
                         if (!module.isWebModule()) {
-                            eventManager.fireStartupEvent(module);
+                            ContextBeginnerEnder contextBeginnerEnder = cdiRuntime.getContextBeginnerEnder();
+                            String id = module.getId();
+                            try {
+                                WebSphereBeanDeploymentArchive bda = deployment.getBeanDeploymentArchive(id);
+                                if (bda != null) {
+                                    if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                                        Tr.debug(tc, "applicationStarted", "Setting context for " + id);
+                                    }
```
Azquelt: See if we can move this debug into `ContextBeginnerEnder` (might not be able to)

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/ContextBeginnerEnderImpl.java
Azquelt: Ben pointed out that it's not actually a singleton, the getter is actually a factory method so we could just rename it from `get` to `create` so that's clear.

We agreed it's not necessary to have two separate objects as long as you are only using the object to apply context once - which is currently guarded against - but that implementing `AutoClosable` and providing a usage example on the interface would be good.

dev/com.ibm.ws.cdi.ejb_fat/bnd.bnd.orig
```diff
@@ -0,0 +1,78 @@
...
+# Contributors:
+#     IBM Corporation - initial API and implementation
+#*******************************************************************************
+-include= ~../cnf/resources/bnd/bundle.props
+bVersion=1.0
+
+# Define the bundle for this FAT
+
+src: \
+	fat/src
```
Azquelt: File should be removed.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/ApplicationImpl.java
```diff
@@ -233,4 +234,16 @@ public boolean getUseJandex() {
...
+    @Override
+    public void setTCCL(ClassLoader tccl) {
+        this.tccl = tccl;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public ClassLoader getTCCL() {
+        return tccl;
+    }
```
Azquelt, line 248: Copyright date needs updating.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/Application.java
```diff
@@ -83,9 +83,19 @@ public interface Application {
...
 
+    /**
+     * Register the thread context class loader associated with this app
+     */
+    void setTCCL(ClassLoader tccl);
+
+    /**
+     * Get the thread context class loader associated with this app
+     */
+    ClassLoader getTCCL();
```
Azquelt: Can this method return `null`? If so then this should be documented.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/CDIRuntime.java
```diff
@@ -105,14 +105,10 @@ public interface CDIRuntime extends CDIService {
...
+     * @return true if a context was begun, otherwise false.
      * @throws CDIException
      */
-    void beginContext(CDIArchive archive) throws CDIException;
-
-    /**
-     *
-     */
-    void endContext();
+    ContextBeginnerEnder createContextBeginnerEnder();
```
Azquelt: Javadoc needs to be corrected.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/ContextBeginnerEnder.java
```diff
@@ -0,0 +1,81 @@
...
+ * You should not use a second ContextBeginnerEnder before closing the first one, nor should you modify any of the context handled by this object before you
+ * have finished with it.
+ * 
+ * Example of use:
+ * 
+ * try (ContextBeginnerEnder contextBeginnerEnder = cdiRuntime.createContextBeginnerEnder().extractComponentMetaData(bda.getArchive()).extractTCCL(application).beginContext()) {
+ *     eventManager.fireStartupEvent(module);
+ * }
+ */
+public interface ContextBeginnerEnder extends AutoCloseable {
```
Azquelt: Please add formatting to the Javadoc:
- `<p>` between each paragraph
- `<pre>` around the code example

Although we don't render HTML Javadoc pages, most IDEs will mush everything together unless you include the formatting tags.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
```diff
@@ -224,7 +221,15 @@ public void applicationStarted(Application application) throws CDIException {
...
-                            eventManager.fireStartupEvent(module);
+                            String id = module.getId();
+                            WebSphereBeanDeploymentArchive bda = deployment.getBeanDeploymentArchive(id);
+                            if (bda != null) {
+                                try (ContextBeginnerEnder contextBeginnerEnder = cdiRuntime.createContextBeginnerEnder().extractComponentMetaData(bda.getArchive()).extractTCCL(application).beginContext()) {
+                                    eventManager.fireStartupEvent(module);
+                                }
+                            } else if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                                Tr.debug(tc, "applicationStarted", "Could not set context for " + id);
+                            }
```
Azquelt: This change means that we will not fire the startup event if the bda was not found.

That seems like it warrants more action that just a debug message. Possibly even an exception to halt the app start.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/CDIRuntimeImpl.java
```diff
@@ -474,24 +468,11 @@ public void applicationStarting(ApplicationInfo appInfo) throws StateChangeExcep
...
-            //Origonally I tried to setup JNDI so only application metadata was on the thread but
-            //that didn't work so I use give classic utils one of the module archives.
-
-            if (application.getModuleArchives().size() > 0 &&
-                application.getApplicationMetaData() != null) {
-                CDIArchive archive = application.getModuleArchives().iterator().next();
-                beginContext(archive);
-                setContext = true;
-            }
+            contextBeginnerEnder.extractComponentMetaData(application).extractTCCL(application).beginContext();
```
Azquelt: This should be in a try-with-resources. At the moment it's missing a `close()`.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/CDIRuntimeImpl.java
```diff
@@ -608,23 +582,14 @@ public ProxyServices getProxyServices() {
...
+        ComponentMetaDataAccessorImpl accessor = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor();
+        accessor.endContext();
 
+        if (application != null) {
+            classLoadingSRRef.getServiceWithException().destroyThreadContextClassLoader(application.getTCCL());
         }
-
-        ComponentMetaDataAccessorImpl accessor = ComponentMetaDataAccessorImpl.getComponentMetaDataAccessor();
-        accessor.beginContext(cmd);
     }
```
Azquelt: Should this new `endContext()` method be removed? That logic is in `ContextBeginnerEnder`, right?

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/ContextBeginnerEnderImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public void close() {
+
+        if (!began) {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(tc, "close", "close invoked without beginContext");
+            }
+        }
```
Azquelt: Given this means we have a liberty bug, this should throw `IllegalStateException`.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/ContextBeginnerEnderImpl.java
```diff
@@ -0,0 +1,131 @@
...
+
+            //Because weld fires observes in all modules when endInitialization() is called
+            //We can only set the jndi context once. This is sufficent for the java:app namespace
+            //but not for the java module namespace.
+
+            //Origonally I tried to setup JNDI so only application metadata was on the thread but
+            //that didn't work so I use give classic utils one of the module archives.
+
+            ModuleMetaData moduleMetaData = (ModuleMetaData) metaData;
+            cmd = new JndiHelperComponentMetaData(moduleMetaData);
```
Azquelt: I don't understand this comment here.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/ContextBeginnerEnderImpl.java
```diff
@@ -0,0 +1,131 @@
...
+
+        return this;
+    }
+
+    @Override
+    public ContextBeginnerEnder beginContext() {
+        if (began) {
+            throw new IllegalStateException("beginContext cannot be called twice");
+        }
+        boolean begun = true;
```
Azquelt: Nothing sets `this.began`.

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/CDIArchiveImpl.java
```diff
@@ -274,57 +272,58 @@ public ExtendedModuleInfo getModuleInfo() throws CDIException {
...
 
     @Override
     public MetaData getMetaData() throws CDIException {
-        if ( isModule() ) {
+        if (isModule()) {
             return getModuleMetaData();
-        } else {
+        } else if (application != null) {
             return application.getApplicationMetaData();
         }
```
Azquelt: The Javadoc for this method indicates that it must not be called for non-application archives.

Looking at the rest of the class, `application` should only be `null` for non-application archives, right?

Maybe an `OperationNotSupportedException` if `application` is `null` would be more appropriate?

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/Application.java
Azquelt: Copyright date needs updating

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/CDIArchiveImpl.java
```diff
@@ -274,57 +272,58 @@ public ExtendedModuleInfo getModuleInfo() throws CDIException {
...
+        return null;
     }
```
Azquelt: Copyright date needs updating

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/weld/injection/EEValidationUtils.java
```diff
@@ -164,8 +165,8 @@ private static void validateJndiLookup(String lookupString, Annotated annotated,
         }
         // We need to set a current component before doing anything to do with JNDI
         CDIRuntime cdiRuntime = cdiArchive.getCDIRuntime();
-        try {
-            cdiRuntime.beginContext(cdiArchive);
+
+        try (ContextBeginnerEnder cbe = cdiRuntime.createContextBeginnerEnder().extractComponentMetaData(cdiArchive).beginContext()) {
```
Azquelt, line 169: Copyright date needs updating

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/JndiHelperModuleMetaData.java
```diff
@@ -27,14 +27,15 @@ public class JndiHelperModuleMetaData implements ModuleMetaData {
...
 
     @Override
-    public void setMetaData(MetaDataSlot slot, Object metadata) {}
+    public void setMetaData(MetaDataSlot slot, Object metadata) {
+    }
 
     @Override
-    public void release() {}
+    public void release() {
+    }
```
Azquelt: No real change, please remove this file from the PR.

Azquelt (changes requested):

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/interfaces/Application.java
benjamin-confino: It can if the app has no modules. E.G. an OSGi app

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/CDIRuntimeImpl.java
benjamin-confino: It is in a try-with-resources. `contextBeginnerEnder` is created within the try statement, and its contents are filled in here after `application` has been created.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/CDIRuntimeImpl.java
benjamin-confino: Yeah

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/ContextBeginnerEnderImpl.java
benjamin-confino: Oooh. Good catch!

dev/com.ibm.ws.cdi.internal/src/com/ibm/ws/cdi/internal/archive/liberty/CDIArchiveImpl.java
benjamin-confino: It should be safe, the old behaviour was an NPE so we know nobody relies on it.

benjamin-confino: I think that should be safe. Assuming all non-application archives are also not modules they would NPE before so an exception now won't break anything.

benjamin-confino: Looking at containerInfo I can't make that assumption without digging into it. But I can put the new exception where the old NPE was.

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/liberty/CDIRuntimeImpl.java
Azquelt: The pattern with try-with-resources is that you open a resource (i.e. apply the context) in the try() and it gets closed (i.e. the context is removed) at the end of the block. Applying the context in the middle like this is horrible.

Splitting `ContextBeginnerEnder` into two objects as I originally suggested would prevent this because the `AutoClosable` object would only be created when `beginContext()` is called.

In addition, if an exception is thrown before `beginContext()` is called, `close()` will also throw an exception.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
```diff
@@ -224,7 +221,15 @@ public void applicationStarted(Application application) throws CDIException {
                 if (modules != null) {
                     for (BeanDeploymentModule module : modules) {
                         if (!module.isWebModule()) {
-                            eventManager.fireStartupEvent(module);
+                            String id = module.getId();
+                            WebSphereBeanDeploymentArchive bda = deployment.getBeanDeploymentArchive(id);
```
Azquelt, line 225: Do we have any guarantee that the BDA for the module has the same ID as the `BeanDeploymentModule`?

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
 unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
+
+no.bda.for.module.CWOWB1019E=CWOWB1019E: There was no bean definition archive for module {0}
+no.bda.for.module.CWOWB1019E.explanation=Every module in a CDI enabled application must have a bean definition archive to store CDI's knowledge of that modules contests. However a module was missing a bean definition archive.
+no.bda.for.module.CWOWB1019E.useraction=Review logs to identify problems in application startup which may cause this issue.
```
Azquelt: This message talks about things that we don't expose to users. We should probably just use a generic "A CDI internal error occurred while doing X" type error message, with the exception appended.

Azquelt:
If possible, we should have some tests for this. There are existing tests for startup events so adding something which tests for the correct context there would be enough.

dev/com.ibm.ws.cdi.weld/src/com/ibm/ws/cdi/impl/CDIContainerImpl.java
benjamin-confino: The ID in the BeanDeploymentModule and the BDA are both built from the same String

BDAFactory.java line 84
`EEModuleDescriptor eeModuleDescriptor = new WebSphereEEModuleDescriptor(archiveID, extensionArchive.getType());`

This sets the EEModuleDescriptor's id to archiveID And eventually the BeanDeploymentModule gets its ID from the EEModuleDescriptor's id.

`module = new BeanDeploymentModule(descriptor.getId()` in weld's class BeanDeploymentModules

Then a few lines later in BDAFactory.java we pass `archiveID` into the constructor BeanDeploymentModule where it becomes the ID. 

The exception is for library modules in `CDIContainerImpl.processModuleLibraries` where the module descriptor will have the ID of the parent. So for those we'll be setting our TCCL and CMD based on the library's parent.

benjamin-confino:
#build

benjamin-confino:
#build

benjamin-confino:
#build

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
 unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
+
+no.bda.for.module.CWOWB1019E=CWOWB1019E: WebSphere Application Server internal error occurred. Please contact WebSphere Application Server support with the following data: {0} {1}
+no.bda.for.module.CWOWB1019E.explanation=CMSG0001
+no.bda.for.module.CWOWB1019E.useraction=See the problem determination information on the WebSphere Application Server Support web page: http://www.ibm.com/software/webservers/appserv/was/support.
```
Azquelt: This message doesn't look great, but maybe it matches existing messages?

In this case it certainly is an internal error and not something the user should have to deal with.

benjamin-confino: Its a copy of CWOWB1003E, there are several identical messages CWOWB1003E - CWOWB1006E. Isn't that what we agreed on in the call?

Azquelt: We didn't say what it was a copy of on the call and I can't tell where you got it from when I'm looking at the diff which is why I asked.

The language here is contrary to the current message guidelines (likely because the place you copied it from is old) so it's not what I expected.

Azquelt: In other features, we just have one general "An internal exception occurred" message which is meant to have the exception appended.

In CDI, it looks like we've got a separate identical message for each different internal error case (as can be seen further up in the file) so, although it doesn't follow the current message guidelines, this message is consistent with existing messages for this component.

Azquelt (dismissed):

Azquelt:
#libby

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
 unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
+
+no.bda.for.module.CWOWB1019E=CWOWB1019E: WebSphere Application Server internal error occurred. Please contact WebSphere Application Server support with the following data: {0} {1}
```
jantley-ibm: Despite the fact that this mirrors several other CWOWB messages, it is against IBM style to use "please" in technical communication. https://ibmdocs-test.dcs.ibm.com/docs/en/ibm-style?topic=word-usage#p
Recommend changing the second sentence to:
Contact WebSphere Application Server support with the following data: {0} {1}

benjamin-confino:
#build

benjamin-confino:
#libby

Azquelt (approved):

dev/com.ibm.ws.cdi.internal/resources/com/ibm/ws/cdi/internal/resources/CDI.nlsprops
```diff
@@ -104,3 +104,7 @@ enableImplicitBeanArchives.conflict.CWOWB1017W.useraction=Update your server.xml
 unversioned.beansXML.warning.CWOWB1018W=CWOWB1018W: Using an unversioned CDI beans.xml file might lead to unexpected behaviour when using CDI 4.0 or later: {0}
 unversioned.beansXML.warning.CWOWB1018W.explanation=Prior to CDI 4.0, a non-empty but unversioned beans.xml file meant that the archive was treated as an explicit bean archive (all classes are eligible to be beans). In CDI 4.0 or later, the archive is taken to be an implicit bean archive (only annotated classes are eligible be beans).
 unversioned.beansXML.warning.CWOWB1018W.useraction=Update the beans.xml file within your application to use a newer version of the schema and ensure that the version and bean-discovery-mode attributes are properly set.
+
+no.bda.for.module.CWOWB1019E=CWOWB1019E: WebSphere Application Server internal error occurred. Contact WebSphere Application Server support with the following data: {0} {1}
+no.bda.for.module.CWOWB1019E.explanation=CMSG0001
+no.bda.for.module.CWOWB1019E.useraction=See the problem determination information on the WebSphere Application Server Support web page: http://www.ibm.com/software/webservers/appserv/was/support.
```
posmith, line 110: The no.bda.for.module.CWOWB1019E.useraction provides a link that redirects to the WAS product page (not support).  No "problem determination information" is found on that page.  Was this intended to link to a different page?

PR #26691, opened 2023-10-23
Test Kafka with post-process subscriber method

Azquelt:
Test receiving messages from Kafka using the connector and sending them on to a simple subscriber method which does post-process acknowledgements.


Azquelt:
Superseded by #26728

PR #26706, opened 2023-10-24
Initialize single instance of repository just once

njr-11:
This pull includes a couple of minor changes, along with a fix to ensure we are using a single instance of the repository implementation which is initialized just once.  The other changes optimize to ignore the static metamodel initialization when `@Generated` is present, and to rearrange a section of code and remove obsolete handling.


njr-11:
#build

KyleAure (approved):

PR #26707, opened 2023-10-24
remove async send timeToLive test

ngmr:
This test assumes that a JMS message that has exceeded its timeToLive value will not be returned from a MessageConsumer.receive() call.

Whilst the JMS specification says that this "should" happen, it explicitly states that this behaviour is _not_ guaranteed by the JMS specification.

Therefore it is not behaviour upon which a test using the JMS API can legitimately rely.


ngmr:
%build

ngmr:
#build

joe-chacko (approved):
About time!

PR #26712, opened 2023-10-24
Reduce trace in EJB Checkpoint FAT

tkburroughs:
Trace of EJB BackgroundLRUEvictionStrategy was interfering with the Checkpoint stub, so reduced EJBContainer trace to just trace EJB metadata processing that is required by the FAT.


tkburroughs:
#build

brideck (approved):

PR #26716, opened 2023-10-24
Fix Jakarta Data messaging issues

KyleAure:
The messages directory was incorrect, moved the messages to the correct directory. Otherwise, they are not included in the bundle.


njr-11 (approved):

KyleAure:
#build

PR #26728, opened 2023-10-25
Add fast.ack option to Kafka Reactive Messaging Connector

### LINKED ISSUE #26692: Kafka message acknowledgement is slow if messages are fed to a subscriber method
**Describe the bug**  
- A simple subscriber method with post-process acknowledgements will wait for an `ack()` to complete before accepting another message.
- The liberty kafka connector buffers acknowledgements to improve throughput.

These two behaviours mean that if messages from the kafka connector are routed to a simple subscriber method, either there's a delay between each message while the kafka connector waits to see if there are more messages to add to the batch, or the whole stream stalls because no more messages are requested, the kafka connector does not poll and the commit required to complete the `ack()` never occurs.

**Steps to Reproduce**  

Connect an incoming kafka channel to a method like this:

```java
@Incoming("myChannel")
public void myMethod(String data) {
}
```

**Expected behavior**  

Messages should be acknowledged quickly.

**Diagnostic information:**  
 - OpenLiberty Version: latest integration
 - Affected feature(s): `mpReactiveMessaging-1.0`

**Additional context**  

I raised a question about this behaviour previously in eclipse/microprofile-reactive-messaging#38. Despite the closed issue, I'm not sure what the resolution was supposed to be.

With the current behaviour, either:
- Connectors can't batch up acknowledgements
- Connectors have to lie and say that acknowledgement has been done when it hasn't

It doesn't look like smallrye reactive messaging has any option to change this behaviour.

This is a configuration which is demonstrated in the liberty reactive guide.

Azquelt:
When `fast.ack` is set to `true`, the Kafka connector will report that an acknowledgement has been completed as soon as it has received the acknowledgement. This is the default for `mpReactiveMessaging-3.0`.

When `fast.ack` is set to `false`, the Kafka connector will not report that acknowledgement has been completed until the partition offset has been successfully committed to the Kafka broker. There will usually be a delay before this happens because the Kafka connector attempts to batch up commits. In addition, if the partition offset cannot be committed, it will report that the acknowledgement failed. This can occur when another instance of the service joins the consumer group and some partitions are moved from one service to the other. This is the default for `mpReactiveMessaging-1.0` to preserve the existing behaviour.

We're adding this config property and changing the default behaviour for RM 3.0 because it works better with the way the Reactive Messaging container behaves. In particular, a subscriber method like this:

```java
@Incoming("myChannel")
public void myMethod(String payload) {}
```

which uses the `POST_PROCESS` acknowledgement mode will wait for each message to be acknowledged before it starts processing the next one. Since the Kafka connector tries to batch up acknowledgements, this results in a delay between the processing of each message.

Fixes #26692

### DIFF
--- a/dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/tck/KafkaPublisherVerification.java
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/tck/KafkaPublisherVerification.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2019, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -103,7 +103,7 @@ public class KafkaPublisherVerification extends PublisherVerification<Message<St
         PartitionTrackerFactory trackerFactory = new PartitionTrackerFactory();
         trackerFactory.setExecutor(executor);
         trackerFactory.setAutoCommitEnabled(false);
-        KafkaInput<String, String> kafkaInput = new KafkaInput<>(kafkaAdapterFactory, trackerFactory, kafkaConsumer, executor, topicName, 100);
+        KafkaInput<String, String> kafkaInput = new KafkaInput<>(kafkaAdapterFactory, trackerFactory, kafkaConsumer, executor, topicName, 100, false);
         kafkaInputs.add(kafkaInput);
         return kafkaInput.getPublisher().buildRs();
     }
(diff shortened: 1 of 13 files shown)


Azquelt:
#build

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/tck/KafkaPublisherVerification.java
```diff
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2023 IBM Corporation and others.
```
abutch3r: Was this only created this year, or should it be `2019, 2023`

abutch3r (commented):
Overall looks good. only minor point on copyright. 

Will need to look at the test failure in for the nack test case, given the changes here might have caused that, though that is not immediately obvious as to why, given that the connector is unable to communicate with kafka itself

Azquelt:
The nack failure is certainly odd. Looking at the trace, it looks like the `MessageAccess30` service was never registered, certainly `MessageAccessProvider.setMessageAccess` was never called.

I'll see if I can add some more OSGi trace to show components and services being registered to see if we can see why.

Azquelt:
For some reason, the changes to `dev/io.openliberty.microprofile.reactive.messaging.3.0.internal/bnd.bnd` are causing the component definition file for `MessageAccess30` not to be added to the bundle.

Azquelt:
Actually, all the classes are also missing from the bundle!

The addition of `Include-Resource` appears to disable the default behaviour of including all classes. I have to explicitly list them in `Private-Package` for them to be included.

Azquelt:
#libby
#build
#spawn.fullfat.buckets=com.ibm.ws.microprofile.reactive.messaging_fat,com.ibm.ws.microprofile.reactive.streams.operators_fat,io.openliberty.microprofile.reactive.messaging.internal_fat,com.ibm.ws.microprofile.reactive.messaging_fat_tck,com.ibm.ws.microprofile.reactive.streams.operators_fat_tck,io.openliberty.microprofile.reactive.messaging30.internal_fat_tck,io.openliberty.microprofile.reactive.streams.operators30.internal_fat_tck

FAT test is passing locally now. I've pushed up the copyright header fix and the fix to the bnd file.

abutch3r (approved):
Tests are now all passing and copyright issues resolved. So loog good to go

PR #26735, opened 2023-10-25
fix-ltpa-expiration

mshah0722:
Fix the logic behind the exception provided for expired validation keys and how we do not want to continue processing them. Adding in a null check to avoid NPE. Small file name change. Also, adding a FAT to ensure that the element works properly.


mshah0722:
#libby
#build
#spawn.fullfat.buckets=com.ibm.ws.security.token.ltpa_fat

utle (approved):
We need to fix these two // later 
processAllKeysFiles(createdFiles, modifiedFiles, deletedFiles); // we've got validation3.keys

 processValidationKeys(); //

Issue #26737, opened 2023-10-25
Improve Test Timing in ConnectionPoolStatsTest for abortConnections tests


brideck:
The abortConnections tests can fail intermittently due to the tight timings between connection creation and the calculation of the maxInUseIdleValue. We should update the tests to relax these timings and make failures a less likely occurrence.

PR #26738, opened 2023-10-25
26737: Relax ConnectionPoolStats abortConnections Test Timing

brideck:
Updating all 3 of the similarly structured tests to help make the sets of connections more distinct and to put the threshold used in the test more firmly in between the two sets. With the current setup, in local testing, I often see the last connection from set 1 and/or the first connection from set 2 end up dangerously close to the threshold. I've seen more wiggle room after making the proposed change locally.


brideck:
#build

jmstephensgit (approved):

brideck:
fixes #26737

PR #26741, opened 2023-10-26
replace liberty Update annotation

njr-11:
Because Jakarta Data introduced its own Update annotation that has a different purpose than the experimental Update annotation in Liberty, replace Liberty's with experimental parameter level annotations that fit with what Jakarta Data has.


njr-11:
#build

njr-11:
The test failures in the build are caused because I forgot to include the file that makes parameter names visible when running the data jpa test bucket.  Resubmitting now that it is added:
#build
#libby

jmstephensgit (approved):

Issue #26748, opened 2023-10-26
Transaction timeout failure in NextTimeoutServlet.testGetNextTimeoutWithActualExpiration


brideck:
The test waits for timers to expire for up to 4 minutes, but the transaction times out after just 2, which fails the test.

https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.ejbcontainer.timer.calendar_fat/test-applications/TimerCalTestWeb.war/src/com/ibm/ws/ejbcontainer/timer/cal/web/NextTimeoutServlet.java#L209

PR #26749, opened 2023-10-26
Prevent tran timeout in NextTimeoutServlet timer waits

brideck:
waitForTimer method should not use a transaction, so the test can wait for its full duration


brideck:
fixes #26748

brideck:
#build

tkburroughs (approved):

PR #26752, opened 2023-10-26
set useSeparateSessionInvalidatorThreadPool property to false in the metrics monitor FATS

Channyboy:
#build

Potential solution to  #24590.
With discussions with sessions team, it was identified that the intermittent failures with the Metric Monitor FATS where the session for the `/metrics` endpoint is being invalidated at it's first invocation may be a timing issue between the creation of the session and the background thread which cleans up timed out sessions.

During the creation process of a session during the first invocation of a servlet the time out attribute for the session is initially set at 0. Subsequently, the default of 30 minutes will be set for the session. It appears that the background process which cleans up expired sessions occured simultaneously as when the session was created. Therefore, the session, although it was in the process of being created, was identified as being expired and removed. The session was then considered invalidated.


pgunapal (approved):
LGTM

Issue #26768, opened 2023-10-30
InitialPollingTest.testRescheduleUnderConfigUpdate can fail on overloaded system


brideck:
The test system gets so stressed by all of the async work on the server that the test process can't read the log within the default timeout value.

tkburroughs:
fix was merged; closing

PR #26770, opened 2023-10-30
Double config update timeout for slow machines

brideck:
All of the concurrent activity on the server bogs things down so that log scraping doesn't happen fast enough on a slow system. Based on the times shown in failure logs, doubling the default should be more than enough. If that ends up not being the case, this can be increased more.


brideck:
#build

brideck:
fixes #26768

KyleAure (approved):

PR #26772, opened 2023-10-30
Clean up endpointSessionMap and httpSessionMap during http session expiration

### LINKED ISSUE #26771: Websocket Out of Memory Leak caused by Expired Sessions
**Describe the bug**  
A map within the websocket code is not properly cleaned up during session expiration and may result in a out-of-memory error.  Websocket sessions are fairly small, so it take a long time to notice.  However, depends on the number of websocket connections the server handles. 

**Steps to Reproduce**  
Open a websocket connection and let the session expire (default 30 minutes). If the websocket connection is closed manually, the session would be properly removed from the map. 

**Expected behavior**  
This map should be cleaned up. 

**Diagnostic information:**  
 - OpenLiberty Version: All 
 - Affected feature(s) websocket-1.0, websocket-1.1, websocket-2.0, websocket-2.1
 - Java Version: N/A
 - server.xml configuration: Session can be configured as so: 
 `<httpSession cookieHttpOnly="false" cookieName="JSESSIONID" invalidationTimeout="15m"/>` 

**Additional context**  
None

volosied:
fixes #26771

There are two leaks fixed in this PR. For the descriptions below, keep in mind there are two sessions -- the HTTP Session and the WebSocket session. 

**httpSessionMap**
- This main leak reported by the customer.
- This leak is caused by a http session id change.  The fix is to create the WsocHttpSessionIdListener and update the http session id key when id changes.
- This is used to track the http sessions to websocket sessions for when the former expire. Once expired, we'll close websocket session (see note below about the follow up)

**endpointSessionMap** 
- This is used to track the wsoc sessions to endpoints. 
- javax.websocket.Session#getOpenSessions basically forces to track the sessions for a particular endpoint. This means we need this object.

I don't like having two ArrayLists for the websocket sessions (see endpointSessionMap & httpSessionMap); however, I don't see a way around it.  We could merge the two, but code would be more complex and I prefer a simpler implementation here.

Follow up: The websocket close message is not sent once the session expired. Maybe only on secure connections per the code? Work / Investigation can be completed via https://github.com/OpenLiberty/open-liberty/issues/28021 

FAT:
- Our websocket FAT is not designed for this type of scenario, so I'd rather opt of of adding a FAT test (especially with other work prioritized). 
- Extra logging has been added to debug this further, if necessary (unlikely).  
- Test app 
  - Use a shorter session timeout (i.e ` <httpSession cookieHttpOnly="false" cookieName="JSESSIONID" invalidationTimeout="15s"/>`)
  [websocket-demo.war.zip](https://github.com/OpenLiberty/open-liberty/files/14764561/websocket-demo.war.zip)
  - the change session id is invoked is via the iframe refresh
  - clicking the websocket button creates a new websocket session which echos hello and a count. 
  - NOTE: use two browsers so that cookie aren't shared / use different urls (localhost / 127.0.0.1 ) to ensure cookies aren't shared

### DIFF
--- a/dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
+++ b/dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
@@ -1,14 +1,11 @@
 /*******************************************************************************
- * Copyright (c) 2017 IBM Corporation and others.
+ * Copyright (c) 2017, 2024 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
  * 
  * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *     IBM Corporation - initial API and implementation
  *******************************************************************************/
 package com.ibm.ws.wsoc;
 
@@ -40,11 +37,11 @@ public class EndpointManager {
     // Map the Endpoint config object to an instance of the Annotated Endpoint it is a config for.  For cache and clone generation.
     private final ConcurrentHashMap<Class<? extends Object>, AnnotatedEndpoint> annotatedEndpointMap = new ConcurrentHashMap<Class<? extends Object>, AnnotatedEndpoint>();
 
-    // Map of active sessions for a given endpoint class
+    // Map of active websocket sessions for a given endpoint class - for getOpenSessions
     private final ConcurrentHashMap<Class<?>, ArrayList<Session>> endpointSessionMap = new ConcurrentHashMap<Class<?>, ArrayList<Session>>();
 
-    // Map of  sessions with corresponding active HttpSession
-    private final ConcurrentHashMap<String, SessionExt> httpSessionMap = new ConcurrentHashMap<String, SessionExt>();
+    // Map of websocket sessions with corresponding active HttpSession (httpSession id is the key) - for HttpSession expiration
+    private final ConcurrentHashMap<String, ArrayList<Session>> httpSessionMap = new ConcurrentHashMap<String, ArrayList<Session>>();
 
     public EndpointManager() {
         serverEndpointConfigMap.clear();
@@ -63,27 +60,36 @@ public class EndpointManager {
             cl = ae.getServerEndpointClass();
         }
 
-        // find the session array for the given endpoint
+        // find the websocket session array for the given endpoint
         ArrayList<Session> sa = endpointSessionMap.get(cl);
 
-        // create a new session list if none was found
+        // create a new websocket session list if none was found
         if (sa == null) {
             sa = new ArrayList<Session>();
         }
 
-        // add the new session to the list for this endpoint
+        // add the new websocket session to the list for this endpoint
         sa.add(sess);
 
         // put the updated list back into the Map
         endpointSessionMap.put(cl, sa);
 
-        if (tc.isDebugEnabled()) {
-            Tr.debug(tc, "added session of: " + sess + "  to endpoint class of: " + cl + " using list of: " + sa + " in endpointmanager of: " + this);
+         String id = sess.getSessionImpl().getHttpSessionID();
+
+        if (id != null) {
+            ArrayList<Session> http_sa = httpSessionMap.get(id);
+            if (http_sa == null) {
+                // a httpSession may have multiple websocket connections
+                http_sa = new ArrayList<Session>();
+            }
+            http_sa.add(sess);
+            httpSessionMap.put(id, http_sa);
         }
 
-        String id = sess.getSessionImpl().getHttpSessionID();
-        if (id != null) {
-            httpSessionMap.put(id, sess);
+        if (tc.isDebugEnabled()) {
+            Tr.debug(tc, "added session of: " + sess + " id=" + id + "  to endpoint class of: " + cl + " using list of: " + sa + " in endpointmanager of: " + this);
+            Tr.debug(tc, "[add] httpSessionMap count: " + httpSessionMap.size() + " in endpointmanager of: " + this);
+            Tr.debug(tc, "[add] active websocket session count: " + sa.size() + " in endpointmanager of: " + this);
         }
     }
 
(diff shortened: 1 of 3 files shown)


volosied:
#build #spawn.fullfat.buckets=io.openliberty.wsoc.2.1.internal_fat,io.openliberty.wsoc.internal_fat,com.ibm.ws.wsoc_fat

dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
```diff
@@ -174,6 +178,9 @@ public synchronized Set<Session> getOpenSessions(Endpoint ep) {
     public void httpSessionExpired(String httpSessionID) {
         if (httpSessionID != null) {
             ArrayList<Session> sessionList = httpSessionMap.remove(httpSessionID);
+            if (tc.isDebugEnabled()) {
+                Tr.debug(tc, "sessionList count: " + sessionList.size());
+            }
```
volosied: Move this below the null check

dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
```diff
@@ -171,15 +177,55 @@ public synchronized Set<Session> getOpenSessions(Endpoint ep) {
...
+                            Tr.debug(tc, "removing session from http_sa: " + session);
+                        }
+                        http_sa.remove(session);
+                    }
+                    endpointSessionMap.put(endPointKey, http_sa);
+                    if (tc.isDebugEnabled()) {
+                        Tr.debug(tc, "removed expired session of: " + httpSessionID + "  from endpoint class of: " + endPointKey.getName() + " with list " + http_sa
+                                    + " in endpointmanager of: " + this);
+                        Tr.debug(tc, "[expire] http active session count: " + http_sa.size() + " in endpointmanager of: " + this);
+                    }
```
volosied: Log httpSessionMap count as well here.

volosied:
#build #spawn.fullfat.buckets=io.openliberty.wsoc.2.1.internal_fat,io.openliberty.wsoc.internal_fat,com.ibm.ws.wsoc_fat

dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
```diff
@@ -77,13 +74,21 @@ public synchronized void addSession(Endpoint ep, SessionExt sess) {
...
-            Tr.debug(tc, "added session of: " + sess + "  to endpoint class of: " + cl + " using list of: " + sa + " in endpointmanager of: " + this);
-        }
+         String id = sess.getSessionImpl().getHttpSessionID();
 
-        String id = sess.getSessionImpl().getHttpSessionID();
         if (id != null) {
-            httpSessionMap.put(id, sess);
+            ArrayList<Session> http_sa = httpSessionMap.get(id);
+            if (http_sa == null) {
+                http_sa = new ArrayList<Session>();
```
volosied, line 83: I created a list here since a single http session could have multiple websocket connections.

dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
```diff
@@ -44,7 +41,7 @@ public class EndpointManager {
     private final ConcurrentHashMap<Class<?>, ArrayList<Session>> endpointSessionMap = new ConcurrentHashMap<Class<?>, ArrayList<Session>>();
 
     // Map of  sessions with corresponding active HttpSession
-    private final ConcurrentHashMap<String, SessionExt> httpSessionMap = new ConcurrentHashMap<String, SessionExt>();
+    private final ConcurrentHashMap<String, ArrayList<Session>> httpSessionMap = new ConcurrentHashMap<String, ArrayList<Session>>();
```
pnicolucci, line 44: Changing from`SessionExt` to `Session` is ok here? I'm assuming we don't need to call anything in the concrete `SessionExt` class and we're fine with the `Session` API?

dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
```diff
@@ -77,13 +74,21 @@ public synchronized void addSession(Endpoint ep, SessionExt sess) {
...
 
-        if (tc.isDebugEnabled()) {
-            Tr.debug(tc, "added session of: " + sess + "  to endpoint class of: " + cl + " using list of: " + sa + " in endpointmanager of: " + this);
-        }
+         String id = sess.getSessionImpl().getHttpSessionID();
 
-        String id = sess.getSessionImpl().getHttpSessionID();
         if (id != null) {
-            httpSessionMap.put(id, sess);
+            ArrayList<Session> http_sa = httpSessionMap.get(id);
```
pnicolucci, line 80: A bit confusing are these httpSessions or wsoc sessions?

dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
volosied: I guess I could have kept it as SessionExt. I later use casting to ensure we call the right methods: ((SessionExt) session).

We can discuss more in our meeting.

dev/com.ibm.ws.wsoc/src/com/ibm/ws/wsoc/EndpointManager.java
volosied: We only import `javax.websocket.Session`, so the Sessions are websocket. 

Only part of the HttpSesssion used in this class is the Id.

volosied:
#build #spawn.fullfat.buckets=io.openliberty.wsoc.2.1.internal_fat,io.openliberty.wsoc.internal_fat,com.ibm.ws.wsoc_fat

pnicolucci (approved):

pnicolucci:
#libby

pnicolucci:
One bulid failure in the latest personal build associated with https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.workitem.viewWorkItem&id=292746

PR #26778, opened 2023-10-31
Validation tests for Reactive Messaging DeploymentExceptions

### LINKED ISSUE #26274: Add Reactive Messaging 3.0 tests for Validation Errors.
While the RM 3.0 TCK does check a deployment exception occurs, the contents of the message is not checked so some fats to make sure we are getting useful and expected messages should be added

ryan-storey:
These tests check that DeploymentExceptions are correctly thrown as [per the spec](https://download.eclipse.org/microprofile/microprofile-reactive-messaging-3.0/microprofile-reactive-messaging-spec-3.0.html#_assembly_and_validation).

Closes #26274


dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/validation/IncomingConnectorMultipleDownstreams.java
```diff
@@ -0,0 +1,33 @@
+/*******************************************************************************
+ * Copyright (c) 2017 IBM Corporation and others.
```
abutch3r: is 2017 the right year here? or should it be 2023

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/validation/KafkaValidationTests.java
```diff
@@ -0,0 +1,148 @@
...
+    public void testMissingConnector() throws Exception {
+        // Configure delivery to a non-existant connector
+        ConnectorProperties incoming = ConnectorProperties.simpleIncomingChannel(KafkaTests.connectionProperties(), "missingConnector", ValidConfig.INCOMING_CHANNEL, "Connector",
+                                                                                 "Topic");
+        PropertiesAsset appConfig = new PropertiesAsset().include(incoming);
+
+        AppValidator.validateAppOn(server)
+                        .withClass(ValidConfig.class)
+                        .withAppConfig(appConfig)
+                        .failsWith("Unknown connector for `" + ValidConfig.INCOMING_CHANNEL + "`.")
```
abutch3r: should try with a regex similar to `"SRMSG00072*" + ValidConfig.INCOMING_CHANNEL` that uses the Weld  Msg Code and the incoming channel in case the rest of the message content does change for some reason in the future

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/validation/KafkaValidationTests.java
```diff
@@ -0,0 +1,159 @@
...
+        server.startServer();
+    }
+
+    @AfterClass
+    public static void shutdown() throws Exception {
+        server.stopServer("CWMRX1006E");
+    }
+
+    @Test
+    public void testMissingConnector() throws Exception {
```
abutch3r, line 64: At the end of each test you should `Mark` your place, this should prevent the chance of a false positive if another test causes the same error signature. This can be done with `server.setMarkToEndOfLog();`

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/validation/AppValidator.java
```diff
@@ -0,0 +1,167 @@
...
+    private PropertiesAsset appConfig;
+
+    private static String APP_START_CODE = "CWWKZ000[13]I";
+    private static String APP_FAIL_CODE = "CWWKZ000[24]E";
+    private static int appCount = 0;
+
+    /**
+     * Begin setting up an app validation to run on the given liberty server
+     */
+    public static AppValidator validateAppOn(LibertyServer server) {
```
abutch3r: One frustration I have with this class is that we effectively end up with an arbitary testApp name which can only be matched by a regex as you have done, but there is no way to guarantee within a test what that app name is and we are checking that in the logs. I would recommend added a property called appName that can be set on the AppValidator before `.run()` is called and uses that. you can then set them in the individual tests and then check for that explicit appName and that will reduce potential false-postivites. It should use the mechanism already used here if appName is null

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/kafka/validation/KafkaValidationTests.java
abutch3r: reviewing the AppValidator class, that does set the mark before it actually deploys the application. so you shouldn't need to set in the tests themselves

ryan-storey:
#build

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/validation/AppValidator.java
```diff
@@ -0,0 +1,180 @@
...
+            }
+
+            try {
+                WebArchive app = ShrinkWrap.create(WebArchive.class, archiveName);
+
+                for (Class<?> clazz : classes) {
+                    app = app.addClass(clazz);
+                }
+                if (appConfig != null) {
+                    app = app.addAsLibraries(kafkaClientLibs())
```
abutch3r: I would have a generic methods on the class that could add libraries and manifests to the app, so that we don't have to include kafka specific entities within an application that might not use kakfka.

Also putting this in the appConfig block when they are unrelated doesn't make sense.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat/fat/src/io/openliberty/microprofile/reactive/messaging/fat/validation/AppValidator.java
```diff
@@ -0,0 +1,180 @@
...
+ *                 .failsWith("CWMFT5001E")
+ *                 .run();
+ * }
+ * </pre>
+ *
+ * <p>
+ * This creates a web application containing just {@code MyInvalidClass}, deploys it, checks that it fails to start and that a message containing {@code CWMFT5001E} is found in the
+ * log and then undeploys it.
+ */
+public class AppValidator {
```
abutch3r: I would move based on later comments to either a `utils` package in this project or into the RM Common FAT project

ryan-storey:
#build

ryan-storey:
#build

ryan-storey:
#build

abutch3r:
#libby
#build
#spawn.fullfat.butckets=io.openliberty.microprofile.reactive.messaging.internal_fat

dev/io.openliberty.microprofile.reactive.messaging.internal_fat.common/src/com/ibm/ws/microprofile/reactive/messaging/fat/AppValidator.java
```diff
@@ -0,0 +1,202 @@
...
+ * {@code
+ * AppValidator.validateAppOn(SHARED_SERVER)
+ *                 .withClass(MyInvalidClass.class)
+ *                 .failsWith("CWMFT5001E")
+ *                 .run();
+ * }
+ * </pre>
+ *
+ * <p>
+ * This creates a web application containing just {@code MyInvalidClass}, deploys it, checks that it fails to start and that a message containing {@code CWMFT5001E} is found in the
```
abutch3r: given this can do both successful and failure scenarios, I would correct the comment about it only being for failures

abutch3r (dismissed):
Only small bit about the AppValidator javadoc comments given that it can do either success or failure, but otherwise looks pretty much ready to go

abutch3r:
#build
#spawn.fullfat.butckets=io.openliberty.microprofile.reactive.messaging.internal_fat

abutch3r:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.reactive.messaging.internal_fat

abutch3r:
#libby

abutch3r (approved):

abutch3r:
we should look at squashing these commits into a smaller number of PRs, as 10 is quite a few, especially for something like this

ryan-storey:
#build

abutch3r:
#libby

abutch3r:
CI Orchestrator failures are known issues with other those buckets and unrelated to the test changes here

PR #26783, opened 2023-10-31
Use life cycle annotations rather than method name

njr-11:
Updates Jakarta Data tests and code to rely on life cycle annotations rather than method names.
Also, experiments with an annotation for specifying the entity attribute name for comparisons when using parameter-based queries.


njr-11:
#build

jmstephensgit (dismissed):

njr-11:
#build
#libby

jmstephensgit (approved):

Issue #26780, opened 2023-10-31
FAT com.ibm.ws.app.manager.wab.installer.fat.ConfigurableWABTests gets SRVE8059E 


epj:
FAT test com.ibm.ws.app.manager.wab.installer.fat.ConfigurableWABTests intermittently gets:
SRVE8059E: An unexpected exception occurred when trying to retrieve the session context 
This message is not causing tests to fail, but it does cause a build break.   Fat will be updated to ignore this message since it isn't otherwise impacting test results.   Likely the SRVE8059E is caused by intermittent slow performance of the cloud vms but the web container recovers and operates normally.

PR #26784, opened 2023-10-31
ConfigurableWABTests skip session context error

epj:
FAT test com.ibm.ws.app.manager.wab.installer.fat.ConfigurableWABTests intermittently gets:
SRVE8059E: An unexpected exception occurred when trying to retrieve the session context
This message is not causing tests to fail, but it does cause a build break. Fat will be updated to ignore this message since it isn't otherwise impacting test results. Likely the SRVE8059E is caused by intermittent slow performance of the cloud vms but the web container recovers and operates normally.

### DIFF
--- a/dev/com.ibm.ws.app.manager.wab.installer_fat/fat/src/com/ibm/ws/app/manager/wab/installer/fat/ConfigurableWABTests.java
+++ b/dev/com.ibm.ws.app.manager.wab.installer_fat/fat/src/com/ibm/ws/app/manager/wab/installer/fat/ConfigurableWABTests.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2021 IBM Corporation and others.
+ * Copyright (c) 2015, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -68,7 +68,8 @@ public class ConfigurableWABTests extends AbstractWABTests {
         if (server != null && server.isStarted()) {
             // ignore conflict error from the conflict test
             // ignore warning about virtual host not found
-            server.stopServer("CWWKZ0208E", "SRVE9956W");
+            // ignore session context error
+            server.stopServer("CWWKZ0208E", "SRVE9956W", "SRVE8059E");
         }
     }
 


epj:
#build

epj:
fixes #26780

pnicolucci:
> FAT test com.ibm.ws.app.manager.wab.installer.fat.ConfigurableWABTests intermittently gets: SRVE8059E: An unexpected exception occurred when trying to retrieve the session context This message is not causing tests to fail, but it does cause a build break. Fat will be updated to ignore this message since it isn't otherwise impacting test results. Likely the SRVE8059E is caused by intermittent slow performance of the cloud vms but the web container recovers and operates normally.

@epj  have you taken a look to determine that there isn't something we can do to prevent the NPE to begin with?

epj:
I spent time looking at trace.log.  I'm not seeing enough information to show why the null pointer is happening at this point.  Do we want to spend more time considering the intermittent nature of the problem?

pnicolucci:
> I spent time looking at trace.log. I'm not seeing enough information to show why the null pointer is happening at this point. Do we want to spend more time considering the intermittent nature of the problem?

I do think taking a look at it in more detail is good. It has been reported against 164 builds so it seems to be happening a lot!

epj:
From Tom, regarding the unavailable service result:
And that is causing the WebContainer to deactivate.  But that must be happening because of some transient dependency on this service.  When this service unregisters then it causes the WebContainer to deactivate and re-activate.  I suspect there is a background thread still using the old WebContainer instance.  But then when the new WebContainer instance comes on line the apps all get re-deployed and work fine.  This is why the test case seems to succeed but then fails at the very end with unexpected error in logs.  **I suspect we could simply ignore the error message**, but it probably would be better to understand why the WebContainer is deactivated more and if it is expected.  **If it is expected then I would suggest we could update the test to ignore this message on shutdown in the FAT test class.**

epj:
I think this is to be expected since the test case is changing the feature list, so when the feature manager is triggered the web container deactivates and activates normally.  It's just that intermittently the session context isn't available yet when WC is reactivating, which then causes the message SRVE8059E and a WC recovery.   Therefore I suggest we move forward with this PR to allow SRVE8059E in the FAT.

dev/com.ibm.ws.app.manager.wab.installer_fat/fat/src/com/ibm/ws/app/manager/wab/installer/fat/ConfigurableWABTests.java
```diff
@@ -4,7 +4,7 @@
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- *
```
pnicolucci, line 7: Can you update the copyright to include 2023 which is when this change was first committed?

epj:
#build

epj:
No errors in build.

pnicolucci (approved):

epj:
#libbybot

PR #26798, opened 2023-11-01
mpTelemetry-1.1 EE9 compatibility review comments

### LINKED ISSUE #26792: Follow-up actions for mpTelemetry-1.1 support for EE9/10
#26391 was merged with a lot of issues that we now need to address.

Azquelt:
Small tidyup changes from the review comments on #26391

JaxRsIntegrationWithConcurrency required some small changes to work with mpTelemetry-1.1.

Removed use of Nullable annotation since it's not available at runtime and we don't have any analysis tools which would use it at build time either.

Fixes #26792


Azquelt:
#build

Yas-Franco (approved):
Looks a lot better- Thanks!

PR #26799, opened 2023-11-01
Lazily create instrumenter in Telemetry filter classes for InstantOn to avoid early reading of config properties on checkpoint

### LINKED ISSUE #26522: InstantOn: Problem with checkpoint of mpTelemetry-1.x.
I have an application with the microProfile-6.0 feature enabled and this property set in microprofile-config.properties
```
otel.sdk.disabled=false
```

I see this error in the checkpoint step.

```
STEP 7/7: RUN checkpoint.sh afterAppStart
Performing checkpoint --at=afterAppStart

Launching defaultServer (Open Liberty 23.0.0.10/wlp-1.0.82.cl231020231002-1201) on Eclipse OpenJ9 VM, version 17.0.8+7 (en_US)
[AUDIT   ] CWWKE0001I: The server defaultServer has been launched.
[AUDIT   ] CWWKG0093A: Processing configuration drop-ins resource: /opt/ol/wlp/usr/servers/defaultServer/configDropins/defaults/keystore.xml
[AUDIT   ] CWWKG0093A: Processing configuration drop-ins resource: /opt/ol/wlp/usr/servers/defaultServer/configDropins/defaults/open-default-port.xml
[AUDIT   ] CWWKZ0058I: Monitoring dropins for applications.
[ERROR   ] CWMOT5002E: An internal error occurred in MicroProfile Telemetry. The error is: java.security.NoSuchAlgorithmException: PKIX TrustManagerFactory not available.
[AUDIT   ] CWWKZ0001I: Application acmeair-authservice started in 1.578 seconds.
[AUDIT   ] CWWKC0451I: A server checkpoint "afterAppStart" was requested. When the checkpoint completes, the server stops.
COMMIT io-acmeauth_test
```

### LINKED ISSUE #26744: Re-introduce change reverted from 26584
Re-introduce change reverted from https://github.com/OpenLiberty/open-liberty/pull/26584

anjumfatima90:
Add more MPTelemetry tests and repeat the tests for EE10.

fixes #26522
fixes #26744

### DIFF
--- a/dev/io.openliberty.checkpoint_fat/bnd.bnd
+++ b/dev/io.openliberty.checkpoint_fat/bnd.bnd
@@ -80,7 +80,8 @@ tested.features: \
 	sociallogin-1.0,\
 	mpMetrics-5.1,\
 	passwordUtilities-1.0,\
-	appSecurity-1.0
+	appSecurity-1.0,\
+	mpTelemetry-1.1
 
 # To define a global minimum java level for the FAT, use the following property.
 # If unspecified, the default value is ${javac.source}
--- a/dev/io.openliberty.checkpoint_fat/bnd.bnd
+++ b/dev/io.openliberty.checkpoint_fat/bnd.bnd
@@ -50,12 +50,7 @@ src: \
 	test-applications/webCacheApp/src,\
 	test-applications/xmlBindings/src,\
 	test-bundles/test-mbeans/src,\
-	test-applications/jaxrspropagation,\
-	test-applications/jaxrspropagation/common,\
-	test-applications/jaxrspropagation/methods,\
-	test-applications/jaxrspropagation/responses,\
-	test-applications/jaxrspropagation/spanexporter,\
-	test-applications/jaxrspropagation/transports
+	test-applications/jaxrspropagation/src
 
 fat.project: true
 
@@ -89,8 +84,9 @@ tested.features: \
 	appSecurity-1.0,\
 	mpTelemetry-1.1,\
 	mpConfig-3.1,\
-    restfulWS-3.1,\
-    servlet-6.0
+	restfulWS-3.1,\
+	servlet-5.0,\
+	servlet-6.0
 
 # To define a global minimum java level for the FAT, use the following property.
 # If unspecified, the default value is ${javac.source}
--- a/dev/io.openliberty.checkpoint_fat/bnd.bnd
+++ b/dev/io.openliberty.checkpoint_fat/bnd.bnd
@@ -49,7 +49,13 @@ src: \
 	test-applications/mapCache/src,\
 	test-applications/webCacheApp/src,\
 	test-applications/xmlBindings/src,\
-	test-bundles/test-mbeans/src
+	test-bundles/test-mbeans/src,\
+	test-applications/jaxrspropagation,\
+	test-applications/jaxrspropagation/common,\
+	test-applications/jaxrspropagation/methods,\
+	test-applications/jaxrspropagation/responses,\
+	test-applications/jaxrspropagation/spanexporter,\
+	test-applications/jaxrspropagation/transports
 
 fat.project: true
 
@@ -81,7 +87,10 @@ tested.features: \
 	mpMetrics-5.1,\
 	passwordUtilities-1.0,\
 	appSecurity-1.0,\
-	mpTelemetry-1.1
+	mpTelemetry-1.1,\
+	mpConfig-3.1,\
+    restfulWS-3.1,\
+    servlet-6.0
 
 # To define a global minimum java level for the FAT, use the following property.
 # If unspecified, the default value is ${javac.source}
@@ -126,7 +135,6 @@ tested.features: \
 	com.ibm.ws.org.slf4j.simple;version=latest,\
 	com.ibm.ws.org.slf4j.api;version=latest,\
 	com.ibm.websphere.org.eclipse.microprofile.faulttolerance.1.0,\
-	com.ibm.websphere.org.eclipse.microprofile.rest.client.1.4,\
 	io.openliberty.jakarta.restfulWS.3.0,\
 	jakarta.ws.rs-api,\
 	io.openliberty.jakarta.cdi.3.0,\
@@ -178,6 +186,10 @@ tested.features: \
 	com.ibm.websphere.appserver.api.distributedMap;version=latest,\
 	org.testng,\
 	com.ibm.ws.jmx,\
-	com.ibm.ws.kernel.service
+	com.ibm.ws.kernel.service,\
+	io.opentelemetry:opentelemetry-opentracing-shim;version=1.19.0.alpha,\
+	io.opentracing:opentracing-api;version=0.33.0,\
+	com.ibm.ws.kernel.boot.core;version=latest,\
+	com.ibm.ws.kernel.boot.common;version=latest
 
 -sub: *.bnd
--- a/dev/io.openliberty.checkpoint_fat/test-applications/jaxrspropagation/src/jaxrspropagation/JaxRsEndpoints.java
+++ b/dev/io.openliberty.checkpoint_fat/test-applications/jaxrspropagation/src/jaxrspropagation/JaxRsEndpoints.java
@@ -1,0 +1,396 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package jaxrspropagation;
+
+import static io.opentelemetry.api.trace.SpanKind.CLIENT;
+import static io.opentelemetry.api.trace.SpanKind.SERVER;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.HTTP_METHOD;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.HTTP_SCHEME;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.HTTP_STATUS_CODE;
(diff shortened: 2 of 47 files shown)


dev/io.openliberty.checkpoint_fat/bnd.bnd
```diff
@@ -49,7 +49,13 @@ src: \
...
 	test-applications/webCacheApp/src,\
 	test-applications/xmlBindings/src,\
-	test-bundles/test-mbeans/src
+	test-bundles/test-mbeans/src,\
+	test-applications/jaxrspropagation,\
+	test-applications/jaxrspropagation/common,\
+	test-applications/jaxrspropagation/methods,\
+	test-applications/jaxrspropagation/responses,\
+	test-applications/jaxrspropagation/spanexporter,\
+	test-applications/jaxrspropagation/transports
```
tjwatson: I don't understand why you need all these sub packages on the `src:` path like this when the source seems to be under `test-applications/jaxrspropagation/src/`

dev/io.openliberty.checkpoint_fat/test-applications/jaxrspropagation/src/jaxrspropagation/JaxRsEndpoints.java
```diff
@@ -0,0 +1,396 @@
...
+        client = ClientBuilder.newClient();
+    }
+
+    @PreDestroy
+    private void closeClient() {
+        if (client != null) {
+            client.close();
+            client = null;
+        }
+    }
```
tjwatson, line 133: Nobody is using this `client` instance.  I think you can delete this.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryClientFilter.java
```diff
@@ -49,48 +51,75 @@ public class TelemetryClientFilter extends AbstractTelemetryClientFilter impleme
...
 
     private final String configString = "otel.span.client.";
 
     private static final NetClientAttributesGetterImpl NET_CLIENT_ATTRIBUTES_GETTER = new NetClientAttributesGetterImpl();
     private static final HttpClientAttributesGetterImpl HTTP_CLIENT_ATTRIBUTES_GETTER = new HttpClientAttributesGetterImpl();
 
     TelemetryClientFilter() {
+        if (instrumenter == null) {
+            if (!CheckpointPhase.getPhase().restored()) {
+                lazyCreate = true;
```
tjwatson: We should only set `lazyCreate` to true if `openTelemetryInfo.getEnabled() && !AgentDetection.isAgentActive()` is true.  Otherwise it looks like we will continually try to create it over and over.  This applies to all cases here.

There is also a method on this class:

```java
    @Override
    public boolean isEnabled() {
        return instrumenter != null;
    }
```
This method will need to be updated to `return instrumenter != null || lazyCreate;`

That can only be done if you ensure `lazyCreate` is set to true when `openTelemetryInfo.getEnabled() && !AgentDetection.isAgentActive()` is true like I mention above

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryClientFilter.java
```diff
@@ -49,48 +51,75 @@ public class TelemetryClientFilter extends AbstractTelemetryClientFilter impleme
...
+    private volatile boolean lazyCreate = false;
+    private final AtomicReference<Instrumenter<ClientRequestContext, ClientResponseContext>> lazyInstrumenter = new AtomicReference<>();
 
     private final String configString = "otel.span.client.";
 
     private static final NetClientAttributesGetterImpl NET_CLIENT_ATTRIBUTES_GETTER = new NetClientAttributesGetterImpl();
     private static final HttpClientAttributesGetterImpl HTTP_CLIENT_ATTRIBUTES_GETTER = new HttpClientAttributesGetterImpl();
 
     TelemetryClientFilter() {
+        if (instrumenter == null) {
```
tjwatson: I don't see why we need a null check in the constructor.  There is no way it could be anything but null here.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryContainerFilter.java
```diff
@@ -69,11 +71,41 @@ public class TelemetryContainerFilter extends AbstractTelemetryContainerFilter i
...
     private Instrumenter<ContainerRequestContext, ContainerResponseContext> instrumenter;
+    
+    private volatile boolean lazyCreate = false;
+    private final AtomicReference<Instrumenter<ContainerRequestContext, ContainerResponseContext>> lazyInstrumenter = new AtomicReference<>();
 
     @jakarta.ws.rs.core.Context
     private ResourceInfo resourceInfo;
 
     public TelemetryContainerFilter() {
+        if (instrumenter == null) {
```
tjwatson: No need to null check here

dev/io.openliberty.checkpoint_fat/bnd.bnd
anjumfatima90: done!

dev/io.openliberty.checkpoint_fat/test-applications/jaxrspropagation/src/jaxrspropagation/JaxRsEndpoints.java
anjumfatima90: used the client instance in the test.

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryClientFilter.java
anjumfatima90: done!

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryContainerFilter.java
anjumfatima90: done!

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryClientFilter.java
anjumfatima90: Fixed isEnabled() for checkpoint and restore scenario.

anjumfatima90:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryClientFilter.java
```diff
@@ -49,48 +51,74 @@ public class TelemetryClientFilter extends AbstractTelemetryClientFilter impleme
...
+    }
+    
+    private Instrumenter<ClientRequestContext, ClientResponseContext> getInstrumenter() {
+        if (instrumenter != null) {
+            return instrumenter;
+        }
+        if (lazyCreate) {
+            instrumenter = lazyInstrumenter.updateAndGet((i) -> {
+                if (i == null) {
+                    lazyCreate = false;
```
tjwatson: Setting this here introduces a timing issue where more than one thread can enter the `getInstrumenter()` method while `instrumeter` is still `null`.  One thread could enter the `if (lazyCreate) {}` block and set this `lazyCreate` before the other one gets to the if block.  That may result in returning a `null` value from this method.

The fix is to set the `lazyCreate` to `true` after the assignment to `instrumenter`.  This applies to all other implementations of `getInstrumenter()` in the other classes.

tjwatson (dismissed):
LGTM

anjumfatima90:
#build

tjwatson (approved):

anjumfatima90:
#build

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryClientFilter.java
```diff
@@ -107,10 +139,14 @@ public Void run() {
...
+     * @return true when instrumenter is not null or if it is called during checkpoint
      */
     @Override
     public boolean isEnabled() {
-        return instrumenter != null;
+        if (!CheckpointPhase.getPhase().restored()) {
+            return true;
+        }         
+        return getInstrumenter() != null;  
     }
```
Azquelt, line 150: I guess this makes sense. We're using `isEnabled` to decide whether or not to register the filter for performance reasons. If we're doing this sort of check before checkpoint, we don't know whether or not the filter will end up being enabled once the server is restored, so we have to register the filter in case it is enabled on restore.

Azquelt (approved):
Telemetry changes look good.

tjwatson:
> We're using isEnabled to decide whether or not to register the filter for performance reasons.

I meant to ask about this.  Would the filter be optimized for this case such that even if it is registered it will not impact performance much and mostly be a no-op filter?

Azquelt:
>  Would the filter be optimized for this case such that even if it is registered it will not impact performance much and mostly be a no-op filter?

Looking at it I think so. If `getInstrumenter()` returns `null`, the `filter` method returns straight away and I think that's the best we can do if we have to register the filter.

The only alternative would be to see if we can dynamically register the filter later.

anjumfatima90:
Instant On build:
https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_fMLckHmUEe6p7N33NVgeoQ

anjumfatima90:
#libby

PR #26800, opened 2023-11-02
Redirect makeCurrent and storeInContext

### LINKED ISSUE #26134: SpanProxy should override makeCurrent
`SpanProxy` should probably override `makeCurrent()` from [ImplicitContextKeyed](https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-context/latest/io/opentelemetry/context/ImplicitContextKeyed.html) as a no-op.

We should also delegate `storeInContext()`.

Setting this class as the current span would be bad as it would redirect to itself and cause a stack overflow.

benjamin-confino:
fixes #26134

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/cdi/SpanProxy.java
+++ b/dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/cdi/SpanProxy.java
@@ -11,14 +11,16 @@ package io.openliberty.microprofile.telemetry.internal.common.cdi;
 
 import java.util.concurrent.TimeUnit;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.api.trace.StatusCode;
-
-import com.ibm.websphere.ras.Tr;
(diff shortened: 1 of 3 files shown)


dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/cdi/SpanProxy.java
```diff
@@ -27,6 +29,17 @@ public class SpanProxy implements Span {
 
     private static final TraceComponent tc = Tr.register(SpanProxy.class);
 
+    @Override
+    public Scope makeCurrent() {
+        //We can't do anything useful here but this will prevent the proxy itself becoming the current span.
+        return Span.current().makeCurrent();
```
Azquelt: I think this change will work, but it would be simpler to return `Scope.noop()`

benjamin-confino: Any risk that code which somehow ends up calling makeCurrent on our SpanProxy will expect the real current span from makeCurrent() and cause problems if it gets a noop?

Azquelt: Should be ok because this method doesn't return the current span, it returns a `Scope` which can be used to stop it being current.

In this case we're not changing the current span (this object always represents the current span, so calling `makeCurrent` will never change anything), so returning a `Scope` that does nothing is fine.

benjamin-confino:
#build

Azquelt (approved):

benjamin-confino:
#libby

PR #26813, opened 2023-11-02
Fix defect 293864

volosied:
Scenario where the web-socket connection is opened and immediately closed.  First thread (which handles the upgrade connection) does not complete properly since the close connection thread interrupts and calls destroy.  This results in the getObjectFactory returning null.

Only issue is that the HTTPAccess entry is skipped (wsoc upgrade call - see below) in this scenario since getRequest() throws an NPE due to getObjectFactory returning null (destroy occurs before the close). I've opted to check for null within getRequestImpl(), and return null instead. 

Example of the upgrade within the access log: 
```127.0.0.1 - [02/Nov/2023:14:53:44 -0400] "GET /basic/codedText HTTP/1.1" 101 -```

If the request is null within logFinalResponse then we log an entry within the trace.


volosied:
#build

volosied:
Build was good except for existing defects. 
image

mrsaldana (approved):
LGTM!

volosied:
#libby

PR #26817, opened 2023-11-02
transition from method name prefix to annotation based

njr-11:
The use of name prefixes on repository methods such as count and exists are now outside of the Jakarta Data specification, and should be replaced with the annotative approach that aligns better with parameter-based queries.


njr-11:
#build

mswatosh (approved):

PR #26820, opened 2023-11-03
Fix failing OpenAPI CORS test

ryan-storey:
Fixes https://wasrtc.hursley.ibm.com:9443/jazz/resource/itemName/com.ibm.team.workitem.WorkItem/297734


ryan-storey:
#build

ryan-storey:
#build

abutch3r (commented):
Changes to the pom.xml should be removed as they shouldn't be part of this PR, as they are likely due to something else within your workspace that has caused this change in reference and should be part of Otel changes.

Happy with the change for the generation of the URL as that does include getting the default HTTP port.

For future awareness, if you need to use HTTPS, then what you have here will not work as it will use the unsecure port

ryan-storey:
#libby

abutch3r (approved):

Issue #26053, opened 2023-08-25
Add getTracer(String instrumentationScopeName, String instrumentationScopeVersion) to GlobalOpenTemetry class in Telemetry 1.0


Yas-Franco:
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/io.openliberty.io.opentelemetry/src/io/opentelemetry/api/GlobalOpenTelemetry.java

Issue #25187, opened 2023-05-03
MicroProfile Telemetry 1.1


Emily-Jiang:
## Description

Implement MicroProfile Telemetry 1.1, which declares the dependencies on the latest release of OpenTelemetry 1.25 or later.
Get this feature to work with Java/Jakarta EE 7, 8, 9.1 and EE 10.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Documents

When available, add links to required feature documents. Use "N/A" to mark particular documents which are not required by the feature.

- Aha: Externally raised RFE ([Aha](https://cloud-platform.ideas.aha.io/))
  - Link the RFE with this issue
- UFO: https://ibm.box.com/s/7lqz6kpmyole7k1bp9kih85789ma4y8x
- FTS: https://github.com/OpenLiberty/open-liberty/issues/26445
- Beta Blog: https://github.com/OpenLiberty/open-liberty/issues/26183
- GA Blog: https://github.com/OpenLiberty/open-liberty/issues/26945

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## Process Overview

- [Prioritization](#prioritization)
- [Design](#design)
- [Implementation](#implementation)
- [Legal and Translation](#legal-and-translation)
- [Beta](#beta)
- [GA](#ga)
  - [Focal Point Approvals](#focal-point-approvals-complete-by-feature-complete-date)
- [Other Deliverables](#other-deliverables)

### General Instructions

The process steps occur roughly in the order as presented. Process steps occasionally overlap.

Each process step has a number of tasks which must be completed or must be marked as not applicable ("N/A").

Unless otherwise indicated, the tasks are the responsibility of the Feature Owner or a Delegate of the Feature Owner.

If you need assistance, reach out to the [OpenLiberty/release-architect](https://github.com/orgs/OpenLiberty/teams/release-architect).

**Important: Labels are used to trigger particular steps and must be added as indicated.** 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
## **Prioritization** (Complete Before Development Starts)

The ([OpenLiberty/chief-architect](https://github.com/orgs/OpenLiberty/teams/chief-architect)) and area leads are responsible for prioritizing the features and determining which features are being actively worked on.

### **Prioritization**
- [x] Feature added to the "New" column of the [Open Liberty project board](https://github.com/orgs/OpenLiberty/projects/2)
[...]

PR #26849, opened 2023-11-07
Telemetry: Add missing method to GlobalOpenTelemetry

Yas-Franco:
For issue #26053

### DIFF
--- a/dev/io.openliberty.io.opentelemetry/src/io/opentelemetry/api/GlobalOpenTelemetry.java
+++ b/dev/io.openliberty.io.opentelemetry/src/io/opentelemetry/api/GlobalOpenTelemetry.java
@@ -49,7 +49,7 @@ public final class GlobalOpenTelemetry {
         return OpenTelemetry.noop();
     }
 
-    //GlobalOpenTelemetry cannot be set twice
+    //GlobalOpenTelemetry cannot be set
     public static void set(OpenTelemetry openTelemetry) {
         throw new IllegalStateException(Tr.formatMessage(tc, "CWMOT5001.cannot.set.globalopentelemetry"));
     }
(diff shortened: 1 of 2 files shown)


dev/io.openliberty.io.opentelemetry/src/io/opentelemetry/api/GlobalOpenTelemetry.java
```diff
@@ -49,7 +49,7 @@ public static OpenTelemetry get() {
         return OpenTelemetry.noop();
     }
 
-    //GlobalOpenTelemetry cannot be set twice
+    //GlobalOpenTelemetry is only set by the Java Agent
     public static void set(OpenTelemetry openTelemetry) {
         throw new IllegalStateException(Tr.formatMessage(tc, "CWMOT5001.cannot.set.globalopentelemetry"));
     }
```
Azquelt: The comment here is a bit mystifying - the `set` method clearly always throws an exception, so it's not clear how it would be set by the Java Agent.

Azquelt: When using the Agent, this class is replaced entirely by the one provided by the agent, so nothing on this class is called.

Yas-Franco: Would it be more appropriate to just say "GlobalOpenTelemetry cannot be set" 

This is addressing this problem: https://github.com/OpenLiberty/open-liberty/issues/25187#issuecomment-1760367683

Azquelt: Ah, that makes sense.

Generally, comments should usually add some context to why things are like they are. It's obvious from looking at the code that `set()` will always throw an exception, but not why it's like that.

In this case, it's because we want users to inject the `OpenTelemetry` or `Tracer` object instead. We do put that information at the top of the file, and in the warning messages though so maybe `//GlobalOpenTelemetry cannot be set` is enough.

Aside: Looking at it again, I'm not sure why we didn't decide to delegate `get()` to get the `OpenTelemetry` object created by the mpTelemetry feature for the current application - possibly because at the time we were reliant on CDI to create that instance. Either way, we don't, we always return a no-op.

Azquelt (approved):

Yas-Franco:
#build

Yas-Franco:
#libby

PR #26858, opened 2023-11-07
Retrieve gc.time from MP Metrics, instead of GC MXBean for computed m…

### LINKED ISSUE #26835: Use gc.time and gc.count metric data from MP Metrics 5.x, when calculating gc.time.per.cycle computed metric
Currently, we use the GarbageCollection MXBean to retrieve `gc.time` and `gc.count` to compute the `gc.time.per.cycle` metric. For consistency with the other computed metrics, need to retrieve those metrics from the MP Metrics metric registry, instead.

pgunapal:
fixes #26835 
- For consistency with the other computed metrics, updated the code to retrieve the `gc.time` and `gc.count` from MP Metrics 5.x registry, instead of explicitly getting it from the GC MXBean.
- Replaced the extra step of retrieving and summing the Total Seconds and Total nanoseconds for the total REST elapsed time, with directly getting the total elapsed time in Nanoseconds and converted that into seconds.


pgunapal:
#build
#spawn.fullfat.buckets=io.openliberty.microprofile.metrics.internal.3.x.monitor_fat,io.openliberty.microprofile.metrics.internal.4.x.monitor_fat,io.openliberty.microprofile.metrics.internal.5.x.monitor_fat

Channyboy (approved):

pgunapal:
#libby

PR #26898, opened 2023-11-09
replace Filter functions and remove Filter

njr-11:
Replace the functions provided by Filter and instead use parameter annotations that line up with changes that were recently made to Jakarta Data.  Also remove Filter under this pull.


mswatosh (approved):

njr-11:
#build

PR #26901, opened 2023-11-09
spec updates to use Optional for Cursor on Pageable

njr-11:
Align with updates that were recently made to the spec changing the return type of pageable.cursor() to `Optional<Cursor>`

### DIFF
--- a/dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/page/Pagination.java
+++ b/dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/page/Pagination.java
@@ -14,6 +14,7 @@ package jakarta.data.page;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


njr-11:
#build

dev/io.openliberty.jakarta.data.1.0/src/jakarta/data/page/Pagination.java
```diff
@@ -60,6 +61,11 @@ public Pageable beforeKeysetCursor(Pageable.Cursor cursor) {
         return new Pagination(page, size, sorts, Mode.CURSOR_PREVIOUS, cursor);
     }
 
+    @Override
+    public Optional<Cursor> cursor() {
+        return type == null ? Optional.empty() : Optional.of(type);
```
KyleAure, line 66: I'm guessing the spec is defined this way, but the check is unnecessary we should have just written: 
```suggestion
        return Optional.ofNullable(type);
```

KyleAure (approved):

PR #26921, opened 2023-11-13
Telemetry: Add subresource test for java agent

### LINKED ISSUE #26882: Telemetry AgentTest adjustments for EE7/8
`AgentTest.testBasic` fails because the value returned by the agent for http.route is slightly different between EE7/8 and EE9/10 - probably due to differences in the way it instruments CXF and RestEasy.

We don't support the agent itself, we just want to test that we don't break when it's used, so as long as it doesn't look like something is going wrong, our test should accept whatever value the agent produces.

We should:
* Update these tests to accept the values that the agent produces for http.route
* Extend the tests to exercise the http.route calculation further so that we have a better idea of the agent's limitations. The things that make this computation more difficult are:
  * use of path parameters
  * use of subresources

Yas-Franco:
Fixes issue [#26882](https://github.com/OpenLiberty/open-liberty/issues/26882)


jakub-pomykala (approved):
LGTM

Yas-Franco:
#build

Azquelt (approved):

Azquelt:
I think this is the last thing for #26882? If so please mark it as fixing that issue.

PR #26925, opened 2023-11-13
Telemetry: Removing duplicate config in http tracing

### LINKED ISSUE #26804: Remove duplicated config reading logic in MP Telemetry HTTP tracing
`TelemetryServletFilter` [reads the telemetry config](https://github.com/OpenLiberty/open-liberty/blob/a3f0bd2d8da5f658c7c906b5bc2e7b55d967fd2e/dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryServletFilter.java#L283-L292) ~to [determine whether MP Telemetry is enabled](https://github.com/OpenLiberty/open-liberty/blob/a3f0bd2d8da5f658c7c906b5bc2e7b55d967fd2e/dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryServletFilter.java#L301-L309) before enabling the filter~.

However, this logic is [already done in `OpenTelemetryInfoFactory`](https://github.com/OpenLiberty/open-liberty/blob/a3f0bd2d8da5f658c7c906b5bc2e7b55d967fd2e/dev/io.openliberty.microprofile.telemetry.internal.common/src/io/openliberty/microprofile/telemetry/internal/common/info/OpenTelemetryInfoFactoryImpl.java#L191-L211). `TelemetryServletFilter` should not need to re-read this config and repeat this check.

Remove the duplicated code from `TelemetryServletFilter`.

_Update: `TelemetryServletFilter` actually checks a **different** property to determine whether HTTP tracing specifically has been enabled or not. However, it shouldn't extract the config into a map first (this is complex, expensive and unnecessary) and should just query MP Config for the property directly._

jakub-pomykala:
- Update the logic that checks `"otel.trace.http.disabled"` to avoid extracting all of the otel config into a map (which is complex and expensive)
  - it's not necessary here, we can just query MP Config directly
- Adding in a test to check the correct amount of spans is returned when `"otel.trace.http.disabled"` is `true` 
- Adding in a test where servlet tracing is disabled and we call a JAX-RS resource

Fixes #26804

Also:
- Removing a left over `system.out.println()` in `JaxRsRouteTestServlet` class


dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryDisabledServletTest.java
```diff
@@ -0,0 +1,87 @@
...
+import io.openliberty.microprofile.telemetry.internal_fat.apps.servlet.DisabledServlet;
+import io.openliberty.microprofile.telemetry.internal_fat.apps.servlet.SimpleServlet;
+import io.openliberty.microprofile.telemetry.internal_fat.common.TestSpans;
+import io.openliberty.microprofile.telemetry.internal_fat.common.spanexporter.InMemorySpanExporter;
+import io.openliberty.microprofile.telemetry.internal_fat.common.spanexporter.InMemorySpanExporterProvider;
+import io.openliberty.microprofile.telemetry.internal_fat.shared.spans.AbstractSpanMatcher;
+import io.opentelemetry.sdk.autoconfigure.spi.traces.ConfigurableSpanExporterProvider;
+
+/**
+ * HTTP request tracing tests
```
Yas-Franco: This comment needs updating

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
```diff
@@ -0,0 +1,79 @@
...
+@WebServlet("/testDisabledHttpTrace")
+public class DisabledServlet extends FATServlet {
+
+    public static final String APP_NAME = "TelemetryServletTestApp";
+    public static final String SIMPLE_SERVLET = "simple";
+    public static final String SIMPLE_ASYNC_SERVLET = "simpleAsync";
+    public static final String CONTEXT_ASYNC_SERVLET = "contextAsync";
+    public static final String PLACEHOLDER_SERVLET = "placeholder";
+    public static final String HELLO_HTML = "hello.html";
+    public static final String DICE_JSP = "dice.jsp";
```
Yas-Franco: You don't need most of these constants

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryDisabledServletTest.java
```diff
@@ -0,0 +1,87 @@
...
+                        .addProperty("otel.bsp.schedule.delay", "100");
+        BeansAsset beans = BeansAsset.getBeansAsset(DiscoveryMode.ALL);
+        WebArchive app = ShrinkWrap.create(WebArchive.class, APP_NAME + ".war")
+                        .addPackage(SimpleServlet.class.getPackage())
+                        .addPackage(InMemorySpanExporter.class.getPackage())
+                        .addPackage(TestSpans.class.getPackage())
+                        .addPackage(AbstractSpanMatcher.class.getPackage())
+                        .addAsServiceProvider(ConfigurableSpanExporterProvider.class, InMemorySpanExporterProvider.class)
+                        .addAsWebResource(new File("test-applications/TelemetryServletTestApp.war/hello.html"))
+                        .addAsWebResource(new File("test-applications/TelemetryServletTestApp.war/dice.jsp"))
```
Yas-Franco: I don't think you use those web resources in this test so you can probably remove them

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryDisabledServletTest.java
jakub-pomykala: Done

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryDisabledServletTest.java
jakub-pomykala: Done

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
jakub-pomykala: Done

Yas-Franco (dismissed):

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryServletFilter.java
```diff
@@ -312,17 +309,6 @@ public List<String> getHttpResponseHeader(final ServletRequest request, final Se
...
-        HashMap<String, String> telemetryProperties = new HashMap<>();
-        for (String propertyName : config.getPropertyNames()) {
-            if (propertyName.startsWith("otel") || propertyName.startsWith("OTEL")) {
-                config.getOptionalValue(propertyName.toLowerCase().replace('_', '.'), String.class).ifPresent(
-                    value -> telemetryProperties.put(propertyName.toLowerCase().replace('_', '.'), value));
-            }
-        }
-        return telemetryProperties;
-    }
-
```
Azquelt, line 325: The method below this one can be removed as well I think.

Azquelt: I also think these changes apply to the other version of `TelemetryServletFilter` that we have.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
```diff
@@ -0,0 +1,74 @@
...
+        String serverName = request.getServerName();
+        int serverPort = request.getServerPort();
+        URL url = new URL(scheme + "://" + serverName + ":" + serverPort + "/" + APP_NAME + "/" + SIMPLE_SERVLET);
+
+        String traceId = httpGet(url); // The servlet outputs the traceId
+
+        // The simple servlet will return the default invalid traceId
+        assertEquals(traceId, INVALID_TRACE_ID);
+
+    }
```
Azquelt: This test is valid and I think we should keep it, but it doesn't really test what you're changing here.

I think the only way to actually find out if the change has broken something would be to enable trace and look for the "instrumenter is initialized" or "instrumenter is set to null" messages.

We could add that check to this test, but I think it's sufficient to check it yourself manually before merging this PR.

Azquelt (changes requested):
These changes look mostly good. It would be helpful to add a justification in the PR description explaining why removing the call to `checkDisabled()` is correct and where the duplicated code is located.

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryServletFilter.java
jakub-pomykala: Done

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
jakub-pomykala: Run with tracing enabled which did return "instrumenter is set to null"

jakub-pomykala:
#build

jakub-pomykala:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryServletFilter.java
```diff
@@ -101,14 +99,16 @@ private Instrumenter<ServletRequest, ServletResponse> getInstrumenter() {
     }
 
     private Instrumenter<ServletRequest, ServletResponse> createInstrumenter() {
+        // Check if the HTTP tracing should be disabled
+        boolean httpTracingDisabled = config.getOptionalValue("otel.trace.http.disabled", Boolean.class).orElse(false);
         OpenTelemetryInfo otelInfo = OpenTelemetryAccessor.getOpenTelemetryInfo();
         if (tc.isDebugEnabled()) {
             Tr.debug(tc, "otelInfo.getEnabled()=" + otelInfo.getEnabled());
         }
```
Azquelt: Please trace the value of `httpTracingDisabled`

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryDisabledServletTest.java
```diff
@@ -0,0 +1,83 @@
...
+import io.openliberty.microprofile.telemetry.internal_fat.common.spanexporter.InMemorySpanExporter;
+import io.openliberty.microprofile.telemetry.internal_fat.common.spanexporter.InMemorySpanExporterProvider;
+import io.openliberty.microprofile.telemetry.internal_fat.shared.spans.AbstractSpanMatcher;
+import io.opentelemetry.sdk.autoconfigure.spi.traces.ConfigurableSpanExporterProvider;
+
+/**
+ * Test with opentelemetry disabled
+ */
+@RunWith(FATRunner.class)
+public class TelemetryDisabledServletTest extends FATServletClient {
```
Azquelt: We already have a `TelemetryDisabledTest`. It looks like this test is also disabling otel with `otel.sdk.disabled`.

If there's something new to test, should it be added to the existing test instead?

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+
+import org.junit.Test;
+
+import componenttest.app.FATServlet;
+
+@SuppressWarnings("serial")
+@WebServlet("/testDisabledHttpTrace")
+public class DisabledServlet extends FATServlet {
```
Azquelt: Please rename the class to `HttpTraceDisabledServlet` to better describe what it's testing.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
```diff
@@ -0,0 +1,74 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.telemetry.internal_fat.apps.servlet;
```
Azquelt: This servlet does not run in the same app as the rest of the classes in this package.

If you need a new app to test this (and I think you do, because you're testing different configuration), put it in a new package.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/TelemetryHttpTraceDisabledTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public static final String APP_NAME = "TelemetryServletTestApp";
+    public static final String SERVER_NAME = "Telemetry10Servlet";
+
+    @TestServlets({
+                    @TestServlet(contextRoot = APP_NAME, servlet = DisabledServlet.class),
+    })
+
+    @Server(SERVER_NAME)
+    public static LibertyServer server;
```
Azquelt: I don't think you need a new test class for this test. It could probably be a second app in `TelemetryServletTest`. This will avoid starting the server an extra time and make the test run more quickly.

dev/io.openliberty.microprofile.telemetry.1.1.internal/src/io/openliberty/microprofile/telemetry11/internal/rest/TelemetryServletFilter.java
```diff
@@ -96,16 +93,18 @@ private Instrumenter<ServletRequest, ServletResponse> getInstrumenter() {
...
     }
-    
+
     private Instrumenter<ServletRequest, ServletResponse> createInstrumenter() {
+        // Check if the HTTP tracing should be disabled
+        boolean httpTracingDisabled = config.getOptionalValue("otel.trace.http.disabled", Boolean.class).orElse(false);
         OpenTelemetryInfo otelInfo = OpenTelemetryAccessor.getOpenTelemetryInfo();
         if (tc.isDebugEnabled()) {
             Tr.debug(tc, "otelInfo.getEnabled()=" + otelInfo.getEnabled());
         }
```
Azquelt: Please trace out the value of `httpTracingDisabled` here.

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
Azquelt: This test is fine, but we should also add another test where servlet tracing is disabled and we call a JAX-RS resource (so we still expect spans from JAX-RS)

Azquelt (changes requested):

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryServletFilter.java
jakub-pomykala: Done

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/fat/src/io/openliberty/microprofile/telemetry/internal_fat/apps/servlet/DisabledServlet.java
jakub-pomykala: Done

dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/rest/TelemetryServletFilter.java
```diff
@@ -101,14 +99,17 @@ private Instrumenter<ServletRequest, ServletResponse> getInstrumenter() {
     }
 
     private Instrumenter<ServletRequest, ServletResponse> createInstrumenter() {
+        // Check if the HTTP tracing should be disabled
+        boolean httpTracingDisabled = config.getOptionalValue("otel.trace.http.disabled", Boolean.class).orElse(false);
```
Azquelt: Should use the `CONFIG_DISABLE_HTTP_TRACING_PROPERTY` constant that the removed code used for the config key.

jakub-pomykala:
#build

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/publish/servers/Telemetry10Servlet/server.xml
```diff
@@ -9,11 +9,14 @@
...
 
-    <application id="TelemetryServletTestApp" name="TelemetryServletTestApp"
-        type="war" location="TelemetryServletTestApp.war">
+    <application type="war" location="TelemetryServletTestApp.war">
         <classloader apiTypeVisibility="+third-party" />
     </application>
     
+<application type="war" location="HttpTraceDisabledServletTestApp.war">
+		<classloader apiTypeVisibility="+third-party" />
+	</application>
```
Azquelt: Try to match the indentation in the rest of the file.

Azquelt (dismissed):
There's one comment above but it's not important enough to prevent merging.

Azquelt:
#libby

jakub-pomykala:
#build

jakub-pomykala:
#libby

dev/io.openliberty.microprofile.telemetry.1.0.internal_fat/publish/servers/Telemetry10Servlet/server.xml
```diff
@@ -9,11 +9,14 @@
...
 
-    <application id="TelemetryServletTestApp" name="TelemetryServletTestApp"
-        type="war" location="TelemetryServletTestApp.war">
+    <application type="war" location="TelemetryServletTestApp.war">
         <classloader apiTypeVisibility="+third-party" />
     </application>
     
+    <application type="war" location="HttpTraceDisabledServletTestApp.war">
+		<classloader apiTypeVisibility="+third-party" />
+    </application>
```
Azquelt: Indentation is inconsistent here.

Azquelt (approved):

jakub-pomykala:
#build

jakub-pomykala:
#libby

jakub-pomykala:
The build failed due to unrelated defects.

Everything required for Telemetry passed.

PR #26926, opened 2023-11-13
Shared lib cache

### LINKED ISSUE #25944: Reuse Shared Library Containers PH55953
## Description

Updates for APAR PH55953:  Excess Artifact File System heap use by shared libraries used by multiple applications.

The update changes the initialization code such that a single overlay container is created for each file of the shared library, and is shared by applications.

A static store holds the shared library containers.  Usage is tracked by reference counts.  When a reference count is reduced to zero, the referenced library container is removed from the store.

The uninstall API is used clear references from applications which were uninstalled.  Finalize is used to ensure that applications which create references but which otherwise fail installation properly release their referenced library containers.

tbitonti:
Re-use shared library containers between applications.  Containers used by shared libraries are tracked by path using reference counts.  Each use of a shared library increases the reference count.  Reference counts are reduced when uninstalling an application.  (Or, when the application is garbage collected, if the usual install/uninstall APIs fail.)

FATs are coded and included in the PR.  See updates, below, for a test summary.


tbitonti:
#build

tbitonti:
#build

tbitonti:
#build

tbitonti:
<h3>FAT Testing</h3>

FAT tests are performed by the new application manager FAT class:

* open-liberty/dev/com.ibm.ws.app.manager_fat/fat/src/componenttest/application/manager/test/SharedLibUpdateConfigTest.java
  * testSharedLibsUpdateConfig()
* open-liberty/dev/com.ibm.ws.app.manager_fat/fat/src/componenttest/application/manager/test/SharedLibUpdateJarTest.java
  * testSharedLibsUpdateJar()

For the first test, the server is initially configured with sixteen applications and three shared libraries, and with four library JARs.

The state of the shared library cache is examined immediately to verify that the shared library cache has the expected state to verify that the expected initial operations were performed.

The initial verification shows that the cache operates correctly for a usual, static, configuration.

Next, using the same server, the configuration is updated to change details of the shared library configuration.  The test cycles several times through four configurations:

* one application and one shared library
* four applications and two shared libraries
* eight applications and two shared libraries
* sixteen applications and three shared libraries

The tests use four different shared library elements.

The cache state and of cache operations are verified after each update.

The second test configures two shared libraries each with two jar files, and does a file update on the jar files.

Currently, file notification is not firing, meaning the shared library updates are not occurring as expected.

However, when the configuration is updated, changes to the shared library jars are noticed and the shared libraries are updated.

tbitonti:
Build was good, except for a problem handling out-of-order cache trace output.  Correcting and rerunning.

tbitonti:
#build

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CollectionUtils.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package com.ibm.ws.app.manager.module.internal;
+
+import java.util.AbstractMap;
+import java.util.Map;
+
+//@formatter:off
```
dazavala: Why does this solution circumvent the eclipse project's code format configuration rather than modify it?

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
```diff
@@ -0,0 +1,434 @@
...
+ * capture cache itself is independent of synchronization within the
+ * capture supplier.  This is important because supplier invocations are
+ * expected to be expensive, and cache operations should not be impacted
+ * by supplier overhead.
+ *
+ * Callers must reliable release suppliers to avoid memory leaks.
+ */
+public class CaptureCache<T> {
+    // Trace with the application manager.
+    // TODO: Move DeferredCache to a more common location.
```
dazavala: Remove this comment. DeferedCache.java is not a part in this solution.   Also, the previous comment about Trace is superfluous as it repeats exactly what the code is doing.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
```diff
@@ -0,0 +1,434 @@
...
+        // "componenttest/application/manager/test/SharedLibTest.java"
+        // scans for and parses these log entries.
+
+        if ( isDebugEnabled ) {
+            debug( getTag(), "release",
+                   "[ " + key + " ] [ " + newCount + " ]: [ " + supplier + " ]" );
+        }
+    }
+
+    // Store a caching Supplier.  That moves production outside of management of
```
dazavala: Perhaps I misunderstood this comment. It seems more appropriate to the class's javadoc description as it explains why the cache utilizes Supplier semantics.  Also, is the TODO still a concern?  It's not a bad idea to track library PID as it will enhance serviceability, especially when examining the cache a heap dump.  If you can correlate the libPath for a cache entry using  the trace and/or the introspection, then I don't see a reason for concern.  Trace seemed to have all the necessary lib info when a cache entry is acquired.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    //
+
+    /**
+     * A single lock is used for key and supplier storage updates:
+     * Key data and supplier storage have linked state and are
+     * satisfied by a single lock.
+     */
+    private final Object cacheLock;
```
dazavala, line 147: The CaptureCache instance is shared among the deployed apps in the app manager.  If there's contention on this lock, consider revising the solution to use a read/write lock or a more granular lock that manages cache access using  libPath.

tbitonti:
#build

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
```diff
@@ -0,0 +1,434 @@
...
+ * The state of any particular capture supplier is completely independent
+ * of the state of the capture cache itself.  Synchronization of of the
+ * capture cache itself is independent of synchronization within the
+ * capture supplier.  This is important because supplier invocations are
+ * expected to be expensive, and cache operations should not be impacted
+ * by supplier overhead.
+ *
+ * Callers must reliable release suppliers to avoid memory leaks.
+ */
+public class CaptureCache<T> {
```
dazavala, line 57: FWICT, the CaptureCache class is used to pool container suppliers keyed by library path and to maintain a reference count for each supplier. It's an object pool of sort, and I think the solution in DeployedAppInfoBase should mention that or rename the instance variable to reflect its usage.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
```diff
@@ -0,0 +1,434 @@
...
+        synchronized( cacheLock ) {
+            KeyData keyData = keys.get(key);
+            if ( keyData == null ) {
+                uniqueKey = null;
+                newCount = -1;
+                supplier = null;
+            } else {
+                uniqueKey = keyData.getKey();
+                newCount = keyData.decrement();//++
+                if ( newCount == 0 ) {
```
dazavala: What happens if count goes negative?  I didn't see any code that would prevent this unexpected condition. 
 Consider changing to `if (newCount <= 0) {`, prohibit a negative count, and emit a debug warning if the condition occurs.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
```diff
@@ -118,15 +122,137 @@ public List<ContainerInfo> getClassesContainerInfo() {
...
+
+    /**
+     * Static cache of library containers.
+     *
+     * Keys are paths to the library containers.
+     *
+     * Use of a static cache requires that all library containers
+     * for a given file path are identical.
+     */
+    protected static final CaptureCache<LibraryContainerData> containerCache = new CaptureCache<>("container");
```
dazavala: Names of operations and fields become more general here.  This is a libraryContainerPool or libraryContainerCache.  I make similar comments below.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
```diff
@@ -118,15 +122,137 @@ public List<ContainerInfo> getClassesContainerInfo() {
...
+    /**
+     * Capture creation of a library container.
+     *
+     * @param libPath The path to the library container.
+     * @param supplier The supplier of the library container.
+     *
+     * @return A new supplier of the library container which will capture the
+     *         result of the base supplier.
+     */
+    protected static CaptureSupplier<LibraryContainerData> capture(String libPath,
```
dazavala: Name should indicate function is specific to library support, like captureLibrary or captureLibraryContainer.  Same for the relase methods.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
```diff
@@ -118,15 +122,137 @@ public List<ContainerInfo> getClassesContainerInfo() {
         }
     }
 
+    // Data for caching a container ...
+
+    protected static class LibraryContainerData {
```
dazavala, line 127: Don't need the protected access here.  Also, any reason why the class does not declare private access to its fields along with public accessor methods?

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
```diff
@@ -175,88 +299,238 @@ public List<ContainerInfo> getClassesContainerInfo() {
...
+        // TODO: This seems different than the service API:
+        // com.ibm.ws.app.manager.module.
+        //   DeployedAppServices.setupContainer(String, File)
+        //   DeployedAppInfoFactoryBase.setupContainer(String, File)
+        //   DeployedAppServicesImpl.setupContainer(String, File)
+        // Those seem to create containers for the application itself.  Such
+        // containers will have data specific to the application, and likely
+        // cannot be shared.
+
+        private Container setupContainer(String libraryPid, File libraryFile) {
```
dazavala, line 392: An observation: It's not clear to me whether this revised implementation of setupContainer() returns null for each of the conditions it used to.  That could be a change in behavior for calling method, `getContainers()`.

dazavala (changes requested):
Hey, @tbitonti.  I'll approve the solution once you've had a look at my comments.  I traced various scenarios we discussed and some we did not.  AFAIS the solution shares library container instances for common library paths among all applications that reference the libraries, and it correctly maintains reference counts to the containers.  Though the FAT verifies reference counting, I observed the solution maintains reference counts as expected by managing apps with the admin console.  Traces look serviceable; I did not dump the server to see the introspection.

I think you should strongly consider safeguarding the cached supplier (container) reference count from going negative.  Where possible remove or reduce the comments.  The code is readable.  Also, as I do not typically service the app manager I would like someone more familiar with the App Mgr to offer a second opinion.

Some concerns we've discussed:
1. What happens when a library path is shared by multiple apps via common lib reference, and the jar file deployed in that path is updated?  The app manager appears insensitive to dynamic updates to library jars.  I don't see any code that would force an app restart in this scenario.  And if the app manager were sensitive, then it must restart all apps that reference the library, just as it does when a `<library/>` config is changed dynamically.  

2. What about apps that reference private libraries or the global shared library?  I looked into this.  The app manager processes containers for private libraries and global libraries the same way it does for common libraries.  Running a few scenarios I saw no difference in how solution acquires and releases (pools) library containers for any of these library types, even when dynamically changing lib references from one type to another, including the global lib.

> FYI: It would be easy to add global and private library refs to the test scenarios.  Apps need not reference the `global` library.  Deploy a jar to `${server.config.dir}/lib/global`.  The jar will be visible to all deployed apps lacking a  `<classloading/>` element.  Apps that have a `<classloading/>` config must reference the `global` library to see the jar, and the deployment must declare a `<library/>` with `id=global`.  

3. What happens when a library path is updated in the library config?  WRT to app mgmt, any app referencing the library is restarted. For common lib references the restarted app classloaders will delegate to a common lib classloader having the new lib path (container) in its classpath; for private libs, the loaders will have the new lib path on their classpaths; and for global library, the loaders will share a parent "gsl" loader  having the new lib path in its classpath.
 
4. Any cache (pool) is vulnerable to leakage and inconsistency. What's happens if the user replaces a jar in an existing lib path and then restarts an app that references the lib, where the lib is also shared by another app.  Does the app manager pick up the new jar content -- i.e. does it send the updated the container to the new app classloader -- and will this cause an inconsistency with the cached container supplier?

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CollectionUtils.java
tbitonti: CollectionUtils is no longer in use and has been removed.
(Generally, the eclipse formatting often badly formats code; for entirely new classes, turning off the auto-format is easier than fighting it.)

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
tbitonti: Removed comment.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
tbitonti: The java doc for 'storage' was clarified.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
tbitonti: For now, a simple object lock seems sufficient.  Requests for containers for shared libraries are not expected to be performed with a high enough frequency for the lock implementation to make a difference.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
tbitonti: Ah, it's not a pool.  Each of the containers is actively in use, and is discarded as soon as the use count reaches zero.  None of the producers or containers is retained in a pool for later reuse.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
tbitonti: Additional logic added to handle a '< 0' case.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
tbitonti: Renamed from 'containerCache' to 'libraryContainerCache'.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
tbitonti: Rename to 'captureLibraryContainer' and 'releaseLibraryContainers'.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
tbitonti: Protected in case a subclass has a need to use this data structure.
'public final' with no accessors to keep the class lightweight.  It's intended to be a lightweight record type data structure.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/DeployedAppInfoBase.java
tbitonti: The cases *should* be the same; the intent was to answer null in exactly the same cases as before.  The main change is the addition of trace to the base container creation steps, which were added as an assist. That the container can ever be null is problematic.  That means we are simply ignoring class path entries that fail to resolve to a container for whatever reason.  The added trace is very useful when diagnosing problems in the container resolution.

tbitonti: Even in the case of dynamic updates, the new code should still obtain new contains every time the same container is resolved and the underlying file was updated (that is, if the file size or time stamp was updated).  Each call to resolve the container will check for changes and will only answer the cached container if there are no changes.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
dazavala: Please pardon the error.  Definitely not a pool.

dev/com.ibm.ws.app.manager.module/src/com/ibm/ws/app/manager/module/internal/CaptureCache.java
dazavala: Thanks.  I've seen many build breaks involving fields that contain inexplicable values.

dazavala (approved):
Approved.  All concerns addressed.

tbitonti:
#build

tbitonti:
#libby

tbitonti:
#build

tbitonti:
Build has a single unrelated failure:

```
no component | 97336 | 0 | 1 |   |  
com.ibm.ws.springboot.fat20_fat FAT tests | 1 | 0 | 1 |   |  
com.ibm.ws.springboot.support.fat.FATSuite | 1 | 0 | 1 | 0 s | 1
testBlockingIOServlet40

```

tbitonti:
Fixes #PH55953

PR #26951, opened 2023-11-14
Rename SkipIfCheckpointNotSupported to CheckpointTest

tjwatson:


anjumfatima90 (approved):
LGTM

dev/io.openliberty.checkpoint_fat_transaction/fat/src/io/openliberty/checkpoint/fat/MockServletTest.java
```diff
@@ -46,7 +46,7 @@
 import io.openliberty.checkpoint.spi.CheckpointPhase;
 
 @RunWith(FATRunner.class)
-@SkipIfCheckpointNotSupported
```
dazavala, line 49: @tjwatson: Noticed I declared a skip annotation on this mock checkpoint test. That was unnecessary, but it got me thinking: Given the intent to automagically run all checkpoint FATs on specialized build definitions, should we annotate mock tests as well to ensure these builds provide the entire result set and to potentially reduce the overhead of these FATs on other specialized build definitions?  The overhead is minor, but why would we care to exercise mock checkpoint FATs on windows, etc.?

dazavala (approved):
Approved.  I left a question for you in `MockServletTest` within the transaction FAT.  No change is necessary, but if you agree with my question, you may care to annotate the EJB and JPA checkpoint FATs.  All tests in those projects use the stubbed-out (mocked) checkpoint support in the server.

dev/io.openliberty.checkpoint_fat_transaction/fat/src/io/openliberty/checkpoint/fat/MockServletTest.java
tjwatson: I don't think we should annotate mock tests to not run on non-Linux.  Mock tests help enable developers running on non-linux to test out their code locally.  We should use the `alwaysRun` attribute on them to mark them as `CheckpointTest` though I think.

tjwatson:
#build

tjwatson:
#libby

PR #26953, opened 2023-11-14
JPA entity with attribute converter

njr-11:
Write a Jakarta Data test case that has an entity attribute annotated with Convert to supply a custom converter.
This required working around a problem where JPA/EclipseLink doesn't detect the converter class even though it clearly sees the Convert annotation.  To work around this, I added detection of the Convert annotation and explicitly added a converter element for it in the generated orm.xml

Also, this adds a test case to the validation/data bucket where constraints are placed on both the parameters and return type of a method rather than on the entity.


njr-11:
#build

KyleAure (approved):

PR #26954, opened 2023-11-14
javaMail-1.x: Export Activation APIs as specification APIs

neuwerk:
Because Java 9 dropped the `javax.activation` APIs, and we don't offer the Activation APIs within a seperate feature, we should export the `javax.activation` as `type="spec"` in the `IBM-API-Header`. Doing so will make it so the `javaMail-1.5` and `javaMail-1.6` features can load the MailCommandMap files that depend on Activation, when running on Java 9+, and prevent failures like this:

```
Caused by: java.lang.ClassNotFoundException: CWWKL0084W: The javax.activation.FileTypeMap class could not be loaded. Try enabling the jaxb-2.2 feature or a newer version of the feature in the server.xml file.
```


jhanders34 (approved):

neuwerk:
#build
#spawn.fullfat.buckets=com.ibm.ws.javamail.1.6_fat

neuwerk:
Failure in PB is unrelated to my changes, merging PR.

PR #26959, opened 2023-11-15
Reactive messaging mtls tests

### LINKED ISSUE #26852: Add tests for testing MTLS with RM Liberty-Kafka Connector
There are currently no RM - Kafka tests that use MTLS authentication. There are some tests that do some level of certificate testing. though that is primarily adding the Kafka Broker cert to the Liberty server to make sure we can validate a TLS secured Kafka broker.

As such we should test that the connector can support MTLS as it is a more and more common scenario for how to secure deployed microservices. 

There are three potential locations defined for Outbound certificate locations
- Channel Definition in mpConfig - takes precedence over Connector
- Connector Definition in mpConfig
- ~~Outbound transport - https://openliberty.io/docs/latest/reference/feature/transportSecurity-1.0.html#outbound~~

For outbound transport, as the host for the broker is the same regardless of connected topics. So would be the equivelant to not 

Test scenarios
- [x] Single Incoming and Outgoing to Broker with keystore defined on Connector in mpConfig
- [x] Single Incoming and Outgoing to Broker with keystore defined on the Channels in mpConfig
- [x] Single Incoming and Outgoing to Broker with different keystores defined on the Connector(invalid) and Channels(Valid) in mpConfig
- [x] Multiple Incoming and Outgoing Channel definitions in Single App to Broker using different certificates with keystores defined in mpConfig on the channels. The channels will be in pairs so that Incoming and outgoing
- [x] Multiple Applications with a single connector per app to the broker using different certificates in separate keystores defined in mpConfig on the Channel
- [x] ~~Single Channel to broker with keystore defined in SSL Outbound Config~~
- [x] ~~Mutliple channels to broker with keystore defined in SSL Outbound Config~~

abutch3r:
Add Mutual TLS tests for the liberty-kafka Connector

Reduce the number of repeats run due to time taken in FULL MODE is now very long

The sole RM 30 difference between running with MP60 vs MP61 is mpConfig 3.0 vs 3.1. These are technically the same apis. So for the majority of tests we don't need to run both. only a small set are kept to allow for testing in these scenarios that cover channels and connector details

Fixes #26852


abutch3r:
#build

abutch3r:
#liiby
#build
#spawn.fullfat.buckets=com.ibm.ws.microprofile.reactive.messaging_fat,com.ibm.ws.microprofile.reactive.messaging_fat_tck,com.ibm.ws.microprofile.reactive.streams.operators_fat,com.ibm.ws.microprofile.reactive.streams.operators_fat_tck,io.openliberty.microprofile.reactive.messaging30.internal_fat_tck,io.openliberty.microprofile.reactive.messaging.internal_fat,io.openliberty.microprofile.reactive.streams.operators30.internal_fat_tck

abutch3r:
#libby

abutch3r:
#build

abutch3r:
#build
#spawn.fullfat.buckets=com.ibm.ws.microprofile.reactive.messaging_fat,com.ibm.ws.microprofile.reactive.messaging_fat_tck,com.ibm.ws.microprofile.reactive.streams.operators_fat,com.ibm.ws.microprofile.reactive.streams.operators_fat_tck,io.openliberty.microprofile.reactive.messaging30.internal_fat_tck,io.openliberty.microprofile.reactive.messaging.internal_fat,io.openliberty.microprofile.reactive.streams.operators30.internal_fat_tck

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsMultipleAppsTest.java
```diff
@@ -0,0 +1,163 @@
...
+        // Producers do not have group.id or have a client.id set. By default, the client.id = `producer-<some number>`. with some number being incremental
+        // When multiple RM applications are running with at least one producer each, we start to see issues of MBeans already existing and not existing
+        // Not sure why that would be, but all the errors relate to the producer-metrics type, no other type seems to appear in the logs
+        // by setting at least one outgoings client.id, the issue is resolved.
+        //
+        // This does only seem to impact multiple applications on the same server, there is no evidence other tests have issues based on their logs.
+        //
+        // Setting on the Connectors does not resolve the issue as it then generates slightly different, but still related issues.
+
+        // TODO: Check with Andrew if I have a Scooby Doo
```
Azquelt: You haven't done this ;-p

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/tck/ReactiveStreamsTckTest.java
```diff
@@ -40,7 +40,9 @@
 import io.smallrye.reactive.streams.Engine;
 
 /**
- * Runs the Reactive Streams TCK against our Kafka connector
+ * Runs the Reactive Streams 1.0 TCK against our Kafka connector
+ *
+ * As this is the 1.0 TCK, it is not running the RSO 3.0 TCK, as such, it should not be repeated for RM30.
  * <p>
  * The Reactive Streams TCK is written using TestNG, so this test is a wrapper which launches TestNG to run the tests.
  */
```
Azquelt: This comment is not correct. This test is running the Reactive Streams TCK (not the MP RSO or the MP RM TCK) which tests that the `Producer`s and `Subscriber`s created by the Kafka connector conform to the rules for those interfaces.

This test is really a unit test and doesn't run a liberty server at all but we need to run it in the FAT bucket because it requires a Kafka server.

I don't think there's any need to repeat this test and the existing `RepeatTests` rule looks like a mistake.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat.common/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/containers/ExtendedKafkaContainer.java
```diff
@@ -115,6 +132,10 @@ public File getKeystoreFile() {
         return new File(KEYSTORE_FILENAME);
     }
 
+    public File getKeystoreFile2() {
+        return new File(KEYSTORE_FILENAME2);
+    }
+
```
Azquelt: We need to indicate what "keystore" and "keystore2" are for.

At the very least these methods should have javadoc, but renaming the method might be better.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat.common/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/containers/ExtendedKafkaContainer.java
```diff
@@ -143,6 +164,20 @@ protected void configure() {
             withEnv("KAFKA_SSL_TRUSTSTORE_FILENAME", KEYSTORE_FILENAME);
             withEnv("KAFKA_SSL_TRUSTSTORE_CREDENTIALS", KEYSTORE_PASSWORD_FILENAME);
             withEnv("KAFKA_SSL_ENABLED_PROTOCOLS", "TLSv1.2");
+            //MTLS only works if TLS is enabled.
+            if(mtls){
```
Azquelt: File needs formatting.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat.common/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/common/KafkaUtils.java
```diff
@@ -48,6 +50,16 @@ public static void copyTrustStore(ExtendedKafkaContainer container, LibertyServe
         copyFileToServer(container.getKeystoreFile(), server);
     }
 
+    public static void copyKeyStoresToServer(ExtendedKafkaContainer container, LibertyServer server) throws Exception {
+        // Copy First Keystore
```
Azquelt, line 66: We might want to refactor the methods which copy files in this class, they have lots of secret built-in constants and aren't documented.

dev/io.openliberty.microprofile.reactive.messaging.internal_fat.common/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/containers/ExtendedKafkaContainer.java
```diff
@@ -143,6 +164,20 @@ protected void configure() {
...
+                withEnv("KAFKA_SSL_CLIENT_AUTH", "required");
+                // Generate Second Keystore
+                subCommands.add(getCertGenerationCommand(KEYSTORE_FILEPATH2, KEYSTORE_PASSWORD, getHost()));
+                // Combine the keys into a Third option
+                if(mergeKeystores){
+                    subCommands.add(mergeKeyStoresCommand(KEYSTORE_FILEPATH, KEYSTORE_FILEPATH_COMBINED, "kafka-testcontainers", KEYSTORE_PASSWORD));
+                    subCommands.add(mergeKeyStoresCommand(KEYSTORE_FILEPATH2, KEYSTORE_FILEPATH_COMBINED, "kafka-testcontainers2", KEYSTORE_PASSWORD));
+                    // Change TrustStore location to combined store
+                    withEnv("KAFKA_SSL_TRUSTSTORE_FILENAME", KEYSTORE_FILENAME_COMBINED);
+                }
```
Azquelt: Not sure about the way we're setting up mTLS here, everything is just getting lumped in one keystore.

The existing code for setting up TLS isn't great either - the client gets the keystore which has the server's private key in it.

We might want to re-think this and there was some work done recently for Telemetry which makes it easier to generate certs and add them to the container.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsMultipleAppsTest.java
```diff
@@ -0,0 +1,163 @@
...
+        // When Kafka is initializing its producers and consumers, it also creates additional ones under the covers, in particular for metrics.
+        // As consumers typically have group.id set, the client.id is a derivation of the `<group.id>-<some number>`
+        // Producers do not have group.id or have a client.id set. By default, the client.id = `producer-<some number>`. with some number being incremental
+        // When multiple RM applications are running with at least one producer each, we start to see issues of MBeans already existing and not existing
+        // Not sure why that would be, but all the errors relate to the producer-metrics type, no other type seems to appear in the logs
+        // by setting at least one outgoings client.id, the issue is resolved.
+        //
+        // This does only seem to impact multiple applications on the same server, there is no evidence other tests have issues based on their logs.
+        //
+        // Setting on the Connectors does not resolve the issue as it then generates slightly different, but still related issues.
```
Azquelt, line 80: This seems odd. Do you think we need an issue? Is there anything we should either default in the connector, or recommend in the docs?

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsMultipleAppsTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+    @Test
+    public void testMutlipleMtlsApps() throws Exception {
+        runTest(server, APP1_NAME + "/KafkaMtlsServletOne", "putMessages");
+        runTest(server, APP2_NAME + "/KafkaMtlsServletTwo", "putMessages");
+
+        // the Kafka JMX issues are reported as Errors, not expcetions/ffdcs or have codes
+        // so is not picked up on shutdown of the server, meaning that it passes
+        assertNull("Unexpected Kafka Error",
+                server.waitForStringInLog("E Error.*kafka", 1000));
```
Azquelt: Since you've spotted this, we should probably wrap up this logic and use it in every test - ideally checking for error messages from Kafka after the server has stopped so we don't need to wait.

Could be done later.

Azquelt: If we do move this, we can also run the tests as regular test servlets, rather than calling both servlets here.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsTestServletOne.java
```diff
@@ -0,0 +1,49 @@
...
+@WebServlet("/KafkaMtlsServletOne")
+@ApplicationScoped
+public class KafkaMtlsTestServletOne extends FATServlet {
+
+    @Inject
+    private KafkaTestClient kafkaTestClient;
+
+    private static final long serialVersionUID = 1L;
+
+    public void putMessages() throws Exception {
```
Azquelt: Add a comment to note that this is called manually from the test class.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsTestServletTwo.java
```diff
@@ -0,0 +1,49 @@
...
+@WebServlet("/KafkaMtlsServletTwo")
+@ApplicationScoped
+public class KafkaMtlsTestServletTwo extends FATServlet {
+
+    @Inject
+    private KafkaTestClient kafkaTestClient;
+
+    private static final long serialVersionUID = 1L;
+
+    public void putMessages() throws Exception {
```
Azquelt, line 41: Add a comment to note that this is called manually from the test class.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/singleapp/KafkaMtlsMultipleChannelsTest.java
```diff
@@ -0,0 +1,130 @@
...
+import static com.ibm.ws.microprofile.reactive.messaging.fat.kafka.common.ConnectorProperties.simpleIncomingChannel;
+import static com.ibm.ws.microprofile.reactive.messaging.fat.kafka.common.ConnectorProperties.simpleOutgoingChannel;
+import static com.ibm.ws.microprofile.reactive.messaging.fat.kafka.common.KafkaUtils.kafkaClientLibs;
+import static com.ibm.ws.microprofile.reactive.messaging.fat.kafka.common.KafkaUtils.kafkaPermissions;
+import static componenttest.topology.utils.FATServletClient.runTest;
+
+
+@RunWith(FATRunner.class)
+@Mode(TestMode.FULL)
+public class KafkaMtlsMultipleChannelsTest {
```
Azquelt: I think the point of this test is to test two different channels in the same application authenticating using different certificates?

1. Since the point of this test is a bit subtle, it needs a Javadoc comment saying what you're trying to test.
2. I'm not sure that this test is really testing what you want it to. If the certificate from the first keystore was used for both connections, would the test still pass?

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/singleapp/KafkaMtlsMultipleChannelsTest.java
```diff
@@ -0,0 +1,130 @@
...
+        KafkaUtils.copyTrustStore(MtlsMultipleKeyStoresTests.kafkaContainer, server);
+        KafkaUtils.copyKeyStoresToServer(MtlsMultipleKeyStoresTests.kafkaContainer, server);
+
+        server.startServer();
+    }
+
+    @Test
+    public void testMutlipleMtlsChannels() throws Exception {
+        runTest(server, APP_NAME + "/KafkaMultipleMtlsTestServlet", "testMtls");
+    }
```
Azquelt: Use `@Test` on `KafkaMultipleMtlsTestServlet.testMtls` and `@TestServlet` on the `LibertyServer` field.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/single/KafkaMtlsChannelConnectorTest.java
```diff
@@ -0,0 +1,118 @@
...
+        KafkaUtils.copyTrustStore(MtlsTests.kafkaContainer, server);
+        KafkaUtils.copyKeyStoresToServer(MtlsTests.kafkaContainer, server);
+
+        server.startServer();
+    }
+
+    @Test
+    public void testMtlsConnector() throws Exception {
+        runTest(server, APP_NAME + "/KafkaMtlsTestServlet", "testMtls");
+    }
```
Azquelt: As before, use `@Test` on the servlet method and `@TestServlet` on the `LibertyServer` field.

Azquelt (changes requested):

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsMultipleAppsTest.java
abutch3r: We should definitely put recommendations in the docs for this. The issue seen here would impact someone using the kafka client directly or if for some reason they used a different kafka connector. 

if we were to try and resolve this in the kafka connector we would have to find a way to generate unique `client.id`'s that are server wide and that might require some documentation in itself.

abutch3r:
#libby
#build

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsMultipleAppsTest.java
abutch3r: [#27120](https://github.com/OpenLiberty/open-liberty/issues/27120) has been opened to address the kafka logging issue

abutch3r:
#libby
#build
#pbbeta

dev/io.openliberty.microprofile.reactive.messaging.internal_fat.common/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/containers/ExtendedKafkaContainer.java
```diff
@@ -111,8 +145,45 @@ public ExtendedKafkaContainer withKafkaJaasConfig(String kafkaJaasConfig) {
...
+     * to created a merged server truststore.
+     *
+     * @return
+     */
+    public File getKeystoreFile2() {
+        if(mtls) {
+            return new File(KEYSTORE_FILENAME2);
+        } else {
+            return null;
+        }
```
Azquelt, line 186: I don't like this because it doesn't match how we expect users to deploy mTLS and it's hard to tell what's in each keystore and where it should be used.

We should re-write how we create the key and trust stores and ensure they all have appropriate names.

In addition, the client and server should not be using the same keystore.

dev/com.ibm.ws.microprofile.reactive.messaging_fat/fat/src/com/ibm/ws/microprofile/reactive/messaging/fat/kafka/mtls/mutliple/multipleapp/KafkaMtlsMultipleAppsTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Test
+    public void testMutlipleMtlsApps() throws Exception {
+        runTest(server, APP1_NAME + "/KafkaMtlsServletOne", "putMessages");
+        runTest(server, APP2_NAME + "/KafkaMtlsServletTwo", "putMessages");
+
+        // the Kafka JMX issues are reported as Errors, not exceptions/ffdcs or have codes
+        // so is not picked up on shutdown of the server, meaning that it passes
+        assertNull("Unexpected Kafka Error",
+                server.waitForStringInLog("E Error.*kafka", 1000));
```
Azquelt, line 150: We should be using `KafkaUtils.kafkaStopServer()` on shutdown to catch errors logged by the Kafka connector.

Azquelt (approved):
I did note two things that should be improved, but I think it's worth getting these tests merged in their current state so that we can have them run on more platforms sooner.

PR #26983, opened 2023-11-16
Save config when preparing to checkpoint

tjwatson:


tjwatson:
#build

tjwatson:
#build

dazavala (dismissed):
On inspection I could not determine whether `ExtendedConfigurationImpl.delete()` may be called more than once during checkpoint.  This method invokes `ConfigurationStore.removeConfiguration()`, which calls `ConfigurationStore.save()`.  If `save()` may be called more than once on the same instance of `configurationStore`, then after the first call any additional calls to `save()` will not result in additional calls to `saveConfigurationDatas()` via the prepare hook. Rather, the `saveTask` will be scheduled.  That's the only potential gotcha I see.  Solution looks right and readable.

tjwatson:
> On inspection I could not determine whether `ExtendedConfigurationImpl.delete()` may be called more than once during checkpoint. This method invokes `ConfigurationStore.removeConfiguration()`, which calls `ConfigurationStore.save()`. If `save()` may be called more than once on the same instance of `configurationStore`, then after the first call any additional calls to `save()` will not result in additional calls to `saveConfigurationDatas()` via the prepare hook. Rather, the `saveTask` will be scheduled. That's the only potential gotcha I see. Solution looks right and readable.

`ConfigurationStore` is a singleton for the active `com.ibm.ws.config` bundle.  The `save` method not only gets called by `removeConfiguration` but also by 3 methods on `ExtendedConfigurationImpl`.  The `ConfigurationStore` maintains a single file to store all configuration object instances.  So any update/delete etc to the individual configuration objects will result in save being called which stores the complete file, which in turn stores all configuration instances currently in the `ConfigurationStore`.

The design of `ConfigurationStore` is to not force it to persist each time `save` is called.  Instead, during non-checkpoint cases, it adds an async task to run in the background to save the full file.  But the first save requests waits 30 seconds before doing the first save.  The reason is because during initial startup `save` gets called, potentially, 100s of times.  Waiting 30 seconds allows it to wait until, hopefully, all 100 calls to `save` have been made so that we can save the file only once on startup, making it much more efficient.

For the checkpoint case, a similar logic is taken, but instead of an async task to save we do the save in the `prepare` before the checkpoint happens.  At this point our startup process dealing with configuration settings should be all done, and all the `save` calls will have been completed.  This change ensures we call the `saveConfigurationDatas` no more than once before the checkpoint.  After restore, if configuration changes are made then it should go back to the "normal" behavior and create an async task to do the save after 30 seconds.

tjwatson:
#build

dazavala (approved):
LGTM.  Changed behavior at line 314.

dazavala:
#libby

PR #26990, opened 2023-11-17
Telemetry: Use correct @WithSpan binding

### LINKED ISSUE #25467: A better error for the NullPointer we get if WithSpan is on the class level
In order to make the WithSpanInterceptor we had to allow the WithSpan annotation to be put onto a Type.

However in a fat test if you put WithSpan onto an application class you get an NPE from WithSpanInterceptor.MethodRequestSpanNameExtractor.extract - we should either remove Type as a target for WithSpan or put a better error method in.

Azquelt:
Use CDI to look up the actual binding used to bind the interceptor

Fixes #25467

### DIFF
--- a/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/cdi/WithSpanInterceptor.java
+++ b/dev/io.openliberty.microprofile.telemetry.1.0.internal/src/io/openliberty/microprofile/telemetry10/internal/cdi/WithSpanInterceptor.java
@@ -10,15 +10,14 @@
  *******************************************************************************/
 package io.openliberty.microprofile.telemetry10.internal.cdi;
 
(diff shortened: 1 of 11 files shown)


Azquelt:
I want to check the performance impact of doing the service lookup this way before merging this.

Azquelt:
#build

Azquelt:
Currently the service lookup takes 64% of the method call time with J2S disabled, or 90% with J2S enabled - which seems a little much.

Using the streams API to find the WithSpan annotation from the list returned also takes longer than I'd expect, given that in most circumstances there will only be one, so a `for` loop there might be worthwhile.

Azquelt:
#libby
#build

Yas-Franco (approved):
Looks good

Azquelt:
Failures in the CI Orchestrator Pipeline build do not appear to be related to this change.

PR #27005, opened 2023-11-20
26942 liberty startup script does not resolve symbolic link to bin directory

### LINKED ISSUE #26942: Liberty startup script does not resolve symbolic link to bin directory
If a symbolic link is created to the bin directory of Liberty, the linked `server` script works to start the server but does not find the java9.options file. This can result in module access errors while Liberty is running.

jimblye:
Fixes #26942 

I needed to add a call early in the script to a function that is defined later in the script.    Before doing that, I decided to move the main program statements to the bottom, past all of the function definitions.   This is in separate commits.  The first commit should have no effect.   The second commit is the fix for symbolic links.

I also noticed a bug in the resolution of WLP_INSTALL_DIR in a secondary cygwin path.   Given that the bug has existed since before time, I suppose it is extremely unlikely that anyone is using that path.  I fixed it anyway.

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server
@@ -97,217 +97,6 @@
 #              action is done and when the server is restored.
 #
 ###############################################################################
-SERVER_UNKNOWN_STATUS=5
-
-##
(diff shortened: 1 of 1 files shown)


jimblye:
#build
#libby

hlhoots (approved):
LGTM

PR #27019, opened 2023-11-22
Wait for schedule in concurrent persistent JCA FAT

tkburroughs:
Update FAT to wait for the RAR start to schedule a task before running any test that also may schedule a task to avoid a know issue with Derby


tkburroughs:
#build

gjwatts (approved):

PR #27032, opened 2023-11-27
Increase DS lock timeout for EJB Async FAT

tkburroughs:
These FAT tests have DS logging enabled, which seems to result in an error in server logs when run on slower systems. Although test methods pass, the error fails the fat.

Increase the lock timeout to avoid the error:

CWWKE0701E: bundle com.ibm.ws.app.manager.war:1.0.81.cl230920230830-0401 (72)[com.ibm.ws.app.manager.ear.internal.EARDeployedAppInfoFactoryImpl(205)] : Timeout waiting for reg change to complete registered


tkburroughs:
#build

jmstephensgit (approved):

Issue #26976, opened 2023-11-16
Add repeats for TelemetryAgentConfig test


Yas-Franco:
The server name passed through [here](https://github.com/OpenLiberty/open-liberty/blob/integration/dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/AgentConfigTest.java#L80) is not the server that the application is deployed to. This means that the features are not actually updated by the repeat and the tests only run on mpTelemetry-1.0. 

It is worth checking if we have made this mistake elsewhere in the FAT bucket.

PR #27067, opened 2023-11-29
Telemetry repeat container tests

Yas-Franco:
For issue #26976


dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/AgentConfigTest.java
```diff
@@ -70,14 +72,19 @@
...
 
     @ClassRule
     public static JaegerContainer jaegerContainer = new JaegerContainer().withLogConsumer(new SimpleLogConsumer(JaegerBaseTest.class, "jaeger"));
 
+    // In contrast to most tests, this test needs a new jaeger instance for each repeat
+    // so that it can check for any trace IDs not accounted for
+
     @ClassRule
-    public static RepeatTests r = MicroProfileActions.repeat("spanTestServer", MicroProfileActions.MP61, MicroProfileActions.MP60);
+    public static RepeatTests r = FATSuite.allMPRepeats(SERVER_NAME);
```
Azquelt: Is this comment correct I don't see anything that's checking for traces not accounted for?

Either way, you should probably use a `RuleChain` to ensure the rules execute in the right order.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/AgentConfigTest.java
```diff
@@ -70,14 +72,19 @@
...
 
     public static JaegerQueryClient client;
```
Azquelt, line 89: This is an existing problem rather than something you've introduced, but I don't see anything closing the client.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/fat/src/io/openliberty/microprofile/telemetry/internal/tests/AgentConfigTest.java
```diff
@@ -96,10 +103,9 @@ public static void setup() throws Exception {
     @Before
     public void resetServer() throws Exception {
         // Replace any test files with their original versions
-        copyToServer("server.xml-original", "server.xml");
-        copyToServer("jvm.options-original", "jvm.options");
-        copyToServer("bootstrap.properties-original", "bootstrap.properties");
         deleteFromServer("agent-config.properties");
+        deleteFromServer("jvm.options");
+        deleteFromServer("bootstrap.properties");
```
Azquelt: I can see why you changed the `server.xml` so that we don't copy and edit it for each test, that makes doing repeats much easier, but I'm not sure why you changed the way `bootstrap.properties` and `jvm.options` are configured.

dev/io.openliberty.microprofile.telemetry.1.0.internal.jaeger_fat/publish/servers/TelemetryAgentConfig/bootstrap.properties
```diff
@@ -0,0 +1,4 @@
+bootstrap.include=../testports.properties
+
+# Exempt because the OTel agent does not work with Java 2 Security enabled
+websphere.java.security.exempt=true
```
Azquelt: This file never gets used because the `@Before` method deletes it before the test even runs.

Azquelt (commented):
A few questions but looks good otherwise.

jakub-pomykala (approved):
LGTM

Yas-Franco:
#build

Yas-Franco:
#libby

PR #27126, opened 2023-12-08
Reactive Messaging context service configuration

### LINKED ISSUE #26678: Async improvements for Reactive Messaging 3.0
During review of #26352, we found some improvements that should be made to the way Reactive Messaging 3.0 handles async work.

When a feature does async work which needs to have the appropriate context propagated, it can take one of three approaches:
- Use the liberty global thread pool to execute tasks and a built-in `WSContextService` instance with a fixed list of context types to propagate to transfer context.
  - This gives the user no control over what context is propagated, so one of the other options is usually preferred
- Use the liberty global thread pool to execute tasks and a `WSContextService` instance configurable by the user
  - Users can configure context services in server.xml if `concurrent-x.x` feature is used
  - The feature should allow the user to select which context service to use
- Use a managed executor service, configurable by the user which will execute tasks and automatically propagate context
  - Users can configure managed executors (including propagation options and concurrency limits) in server.xml if `concurrent-x.x` feature is used
  - The feature should allow the user to select which context service to use

In RM 3.0, we should take the second approach, falling back to the first approach if the `concurrent-x.x` feature is not enabled.

Azquelt:
Several changes here:

- Remove the RM 3.0 dependency on concurrent
  - Use global thread pool and internal `WSContextService` instead of  `ManagedExecutorService` and `ManagedScheduledExecutorService`
- Add explicit tests for context propagation with RM
- Add new config option to select a custom context service

With this change, users who are using the `concurrent-x.x` feature, can define a context service in server.xml:

```xml
    <contextService id="myContextService">
        <classloaderContext/>
        <jeeMetadataContext/>
    </contextService>
```

They can then configure the kafka connector to use it by setting MP Config properties. It can be set globally:
```properties
mp.messaging.connector.liberty-kafka.context.service=myContextService
```

It can also be set for a specific channel:
```properties
mp.messaging.incoming.myChannel.context.service=myContextService
```

If this property is not set, it will use the default context service if `concurrent-x.x` is enabled, or the built-in context service otherwise.

Fixes #26678

### DIFF
--- a/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTracker.java
+++ b/dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTracker.java
@@ -4,7 +4,7 @@
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -20,6 +20,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import org.eclipse.microprofile.reactive.messaging.Message;
 
 import com.ibm.ws.microprofile.reactive.messaging.kafka.adapter.TopicPartition;
+
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.RMContext;
 
 /**
  * Tracks a particular assignment of a partition to this consumer
@@ -72,9 +74,10 @@ public class PartitionTracker {
      *
      * @param offset the record offset
      * @param leaderEpoch the record leaderEpoch
+     * @param context the context to apply around the completion of the result, if the completion stage is completed asynchronously
      * @return a CompletionStage which completes when any associated processing has been completed (e.g. when the message offset has been committed)
      */
-    public CompletionStage<Void> recordDone(long offset, Optional<Integer> leaderEpoch) {
+    public CompletionStage<Void> recordDone(long offset, Optional<Integer> leaderEpoch, RMContext context) {
         return CompletableFuture.completedFuture(null);
     }
 
--- a/dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/QuiesceRegisterImpl.java
+++ b/dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/QuiesceRegisterImpl.java
@@ -9,6 +9,7 @@
  *******************************************************************************/
 package io.openliberty.microprofile.reactive.messaging.internal;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -21,16 +22,22 @@ import io.openliberty.microprofile.reactive.messaging.internal.interfaces.Quiesc
 import io.openliberty.microprofile.reactive.messaging.internal.interfaces.QuiesceRegister;
 
 /**
- *
+ * The singleton implementation of {@link QuiesceRegister}
  */
 @Component(configurationPolicy = ConfigurationPolicy.IGNORE)
 public class QuiesceRegisterImpl implements ServerQuiesceListener, QuiesceRegister {
 
-    private final Set<QuiesceParticipant> participants = new HashSet<>();
+    private final Set<QuiesceParticipant> participants = Collections.synchronizedSet(new HashSet<>());
 
     @Override
     public void serverStopping() {
-        for (QuiesceParticipant participant : participants) {
+        Set<QuiesceParticipant> participantsCopy;
+        // Avoid holding a lock on participants while calling quiesce() to avoid deadlock in case remove() is called at the same time
+        synchronized (participants) {
+            participantsCopy = new HashSet<>(participants);
+        }
+
+        for (QuiesceParticipant participant : participantsCopy) {
             participant.quiesce();
         }
     }
--- a/dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/QuiesceRegisterImpl.java
+++ b/dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/QuiesceRegisterImpl.java
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.reactive.messaging.internal;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+
+import com.ibm.wsspi.kernel.service.utils.ServerQuiesceListener;
+
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.QuiesceParticipant;
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.QuiesceRegister;
+
+/**
+ *
+ */
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE)
+public class QuiesceRegisterImpl implements ServerQuiesceListener, QuiesceRegister {
+
+    private final Set<QuiesceParticipant> participants = new HashSet<>();
+
+    @Override
+    public void serverStopping() {
+        for (QuiesceParticipant participant : participants) {
+            participant.quiesce();
+        }
+    }
+
+    @Override
+    public void register(QuiesceParticipant participant) {
+        participants.add(participant);
+    }
+
+    @Override
+    public void remove(QuiesceParticipant participant) {
+        participants.remove(participant);
+    }
+
+}
--- a/dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
+++ b/dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
@@ -11,12 +11,13 @@ package io.openliberty.microprofile.reactive.messaging.internal;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicy;
@@ -32,7 +33,7 @@ import io.openliberty.microprofile.reactive.messaging.internal.interfaces.RMAsyn
 import io.openliberty.microprofile.reactive.messaging.internal.interfaces.RMAsyncProviderFactory;
 import io.openliberty.microprofile.reactive.messaging.internal.interfaces.RMContext;
 
-@Component
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE)
 public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
 
     private static final TraceComponent tc = Tr.register(RMAsyncProviderFactoryImpl.class);
@@ -59,7 +60,14 @@ public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
     @Reference(target = "(id=DefaultContextService)", cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
     private volatile WSContextService defaultContextService;
 
-    private final Map<String, WSContextService> namedContextServices = Collections.synchronizedMap(new HashMap<>());
+    /**
+     * A map of named context services configured in the server.xml
+     * <p>
+     * Maps from the context service ID to the context service.
+     * <p>
+     * Updated dynamically if context services are added or removed while the server is running.
+     */
+    private final Map<String, WSContextService> namedContextServices = new ConcurrentHashMap<>();
 
     /**
      * The liberty global thread pool executor
@@ -73,6 +81,9 @@ public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
     @Reference(target = "(deferrable=false)", policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
     private volatile ScheduledExecutorService scheduledExecutor;
 
+    /**
+     * Updates {@link #namedContextServices} when a new service is added
+     */
     @Reference(service = WSContextService.class, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY, cardinality = ReferenceCardinality.MULTIPLE)
     protected void addContextService(WSContextService contextService, Map<String, Object> properties) {
         String name = getContextServiceName(properties);
@@ -81,21 +92,35 @@ public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
         }
     }
 
+    /**
+     * Updates {@link #namedContextServices} when the properties of a context service are updated
+     * <p>
+     * Not sure if this actually happens, but since the properties includes the name, that could in theory change.
+     */
     protected void updatedContextService(WSContextService contextService, Map<String, Object> properties) {
         String name = getContextServiceName(properties);
         if (name == null) {
-            namedContextServices.entrySet().removeIf(e -> e.getValue() == contextService);
-        } else if (namedContextServices.get(name) != contextService) {
+            namedContextServices.values().remove(contextService);
+        } else if (!contextService.equals(namedContextServices.get(name))) {
             // If name has changed, remove and re-add
-            namedContextServices.entrySet().removeIf(e -> e.getValue() == contextService);
+            namedContextServices.values().remove(contextService);
             namedContextServices.put(name, contextService);
         }
     }
 
+    /**
+     * Updates {@link #namedContextServices} when a context service is removed
+     */
     protected void removeContextService(WSContextService contextService) {
-        namedContextServices.entrySet().removeIf(e -> e.getValue() == contextService);
+        namedContextServices.values().remove(contextService);
     }
 
+    /**
+     * Extract the name of a context service from its OSGi service properties
+     *
+     * @param properties the OSGi service properties
+     * @return the name, or {@code null} if the context service doesn't have a name (e.g. because it's an internal one)
+     */
     private String getContextServiceName(Map<String, Object> properties) {
         Object name = properties.get("id");
         if (name instanceof String) {
@@ -110,11 +135,18 @@ public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
         return new NamedAsyncProvider(contextServiceRef);
     }
 
+    /**
+     * Look up the context service with the given name and use it to capture the thread context
+     *
+     * @param contextServiceName the context service to use, or {@code null} to use the default context service
+     * @return the captured thread context
+     */
     @SuppressWarnings("unchecked")
     private RMContext captureContext(String contextServiceName) {
         WSContextService namedContextService = null;
         if (contextServiceName != null) {
             namedContextService = namedContextServices.get(contextServiceName);
+            // TODO: confirm behavior if named context service not found
         }
 
         if (namedContextService != null) {
@@ -135,10 +167,20 @@ public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
         }
     }
 
+    /**
+     * An RMAsyncProvider which uses a named context service to capture thread context.
+     * <p>
+     * {@code null} can be provided as the name to use the default context service.
+     */
     private class NamedAsyncProvider implements RMAsyncProvider {
 
         private final String name;
 
+        /**
+         * Create a new AsyncProvider
+         *
+         * @param name the name of the context service to use for capturing thread context, or {@code null} to use the default
+         */
         public NamedAsyncProvider(String name) {
             this.name = name;
         }
@@ -163,6 +205,9 @@ public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
 
         private final ThreadContextDescriptor context;
 
+        /**
+         * @param context the thread context captured with {@link WSContextService#captureThreadContext(Map, Map...)}
+         */
         public RMContextImpl(ThreadContextDescriptor context) {
             this.context = context;
         }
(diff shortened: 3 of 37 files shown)


Azquelt:
#build

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTracker.java
```diff
@@ -72,9 +74,10 @@ public boolean isClosed() {
      *
      * @param offset the record offset
      * @param leaderEpoch the record leaderEpoch
+     * @param context the context to apply around the completion of the result, if the completion stage is completed asynchronously
      * @return a CompletionStage which completes when any associated processing has been completed (e.g. when the message offset has been committed)
      */
-    public CompletionStage<Void> recordDone(long offset, Optional<Integer> leaderEpoch) {
+    public CompletionStage<Void> recordDone(long offset, Optional<Integer> leaderEpoch, RMContext context) {
```
abutch3r, line 80: How is the context here being applied to the result, given we currently aren't doing anything with it.

abutch3r:
I can't find how the new tests are going to get run. there are no changes in any of the parent test classes in the `suite` package to either add a new Context set or an update to an existing one

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,183 @@
...
+    @Reference(target = "(id=DefaultContextService)", cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
+    private volatile WSContextService defaultContextService;
+
+    private final Map<String, WSContextService> namedContextServices = Collections.synchronizedMap(new HashMap<>());
+
+    /**
+     * The liberty global thread pool executor
+     */
+    @Reference(target = "(component.name=com.ibm.ws.threading)", policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
+//    @Reference(service = ManagedExecutorService.class, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
```
abutch3r: Is this commented out item needed, looks like it was has been replaced?

abutch3r:
Overall looks good, though the main concern is around the tests and if they actually run, it looks like they were setup to be run as part of the the `PlainTextTests`, but there is no update in that file

dev/com.ibm.ws.microprofile.reactive.messaging.kafka/src/com/ibm/ws/microprofile/reactive/messaging/kafka/PartitionTracker.java
Azquelt: Here the context is unused. We don't need to apply it in this method because we return a `CompletableFuture` which is already complete.

However, this method takes the context as an argument because the method is overridden in `CommittingPartitionTracker` which returns an incomplete `CompletableFuture` and applies the context around the call to `complete()` when that's called asynchronously.

Azquelt:
> I can't find how the new tests are going to get run. there are no changes in any of the parent test classes in the `suite` package to either add a new Context set or an update to an existing one

Yeah, good spot, they should have been added to `PlaintextTests`. I was testing with the suite modified to only run the new tests, then I undid my changes to run all the tests again without actually committing the change to add the new test classes.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
Azquelt: This is left over from when I was tracking down the cause of a behaviour change when using the global thread pool vs. `ManagedExecutorService`. I will remove it.

Azquelt:
#build
#libby

I think I've addressed all the comments, this is ready for re-review.

abutch3r (dismissed):
Assuming the Tests pass on the latest PB, LGTM

abutch3r:
All failures are unrelated to Reactive messaging and have confirmed that the new tests have been run in the rtc with all repeats

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+import com.ibm.wsspi.threadcontext.ThreadContext;
+import com.ibm.wsspi.threadcontext.ThreadContextDescriptor;
+import com.ibm.wsspi.threadcontext.WSContextService;
+
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.RMAsyncProvider;
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.RMAsyncProviderFactory;
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.RMContext;
+
+@Component
+public class RMAsyncProviderFactoryImpl implements RMAsyncProviderFactory {
```
Azquelt: Set `configurationPolicy` for this component

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+    @Reference(target = "(service.pid=com.ibm.ws.context.manager)", policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
+    private volatile WSContextService builtInContextService;
+
+    /**
+     * The default context service provided by the concurrent feature
+     */
+    @Reference(target = "(id=DefaultContextService)", cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
+    private volatile WSContextService defaultContextService;
+
+    private final Map<String, WSContextService> namedContextServices = Collections.synchronizedMap(new HashMap<>());
```
Azquelt: Add Javadoc, maybe use `ConcurrentHashMap`?

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+        } else if (namedContextServices.get(name) != contextService) {
+            // If name has changed, remove and re-add
+            namedContextServices.entrySet().removeIf(e -> e.getValue() == contextService);
+            namedContextServices.put(name, contextService);
+        }
+    }
+
+    protected void removeContextService(WSContextService contextService) {
+        namedContextServices.entrySet().removeIf(e -> e.getValue() == contextService);
+    }
```
Azquelt: This seems simpler:
```suggestion
    protected void removeContextService(WSContextService contextService) {
        namedContextServices.values().remove(contextService);
    }
```

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private String getContextServiceName(Map<String, Object> properties) {
+        Object name = properties.get("id");
```
Azquelt, line 125: Add Javadoc

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    @Override
+    public RMAsyncProvider getAsyncProvider(String contextServiceRef) {
+        return new NamedAsyncProvider(contextServiceRef);
+    }
+
+    @SuppressWarnings("unchecked")
+    private RMContext captureContext(String contextServiceName) {
```
Azquelt, line 155: Add Javadoc

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+            return new RMContextImpl(defaultContextService.captureThreadContext(null));
+        } else {
+            if (TraceComponent.isAnyTracingEnabled() && tc.isDebugEnabled()) {
+                Tr.debug(this, tc, "Capturing context with build-in context service");
+            }
+            return new RMContextImpl(builtInContextService.captureThreadContext(null, DEFAULT_CONTEXT_PROVIDERS));
+        }
+    }
+
+    private class NamedAsyncProvider implements RMAsyncProvider {
```
Azquelt, line 185: Add Javadoc

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    private static class RMContextImpl implements RMContext {
+
+        private final ThreadContextDescriptor context;
+
+        public RMContextImpl(ThreadContextDescriptor context) {
+            this.context = context;
+        }
```
Azquelt, line 223: Add Javadoc

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/interfaces/QuiesceParticipant.java
```diff
@@ -0,0 +1,18 @@
...
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.reactive.messaging.internal.interfaces;
+
+/**
+ *
+ */
+public interface QuiesceParticipant {
```
Azquelt: Add Javavdoc

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/interfaces/QuiesceRegister.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.reactive.messaging.internal.interfaces;
+
+public interface QuiesceRegister {
+
+    void register(QuiesceParticipant participant);
+
+    void remove(QuiesceParticipant participant);
+}
```
Azquelt: Add Javadoc

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/interfaces/RMAsyncProviderFactory.java
```diff
@@ -0,0 +1,18 @@
...
+ *******************************************************************************/
+package io.openliberty.microprofile.reactive.messaging.internal.interfaces;
+
+/**
+ *
+ */
+public interface RMAsyncProviderFactory {
+
+    RMAsyncProvider getAsyncProvider(String contextServiceRef);
+}
```
Azquelt: Add Javadoc, including how to obtain an instance.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/QuiesceRegisterImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import com.ibm.wsspi.kernel.service.utils.ServerQuiesceListener;
+
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.QuiesceParticipant;
+import io.openliberty.microprofile.reactive.messaging.internal.interfaces.QuiesceRegister;
+
+/**
+ *
+ */
+@Component(configurationPolicy = ConfigurationPolicy.IGNORE)
+public class QuiesceRegisterImpl implements ServerQuiesceListener, QuiesceRegister {
```
Azquelt: Add Javadoc

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/QuiesceRegisterImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final Set<QuiesceParticipant> participants = new HashSet<>();
```
Azquelt: I think this should be a `ConcurrentHashMap` or a synchronized set to ensure thread safety.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/QuiesceRegisterImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public void serverStopping() {
+        for (QuiesceParticipant participant : participants) {
+            participant.quiesce();
+        }
```
Azquelt: Consider thread safety here also.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/interfaces/RMAsyncProvider.java
```diff
@@ -0,0 +1,44 @@
...
+ *******************************************************************************/
+package io.openliberty.microprofile.reactive.messaging.internal.interfaces;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * Provides the services needed for Reactive Messaging components to do asynchronous tasks
+ */
+public interface RMAsyncProvider {
```
Azquelt: Doc how to obtain an instance.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/interfaces/RMContext.java
```diff
@@ -0,0 +1,28 @@
...
+ * http://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package io.openliberty.microprofile.reactive.messaging.internal.interfaces;
+
+/**
+ * Represents a context which can be applied to a thread
+ */
+public interface RMContext {
```
Azquelt: Doc how to obtain an instance.

Azquelt (commented):
Noticed some missing doc and a possible thread-safety issue.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,181 @@
...
+        return new NamedAsyncProvider(contextServiceRef);
+    }
+
+    @SuppressWarnings("unchecked")
+    private RMContext captureContext(String contextServiceName) {
+        WSContextService namedContextService = null;
+        if (contextServiceName != null) {
+            namedContextService = namedContextServices.get(contextServiceName);
+        }
+
```
Azquelt: If a context service is not found for a given name, (i.e. `namedContextServices.get` reutrns `null`) no error or warning is reported.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * Updates {@link #namedContextServices} when a new service is added
+     */
+    @Reference(service = WSContextService.class, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY, cardinality = ReferenceCardinality.MULTIPLE)
+    protected void addContextService(WSContextService contextService, Map<String, Object> properties) {
+        String name = getContextServiceName(properties);
+        if (name != null) {
+            namedContextServices.put(name, contextService);
+        }
+    }
```
njr-11: This seems inefficient to be tracking and receiving notifications for every configured contextService, including the ones you don't need and will never use.  Does this have any advantage over simply looking for the one you want in the service registry when you need it and tracking only that instance?
Also, if you keep it this way, it seems you are later manually filtering out the instances without an id. You could write a target filter that requires a non-null id, and then OSGi will only give you the ones with ids and you can remove the manual code.

njr-11 (commented):
I added a few comments/questions.

Also, I should point out that server.xml isn't the only way for users to define a ContextService. As of concurrent-3.0, they can also define a ContextService within the application by using `@ContextServiceDefinition` or `<context-service>` in the deployment descriptor.

```
@ContextServiceDefinition(name = "java:module/env/MyContextService", ...
```

When a user does that, Liberty writes equivalent configuration via ConfigAdmin. The resulting configuration will have
  name = "java:module/env/MyContextService"
  application = "MyApp" // whichever app it came from
  module = "MyModule" // whichever module it came from
  id = "application[MyApp]/module[MyModule]/contextService[java:module/env/MyContextService]"
  ...
(You might want to double check the above before relying on it because I wrote the example from memory)
The main thing to point out here is that although it has a generated id, the user doesn't know that id. Instead, the user knows it by the name they gave it and which application, module, and component they defined it in.
The solution that you have looks like it will sort of work with this pattern (whether you intended it to or not) but has a gap where the user probably doesn't know which value to supply as context.service

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
Azquelt: My thinking was:
- we need to find a context service by name regularly (usually at least once per message processed)
  - we need to look it up every time we use it because context services defined in the server.xml can change dynamically
- the total number of context services will be small
- context services don't come and go very often

I assumed (without testing) that it would be quicker to maintain a map from name to context service than to do an OSGi service lookup every time we wanted to use it. I wasn't aware of an easy way to do a lookup the first time we need a context service with a given name and then track it to see if it goes away.

I guess we could create a `ServiceTracker` every time we need a service with a given name and maintain a map of name to corresponding `ServiceTracker`? The tracker could remove itself from the map if the service goes away. I was assuming that the total number of context services would be small so the overhead of maintaining the map would be insignificant and didn't consider doing something more complex.

The suggestion to use a filter to ignore services without an ID is a reasonable one.

Azquelt:
> Also, I should point out that server.xml isn't the only way for users to define a ContextService. As of concurrent-3.0, they can also define a ContextService within the application by using `@ContextServiceDefinition` or `<context-service>` in the deployment descriptor.
> 
> ```
> @ContextServiceDefinition(name = "java:module/env/MyContextService", ...
> ```
> 
> When a user does that, Liberty writes equivalent configuration via ConfigAdmin. The resulting configuration will have
> ```
> name = "java:module/env/MyContextService"
> application = "MyApp" // whichever app it came from
> module = "MyModule" // whichever module it came from
> id = "application[MyApp]/module[MyModule]/contextService[java:module/env/MyContextService]"
> ...
> ```
> (You might want to double check the above before relying on it because I wrote the example from memory)
>
> The main thing to point out here is that although it has a generated id, the user doesn't know that id. Instead, the user knows it by the name they gave it and which application, module, and component they defined it in. The solution that you have looks like it will sort of work with this pattern (whether you intended it to or not) but has a gap where the user probably doesn't know which value to supply as context.service

Yeah, I didn't consider that and wasn't assuming that the application-defined ContextServices would be available as OSGi services. Supporting application-defined context services wasn't included in the design, though I did have a think about it. I assumed I'd need to look up the `ContextService` from JNDI and have a second implementation of `RMAsyncProvider` which could capture context using a `ContextService`.

We're always looking up a context service within the context of an application, so we could extend the definition of the `context.service` parameter so that it can be used to select an application-defined `ContextService` from the current application.

Is that something we need to consider doing here?

njr-11:
> Is that something we need to consider doing here?

The most important thing will be to ensure it doesn't partly work, after which we will be stuck with needing to preserve that behavior. It should either be completely disallowed or properly supported.

It looks like you very nearly have it supported already but would need to figure out what the correct externals of it are. We would either need to document the way of forming the id that can be used, or you would need to allow a JNDI name to be configured as the context.service and then either obtain the application/module/component if needed from the J2EEName on the thread, or if that isn't possible then provide a way for the user to supply the application/module/component.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
njr-11: I made the comment because it seemed odd to me to be collecting and tracking extra services that aren't needed, but your assumptions here are valid and so your approach does make sense.  Most other users of ContextService have server.xml config (or app config) that directly references a single instance or defaults, in which case the dependency can be enforced in OSGi on the one instance that is needed.  You have a more complex situation with the ability to receive requests for multiple different context services.

njr-11:
I don't want to hold up progress on this with my comment about application-defined ContextServices.  One option would be to temporarily force that code path to fail, allowing that work to be split out and done/decided separately.

Azquelt:
> The most important thing will be to ensure it doesn't partly work, after which we will be stuck with needing to preserve that behavior. It should either be completely disallowed or properly supported.

Yes, that's a good point.

> One option would be to temporarily force that code path to fail, allowing that work to be split out and done/decided separately.

I will do this for now by excluding services where the `application` and `module` properties are set. I'll raise another issue to decide whether to leave it like that or implement support for application-defined context services.

njr-11:
> will do this for now by excluding services where the `application` and `module` properties are set. I'll raise another issue to decide whether to leave it like that or implement support for application-defined context services.

For application-defined context services that are in the java:app scope, only the `application` property will be present, not `module`.

For application-defined context services that are in the java:global scope, neither `application` nor `module` properties will be present, although in this case, if someone puts in the java:global JNDI name as the context.service value, then I expect it will work.

dev/io.openliberty.microprofile.reactive.messaging.internal/src/io/openliberty/microprofile/reactive/messaging/internal/RMAsyncProviderFactoryImpl.java
```diff
@@ -130,6 +130,16 @@ private String getContextServiceName(Map<String, Object> properties) {
...
     }
 
+    /**
+     * Returns whether a context service is one defined within an application using {@code @ContextServiceDefinition}.
+     *
+     * @param properties the service properties of the {@code WSContextService} service to check
+     * @return {@code true} if the service properties are for an application-defined context service, otherwise {@code false}
+     */
+    private boolean isApplicationDefinedContextService(Map<String, Object> properties) {
+        return properties.containsKey("application") && properties.containsKey("module");
```
njr-11: Identify application-defined context services in java:global, java:app, and java:comp as well as those in java:module
```suggestion
        return !"file".equals(properties.get("config.source"));
```

Azquelt: Thanks, I'll squash this in

njr-11 (approved):

Azquelt:
#libby
#build

PR #27136, opened 2023-12-11
Set OpenAPI update workflow permissions

### LINKED ISSUE #27137: Set permissions on OpenAPI Update action
Currently, the OpenAPI Action is reporting as having no permissions set, so uses the permissions on the github token which likely will be greater then is required depending on who makes the request.

abutch3r:
Currently, the OpenAPI workflow does not define its permissions, therefore it inherits whatever token it uses has, which is more then it needs.

Add deny-all permissions to the workflow and add `contents` and `pull-request permissions` to the openapi job so if any new jobs are added they need to add their own permissions. As the job creates new branches and PRs it requires write access to both

Update checkout and pull-request actions to latest to resolve new deprecations 

Fixes [#27137](https://github.com/OpenLiberty/open-liberty/issues/27137)

### DIFF
--- a/.github/workflows/update-openapi-ui.yml
+++ b/.github/workflows/update-openapi-ui.yml
@@ -4,8 +4,13 @@ name: Update OpenAPI UI dependencies
 
 on: workflow_dispatch
 
+permissions: {}
+
 jobs:
   update-openapi-ui:
+    permissions:
+      contents: write
+      pull-requests: write
     name: Update OpenAPI UI Dependencies
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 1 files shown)


abutch3r:
#build

.github/workflows/update-openapi-ui.yml
```diff
@@ -51,7 +56,7 @@ jobs:
         git commit -m "Update Swagger UI to $SWAGGER_UI_VERSION [auto]"
         git push
     - name: Create Pull Request
-      uses: peter-evans/create-pull-request@18f7dc018cc2cd597073088f7c7591b9d1c02672 #3.14.0
+      uses: peter-evans/create-pull-request@v5.0.2
```
Azquelt: It was brought up before that using a version allows the author of the action to make nefarious changes to that version which we would then consume.

abutch3r: Have replaced version with the commit sha

abutch3r:
#libby

Azquelt (approved):
Looks sensible to me, thanks.

abutch3r:
#build

PR #27140, opened 2023-12-11
static metamodel with Attribute subtypes

njr-11:
Update the code that experiments with static metamodel to include subtypes


njr-11:
#build

KyleAure (approved):

PR #27148, opened 2023-12-11
Update deployment descriptors for Concurrency 3.1

KyleAure:
Updates our deployment descriptors for new values added for Concurrency 3.1 from the following sources: 
- `virtual=true` annotation atrribute (https://github.com/jakartaee/concurrency/pull/349)
- `qualifiers` annotation attribute (https://github.com/jakartaee/concurrency/pull/348)
- Deployment Descriptor schema (https://github.com/jakartaee/jakartaee-schemas/pull/44)

I think I'm most of the way there just have some questions tagged with //TODO that need some further review.
Would appreciate some feedback @tbitonti.

### DIFF
--- a/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ContextService.java
+++ b/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ContextService.java
@@ -36,7 +36,7 @@ public interface ContextService extends JNDIEnvironmentRef, Describable {
     /**
      * @return &lt;qualifier&gt; elements as a read-only list
      */
-    String[] getQualifier();
+    String[] getQualifiers();
 
     /**
      * @return &lt;property&gt; elements as a read-only list
--- a/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ContextService.java
+++ b/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ContextService.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2021,2022 IBM Corporation and others.
+ * Copyright (c) 2021, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -32,6 +32,11 @@ public interface ContextService extends JNDIEnvironmentRef, Describable {
      * @return &lt;unchanged&gt; elements as a read-only list
      */
     String[] getUnchanged();
+
+    /**
+     * @return &lt;qualifier&gt; elements as a read-only list
+     */
+    String[] getQualifier();
 
     /**
      * @return &lt;property&gt; elements as a read-only list
--- a/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedExecutor.java
+++ b/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedExecutor.java
@@ -62,7 +62,7 @@ public interface ManagedExecutor extends JNDIEnvironmentRef, Describable {
     /**
      * @return &lt;qualifier&gt; elements as a read-only list
      */
-    String[] getQualifier();
+    String[] getQualifiers();
 
     /**
      * @return &lt;property&gt; elements as a read-only list
--- a/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedExecutor.java
+++ b/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedExecutor.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2021,2022 IBM Corporation and others.
+ * Copyright (c) 2021, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -17,7 +17,7 @@ import java.util.List;
 /**
  * Represents &lt;managed-executor&gt;.
  */
-public interface ManagedExecutor extends JNDIEnvironmentRef, Describable {   
+public interface ManagedExecutor extends JNDIEnvironmentRef, Describable {
     /**
      * @return &lt;context-service-ref&gt;, or null if unspecified
      */
@@ -48,7 +48,24 @@ public interface ManagedExecutor extends JNDIEnvironmentRef, Describable {
     boolean isSetMaxAsync();
 
     /**
+     * @return &lt;virtual&gt; if specified
+     * @see #isSetVirutal
+     */
+    boolean isVirtual();
+
+    /**
+     * @return true if &lt;virtual&gt; is specified
+     * @see #isVirtual
+     */
+    boolean isSetVirtual();
+
+    /**
+     * @return &lt;qualifier&gt; elements as a read-only list
+     */
+    String[] getQualifier();
+
+    /**
      * @return &lt;property&gt; elements as a read-only list
      */
-    List<Property> getProperties();    
+    List<Property> getProperties();
 }
--- a/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedScheduledExecutor.java
+++ b/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedScheduledExecutor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021,2022 IBM Corporation and others.
+ * Copyright (c) 2021, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -62,7 +62,7 @@ public interface ManagedScheduledExecutor extends JNDIEnvironmentRef, Describabl
     /**
      * @return &lt;qualifier&gt; elements as a read-only list
      */
-    String[] getQualifier();
+    String[] getQualifiers();
 
     /**
      * @return &lt;property&gt; elements as a read-only list
--- a/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedScheduledExecutor.java
+++ b/dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedScheduledExecutor.java
@@ -4,7 +4,7 @@
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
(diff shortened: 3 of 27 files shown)


dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedExecutor.java
```diff
@@ -47,8 +47,25 @@ public interface ManagedExecutor extends JNDIEnvironmentRef, Describable {
...
+    /**
+     * @return true if &lt;virtual&gt; is specified
+     * @see #isVirtual
+     */
+    boolean isSetVirtual();
+
+    /**
+     * @return &lt;qualifier&gt; elements as a read-only list
+     */
+    String[] getQualifier();
```
njr-11: The code would read better if we can make this and similar places plural:
`String[] getQualifiers();`
I don't know if that is allowed or if the names need to match exactly, but it looks like a similar pattern is used on the following lines using the plural form `getProperties` to obtain multiple `property` elements

dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedScheduledExecutor.java
tbitonti, line 1: Needs copyright update.

dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ManagedThreadFactory.java
tbitonti, line 1: Needs copyright update.

dev/io.openliberty.concurrent.internal/resources/io/openliberty/concurrent/internal/resources/CWWKCMessages.nlsprops
```diff
@@ -45,3 +45,7 @@ CWWKC1203.duplicate.context.useraction=Update the application, libraries, or bot
 CWWKC1204.not.serializable=CWWKC1204E: Cannot create a serializable contextual proxy that propagates thread context {0} which is not serializable.
 CWWKC1204.not.serializable.explanation=Third-party thread context types are not compatible with serialization and cannot be configured to be propagated for serializable contextual proxies.
 CWWKC1204.not.serializable.useraction=Update the application to request a non-serializable contextual proxy or ensure that no third-party context types are configured to be propagated.
+
+CWWKC1205.qualifier.class.not.found=CWWKC1205E: Qualifier annotation {0}, which is defined by the application deployment descriptor, could not be loaded by application classloader.
```
jantley-ibm: Would it be OK to instead say this?
The {0} qualifier annotation
IBM Style is to place the variable before the noun object. If the variable is such that it makes the sentence difficult to read or understand, we can leave it as is.

Also, recommend changing
...could not be loaded by application classloader.
to
...could not be loaded by the application classloader.

dev/io.openliberty.concurrent.internal/resources/io/openliberty/concurrent/internal/resources/CWWKCMessages.nlsprops
```diff
@@ -45,3 +45,7 @@ CWWKC1203.duplicate.context.useraction=Update the application, libraries, or bot
...
+CWWKC1205.qualifier.class.not.found.explanation=The <qualifier> element(s) configured via application.xml or web.xml contained a class name which could not be loaded.
```
jantley-ibm: 1. Is it necessary to place the word qualifier in angle brackets?
2. Do not use (s) to indicate that it can be singular or plural
3. Do not use the word via.
Suggested rewrite:
Qualifier elements that are configured by the application.xml file or the web.xml file contain a class name that could not be loaded.

dev/io.openliberty.concurrent.internal/resources/io/openliberty/concurrent/internal/resources/CWWKCMessages.nlsprops
```diff
@@ -45,3 +45,7 @@ CWWKC1203.duplicate.context.useraction=Update the application, libraries, or bot
...
+CWWKC1205.qualifier.class.not.found.useraction=Verify the <qualifier> element(s) were configured with a fully qualified class name, and the class is present within the application.
```
jantley-ibm: Suggested rewrite:
Verify that the qualifier elements are configured with a fully qualified class name, and that the class is present within the application.

jantley-ibm (commented):
ID review comments attached. Let me know if you have questions or concerns about suggested changes. When we have agreed on changes, I will add the "ID reviewed" label.

dev/com.ibm.ws.javaee.ddmodel/test/com/ibm/ws/javaee/ddmodel/app/AppTest.java
tbitonti, line 1: Re: The TODO: 'handleChild' *could* look for attributes which are not valid for the current schema version.  A review of several of the implementations shows that we aren't doing that.  The implementations seem to usually accept attributes from all schema levels, regardless of whatever level is being used.  This implies that the runtime has to know when to ignore particular attributes.

dev/com.ibm.ws.javaee.ddmodel/src/com/ibm/ws/javaee/ddmodel/app/ApplicationDDParser.java
tbitonti, line 1: 'getMaxImplemented' is used by a reporting test class:
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.javaee.ddmodel/test/com/ibm/ws/javaee/ddmodel/DDParserSpecList.java
We should update this to what is actually implemented, and update the spec-list test data.  However, this is only for reporting, not anything functional in regards to what we support.

dev/com.ibm.ws.javaee.ddmodel/src/com/ibm/ws/javaee/ddmodel/web/WebAppDDParser.java
tbitonti, line 1: See the comment on ApplicationDDParser re: getMaxImplemented.

dev/com.ibm.ws.javaee.dd.common/src/com/ibm/ws/javaee/dd/common/ContextService.java
tbitonti, line 1: The plural 'getQualifiers' seems more consistent with naming for other attributes.

tbitonti (changes requested):
I reviewed the parser changes.  I did not look at the runtime updates.  I see a few very minor updates (missing copyright, a possible name change).  The function looks good.

KyleAure:
@njr-11 @tbitonti @jantley-ibm Updated based off your feedback.

dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ContextServiceDefinitionBinding.java
```diff
@@ -110,7 +110,7 @@ public void merge(ContextServiceDefinition annotation, Class<?> instanceClass, M
 
         unchanged = mergeAnnotationValue(unchanged, XMLunchanged, annotation.unchanged(), KEY_UNCHANGED, DEFAULT_UNCHANGED);
 
-        qualifier = mergeAnnotationValue(qualifier, XMLqualifer, annotation.qualifiers(), KEY_QUALIFIER, DEFAULT_QUALIFIER);
+        qualifiers = mergeAnnotationValue(qualifiers, XMLqualifers, annotation.qualifiers(), KEY_QUALIFIERS, DEFAULT_QUALIFIERS);
```
njr-11, line 113: I don't think this will behave correctly for a more complex type like `Class[]`.
mergeAnnotationValue in InjectionBinding does `newValue.equals(oldValue)` which will be an instance comparison for arrays.  It looks like there is a special variant, mergeAnnotationProperties, to handle merging for the properties, and it is even more complex in that it appears to combine individual properties from one list into the other rather than just overwriting.  Should the qualifiers behave similarly where the merge is a combination of the annotation values and `<qualifier>` elements?  Or should the set of the latter replace the former?

This same comment will apply to all 4 resource types.

njr-11 (commented):
Added a comment about merging.  I'm fine if you defer this to a separate pull.

dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ContextServiceDefinitionBinding.java
KyleAure: Haven't done any builds on this yet and still waiting on reviews.  I think I'll tackle this now.

posmith:
Support message review completed.

KyleAure:
Force pushed to resolve merge conflicts.

dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ContextServiceDefinitionBinding.java
KyleAure: @njr-11 Resolved this by adding an extra check when we merge XML values under this commit [`b181523` (#27148)](https://github.com/OpenLiberty/open-liberty/pull/27148/commits/b181523bd1b1263ccf3ce1a2894b7d1a1b07f7c0)

dev/com.ibm.ws.javaee.ddmodel/src/com/ibm/ws/javaee/ddmodel/common/XSDTokenType.java
```diff
@@ -45,6 +45,20 @@ public List<String> getList() {
...
+                stringArray[i] = list.get(i).getValue();
+            }
+            return stringArray;
+        }
+
+        public static final String[] EMPTY_ARRAY = new String[] {};
+
+        public static String[] getEmptyArray() {
+            return EMPTY_ARRAY;
+        }
```
njr-11: Optional review comment - You could remove this method and just directly access the constant, which would make it more clear that you are getting the same instance every time.

dev/com.ibm.ws.javaee.ddmodel/test/com/ibm/ws/javaee/ddmodel/DDJakarta11Elements.java
```diff
@@ -0,0 +1,424 @@
...
+*/
+public class DDJakarta11Elements {
+
+    public static List<String> names(String... name0) {
+        List<String> names = new ArrayList<String>(6);
+        for (String name : names) {
+            names.add(name);
+        }
+        return names;
+    }
```
njr-11: The parameter to this method isn't used and the loop is a no-op.  Should it be this instead?
```suggestion
    public static List<String> names(String... name0) {
        List<String> names = new ArrayList<String>(name0 == null ? 0 : name0.length);
        if (name0 != null)
            for (String name : name0) {
                names.add(name);
            }
        return names;
    }
```

dev/io.openliberty.concurrent.internal/resources/io/openliberty/concurrent/internal/resources/CWWKCMessages.nlsprops
```diff
@@ -46,7 +46,9 @@ CWWKC1204.not.serializable=CWWKC1204E: Cannot create a serializable contextual p
 CWWKC1204.not.serializable.explanation=Third-party thread context types are not compatible with serialization and cannot be configured to be propagated for serializable contextual proxies.
 CWWKC1204.not.serializable.useraction=Update the application to request a non-serializable contextual proxy or ensure that no third-party context types are configured to be propagated.
 
-
+CWWKC1205.qualifier.class.not.found=CWWKC1205E: The {0} qualifier annotation, which is defined by a deployment descriptor, could not be loaded by the application classloader.
```
njr-11:
```suggestion
CWWKC1205.qualifier.class.not.found=CWWKC1205E: The {0} qualifier annotation class, which is defined by a deployment descriptor, could not be loaded by the application classloader.
```

dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ContextServiceDefinitionBinding.java
```diff
@@ -198,17 +225,33 @@ static final String toString(ContextServiceDefinition anno) {
...
         return none ? "None" : Arrays.toString(list);
     }
+
+    @Trivial
+    private static final Class<?>[] toQualifierClassArray(String[] classList) throws IllegalArgumentException {
+        Class<?>[] clazzArray = new Class<?>[classList.length];
+        for (int i = 0; i < classList.length; i++) {
+            try {
+                //TODO is there a certain classloader I should be using to load this class? ApplicationClassLoader?
+                clazzArray[i] = Class.forName(classList[i]);
```
njr-11, line 249: Yes, we need to be careful with making it a requirement to load classes if we aren't certain we can have the right class loader when doing it.  There might be different class loaders between modules and for the application.  If we don't have access to the correct class loader here, another option would be to perform the merge as String class names, to be loaded later.  I'm not sure what is available.  You'll need to work with Tom on this part.

dev/com.ibm.ws.javaee.ddmodel/test/com/ibm/ws/javaee/ddmodel/DDJakarta11Elements.java
KyleAure: This was copied over from the EE 10 element object. Fixing in next commit.

njr-11 (approved):

PR #27166, opened 2023-12-13
CIOrch - Removing Schedules from TestPBs 

websterc87:
we have enough data without the scheduled builds. convert scheduled PB to be on manual request rather than all the time.

I have also updated to use the latest groups concept so that OL builds appear on the right page in the new cognitive UI.

No need for a PB as its only changing the config files for CI Orchestrator and is not exercises by a PB.


dave-waddling (approved):

wraschke:
#libby

PR #27187, opened 2023-12-15
Update MP OpenAPI 3.1 TCK to run against different MP Config versions

abutch3r:
By default OpenAPI 3.1 will enable MP Config 3.0, so currently it does not test it against MP Config 3.1. so effectively not against MicroProfile 6.1.

Update server.xml to explicitly include Mp Config so that the right version is used for the repeats and add a repeat to the test to cover MicroProfile 6.0 and 6.1


abutch3r:
#build

tevans78 (approved):

PR #27195, opened 2023-12-15
split the concurrency cdi bundle between EE 10 and 11

njr-11:
The Concurrency/CDI integration bundle has a small amount of code that is common to both EE 10 and 11, but most of the code is specific to new features in Concurrency 3.1 in Jakarta EE 11, with much more code still to be added that will require accessing new API classes in Concurrency 3.1.  A better approach is to split into separate bundles, which is done under this pull.


njr-11:
@OpenLiberty/message-reviewer - none of the messages in this code are new. The files are just copied from one place to another due to some refactoring.

njr-11:
#build

njr-11:
#build
#libby

dev/io.openliberty.concurrent.internal.cdi/bnd.bnd
```diff
@@ -53,7 +53,7 @@ instrument.classesExcludes: io/openliberty/concurrent/internal/cdi/resources/*.c
   com.ibm.ws.kernel.service,\
   com.ibm.ws.threading,\
   io.openliberty.jakarta.annotation.2.1,\
-  io.openliberty.jakarta.cdi.4.0,\
-  io.openliberty.jakarta.concurrency.3.0,\
+  io.openliberty.jakarta.cdi.4.1,\
```
tevans78, line 54: If this bundle is now only for EE11, should the Java version be moved up to 21 and the explicit imports be removed?

dev/io.openliberty.concurrent.internal.cdi.4.0/resources/io/openliberty/concurrent/internal/cdi4/resources/CWWKCMessages.nlsprops
```diff
@@ -0,0 +1,43 @@
...
+#COMPONENTNAMEFOR CWWKC Jakarta EE Concurrency
+#
+# NLS_MESSAGEFORMAT_VAR
+#
+#   Strings in this file which contain replacement variables are processed by the MessageFormat 
+#   class (single quote must be coded as 2 consecutive single quotes ''). Strings in this file 
+#   which do NOT contain replacement variables are NOT processed by the MessageFormat class 
+#   (single quote must be coded as one single quote '). 
+
+# All messages must use the range CWWCK1400 to CWWCK1499
```
tevans78: This is still the message range used by `io.openliberty.concurrent.internal.cdi` so you can't re-use it here. Either the messages need to be in a common bundle or duplicated using a new, unique, range.

njr-11: That's a good point. I moved the messages to a common bundle under https://github.com/OpenLiberty/open-liberty/pull/27195/commits/3f2b926ddb227b5c0a5ea1f70ee8ff700c8ed5c8 
Please review again.

dev/io.openliberty.concurrent.internal.cdi/bnd.bnd
njr-11: Sure, I can include that in the pull. I updated it under the latest commit.

njr-11:
#libby
#build

posmith:
Support message review not required.

KyleAure (approved):

PR #27229, opened 2023-12-29
Shutdown Liberty directly when administrator issue server stop xxx --force option

YuLimin:
After for 10 * 1000 ms and shutdown Liberty directly when administrator issue server stop xxx --force option.

Especially there have some threads launched by customer self and not managed by Liberty server, Liberty can't not stop these threads smoothly and will hung.


hlhoots:
@YuLimin We cant accept this PR as it would break our Open Liberty Zero Migration policy as it changes the function of the --force command.  Issue will continue to be discussed in your Support ticket.

YuLimin:
> @YuLimin We cant accept this PR as it would break our Open Liberty Zero Migration policy as it changes the function of the --force command. Issue will continue to be discussed in your Support ticket.

I don't know what's the Open Liberty Zero Migration policy will be affected.

or we can add another option such **--shutdown** to shutdown Liberty Server which there have some threads launched by customer self and not managed by Liberty server and still running.

PR #27232, opened 2024-01-02
Exp jsfurl transformer

### LINKED ISSUE #27231: Remove IBMViewHandlerProxy.java, FacesUrlTransformer.java, and associated
We have a horrible situation where we've copied some weld code from ConversationAwareViewHandler.java into Liberty and now have a test making sure the upstream files haven't changed via checksums. 

We should fix this, and I believe I can do it by extending ConversationAwareViewHandler

benjamin-confino:
fixes #27231


benjamin-confino:
Further research suggests this is an unsolvable issue

PR #27238, opened 2024-01-02
Add Beta guards to Liberty-kafka connector new options

abutch3r:
fast.ack and context.service are new attributes for the liberty-kafka connector. However, this are not currently ready to GA and as such require beta guards.

If they are specified based on the RM specification they should be ignored as they outside of beta should not have any impact upon the operations.

For `fast.ack` this is applied at the point where it is checked to determine the kafka acknowledgemnt behavior.
For `context.service` this is applied when we attempt to look up the defined `context.service` and we set it to use the default provided. This does mean that for both RM 1.0 and 3.0 proper looks are still performed in relation to using the context service provided by the `concurrent` feature if it is enabled (which is always the case for 1.0) or the default Liberty context service.

Update relate kafka tests to set the beta flag.


abutch3r:
#build

abutch3r:
#libby
#build

abutch3r:
#libby
#build

abutch3r:
#libby

Azquelt (approved):

PR #27245, opened 2024-01-02
Remove daytrader7 from runJandex.bat

### LINKED ISSUE #27244: Remove daytrader7 references from runJandex.bat utility
This will remove references to struts 2.3.24.    Scans for the Apache Struts File Upload Exploit flag the runJandex.bat file.    This is not a vulnerability.   The runJandex.bat utility is an internal tool.   It is not shipped.   Furthermore, the tool is only using the struts jar as input; i.e data.   It is not running code in the struts JAR.   It's basically creating a jandex index of annotations in the JAR.  

While this is not a vulnerability, it is bothersome that it shows up in the scan, and the reference to daytrader 7 isn't needed.  If we simply remove the daytrader7 from the tool, it will no longer be flagged by a security scan.

jimblye:
fixes #27244


hlhoots (approved):
LGTM

PR #27246, opened 2024-01-02
Fix EJB remote FAT http port

tkburroughs:
http ports not configured, resulting in use of defaults that are not available on build systems
- update to use ports provided by build system


tkburroughs:
#build

brideck (approved):

PR #27255, opened 2024-01-03
Increase EJB Timer FAT delay tolerance

tkburroughs:
Accuracy of EJB timer schedule is diminished on slower VMs, especially over longer intervals
- increase the EJB timer delay tolerance for timer scheduled over longer intervals


tkburroughs:
#build

brideck (approved):

PR #27275, opened 2024-01-04
Updating MP Config 3.1 API maven coordinates to match version in maven central

### LINKED ISSUE #27125: Fix MP Config 3.1 API maven co-ords
In some places in the OL code the version is listed as `3.1.0` but the version in maven central is `3.1`. Although I suspect this is technically valid, it does cause some issues and so we should fix it.

https://github.com/OpenLiberty/open-liberty/blob/6bd279974d9af9a96e73971bc0bac1378b99bfb6/dev/com.ibm.websphere.appserver.features/visibility/private/io.openliberty.org.eclipse.microprofile.config-3.1.feature#L6

There may be other places that the version is listed??

jakub-pomykala:
- Updating the maven coordinates for MP Config 3.1. From 3.1.0 >> 3.1

Fixes #27125


jakub-pomykala:
#build

tevans78 (approved):

PR #27281, opened 2024-01-05
Create script to update OpenAPI UI with latest Swagger UI version

### LINKED ISSUE #27142: Convert the update-openapi-ui action into a shell script
We have been asked to remove the update-openapi-ui action. Given this was created to reduce the potential for mistakes we should convert this into a shell script which anyone can run

The shell script should in addition to running the commands from the action, perform the following steps
1. Ensure it is being build in a Node16 environment
2. Using a suitable version of npm (>=8)
3. anything else that the action used to setup for us

ryan-storey:
This script replaces the GitHub action we recently discontinued. 
Closes #27142

### DIFF
--- a/dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
+++ b/dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
@@ -15,7 +15,7 @@
 # Ensure that a Node 16 or later LTS environment is used
 node_version=`node -v | grep "^v"`
 version_number=${node_version:1:2}
-if [[ version_number -lt 16 ]] then
+if [[ version_number -lt 16 ]]; then
     echo "A Node 16 or later LTS environment must be used when running this script."
     exit 1
 else
--- a/dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
+++ b/dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
@@ -1,0 +1,68 @@
+#*******************************************************************************
+# Copyright (c) 2024 IBM Corporation and others.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License 2.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-2.0/
+# 
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
+#*******************************************************************************
+#!/usr/bin/env bash
+
+# Ensure that a Node 16 or later LTS environment is used
+node_version=`node -v | grep "^v"`
+version_number=${node_version:1:2}
+if [[ version_number -lt 16 ]] then
+    echo "A Node 16 or later LTS environment must be used when running this script."
+    exit 1
+else
+    if ! [ $((version_number % 2)) -eq 0 ]; then
+        echo "A Node 16 or later LTS environment must be used when running this script."
+        exit 1        
+    fi
+fi
+
+# Check version of npm is 8 or higher
+npm=`npm -v | sed 's/^[^0-9]*\([0-9]*\).*/\1/'`
+if [[ "${npm%.*}" -lt 8 ]]; then
+    echo "A version of npm 8 or higher must be used when running this script."
+fi
+
+# Check script is run from com.ibm.ws.openapi.ui directory
+if ! pwd | grep -q -E 'open-liberty/dev/com.ibm.ws.openapi.ui$' ; then
+    echo "This script must be executed from the open-liberty/dev/com.ibm.ws.openapi.ui directory."
+    exit 1
+fi
+
+# Checking script dependencies
+if ! which jq >/dev/null; then
+    echo "Please install jq to run this script."
+    exit 1
+fi
+
+# Find new Swagger UI version
+export SWAGGER_UI_VERSION=$(jq -r '.packages."node_modules/swagger-ui".version' ./swagger-ui/package-lock.json)
+echo "Swagger UI version: $SWAGGER_UI_VERSION"
+
+# Update OpenAPI UI dependencies
+cd swagger-ui
+npx --yes npm-check-updates -u
+npm update
+
+# Copy SCSS from Swagger UI source
+rm -fr swagger-ui-src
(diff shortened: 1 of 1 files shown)


dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
```diff
@@ -0,0 +1,77 @@
...
+fi
+
+# Check version of npm is 8 or higher
+npm=`npm -v | sed 's/^[^0-9]*\([0-9]*\).*/\1/'`
+if [[ "${npm%.*}" -lt 8 ]]; then
+    echo "A version of npm 8 or higher must be used when running this script."
+fi
+
+# Installing script dependencies
+brew list jq >/dev/null 2>&1 || brew install jq
```
abutch3r: Linux OSes don't have Brew as that is Mac specific, so this will need an OS gate around it. so if it is MACOS, do these steps. Otherwise you will need to have a look at either apt or similar linux based package managers to do the checks. jq can be validated easily given it is a command. I am less sure about coreutils and what utilities that provides as this might already be part of standard Linux distros

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
```diff
@@ -0,0 +1,77 @@
...
+# Check git username was provided as an argument
+if ! [ $# -eq 1 ]; then
+    echo "Missing git username argument."
+    exit 2
+else
+    git=$1
+fi
+
+# Ensure that a Node 16 environment is used
+if ! node -v | grep -q "^v16"; then
```
abutch3r: We should see if we can use Node 18 and 20. 16 is now EoL so saying we must be on 16 is restrictive unless we find the build process doesn't work with newer Node versions. So I would try with 18 and 20 to see if the generated result of the npm build process is different. if it isn't then we should change the check to be >16 and divisible by 2 as we want it built with an LTS version

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# Copy SCSS from Swagger UI source
+rm -fr swagger-ui-src
+git clone --depth 1 -b v${SWAGGER_UI_VERSION} -c advice.detachedHead=false https://github.com/swagger-api/swagger-ui.git swagger-ui-src
+rm src/style/original/*
+rm -r src/style/core
+cp swagger-ui-src/src/style/* src/style/original
+shopt -s globstar
+cd swagger-ui-src/src
+gcp --parents core/**/*.scss ../../src/style
```
Azquelt: What's `gcp`?

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
Azquelt: A script like this shouldn't install its own dependencies. It would be better to print out an error message about what's needed that isn't available (as you're already doing for `npm` and `node`)

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
```diff
@@ -0,0 +1,77 @@
...
+git reset --hard $remote/integration
+git merge $remote/integration
+
+# Find new Swagger UI version
+export SWAGGER_UI_VERSION=$(jq -r '.packages."node_modules/swagger-ui".version' ./swagger-ui/package-lock.json)
+echo "Swagger UI version: $SWAGGER_UI_VERSION"
+
+# Update OpenAPI UI dependencies
+git branch "update-swagger-ui-$SWAGGER_UI_VERSION"
+git switch "update-swagger-ui-$SWAGGER_UI_VERSION"
```
Azquelt: Don't do any of this. If you're expecting users to run this in their local workspace, you must not take any git actions that would affect their workspace. Leave the user to set up their branch and manually commit the result.

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
Azquelt: GNU `cp`?

Maybe replace this with `find`,`xargs` and regular `cp`.

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
```diff
@@ -0,0 +1,77 @@
...
+cd ../..
+rm -rf swagger-ui-src
+
+# Build OpenAPI UI
+npm run build -- --mode=production
+
+# Commit and push
+git add .
+git commit -m "Update Swagger UI to $SWAGGER_UI_VERSION [auto]"
+git push --set-upstream "git@github.com:$git/open-liberty.git" "update-swagger-ui-$SWAGGER_UI_VERSION"
```
Azquelt: Again, never do this for a script that a developer is supposed to run on their own workspace.

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
```diff
@@ -0,0 +1,69 @@
...
+npx --yes npm-check-updates -u
+npm update
+
+# Copy SCSS from Swagger UI source
+rm -fr swagger-ui-src
+git clone --depth 1 -b v${SWAGGER_UI_VERSION} -c advice.detachedHead=false https://github.com/swagger-api/swagger-ui.git swagger-ui-src
+rm src/style/original/*
+rm -r src/style/core
+cp swagger-ui-src/src/style/* src/style/original
+shopt -s globstar
```
Azquelt: This is no longer needed.

Azquelt (dismissed):

dev/com.ibm.ws.openapi.ui/update-openapi-ui.sh
```diff
@@ -0,0 +1,68 @@
...
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
+#*******************************************************************************
+#!/usr/bin/env bash
+
+# Ensure that a Node 16 or later LTS environment is used
+node_version=`node -v | grep "^v"`
+version_number=${node_version:1:2}
+if [[ version_number -lt 16 ]] then
```
abutch3r: needs a `;` after the `]]`

ryan-storey:
#libby #build #pbbeta

abutch3r (approved):
LGTM

PR #27284, opened 2024-01-05
context propagation to scheduled async methods

njr-11:
Thread context propagation to each execution of scheduled asynchronous methods.
This pull also includes handling for schedule components that are assign to empty (They are excluded from the schedule, except for the most granular unit of seconds which much be specified or left at its default value of {0}) and test coverage to verify that scheduled asynchronous methods that should have completed are not continuing to run additional executions.


njr-11:
#build

mswatosh (approved):

PR #27293, opened 2024-01-06
scheduled asynchronous methods on virtual threads

njr-11:
Run scheduled asynchronous methods on virtual threads by supplying and executor with virtual=true.
Because virtual=true on ManagedExecutorDefinition isn't implemented yet, this temporarily uses an internal setting on the concurrency policy to simulate it.


njr-11:
#build

njr-11:
#build

jmstephensgit (dismissed):

njr-11:
2 tests failed because I did not update them after changing how the virtual thread name is computed.  I added a commit to get them updated.
#build
#libby

njr-11:
rebased on to latest
#build
#libby

jmstephensgit (approved):

njr-11:
The personal build shows that the second commit cleared up the failures in the test case that I hadn't previously updated.

PR #27317, opened 2024-01-09
xmlBinding-4.0: Prevent test ordering from causing CNFE in JAXBContextTest

neuwerk:
This PR addresses the following failure: 
```
jakarta.xml.bind.JAXBException: Implementation of Jakarta XML Binding-API has not been found on module path or classpath.
- with linked exception:
[java.lang.ClassNotFoundException: org.eclipse.persistence.jaxb.XMLBindingContextFactory]
```
This is caused when the negative test cases run prior to the positive test cases, and prevent the right JAXBContext from being loaded for the positive cases.


neuwerk:
#build

BerksanAtes (approved):
Looks good to me!

PR #27321, opened 2024-01-09
first steps for managed thread factory with virtual=true

njr-11:
Simulate virtual=true for ManagedThreadFactory with an internal property.
Include a test case to cover the code path.


njr-11:
#build

mswatosh (approved):

PR #27333, opened 2024-01-10
Fix large string persistence for config

### LINKED ISSUE #27342: Persisted configurations cannot store long string values.
**Describe the bug**  
Server configuration elements which generate long string values cannot be persisted.

The problem occurs, in particular, when an application binding which has many users is specified in the server configuration.

The problem case was produced with a combination of approximately 2300 users and twenty security roles (see below).  Other "large" configuration elements will likely have the same problem.

**Steps to Reproduce**  
The problem is reproduced by creating a server that has a configuration with a large configuration element (in particular, having many users and security roles), then starting or restarting the server.

**Expected behavior**  
Large configuration elements should be handled without error.

**Diagnostic information:**  
 - OpenLiberty Version: all up to 24.0.0.2.
 - Affected feature(s): potentially all
 - Java Version: all
 - server.xml configuration: For example, with 2400 users and 20 security roles:
```
<server>
  <application-bnd>
    <security-role name="name1">
      <user name="${var.name}"/>
      <user name="name1"/>
      <user name="name2"/>
      ...
```
 - messages: The problem produces two FFDC, one for an initial failure to save a large value, and one for later failures to read the partially persisted configuration:
   - java.io.UTFDataFormatException com.ibm.ws.config.admin.internal.ConfigurationStorageHelper/ffdc_23.12.08_15.18.58.0.log
   - java.io.EOFException com.ibm.ws.config.admin.internal.ConfigurationStorageHelper 

**Additional context**  
The problem, while related to security configuration, is not a security vulnerability.

tbitonti:
Fix for the large strings issue.  Details per the issue.

### DIFF
--- a/dev/com.ibm.ws.config/src/com/ibm/ws/config/admin/internal/ConfigurationStorageHelper.java
+++ b/dev/com.ibm.ws.config/src/com/ibm/ws/config/admin/internal/ConfigurationStorageHelper.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2019 IBM Corporation and others.
+ * Copyright (c) 2019,2024 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -21,6 +21,7 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Array;
+import java.nio.charset.StandardCharsets;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -33,6 +34,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import com.ibm.websphere.ras.Tr;
+import com.ibm.websphere.ras.TraceComponent;
+import com.ibm.websphere.ras.annotation.Trivial;
 import com.ibm.ws.config.admin.ConfigID;
 import com.ibm.ws.config.admin.ConfigurationDictionary;
 import com.ibm.ws.config.admin.ExtendedConfiguration;
@@ -44,7 +48,7 @@ import com.ibm.wsspi.kernel.service.utils.SerializableProtectedString;
  * A helper class for storing and retrieving the persisted config information.
  * <p>
  * {@link #load(File, ConfigStorageConsumer)} / {@link #store(File, Collection)} format: <br>
- * byte - version, set to constant {@link #VERSION} <br>
+ * byte - version <br>
  * int - number of configurations <br>
  * followed by the list of conditions as loaded by {@link #load(DataInputStream, Set, Set, ConfigurationDictionary)}
  * <p>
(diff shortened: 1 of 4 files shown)


tbitonti:
#build

jimblye (dismissed):

tbitonti:
* Test Summary
  * All tests in "open-liberty/dev/com.ibm.ws.config/test"

---

com.ibm.ws.config.admin.internal.ConfigurationStorageHelperTest:

- Renamed "testStoreLoadCycle()" to "testStoreLoadCycleShort()"
- Added "testStoreLoadCycleLong()"
  - Split into cases using the old "short string" format and a using the new "long string format".

- Added "testStoreLoadCycleDowngrade()"
  - Test attempt to read the new "long string format" using the "short string" reader.
- Added "testStoreLoadCycleUpgrade()"
  - Test attempt to read the old "short string format" using the "long string" reader.

- Renamed "testStoreLoadCycleWithNulls()" to "testStoreLoadCycleWithNullsShort()"
- Added "testStoreLoadCycleWithNullsLong()"
  - Split into cases using the old "short string" format and a using the new "long string format".

- Renamed "testMultiConfigStore()" to "testMultiConfigStoreShort()"
- Added "testMultiConfigStoreLong()"
  - Split into cases using the old "short string" format and a using the new "long string format".

- Renamed "testMapStore()" to "testMapStoreShort()"
- Added "testMapStoreLong()"
  - Split into cases using the old "short string" format and a using the new "long string format".

---

com.ibm.ws.config.xml.internal.XMLConfigParserTest:

- Updated "testIncludesWithVariables()"
  - Fixed an unrelated problem where URLs are incorrectly compared when running in a BASH shell on WINDOWS, which caused a false test failure.

---

tbitonti:
Lots of build failures ... retrying.

tbitonti:
#build

anjumfatima90:
I am going to defer the review to @tjwatson. Kindly review this Tom!

dev/com.ibm.ws.config/src/com/ibm/ws/config/admin/internal/ConfigurationStorageHelper.java
```diff
@@ -70,14 +73,97 @@
...
+         * Answer the key of the configuration data. This is usually a particular value of the
+         * data.
+         *
+         * @param config The configuration from which to obtain a key value.
+         * @return A key used to store the configuration.
+         */
+        K getKey(C config);
     }
 
-    private static final byte VERSION = 0;
```
tjwatson, line 80: This has over complicated the solution.  The original idea was that if the persistent format changed then you would monotonically increment this constant by 1.  Then if the new version of config encounters a version that does not match it simply throws away the persistent data and starts over.  But this exposes a greater issue which causes the behavior to fail if something like that happens.

There are two types of issues.  What happens if load fails for any reason (file does not exists, corrupted file etc.) and what happens if it encounters a version of the persistent data it doesn't know how to read.  The first case will probably throw an `IOException` out of `ConfigurationStorageHelper.load(File, ConfigStorageConsumer<String, ExtendedConfigurationImpl>)`.  That is then caught with and auto FFDC is logged and an empty `Map` is used.  The second case will return an empty `Map`.  An empty `Map` here really throws things off when launching from a cached state because we will start without any persisted configuration objects and we will bypass re-populating it with the ServerConfiguration content.

`ConfigurationStorageHelper.load(File, ConfigStorageConsumer<String, ExtendedConfigurationImpl>)` should be changed to throw an IOException in all cases where it could not read the stored config.  That is for all these cases:

- The persistent file does not exists
- The persistent version does not match the expected version to read
- Any other IO exception while reading the persisted file

Then `com.ibm.ws.config.admin.internal.ConfigurationStore.loadConfigurationDatas(File)` should catch the `IOException` and do something that will force `com.ibm.ws.config.xml.internal.SystemConfiguration.start()` to re-process the configuration such that it enters the code at:
https://github.com/OpenLiberty/open-liberty/blob/095b44578b63d7f129edb7ca38c62088139420d9/dev/com.ibm.ws.config/src/com/ibm/ws/config/xml/internal/SystemConfiguration.java#L201-L204

If this is done then I believe the solution here becomes much more simple because you can choose to write all Strings in the same way without having to invent all these new types of short/long string types.  It will also fix an existing issue with what happens when trying to load a corrupted persistent store file.

tjwatson (changes requested):

dev/com.ibm.ws.config/src/com/ibm/ws/config/admin/internal/ConfigurationStorageHelper.java
tbitonti: Discussed in sametime.  Decided:
*The version change should be removed.  Version is not serving the original purpose and does not need to be updated.
*The new short_string format is not necessary; string and long_string should be used.
*Because the version is not changing, the error handling code path does not need to be updated.
Changes will be made with a following commit, including test changes.

tbitonti:
#build

tbitonti:
#build

tjwatson (dismissed):

tbitonti:
#build

tbitonti:
#build

tbitonti:
#build

tbitonti:
#build

jimblye (approved):

tbitonti:
#libby

PR #27340, opened 2024-01-11
Concurrency 3.1 injection binding for qualifier and virtual

KyleAure:
Closes #27148 - original PR was split.

### DIFF
--- a/dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ManagedExecutorDefinitionBinding.java
+++ b/dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ManagedExecutorDefinitionBinding.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2021,2022 IBM Corporation and others.
+ * Copyright (c) 2021, 2023 IBM Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-2.0/
- * 
+ *
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
@@ -13,6 +13,7 @@
 package io.openliberty.concurrent.internal.processor;
 
 import java.lang.reflect.Member;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ManagedExecutorDefinitionBinding.java
```diff
@@ -80,10 +101,12 @@ protected JNDIEnvironmentRefType getJNDIEnvironmentRefType() {
...
         if (trace)
-            Tr.entry(this, tc, "merge", toString(annotation), instanceClass, member,
+            Tr.entry(this, tc, "merge", toString(annotation, eeVersion), instanceClass, member,
                      (XMLContextServiceRef ? "(xml)" : "     ") + "contextServiceRef: " + contextServiceJndiName + " << " + annotation.context(),
                      (XMLHungTaskThreshold ? "(xml)" : "     ") + "hungTaskThreshold: " + hungTaskThreshold + " << " + annotation.hungTaskThreshold(),
-                     (XMLMaxAsync ? "         (xml)" : "              ") + "maxAsync: " + maxAsync + " << " + annotation.maxAsync());
+                     (XMLMaxAsync ? "         (xml)" : "              ") + "maxAsync: " + maxAsync + " << " + annotation.maxAsync(),
+                     (XMLvirtual ? "          (xml)" : "               ") + "virtual: " + virtual + " << " + (eeVersion >= 11 ? annotation.virtual() : "Unspecified"),
+                     (XMLqualifers ? "        (xml)" : "            ") + "qualifiers: " + toString(qualifiers) + " << " + (eeVersion >= 11 ? toString(annotation.qualifiers()) : "Unspecified"));
```
njr-11, line 109: Even though we don't support these values prior to EE 11, would it be valuable to see if anything unexpected is coming in for them rather than showing Unspecified?  I'm wondering if Unspecified could mislead someone who is debugging issues

njr-11 (approved):

dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ManagedExecutorDefinitionBinding.java
KyleAure: If a user compiled their application using EE 11 and used the qualifiers attribute on their service annotation, but then ran with EE 10 I would assume we'd throw a runtime exception when we attempted to load the class containing the annotation.  So I don't think we'd ever have any unexpected values coming in via the annotation, but rather we might get values coming in via the deployment descriptor.

KyleAure:
#libby - check headers
#build - personal build

dev/io.openliberty.concurrent.internal/src/io/openliberty/concurrent/internal/processor/ManagedExecutorDefinitionBinding.java
KyleAure: I used the word `Unspecified` since that matches what the injection binding superclass uses when no value was provided.  But in this case we could use something like `Unsupported` to make it clearer why no value was merged.

PR #27373, opened 2024-01-16
Fix InstantOn persistence FAT break 298584

dazavala:
A feature replacement action introduces an incompatible feature into the server used by test `JPATest` of the `io.openliberty.checkpoint_fat_persistence` FAT suite. Modify the repeat rules in the suite to apply only to servers named in the FAT test classes. 

> Stack Dump = java.lang.IllegalArgumentException: Unable to load conflicting versions of features "com.ibm.websphere.appserver.jdbc-4.1" and "com.ibm.websphere.appserver.jdbc-4.2". The feature dependency chains that led to the conflict are: com.ibm.websphere.appserver.jdbc-4.1 and com.ibm.websphere.appserver.jpa-2.2 -> com.ibm.websphere.appserver.jdbc-4.2


dazavala:
#build

tjwatson (approved):

PR #27380, opened 2024-01-16
virtual true on managed thread factory

njr-11:
Configure virtual=true on ManagedThreadFactory definition and on managed-thread-factory in web.xml.


njr-11:
#libby

KyleAure (approved):

njr-11:
#build

PR #27393, opened 2024-01-17
Fix BundleFactory usage of MessageDigest

### LINKED ISSUE #27421: Resource adapter install fails due to ArrayIndexOutOfBoundsException
**Describe the bug**  
During startup the server logs error `J2CA7002E` indicating an ArrayIndexOutOfBoundsException occurred while installing a resource adapter. The error is caused by a regression in the server's BundleFactory internal which may fail with exception whenever the feature runtime starts multiple resource adapter modules on concurrent threads. This regression may impact server configurations that enable the Jakarta Connectors feature and deploy two or more resource adapters. 

For example: 

`J2CA7002E: An exception occurred while installing the resource adapter <your-resource-adapter>. The exception message is: com.ibm.ws.container.service.state.StateChangeException: java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 98 out of bounds for byte[64]`

The error is preceded by an FFDC incident which shows the causal exception:

`Caused by: java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 98 out of bounds for byte[64]
	at java.base/java.lang.System.arraycopy(Native Method)
	at java.base/sun.security.provider.DigestBase.engineUpdate(DigestBase.java:117)
	at java.base/java.security.MessageDigest$Delegate.engineUpdate(MessageDigest.java:653)
	at java.base/java.security.MessageDigest.update(MessageDigest.java:357)
	at com.ibm.ws.dynamic.bundle.BundleFactory.computeExtrasSha(BundleFactory.java:200)
	at com.ibm.ws.dynamic.bundle.BundleFactory.createBundle(BundleFactory.java:142)
	at com.ibm.ws.jca.internal.ConnectorModuleMetatypeBundleImpl.registerMetatype(ConnectorModuleMetatypeBundleImpl.java:216)
	at com.ibm.ws.jca.internal.ConnectorModuleRuntimeContainer.startModule(ConnectorModuleRuntimeContainer.java:368)
	... 14 more`

**Steps to reproduce**  
1. Configure a Jakarta Connectors feature and two or more resource adapters in server.xml
2. Start the server

**Expected behavior**  
All resource adapters install and start during server startup and remain available over the life of the server.

**Diagnostic information:**  
 - OpenLiberty Version: 21.0.0.4
 - Affected feature(s): jca-1.7, connectors-2.0, connectors-2.1  
 - Java Version: Any supported Java version

```
<server>
     <featureManager>
          ...
          <feature>connectors-2.1</feature>
     </featureManager>

     ... 

     <resourceAdapter name="ra/myRA1 id="RA1" location="${server.config.dir}/connectors/RA1.rar"> 
          ...
     </resourceAdapter>

     <resourceAdapter name="ra/myRA2" id="RA2" location="${server.config.dir}/connectors/RA2.rar">
          ...
     </resourceAdapter>
     ...
</server>

```
**Additional context**

dazavala:
Modify BundleFactory to fix the digest algorithm rather than the digest instance.

Fixes #27421

### DIFF
--- a/dev/com.ibm.ws.dynamic.bundle/src/com/ibm/ws/dynamic/bundle/BundleFactory.java
+++ b/dev/com.ibm.ws.dynamic.bundle/src/com/ibm/ws/dynamic/bundle/BundleFactory.java
@@ -208,19 +208,29 @@ public class BundleFactory extends ManifestFactory {
         }
     }
 
-    private static MessageDigest shaDigest;
+    // The message digest algorithm remains fixed over the server
+    // life cycle and is set during the first call to getShaDigest()
+    private static String shaAlgorithm;
 
     @FFDCIgnore(NoSuchAlgorithmException.class)
     private MessageDigest getShaDigest() throws NoSuchAlgorithmException {
-        if (shaDigest == null) {
+        if (shaAlgorithm == null) {
+            // Set the message digest algorithm
             try {
-                shaDigest = MessageDigest.getInstance(SHA_ALGORITHM);
+                return MessageDigest.getInstance(shaAlgorithm = SHA_ALGORITHM);
             } catch (NoSuchAlgorithmException e) {
-                // Preferred algorithm may be unavailable during server checkpoint
-                shaDigest = MessageDigest.getInstance(SHA_ALGORITHM_DEFAULT);
+                // The preferred algorithm may be unavailable during server checkpoint.
+                // Default to an available algorithm for checkpoint/restore.
+                try {
+                    return MessageDigest.getInstance(shaAlgorithm = SHA_ALGORITHM_DEFAULT);
+                } catch (NoSuchAlgorithmException e2) {
+                    shaAlgorithm = null;
+                    throw e2;
+                }
             }
+        } else {
+            return MessageDigest.getInstance(shaAlgorithm);
         }
-        return shaDigest;
     }
 
     static String getHexSHA(MessageDigest digest) {


dazavala:
#build

dev/com.ibm.ws.dynamic.bundle/src/com/ibm/ws/dynamic/bundle/BundleFactory.java
```diff
@@ -208,19 +208,26 @@ private void computeExtrasSha() {
         }
     }
 
-    private static MessageDigest shaDigest;
+    private static String shaDigestAlgorithm;
```
tevans78: I think some more comments should be added here to explain why the code is crafted in this way. It isn't clear and obvious, particularly since this is partly to workaround an OpenJDK bug.

dazavala: Thanks, @tevans78. I added comments indicating the SHA algorithm remains fixed over the life of the server and is set during the first usage of BundleFactory. During checkpoint the preferred algorithm is not available, so BundleFactory defaults to a known available algorithm. The default algorithm is used in both the checkpoint and restored server processes.

The comment does not address the OpenJDK bug exposed by the build break. The break was caused by a change in the use of MessageDigest which allowed multiple threads to access/share the same MessageDigest instance. The solution here reinstates the original, correct usage.  So technically, we're not working around the JDK bug. Of course, the bug symptom will be addressed in the pertinent support documentation.

dazavala:
#build
Amended initial solution with comments per @tevans78 review.

tjwatson (approved):

dazavala:
#libby

dazavala:
#libby

dazavala:
Pursuant to the ifx delivery process step 2.7, here's the link to the 24001 iFix build: 

https://wasrtc.hursley.ibm.com:9443/jazz/web/projects/WS-CD#action=com.ibm.team.build.viewResult&id=_a9OZ8LutEe64zdT7X42CIQ

PR #27404, opened 2024-01-18
qualifiers on context service and executor definitions

njr-11:
Provide a more correct implementation of support for qualifiers on ContextServiceDefinition, ManagedExecutorDefinition, and ManagedScheduledExecutorDefinition that allows for CDI.current.select as well as injection.


njr-11:
#build

KyleAure (approved):

PR #27422, opened 2024-01-22
Fix Concurrency 3.1 Test Suite

KyleAure:
When attempting to run the `com.ibm.ws.concurrent_fat_jakarta` test suite against Concurrency 3.0 and Java 11 the following will occur:

1. We will compile the concurrency 3.1 bundle using Java 17
2. We will compile our test classes using Java 11
3. At runtime our test infrastructure (running on java 11) will attempt to load classes (such as the servlet class, or packages/classes passed to shrinkwrap).
4. Our test infrastructure will fail to load these classes since we only have the concurrency 3.1 API available and it was compiled at Java 17.

We will need to separate out the concurrency 3.1 tests into their own test bucket to resolve this issue.


dev/com.ibm.ws.concurrent_fat_jakarta_11/fat/src/test/jakarta/concurrency/Concurrency31Test.java
```diff
@@ -33,7 +33,7 @@
 import test.jakarta.concurrency31.web.Concurrency31TestServlet;
 
 @RunWith(FATRunner.class)
-@MinimumJavaLevel(javaLevel = 21)
+@MinimumJavaLevel(javaLevel = 17)
```
njr-11: I don't think some of the tests in this bucket will pass on Java SE 17 without updates.

KyleAure: Yeah, I was testing locally using java 21.  I reverted this change.

KyleAure:
#build

njr-11 (approved):

KyleAure:
#libby - recheck copyright headers

PR #27426, opened 2024-01-22
Increase tolerance in EJB timer FAT

tkburroughs:
- tolerate pauses in test hardware that may result in timer catchup
- tolerate scans of test hardware that may result in unrelated FFDC


tkburroughs:
#build

KyleAure (approved):

PR #27427, opened 2024-01-22
make managed thread factory bean application scoped

njr-11:
Make the ManagedThreadFactory bean be application scoped like the others, and verify that behavior for context capture when using qualifiers matches the behavior for context capture when using `@Resource` which is that context is captured once upon creation, as required by the spec.


njr-11:
#build

njr-11:
#libby
#build

KyleAure (approved):

PR #27430, opened 2024-01-23
Add initial test for multiple connectors with single group.id

abutch3r:
We need to investigate behavior relating to have multiple ingoing connectors that use the same group.id with kafka for different topics to establish a baseline of behavior that is likely to result in rebalancing.

Do not merge - this is currently for testing only


abutch3r:
#build

abutch3r:
#pbbeta

PR #27444, opened 2024-01-24
approach to inherit from abstract superclass

njr-11:
The pull demonstrates the approach to dealing with multiple implementation of incompatible versions of spec API by isolating differences (which are minimal) to an abstract superclass.  The enabled feature selects which superclass to bring in, either the no-op one for Concurrency 3.0 and earlier, or the one for Concurrency 3.1+ which includes a minimal amount of new methods that require a newer Java SE level to implement.


njr-11:
#build

dev/io.openliberty.concurrent.internal.compat31/src/io/openliberty/concurrent/internal/compat/ContextServiceBase.java
```diff
@@ -0,0 +1,37 @@
...
+package io.openliberty.concurrent.internal.compat;
+
+import java.util.concurrent.Flow.Subscriber;
+
+import com.ibm.wsspi.threadcontext.ThreadContextDescriptor;
+
+import io.openliberty.concurrent.internal.compat.impl.ContextualSubscriber;
+
+/**
+ * Superclass for ContextServiceImpl when using Concurrency 3.0 or less.
```
mswatosh: This should be Concurrency 3.1 or later, but can be changed in a later pull

mswatosh (dismissed):

dev/io.openliberty.concurrent.internal.compat31/src/io/openliberty/concurrent/internal/compat/ContextServiceBase.java
njr-11: Good catch -- too much copying and pasting there!

mswatosh (approved):

njr-11:
#build
#libby

Issue #27432, opened 2024-01-23
GRPC MXBeans not deregistered upon application shutdown.


Channyboy:
The MXBeans created in `io.openliberty.grpc.1.0.internal.client.monitor` (i.e., `GrpcClientStatsMXBean`) and `io.openliberty.grpc.1.0.internal.monitor` (i.e., `GrpcServerStatsMXBean`) are not deregistered upon application shutdown.

hutchig:
For background on this see: https://github.com/OpenLiberty/open-liberty/pull/13112

PR #27447, opened 2024-01-24
Remove app bound vendor metrics during application shutdown

### LINKED ISSUE #27439: Remove application bound (Vendor scoped) metrics upon application shut-down.
The mapping of application name -> MetricID within the vendor scoped metric registry is not cleared upon application shut-down.
Furthermore, the use of a Concurrent LinkedList should be replaced with a Set for memory improvements.

Channyboy:
Remove app bound vendor metrics during application shutdown; use set over list for memory savings; temp work-around grpc fat

GRPC issue encountered is that MBeans are not deregistered during application shut down.
Tracked with: https://github.com/OpenLiberty/open-liberty/issues/27432

This PR will cause issue with the `GRPCMetricsTest` as it runs two test on the client server. Upon server config update, this would force an application restart. With this PR, the GRPC metrics would be unloaded from the vendor metric registry. However, since the GRPC MBeans persist from a previous runtime of the application, no "new" Mbeans are created and therefore no new grpc metrics are created. The temprorary work around is to restart the servers for each test in the `GrpcMetricsTest`.

fixes #27439 

#build

### DIFF
--- a/dev/com.ibm.ws.grpc_fat/fat/src/com/ibm/ws/fat/grpc/GrpcMetricsTest.java
+++ b/dev/com.ibm.ws.grpc_fat/fat/src/com/ibm/ws/fat/grpc/GrpcMetricsTest.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2022 IBM Corporation and others.
+ * Copyright (c) 2020, 2024 IBM Corporation and others.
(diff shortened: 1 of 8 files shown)


Channyboy:
#build
#libby

pgunapal (approved):
Metrics portion, looks good.

hutchig (approved):
LGTM

PR #27463, opened 2024-01-25
Adding ContextualSubscriber and ContextualProcessor

mswatosh:
Adding ContextualSubscriber and ContextualProcessor implementations and tests

### DIFF
--- a/dev/io.openliberty.concurrent.internal.compat31/src/io/openliberty/concurrent/internal/compat/impl/ContextualProcessor.java
+++ b/dev/io.openliberty.concurrent.internal.compat31/src/io/openliberty/concurrent/internal/compat/impl/ContextualProcessor.java
@@ -6,41 +6,36 @@
  * http://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *     IBM Corporation - initial API and implementation
  *******************************************************************************/
 package io.openliberty.concurrent.internal.compat.impl;
 
 import java.util.ArrayList;
+import java.util.concurrent.Flow.Processor;
 import java.util.concurrent.Flow.Subscriber;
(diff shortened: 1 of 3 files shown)


dev/io.openliberty.concurrent.internal.compat31/src/io/openliberty/concurrent/internal/compat/impl/ContextualProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public Processor<T, R> getAction() {
+        return action;
+    }
+
+    @Override
+    public ThreadContextDescriptor getContextDescriptor() {
+        return threadContextDescriptor;
+    }
```
njr-11: ContextualSubscriber includes the Trivial annotation to avoid tracing these. We should probably either add the same here, or remove it from ContextualSubscriber to make them consistent.

```suggestion
    @Override
    @Trivial
    public Processor<T, R> getAction() {
        return action;
    }

    @Override
    @Trivial
    public ThreadContextDescriptor getContextDescriptor() {
        return threadContextDescriptor;
    }
```

njr-11 (dismissed):
Looks good. I added one minor comment which is optional.

dev/io.openliberty.concurrent.internal.compat31/src/io/openliberty/concurrent/internal/compat/impl/ContextualProcessor.java
mswatosh: Good catch, `@Trivial` is a good idea for these

njr-11 (dismissed):

mswatosh:
#run-libby-bot

mswatosh:
#build

njr-11 (approved):

PR #27474, opened 2024-01-26
Qualifier attribute comparisons and qualifiers in web deployment descriptor

njr-11:
This pull adds tests for qualifiers defined on the resource definitions for the four types of Concurrency resources in the web.xml deployment descriptor.

It also implements the annotation methods of generated qualifiers (toString, hashCode, equals, annotationType) according to the requirements of the Annotation JavaDoc, and then adds test coverage for each of these methods.

It covers nonbinding qualifier attributes, which do count for annotation equals comparisons, but do not count for resolving beans.


jmstephensgit (approved):

njr-11:
#build

Issue #27377, opened 2024-01-16
Create a common OSGi utils class


benjamin-confino:
I've noticed duplicate code in different microprofile bundles for injecting an OSGi `@Reference` manually. A common class in the kernel could de-duplicate things and ensure everyone's using best practices.

PR #27481, opened 2024-01-28
use service caller for OSGi

benjamin-confino:
Another followup to #27377

This will improve performance and guard against the risk of overrunin the use count in the framework


benjamin-confino:
#build

njr-11 (approved):

arturdzm (approved):

PR #27502, opened 2024-01-30
add more doPriv statements

benjamin-confino:
A follow up to #27377 we found a need for doPriv statements when doing local testing for OpenTelemetry

TODO - does this need a new issue number. Discuss with Tom Watson


benjamin-confino:
#build

tjwatson (changes requested):
I don't agree that this utility should protect callers from the `org.osgi.framework.ServicePermission` `get` operation.  It should only protect the caller from the `AdminPermission` required to access a bundle `CONTEXT` which is already handled by the doPriv in the private method `getContext`.  All other security checks must be propagated to the caller of this utility class.  The caller must decide if it should use `doPriv` or not.  This is the same as any other APIs in OSGi that access the service registry to `get` services.

PR #27504, opened 2024-01-30
Add doPriv to calls to ServiceCaller

benjamin-confino:
Followup to #27377 adds some needed doPriv calls

### DIFF
--- a/dev/com.ibm.ws.kernel.service/src/com/ibm/ws/kernel/service/util/ServiceCaller.java
+++ b/dev/com.ibm.ws.kernel.service/src/com/ibm/ws/kernel/service/util/ServiceCaller.java
@@ -36,6 +36,7 @@ import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.framework.ServiceEvent;
 import org.osgi.framework.ServiceListener;
+import org.osgi.framework.ServicePermission;
 import org.osgi.framework.ServiceReference;
 import org.osgi.framework.SynchronousBundleListener;
 import org.osgi.util.tracker.ServiceTracker;
@@ -118,6 +119,7 @@ import org.osgi.util.tracker.ServiceTracker;
  * @since 3.13
  */
 public class ServiceCaller<S> {
+
     /**
      * Calls an OSGi service by dynamically looking it up and passing it to the
      * given consumer.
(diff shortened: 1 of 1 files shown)


benjamin-confino:
#build

dev/com.ibm.ws.kernel.service/src/com/ibm/ws/kernel/service/util/ServiceCaller.java
```diff
@@ -371,6 +383,8 @@ public ServiceCaller(Class<?> caller, Class<S> serviceType, String filter) {
         this.serviceType = Objects.requireNonNull(serviceType);
         this.bundle = Optional.of(Objects.requireNonNull(caller)).map(FrameworkUtil::getBundle).orElseThrow(IllegalStateException::new);
         this.filter = filter;
+        this.servicePermission = new ServicePermission(serviceType.getSimpleName(), ServicePermission.GET);
```
tjwatson: `getSimpleName` does not include the package name which is required to do the `ServicePermission` check.  The method `serviceType.getName()` should be used here.

dev/com.ibm.ws.kernel.service/src/com/ibm/ws/kernel/service/util/ServiceCaller.java
```diff
@@ -494,5 +520,8 @@ public void unget() {
             current.unget();
         }
     }
-}
 
+    private boolean bundleHasPermission(BundleContext bc, Object permission) {
+        return bc.getBundle().hasPermission(permission);
```
tjwatson: we already have the `bundle` field and we should use that to avoid exceptions when the bundle is stopped and the BundleContext starts throwing IllegalStateExceptions.

dev/com.ibm.ws.kernel.service/src/com/ibm/ws/kernel/service/util/ServiceCaller.java
```diff
@@ -462,6 +476,17 @@ private Optional<ReferenceAndService> trackCurrent() {
 
     private Optional<ReferenceAndService> getCurrent() {
         BundleContext context = getContext();
+        if (System.getSecurityManager() != null && bundleHasPermission(context, servicePermission)) {
```
tjwatson: would be more clear if this was

```java
        if (System.getSecurityManager() != null && bundle.hasPermission(servicePermission)) {
```

dev/com.ibm.ws.kernel.service/src/com/ibm/ws/kernel/service/util/ServiceCaller.java
```diff
@@ -462,6 +476,17 @@ private Optional<ReferenceAndService> trackCurrent() {
...
+            return AccessController.doPrivileged(
+                                                 (PrivilegedAction<Optional<ReferenceAndService>>) () -> {
+                                                     return getCurrentInternal(context);
+                                                 });
+        } else {
+            return getCurrentInternal(context);
```
tjwatson, line 486: I think this is fine because it should end up throwing a `SecurityException` if the security manager is not `null` and the calling bundle did not have permissions above.  Maybe add a comment if that was the intention here.

benjamin-confino: That was the intention.

tjwatson (approved):

benjamin-confino:
#build

benjamin-confino:
#libby

PR #27515, opened 2024-01-31
Add Java 22 specific test to FAT

gjwatts:
Add a Java 22 specific test to the `io.openliberty.java.internal_fat` FAT to ensure we are running on Java 22.

Note: The source files in this PR under `io.openliberty.java.internal_fat/src-reference/...` are not actually used to build and run the test since they need to be compiled at Java 22.  They are just convenient copies of the source for the test which is kept here -> https://github.com/OpenLiberty/open-liberty-misc/tree/main/io.openliberty.java.internal_fat_22


gjwatts:
#build

gjwatts:
#libby

KyleAure (approved):

gjwatts:
#libby

PR #27524, opened 2024-01-31
CDI 4.1 repeats for visibility bucket

tevans78:
Repeat the CDI Visibility bucket against EE11.
Also included a small improvement to the code which runs the transformer, to ensure that failures result in an exception ... which in turn means that the logs are kept.


tevans78:
#build

benjamin-confino (approved):

tevans78:
Failures were not in the CDI buckets

PR #27531, opened 2024-01-31
Avoid parse error when unmarshalling server.xml

KyleAure:
After recent update to the JAXB implementation used in fattest.simplicity https://github.com/OpenLiberty/open-liberty/pull/26129
the com.ibm.ws.concurrent.persistent.fat.failovertimers.FailoverTimersTest has started to exploit a bug in the java.xml.xpath package which fails to handle concurrent parsing.
https://bugs.openjdk.org/browse/JDK-8047329

Long term fix is to ensure unmarshalling is synchronized.


KyleAure:
#build

mswatosh (approved):

KyleAure:
#libby re-check headers
Both personal builds failed with known defects unrelated to this change.

PR #27539, opened 2024-02-01
Utilise new kafkaStopServer method in RM Kafka FATs

### LINKED ISSUE #27537: Update RM Kafka tests to use new kafkaStopServer method
As part of work on Kafka, we found that the server might contain Kafka errors that are not picked up on Server stop

As such `kafkaServerStop(Server server)` was added to the KafkaUtils class

We need to go through the kafka related tests and where we do not expect errors to occur replace the standard stopServer() call

ryan-storey:
Closes #27537


ryan-storey:
#build

abutch3r (approved):
LGTM - as long as these tests are green in the build then good to merge

wraschke:
#libby

PR #27548, opened 2024-02-02
prevent double contextualization of processor via contextualSubscriber

njr-11:
There is checking in place to prevent contextualSubscriber from doubly contextualizing a Subscriber, but not to guard against doubly contextualizing a Processor that is supplied to the method. This is possible because Processor inherits from Subscriber.  This pull prevents this.


njr-11:
#build

gjwatts (approved):

njr-11:
#build

njr-11:
Build has 1 test failure, which is a duplicate of a pre-existing issue in unrelated tests.

PR #27577, opened 2024-02-05
Assume.assumeTrue exception caused test failure

zhang-ca:
The following exception seems caused AIX and Java 17 test failure:

`[01/26/2024 16:04:46:988 EST] 001 FATRunner                      evaluate                       I assumption violated: org.junit.internal.AssumptionViolatedException: got: <false>, expected: is <true>`


zhang-ca:
#build

karel-harjono (approved):

PR #27583, opened 2024-02-06
Remove test that times out because it legitimately needs longer than …

benjamin-confino:
…the timeout to run

### DIFF
--- a/dev/com.ibm.ws.microprofile.faulttolerance.2.0_fat_tck/publish/tckRunner/tck/tck-suite-lite.xml
+++ b/dev/com.ibm.ws.microprofile.faulttolerance.2.0_fat_tck/publish/tckRunner/tck/tck-suite-lite.xml
@@ -23,11 +23,6 @@
                     <include name="testAsyncTimeout"/>
                 </methods>
             </class>
-            <class name="org.eclipse.microprofile.fault.tolerance.tck.CircuitBreakerBulkheadTest">
-                <methods>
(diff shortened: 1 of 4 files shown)


Azquelt (dismissed):

benjamin-confino:
#build

dev/com.ibm.ws.microprofile.faulttolerance.2.0_fat_tck/publish/tckRunner/tck/tck-suite-lite.xml
```diff
@@ -23,11 +23,13 @@
...
                 </methods>
             </class>
+            <!--
+            This was removed because the test could legitimately take longer than the 2second timeout. This bug is fixed in TCK version 3
             <class name="org.eclipse.microprofile.fault.tolerance.tck.CircuitBreakerBulkheadTest">
                 <methods>
                     <include name="testCircuitBreakerAroundBulkheadSync"/>
                 </methods>
-            </class>
+            </class> -->
```
Azquelt: Actually, I don't think we need to document this.

It's sufficient that we document it being excluded in the full suite XML and here it adds clutter, making it harder to check what is included.

Azquelt (approved):

benjamin-confino:
The previous build hasn't started yet so it will have the latest change. (Not that we really need a build)

benjamin-confino:
#libby

PR #27618, opened 2024-02-09
add the find annotation

njr-11:
Add, implement, and test the new Find annotation from Jakarta Data.
This pull also includes a commit from another pull that this depends on which is unfortunately blocked from merging because the stream is locked down.


njr-11:
#build

jmstephensgit (approved):

njr-11:
Build and 1 test failure which is a duplicate of a pre-existing issue

PR #27619, opened 2024-02-10
Fix JPA Spec 10 FAT with JPA 32

tkburroughs:
JPA FAT is incorrectly running Jakarta EE 11 repeat for Java 8 and 11
- delete duplicated repeat actions that caused JPA 31 to use JPA 30
- remove isEnabled override that resulted in EE 11 FAT with Java 8 and 11
- added fullFATOnly support to JPA repeat actions and run some in Lite mode


tkburroughs:
#build

tkburroughs:
#build

KyleAure (approved):

PR #27648, opened 2024-02-13
qualifiers on context service in application deployment descriptor

njr-11:
Add the ability to handle qualifier class names that are specified on a context-service definition in application.xml. We need to be able to load these from the application, while still being able to load qualifier classes from web.xml out of the web module. The injection engine already has the correct class loader, and this pull adds code to get it from there and centralizes loading the qualifier classes into a single place.  This was sufficient to get a single test that is added under this pull working.  However, splitting the test application into both web and ejb modules exposed a separate problem where the CDI extension cannot rely on any particular module's thread context being present.  For now, I commented out tests of managed thread factory that are intermittently hitting that issue, with TODO comments where that needs to be addressed, which will be covered in a separate pull.  It should also be noted that more tests need to be written for other 3 resource types and so forth. Those will be covered under another pull as well.  This pull is limited to being a proof of concept of the general solution and only tests out the context-service definition.

### DIFF
--- a/dev/com.ibm.ws.concurrent_fat_cdi/bnd.bnd
+++ b/dev/com.ibm.ws.concurrent_fat_cdi/bnd.bnd
@@ -31,5 +31,6 @@ fat.project: true
 	io.openliberty.jakarta.annotation.2.1;version=latest,\
 	io.openliberty.jakarta.cdi.4.1;version=latest,\
 	io.openliberty.jakarta.concurrency.3.1;version=latest,\
+	io.openliberty.jakarta.enterpriseBeans.4.0;version=latest,\
 	io.openliberty.jakarta.servlet.6.1;version=latest,\
 	io.openliberty.jakarta.transaction.2.0;version=latest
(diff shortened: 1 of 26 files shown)


njr-11:
#build

njr-11:
#build

tkburroughs (approved):
The InjectionEngine changes look good to me.

Perhaps the next best alternative would be to just change the thread context classloader, but that could very well result in errors; other resources are likely not expecting an application specific classloader.

Does the use of the "declaringClassloader" take into consideration what it means if the resource is defined in the `java:global` name space?  Things in `java:global` may be used by any installed application, which could be problematic (and potentially a memory leak) if the "declaringClassloader" application is uninstalled/removed.

Despite the question about `java:global`; the actual changes to InjectionEngine appear to be the best (and safest) way to provide the application classloader to the resource factory.

njr-11:
> Does the use of the "declaringClassloader" take into consideration what it means if the resource is defined in the `java:global` name space? Things in `java:global` may be used by any installed application, which could be problematic (and potentially a memory leak) if the "declaringClassloader" application is uninstalled/removed.

This is an excellent question. The approach we have taken in the Concurrency specification is to require that `java:global` names must not be used on resource definitions that specify any qualifiers. (The class loader is used to load qualifier annotation classes)

PR #27653, opened 2024-02-14
Increase FeatureList generation timeout

mswatosh:
The Feature List generation timeout is being hit in builds. When builds are quick it won't make a difference, but it they're running slow this gives some additional overhead.


mswatosh:
#build

mswatosh:
The one failure was caused by a 503 response from an external resource, unrelated to this change.

KyleAure (approved):

PR #27654, opened 2024-02-14
server command not handle space in path on Windows

### LINKED ISSUE #27652: Windows: server command doesn't handle space in path unless JAVA_HOME set
**Describe the bug**  
On Windows, the server script cannot find java.exe when java.exe resides in a directory path containing a space; for example:
`C:\Program Files\IBM\java8\bin\java.exe`

**Steps to Reproduce**  
Install java in a directory location which contains a space in the path.  Run any server command which invokes java.exe; e.g. server list

**Expected behavior**  
The server list command should list the servers.  Instead it says
```
The system cannot find the batch label specified - eof
'" \java"' is not recognized as an internal or external command,
operable program or batch file.
" \java"
```

**Diagnostic information:**  
 - OpenLiberty Version: [e.g. 23.0.0.10-24.0.0.2]

**Additional context**  
Works ok if JAVA_HOME is set.

jimblye:
Fixes #27652

### DIFF
--- a/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server.bat
+++ b/dev/com.ibm.ws.kernel.boot.ws-server/publish/bin/server.bat
@@ -1,7 +1,7 @@
 @echo off
 @REM WebSphere Application Server liberty launch script
 @REM
-@REM Copyright (c) 2011, 2023 IBM Corporation and others.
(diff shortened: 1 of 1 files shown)


jimblye:
#build

hlhoots (approved):
LGTM

PR #27658, opened 2024-02-15
Make sure that all JWT Consumer test classes wait for the jwtconsumerclient app to start

c00crane:
The tests in the com.ibm.ws.security.jwt_fat.consumer need to make sure that the jwtconsumerclient test app is ready for requests.  3 of the test classes weren't waiting.  The app usually came up fast enough that there wasn't an issue when the first test made a request.  But, we finally had an instance where the app wasn't ready.
I'm adding the app to the list of applications that the server startup should wait for.


c00crane:
#build
#spawn.fullfat.buckets=com.ibm.ws.security.jwt_fat.consumer

jimmy1wu (approved):

PR #27357, opened 2024-01-13
Fix JavaIllegalAccessTest dns test values

gjwatts:
Fixes the problem where in our testing we see this error:
```
Test Failure: io.openliberty.java.internal.fat.JavaIllegalAccessTest.IllegalAccessTestServlet.testJdkNamingDnsDoesNotExportComSunJndiUurlDns

IllegalAccessTestServlet.testJdkNamingDnsDoesNotExportComSunJndiUurlDns:junit.framework.AssertionFailedError: 2023-08-05-08:12:27:454 ERROR: Caught exception attempting to call test method testJdkNamingDnsDoesNotExportComSunJndiUurlDns on servlet test.web.IllegalAccessTestServlet
javax.naming.CommunicationException: DNS error [Root exception is java.net.SocketTimeoutException: Receive timed out]; remaining name '[www.github.com](http://www.github.com/)'
at com.sun.jndi.dns.DnsClient.query(DnsClient.java:321)
at com.sun.jndi.dns.Resolver.query(Resolver.java:81)
at com.sun.jndi.dns.DnsContext.c_getAttributes(DnsContext.java:434)
at java.naming/com.sun.jndi.toolkit.ctx.ComponentDirContext.p_getAttributes(ComponentDirContext.java:235)
at java.naming/com.sun.jndi.toolkit.ctx.PartialCompositeDirContext.getAttributes(PartialCompositeDirContext.java:141)
at java.naming/com.sun.jndi.toolkit.url.GenericURLDirContext.getAttributes(GenericURLDirContext.java:103)
at org.apache.aries.jndi.DelegateContext.getAttributes(DelegateContext.java:263)
...
```

Originally, the DNS host name value to be resolved was `www.github.com` and that would fail sometimes (servers the tests ran on could not resolve internet host names?).  Rather than rely on name resolution for external servers, now the test DNS value to be resolved is the local host name of the test server.

PR #27671, opened 2024-02-16
Fix CommunicationException in JavaIllegalAccessTest

gjwatts:
This is a continuation of fix #27357.

Previously I attempted a DNS lookup of the local host name using the underlying machine's DNS config, but intermittently, a `javax.naming.CommunicationException` is thrown when trying the DNS lookup.  This exception is permissible since the purpose of this test is to attempt the DNS connection to ensure the illegalAccessException FFDC is not created, not to ensure the DNS resolution is successful.

So added `CommunicationException` to the catch block to ignore any of those exceptions.


njr-11 (approved):

gjwatts:
#build

PR #27672, opened 2024-02-16
apply component metadata from ejb or create web component metadata

njr-11:
This pull gets a little further. It doesn't address resource definitions from application.xml (which will only have application metadata) yet, but it does identify component metadata and in the case of web, module metadata from the application artifact that defines the managed thread factory definition.  In the case of component metadata, it's already what we want and we can just us it.  In the case of web module metadata, we can use the web containers own deferred metadata factory to create component metadata for web based on the module metadata.  This is preferable to creating our own component metadata in these cases where it is possible.

A separate pull will cover the places where it is not possible (managed thread factories defined in application.xml and a CDI bean for a per-application default instance) which is where we will need to create dummy component metadata, which should be more appropriate for that path because it is only representing the application, not any of the modules in particular.

### DIFF
--- a/dev/com.ibm.ws.concurrent/bnd.bnd
+++ b/dev/com.ibm.ws.concurrent/bnd.bnd
@@ -23,6 +23,7 @@ IBM-Default-Config: OSGI-INF/wlp/defaultInstances.xml
 
 Export-Package:\
   com.ibm.ws.concurrent,\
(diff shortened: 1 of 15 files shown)


njr-11:
#build

jmstephensgit (approved):

PR #27723, opened 2024-02-22
Revert "Reintroduce Container Auth Support for TX DB Recovery logs"

neilgsyoung:
Reverts OpenLiberty/open-liberty#27605

Reverting because of product regression identified by InstantOn team.


dazavala (approved):
Thank you!

